PR #94, opened 2020-02-08
feat(implementations/c): add basic transport

robinbudd:
Implement posix socket, single-threaded TCP transport library and test

Thank you for sending a pull request :heart:

### Proposed Changes
Please describe the changes in your pull request.

If this pull request resolves an already recorded bug or a feature request, be sure to link to that issue.


mrinalwadhwa:
@robinbudd some files in the PR have tabs as indents, while the rest of the C code is using spaces per the editorconfig.

Here's a easy tool to lint for compliance with the editorconfig:

With the `builder-debian` running:
```
vagrant ssh builder-debian -c "npx eclint@2.8.1 check implementations/c/source/ockam/transport"
```

If you have node on your host machine, you can also invoke npx directly:

```
npx eclint@2.8.1 check implementations/c/source/ockam/transport
```

implementations/c/include/ockam/transport.h
```diff
@@ -0,0 +1,154 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netinet/tcp.h>
+#include <arpa/inet.h>
+#include <netinet/in.h>
+#include <netdb.h>
+#include <errno.h>
+#include <time.h>
```
mrinalwadhwa: These includes look like they tightly couple the transport interface to a platform. 
Which ones here do we really need in `transport.h`?

`stdlib.h` and `error.h` make sense, although we could debate if we need to decouple from `stdlib.h`.

implementations/c/source/ockam/transport/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+cmake_minimum_required(VERSION 3.13)
```
mrinalwadhwa: other cmake files in the PR set min version to `3.14`, lets make that consistent.

implementations/c/test/ockam/transport/client/ipaddress.txt
```diff
@@ -0,0 +1,2 @@
+192.168.0.77
+8000
```
mrinalwadhwa: Two things about this file:

1. since this file is committed to the code repo, keeping it separate from the test code doesn't seem very useful. It could instead be variables in the test code with default values that can be overridden with CLI arguments.

2. Was there a reason to chose `192.168.0.77` as a default value? I fear this test, if automated, will fail if the machine running the test doesn't have this subnet defined `192.168.*.*` ... could `127.0.0.1` or `0.0.0.0` be better defaults?

implementations/c/test/ockam/transport/server/test_data_client.txt
```diff
@@ -0,0 +1,358 @@
+# CMAKE generated file: DO NOT EDIT!
```
mrinalwadhwa: about this and other test data files:

1. could we replace them with output from `/dev/urandom` written to an *uncommitted temp files*?
2. if there is a good reason we can identify to commit these files then I think it would be better to use data that looks random (lorem impsum or random hex) .. the current log output may confuse casual readers / beginners grep-ing in the code base.

robinbudd: Agree about using a file with generic content. But, I'm not sure how to use /dev/urandom as a test case, when the client and server might be on different machines. For now, I'll change the content as you suggest.

implementations/c/test/ockam/transport/client/test_data_client.txt
```diff
@@ -0,0 +1,358 @@
+# CMAKE generated file: DO NOT EDIT!
```
mtm0183: Does this file need to be committed?

implementations/c/test/ockam/transport/client/test_data_compare.txt
```diff
@@ -0,0 +1,358 @@
+# CMAKE generated file: DO NOT EDIT!
```
mtm0183: Does this file need to be committed?

implementations/c/test/ockam/transport/client/test_data_server.txt
```diff
@@ -0,0 +1,358 @@
+# CMAKE generated file: DO NOT EDIT!
```
mtm0183: Does this file need to be committed?

implementations/c/test/ockam/transport/server/test_data_compare.txt
```diff
@@ -0,0 +1,358 @@
+# CMAKE generated file: DO NOT EDIT!
```
mtm0183: Does this file need to be committed?

implementations/c/test/ockam/transport/server/test_data_server.txt
```diff
@@ -0,0 +1,358 @@
+# CMAKE generated file: DO NOT EDIT!
```
mtm0183: Does this file need to be committed?

implementations/c/test/ockam/transport/client/test_data_client.txt
robinbudd: I've addressed that too

implementations/c/include/ockam/transport.h
robinbudd: I moved all but stdlib.h, string.h, and stdint.h into the posix-specific include file in the transport source tree

implementations/c/test/ockam/transport/client/ipaddress.txt
robinbudd: If there are command line args I use those, otherwise I default to loopback address 127.0.0.1

implementations/c/test/ockam/transport/server/test_data_client.txt
robinbudd: I created a test file with "lorem ipsum"...etc. For now it lives in the source directory. At some point we may want to have a strategy for where such files live.

implementations/c/include/ockam/syslog.h
```diff
@@ -0,0 +1,6 @@
+#include <stdio.h>
+#include "error.h"
+
+void init_err_log(FILE* fp);
+
+void log_error(OCKAM_ERR error, char* message);
```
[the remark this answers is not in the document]
robinbudd: Leaving as-is for now, until we agree on how we will handle logging.

implementations/c/source/ockam/transport/connection/socket/posix/posix_tcp_connection.c
```diff
@@ -0,0 +1,385 @@
...
+ *                                             INCLUDE FILES                                            *
+ ********************************************************************************************************
+ */
+
+#include <unistd.h>
+#include "ockam/syslog.h"
+#include "ockam/error.h"
+#include "ockam/transport.h"
+#include "../../../include/connection.h"
+#include "../../../include/posix_socket.h"
```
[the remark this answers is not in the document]
robinbudd: Done.

implementations/c/source/ockam/transport/include/server.h
```diff
@@ -0,0 +1,43 @@
+//
+// Created by Robin Budd on 191220.
+//
+
```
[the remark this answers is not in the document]
robinbudd: Done and done

implementations/c/source/ockam/transport/connection/socket/posix/posix_socket.c
```diff
@@ -0,0 +1,40 @@
+#include <stdlib.h>
```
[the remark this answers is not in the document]
robinbudd: Done

PR #100, opened 2020-02-10
Support building with Docker

bitwalker:
I'll update this with more details later, but I wanted to push this PR up in case Mark wants to try it to work around the build issues he is experiencing.

## Summary

This changeset enables Docker as another provider for Vagrant, using the same init scripts we use for the VirtualBox VM (with some scripts skipped as they aren't applicable). You can specify to use Docker as your provider rather than VirtualBox by exporting `VAGRANT_DEFAULT_PROVIDER=docker` before running any Vagrant commands. The included `Makefile` simplifies this by providing a couple psuedo-targets which run `gradlew build` with this environment variable set to either `docker` or `virtualbox` respectively, i.e. `make docker` and `make virtualbox`.

The Docker image used must be built the first time it is used (or fetched from an image repository, but we currently do not have one set up). You can run this build with `make builder`. This command takes some time to complete, since it has to install all of our dependencies, including a few which are built from source. I believe it took my machine ~15 minutes. Once built, you do not need to build it again unless you change dependencies, in which case you run `make builder` again, and it will rebuild the image, using cached layers where possible.

## Benefits

The Docker build itself runs significantly faster than the VirtualBox build. A significant reason for this is that `rsync` is not used or needed. A complete build runs in under a minute for me currently.

Another benefit is that our CI environment could use the same base image for builds, rather than spin up VMs for that purpose.

Lastly, it should help with community involvement, since it provides a simpler option for building without needing to install VirtualBox and set it up. Docker tends to be more commonly used for this in open source projects than VirtualBox, but having both means that one can choose whatever is easiest or preferred.

## Non-goals

- This does not replace VirtualBox in the build.
- This does not provide a solution for building against embedded hardware targets (unless we set up a cross-compiler toolchain, which is already done for Rust w/ ARMv7, but not for the C code at this time).

## Future Work

- Set up an image repository (probably on Docker Hub) and an automated builder that constructs the Docker image when changes are made to the scripts used to build it, i.e. when we update dependencies.
- Maybe provide a cross-compilation image for building against non-x86_64 targets? If we could have a build matrix of images that cover all of our targets, that would be super helpful for automating releases. Unfortunately we would likely always need a seperate set of builders that handle running tests against real hardware - but we may be able to cover some targets with QEmu or similar emulation strategies for regular builds.


tools/gradle/plugins/builder/src/main/groovy/network/ockam/gradle/builders/BuildersPlugin.groovy
```diff
@@ -68,8 +68,8 @@ class BuildersPlugin implements Plugin<Project> {
   }
 
   // starting from the current directory, look for a parent directory that has a Vagrantfile
-  static String findVagrantfileDir() {
-    def path = new File('.')
+  static String findVagrantfileDir(File currentDir) {
+    def path = new File(currentDir.getAbsolutePath())
```
mrinalwadhwa: this was fixed in develop https://github.com/ockam-network/ockam/commit/a16cda2ee7709922d5d557a8722c8a7ce1143335

tools/builder/debian/Dockerfile
```diff
@@ -0,0 +1,155 @@
+# Stage 1: Create the base image with all necessary components
+FROM debian:9 as vagrant-base
```
mrinalwadhwa: `debian:9` is not deterministic, we should pin the hash of the image

tools/builder/debian/Dockerfile
```diff
@@ -0,0 +1,155 @@
...
+        systemd-sysv \
+        cron         \
+        anacron \
+        curl \
+        wget \
+        build-essential \
+        gnupg \
+        openssh-server \
+        openssh-client && \
+    apt-get clean
```
mrinalwadhwa: apt-get also non determinism, we should bake the needed packages into the base image

tools/builder/debian/provision/rust.sh
```diff
@@ -1,15 +1,25 @@
...
 export CARGO_HOME=/opt/rust/cargo
+export RUSTUP_HOME=/opt/rust/rustup
 export PATH=$CARGO_HOME/bin:$PATH
-chmod +x rustup-init
-./rustup-init -y --no-modify-path --default-toolchain "1.35.0"
-rm rustup-init
 
-echo 'export CARGO_HOME=/vagrant/.builder/cargo' > /etc/profile.d/rust.sh
-echo 'export PATH="$CARGO_HOME/bin:/opt/rust/cargo/bin:$PATH"' >> /etc/profile.d/rust.sh
+curl --proto '=https' --tlsv1.2 -sSf 'https://sh.rustup.rs' > /tmp/rustup-init
```
mrinalwadhwa: we would want to pin the hash of the download.

Vagrantfile
```diff
@@ -16,15 +16,27 @@ Vagrant.configure("2") do |config|
...
+    config.vm.provider :docker do |docker, override|
+      override.vm.box = nil
+      override.ssh.insert_key = true
+      override.vm.synced_folder ".", "/vagrant", docker_consistency: "cached"
+
+      docker.image = "ockam-builder-debian-base:latest"
+      docker.name = "builder-debian"
+      docker.remains_running = true
+      docker.has_ssh = true
+      docker.create_args = ['--cap-add', 'SYS_ADMIN', '--tmpfs', '/tmp:exec', '--tmpfs', '/run', '-v', '/sys/fs/cgroup:/sys/fs/cgroup:ro']
```
mrinalwadhwa: does this really need SYS_ADMIN on the docker host?

tools/builder/debian/Dockerfile
bitwalker: This is the base image

tools/builder/debian/Dockerfile
bitwalker: The image we build from this is what matters, we'd pin the hash of the resulting image in our CI and tag it, then build against that, and point our tooling to the tagged image.

Vagrantfile
bitwalker: Yes, it's required for filesystem mounts

bitwalker: To be clear, the mounts are necessary to allow `systemd` to run in the container, which is how we can run an SSH daemon and treat it like a virtual machine, rather than a single-process container

tools/builder/debian/provision/rust.sh
bitwalker: That doesn't make much sense for the `rustup-init` binary, it is an installer, it's not pinned to a version of Rust, and will always fetch the latest from the given channel (in our case, stable), which is by design.

bitwalker:
@mrinalwadhwa This is ready to merge now, commits have been rebased on develop, Makefile tasks I felt would be useful have been moved into the project-level `build.gradle`. #110 can be merged right after this. #108 needs a new VM built, so we can update the hash, then that will be ready also.

mrinalwadhwa:
@bitwalker could you amend the first 5 commits that don't follow the commit message convention or squash to one commit.

bitwalker:
@mrinalwadhwa Done!

mrinalwadhwa:
@bitwalker you missed a few in the very beginning.
Screen Shot 2020-03-18 at 6 37 11 PM

mrinalwadhwa:
@bitwalker gradle doesn't seem to like this line before the call to `plugins`
https://github.com/ockam-network/ockam/pull/100/files#diff-c197962302397baf3a4cc36463dce5eaR1

https://dev.azure.com/ockam-network/ockam/_build/results?buildId=835&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=5caf77c8-9b10-50ef-b5c7-ca89c63e1c86&l=24

Also, there are a couple of capital letters in the commit subject lines.

PR #101, opened 2020-02-10
feat(implementations/c): Add CMocka unit test framework to Vault tests

mtm0183:
Add CMocka as a submodule
Modify Azure Pipelines to pick up the resulting XML files
Remove custom compare and print functions for CMocka asserts

Addresses Issue #89

PR #102, opened 2020-02-12
feat(implementations/c): Add Infineon Optiga Trust X

mtm0183:
Add support for the following operations:
Random Number Generation
NIST P-256 Key Pair Generation
ECDH
SHA-256


mrinalwadhwa:
When we merge this ... 53ddf20 would need to be handled with case since its also present in #101 and this pr branch is not rebased to the branch in #101

mrinalwadhwa:
merged https://github.com/ockam-network/ockam/commit/48713d4c17528782daec805b79afff1958227511

Issue #67, opened 2020-01-07
Implement XX key agreement protocol


mrinalwadhwa:
https://github.com/ockam-network/proposals/issues/14

robinbudd:
![](https://github.trello.services/images/mini-trello-icon.png) [Implement XX key agreement protocol in C](https://trello.com/c/ydR0GERL/8-implement-xx-key-agreement-protocol-in-c)

PR #105, opened 2020-02-14
feat(implementations/c): implement xx key exchange, test, and demo

robinbudd:
Fixes: #67 

#### Proposed Changes
Implement the xx key exchange functions.
Test with known test set data to verify correctness.
Interactive demo code that encrypts and decrypts text entered from console, sending it across a secure channel.


implementations/c/include/ockam/handshake.h
```diff
@@ -0,0 +1,124 @@
...
+ ********************************************************************************************************
+ */
+
+/**
+ * XX_HANDSHAKE - the handshake structure is passed to all handshake functions.
+ */
+typedef struct  {
+
+    uint64_t    nonce;
+    uint8_t     s[KEY_SIZE];
```
bitwalker: Copying my comment from the previous iteration: These field names need documentation or descriptive text rather than opaque variable names. If there is a document that describes their meaning and use already, linking to that works fine

mrinalwadhwa: @bitwalker good point, the details are here:
https://github.com/ockam-network/proposals/issues/14#issuecomment-570505343
https://github.com/ockam-network/proposals/tree/master/design/0003-key-agreement-xx

bitwalker: Perfect, copying that description into a comment on the struct would be fine, though in general I think it’s best to make the names themselves reflect their purpose as the code ends up being a lot easier to read. Within the context of a single function, abbreviations are easy to hold in your head, but across larger contexts it’s too easy to lose track and potentially make mistakes as a result.

robinbudd: @bitwalker I ordinarily agree. In this case however, the code is arcane enough that anyone who is going to muck around in should read the paper, at which point the variable name make sense. I expect we'll want to move the structure out of the public-facing header file though, and implement a handle-type interface instead of passing around an explicit function pointer. I'll open an issue for that, but will leave as-is for now unless someone strongly objects (@mrinalwadhwa ).

mrinalwadhwa: I agree, this will go through some iteration soon.

robinbudd:
[Implement XX key agreement protocol in C](https://trello.com/c/ydR0GERL/8-implement-xx-key-agreement-protocol-in-c)

implementations/c/include/ockam/handshake.h
bitwalker: I mean, the reason I posted this comment originally was due to reading the code and being unable to infer the meaning of those fields, both because there was no comment linking to a paper, and because the surrounding code gave no context, only the few scattered comments did that. I don't think this obfuscated form is useful whether the people writing the code know the paper or not - it's just an extra layer of indirection to get to the actual purpose of those variables/fields.

In any case, I won't fight that hard for this specific case, but I'd strongly recommend reconsidering.

PR #110, opened 2020-03-18
build: add raspberry pi tools to debian builder

mtm0183:
Currently the raspberry pi compiler and sysroot are a git submodule
being pulled in at the start of every build. This is adding 5+ minutes
to the start of every build. This script will install the raspberry pi
tools into the debian builder so the download does not take place during
every build.

### DIFF
--- a/tools/builder/debian/provision/arm-rpi.sh
+++ b/tools/builder/debian/provision/arm-rpi.sh
@@ -1,0 +1,13 @@
+#!/bin/sh -eux
+
+cd /tmp
+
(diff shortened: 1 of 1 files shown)


mrinalwadhwa:
@mtm0183 thanks, this looks great.
Is there a quick test we could add to this file to check that it worked?
https://github.com/ockam-network/ockam/blob/de47a042538f22e44dad9d230f523f726df26d79/tools/builder/debian/tests/test.bats

mtm0183:
I ran it in my Debian VirtualBox image to verify it worked. You should be able to drop it in as well and run it to verify it works. That seemed like the easiest way to test it.

mrinalwadhwa:
merged https://github.com/ockam-network/ockam/commit/27df03d730f037d414207fe7f73df303b8ece885

PR #112, opened 2020-03-20
refactor: implement new directory, interface, build structures c code

robinbudd:
Thank you for sending a pull request :heart:

### Proposed Changes
implement new directory structure for implementations/c

replace gradle-based build system with CMake (invoked by gradle)

refactor interface for vault and transport

rename handshake project to key_agreement_xx


implementations/c/include/ockam/error.h
```diff
@@ -1,103 +1,30 @@
...
-
-    OCKAM_ERR_XX_HANDSHAKE_TEST_FAILED                = 0x0601,
-    OCKAM_ERR_XX_HANDSHAKE_FAILED                     = 0x0602
+typedef int OckamError;
 
-} OCKAM_ERR;
+enum {
+  kOckamErrorNone = 0,
+  kOckamError = 1,
+};
```
mrinalwadhwa: this should move below to common error codes section

implementations/c/include/ockam/log.h
```diff
@@ -88,9 +89,9 @@ typedef enum {
 extern "C" {
```
mrinalwadhwa: let's confirm that the way we `extern "C"` is consistent across all all our headers

implementations/c/include/ockam/memory.h
```diff
@@ -48,47 +46,120 @@
...
+   * @brief   Create an ockam memory module
+   *
+   * @param   p_arg Configuration structure
+   *
+   * @return  kOckamErrorNone on success.
+   *
+   ****************************************************************************************************
+   */
+
+  OckamError (*Create)(void *p_arg);
```
mrinalwadhwa: return interface specific error types

implementations/c/include/ockam/syslog.h
```diff
@@ -1,6 +1,6 @@
 #include <stdio.h>
```
mrinalwadhwa: we can remove `stdio.h` from all headers in `implementations/c/include/ockam`

implementations/c/include/ockam/transport.h
```diff
@@ -1,147 +1,168 @@
...
+#include <errno.h>
+#include <netdb.h>
+#include <netinet/in.h>
+#include <netinet/tcp.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
-#include <stdint.h>
+#include <sys/socket.h>
+#include <sys/types.h>
```
mrinalwadhwa: most of these can likely move to an implementations, except `stdint.h`

implementations/c/include/ockam/vault.h
```diff
@@ -76,88 +68,142 @@ typedef enum {
...
+                                     uint8_t *p_input, size_t input_size,
+                                     uint8_t *p_output, size_t output_size);
+
+OckamError VaultBearsslAesGcmDecrypt(OckamVaultCtx *ctx,
+                                     uint8_t *p_key, size_t key_size,
+                                     uint8_t *p_iv, size_t iv_size,
+                                     uint8_t *p_aad, size_t aad_size,
+                                     uint8_t *p_tag, size_t tag_size,
+                                     uint8_t *p_input, size_t input_size,
+                                     uint8_t *p_output, size_t output_size);
```
mrinalwadhwa: rename `bearssl` vault to `default` and move these functions to that impl header.

implementations/c/include/ockam/transport.h
```diff
@@ -1,147 +1,168 @@
...
-} OCKAM_LISTEN_ADDRESS;
+typedef struct {
+  uint16_t isConnected;                // connection with remote is established
+  OckamInternetAddress localAddress;   // human-friendly local address
+  OckamInternetAddress remoteAddress;  // human-friendly remote address
+  int socket;                           // posix socket identifier
+  struct sockaddr_in socketAddress;    // network-friendly socket information
+  Transmission receiveTransmission;
+  Transmission sendTransmission;
+} PosixSocket;
```
mrinalwadhwa: move all details related to Posix and TCP sockets to that impl header.

implementations/c/include/ockam/transport.h
```diff
@@ -1,147 +1,168 @@
...
+typedef struct {
+  enum  {
+    kBlocking,
+    kNonBlocking
+  } block;
+} OckamTransportConfig;
+
+typedef struct {
+  // Initialize
+  TransportError (*Create)(OckamTransportConfig*, OckamTransportCtx*);
```
mrinalwadhwa: make context the fist arg

mrinalwadhwa:
@robinbudd could you squash this to a commit that passes commitlint and rebase to develop.
@mtm0183 with the rebase to develop, this branch would have the right builder image, could you change the use of rpi toolchain so it doesn't need the submodule any more.

mrinalwadhwa:
The test folders in some place in the C code are called `test` and other places `tests` .. let's rename them consistently to tests @mtm0183 @robinbudd

mrinalwadhwa:
@mtm0183 since have a folder dedicated to vaults `lib/vault` ... I think vault tests can live there. If that is a minor change, then let's do that with this change ... if it will need some time. Please file an issue for that change.

mrinalwadhwa:
@robinbudd 
1. `lib/transport/posix` I think should be call `lib/transport/posix_socket`
2. tests in `lib/transport/tests` look like they are related to posix sockets in general or tcp sockets, they should live under `lib/transport/posix_socket`

if this is quick to change, let's do that now or file please file it as an issue.

mtm0183:
> @mtm0183 since have a folder dedicated to vaults `lib/vault` ... I think vault tests can live there. If that is a minor change, then let's do that with this change ... if it will need some time. Please file an issue for that change.

@mrinalwadhwa Are we removing the top-level `tests` directory?

mrinalwadhwa:
no we'll use it for more complex integrations tests ... in this case I think these tests belong in `lib/vault`

implementations/c/include/ockam/transport.h
robinbudd: done

implementations/c/include/ockam/transport.h
robinbudd: done

implementations/c/include/ockam/transport.h
robinbudd: done

implementations/c/include/ockam/transport.h
mrinalwadhwa: In `transport.h` there is:
```
#include <stdlib.h>
#include <string.h>
```
Do we need them?

mrinalwadhwa:
@robinbudd could change the `clean` task in `implementations/c/build.gradle` to 

```
task clean {
  doLast {
    delete '_build'
  }
}
```

The way it currently is it causes the debian build to get created on the clean path.

PR #115, opened 2020-03-25
build: update openssl to v1.1.1

bitwalker:
@mrinalwadhwa This addresses the OpenSSL versioning in the builder; unfortunately it likely requires updating the VM, but I'm going to update my other PR with an attempt at applying the changes from the Vagrantfile and see how that goes


mrinalwadhwa:
This needed `export DEBIAN_FRONTEND=noninteractive` to succeed, same with the patch in your branch. Otherwise the build waits infinitely for use input.
I squashed this with another commit and merged here https://github.com/ockam-network/ockam/commit/a8909b0bb4231ac2dc761ad471c5de4f88a09b06

PR #117, opened 2020-03-27
feat(c): add size parameter to memory free

mtm0183:
In order to integrate the C memory functions with Rust's
Alloc/GlobalAlloc, a size parameter is needed in OckamMemory's Free
function. Caller's of Free should always pass in the size they
allocated, but the underlying implementation of Free may choose to
enforce or ignore the size parameter.


mrinalwadhwa:
@mtm0183 the changes look good to me. Since we want a change like this to be reflected in changelogs, could you change the commit message to say `feat(c):` In the changelogs we'll consider `feat:` and `fix:` type commits.

PR #128, opened 2020-04-14
Improve Vault and Transport tests

mtm0183:
This PR contains two improvements:

1. Add getopts to the TCP test to allow for a non-fixed path to the fixtures directory. IP and Port are also added to getopts.
2. Check the error return for the default vault test create step.


mrinalwadhwa:
@mtm0183 could you resolve conflicts.
@robinbudd could you review after @mtm0183 has resolved the conflicts. I'll merge after Robin's review.

PR #129, opened 2020-04-14
build(c): add cmake support to be built by external projects

mtm0183:
CMake builds up a new include directory to be included by external
projects.

Modify CMake variables to allow for the Ockam C Library to be
included in other projects via FetchContent.

Only build for hardware targets on specific platforms.


implementations/c/CMakeLists.txt
```diff
@@ -39,20 +39,22 @@ include(ockam_triples)
...
 get_host_triple(OCKAM_HOST_TRIPLE OCKAM_HOST_ARCHITECTURE OCKAM_HOST_VENDOR OCKAM_HOST_OS OCKAM_HOST_ABI)
 
 # If the install prefix wasn't set, place the output in the build directory
-set(CMAKE_INSTALL_PREFIX ${CMAKE_BINARY_DIR} CACHE PATH "Force install prefix" FORCE)
-
-#TODO Clean-up install
-#if((NOT DEFINED CMAKE_INSTALL_PREFIX) OR CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
-#  set(CMAKE_INSTALL_PREFIX ${CMAKE_BINARY_DIR} CACHE PATH "Force install prefix" FORCE)
-#endif()
+if(CMAKE_INSTALL_PREFIX MATCHES "^C:/Program Files" OR CMAKE_INSTALL_PREFIX MATCHES "^/usr/local$")
```
mrinalwadhwa, line 53: that windows path looks suspicious, do we need it?

mrinalwadhwa:
@bitwalker could you review the changes in this branch. I'll merge after your review

implementations/c/CMakeLists.txt
mtm0183: It is to detect if the install prefix has been left to the default or not. For now the build system is going under the assumption that if you don't specify an install path, put the output in the build directory under a target triple.

In order to detect if the install prefix is set to default, CMake has `CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT` but its only set the first time CMake is run. If you rerun CMake without deleting the cache `CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT` becomes false, which I kinda think defeats the purpose of that variable. We run into this issue since we call CMake for Debug and Release from the same build directory. The regex is there to detect if its the default path or not for multiple CMake runs.

mrinalwadhwa: got it, thanks.

PR #133, opened 2020-04-17
feat: initial wire protocol impl for elixir

bitwalker:
- Define reusable encoding/decoding infrastructure
- Define messages that exist in proposed schema
- Define a message that contains nested types for testing
- Convert handshake implementation to use wire protocol

NOTE: This is not ready to be merged until upstream implementation in C
is ready, then the integration tests can be updated


mrinalwadhwa:
@bitwalker is there something in this PR that wasn't merged in #136 ?

bitwalker:
No, we can close this.

PR #135, opened 2020-05-08
docs: fix the link to the vault proposal, clarify its draft state

gustin:
### Proposed Changes
Documentation change to fix the broken link on the Vault concept page that references the Vault proposal. 

I could only find the Vault proposal as a WIP on a branch and PR. Maybe the link should go to the Vault guide instead.


mrinalwadhwa:
Hey @gustin thank you for the PR ❤️ 

Your changes make sense for now, until we get get the new vault interface proposal merged.

@mtm0183 is planning on an update to the vault interface proposal soon. Here's what the current draft looks like, incase you're curious.
https://github.com/ockam-network/ockam/blob/mark/new_interface/implementations/c/include/ockam/vault.h

PR #136, opened 2020-05-09
WIP: Update Elixir/Rust implementations to match new C interface

bitwalker:
In addition to updating the Elixir/Rust Vault APIs to use the new C interface, this changset also moves Elixir from the placeholder OpenSSL-based Vault to the C vault, updating the tests and internal implementation details which used the previous API.

This PR currently has two fixes in the C default software vault to address two issues I found during porting:

- When generating secrets, a secret size mismatch check had an incorrect guard that would always result in failure when generating new secrets (but likely worked fine for regenerating or importing secrets)
- When creating a new secret (i.e. import), there was an invalid guard that prevented creating secrets with non-null, but empty content. This is a edge case present in key agreement, as the channel keys created when finalizing the handshake are generated by performing `ecdh` on the key contained in the cipher state, and an empty key. There are tests based on this behavior, so it is not strictly necessary to allow this edge case, if we disallow it, we'd ideally need to find another pattern which has well-known test data we can verify against.

This is marked work-in-progress, as it is based against Mark's work-in-progress branch, and will need to be rebased once that branch is merged, and before this branch is merged.


mrinalwadhwa:
merged

PR #144, opened 2020-05-14
modify memory interface from uint8_t to void

mtm0183:
This PR modifies the Ockam memory interface to use `void*` rather than `uint8_t*` to closer mimic the stdlib implementation and prevent unnecessary casting. The one exception is the `allocate_zeroed()` which requires a `void**` to allow the function to be consistent with all other Ockam interface calls and return `ockam_error_t`

In addition to updating the interface, the vault's use of the memory interface and rust's use of the memory interface have been updated.


mrinalwadhwa:
@bitwalker could you review this please.

mrinalwadhwa:
merged

PR #145, opened 2020-05-14
docs: create contact links

mattgreg:
This PR routes users that click on 'Issues' in GitHub to the appropriate action.

Specifically, some are routed to 'Discussions' since that is a new UX in GitHub that many may not know about. 

This workflow encourages issues to be placed in issues, and discussions to happen in discussions.


mrinalwadhwa:
merged

PR #146, opened 2020-05-14
feat: draft of influx integration

bitwalker:
To run:

- Make sure you have Docker installed/running
- Run `implementations/elixir/influx.sh`
- Run `cd implementations/elixir && mix test --no-start
test/integration/influx_test.exs`

The way this is put together is not ideal, but works as a rough draft to
iterate from.

## NOTES

- The infrastructure I've set up to allow dispatching from the connection handler to the router, and from there to some arbitrary servicer has not been fully fleshed out so that all errors and edge cases are checked/handled/validated. It at least keeps things relatively disconnected though.
- This draft introduces four new message types `Ack`, `Error`, `Send`, and `Request`. These can obviously be changed/removed/whatever, but here's the reason for each in this PR:
  - `Ack` - used to acknowledge that a message was received and processed successfully
  - `Error` - used to indicate to the sender that an error occurred, and what it was
  - `Send` - used to indicate that the message being sent is not expected to return a value, just an acknowledgement, the sender doesn't need to wait
  - `Request` - used to indicate that the message being sent is expected to return a value and the sender will wait for it
  - `Send` and `Request` could certainly be merged, but initially I wanted to model whether or not the request was intended to produce a value that the sender wants
- This introduces a barebones Registry module used to register and locate services, such as Influx, that can be dispatched to
- The Influx service itself is modeled to reflect one way in which all services could be structured:
  - Each "connection" to a service is managed by a process dedicated to that connection, and knows about the transport-level connection that spawned it.
  - The transport-level connection will dispatch Connect requests to the router, using the endpoint of the `send_to` header to determine what to connect to. For local endpoints, the router will consult the registry for registered services by that name, and if one is found, it will relay the connection request to that service instance.
  - The service instance itself is responsible for decoding messages it can handle; this keep the lower-level protocol unaware of the service-specific message protocol, and leaves the encoding entirely up to the service itself.
  - The API for interacting with the service is described by a set of messages wrapped in a top-level message type. For example, the `Influx.Message` type wraps the `Influx.Message.Write` and `Influx.Message.Query` message types, and knows how to decode them, and also handles invalid message types. If the Influx service tries to decode a message and does not get an `Influx.Message` to dispatch on, it returns an error. Other services can easily follow this model.
- The interactions with Influx are currently happening over HTTP for testing; the Fluxter library is present, and can be wired up easily to use UDP for writing, but it does a lot of buffering/caching of things as it is designed for high-throughput use cases, so ended up not working well for simple tests like I have set up here (since there is no way to force-flush data to Influx via Fluxter).

If I've forgotten anything, feel free to ask any questions.

### DIFF
--- a/implementations/elixir/config/config.exs
+++ b/implementations/elixir/config/config.exs
@@ -2,5 +2,6 @@ import Config
 
 config :ockam, vault: []
 config :ockam, transports: []
+config :ockam, services: []
 
 import_config "#{Mix.env()}.exs"
--- a/implementations/elixir/config/dev.exs
+++ b/implementations/elixir/config/dev.exs
@@ -18,3 +18,19 @@ config :ockam, :transports,
     listen_address: "0.0.0.0",
     listen_port: 4000
   ]
+
+config :ockam, :services,
+  influx_example: [
+    service: Ockam.Services.Influx,
+    database: "test",
+    http: [
+      host: "127.0.0.1",
+      port: 8086
+    ]
+  ]
+
+config :fluxter, Ockam.Services.Influx.Fluxter,
+  host: "127.0.0.1",
+  port: 8089,
+  pool_size: 5,
+  prefix: nil
--- a/implementations/elixir/config/test.exs
+++ b/implementations/elixir/config/test.exs
@@ -3,4 +3,8 @@ import Config
 config :logger,
   level: :warn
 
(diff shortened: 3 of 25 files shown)


mrinalwadhwa:
merged 6bc23e5ee35967584dd62403e007000aec6e8a72

PR #184, opened 2020-05-20
feat(c): implement encrypted channel with wire protocol for key excha…

robinbudd:
implement secure channels, perform key exchange upon connection

implement wire protocol for key exchange messages and payload_test

onward and return routes not implemented


mrinalwadhwa:
merged.

PR #205, opened 2020-06-18
feat(c): add memory compare interface

SanjoDeundiak:
feat(c): add compare to memory interface, add compare to stdlib memory implementation


mrinalwadhwa:
@SanjoDeundiak thank you for sending this.
Please sign all commits before we merge.

SanjoDeundiak:
@mrinalwadhwa I signed commits and rebased against latest develop. Though, I was thinking about adding proper tests for rust memory implementation (at least for memory compare). But we can merge this PR for now and add tests later according to priorities

mrinalwadhwa:
let's wait for you to be ready with tests, assign this PR to @robinbudd when you're ready for review.

implementations/c/lib/memory/stdlib/tests/stdlib_memory_test.c
```diff
@@ -0,0 +1,157 @@
+#include <stdarg.h>
```
mtm0183, line 1: This file looks great! One suggestion, since the test is calling the ockam_memory interface we should consider if this is a valid test for all implementations of ockam_memory or if its specific to stdlib.

I *think* this should apply to any implementation of ockam_memory so we might be better served if we moved this to something like compare.c and have it live in lib/memory/tests. Then this stdlib_memory_test.c file can call compare.c functions and as we grow the memory test suite other memory test functions.

implementations/c/lib/memory/stdlib/CMakeLists.txt
```diff
@@ -18,3 +18,7 @@ ockam_cc_library(
      ockam::memory
    PUBLIC)
 
+if(${OCKAM_BUILD_TESTS})
+    add_subdirectory(tests)
+    message(STATUS "-----TEST ockam_codec--------")
```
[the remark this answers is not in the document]
SanjoDeundiak: Fixed

implementations/c/lib/memory/stdlib/tests/stdlib_memory_test.c
SanjoDeundiak: Good point, fixed that. Also, new version should allow to easily choose what implementations to test according to defines (e.g. different platform may have different implementations available)

implementations/c/lib/memory/tests/memory_compare_test.c
```diff
@@ -0,0 +1,201 @@
...
+  size_t           implementations_count;
+} test_state_t;
+
+static int test_setup(void **state)
+{
+  static test_state_t test_state;
+
+  size_t i = 0;
+
+  // ifdef should be added here when more memory implementations added
```
mtm0183, line 22: To make sure I understand it right, in this code there's an executable that is built for whatever target we compile for, but the executable exercises all memory implementations? My only concern with this is if there are certain memory implementations that can't run on a specific target we need a way for them to not be built, easy enough to do with a compiler flag or similar I suppose.

My original thought for a layout would have been:
`lib/memory/stdlib/test/test_stdlib.c` -> Contains `main()`, calls `test_compare()`
`lib/memory/test/test_compare.c` -> Contains `test_compare()`
The upside here is the stdlib directory only gets built for the targets its supported on. 

I think either solution works, we just need to make sure we have a way to disable certain implementations when compiling for certain targets. Thoughts?

SanjoDeundiak: @mtm0183 I understand that at some point we'll need to enable/disable specific implementations, my plan was:
- define cmake variable, e.g. OCKAM_STD_MEMORY_ENABLED
- based on that variable we decide whether we a) are building stdlib library b) are linking it to other libraries
- tests can simply check #ifdef OCKAM_STD_MEMORY_ENABLED in setup function and add corresponding implementation to test_state object
- tests themselves don't care about all of that, they are given state object with array of implementations, and they test them, they don't even know what implementations are

mtm0183: Ok. We may find as we move forward we end up with too many cmake variables for the various implementations but for now I don't think it is an issue. This is a fine solution and we can always revisit it as we add more implementations.

mrinalwadhwa:
rebased and force pushed to the branch to make sure it builds in ci

PR #227, opened 2020-07-07
Vault ATECC608A

mtm0183:
This PR has 3 components:

1. Add support for the Microchip ATECC608A. This implementation works with the ockam vault secrets to hold keys. The private keys live in the ATECC608A but shared secrets and AES keys are pulled out of the device and stored in buffers due to limitations on the ATECC608A.
2. Add P256 support to the default vault to allow for key agreement with the ATECC608A.
3. Add Mutex interface for pthreads


implementations/c/lib/vault/impl.h
```diff
@@ -211,7 +211,8 @@ typedef struct {
  */
 struct ockam_vault {
   ockam_vault_dispatch_table_t* dispatch;
-  void*                         context;
+  void*                         default_context;
+  void*                         impl_context;
```
SanjoDeundiak, line 215: It's not used anywhere

mrinalwadhwa:
rebased and merged

PR #230, opened 2020-07-15
Added errors

mikelodder7:
Adds crate specific error that also converts to FFI integer values found in implementations/c/include/ockam/{vault.h, error.h}

### DIFF
--- a/implementations/rust/Cargo.toml
+++ b/implementations/rust/Cargo.toml
@@ -3,5 +3,8 @@
 members = [
   "crates/ockam-vault-sys",
   "crates/ockam",
-  "vault"
+]
+
+exclude = [
+    "vault"
 ]
--- a/implementations/rust/vault/src/error.rs
+++ b/implementations/rust/vault/src/error.rs
@@ -1,0 +1,281 @@
+use failure::{Backtrace, Context, Fail};
+use std::{fmt, io};
+
+/// Represents the failures that can occur in
+/// an Ockam Vault
+#[derive(Clone, Copy, Fail, Debug)]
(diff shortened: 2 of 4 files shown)


implementations/rust/vault/src/error.rs
```diff
@@ -0,0 +1,281 @@
...
+    SecretSizeMismatch,
+    /// An error occurred while reading from I/O
+    #[fail(display = "An error occurred while reading from I/O")]
+    IOError,
+}
+
+impl VaultFailErrorKind {
+    pub(crate) const ERROR_INTERFACE_VAULT: usize = 3 << 24;
+    /// Convert to an integer
+    pub fn to_usize(&self) -> usize {
```
SanjoDeundiak, line 97: @mrinalwadhwa @robinbudd there was an idea of revisiting error handling in C, which probably will also affect this part of a rust code. Any news on that?

implementations/rust/Cargo.toml
```diff
@@ -3,5 +3,8 @@
 members = [
   "crates/ockam-vault-sys",
   "crates/ockam",
-  "vault"
+]
+
+exclude = [
```
SanjoDeundiak, line 8: Why is vault excluded?

mikelodder7: For now, I'm building it separately to avoid some library conflict issues.

implementations/rust/vault/src/error.rs
mrinalwadhwa: Yes, but let's not block the rust code on that change.
We can iterate one both pieces together, once they're working together.

PR #232, opened 2020-07-17
feat(rust): add types and trait

mikelodder7:
Adds the vault trait and types associated with calling a vault.


implementations/rust/vault/src/types.rs
```diff
@@ -0,0 +1,171 @@
...
+/// Key is backed by an Intel SGX enclave
+/// IntelSgx,
+/// Key is backed by an AMD's PSP enclave
+/// AmdPsp,
+/// Key is backed by an ARM's Trustzone enclave
+/// ArmTrustzone,
+/// Key is backed a Apple's secure enclave
+/// SecureEnclave,
+#[derive(Clone, Copy, Debug)]
+pub enum SecretKeyContext {
```
mrinalwadhwa, line 133: Would exposing type of vault/secret here create coupling we don't want?
For instance, we want someone to be able to plug-in a custom vault implementation without having to change Ockam code.

SanjoDeundiak: In such cases to allow flexible usage it's convenient to use string as a type (or a subtype). Also, we can provide constants for some predefined types

mrinalwadhwa:
rebased and merged https://github.com/ockam-network/ockam/commit/a9cb8ba5c8d380c58743553d883efee70f77dd8a

implementations/rust/vault/src/lib.rs
```diff
@@ -1,24 +1,96 @@
...
+        xtype: SecretKeyType,
+    ) -> Result<(), VaultFailError>;
+    /// Remove a secret key from the vault
+    fn secret_destroy(&mut self, context: SecretKeyContext) -> Result<(), VaultFailError>;
+    /// Compute Elliptic-Curve Diffie-Hellman using this secret key
+    /// and the specified public key
+    fn ec_diffie_hellman(
+        &mut self,
+        context: SecretKeyContext,
+        peer_public_key: [u8; 32],
```
SanjoDeundiak, line 76: P256 public keys used to be 65 bytes

implementations/rust/vault/src/lib.rs
```diff
@@ -1,24 +1,96 @@
...
+    ) -> Result<SecretKey, VaultFailError>;
+    /// Compute the HKDF-SHA256 using the specified salt and input key material
+    /// and return the output key material of the specified length
+    fn hkdf_sha256<B: AsRef<[u8]>>(&mut self, salt: B, ikm: B, okm_len: usize) -> Result<Vec<u8>, VaultFailError>;
```
SanjoDeundiak, line 80: Isn't it supposed to have separate template type for salt and for ikm?

implementations/rust/vault/src/lib.rs
mikelodder7: That depends on if they are compressed or not. But you are right. Compressed for is 33 bytes, uncompressed is 65 bytes. We need to decide which form to support.

implementations/rust/vault/src/lib.rs
mikelodder7: HKDF extract takes `salt` and `ikm` together and outputs the `prk`. 
HKDF expand takes optional `info` which is usually left blank with the `prk`

See [here](https://tools.ietf.org/html/rfc5869)

implementations/rust/vault/src/types.rs
mikelodder7: Context is where the key is and how is addressed.

implementations/rust/vault/src/lib.rs
SanjoDeundiak: I know how HKDF works, the question is about template <B: AsRef<[u8]>. As far as I understood it's to make calling the function more convenient e.g. you can call it with slice or with vector. But what if ikm is vector and salt is slice (which is very likely to happen because ikm is something you can get after ECDH and it can be Vector, while salt may be some static const represented as slice). Wouldn't this function fail in that case? Does it make sense to have template <B1: AsRef<[u8]>, <B2: AsRef<[u8]>> and salt: B1, ikm: B2?

implementations/rust/vault/src/lib.rs
SanjoDeundiak: ATECC608A supports only uncompressed format and from my experience P256 public keys are represented in uncompressed format for historical reasons. Maybe @mrinalwadhwa has ideas about that

mikelodder7: When they only support uncompressed, it’s usually a shortcoming if the library or a fear of calculating the `y` coordinate incorrectly. Usually it’s not a problem I’ve found so the former is the most common

implementations/rust/vault/src/lib.rs
mikelodder7: No because both of them implement the AsRef<[u8]> trait.

SanjoDeundiak: @mikelodder7 I prepared an example to explain what I'm talking about https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=324e36f5bc058b23dc0f81c1c43126dc

mikelodder7: I haven't ever seen this problem in production. Perhaps we consider this change later. Thanks for the suggestion.

PR #234, opened 2020-07-20
Mikelodder7/vault updates

mikelodder7:
This implements the default vault in memory. The Rust API might be subject to change depending on how we optimize it from here.


mrinalwadhwa:
@mikelodder7 could you update the commit message, that's why the build is failing.

PR #268, opened 2020-07-26
feat(rust): implement wire protocol codec for ockam messages

robinbudd:


.gitignore
```diff
@@ -13,3 +13,4 @@ cmake_install.cmake
 *.gch
 .idea
 .tool-versions
+Cargo.lock
```
mrinalwadhwa, line 16: The Cargo.lock file shouldn't be gitignored or removed.
It pins our dependencies.

Since your Cargo.toml changes added a new member to the workspace. The updates to Cargo.lock should be part of this PR

robinbudd: Done

implementations/rust/message/src/lib.rs
```diff
@@ -0,0 +1,540 @@
+#![allow(unused)]
+
+// Definition and implementation of an Ockam message and message components.
+// Each message component, and the message overall, implements the "Codec" trait
+// allowing it to be encoded/decoded for transmission over a transport.
+
+pub mod message {
```
mikelodder7, line 7: I would move this to a file named `message.rs`. Then you can dispense with the pub mod group

implementations/rust/message/src/lib.rs
```diff
@@ -0,0 +1,540 @@
...
+          w = u1;
+        }
+        Err(s) => { return Err(s); }
+      }
+      Ok((msg, w))
+    }
+  }
+
+  /* Addresses */
+  enum AddressType {
```
mikelodder7, line 78: Is this meant to be used by library consumers? If so, insert `pub`

implementations/rust/message/src/lib.rs
```diff
@@ -0,0 +1,540 @@
...
+    Local = 0,
+    Tcp = 1,
+    Udp = 2,
+  }
+
+  #[derive(PartialEq)]
+  #[derive(Debug)]
+  pub struct LocalAddress {
```
mikelodder7, line 86: This can be collapsed into `#[derive(Debug, PartialEq, Eq)]

implementations/rust/message/src/lib.rs
```diff
@@ -0,0 +1,540 @@
...
+    pub length: u8,
+    pub address: Vec<u8>,
+  }
+
+  #[derive(PartialEq)]
+  #[derive(Debug)]
+  pub enum Address {
```
mikelodder7, line 93: See previous

implementations/rust/message/src/lib.rs
```diff
@@ -0,0 +1,540 @@
...
+    LocalAddress(LocalAddress),
+    TcpAddress(IpAddr, u16),
+    UdpAddress(IpAddr, u16),
+  }
+
+  enum HostAddressType {
```
mikelodder7, line 99: Is this meant to be public? If so, insert `pub`

implementations/rust/message/src/lib.rs
```diff
@@ -0,0 +1,540 @@
...
+      let length =  u[0] as usize;
+      let address =  u[1..length+1].to_vec();
+      Ok((LocalAddress { length: u[0], address }, &u[length+1..]))
+    }
+  }
+
+  /* Routes */
+  #[derive(PartialEq)]
+  #[derive(Debug)]
+  pub struct Route {
```
mikelodder7, line 205: Can be collapsed into `#[derive(Debug, PartialEq, Eq)]`

implementations/rust/message/src/lib.rs
mikelodder7: Consider adding `Copy, Clone` to the derive. Trust me, it helps.

implementations/rust/message/src/lib.rs
mikelodder7: Consider adding `Copy, Clone` to the derive.

implementations/rust/message/src/lib.rs
mikelodder7: Consider adding `Copy, Clone`.

implementations/rust/message/src/lib.rs
```diff
@@ -0,0 +1,540 @@
...
+            ()
+          }
+        }
+      }
+      Ok((route, w))
+    }
+  }
+
+  #[derive(Debug)]
+  pub enum MessageBody {
```
mikelodder7, line 240: Consider adding `PartialEq, Eq, Copy, Clone` to **derive**

implementations/rust/message/src/lib.rs
```diff
@@ -0,0 +1,540 @@
...
+      match data {
+        0 => Ok(MessageBody::Ping),
+        1 => Ok(MessageBody::Pong),
+        _ => Err("Not Implemented".to_string())
+      }
+    }
+  }
+
+  #[allow(arithmetic_overflow)]
+  impl Codec<u16> for u16 {
```
mikelodder7, line 281: usually not a good idea to allow arithmetic_overflow. I wonder which part is causing this error because it's usually `+, -, *, /, %` and I don't see that anywhere except line 309.

implementations/rust/message/src/lib.rs
```diff
@@ -0,0 +1,540 @@
...
+    assert_eq!(v.len(), 1);
+    match Route::decode(&v) {
+      Ok((r, v)) => {
+        assert_eq!(r.addresses.len(), 0);
+        assert_eq!(v.len(), 0);
+      },
+      Err(s) => { panic!(); }
+    }
+
+  }
```
mikelodder7, line 431: Nice tests!!!

implementations/rust/message/src/lib.rs
```diff
@@ -0,0 +1,540 @@
...
+    u16::encode(&mut n, &mut u);
+  }
+
+  #[test]
+  fn u16_codec() {
+    let mut u: Vec<u8> = vec![];
+    let mut n: u16 = 0x7f;
+    u16::encode( &mut n, &mut u);
+    assert_eq!(u.len(), 1);
+    assert_eq!(u[0], 0x7f);
```
mikelodder7, line 447: What are these magic byte sequences like `0x7f, 0x80, 0x01`? Could you add a comment for these particular test values

implementations/rust/message/src/lib.rs
robinbudd: done

implementations/rust/message/src/lib.rs
robinbudd: don't need it after all!

implementations/rust/message/src/lib.rs
robinbudd: Yes comments are in order. They are used when encode variable-length integers. Bit twiddling.

Issue #265, opened 2020-07-24
switch to using cmocka-static in all tests


mrinalwadhwa:
#237

mrinalwadhwa:
fixed in https://github.com/ockam-network/ockam/pull/269/commits/3da8b1dc71fbf1247563e9a38ae9fee6199174ff

Issue #237, opened 2020-07-21
C build fails on windows


mrinalwadhwa:
https://github.com/ockam-network/ockam/runs/895147485?check_suite_focus=true#step:4:201

mrinalwadhwa:
Fixed in https://github.com/ockam-network/ockam/pull/269

New issue added fro visual studio: #272

SanjoDeundiak:
This one can be closed I think

Issue #262, opened 2020-07-24
u_int16_t breaks builds with multiple toolchains


mrinalwadhwa:
```
» grep -R "u_int" *
ockam/key_agreement/xx/xx_initiator.c:  u_int16_t     offset                = 0;
ockam/queue/pthread/queue.c:  u_int16_t       tail;
```

```
/work/implementations/c/ockam/key_agreement/xx/xx_initiator.c:171:3: error: unknown type name 'u_int16_t'; did you mean 'uint16_t'?
  u_int16_t     offset                = 0;
  ^~~~~~~~~
  uint16_t
/emsdk_portable/fastcomp/emscripten/system/lib/libc/musl/arch/emscripten/bits/alltypes.h:186:25: note: 'uint16_t' declared here
typedef unsigned short  uint16_t;
```

mrinalwadhwa:
@robinbudd can I assume that its safe to replace `u_int16_t` with `uint16_t`?

SanjoDeundiak:
@mrinalwadhwa works for gcc. Should work for others

mrinalwadhwa:
fixed in https://github.com/ockam-network/ockam/pull/269/commits/45e5f0701e979e4fc26b194a991d8670bd9903c2

PR #269, opened 2020-07-27
Fix windows build

SanjoDeundiak:
Use static cmocka library for tests. Enable windows CI build. Fix for #237 and #265

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,10 +18,22 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - id: Build
+      - id: Build_windows
+        if: startsWith(matrix.os, 'windows')
+        run: ../../gradlew build -PcmakeGenerator="Unix Makefiles"
+        working-directory: implementations/c
+
+      - id: Build_not_windows
+        if: "!startsWith(matrix.os, 'windows')"
         run: ../../gradlew build
         working-directory: implementations/c
 
-      - id: Test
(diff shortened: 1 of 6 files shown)


SanjoDeundiak:
I have read the CLA Document and I hereby sign the CLA

SanjoDeundiak:
@mrinalwadhwa after your recent gradle changes CI build on windows started to use MVSC compiler (instead of gcc), which caused new problems with build (e.g. absence of pthread). Do we want to be able to build with gcc on windows (which should be possible with this PR), or make everything work with mvsc?

SanjoDeundiak:
#262

mrinalwadhwa:
That's really interesting. The gradle changes from last night make it so that gradle doesn't pick the default CMake generator. Instead CMake picks the default generator.

The idea there is that we want to test for / optimize the default experience most people will have on a platform.

In case of windows (or at least the windows ci machine) it seems CMake picked the Visual Studio generator. Which also picks visual studio's default compiler.

For this PR let's change `ci.yml` to specify `Unix Makefiles` generator when the OS is windows.

```
../../gradlew build -PcmakeGenerator="Unix Makefiles"
```

mrinalwadhwa:
@SanjoDeundiak I forgot to mention, the test run needs the `cmakeGenerator` option too

mrinalwadhwa:
It took me a second to realize why the github workflow didn't run
https://github.com/ockam-network/ockam/actions/runs/184488829#L26

Issue #243, opened 2020-07-22
Improve Logging in C


mrinalwadhwa:
- [x] deal with multiple log modules
- [x] figure out the interface we should expose as `ockam/log.h`
- [x] figure out how the logger will be chosen - CMake?
- [x] figure out how logging will be used.
- [x] how does this affect log/CMakeLists.txt
- [ ] how does logging work with C and Rust together

mrinalwadhwa:
@SanjoDeundiak proposal 
https://github.com/SanjoDeundiak/cmake-test1
https://github.com/SanjoDeundiak/cmake-test2

SanjoDeundiak:
C <-> Rust log interference should be discussed, other items should be resolved by merging #289

PR #289, opened 2020-08-03
feat(c): upgrade log

SanjoDeundiak:
Switch to new log


SanjoDeundiak:
#243

implementations/c/ockam/log/CMakeLists.txt
```diff
@@ -1,18 +1,40 @@
...
+
+target_link_libraries(
+  ockam_log
+  PUBLIC
+    ockam::error_interface
+)
+
+if (OCKAM_DISABLE_LOG)
+  message(STATUS "Log is disabled")
+  add_definitions(-DOCKAM_DISABLE_LOG)
```
mrinalwadhwa, line 30: I suspect this may need to be tied to the target, with `target_compile_definitions ` maybe.

SanjoDeundiak: @mrinalwadhwa tnx for pointing out. I was expecting target_add_definitions to exist, but I was wrong. Fixed

PR #295, opened 2020-08-05
Transport integer casts fixes, udp test hanging fix

SanjoDeundiak:
1. Transport integer casts fixes: posix recv/send socket functions return ssize_t which should be checked for negative values before assigning to size_t variable, because that causes integer overflow for buffer_size variables, which can lead to all sorts of problems (including security)
2. UDP test hanging fix: udp doesn't involve acknowledges, so it's possible to send any amount of data to udp socket and this data should be put into some OS buffer, waiting for user program to get it from there. Consequently this buffer can overflow, this way receiving side won't get termination string (which is defined by our test logic) and just hang on some read(). Size of this buffer is adjusted by OS and who knows what the value is for GitHub actions. Proposed fix: add occasional delays on sending side to give receiving side some time to get data from the buffer.

### DIFF
--- a/implementations/c/ockam/transport/posix_socket/socket.h
+++ b/implementations/c/ockam/transport/posix_socket/socket.h
@@ -14,6 +14,8 @@
 #include "ockam/transport/impl.h"
 #include "socket.h"
 
+// TODO: add ockam_ prefix to types declared here. Review which of them indeed need to be public.
+
 /**
  * The PosixSocket is the posix socket specific class data for a posix socket
  * connection (TCP or UDP). Note that TCP sockets are further defined by the
--- a/implementations/c/ockam/transport/posix_socket/socket_tcp.c
+++ b/implementations/c/ockam/transport/posix_socket/socket_tcp.c
@@ -255,7 +255,7 @@ ockam_error_t socket_tcp_read(void* ctx, uint8_t* buffer, size_t buffer_size, si
     bytes_read        = recv(p_socket->socket_fd, &recv_len, sizeof(uint16_t), 0);
     // Must convert from network order to native endianness
     p_transmission->transmit_length = ntohs(recv_len);
-    if (-1 == bytes_read) {
+    if (sizeof(uint16_t) != bytes_read) {
       error = TRANSPORT_ERROR_RECEIVE;
       goto exit;
     }
@@ -264,12 +264,11 @@ ockam_error_t socket_tcp_read(void* ctx, uint8_t* buffer, size_t buffer_size, si
 
   bytes_read = 0;
   while ((TRANSPORT_ERROR_MORE_DATA == p_transmission->status) && (p_transmission->buffer_remaining > 0)) {
-    uint16_t bytes_to_read = 0;
-    ssize_t  recv_status   = 0;
-    bytes_to_read          = p_transmission->transmit_length - p_transmission->bytes_transmitted;
+    ssize_t recv_status   = 0;
+    size_t  bytes_to_read = p_transmission->transmit_length - p_transmission->bytes_transmitted;
     if (bytes_to_read > p_transmission->buffer_remaining) bytes_to_read = p_transmission->buffer_remaining;
     recv_status = recv(p_socket->socket_fd, p_transmission->buffer + bytes_read, bytes_to_read, 0);
-    if (-1 == recv_status) {
+    if (0 > recv_status) {
       ockam_log_error("receive failed: %ll", recv_status);
       goto exit;
     }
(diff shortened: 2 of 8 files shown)


implementations/c/ockam/transport/posix_socket/socket_tcp.c
```diff
@@ -264,12 +264,11 @@ ockam_error_t socket_tcp_read(void* ctx, uint8_t* buffer, size_t buffer_size, si
 
   bytes_read = 0;
   while ((TRANSPORT_ERROR_MORE_DATA == p_transmission->status) && (p_transmission->buffer_remaining > 0)) {
-    uint16_t bytes_to_read = 0;
-    ssize_t  recv_status   = 0;
-    bytes_to_read          = p_transmission->transmit_length - p_transmission->bytes_transmitted;
+    ssize_t recv_status   = 0;
+    size_t  bytes_to_read   = p_transmission->transmit_length - p_transmission->bytes_transmitted;
```
mikelodder7, line 268: Good change.

implementations/c/ockam/transport/posix_socket/socket_tcp.c
```diff
@@ -255,7 +255,7 @@ ockam_error_t socket_tcp_read(void* ctx, uint8_t* buffer, size_t buffer_size, si
     bytes_read        = recv(p_socket->socket_fd, &recv_len, sizeof(uint16_t), 0);
     // Must convert from network order to native endianness
     p_transmission->transmit_length = ntohs(recv_len);
-    if (-1 == bytes_read) {
+    if (sizeof(uint16_t) != bytes_read) {
       error = TRANSPORT_ERROR_RECEIVE;
```
mikelodder7, line 259: Do we always expect a fixed amount to be read?

implementations/c/ockam/transport/posix_socket/socket_tcp.c
```diff
@@ -296,15 +295,15 @@ ockam_error_t socket_tcp_read(void* ctx, uint8_t* buffer, size_t buffer_size, si
...
   ockam_error_t   error       = OCKAM_ERROR_NONE;
   tcp_socket_t*   p_tcp_ctx   = (tcp_socket_t*) ctx;
   posix_socket_t* p_socket    = &p_tcp_ctx->posix_socket;
   uint16_t        send_length = 0;
-  size_t          bytes_sent  = 0;
 
   // Convert from native endianness to network order
   send_length = htons(buffer_length);
-  bytes_sent  = send(p_socket->socket_fd, (void*) &send_length, sizeof(uint16_t), 0);
+  ssize_t bytes_sent = send(p_socket->socket_fd, (void*) &send_length, sizeof(uint16_t), 0);
```
mikelodder7, line 306: is this ever negative?

implementations/c/ockam/transport/posix_socket/socket_udp.c
```diff
@@ -181,17 +181,17 @@ ockam_error_t socket_udp_read(void* ctx, uint8_t* buffer, size_t buffer_size, si
...
-                      (struct sockaddr*) &p_socket->remote_sockaddr,
-                      sizeof(p_udp_ctx->posix_socket.remote_sockaddr));
+
+  ssize_t bytes_sent = sendto(p_socket->socket_fd,
+                              buffer,
+                              buffer_length,
+                              0,
+                              (struct sockaddr*) &p_socket->remote_sockaddr,
+                              sizeof(p_udp_ctx->posix_socket.remote_sockaddr));
   if (bytes_sent != buffer_length) {
```
mikelodder7, line 195: one is signed the other is unsigned. Should probably do an explicit cast and saturate.

implementations/c/ockam/transport/posix_socket/socket_tcp.c
SanjoDeundiak: AFAIK for tcp we first send size of the following data. Also, we pass sizeof(uint16_t) as a size of buffer where we want to receive data recv, so bytes_read cannot be more than that. Negative values are errors. Don't know if results from 0 to sizeof(uint16_t) could happen and how to handle them. @mikelodder7 @robinbudd thought?

implementations/c/ockam/transport/posix_socket/socket_tcp.c
SanjoDeundiak: It can be -1

implementations/c/ockam/transport/posix_socket/socket_udp.c
SanjoDeundiak: I think worth checking bytes_sent for negative values

implementations/c/ockam/transport/posix_socket/socket_tcp.c
mikelodder7: if -1 is cast to unsigned, does it saturate at uint16_t.MAX?

SanjoDeundiak: @mikelodder7 yes, but it should not be problem here.

implementations/c/ockam/transport/posix_socket/socket_udp.c
SanjoDeundiak: Added

implementations/c/ockam/transport/posix_socket/socket_tcp.c
mikelodder7: A saturation cast means -1 == uint16_t.MAX, i.e. the if statement will succeed. Does this lead to unexpected behavior?

SanjoDeundiak: @mikelodder7 sizeof(uint16_t) is 2

implementations/c/ockam/transport/posix_socket/socket.h
```diff
@@ -14,6 +14,8 @@
 #include "ockam/transport/impl.h"
 #include "socket.h"
 
+// TODO: add ockam_ prefix to types declared here. Review which of them indeed need to be public.
```
mrinalwadhwa, line 17: let's try to note these types of TODOs in github issues, otherwise there is high likely hood that they will get missed.

mrinalwadhwa:
@robinbudd a quick review from you would be really helpful with this PR.

implementations/c/ockam/transport/posix_socket/socket_tcp.c
robinbudd: @mrinalwadhwa @mikelodder7 You'll never get 0 back from receive. If there's no data, you get -1 with errno set to EWOULD_BLOCK. If you get 1 byte, then the function fails.

And you'll never get more than sizeof(uint16_t) bytes b/c you've told TCP that that is your max buffer size.

So I think all cases were handled properly in the original code, although these changes also result in correct behavior.

mikelodder7:
Merged

implementations/c/ockam/transport/posix_socket/socket_tcp.c
SanjoDeundiak: @robinbudd I wonder if bytes_read could be equal to 1 here...

PR #297, opened 2020-08-05
fix(c): fix target names

SanjoDeundiak:


mrinalwadhwa:
the check failure is likely unrelated. so we should be able to merge this.

mikelodder7:
Merged in d70d0b27cca78377522a0d020470d062a7492307

PR #300, opened 2020-08-05
Add initial ockam key exchange prologue code

mikelodder7:
Doesn't handle the exchange yet. Just does the prologue whether its for initiator or responder


mrinalwadhwa:
minor suggestion for the commit message -
```
feat(rust): add initial key exchange
```

and then mention the parts of the key exchange feature added in the commit message body - prologue, mix_key, mix_hash etc.

mrinalwadhwa:
https://github.com/ockam-network/ockam/issues/301

PR #304, opened 2020-08-06
feat(rust): refactor vault and add handle message1 to kex

mikelodder7:
Add handling of message 1 from initiator to responder

### DIFF
--- a/implementations/rust/common/Cargo.toml
+++ b/implementations/rust/common/Cargo.toml
@@ -1,9 +1,7 @@
 [package]
-name = "common"
+name = "ockam-common"
 version = "0.1.0"
 authors = ["Ockam Developers"]
 edition = "2018"
-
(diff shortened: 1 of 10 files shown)


implementations/rust/common/Cargo.toml
```diff
@@ -1,9 +1,7 @@
 [package]
-name = "common"
+name = "ockam-common"
```
mrinalwadhwa, line 2: My experience has been that modules called "common", "shared" etc. by definition tend to violate the single responsibility principle. They change a lot, for lot of reasons. This usually worsens the modularity of the code and increases coupling.

I think, it would be better to have an error module even if it is rather simple.

mrinalwadhwa:
apart from the one comment above (which we can address in different PR later) ... the rest looks good to go 👍

implementations/rust/common/Cargo.toml
mikelodder7: Okay.

PR #307, opened 2020-08-10
feat(rust): implement encode and decode message 2 in key exchange

mikelodder7:


implementations/rust/kex/src/lib.rs
```diff
@@ -180,7 +180,27 @@ impl<'a, V: Vault> Initiator<'a, V> {
...
+        re.copy_from_slice(&message[..32]);
+        let encrypted_rs_and_tag = &message[32..80];
+        let encrypted_payload_and_tag = &message[80..];
+
+        let re = PublicKey::Curve25519(re);
+        self.0.handshake.remote_ephemeral_public_key = Some(re);
+
+        self.0.mix_hash(&re)?;
+        let shared_secret_ctx = self.0.vault.ec_diffie_hellman(self.0.handshake.ephemeral_secret_handle, re)?;
+        let shared_secret = self.0.vault.secret_export(shared_secret_ctx)?;
```
mrinalwadhwa, line 194: We should talk about what we could do to make key agreement feasible without needing to export this shared secret.

mikelodder7: I know what I can do, allow the vault to take a handle for the IKM argument. However, not all enclaves allow that but we can cross that bridge when we get there.

PR #311, opened 2020-08-12
feat(rust): add first tests for key exchange

mikelodder7:
Finishes key exchange and passes test 1

### DIFF
--- a/implementations/rust/kex/src/lib.rs
+++ b/implementations/rust/kex/src/lib.rs
@@ -106,8 +106,10 @@ impl<'a, V: Vault> XXSymmetricState<'a, V> {
         // 5. h = SHA256(h || prologue),
         // prologue is empty
         // mix_hash(xx, NULL, 0);
-        let h = vault.sha256(Self::CSUITE)?;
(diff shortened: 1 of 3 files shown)


implementations/rust/kex/src/lib.rs
```diff
@@ -106,8 +106,10 @@ impl<'a, V: Vault> XXSymmetricState<'a, V> {
         // 5. h = SHA256(h || prologue),
         // prologue is empty
         // mix_hash(xx, NULL, 0);
-        let h = vault.sha256(Self::CSUITE)?;
+        let mut h = [0u8; 32];
+        h[..Self::CSUITE.len()].copy_from_slice(Self::CSUITE);
```
mrinalwadhwa, line 110: note that the noise paper says -
```
If protocol_name is less than or equal to HASHLEN bytes in length, sets h equal to protocol_name with zero bytes appended to make HASHLEN bytes. Otherwise sets h = HASH(protocol_name).
```

we'll have make sure we do this correctly when we change the name to be an Ockam specific name

PR #329, opened 2020-08-19
feat(c): upgrade error handling

### LINKED ISSUE #317: (C) Improve Errors

SanjoDeundiak:
1. Upgrade error handling
2. Get rid of queue_interface and channel_interface cmake modules, since we don't have real interface/implementation approach there

Closes #317


mrinalwadhwa:
@SanjoDeundiak could you amend the commit to fix the two authors on the commit

Screen Shot 2020-08-19 at 9 10 24 AM

SanjoDeundiak:
@mrinalwadhwa any ideas how to do it? I though this happens because actual GitHub account email (and signing key assigned to it) differs from email I enter while committing (which is working email)

implementations/c/ockam/channel/tests/channel_test.c
```diff
@@ -71,7 +71,7 @@ ockam_error_t parse_opts(int argc, char* argv[])
 
 int main(int argc, char* argv[])
 {
-  ockam_error_t                    error             = OCKAM_ERROR_NONE;
+  int                              status            = 0;
```
mikelodder7, line 74: Should status have its own `_t` struct?

implementations/c/ockam/codec/tests/payload_aead_aes_gcm_test.c
```diff
@@ -23,19 +23,19 @@ int _test_codec_payload_aead_aes_gcm_setup(void** state)
 
   aag_test_payload = malloc(CODEC_MAX_VLU2_SIZE);
   if (NULL == aag_test_payload) {
-    status = OCKAM_ERROR_INTERFACE_CODEC;
+    status = -1;
```
mikelodder7, line 26: Convert `-1` to a macro or const for readability.

implementations/c/ockam/channel/tests/channel_test.c
SanjoDeundiak: The problem is that now errors are defined by error code and error domain (string that indicates which module returned the error). I think creating separate error domains and codes for tests is overkill (as well as adding separate error codes for tests to our non-test code, which was the case somewhere before this PR). We need to improve tests in many ways, particularly add logs, for errors I decided to just stick with simplest way to make it work. Further work IMHO goes out of scope of this PR.

implementations/c/ockam/codec/tests/payload_aead_aes_gcm_test.c
SanjoDeundiak: See answer above

mrinalwadhwa:
> @mrinalwadhwa any ideas how to do it? I thought this happens because actual GitHub account email (and signing key assigned to it) differs from email I enter while committing (which is working email)

yes, usually an email problem.
try to print and see that it matches the email in the signing key
```
git config user.email
```

SanjoDeundiak:
@mrinalwadhwa fixed

implementations/c/ockam/channel/tests/channel_test.c
mikelodder7: Okay, should we add it as furture work?

mrinalwadhwa:
* +1 on smaller PRs
* Let's wait on merging this till we have @robinbudd's upcoming PR merged first.
* In the meantime I'll also review and get familiar with the new error approach from this PR

mrinalwadhwa:
I think this can be merged after #337 is merged and any conflicts resolved @mikelodder7

PR #332, opened 2020-08-20
Add support for natively implemented vault functions, so elixir can call C

mrinalwadhwa:
1. docs(elixir): add readme with build instructions
2. feat(elixir): add software vault nif and nif compiler
3. test(elixir): disable broken channel test 
(this test is broken because of there is not working vault, we can re-enable it once we add that back in)

This doesn't yet tackle the cmake support needed to find other rust/c modules.


mrinalwadhwa:
#319

PR #337, opened 2020-08-22
feat(c): implement asynchronous reads, multi-hop routes

robinbudd:
- all reads are now asynchronous and require the caller to check return status and poll if the read did not complete.
  see the udp transport test for an example
- channels require a route upon creation
- channels assume that they are running on an edge device, i.e. they do not route/forward messages


mikelodder7:
Merged in c07cdefc4b1966447f038ee5399c06459662e51f

PR #387, opened 2020-09-04
Nifs refactor+fix

SanjoDeundiak:


mrinalwadhwa:
A couple of notes:

The NIFs have to eventually land here, the build etc. is already setup there.
https://github.com/ockam-network/ockam/tree/develop/implementations/e/applications/ockam_vault_software

After this PR is merged could you look at moving existing NIFs there and only add new ones there.
`implementations/e` will take the place of `implementations/elixir` in a few days.

mrinalwadhwa:
Also could you amend the commit message of https://github.com/ockam-network/ockam/pull/387/commits/4f5a5b1b80a3995c08f4c00a76f648a20ca4704b

It doesn't sound like (from the commit message) a feature was added so use either `refactor` or `fix` commit message scope, also don't use a period in the middle of the message.

mrinalwadhwa (approved):
we can merge after the commit message is amended.

mikelodder7:
This appears similar to #385 with a little variation. How should we resolve this

PR #389, opened 2020-09-04
Nifs

SanjoDeundiak:


implementations/elixir/test/ockam/vault/software_test.exs
```diff
@@ -27,9 +27,8 @@ defmodule Ockam.Vault.Software.Tests do
 
   describe "Ockam.Vault.Software.secret_generate/2" do
     test "can run natively implemented functions" do
-      handle = Ockam.Vault.Software.default_init
-      # curve25519 key, ephemeral, key agreement
-      attributes = %{type: 3, persistence: 0, purpose: 0}
+      {:ok, handle} = Ockam.Vault.Software.default_init
+      attributes = %{type: :curve25519, persistence: :ephemeral, purpose: :keyagreement}
```
SanjoDeundiak, line 31: @mrinalwadhwa does that look great?

mrinalwadhwa: yes, that's great 👍  ... one small tweak I would make is to use `:key_agreement` in place of `:keyagreement` .. elixir atoms usually follow that convention.

mrinalwadhwa (approved):

PR #411, opened 2020-09-09
feat(rust): implement basic router, modify message and route types

robinbudd:
- Message: implement std::io::Read & std::io::Write traits
- Message: update Route type to only contain one-way route, i.e.
           separate onward and return routes
- Router: implement basic router functionality and test suite


implementations/rust/router/src/router.rs
```diff
@@ -0,0 +1,168 @@
+#![allow(unused)]
+pub mod router {
+    use ockam_message::message::*;
+    use std::collections::HashMap;
+    use std::io::{Error, ErrorKind, Read, Write};
+    use std::net::UdpSocket;
+    use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+    use std::ops::Add;
+    use std::sync::{Arc, Mutex};
+    use std::thread;
```
mrinalwadhwa, line 10: I think we want the router to be `no_std` and have no knowledge of / dependence on transport protocols /  wire encoding / threading / IO etc. 
https://rust-embedded.github.io/book/intro/no-std.html

implementations/rust/router/src/router.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    impl HandleMessage for UdpAddressHandler {
+        fn handle_message(&self, mut m: Box<Message>) -> Result<(), std::io::Error> {
+            println!("In UdpAddressHandler!");
+            Ok(())
+        }
+    }
+
+    pub trait HandleMessage {
```
mrinalwadhwa, line 23: I'm call this `MessageHandler` in elixir, let's use that same name here.

implementations/rust/router/src/router.rs
```diff
@@ -0,0 +1,168 @@
...
+            Router {
+                registry: HashMap::new(),
+            }
+        }
+
+        pub fn register_handler(
+            &mut self,
+            handler: Arc<Mutex<Box<dyn HandleMessage + Send>>>,
+            address: &Address,
+        ) -> Result<(), String> {
```
mrinalwadhwa, line 42: This function, I think, should set a handler by `address type` not by `address`.
The signature in the elixir implementation is:

```
set_message_handler_for_address_type(address_type, handler)
```

implementations/rust/router/src/router.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        pub fn route(&mut self, mut m: Box<Message>) -> Result<(), String> {
+            // Pop the first address in the list
+            // If there are no addresses, route to the controller
+            // Controller key is always 0
+            let mut key: u64 = 0;
+            let address: Address;
+            if !m.onward_route.addresses.is_empty() {
+                address = m.onward_route.addresses.remove(0);
```
mrinalwadhwa, line 55: I think the router shouldn't remove the address.
It just hands over the message to a handler based on the `address type` of the first address in the `onward_route` of the message.

mrinalwadhwa:
@robinbudd the commit here https://github.com/ockam-network/ockam/pull/411/commits/807c8524dc2b47498fecdfe7d642d229235c9a57 has many more things changing than what is reflected in the commit message. Could you break this down into multiple commits to have the commit messages reflect the changes

implementations/rust/router/src/router.rs
robinbudd: The router has no knowledge of transport. The only reason transport-related stuff is in there is for the tests. The actual router code only knows about addresses and dispatch functions.

What do you mean by "no_std"? If I can't use std, it adds weeks of work to implement OS-independent equivalents for hashmap, threads, and all the transport stuff. What do you propose instead? @mrinalwadhwa

implementations/rust/router/src/router.rs
robinbudd: ok

implementations/rust/router/src/router.rs
robinbudd: Ok, it was an incomplete refactor after the design changes from a couple days ago.

implementations/rust/router/src/router.rs
robinbudd: So the handler removes the address? I think the router should be the only thing messing with the routes. All route management done by the router.

mrinalwadhwa (approved):

PR #415, opened 2020-09-11
feat(elixir): add key exchange to rust <-> elixir ffi

SanjoDeundiak:
This PR adds following:
1. C header with kex interface for nifs
2. C nifs
3. Elixir functions for nifs
4. 1 very basic elixir test
5. Placeholders for all kex rust ffi functions

What needs to be done:
1. Improve storage for ffi instances, so that vault handle could be used as a dependency to rust kex
2. Finish rust ffi layer (should be pretty straightforward after 1 is done)


mrinalwadhwa (approved):
looks great 👍

PR #417, opened 2020-09-13
refactor(rust): remove some std dependencies, update message_handler per design spec

robinbudd:
- router: change registry from hashmap to array, removes std::hashmap
          dependency
- router: remove address from message_handler signature
- router: pass entire route to message handler (don't pop first address)
- address: remove address length from local address, add type to addresses


mrinalwadhwa (approved):

mrinalwadhwa:
rebased and merged https://github.com/ockam-network/ockam/commit/000f0e25098602f5d940ad33bac3e570ee8e36cb

Issue #412, opened 2020-09-10
Create OckamD in Rust


mrinalwadhwa:
Ockam Daemon / OckamD is a daemon program written in rust and distributed as a binary for major operating systems / architectures. (we can start with just linux/macos support and add other operating systems later)

* It opens up raw input/output portals for various protocols.
* A raw input portal - turns a message that is received on the input into an ockam routing protocol message.
* The first one we're interested in supporting is an input portal for STDIN that turn every line that is input on stdin into an ockam routing protocol message.

We want this to be compatible with InfluxDB External Plugin
https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md

Steps:
- [x] Create a crate called daemon in https://github.com/ockam-network/ockam/tree/develop/implementations/rust where the code of ockam daemon will live.
- [x] Add a binary program that  will depend on the `ockam` crate https://github.com/ockam-network/ockam/tree/develop/implementations/rust/ockam
- [x] Add some means to define command line arguments
- [x] Instantiate the Ockam Software Vault via persisted FilesystemVault
- [x] Use routing and secure channels form the `ockam` crate to setup a secure channel with another Ockam node.

mrinalwadhwa:
The CLI and Daemon may be the same binary program that can be started in different modes. (interactive or not) #326

nilslice:
@mrinalwadhwa - any objection to using an async runtime in `ockamd`? Since `telegraf ` is written in Go, I don't think `ockamd` for `telegraf` would run in an environment where `tokio` or another Rust runtime would be off-limits from a resource standpoint... but let me know if you are already aware of cases which that would not be true. 

I was thinking that when `ockamd` is run in interactive-mode, it would send messages over TCP or HTTP via localhost to the `ockamd` running in server-mode. Since `telegraf` will have spawned `ockamd` in server-mode (using STDIN for passing Influx messages) I think it will be more difficult to try and re-use STDIN for interactive-mode -> server-mode process communication. `telegraf` will have a handle to fd/0 on the `ockamd` server-mode process, but `ockamd` interactive-mode would need to look it up each time the binary is run, requiring a different implementation for various OSes. 

Instead of TCP or HTTP, other options I've considered for interactive-mode -> server-mode processes include:
- named pipes (likely also to require OS-specific implementations)
- file on disk
- custom IPC (still needs to keep track of the PID or OS-specific process handle)
[...]

PR #436, opened 2020-09-17
#412 ockamd: initiator implementation, configurable UDP transport, no channel use

nilslice:
midpoint PR for #412 

- initial CLI design and usage documentation that covers both initiator and responder configurations
- includes usage of UDP transport, with minor configuration options including multi-hop
- adds a `FilesystemVault` for local disk persistence 
- uses the persistent vault (not checking for previous vault yet, but in the works)


implementations/rust/daemon/src/args.rs
```diff
@@ -0,0 +1,127 @@
...
+
+impl Default for Args {
+    fn default() -> Self {
+        Self {
+            control: false,
+            control_port: DEFAULT_CONFIG_PORT,
+            input: InputKind::Stdin,
+            output: None, // TODO: should it be more literally named & exposed as such to the user, e.g. "onward_route"?
+            transport: OutputKind::Stdout,
+            public_key: PathBuf::from("ockamd.pub"),
```
nilslice, line 118: @mikelodder7 - with regard to the persistent vault, I think the minimal feature set is effectively the software vault plus a way to save the keys to disk, and load one from disk on initialization. @mrinalwadhwa suggested that we modify the initialization of a vault to optionally take bytes in to be used as the keys (Mrinal, check me on this please). 

The `public_key` field above in the `Args` struct may be incorrect. Would I need to load a key-pair path on disk in order to re-create a vault from a past initialization? If so, I'll update this field. Thanks.

mrinalwadhwa: > modify the initialization of a vault to optionally take bytes in to be used as the keys

Here is what I was imagining
1. the persistent vault is initialized with a directory path.
2. It gets to manage private keys in that path.
3. All files stored in that dir are private keys.
4. The name of the file, is the identifier of that private key.
5. Public keys, I think, need not be stored because they can be derived from private keys.

The above is kinda like the SSH home folder.

implementations/rust/daemon/src/bin/ockamd.rs
```diff
@@ -0,0 +1,26 @@
...
+fn main() {
+    let args = Args::parse();
+    println!("args: {:?}", args);
+
+    // read stdin for messages to encrypt and route
+    let input = stdin();
+    let mut buf = String::new();
+
+    loop {
+        match input.read_line(&mut buf) {
```
mikelodder7, line 14: I have a better opinion on how to do this. But that can come in future PR.

mikelodder7: For example, we should allow input to be passed via a pipe, atty, file, or channel.

mikelodder7 (approved):

implementations/rust/daemon/src/args.rs
mikelodder7: I can help with that @nilslice

implementations/rust/daemon/src/args.rs
```diff
@@ -0,0 +1,127 @@
...
+    Control,
+    /// Used to create a long-running process, to be executed with a particular
+    /// input, e.g. Stdin.
+    Server,
+}
+
+#[derive(Debug, StructOpt)]
+pub enum InputKind {
+    Stdin,
+}
```
nilslice, line 102: @mikelodder7 

> For example, we should allow input to be passed via a pipe, atty, file, or channel.

definitely agreed, and here is where I had left that TBD, where we'd enable the daemon to be configured for those various input sources.

implementations/rust/daemon/src/args.rs
nilslice: Thanks, @mikelodder7. I won't have time to sync up today -- but I'll find some time early next week to discuss. If you are in a position to start, don't let me hold you back :)

nilslice:
@mrinalwadhwa  -- just to check here, can you verify that the below output captured from `$ ockamd -h` handles the various options we want to support in the initial release? 

```
ockamd 0.1.0
Ockam Developers (ockam.io)
Encrypt and route messages using the Ockam daemon.

USAGE:
    ockamd [FLAGS] [OPTIONS]

FLAGS:
    -c, --control    Execute `ockamd` in control-mode, otherwise will start as a long-running process
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -p, --port <control-port>    port for runtime configuration updates [default: 11199]
    -i, --input <input>          data source providing input to `ockamd`, either "stdin" or a valid URI [default: stdin]
        --keys <keys>            path on disk to pre-existing private keys [default: ockamd_keys]
    -o, --output <output>        URI of remote Ockam node, e.g. udp://host:port[/channel_id] [default: stdout]
```

I'm not certain the `channel_id` is useful, but it would be the bit of information used in your demo as the "Channel Responder Address".

implementations/rust/daemon/src/cli.rs
```diff
@@ -0,0 +1,366 @@
...
+impl Args {
+    /// Parse the command line options into the Args struct.
+    pub fn parse() -> Args {
+        let mut args = Args::from_args();
+
+        // validate provided arguments & override possibly fallible options
+        match args.output_kind() {
+            OutputKind::Channel(_) => {
+                // disallow output to be decrypted if it's to be sent over a secure channel
+                if args.decrypt_output {
```
mikelodder7, line 159: Mostly curious about why this is the case.

implementations/rust/daemon/src/cli.rs
```diff
@@ -0,0 +1,366 @@
...
+
+use ockam_message::message::{Route, RouterAddress};
+
+use structopt::{clap::ArgSettings::Hidden, StructOpt};
+use url::Url;
+
+/// The port on which the config updater runs and accepts Config messages.
+pub const DEFAULT_CONFIG_PORT: u16 = 11199;
+
+const DEFAULT_LOCAL_SOCKET: &str = "127.0.0.1:11200";
```
mikelodder7, line 14: Are the ports chosen randomly?

implementations/rust/daemon/src/cli.rs
```diff
@@ -0,0 +1,366 @@
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "FILESYSTEM" => Ok(VaultKind::Filesystem),
+            _ => Err("currently, 'FILESYSTEM' is the only supported vault option".into()),
+        }
+    }
+}
+
+/// Specifies which end of the secure channel the instance of `ockamd` is prepared to run in.
+#[derive(Clone, Debug, StructOpt, PartialEq)]
```
mikelodder7, line 215:
```suggestion
#[derive(Clone, Copy, Debug, StructOpt, PartialEq)]
```

implementations/rust/daemon/src/cli.rs
```diff
@@ -0,0 +1,366 @@
...
+        match s {
+            "initiator" | "init" => Ok(ChannelRole::Initiator),
+            "responder" | "resp" => Ok(ChannelRole::Responder),
+            _ => Err("role must be set to either 'initiator' or 'responder'".into()),
+        }
+    }
+}
+
+/// The mode in which `ockamd` is to be run.
+#[derive(Debug, StructOpt)]
```
mikelodder7, line 238:
```suggestion
#[derive(Clone, Copy, Debug, StructOpt)]
```

implementations/rust/daemon/src/vault.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub fn new(path: PathBuf) -> std::io::Result<Self> {
+        let create_path = path.clone();
+        std::fs::create_dir_all(create_path)?;
+
+        let mut vault = DefaultVault::default();
+        let to_secret = |path: PathBuf,
+                         data: &[u8]|
+         -> Result<(SecretKey, SecretKeyAttributes), VaultFailError> {
+            let attrs = SecretKeyAttributes {
+                xtype: SecretKeyType::Buffer(data.len()),
```
mikelodder7, line 22: Why would all the keys be `buffer`?

implementations/rust/daemon/src/vault.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    /// Close and release all resources in use by the vault
+    fn deinit(&mut self) {
+        self.v.deinit()
+    }
+}
+
+impl Zeroize for FilesystemVault {
+    fn zeroize(&mut self) {}
```
mikelodder7, line 212:
```suggestion
    fn zeroize(&mut self) {
        self.v.zeroize();
    }
```

implementations/rust/vault/src/lib.rs
```diff
@@ -72,7 +72,7 @@ pub trait Vault: Zeroize {
     ) -> Result<SecretKeyContext, VaultFailError>;
     /// Export a secret key from the vault
     fn secret_export(&mut self, context: SecretKeyContext) -> Result<SecretKey, VaultFailError>;
-    /// Set the attributes for a secret key
+    /// Get the attributes for a secret key
```
mikelodder7, line 75: Good catch!

implementations/rust/vault/src/lib.rs
```diff
@@ -149,7 +149,7 @@ pub trait DynVault {
...
```
mikelodder7, line 152: Good catch!

implementations/rust/daemon/src/vault.rs
nilslice: @mikelodder7 - ah, they shouldn't be -- I meant to add a `TODO` here and ask how to check which key type the file would be... is there a byte or value in the stored key that indicates its type?

implementations/rust/daemon/src/cli.rs
nilslice: Good callout 😆 a couple reason: In a demo, it is useful to be able to show the data in both states, and as a requirement for the near-term deliverable we need to just write cleartext messages to `stdout` on the responder end.

implementations/rust/daemon/src/vault.rs
nilslice: Also, I don't believe they should all be for `purpose: SecretPurposeType::KeyAgreement`.. I might have to encode these bits of data into the file or the filename somehow.

nilslice: @mikelodder7  -- unless you suggest otherwise, I think I'm going to encode the following `usize` values as a part of the filename using these fn's below. For example, a key with context 1, of type `Aes256`, purpose `KeyAgreement` and persistence `Persistent` would be saved to disk named:

`./ockamd_vault/1_2_0_1` 


```rust
impl SecretKeyType {
    /// Convert enum to a number
    pub fn to_usize(&self) -> usize {
        match *self {
            SecretKeyType::Buffer(..) => 0,
            SecretKeyType::Aes128 => 1,
            SecretKeyType::Aes256 => 2,
            SecretKeyType::Curve25519 => 3,
            SecretKeyType::P256 => 4,
        }
    }

    /// Try to convert from a number to the rust enum
    pub fn from_usize(value: usize) -> Result<Self, VaultFailError> {
        match value {
            0 => Ok(SecretKeyType::Buffer(0)),
            1 => Ok(SecretKeyType::Aes128),
            2 => Ok(SecretKeyType::Aes256),
            3 => Ok(SecretKeyType::Curve25519),
            4 => Ok(SecretKeyType::P256),
            _ => Err(VaultFailErrorKind::InvalidParam(0).into()),
        }
    }
}
```

nilslice: or prefix/append them to the bytes in the file, but I think I'd prefer not to mess with the key itself. what do you think?

implementations/rust/daemon/src/cli.rs
robinbudd: That *should* work, I haven't tested it though.

robinbudd (approved):
Looks good overall. When the Channel is ready you'll need to add it to the polling loop of course. Next step might be to add a Worker.

implementations/rust/daemon/src/vault.rs
mikelodder7: In the OSX implementation I just take the `to_bytes` method for the attributes and put that as the first bytes with the key data after. We went over this once before but it was probably a lot of info at the time. 

See [here](https://github.com/ockam-network/ockam/blob/ml/channel-updates/implementations/rust/vault/src/osx/mod.rs#L197) and [here](https://github.com/ockam-network/ockam/blob/ml/channel-updates/implementations/rust/vault/src/osx/mod.rs#L149) for examples

implementations/rust/daemon/src/node.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        let route_config = config.clone();
+        let remote_addr = route_config
+            .onward_route
+            .expect("invalid address provided for output")
+            .clone();
+        thread::spawn(move || {
+            loop {
+                if let Ok(data) = rx_in.recv() {
```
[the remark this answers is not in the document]
mikelodder7: Right now you're ignoring `Err`. What I mean here is shouldn't you handle it with `eprintln` for now? I agree a logger in the future would be better

implementations/rust/daemon/src/cli.rs
mikelodder7: What I mean is where did `11199` and `11200` come from? Did you randomly pick them? Sorry for not being clear earlier.

implementations/rust/daemon/src/vault.rs
nilslice: Thanks, @mikelodder7 - I do remember that now. I'll follow suit!

implementations/rust/daemon/src/node.rs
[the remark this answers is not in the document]
nilslice: I see what you're saying. Sorry, I was looking at a different `eprintln!` use inside the `if let` block. Yes, I'll print the error, but now that you've brought my attention to it, I wonder if this should actually terminate the program after printing the error.. otherwise I think it will just print errors repeatedly until the process is killed.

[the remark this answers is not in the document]
nilslice:
```rust
eprintln!("fatal error: failed to read from input");
std::os::exit(1);
```

basically, if this CLI stdin reader can't read from it's input channel then it's no use running at all... os::exit(1) is probably the easy way out, but happy to consider alternatives if you have a suggestion.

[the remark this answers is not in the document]
mikelodder7: That sounds like a great solution.

implementations/rust/daemon/src/cli.rs
nilslice: haha oh yes, I randomly picked those.. 111 = ASCII `o`, 99 = ASCII `c` (i.e. **oc**kam). I changed the code to use a any available port though, but it can still be configured at the CLI.

mikelodder7 (approved):

PR #443, opened 2020-09-18
refactor(rust): refactor address type, add commands

robinbudd:
- rename Address to RouterAddress, indicating the router can route it
- pull AddressType and Length out of Address, into RouterAddress
- make RouterAddres.address an enum
- define Commands for inter-module messaging
- note that Router and Transport are currently broken due to new Message type.
-- they have been removed from the build and will be repaired soon

### DIFF
--- a/implementations/rust/Cargo.toml
+++ b/implementations/rust/Cargo.toml
@@ -6,6 +6,7 @@ members = [
     "kex",
     "message",
     "vault",
-    "router",
-    "transport",
+    # "router",
+    # "transport",
+    # "node",
 ]
--- a/implementations/rust/message/src/message.rs
+++ b/implementations/rust/message/src/message.rs
@@ -12,21 +12,20 @@ pub mod message {
     use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
     use std::ops::Add;
     use std::slice;
-    use std::sync::atomic::Ordering::AcqRel;
 
     const WIRE_PROTOCOL_VERSION: u8 = 1;
 
     pub trait Codec {
         type Inner;
 
-        fn encode(t: &Self::Inner, v: &mut Vec<u8>) -> Result<(), String>;
+        fn encode(t: Self::Inner, v: &mut Vec<u8>) -> Result<(), String>;
         fn decode(s: &[u8]) -> Result<(Self::Inner, &[u8]), String>;
         fn decode_boxed(s: &[u8]) -> Result<(Box<Self::Inner>, &[u8]), String> {
             Err("not implemented".to_string())
         }
     }
 
-    #[derive(Debug)]
+    // #[derive(Debug)]
     #[repr(C)]
     pub struct Message {
         pub onward_route: Route,
@@ -45,39 +44,15 @@ pub mod message {
     }
 
     impl Codec for Message {
-        type Inner = Message;
-        fn encode(msg: &Message, u: &mut Vec<u8>) -> Result<(), String> {
-            Route::encode(&msg.onward_route, u);
-            Route::encode(&msg.return_route, u);
+        type Inner = Box<Message>;
+        fn encode(msg: Box<Message>, u: &mut Vec<u8>) -> Result<(), String> {
+            Route::encode(msg.onward_route, u);
+            Route::encode(msg.return_route, u);
             u.extend(&msg.message_body[0..]);
             Ok(())
         }
 
-        fn decode(u: &[u8]) -> Result<(Message, &[u8]), String> {
-            let mut msg = Message::default();
-            let mut w = u;
-            match Route::decode(w) {
-                Ok((r, u1)) => {
-                    msg.onward_route = r;
-                    w = u1;
-                }
-                Err(s) => {
-                    return Err(s);
-                }
-            }
-            match Route::decode(w) {
-                Ok((r, u1)) => {
-                    msg.return_route = r;
-                    w = u1;
-                }
-                Err(s) => {
-                    return Err(s);
-                }
-            }
-            msg.message_body.append(&mut (w.to_vec()));
-            Ok((msg, w))
-        }
-        fn decode_boxed(u: &[u8]) -> Result<(Box<Message>, &[u8]), String> {
+        fn decode(u: &[u8]) -> Result<(Box<Message>, &[u8]), String> {
             let mut msg = Box::new(Message::default());
             let mut w = u;
             match Route::decode(w) {
@@ -98,6 +73,7 @@ pub mod message {
                     return Err(s);
                 }
             }
+
             msg.message_body.append(&mut (w.to_vec()));
             Ok((msg, w))
         }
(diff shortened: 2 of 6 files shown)


implementations/rust/Cargo.toml
```diff
@@ -6,6 +6,7 @@ members = [
     "kex",
     "message",
     "vault",
-    "router",
-    "transport",
+    # "router",
```
mikelodder7, line 9: what's the reason these are commented out?

implementations/rust/message/src/message.rs
```diff
@@ -12,21 +12,20 @@ pub mod message {
...
     use std::slice;
-    use std::sync::atomic::Ordering::AcqRel;
 
     const WIRE_PROTOCOL_VERSION: u8 = 1;
 
     pub trait Codec {
         type Inner;
 
-        fn encode(t: &Self::Inner, v: &mut Vec<u8>) -> Result<(), String>;
+        fn encode(t: Self::Inner, v: &mut Vec<u8>) -> Result<(), String>;
```
mikelodder7, line 21:
```suggestion
        fn encode(&self, v: &mut Vec<u8>) -> Result<(), String>;
```

implementations/rust/message/src/message.rs
```diff
@@ -45,39 +44,15 @@ pub mod message {
     }
 
     impl Codec for Message {
-        type Inner = Message;
-        fn encode(msg: &Message, u: &mut Vec<u8>) -> Result<(), String> {
-            Route::encode(&msg.onward_route, u);
-            Route::encode(&msg.return_route, u);
+        type Inner = Box<Message>;
```
mikelodder7, line 47: Why a Box?

implementations/rust/message/src/message.rs
```diff
@@ -45,39 +44,15 @@ pub mod message {
...
+        fn encode(msg: Box<Message>, u: &mut Vec<u8>) -> Result<(), String> {
```
mikelodder7, line 48:
```suggestion
        fn encode(msg: Box<Message>, u: &mut [u8]) -> Result<(), String> {
```

implementations/rust/message/src/message.rs
```diff
@@ -98,17 +73,49 @@ pub mod message {
...
             }
+
             msg.message_body.append(&mut (w.to_vec()));
             Ok((msg, w))
         }
     }
 
     /* Addresses */
     #[repr(C)]
+    #[derive(Debug, PartialEq)]
```
mikelodder7, line 84:
```suggestion
    #[derive(Debug, PartialEq, Copy, Clone)]
```

implementations/rust/message/src/message.rs
```diff
@@ -98,17 +73,49 @@ pub mod message {
...
+    #[repr(C)]
+    #[derive(Clone, Copy)]
```
mikelodder7, line 86:
```suggestion

```

implementations/rust/message/src/message.rs
```diff
@@ -98,17 +73,49 @@ pub mod message {
...
+        ChannelAddress(u32),
+    }
+
+    pub enum HostAddressType {
+        Ipv4 = 0,
+        Ipv6 = 1,
+    }
+
     pub enum AddressType {
+        Undefined = 255,
```
mikelodder7, line 114: Could we change this to use Option instead where Undefined is None and everything else is Some?

implementations/rust/message/src/message.rs
```diff
@@ -192,56 +186,87 @@ pub mod message {
...
+            match AddressType::try_from(u[0]) {
+                Ok(t) => {
+                    a_type = t;
+                }
+                Err(s) => return Err(s),
+            }
+            match a_type {
+                AddressType::Channel => {
+                    let address =
+                        Address::ChannelAddress(u32::from_le_bytes([u[2], u[3], u[4], u[5]]));
```
mikelodder7, line 239: TIL: we are using little-endian.

implementations/rust/message/src/message.rs
```diff
@@ -645,120 +720,132 @@ mod tests {
...
-            AddressType::Local,
-            LocalAddress {
-                address: 0x00010203,
-            },
-        ));
-        let onward_route = Route {
-            addresses: onward_addresses,
+        let mut onward_route = Route { addresses: vec![] };
+        let mut udp_address = Address::UdpAddress(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 0x8080);
+        let mut router_address = RouterAddress {
```
mikelodder7, line 725: For future reference. You can also use "127.0.0.1".parse()

implementations/rust/message/src/message.rs
```diff
@@ -645,120 +720,132 @@ mod tests {
...
+            a_type: AddressType::Channel,
+            length: 0,
+            address: channel_address,
+        };
+        router_address.length = router_address.size_of();
+        return_route.addresses.push(router_address);
+
         let mut message_body = vec![0];
-        let mut msg = Message {
+        let mut msg = Box::new(Message {
```
mikelodder7, line 780: Wish I understood why Box is necessary.

implementations/rust/Cargo.toml
robinbudd: Not checked in, in this PR.

implementations/rust/message/src/message.rs
robinbudd: Because I thought it was necessary due to how I was passing messages between modules and possibly threads. I've already removed the Box aspect (it's in the next PR).

implementations/rust/message/src/message.rs
robinbudd: It's a Vec because its size is unknown up-front, it needs to be growable.

implementations/rust/message/src/message.rs
robinbudd: Already done in next PR

implementations/rust/Cargo.toml
robinbudd: ...meaning, it wasn't meant to be part of this PR

implementations/rust/message/src/message.rs
robinbudd: That doesn't compile

implementations/rust/message/src/message.rs
robinbudd: We could, but then we have to pattern-match every time we want to use AddressType. I'm making the call that it isn't worth it, at least not now.

implementations/rust/message/src/message.rs
robinbudd: ?

implementations/rust/message/src/message.rs
robinbudd: Don't you though. I addressed that earlier, and Box is gone inn the next PR.

mikelodder7:
Merged with b0d1e38

PR #453, opened 2020-09-26
refactor(rust): make address types consistent, make udp_transport do all comms on same socket

robinbudd:
- Message: now support the following types: Udp, Channel, Worker
- Message: has a new field, message_type, to indicate what is in the message
           body. The following types are recognized: Ping(0), Pong(1), Payload(2),
           KeyAgreementM1..M3(3,4,..)
- UdpTransport: now performs all io over the same socket. When creating a
                new transport, provide the local address/port to bind to.
- UdpTransport: The "Add" command is no longer supported. The transport will
                attempt to send the message to whatever address is in the
                route, from the socket it is bound to


mrinalwadhwa:
@robinbudd rebase to the latest develop should fix the azure pipeline that is stuck for this pr

robinbudd:
This will be superseded by a future PR

PR #455, opened 2020-09-28
refactor(rust): make address types consistent, do all comms on same s…

robinbudd:
…ocket

- Message: now support the following types: Udp, Channel, Worker
- Message: has a new field, message_type, to indicate what is in the message
  body. The following types are recognized: Ping(0), Pong(1), Payload(2),
  KeyAgreementM1..M3(3,4,..)
- UdpTransport: now performs all io over the same socket. When creating a
  new transport, provide the local address/port to bind to.
- UdpTransport: The "Add" command is no longer supported. The transport will
  attempt to send the message to whatever address is in the
  route, from the socket it is bound to
- Node: Redesign the node test to support being run as a separate process.
  Node now takes command line arguments as follows:
  "-l xxx.xxx.xxx.xxx:pppp" - binds local socket to given address (required)
  "-v xxx.xxx.xxx.xxx:pppp" - routes message via this address
  "-r xxx.xxx.xxx.xxx:pppp" - remote (destination) IP address
  "-w xxx..." - worker address
  To wait for incoming message (e.g. responder) you only need "-l". For example,
    ./node -l 127.0.0.1:4052
  To create a route for a message, "-l" and "-r" are required, with "-v" and "-w"
    optional. For example,
    ./node -l 127.0.0.1:4050 -v 127.0.0.1:4051 -r 127.0.0.1:4052 -w 1234
    will create a route 127.0.0.1:4050->127.0.0.1:4051->1234

### DIFF
--- a/implementations/rust/message/src/lib.rs
+++ b/implementations/rust/message/src/lib.rs
@@ -1,517 +1,593 @@
+#![allow(unused)]
+
 // Definition and implementation of an Ockam message and message components.
 // Each message component, and the message overall, implements the "Codec" trait
 // allowing it to be encoded/decoded for transmission over a transport.
 
-use std::{
-    convert::TryFrom,
-    fmt::Formatter,
-    io::ErrorKind,
-    net::{IpAddr, Ipv4Addr, SocketAddr},
-    str::FromStr,
-};
+pub mod message {
+    use crate::message::Address::ChannelAddress;
+    use crate::message::MessageType::Payload;
+    use hex::*;
+    use std::convert::{Into, TryFrom};
+    use std::error::Error;
+    use std::fmt::Formatter;
+    pub use std::io::{ErrorKind, Read, Write};
+    use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
+    use std::ops::Add;
+    use std::slice;
+    use std::str::FromStr;
 
-pub const WIRE_PROTOCOL_VERSION: u8 = 1;
+    const WIRE_PROTOCOL_VERSION: u8 = 1;
 
-pub trait Codec {
-    type Inner;
+    pub trait Codec {
+        type Inner;
 
-    fn encode(t: Self::Inner, v: &mut Vec<u8>) -> Result<(), String>;
-    fn decode(s: &[u8]) -> Result<(Self::Inner, &[u8]), String>;
-}
-
-// #[derive(Debug)]
-#[repr(C)]
-pub struct Message {
-    pub onward_route: Route,
-    pub return_route: Route,
-    pub message_body: Vec<u8>,
-}
-
-impl Default for Message {
-    fn default() -> Message {
-        Message {
-            onward_route: Route { addresses: vec![] },
-            return_route: Route { addresses: vec![] },
-            message_body: vec![0],
-        }
-    }
-}
-
-impl Codec for Message {
-    type Inner = Message;
-    fn encode(msg: Message, u: &mut Vec<u8>) -> Result<(), String> {
-        Route::encode(msg.onward_route, u)?;
-        Route::encode(msg.return_route, u)?;
-        u.extend(&msg.message_body[0..]);
-        Ok(())
+        fn encode(t: Self::Inner, v: &mut Vec<u8>) -> Result<(), String>;
+        fn decode(s: &[u8]) -> Result<(Self::Inner, &[u8]), String>;
     }
 
-    fn decode(u: &[u8]) -> Result<(Message, &[u8]), String> {
-        let mut msg = Message::default();
-        let mut w = u;
-        match Route::decode(w) {
-            Ok((r, u1)) => {
-                msg.onward_route = r;
-                w = u1;
-            }
-            Err(s) => {
-                return Err(s);
+    // #[derive(Debug)]
+    #[repr(C)]
+    pub struct Message {
+        pub onward_route: Route,
+        pub return_route: Route,
+        pub message_type: MessageType,
+        pub message_body: Vec<u8>,
+    }
+
+    #[repr(u8)]
+    pub enum MessageType {
+        Ping = 0,
+        Pong = 1,
+        Payload = 2,
+        KeyAgreementM1 = 3,
+        KeyAgreementM2 = 4,
+        KeyAgreementM3 = 5,
+    }
+
+    impl Default for Message {
+        fn default() -> Message {
+            Message {
+                onward_route: Route { addresses: vec![] },
+                return_route: Route { addresses: vec![] },
+                message_type: Payload,
+                message_body: vec![0],
             }
         }
-        match Route::decode(w) {
-            Ok((r, u1)) => {
-                msg.return_route = r;
-                w = u1;
-            }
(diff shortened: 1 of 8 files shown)


implementations/rust/message/src/lib.rs
```diff
@@ -1,517 +1,593 @@
...
 // allowing it to be encoded/decoded for transmission over a transport.
 
-use std::{
-    convert::TryFrom,
-    fmt::Formatter,
-    io::ErrorKind,
-    net::{IpAddr, Ipv4Addr, SocketAddr},
-    str::FromStr,
-};
+pub mod message {
```
nilslice, line 7: Now that the file is in `message/src/lib/rs` we should omit the `pub mod message` scope wrapping the message code. This is the "path shortening" concern @mikelodder7 mentioned. 

From a user standpoint, including this mod wrapper will require typing `use ockam_message::message::Message` rather than `use ockam_message::Message`. 

I'd suggest we remove the mod wrapper, but am open to alternate feedback.

implementations/rust/message/src/lib.rs
```diff
@@ -1,517 +1,593 @@
...
-    fn decode(u: &[u8]) -> Result<(SocketAddr, &[u8]), String> {
-        match (HostAddressType::try_from(u[0])?, &u[1..]) {
-            (HostAddressType::Ipv4, addr) => {
-                let ip4 = Ipv4Addr::new(addr[0], addr[1], addr[2], addr[3]);
-                let port = u16::from_le_bytes([addr[4], addr[5]]);
-                let sock = SocketAddr::new(IpAddr::V4(ip4), port);
-                Ok((sock, &addr[6..]))
+
+    impl Route {
+        pub fn print_route(&self) {
```
nilslice, line 378: May benefit from implementing `std::fmt::Display` for the different route address types, so anyone can `println!` them.

implementations/rust/node/src/node.rs
```diff
@@ -115,78 +141,161 @@ pub fn get_route() -> Option<Route> {
...
-    match join_thread.join() {
-        Ok(_unused) => {}
-        Err(_unused) => {}
+fn main() {
+    let args = Args::from_args();
+    println!("{:?}", args);
+    let local_socket: RouterAddress;
+    let route: Route;
+    let message: String;
+    match parse_args(args) {
```
nilslice, line 286: Not that important, but you could remove all the let bindings, and pass the tuple values from `Ok` match arm into the `start_node` call and sleep from there. It's common in Rust to try to avoid doing a declaration without assignment. Even though the compiler will reject it, this could ultimately lead to using uninitialized vars.

implementations/rust/router/src/lib.rs
```diff
@@ -1,80 +1,101 @@
-use ockam_common::commands::ockam_commands::*;
-use ockam_message::*;
+#![allow(unused)]
+pub mod router {
```
nilslice, line 2: Same comment as above re: `pub mod router` namespace wrapping -- since this is in `router/src/lib.rs` it's already in the `router` module.

implementations/rust/transport/src/lib.rs
```diff
@@ -1,153 +1,160 @@
...
-};
+#[allow(unused)]
 
-pub struct UdpTransport {
-    connections: HashMap<String, UdpConnection>,
-    rx: Receiver<OckamCommand>,
-    tx: Sender<OckamCommand>,
-    router_tx: Sender<OckamCommand>,
-}
+pub mod transport {
```
nilslice, line 3: Same comment as above for `pub mod transport` wrapper :)

nilslice:
Awesome! If you agree with the module layout suggestions (un-nesting the `pub mod transport` etc), then the other one would be `ockam_common::commands::ockam_commands`, which could use the same un-nesting. Unless we think we will have other scoped commands in the future... but I suppose those shouldn't be in the `ockam_common` crate then. 

Sorry to pile on all of this un-nesting stuff :) Really looks great otherwise, and the updates to the node example are great. I'm going to borrow the Node name in `ockamd`...

implementations/rust/message/src/lib.rs
```diff
@@ -1,517 +1,593 @@
...
 
-impl TryFrom<u8> for HostAddressType {
-    type Error = String;
-    fn try_from(data: u8) -> Result<Self, Self::Error> {
-        match data {
-            0 => Ok(HostAddressType::Ipv4),
-            1 => Ok(HostAddressType::Ipv6),
-            _ => Err("Unknown host address type".to_string()),
+    impl TryFrom<u8> for MessageType {
+        type Error = String;
```
mikelodder7, line 203:
```suggestion
        type Error = &'static str;
```

implementations/rust/message/src/lib.rs
```diff
@@ -1,517 +1,593 @@
...
+            }
+        }
+        pub fn channel_router_address_from_str(a: &str) -> Result<RouterAddress, String> {
+            match hex::decode(a) {
+                Ok(h) => Ok(RouterAddress {
+                    a_type: AddressType::Channel,
+                    length: h.len() as u8,
+                    address: Address::ChannelAddress(h),
+                }),
+                Err(_unused) => Err("string contains non-hex chars".to_string()),
```
mikelodder7, line 439:
```suggestion
                Err(_) => Err("string contains non-hex chars".to_string()),
```

implementations/rust/message/src/lib.rs
```diff
@@ -1,517 +1,593 @@
...
+            }
+        }
+        pub fn worker_router_address_from_str(a: &str) -> Result<RouterAddress, String> {
+            match hex::decode(a) {
+                Ok(h) => Ok(RouterAddress {
+                    a_type: AddressType::Worker,
+                    length: h.len() as u8,
+                    address: Address::WorkerAddress(h),
+                }),
+                Err(_unused) => Err("string contains non-hex chars".to_string()),
```
mikelodder7, line 449:
```suggestion
                Err(_) => Err("string contains non-hex chars".into()),
```

implementations/rust/message/src/lib.rs
```diff
@@ -1,517 +1,593 @@
...
-    }
-    pub fn channel_router_address_from_u32(a: u32) -> Result<RouterAddress, String> {
-        Ok(RouterAddress {
-            a_type: AddressType::Channel,
-            length: 4,
-            address: Address::ChannelAddress(a),
-        })
+    /* Routes */
+    #[repr(C)]
+    pub struct Route {
```
mikelodder7, line 456:
```suggestion
    #[derive(Debug)]
    pub struct Route {
```

nilslice:
@robinbudd, @mikelodder7 & I spoke about this and will merge as-is. Robin is going to create some new issues and each will mention this PR so they are associated on Github (by adding unquoted `#455` as text in the Issue description).

nilslice (approved):

mikelodder7 (approved):
Thanks for capturing in issues

implementations/rust/message/src/lib.rs
robinbudd: Issue opened

PR #469, opened 2020-09-30
feat(rust): add c binding

SanjoDeundiak:


implementations/rust/c/bindings/src/ctypes.rs
```diff
@@ -0,0 +1,10 @@
+#![allow(non_camel_case_types)]
```
mikelodder7, line 1: Not sure if this is necessary with [this](https://doc.rust-lang.org/std/os/raw/index.html).
Can we use the built-in rust ones instead?

implementations/rust/c/rust_memory/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+    ERROR_NONE
+}
+
+unsafe extern "C" fn memcpy_impl(
+    _: *mut ockam_memory_t,
+    dst: *mut core::ffi::c_void,
+    src: *const core::ffi::c_void,
+    size: usize,
+) -> ockam_error_t {
+    core::intrinsics::copy_nonoverlapping(src, dst, size);
```
mikelodder7, line 112: Is there a reason we're using `core::intrinsics` vs [std::ptr](https://doc.rust-lang.org/beta/std/ptr/index.html)?

`core::instrinsics` only works on nightly and I believe we are targeting stable

implementations/rust/c/rust_memory/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+    ERROR_NONE
+}
+
+unsafe extern "C" fn memmove_impl(
+    _: *mut ockam_memory_t,
+    dst: *mut core::ffi::c_void,
+    src: *mut core::ffi::c_void,
+    size: usize,
+) -> ockam_error_t {
+    core::intrinsics::copy(src, dst, size);
```
mikelodder7, line 122: Same as before.

mikelodder7 (commented):
Great PR! Just a few questions to answer then I think it can be merged.

implementations/rust/c/bindings/src/ctypes.rs
SanjoDeundiak: Yes, fixed

implementations/rust/c/rust_memory/src/lib.rs
SanjoDeundiak: Changed to std::ptr

implementations/rust/c/rust_memory/src/lib.rs
SanjoDeundiak: +

PR #472, opened 2020-10-02
chore(rust): add daemon to default-members list for cargo

nilslice:
Adds `daemon` crate to new list of `default-members` in the workspace

This allows more convenient use of `cargo`

### DIFF
--- a/implementations/rust/Cargo.toml
+++ b/implementations/rust/Cargo.toml
@@ -18,6 +18,7 @@ members = [
 default-members = [
     "channel",
     "common",
+    "daemon",
     "kex",
     "message",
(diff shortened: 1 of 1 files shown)


SanjoDeundiak:
@nilslice @mrinalwadhwa @robinbudd @mikelodder7  this is due to my change, now every new workspace member should be added to default-members as well, that's because we don't want to run generate-c-bindings on every cargo build command. Should have warned you all earlier.

nilslice:
no problem! thanks, @SanjoDeundiak!

mikelodder7 (approved):

PR #476, opened 2020-10-02
Finished first draft for channels

mikelodder7:
First implementation of channels.
Doesn't handle key rotation yet but will in a future PR.

### DIFF
--- a/implementations/rust/channel/src/lib.rs
+++ b/implementations/rust/channel/src/lib.rs
@@ -23,6 +23,7 @@
 extern crate ockam_common;
 
 use error::*;
+use failure::_core::marker::PhantomData;
 use ockam_common::commands::ockam_commands::{ChannelCommand, OckamCommand, RouterCommand};
 use ockam_kex::{CompletedKeyExchange, KeyExchanger, NewKeyExchanger};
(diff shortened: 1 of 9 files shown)


robinbudd (approved):

implementations/rust/channel/src/lib.rs
```diff
@@ -23,54 +23,405 @@
...
-use std::io::{Read, Write};
-
-/// Represents an Ockam channel for reading and writing payloads
-#[derive(Debug)]
-pub struct Channel<R: Read, W: Write, KX: KeyExchanger> {
-    exchange_data: Option<CompletedKeyExchange>,
-    key_exchanger: KX,
-    reader: R,
-    writer: W,
+use failure::_core::marker::PhantomData;
```
nilslice, line 26: `
```suggestion
use core::marker::PhantomData;
```
I will update and push

nilslice:
Once builds pass, I think this is good to go. There are a couple spots where @robinbudd  might want to handle the `unwrap` usage, but nice work overall!

nilslice (approved):

nilslice:
@mikelodder7 -- can you check that you have whatever the CLion equivalent of "Format on Save" enabled?

PR #479, opened 2020-10-04
feat(rust): use basic worker to process message

nilslice:
To test, use these commands from `implementations/rust`:
---
no transport, echo stdin -> stdout:
```sh
cargo  run --bin ockamd -- \
        --role=init \
        --output=stdout \
        --channel-responder-address=33333333 \
        --worker-address=01242020 \
        --identity-name=default \
        --responder-public-key=1234
```
---
over UDP transport, no secure channels:
1. start ockamd,
```sh
cargo  run --bin ockamd -- \
        --role=init \
        --output=udp://127.0.0.1:34254 \
        --channel-responder-address=33333333 \
        --worker-address=01242020 \
        --identity-name=default \
        --responder-public-key=1234
```
2. and separately, run:
```rust
use std::net::UdpSocket;

fn main() {
        let socket = UdpSocket::bind("127.0.0.1:34254").expect("failed to bind socket address");
	println!("udp socket: 127.0.0.1:34254");

	loop {
		let mut buf = [0; 32];
                socket.recv_from(&mut buf).expect("failed to read from socket");

		println!("got: {:?}", buf);
	}
}
```

### DIFF
--- a/implementations/rust/daemon/src/worker.rs
+++ b/implementations/rust/daemon/src/worker.rs
@@ -1,0 +1,12 @@
+use std::sync::mpsc::{self, Receiver, Sender};
+
+use ockam_common::commands::ockam_commands::*;
+
+pub struct Worker {
+    router_tx: Sender<OckamCommand>,
+    rx: Receiver<OckamCommand>,
+}
+
+impl Worker {
+    fn new(router_tx: Sender<OckamCommand>, rx: Receiver<OckamCommand>) -> Self {}
+}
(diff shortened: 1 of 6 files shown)


implementations/rust/daemon/src/worker.rs
```diff
@@ -0,0 +1,45 @@
...
+                        eprintln!("worker-error: bad data");
+                    }
+                    true
+                }
+                _ => {
+                    eprintln!("unrecognized worker command");
+                    false
+                }
+            },
+            Err(_) => true,
```
mikelodder7, line 42:
```suggestion
            Err(e) => {
                eprintln!("{:?}", e)
                true
            },
```

mikelodder7 (changes requested):
Great PR!
Either reply to comments or incorporate them.

nilslice:
Thanks, @mikelodder7 -- great feedback, PR updated.

PR #481, opened 2020-10-05
feat(rust): make sensitive data secrets

SanjoDeundiak:


implementations/rust/kex/src/xx.rs
```diff
@@ -106,31 +111,43 @@ impl KeyExchange for SymmetricState {
         &mut self,
         secret_handle: SecretKeyContext,
         public_key: PublicKey,
-    ) -> Result<Vec<u8>, VaultFailError> {
+    ) -> Result<Vec<SecretKeyContext>, VaultFailError> {
         let mut vault = self.vault.lock().unwrap();
```
nilslice, line 115: could we handle this w/o the `unwrap()`?

implementations/rust/kex/src/xx.rs
```diff
@@ -106,31 +111,43 @@ impl KeyExchange for SymmetricState {
...
+            purpose: SecretPurposeType::KeyAgreement,
+            persistence: SecretPersistenceType::Ephemeral,
+        };
+
         vault.ec_diffie_hellman_hkdf_sha256(
             secret_handle,
             public_key,
-            self.state.ck.as_ref(),
-            SHA256_SIZE + AES256_KEYSIZE,
+            self.state.ck.unwrap(),
```
nilslice, line 132: also here -- can this value be handled before passing it into the `ec_diffie_hellman_hkdf_sha256` fn?

implementations/rust/kex/src/xx.rs
```diff
@@ -106,31 +111,43 @@ impl KeyExchange for SymmetricState {
...
-            purpose: SecretPurposeType::KeyAgreement,
-        };
-        let key = SecretKey::Aes256(*array_ref![hash, SHA256_SIZE, AES256_KEYSIZE]);
+    fn mix_key(&mut self, hkdf_output: Vec<SecretKeyContext>) -> Result<(), VaultFailError> {
         let mut vault = self.vault.lock().unwrap();
+
+        if self.state.ck.is_some() {
+            vault.secret_destroy(self.state.ck.unwrap())?;
+        }
+        self.state.ck = Some(hkdf_output[0]);
```
nilslice, line 144: prefer use of `.get(0)` on vec and to handle the `Option` returned, or do a length check before and handle the case where invalid.

implementations/rust/kex/src/xx.rs
SanjoDeundiak: @mikelodder7

implementations/rust/kex/src/xx.rs
SanjoDeundiak: Agree, tnx

implementations/rust/kex/src/xx.rs
```diff
@@ -106,31 +111,43 @@ impl KeyExchange for SymmetricState {
...
+
         if self.key.is_some() {
             vault.secret_destroy(*self.key.as_ref().unwrap())?;
         }
-        self.key = Some(vault.secret_import(&key, attributes)?);
+        // FIXME
+        self.key = Some(hkdf_output[1]);
```
nilslice, line 150: again here, prefer `.get(1)` here and handle the Option.. could just assign `self.key = hdkf_output.get(1);` directly, but unsure if `self.key` is checked later or if it would be better to catch an unexpected state here.

implementations/rust/kex/src/xx.rs
SanjoDeundiak: Yes, will add the check before final review

implementations/rust/kex/src/xx.rs
```diff
@@ -188,35 +205,22 @@ impl KeyExchange for SymmetricState {
     }
 
     /// Split step in Noise protocol
-    fn split(&mut self) -> Result<Vec<u8>, VaultFailError> {
+    fn split(&mut self) -> Result<Vec<SecretKeyContext>, VaultFailError> {
         let mut vault = self.vault.lock().unwrap();
```
nilslice, line 209: best to handle this vs. `unwrap()` in case of poisoning.

SanjoDeundiak:
@nilslice please review and check daemon changes

implementations/rust/kex/src/xx.rs
nilslice: @mikelodder7  -- can you double-check that this is an acceptable `unwrap` here?

implementations/rust/kex/src/xx.rs
```diff
@@ -106,31 +111,57 @@ impl KeyExchange for SymmetricState {
...
-            xtype: SecretKeyType::Aes256,
-            persistence: SecretPersistenceType::Ephemeral,
-            purpose: SecretPurposeType::KeyAgreement,
-        };
-        let key = SecretKey::Aes256(*array_ref![hash, SHA256_SIZE, AES256_KEYSIZE]);
+    fn mix_key(
+        &mut self,
+        hkdf_output: (SecretKeyContext, SecretKeyContext),
+    ) -> Result<(), VaultFailError> {
         let mut vault = self.vault.lock().unwrap();
```
nilslice, line 153: also here -- mainly concerned about catching a https://doc.rust-lang.org/std/sync/struct.PoisonError.html, though we may be able to add a `?` if the From impl exists.

SanjoDeundiak: @nilslice imho unwrapping Mutex is ok, it'll only fail if another thread panicked.

implementations/rust/kex/src/xx.rs
```diff
@@ -550,7 +577,17 @@ mod tests {
         let res = state.prologue();
         assert!(res.is_ok());
```
nilslice, line 578: @mikelodder7 @SanjoDeundiak, there seem to be a lot of panic-inducing calls throughout this code `assert` `assert_eq` `unwrap` etc.. 

We need to either prove their existence is acceptable, or handle the cases in which they fail. I'm not confident that we're ok with these here, but would be happy to know otherwise 😄  ultimately, we should never panic and take down a caller's process unless there is no other logical option. We let them make that decision after they handle our returned error.

implementations/rust/vault/src/ffi/mod.rs
```diff
@@ -340,29 +335,28 @@ pub extern "C" fn ockam_vault_hkdf_sha256(
...
+            let hkdf_output = v
+                .vault
+                .hkdf_sha256(salt_ctx, Some(ikm_ctx), output_attributes)?
+                .iter()
+                .map(|x| x.into_ffi_value())
+                .collect();
+
             // FIXME: Double conversion is happening here
-            let res = OckamSecretList(outputs);
+            let res = OckamSecretList(hkdf_output);
```
nilslice, line 359:
```suggestion
            Ok(OckamSecretList(hkdf_output))
```

implementations/rust/vault/src/ffi/mod.rs
```diff
@@ -340,29 +335,28 @@ pub extern "C" fn ockam_vault_hkdf_sha256(
...
             Ok(res)
```
nilslice, line 360:
```suggestion
            
```

implementations/rust/kex/src/xx.rs
SanjoDeundiak: @nilslice is that so important for test code?

nilslice: @SanjoDeundiak -- ah, no tests are definitely a good place to use those kinds of calls. I apologize if any of my comments in the review are associated with tests. Github doesn't do a great job showing the context of the code point... 

All tests should use those kinds of functions, and we should be very selective when we use them elsewhere.

implementations/rust/kex/src/xx.rs
nilslice: ok, let's get agreement from @mikelodder7. if it is not recoverable, then I agree. It seems that there is some material out there that suggests unwrapping it is fine: https://users.rust-lang.org/t/any-examples-of-recovering-from-a-poisoned-lock/29435/3

maybe we should go with `self.vault.lock().expect("lock failed: mutex was poisoned")` so that the user of this code is made aware, in case the other panicked thread doesn't leave much helpful detail behind.

implementations/rust/vault/src/ffi/mod.rs
SanjoDeundiak: +

SanjoDeundiak:
@mikelodder7 can we merge this and create separate discussion about unlocking Mutex? I think it has nothing to do with this pr, it just happened that we noticed that fact here :)

nilslice (approved):
+1 to merging and discussing unwrap on lock at a later date. we gain no extra capability by expect vs unwrap, but it could add clarity to the end-user.

implementations/rust/kex/src/xx.rs
mikelodder7: The unwrap is okay on the mutex.

implementations/rust/kex/src/xx.rs
mikelodder7: I personally don't like using `expect`. In my experience, it's okay to unwrap on a Mutex.

implementations/rust/kex/src/xx.rs
```diff
@@ -106,31 +111,57 @@ impl KeyExchange for SymmetricState {
...
+        let ck = self
+            .state
+            .ck
+            .ok_or(VaultFailError::from(VaultFailErrorKind::Ecdh))?;
+
         let mut vault = self.vault.lock().unwrap();
-        vault.ec_diffie_hellman_hkdf_sha256(
+
+        let attributes_ck = SecretKeyAttributes {
+            xtype: SecretKeyType::Buffer(32),
```
mikelodder7, line 123:
```suggestion
            xtype: SecretKeyType::Aes256,
```

implementations/rust/vault/src/ffi/mod.rs
```diff
@@ -340,30 +335,28 @@ pub extern "C" fn ockam_vault_hkdf_sha256(
...
-            for derived in hkdf_bytes.chunks(SIZES) {
-                let key = SecretKey::Buffer(derived.to_vec());
-                let h = v.vault.secret_import(&key, attributes)?;
-                let secret = h.into_ffi_value();
-                outputs.push(secret);
+            let mut output_attributes = Vec::<SecretKeyAttributes>::new();
+
+            for _ in 0..derived_outputs_count {
+                output_attributes.push(SecretKeyAttributes {
+                    xtype: SecretKeyType::Buffer(SIZES),
```
mikelodder7, line 345:
```suggestion
                    xtype: SecretKeyType::Aes256,
```

mikelodder7 (commented):
Just a few suggested changes but otherwise I think it looks good.

implementations/rust/kex/src/xx.rs
SanjoDeundiak: @mikelodder7 for atecc608a I ended up creating new secret key type called "ChainKey", probably I'll do the same thing for rust later. Before that I don't understand why would chain key be of AES256 type

PR #488, opened 2020-10-09
feat(rust): add p256aes128 kex

SanjoDeundiak:


SanjoDeundiak:
1. Add P256 Aes128 key agreement
2. Improve key agreement structures to avoid unwrapping (even implicit)

PR #509, opened 2020-10-13
feat(rust): Implement X3dh key agreement

mikelodder7:
Implement the X3dh key agreement along with XEdDSA.

### DIFF
--- a/implementations/rust/Cargo.toml
+++ b/implementations/rust/Cargo.toml
@@ -10,6 +10,7 @@ members = [
     "router",
     "transport",
     "node",
+    "xeddsa",
     "c/generate_bindings",
     "c/bindings",
     "c/rust_memory",
@@ -25,6 +26,8 @@ default-members = [
     "router",
     "transport",
     "node",
+    "xeddsa",
+    "c/generate_bindings",
     "c/bindings",
     "c/rust_memory",
 ]
--- a/implementations/rust/vault/src/types.rs
+++ b/implementations/rust/vault/src/types.rs
@@ -380,6 +380,46 @@ impl PublicKey {
         use PublicKey::*;
         matches!(self, P256(..))
     }
+
+    /// Return the byte type for the enum value
+    pub fn get_type(&self) -> u8 {
+        match *self {
+            PublicKey::Curve25519(_) => 1,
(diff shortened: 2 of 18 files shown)


implementations/rust/Cargo.toml
```diff
@@ -25,6 +26,8 @@ default-members = [
     "router",
     "transport",
     "node",
+    "xeddsa",
+    "c/generate_bindings",
```
SanjoDeundiak, line 30: This was out of default_members list for a purpose

mikelodder7: Except I didn't add c/generate_bindings, that was from a merge.

implementations/rust/vault/src/types.rs
```diff
@@ -380,6 +380,46 @@ impl PublicKey {
         use PublicKey::*;
         matches!(self, P256(..))
     }
+
+    /// Return the byte type for the enum value
```
nilslice, line 384:
```suggestion
    /// Return the byte type for the enum value
    // TODO: consider using discriminant values when supported in Rust: https://github.com/rust-lang/rust/issues/60553
```

nilslice: it would be preferable if/when supported to be able to use the enum as a `u8` instead of doing this match, which currently you must do. just to remind us in the future...

implementations/rust/xeddsa/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "xeddsa"
+version = "0.1.0"
+authors = ["Michael Lodder <mike@ockam.io>"]
```
nilslice, line 4: I've seen us standardize on "Ockam Developers" as the only author entry -- not opposed to start tracking who has worked on these crates and keep the authors entry updated.

nilslice (changes requested):
Just a couple things I'd like your input on in the comments.

implementations/rust/xeddsa/Cargo.toml
mikelodder7: good catch. I'll make the change.

nilslice (approved):

PR #510, opened 2020-10-13
feat(rust): enable access to channel remote public key via channel manager

nilslice:
This enables a channel manager / channel to export the remote end's public key. However, I'm still not sure where in the channel manager (or elsewhere) code would be a good place to return the channel address created so that a later call can find the public key. Suggestions appreciated! (@robinbudd / @mikelodder7)


implementations/rust/channel/src/lib.rs
```diff
@@ -100,6 +100,14 @@ impl<I: KeyExchanger, R: KeyExchanger, E: NewKeyExchanger<I, R>> ChannelManager<
         })
     }
 
+    /// Get the remote public key for a channel
```
nilslice, line 103:
```suggestion
    /// Get the remote public key for a channel
    // TODO: consider using discriminant values when supported in Rust: https://github.com/rust-lang/rust/issues/60553
```

nilslice:
Closing in favor of different approach. Will have another small PR shortly.

PR #512, opened 2020-10-14
feat(rust): add remote public key to the message body for completed kex

nilslice:
This is a requirement for the initiator in `ockamd` to verify the public key of the responder. 

However, I think it is limited to that direction -- if so, would it also be necessary to add this public key data to the message in the `KeyAgreementM3` branch? lastly, then would it be in the case where there is a `pending_message`, or None, or both?


robinbudd (approved):

mikelodder7 (approved):
Nice pretty simple for getting the public key in the last message.
Does it need to go both ways?

nilslice:
> Nice pretty simple for getting the public key in the last message.
credit to @robinbudd for the simple & elegant idea.

> Does it need to go both ways?
I _think_ so, but I'm not 100% sure. The second p.message_body assignment in the PR is in the KeyAgreementM3 block, so I think it's already handled. Again, not 100% sure. but this unblocks the demo use case.

PR #518, opened 2020-10-19
feat(rust): add optional identity key when creating exchange

mikelodder7:
Update key exchange to allow an predefined identity key

### DIFF
--- a/implementations/rust/channel/src/lib.rs
+++ b/implementations/rust/channel/src/lib.rs
@@ -317,7 +317,8 @@ impl<I: KeyExchanger, R: KeyExchanger, E: NewKeyExchanger<I, R>> ChannelManager<
 
         m.onward_route.addresses.remove(0);
 
-        let mut channel = Channel::new(channel_id, Box::new(self.new_key_exchanger.initiator()));
+        let mut channel =
+            Channel::new(channel_id, Box::new(self.new_key_exchanger.initiator(None)));
         let ka_m1 = channel.agreement.process(&[])?;
 
         m.onward_route
@@ -342,7 +343,7 @@ impl<I: KeyExchanger, R: KeyExchanger, E: NewKeyExchanger<I, R>> ChannelManager<
         let mut rng = thread_rng();
         let channel_id = rng.gen::<u32>();
         let channel_address = Address::ChannelAddress(channel_id.to_le_bytes().to_vec());
-        let channel = Channel::new(channel_id, Box::new(self.new_key_exchanger.responder()));
+        let channel = Channel::new(channel_id, Box::new(self.new_key_exchanger.responder(None)));
         self.channels.insert(channel_address.as_string(), channel);
         Ok(channel_address.as_string())
     }
--- a/implementations/rust/kex/src/x3dh.rs
+++ b/implementations/rust/kex/src/x3dh.rs
@@ -104,7 +104,7 @@ const ENROLLMENT_MSG_SIZE: usize = 32 + 32 + 32 + 64 + 16;
 /// The responder of X3DH creates a prekey bundle that can be used to establish a shared
 /// secret key with another party that can use
 pub struct X3dhResponder {
-    identity_key: SecretKeyContext,
+    identity_key: Option<SecretKeyContext>,
     signed_prekey: SecretKeyContext,
     one_time_prekey: SecretKeyContext,
     expected_enrollment_key: Option<PublicKey>,
@@ -114,9 +114,9 @@ pub struct X3dhResponder {
 }
 
 impl X3dhResponder {
-    fn new(v: Arc<Mutex<dyn DynVault + Send>>) -> Self {
+    fn new(v: Arc<Mutex<dyn DynVault + Send>>, identity_key: Option<SecretKeyContext>) -> Self {
         Self {
-            identity_key: SecretKeyContext::Memory(0),
+            identity_key,
             signed_prekey: SecretKeyContext::Memory(0),
             one_time_prekey: SecretKeyContext::Memory(0),
             expected_enrollment_key: None,
@@ -138,7 +138,9 @@ impl X3dhResponder {
             persistence: SecretPersistenceType::Ephemeral,
             xtype: SecretKeyType::Curve25519,
         };
-        self.identity_key = vault.secret_generate(p_atts)?;
+        if self.identity_key.is_none() {
+            self.identity_key = Some(vault.secret_generate(p_atts)?);
+        }
         self.signed_prekey = vault.secret_generate(p_atts)?;
         self.one_time_prekey = vault.secret_generate(e_atts)?;
         self.expected_enrollment_key = None;
@@ -184,16 +186,18 @@ pub struct X3dhInitiator {
     state: InitiatorState,
     vault: Arc<Mutex<dyn DynVault + Send>>,
     completed_key_exchange: Option<CompletedKeyExchange>,
+    identity_key: Option<SecretKeyContext>,
 }
 
 impl X3dhInitiator {
-    fn new(v: Arc<Mutex<dyn DynVault + Send>>) -> Self {
+    fn new(v: Arc<Mutex<dyn DynVault + Send>>, identity_key: Option<SecretKeyContext>) -> Self {
         Self {
             ephemeral_identity_key: SecretKeyContext::Memory(0),
             prekey_bundle: None,
             state: InitiatorState::GenerateEphemeralIdentityKey,
             vault: v,
             completed_key_exchange: None,
+            identity_key,
         }
     }
 
@@ -215,11 +219,12 @@ impl std::fmt::Debug for X3dhInitiator {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         write!(
             f,
-            r#"X3dhInitiator {{ ephemeral_identity_key: {:?}, prekey_bundle: {:?}, state: {:?}, vault, completed_key_exchange: {:?} }}"#,
+            r#"X3dhInitiator {{ ephemeral_identity_key: {:?}, prekey_bundle: {:?}, state: {:?}, vault, completed_key_exchange: {:?}, identity_key: {:?} }}"#,
             self.ephemeral_identity_key,
             self.prekey_bundle,
             self.state,
-            self.completed_key_exchange
+            self.completed_key_exchange,
+            self.identity_key,
         )
     }
 }
@@ -230,9 +235,14 @@ impl KeyExchanger for X3dhResponder {
             ResponderState::GenerateBundle => {
                 self.prologue()?;
                 let mut vault = self.vault.lock().unwrap();
+                let identity_secret_key =
+                    self.identity_key
+                        .ok_or(KeyExchangeFailErrorKind::GeneralError {
+                            msg: "Invalid identity key".to_string(),
+                        })?;
                 let signed_prekey = vault.secret_public_key_get(self.signed_prekey)?;
-                let signature = vault.sign(self.identity_key, signed_prekey.as_ref())?;
-                let identity_key = vault.secret_public_key_get(self.identity_key)?;
+                let signature = vault.sign(identity_secret_key, signed_prekey.as_ref())?;
+                let identity_key = vault.secret_public_key_get(identity_secret_key)?;
                 let one_time_prekey = vault.secret_public_key_get(self.one_time_prekey)?;
                 let bundle = PreKeyBundle {
                     identity_key,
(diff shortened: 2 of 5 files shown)


implementations/rust/channel/src/lib.rs
```diff
@@ -317,7 +317,8 @@ impl<I: KeyExchanger, R: KeyExchanger, E: NewKeyExchanger<I, R>> ChannelManager<
 
         m.onward_route.addresses.remove(0);
 
-        let mut channel = Channel::new(channel_id, Box::new(self.new_key_exchanger.initiator()));
+        let mut channel =
+            Channel::new(channel_id, Box::new(self.new_key_exchanger.initiator(None)));
```
nilslice, line 321: Within `fn initiate_key_exchange`, is the expectation that both parties have no intention of using pre-existing keys? 

e.g. this function is _not_ called in the case that I do want to supply keys?

implementations/rust/kex/src/x3dh.rs
```diff
@@ -215,11 +219,12 @@ impl std::fmt::Debug for X3dhInitiator {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         write!(
             f,
-            r#"X3dhInitiator {{ ephemeral_identity_key: {:?}, prekey_bundle: {:?}, state: {:?}, vault, completed_key_exchange: {:?} }}"#,
+            r#"X3dhInitiator {{ ephemeral_identity_key: {:?}, prekey_bundle: {:?}, state: {:?}, vault, completed_key_exchange: {:?}, identity_key: {:?} }}"#,
```
nilslice, line 222: It would be preferable to implement `Debug` using the `debug_struct` etc method, example here: https://doc.rust-lang.org/std/fmt/trait.Debug.html#examples-1

also for the Debug implementation of `X3dhResponder`

nilslice: The benefit is that using `{:?}` is printed on one line, and `{:#?}` is printed on many lines. this is more important if these formatted strings ever get logged, it becomes really hairy to untangle single line log messages from multi-line.

implementations/rust/kex/src/x3dh.rs
```diff
@@ -274,7 +279,7 @@ impl KeyExchanger for X3dhResponder {
                 let ek = PublicKey::Curve25519(*array_ref![data, 0, 32]);
 
                 let dh1 = vault.ec_diffie_hellman(self.signed_prekey, *eik)?;
-                let dh2 = vault.ec_diffie_hellman(self.identity_key, ek)?;
+                let dh2 = vault.ec_diffie_hellman(self.identity_key.unwrap(), ek)?;
```
nilslice, line 282: does this need to be handled rather than unwrapping?

implementations/rust/kex/src/x3dh.rs
```diff
@@ -329,7 +334,7 @@ impl KeyExchanger for X3dhResponder {
                     h: state_hash,
                     encrypt_key,
                     decrypt_key,
-                    local_static_secret: self.identity_key,
+                    local_static_secret: self.identity_key.unwrap(),
```
nilslice, line 337: does this need to be handled rather than unwrapping?

implementations/rust/kex/src/xx.rs
```diff
@@ -90,10 +90,25 @@ impl SymmetricState {
...
+    pub fn new(
+        cipher_suite: CipherSuite,
+        vault: Arc<Mutex<dyn DynVault + Send>>,
+        identity_key: Option<SecretKeyContext>,
+    ) -> Self {
+        let static_key_pair = match identity_key {
+            None => None,
+            Some(ik) => {
+                let mut v = vault.lock().unwrap();
+                let pk = v.secret_public_key_get(ik).unwrap();
```
nilslice, line 102: does this need to be handled rather than unwrapping?

implementations/rust/kex/src/x3dh.rs
mikelodder7: I get that. The problem is you can't use `debug_struct` on Arc<Mutex<dyn DynVault>>

implementations/rust/kex/src/x3dh.rs
mikelodder7: At this point in the code it's always `Some`

implementations/rust/kex/src/x3dh.rs
mikelodder7: Same as previous

implementations/rust/kex/src/xx.rs
mikelodder7: This is debatable but I think it's okay to call unwrap here.

implementations/rust/channel/src/lib.rs
mikelodder7: I'm assuming @robinbudd will take care of that part. I just put this fix here so the tests pass.

nilslice: Ah, right -- sorry, that's correct as far as I am aware. thanks for calling it out.

implementations/rust/kex/src/x3dh.rs
nilslice: I think you can pretty much use it regardless... but either way,`X3dhInitiator` is the name of the struct you'd pass to `debug_struct` though, not the `Arc`. 

The main problem I have with this is the multi-line printing... if you don't want to do the work to implement with `debug_struct`, I think at minimum you should remove the multi-line value and put it all on a single line.

implementations/rust/kex/src/x3dh.rs
```diff
@@ -231,8 +236,8 @@ impl KeyExchanger for X3dhResponder {
                 self.prologue()?;
                 let mut vault = self.vault.lock().unwrap();
                 let signed_prekey = vault.secret_public_key_get(self.signed_prekey)?;
-                let signature = vault.sign(self.identity_key, signed_prekey.as_ref())?;
-                let identity_key = vault.secret_public_key_get(self.identity_key)?;
+                let signature = vault.sign(self.identity_key.unwrap(), signed_prekey.as_ref())?;
```
nilslice, line 239: does this need to be handled rather than unwrapping?

implementations/rust/kex/src/x3dh.rs
```diff
@@ -231,8 +236,8 @@ impl KeyExchanger for X3dhResponder {
...
+                let identity_key = vault.secret_public_key_get(self.identity_key.unwrap())?;
```
nilslice, line 240: does this need to be handled rather than unwrapping?

implementations/rust/kex/src/x3dh.rs
nilslice: At this point, _maybe_ given the current logic -- but what if that changes? Could you point me to what guarantees this as `Some`? Sorry if I'm missing it. If it's just match logic, I don't think that's enough.

implementations/rust/kex/src/x3dh.rs
nilslice: If this is true because it's "been checked" prior, then shouldn't we use a new variable at the point you prove it's `Some` and use that elsewhere?

implementations/rust/kex/src/xx.rs
nilslice: I especially worry about eliding the Option check here, because this code trusts the underlying Vault implementation. _You_ think it's ok because you know both the XX code and the Vault code being used here now. But in the future, it could be a different Vault implementation used, correct? In that case it could behave in a way that doesn't satisfy your current thought.

implementations/rust/kex/src/x3dh.rs
mikelodder7: If I put in the debug_struct it doesn't compile.

implementations/rust/kex/src/x3dh.rs
mikelodder7: See line 141

implementations/rust/kex/src/x3dh.rs
mikelodder7: No because it has to be passed across function calls which are not directly linked.

implementations/rust/kex/src/x3dh.rs
mikelodder7: See above.

implementations/rust/kex/src/x3dh.rs
mikelodder7: See above

implementations/rust/kex/src/xx.rs
mikelodder7: If we don't unwrap what is the expected behavior? This requires an interface change if not unwrapped

implementations/rust/kex/src/x3dh.rs
nilslice: That's because there are non-Debug impl types in the struct fields. 

Just make it one single line then, to keep it consistent?

implementations/rust/kex/src/x3dh.rs
nilslice: ok, so you should definitely use the value if you can, NOT continuing to call `.unwrap()` everywhere. assigning the result of the `vault.secret_generate(p_atts)?` call to a new variable will ensure that at no point in the future does that `self.identity_key` happen to turn into a `None`... code changes & this is only valid in it's current state.

nilslice (changes requested):
at minimum:

1. let's reduce the noise of any Debug impl so that the print is only using a single line. 
2. fix the `unwrap` issue in `x3dh.rs` so that it's less prone to uncaught changes in the future.

```rust
if self.identity_key.is_none() {
     self.identity_key = Some(vault.secret_generate(p_atts)?);
// ^ assign Some(vault.secret_generate(p_atts)?) to a new variable and use it instead of unwrapping the same call to self.identity_key ?
}
```

implementations/rust/kex/src/xx.rs
nilslice: If the choices are:
1. unwrap and possibly panic
2. change the interface

my opinion is 2.

@mrinalwadhwa may have some thoughts?

SanjoDeundiak:
Should we also add identity public key of the other party?

robinbudd (approved):

implementations/rust/kex/src/xx.rs
mikelodder7: I have adjusted this to store a seeded value then use it later where it uses the ? operator instead of unwrapping.

mikelodder7:
> at minimum:
> 
>     1. let's reduce the noise of any Debug impl so that the print is only using a single line.
> 
>     2. fix the `unwrap` issue in `x3dh.rs` so that it's less prone to uncaught changes in the future.
> 
> 
> ```rust
> if self.identity_key.is_none() {
>      self.identity_key = Some(vault.secret_generate(p_atts)?);
> // ^ assign Some(vault.secret_generate(p_atts)?) to a new variable and use it instead of unwrapping the same call to self.identity_key ?
> }
> ```

Already fixed in my latest push

implementations/rust/kex/src/x3dh.rs
mikelodder7: The compiler complains that Debug is not implemented for trait objects. While it is good practice to use derive(Debug) when you can, its not always feasible.

implementations/rust/kex/src/xx.rs
nilslice: @mikelodder7 where? use Github's code linking feature's to post a comment with the link to the line where this is done please :)

mikelodder7:
> @mikelodder7 - here's a new one. why is this safe to unwrap here? i don't see any prior check within this function.
> 
> you could add a match guard to this arm, `ResponderState::GenerateBundle if self.identity_key.is_some()` and another match arm for the `None` case to return an error? That would also handle the concern of the next line. 


Lets start at the beginning. 

1. `new` on line 117 receives an Optional identity_key and just sets it into Self
2. The next public function they can even call that touches identity_key is `process`. There we see that the first state is GenerateBundle.
3. The first operation in GenerateBundle is a call to prologue
4. On line 141, we see that if identity_key is none, it is immediately set to Some. This identity_key is guaranteed to have a value.
5. Control returns to 237. On line 239 identity_key is unwrapped. This is okay because it was just set.
6. The next call that can touch identity_key is `process` again which is `SetEnrollmentKey`
7. Starting on line 251, identity_key hasn't been touched since then, therefore it still has a proper value.
8. The same happens for the next call to `process` for VerifyEnrollment.

implementations/rust/kex/src/xx.rs
mikelodder7: See [here](https://github.com/ockam-network/ockam/pull/518/files#diff-2b6c53777e3d5dc1c187c0864bf7c32e9811f1839d154e22f736bafa7a36f67fR134)

nilslice:
@mikelodder7 

> Lets start at the beginning.
`new` on line 117 receives an Optional identity_key and just sets it into Self
The next public function they can even call that touches identity_key is process. There we see that the first state is GenerateBundle.
The first operation in GenerateBundle is a call to prologue
On line 141, we see that if identity_key is none, it is immediately set to Some. This identity_key is guaranteed to have a value.
Control returns to 237. On line 239 identity_key is unwrapped. This is okay because it was just set.
The next call that can touch identity_key is process again which is SetEnrollmentKey
Starting on line 251, identity_key hasn't been touched since then, therefore it still has a proper value.
The same happens for the next call to process for VerifyEnrollment.

I understand, but the point is that this code _expects_ all of that to be done (which at current it _is_), but by ignoring the possibility of a `None` from the `unwrap` calls, this code will break in the future if its expectations are ever unmet. This invariant is exactly why we want to do better to check these cases. 

In projects like Ockam, where contributions can come from anyone at any time - it's even more likely that something slips by and could break the code. Does that make sense?

mikelodder7:
Changed them to use `ok_or` which means if they are None, they are converted to an error that is caught and returned.

nilslice (approved):
thank you @mikelodder7!

Issue #530, opened 2020-10-22
vault ffi build failed


mrinalwadhwa:
```
» rustc --version
rustc 1.47.0 (18bf6b4f0 2020-10-07)
```

```
» cargo build --features=ffi
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/channel/Cargo.toml
workspace: /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/Cargo.toml
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/message/Cargo.toml
workspace: /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/Cargo.toml
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/kex/Cargo.toml
workspace: /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/Cargo.toml
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/vault/Cargo.toml
workspace: /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/Cargo.toml
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/router/Cargo.toml
workspace: /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/Cargo.toml
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/transport/Cargo.toml
workspace: /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/Cargo.toml
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
package:   /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/worker/Cargo.toml
workspace: /Users/mrinal/Workspace/ockam-network/ockam/implementations/rust/Cargo.toml
   Compiling log v0.4.11
   Compiling lazy_static v1.4.0
   Compiling ffi-support v0.4.2
[...]

PR #533, opened 2020-10-22
feat(rust): fix vault ffi hkdf

mikelodder7:
Address issue #530

### DIFF
--- a/implementations/rust/vault/src/ffi/mod.rs
+++ b/implementations/rust/vault/src/ffi/mod.rs
@@ -325,7 +325,7 @@ pub extern "C" fn ockam_vault_hkdf_sha256(
     salt: u64,
     input_key_material: u64,
     derived_outputs_count: u8,
-    derived_outputs: &mut [u64],
+    derived_outputs: *mut u64,
 ) -> VaultError {
     let derived_outputs_count = derived_outputs_count as usize;
     let mut err = ExternError::success();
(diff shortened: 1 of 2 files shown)


implementations/rust/vault/src/ffi/mod.rs
```diff
@@ -348,9 +348,10 @@ pub extern "C" fn ockam_vault_hkdf_sha256(
                 });
             }
 
+            // Hard code empty info string for now
```
nilslice, line 351: @mikelodder7 what's the eventual state this should be in? why an empty info string? 

adding more detail to this comment here would be helpful.

nilslice (changes requested):
Looks good -- just want to see a bit more detail in the comment e.g. what the long-term solution should be vs. the hard-coded empty string.

implementations/rust/vault/src/ffi/mod.rs
mikelodder7: Hardcoded to be empty for now because any changes to the C layer requires an API change. This change was necessary to implement Enrollment since the info string is not left blank for that protocol, but is blank for the XX pattern. If we agree to change the API, then this wouldn't be hardcoded but received from a parameter in the C API. Elixir and other consumers would be expected to pass the appropriate flag.

The other option is to not expose the vault directly since it may confuse users about what to pass here and I don't like the idea of yelling at consumers through comments. Instead the vault could be encapsulated in channels and key exchanges. Either way, I don't want to change the API until this decision is finalized.

nilslice: @mikelodder7 - ok, I meant that this detail should be added to the comment _in the source code_ :) can you push an update to include this?

nilslice: Sounds like a good `// TODO:`

nilslice: > yelling at consumers through comments. 

no one is yelling -- but this is too obvious of a comment to be helpful. simplifying the explanation above in a `TODO` is a reasonable expectation so that this isn't forgotten about.

mikelodder7: Add this to the comments as `TODO`

nilslice (approved):
ah, looks l

PR #538, opened 2020-10-22
build: add dockerfiles for ockam-dev as a builder image and ockamd

nilslice:
From the `ockam-network/ockam` project root:

`docker build -t ockam-dev -f tools/docker/ockam-dev/Dockerfile .` 
- _this will take a while, and currently produces a >3GB image_

`docker build -t ockamd:0.1.0 -f tools/docker/rust/Dockerfile.ockamd .`
- _very fast since dependencies are pre-compiled in ockam-dev_

Eventually I would like to push these to Docker hub, but they will be the basis for the docker-compose setup for the telegraf/influx end-to-end demo.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,20 @@
+# from .gitignore 
+.env
+.gradle
+.vagrant
+.envrc
+.builder
+.apple-codesign
+CMakeCache.txt
+CMakeFiles
+Makefile
+cmake_install.cmake
+*.a
+*.DS_store
+*.gch
+.idea
+.tool-versions
+
+.git
+implementations/rust/target
+implementations/rust/ockamd_vault
--- a/tools/docker/ockam-dev/Dockerfile
+++ b/tools/docker/ockam-dev/Dockerfile
@@ -1,0 +1,48 @@
+FROM buildpack-deps:buster-scm
+
+# install base Ockam build system dependencies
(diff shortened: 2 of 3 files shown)


mrinalwadhwa:
for the above example commands to work, this line
https://github.com/ockam-network/ockam/blob/e23965d09d7f9c96f51f947745c4d6c705de2559/tools/docker/rust/Dockerfile.ockamd#L2

needed to be `ockamd-dev` in place of `ockam-dev`

mrinalwadhwa:
```
» docker run --rm -it ockamd:0.1.0 /bin/bash
/bin/bash: 1: [ockamd]: not found
```

mrinalwadhwa (approved):

mrinalwadhwa:
looks good 👍 

There is a little more work needed for the C and Elixir builds to work in the container ..
C seems to fail.
Elixir succeeds but atleast the locale warning important to fix.

nilslice:
@mrinalwadhwa  -- good catch, I mistyped the example and have since corrected it. 

> There is a little more work needed for the C and Elixir builds to work in the container ..
C seems to fail.
Elixir succeeds but atleast the locale warning important to fix.

Yes, this just gets me building the rust code for the demos in docker-compose. We will have to sort through the rest of the dependencies for other implementations 😄

mrinalwadhwa (approved):

PR #545, opened 2020-10-23
feat(rust): integrate new channel api with ockamd, encapsulate messages in outer message at channel layer

robinbudd:


nilslice:
@robinbudd - we may need to back this out. looks like there are a number of commits in here that shouldn't be. it would be safer to PR my earlier branch, merge it & for you to rebase off develop. Although, it might mean manually adding back the changes you've made.

nilslice (changes requested):
see my comment above -- might need to close this and start from a cleaner base branch.

robinbudd:
Superseded by #547

PR #547, opened 2020-10-24
feat(rust): integrate new channel api with ockamd, encapsulate messages in outer message at channel layer

robinbudd:
- channel now has 2 addresses, one for encrypt and one for decrypt.
  On the encrypt address, channel encryptes and encapsulates messages in
  an outer message. On the decrypt address, channel decrypts the payload
  and expects it to be a message.
- channel now requires the initiator to explicitly create a channel, instead
  of automatically initiating a channel when a message is sent to address 0
- ockamd now requires the responder to be running prior to the initiator
- ockamd requires a remote worker address and accepts a predefined static
  key, although it doesn't yet use the key


nilslice (dismissed):
nice! just a couple place we can clean up later in some manifests. e.g. not sure this needs to be in `node` bin
```toml
[lib]
crate-type = ["rlib"]
```

nilslice (approved):

PR #549, opened 2020-10-24
feat(rust): make ockamd responder display static public key on startup

robinbudd:


nilslice:
Still need to give #547 a closer review, but once that is merged can you change this commit message before you rebase from the updated `develop`?

`fix(rust): cargo fmt all files` -> `style(rust): cargo fmt all files`

I learned recently that we are reserving `fix` for cases where the code was incorrect or buggy, and `style` is more in line with that particular commit.

robinbudd:
> Still need to give #547 a closer review, but once that is merged can you change this commit message before you rebase from the updated `develop`?
> 
> `fix(rust): cargo fmt all files` -> `style(rust): cargo fmt all files`
> 
> I learned recently that we are reserving `fix` for cases where the code was incorrect or buggy, and `style` is more in line with that particular commit.

I would if I knew how to go back and edit that one commit message. I've noted that and will comply moving forward. Does it matter for that one commit?

mrinalwadhwa:
@robinbudd 
```
git commit --amend HASH_OF_COMMIT
```

PR #550, opened 2020-10-24
feat(rust): make ockamd panic if supplied public key doesn't match computed key

robinbudd:
- channel now sends responder public key in message-body of channel
  completion message
- ockamd-initiator will panic if a responder public key is supplied and
  it doesn't match the computed responder key


nilslice:
@robinbudd, I'm thinking that I will close #549 and merge this (which includes all of #549's code) instead of both. Saving a rebase, and renaming commit as per `style` vs. `fix` prefixes: `fix(rust): cargo fmt all files` -> `style(rust): cargo fmt all files`

nilslice (approved):

PR #539, opened 2020-10-23
native functions

mrinalwadhwa:
- build(rust): vault crate with ffi feature on by default
- feat(elixir): remove key exchange native functions
- style(elixir): fix unaliased nested modules and format
- build(elixir): compile native functions based on operating system
- build(elixir): ockam_vault_software with native functions
- build: make elixir gradle depend on rust gradle

PR #551, opened 2020-10-25
binary wire protocol

mrinalwadhwa:
- feat(elixir): define serializable protocol
- feat(elixir): define the error module
- feat(elixir): make address value serializable and default type zero
- feat(elixir): make massage payload serializable
- feat(elixir): improve defualt impl and error handling in wire module
- feat(elixir): add varint encode and decode
- feat(elixir): add binary implementation of wire behaviour


mrinalwadhwa:
Based on #539, should be merged after that.

nilslice (approved):

nilslice:
merged via #552

PR #558, opened 2020-10-26
feat(rust): encode nonce as variable-length u16

robinbudd:


nilslice:
@robinbudd - let's check if you have your "Format on save" equivalent set up in your editor. It will help avoid the CI feedback loop for these kinds of commits:  `style(rust): cargo fmt`

nilslice (approved):

robinbudd:
> @robinbudd - let's check if you have your "Format on save" equivalent set up in your editor. It will help avoid the CI feedback loop for these kinds of commits: `style(rust): cargo fmt`

Been meaning to do it, it's now done.

mrinalwadhwa (approved):

PR #577, opened 2020-10-27
feat(rust): move file system vault

mikelodder7:
From daemon folder to vault folder.

The only code changes made are necessary for the code to compile and pass tests. No other modifications were made.

PR #578, opened 2020-10-27
fix(rust): only save key to disk if type is persistent

nilslice:
Restricts writes to disk for keys that are of type persistent.


nilslice:
@mikelodder7, this includes #577, and the rebase, so i'm going to close #577 and merge this.

mikelodder7 (dismissed):

mrinalwadhwa (approved):

PR #579, opened 2020-10-27
feat(rust): make ockamd responder check for valid worker address

robinbudd:
if the address isn't valid, drop the message and print a warning


nilslice:
recheckcla

nilslice:
going to merge #581 before this PR and then do a rebase for `rb/worker_address` from the latest `develop`.

nilslice (approved):

PR #585, opened 2020-10-28
feat(elixir): add ability to call and use rust filesystem vault

mikelodder7:


implementations/elixir/ockam/ockam_vault_software/test/vault_software_test.exs
```diff
@@ -14,6 +14,13 @@ defmodule Ockam.Vault.Software.Tests do
     end
   end
```
nilslice, line 15: @mikelodder7 -- do you mind adding a copy of this test and replacing the line `{:ok, handle} = SoftwareVault.default_init()` with `{:ok, handle} = SoftwareVault.file_init("/tmp")`

to see that it also produces the same hash? just to verify that the two vaults have consistent behavior.

mikelodder7: sure

implementations/rust/vault/src/file/mod.rs
```diff
@@ -138,6 +158,14 @@ impl DynVault for FilesystemVault {
 
     /// Export a secret key from the vault
     fn secret_export(&mut self, context: SecretKeyContext) -> Result<SecretKey, VaultFailError> {
+        use std::io::Write;
```
nilslice, line 161: just a little cleanup here from debugging :)

nilslice (changes requested):
thx for fixing! let's remove the tmp file though and then will keep an eye out for the elixir lint fix

implementations/rust/vault/src/file/mod.rs
mikelodder7: good catch

nilslice:
@mikelodder7, these commit messages should also be updated to conform to the commit linter:

`feat(elixit): add ability to use filesystem vault` -> `feat(elixir): add ability to use filesystem vault` (misspell elixir)
`feat(rust): fix bug in filesystem vault` -> `fix(rust): fix bug in filesystem vault` (re-classify as fix)

nilslice (approved):
👏 👏 👏

PR #591, opened 2020-10-29
feat(rust): implement the influxdb addon configuration, write message over http

nilslice:
implement the full data path: 
`telegraf -> ockamd --(secure channel)--> ockamd -> influxdb`

introduces an "addon" CLI paramater, which is meant to simplify the usage of various integrations. to use the InfluxDB addon, you run the responder as such:

`ockamd --role=responder --local-socket=127.0.0.1:52440 --addon=influxdb,$DB_NAME,http://127.0.0.1:8086`

This will configure `ockamd` to send encrypted messages received by responder to InfluxDB over HTTP using the database name and URL provided.

### DIFF
--- a/implementations/rust/daemon/src/cli.rs
+++ b/implementations/rust/daemon/src/cli.rs
@@ -97,6 +97,12 @@ pub struct Args {
     )]
     service_address: Option<String>,
 
+    #[structopt(
+        long,
+        help = r#"Pre-defined configuration for an official Ockam Add-on, e.g. "influx,http://localhost:8086""#
+    )]
+    addon: Option<Addon>,
+
     // TODO: expose `control` and `control_port` once runtime configuration is needed.
     #[structopt(
         short,
@@ -130,6 +136,7 @@ impl Default for Args {
             service_address: None,
             identity_name: format!("1{}", FILENAME_KEY_SUFFIX),
             service_public_key: None,
+            addon: None,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


implementations/rust/daemon/src/config.rs
```diff
@@ -16,6 +16,11 @@ pub enum Input {
     Stdin,
 }
 
+#[derive(Debug, Clone)]
+pub enum AddonKind {
```
mikelodder7, line 20: Should this be an empty trait similar to vault?

implementations/rust/daemon/src/cli.rs
```diff
@@ -181,6 +188,36 @@ impl Args {
...
         self.identity_name.clone()
     }
+
+    pub fn addon(&self) -> Option<Addon> {
+        self.addon.clone()
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum Addon {
```
mikelodder7, line 198: Maybe some comments here about what an Addon is?

mikelodder7 (commented):
Cool beans. Just one comment to discuss. Otherwise looks good.

implementations/rust/daemon/src/config.rs
nilslice: We should consider that if we intended to ship `ockamd` as a lib, but it's only going to be downloaded as a pre-built binary (or build from source).

implementations/rust/daemon/src/cli.rs
nilslice: Good call. Lots of places need more comments in `ockamd` and elsewhere. But we'll address all of those after this mile is over.

mikelodder7 (approved):

mrinalwadhwa (approved):

PR #607, opened 2020-11-01
feat(rust): implement channel tunneling, update ockamd to use it

robinbudd:


implementations/rust/daemon/src/cli.rs
```diff
@@ -37,15 +37,21 @@ pub struct Args {
         default_value = "stdout",
         help = r#"Route to channel responder, e.g. udp://host:port[,udp://host:port] (note comma-separation) or "stdout""#
     )]
-    route: OutputKind,
+    sink_route: OutputKind,
```
nilslice, line 40: this breaks the CLI -- i don't think we need to disambiguate the `sink_route` to end-users, as (I believe) there's only 1 kind of `route`.

implementations/rust/daemon/src/cli.rs
```diff
@@ -241,19 +257,23 @@ impl FromStr for VaultKind {
 pub enum ChannelRole {
     /// The Initiator role expects a channel responder address and a public key to use in order to
     /// communicate with the Responder end of the channel.
-    Initiator,
+    Source,
     /// The Responder role will create a channel responder, and will instruct the program to print
```
nilslice, line 261: If we're going to commit to using these names in code (which I think @mrinalwadhwa should approve) then can you also update the comments to reflect the change?

implementations/rust/daemon/src/cli.rs
```diff
@@ -241,19 +257,23 @@ impl FromStr for VaultKind {
...
 impl FromStr for ChannelRole {
     type Err = String;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
-            "initiator" | "init" => Ok(ChannelRole::Initiator),
-            "responder" | "resp" => Ok(ChannelRole::Responder),
+            "initiator" | "init" => Ok(ChannelRole::Source),
+            "responder" | "resp" => Ok(ChannelRole::Sink),
+            "router" => Ok(ChannelRole::Router),
```
nilslice, line 276: these should also reflect the naming change (again, if we're going to commit to it & @mrinalwadhwa approves).. this particular function takes the string literal from the CLI input for the `--role` arg, and translates it into the proper enum variant. So we want these to be consistent as well.

implementations/rust/daemon/src/config.rs
```diff
@@ -110,8 +123,9 @@ impl From<cli::Args> for Config {
         }
 
         cfg.role = match args.role() {
-            cli::ChannelRole::Initiator => Role::Initiator,
-            cli::ChannelRole::Responder => Role::Responder,
+            cli::ChannelRole::Source => Role::Initiator,
+            cli::ChannelRole::Sink => Role::Responder,
+            cli::ChannelRole::Router => Role::Router,
```
nilslice, line 128: Same consistency suggestion as before -- I don't think we should have different names between internal / external concerns (e.g. it's all external)

implementations/rust/daemon/src/initiator.rs
```diff
@@ -24,9 +26,31 @@ pub fn run(config: Config) {
...
+    }
+    if let Some(channel_to_sink) = config.channel_to_sink() {
+        onward_route
+            .addresses
+            .push(RouterAddress::channel_router_address_from_str(&channel_to_sink).unwrap());
+    }
+    onward_route
+        .addresses
+        .push(RouterAddress::channel_router_address_from_str(CHANNEL_ZERO).unwrap());
+    println!("Initiating channel with route");
```
nilslice, line 48: let's remove this println

implementations/rust/daemon/src/node.rs
```diff
@@ -128,12 +128,14 @@ impl<'a> Node<'a> {
             )))
             .expect("failed to register worker with router");
 
+        println!("worker registered");
```
nilslice, line 131: lets remove this println

implementations/rust/daemon/src/node.rs
```diff
@@ -128,12 +128,14 @@ impl<'a> Node<'a> {
...
         self.worker = Some(worker);
     }
 
     pub fn run(mut self) {
         match self.worker {
-            Some(worker) => {
+            Some(mut worker) => {
+                println!("will poll worker");
```
nilslice, line 138: lets remove this println

implementations/rust/daemon/src/responder.rs
```diff
@@ -4,14 +4,35 @@ use crate::config::{AddonKind, Config};
...
     let (mut node, router_tx) = Node::new(&config);
 
     let worker_addr = RouterAddress::worker_router_address_from_str("01242020").unwrap();
+    // kick off secure channel to router, if we have a router address
+    match config.router_socket() {
+        Some(socket) => {
+            let route = Route {
+                addresses: vec![
+                    RouterAddress::from_address(Address::UdpAddress(socket)).unwrap(),
+                    RouterAddress::channel_router_address_from_str("00000000").unwrap(),
```
nilslice, line 22: can this be replaced with the CHANNEL_ZERO constant?

implementations/rust/daemon/src/worker.rs
```diff
@@ -52,7 +57,24 @@ impl Worker {
...
+        //     if rpk == encode(&m.message_body) {
+        //         println!("keys agree");
+        //         return Ok(());
+        //     } else {
+        //         println!("keys conflict");
+        //         return Err("remote public key doesn't match expected, possible spoofing".into());
+        //     }
+        // }
+        // Ok(())
+    }
```
nilslice, line 75: does this introduce a regression? are we not validating the public key elsewhere if this is commented out?

implementations/rust/daemon/src/cli.rs
mrinalwadhwa: Please help me understand the thought behind this?

nilslice (changes requested):
Need some updates before it can be tested - namely the identifier name change for Source/Sink and making that consistent, and the approval from @mrinalwadhwa that we want to do so. 

Also, the CLI arg name change should probably be reverted so that the existing docker demo and the slide in mrinal's Influx deck is still valid.

implementations/rust/daemon/src/cli.rs
robinbudd: @mrinalwadhwa The "responder" role is no longer really appropriate for the sink end of the pipe (i.e. the end that pumps data into Influx). The sink acts as initiator when it establishes the channel with the router, and as responder when it gets M1 from the source end of the pipe. "Source" and "sink" are less ambiguous IMO and don't muddy the initiator/responder relationship.

implementations/rust/daemon/src/cli.rs
robinbudd: ok

implementations/rust/daemon/src/cli.rs
robinbudd: done...

implementations/rust/daemon/src/cli.rs
robinbudd: done

implementations/rust/daemon/src/initiator.rs
robinbudd: done

implementations/rust/daemon/src/node.rs
robinbudd: done

implementations/rust/daemon/src/worker.rs
robinbudd: This is the sink, which publishes its public key. It's the source side that validates the public key.

nilslice:
All of this is rolled into #620, so I'm closing this PR in favor of that.

PR #608, opened 2020-11-01
ockam hub

mrinalwadhwa:
- feat(elixir): change telemetry event name order
- feat(elixir): change router telemetry event names
- feat(elixir): allow modules as address types
- feat(elixir): add workers, node process registry and worker routing
- feat(elixir): move udp transport into ockam app and add udp routing
- feat(elixir): make telemetry optional
- refactor(elixir): remove ockam app dependency from ockam node web
- feat(elixir): add ranch tcp transport
- feat(elixir): add gen tcp client
- feat(elixir): add create listener and create client functions
- feat(elixir): add tcp transport registry
- feat(elixir): start tcp transport listener in ockam hub
- build(elixir): update ockam_hub dockerfile for runtime requirements
- docs(elixir): commands for deploying in azure
- style(elixir): fix typos
- refactor(elixir): replace udp address type constant with module attribute
- feat(elixir): add telemetry events to worker init


implementations/elixir/ockam/ockam/lib/ockam/node.ex
```diff
@@ -1,86 +1,124 @@
...
 
   @doc """
-  Returns a random address that is currently not registed with the node.
+  Returns a random address that is currently not registed on the node.
   """
-  def get_unused_address(length_in_bytes \\ @default_address_length_in_bytes) do
+  def get_random_unregistered_address(length_in_bytes \\ @default_address_length_in_bytes) do
     cadidate = :crypto.strong_rand_bytes(length_in_bytes)
 
     case whereis(cadidate) do
```
nilslice, line 58: not sure if this is a typo or not, for `candidate`

mrinalwadhwa: yes it is 🤦

implementations/elixir/ockam/ockam/lib/ockam/node.ex
```diff
@@ -1,86 +1,124 @@
...
       nil -> cadidate
-      _pid -> get_unused_address(length_in_bytes)
+      _pid -> get_random_unregistered_address(length_in_bytes)
+    end
+  end
+
+  @doc false
+  def start_supervised(module, options) do
```
nilslice, line 65: would we benefit from Telemetry added to this method (and `start_link` and `init`) ? maybe it is done behind the scenes.

nilslice: oh, I see some if it is done later.

mrinalwadhwa: the idea is to have telemetry be an optional feature/dependency in the `ockam` lib .. an application like `ockam_hub` will start it as part of its supervision tree.

implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/client/gen_tcp.ex
```diff
@@ -0,0 +1,44 @@
...
+  @doc false
+  @impl true
+  def setup(options, state) do
+    ip = Keyword.get_lazy(options, :ip, &default_ip/0)
+    port = Keyword.get_lazy(options, :port, &default_port/0)
+    tcp_connect_options = [:binary, active: :once]
+
+    state = Map.put(state, :ip, ip)
+    state = Map.put(state, :port, port)
+    state = Map.put(state, :tcp_connect_options, tcp_connect_options)
```
nilslice, line 15:
```suggestion
    state = Map.merge(state, %{
      ip: ip,
      port: port,
      tcp_connect_options: tcp_connect_options,
    })
```

unsure about precise syntax here, but this might grow in the future and could be better handled in a single assignment.

implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/listener/ranch.ex
```diff
@@ -0,0 +1,72 @@
+if Code.ensure_loaded?(:ranch) do
+  defmodule Ockam.Transport.TCP.Listener.Ranch do
+    @moduledoc false
+
+    use Ockam.Worker
+
+    @doc false
+    @impl true
+    def setup(options, state) do
```
nilslice, line 9: not sure how `Telemetry` is/isn't injected into modules, but this might also be a good event to track

implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/listener/ranch.ex
```diff
@@ -0,0 +1,72 @@
...
+      state = Map.put(state, :port, port)
+
+      ref = make_ref()
+      transport = :ranch_tcp
+      transport_options = [port: port]
+      protocol = __MODULE__.Protocol
+      protocol_options = []
+
+      with {:ok, _apps} <- Application.ensure_all_started(:ranch),
+           :ok <- start_listerner(ref, transport, transport_options, protocol, protocol_options) do
```
nilslice, line 23: assuming the Telemetry is wrapped in this call, however I think this is another typo: `start_listerner` -> `start_listener`

implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/client/gen_tcp.ex
mrinalwadhwa: possibly, although I think the more likely case is that individual option validation will move to separate functions.

implementations/elixir/ockam/ockam/lib/ockam/transport/udp.ex
```diff
@@ -0,0 +1,10 @@
+defmodule Ockam.Transport.UDP do
+  @moduledoc false
+
+  alias Ockam.Transport.UDP.Listerner
```
nilslice, line 4: `s/Listerner/Listener`

implementations/elixir/ockam/ockam/lib/ockam/transport/udp.ex
```diff
@@ -0,0 +1,10 @@
...
+
+  @doc false
+  def create_listener(options) do
+    Listerner.create(options)
```
nilslice, line 8: I'm sure you can fix this in one place, and the rest will throw somewhere when you try to run.. or a global search and replace for this and itself in lowercase.

implementations/elixir/ockam/ockam/lib/ockam/transport/udp/listener.ex
```diff
@@ -0,0 +1,118 @@
+defmodule Ockam.Transport.UDP.Listerner do
```
nilslice, line 1: guessing this is the place to fix the typo, and see the other misspell locations fail

implementations/elixir/ockam/ockam/lib/ockam/transport/udp/listener.ex
```diff
@@ -0,0 +1,118 @@
...
+      state = Map.put(state, :socket, socket)
+      state = Map.put(state, :udp_address, udp_address)
+      {:ok, state}
+    end
+  end
+
+  defp setup_routed_message_handler(true, listerner) do
+    handler = fn message -> handle_routed_message(listerner, message) end
+
+    with :ok <- Router.set_message_handler(2, handler),
```
nilslice, line 34: what's `2`?

implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/listener/ranch.ex
mrinalwadhwa: good point, I'll add to Ockam.Worker in a different PR
https://github.com/ockam-network/ockam/blob/mrinal/ockam_hub/implementations/elixir/ockam/ockam/lib/ockam/worker.ex#L61

anything module that does `use Ockam.Worker` already gets Telemetry events on handle_message.
I'll add it for setup as well 👍

implementations/elixir/ockam/ockam/lib/ockam/transport/udp/listener.ex
```diff
@@ -0,0 +1,118 @@
...
+    destination_and_onward_route =
+      message
+      |> Message.onward_route()
+      |> Enum.drop_while(fn a -> a === address end)
+      |> List.pop_at(0)
+
+    case destination_and_onward_route do
+      {nil, []} -> {:error, :no_destination}
+      {%UDPAddress{} = destination, r} -> {:ok, destination, %{message | onward_route: r}}
+      {{2, address}, onward_route} -> deserialize_address(message, address, onward_route)
```
nilslice, line 95: `2` again - maybe a few comments or some exported constant from wherever this should be defined would be more clear.

implementations/elixir/ockam/ockam/lib/ockam/transport/udp/listener.ex
mrinalwadhwa: thanks, its the udp address type in our binary protocol
I'll add module constant 👍

implementations/elixir/ockam/ockam/mix.exs
```diff
@@ -45,8 +45,9 @@ defmodule Ockam.MixProject do
 
   defp deps do
     [
-      {:telemetry, "~> 0.4.2"},
+      {:telemetry, "~> 0.4.2", optional: true},
       {:gen_state_machine, "~> 3.0"},
+      {:ranch, "~> 2.0", optional: true},
       {:ex_doc, "~> 0.23.0", only: :dev, runtime: false},
       {:credo, "~> 1.5", only: [:dev, :test], runtime: false},
       {:dialyxir, "~> 1.0", only: [:dev], runtime: false}
```
nilslice, line 53: out of curiosity, could `dialyxir` being `only: [:dev]` miss the chance that an external PR is accepted and breaks some type contract that the tests dont find? e.g. If the PR was made by someone who didn't run the code locally, just observed that tests pass locally or on CI. 

in the case that this happens, the code could be correct (using a value like 1 instead of some equivalent typed enum or otherwise), but the type checking would fail.

mrinalwadhwa: our CI builds in `dev`, that's the convention when linting elixir projects.
So when we turn on dialyxir in CI, it won't pass CI

mrinalwadhwa: I feel the need to add that since Elixir is dynamically typed, dialyzir will only do so much ... type errors are the cost of the convenience. That's why https://gleam.run/ is exciting as a tool in the BEAM toolbox.

implementations/elixir/ockam/ockam/lib/ockam/node.ex
mrinalwadhwa: fixed.

implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/listener/ranch.ex
mrinalwadhwa: fixed

implementations/elixir/ockam/ockam/lib/ockam/transport/udp/listener.ex
mrinalwadhwa: fixed

nilslice (approved):

PR #613, opened 2020-11-02
Sanjo/ockamd p256

SanjoDeundiak:


implementations/rust/vault/src/lib.rs
```diff
@@ -44,23 +44,34 @@ mod macros;
...
 pub mod types;
 
+use failure::_core::fmt::Debug;
+use std::any::Any;
 use types::*;
 
+/// Secret
+pub trait Secret: Debug + Sync + Send + 'static {
+    /// As any
+    fn as_any(&self) -> &dyn Any;
```
SanjoDeundiak, line 69: Needed to perform downcasts

implementations/rust/vault/src/file/mod.rs
```diff
@@ -99,22 +100,26 @@ fn id_to_path(id: usize) -> PathBuf {
     format!("{}.key", id.to_string()).into()
 }
 
+fn cast_secret(context: &Box<dyn Secret>) -> Result<usize, VaultFailError> {
```
SanjoDeundiak, line 103: Downcast

implementations/rust/vault/src/c/mod.rs
```diff
@@ -1,20 +1,771 @@
...
 use zeroize::Zeroize;
 
+/// C vault secret
+#[derive(Debug)]
+pub struct CVaultSecret {
+    handle: ockam_vault_secret_t,
+}
+
+// FIXME
+unsafe impl Send for CVaultSecret {}
```
SanjoDeundiak, line 20: Thoughts?

implementations/rust/vault/src/c/mod.rs
```diff
@@ -1,20 +1,771 @@
...
+
+// FIXME
+unsafe impl Sync for CVaultSecret {}
```
SanjoDeundiak, line 23: Thoughts?

implementations/rust/kex/src/xx.rs
```diff
@@ -4,37 +4,35 @@ use crate::{CipherSuite, NewKeyExchanger, AES_GCM_TAGSIZE};
...
     public_key: PublicKey,
-    secret_handle: SecretKeyContext,
+    secret_handle: Box<dyn Secret>,
 }
 
 /// Represents the XX Handshake]
 struct SymmetricState {
     cipher_suite: CipherSuite,
-    static_key_pair: Option<KeyPair>,
+    identity_key: Option<Arc<Box<dyn Secret>>>,
```
SanjoDeundiak, line 24: Arc is needed to prevent Secret Copy/Clone for long-term secrets. It looks like for ephemeral secrets it's possible to not use Copy/Clone even without Arc, since they have single owner

implementations/rust/kex/src/lib.rs
```diff
@@ -48,8 +48,9 @@ trait KeyExchange {
     /// Perform the diffie-hellman computation
     fn dh(
         &mut self,
-        secret_handle: SecretKeyContext,
+        secret_handle: &Box<dyn Secret>,
         public_key: PublicKey,
+        is_last: bool,
```
SanjoDeundiak, line 53: Workaround for Atecc...

implementations/rust/c/bindings/src/bindings.rs
```diff
@@ -290,6 +291,7 @@ pub enum ockam_vault_secret_persistence_t {
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub enum ockam_vault_secret_purpose_t {
     OCKAM_VAULT_SECRET_PURPOSE_KEY_AGREEMENT = 0,
+    OCKAM_VAULT_SECRET_PURPOSE_EPILOGUE = 1,
```
SanjoDeundiak, line 294: Workaround for Atecc...

implementations/c/ockam/vault/atecc608a/atecc608a.c
```diff
@@ -93,7 +93,7 @@ typedef struct {            /*!< Byte(s): Description
  * @brief EEPROM slot configuration data
  */
 typedef struct {
-  ockam_vault_secret_t* secret;
+  bool                  taken;
```
mikelodder7, line 96: what does `taken` mean?

implementations/rust/c/bindings/build.rs
```diff
@@ -40,9 +56,35 @@ fn main() {
     );
     println!("cargo:rustc-link-lib=static=ockam_vault_atecc608a");
 
+    println!(
```
mikelodder7, line 59: Is `log` a package? If not, I'm confused as to why we would link to a log directory?

implementations/rust/c/bindings/build.rs
```diff
@@ -40,9 +56,35 @@ fn main() {
...
+    );
+    println!("cargo:rustc-link-lib=static=ockam_mutex");
+
+    println!(
+        "cargo:rustc-link-search=native={}",
+        ockam_vault_output
+            .join("build/_deps/cryptoauth-build/lib")
+            .display()
+    );
+    println!("cargo:rustc-link-lib=static=cryptoauth");
```
mikelodder7, line 89: What is cryptoauth?

implementations/rust/c/bindings/src/bindings.rs
```diff
@@ -1035,3 +1030,51 @@ extern "C" {
...
+    #[doc = " @param   mutex[in] The ockam mutex implementation to use."]
+    #[doc = " @param   lock[in]  Lock object to unlock."]
+    #[doc = " @return  OCKAM_ERROR_NONE on success."]
+    pub unlock: ::core::option::Option<
+        unsafe extern "C" fn(mutex: *mut ockam_mutex_t, lock: ockam_mutex_lock_t) -> ockam_error_t,
+    >,
+}
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct ockam_mutex_t {
```
mikelodder7, line 1077: What are the benefits of this approach over trait objects? This seems rather complex.

implementations/rust/channel/src/lib.rs
```diff
@@ -247,6 +247,11 @@ impl<I: KeyExchanger, R: KeyExchanger, E: NewKeyExchanger<I, R>> ChannelManager<
 
         let channel = self.channels.get_mut(&cipher_address).unwrap();
         let mut channel = &mut *channel.lock().unwrap();
+        let agreement = match &mut channel.agreement {
+            Some(e) => e.as_mut(),
+            None => panic!(), // FIXME
```
mikelodder7, line 252: We shouldn't panic. Return an error instead.

implementations/rust/channel/src/lib.rs
```diff
@@ -359,8 +364,13 @@ impl<I: KeyExchanger, R: KeyExchanger, E: NewKeyExchanger<I, R>> ChannelManager<
 
     fn handle_m1_recv(&self, channel: Arc<Mutex<Channel>>, m: Message) -> Result<(), ChannelError> {
         let channel = &mut *channel.lock().unwrap();
-        channel.agreement.process(&m.message_body)?;
-        let m2 = channel.agreement.process(&[])?;
+        let agreement = match &mut channel.agreement {
+            Some(e) => e.as_mut(),
+            None => panic!(), // FIXME
```
mikelodder7, line 369: Again no panic.

implementations/rust/channel/src/lib.rs
```diff
@@ -379,9 +389,14 @@ impl<I: KeyExchanger, R: KeyExchanger, E: NewKeyExchanger<I, R>> ChannelManager<
 
     fn handle_m2_recv(&self, channel: Arc<Mutex<Channel>>, m: Message) -> Result<(), ChannelError> {
         let mut channel = &mut *channel.lock().unwrap();
+        let mut agreement = match channel.agreement.take() {
+            Some(e) => e,
+            None => panic!(), // FIXME
```
mikelodder7, line 394: Again no panic.

implementations/rust/channel/src/lib.rs
```diff
@@ -419,14 +435,24 @@ impl<I: KeyExchanger, R: KeyExchanger, E: NewKeyExchanger<I, R>> ChannelManager<
 
     fn handle_m3_recv(&self, channel: Arc<Mutex<Channel>>, m: Message) -> Result<(), ChannelError> {
         let mut channel = channel.lock().unwrap();
+        let mut agreement = match channel.agreement.take() {
+            Some(e) => e,
+            None => panic!(), // FIXME
```
mikelodder7, line 440: No panic

implementations/rust/daemon/src/node.rs
```diff
@@ -7,17 +7,24 @@ use crate::cli;
...
 use ockam_router::router::Router;
 use ockam_system::commands::{OckamCommand, RouterCommand};
 use ockam_transport::transport::UdpTransport;
+use ockam_vault::software::DefaultVaultSecret;
 use ockam_vault::types::*;
-use ockam_vault::{file::FilesystemVault, DynVault};
+use ockam_vault::{file::FilesystemVault, DynVault, Secret};
+use cfg_if::cfg_if;
+
+cfg_if! {
```
mikelodder7, line 22:
```suggestion
#[cfg(feature = "atecc608a)]
use ockam_vault::c::Atecc608aVaultBuilder;
#[cfg(feature = "atecc608a)]
use ockam_vault::c::CVault;
```

Removes the dependency on cfg_if

implementations/rust/kex/src/x3dh.rs
```diff
@@ -179,20 +176,19 @@ enum InitiatorState {
...
-    identity_key: Option<SecretKeyContext>,
+    identity_key: Option<Arc<Box<dyn Secret>>>,
 }
 
 impl X3dhInitiator {
-    fn new(v: Arc<Mutex<dyn DynVault + Send>>, identity_key: Option<SecretKeyContext>) -> Self {
+    fn new(v: Arc<Mutex<dyn DynVault + Send>>, identity_key: Option<Arc<Box<dyn Secret>>>) -> Self {
         Self {
-            ephemeral_identity_key: SecretKeyContext::Memory(0),
+            ephemeral_identity_key: Box::new(DefaultVaultSecret::default()), // FIXME: Should be Option<Secret>
```
mikelodder7, line 191: Why isn't it an option?

implementations/rust/kex/src/xx.rs
mikelodder7: How?

implementations/rust/vault/src/c/mod.rs
mikelodder7: Could this be changed to just be a pointer directly rather than wrapping the secret_t? In other words, could we just use *mut ockam_vault_secret_t?

implementations/rust/vault/src/c/mod.rs
```diff
@@ -1,20 +1,771 @@
...
+        CVault::handle_error(error)?;
+
+        Ok(CVault::new(c_vault))
+    }
+}
+
 // TODO: Should be thread-safe?
 /// Represents a single instance of an Ockam vault context
 #[derive(Debug)]
 pub struct CVault {
```
mikelodder7, line 111: Thread safety is better assured by using a Mutex.

mikelodder7: Like using Arc<Mutex

implementations/rust/vault/src/c/mod.rs
```diff
@@ -1,20 +1,771 @@
...
-    context: Cell<ockam_vault_t>,
+    context: ockam_vault_t,
+}
+
+impl Drop for CVault {
```
mikelodder7, line 115: I don't think the drop is necessary since it's not deleting anything and just a reference pointer.

implementations/rust/vault/src/c/mod.rs
```diff
@@ -1,20 +1,771 @@
...
+            let str = unsafe { CStr::from_ptr(error.domain) };
+            println!("ERROR: {:?}, {}", str, error.code);
+            Err(error.into())
+        }
+    }
+
+    fn cast_secret(context: &Box<dyn Secret>) -> Result<&ockam_vault_secret_t, VaultFailError> {
+        match context.as_any().downcast_ref::<CVaultSecret>() {
+            Some(secret) => Ok(&secret.handle),
+            None => panic!(), //FIXME,
```
mikelodder7, line 139: Change to error.

implementations/rust/vault/src/c/mod.rs
```diff
@@ -1,20 +1,771 @@
...
+        }
+    }
+
+    fn cast_secret_consume(
+        context: Box<dyn Secret>,
+    ) -> Result<ockam_vault_secret_t, VaultFailError> {
+        match context.as_any().downcast_ref::<CVaultSecret>() {
+            Some(secret) => Ok(secret.handle),
+            None => panic!(), //FIXME,
```
mikelodder7, line 148: change to error.

implementations/rust/vault/src/c/mod.rs
```diff
@@ -1,20 +1,771 @@
...
+            SecretKeyType::Buffer(len) => len as u16,
+            _ => {
+                println!("TYPE: {:?}", self.xtype);
+                panic!()
+            }
+        };
+
+        ockam_vault_secret_attributes_t {
+            length,
+            type_: self.xtype.into(),
```
mikelodder7, line 271:
```suggestion
            xtype: self.xtype.into(),
```

implementations/rust/c/bindings/src/helpers.rs
```diff
@@ -7,3 +8,34 @@ pub fn ockam_error_is_none(error: &ockam_error_t) -> bool {
 pub fn ockam_error_has_error(error: &ockam_error_t) -> bool {
     error.code != 0
 }
+
+impl Default for ockam_vault_secret_t {
+    fn default() -> Self {
+        ockam_vault_secret_t {
+            attributes: ockam_vault_secret_attributes_t {
+                length: 0,
+                type_: ockam_vault_secret_type_t::OCKAM_VAULT_SECRET_TYPE_BUFFER,
```
mikelodder7, line 17: Here it's using type_ but elsewhere in the code, it's using xtype. Need to be consistent.

implementations/rust/vault/src/c/mod.rs
```diff
@@ -1,20 +1,771 @@
...
+            )
+        };
+        Self::handle_error(error)?;
+
+        match handle.attributes.type_ {
+            ockam_vault_secret_type_t::OCKAM_VAULT_SECRET_TYPE_P256_PRIVATEKEY => {
+                if output_size != 65 {
+                    return Err(VaultFailError::from(VaultFailErrorKind::InvalidSecret));
+                }
+                Ok(PublicKey::P256(*array_ref![buffer, 0, 65]))
```
mikelodder7, line 458:
```suggestion
                Ok(PublicKey::P256(buffer))
```

implementations/rust/vault/src/file/mod.rs
mikelodder7: Renaming to `downcast` would be better.

implementations/rust/vault/src/software/mod.rs
```diff
@@ -41,17 +63,20 @@ impl Default for DefaultVault {
 }
 
 impl DefaultVault {
+    fn cast_secret(context: &Box<dyn Secret>) -> Result<usize, VaultFailError> {
```
mikelodder7, line 66: This feels like an unwrap. Maybe that's a better name for it.

implementations/rust/vault/src/software/mod.rs
```diff
@@ -63,37 +88,43 @@ impl DefaultVault {
...
+        salt: &Box<dyn Secret>,
         info: &[u8],
         ikm: &[u8],
         output_attributes: Vec<SecretKeyAttributes>,
-    ) -> Result<Vec<SecretKeyContext>, VaultFailError> {
+    ) -> Result<Vec<Box<dyn Secret>>, VaultFailError> {
         let salt = self.get_entry(salt, VaultFailErrorKind::Ecdh)?;
 
         // FIXME: Doesn't work for secrets with size more than 32 bytes
         let okm_len = output_attributes.len() * 32;
```
mikelodder7, line 99: Either we should throw an error if more than 32 bytes is asked or handle it under the hood with multiple calls.

implementations/rust/vault/src/types.rs
```diff
@@ -119,20 +123,24 @@ try_from_int_impl!(SecretPersistenceType, u128);
 pub enum SecretPurposeType {
     /// Key exchange
     KeyAgreement,
+    /// Epilogue
+    Epilogue,
```
mikelodder7, line 127: Can we get a more detailed comment about what `Epilogue` is?
What is it used for?

implementations/rust/vault/src/types.rs
```diff
@@ -293,6 +222,8 @@ pub enum SecretKey {
     Curve25519([u8; 32]),
     /// NIST P-256 (secp256r1, prime256v1) secret key
     P256([u8; 32]),
+    /// ChainKey
+    ChainKey([u8; 32]),
```
mikelodder7, line 226: This feels like the opposite of modularity. ChainKey is a key exchange term. Key exchange terms shouldn't bleed into vault.

mikelodder7 (commented):
This is a massive PR. Is it possible to break this into smaller PRs to have a better understanding of what all the changes are for?

implementations/rust/vault/src/c/mod.rs
SanjoDeundiak: Maybe yes, but I don't like the idea of populating C structures all over the place. Native Rust struct that hides all this nasty details under the hood is IMHO better

implementations/rust/kex/src/xx.rs
SanjoDeundiak: If you cannot copy/clone identity key but want to use it from different places/threads you can achieve that by wrapping it inside Arc.
With ephemeral keys it's simpler, cause you just don't use it from different places

implementations/c/ockam/vault/atecc608a/atecc608a.c
SanjoDeundiak: Previously ATECC vault stored pointers to all secrets, but it lead to memory management problems + it wasn't really necessary functionality. So, instead I replaced secret pointer with bool value that just indicates that given slot is used for some secret.

implementations/rust/c/bindings/build.rs
SanjoDeundiak: Microchip's lib

implementations/rust/c/bindings/src/bindings.rs
SanjoDeundiak: This is bindings to C object (we use mutex inside C library). Though, it's not used at this moment from rust

implementations/rust/channel/src/lib.rs
SanjoDeundiak: Agree, this will be fixed in actual mergeable PRs

implementations/rust/daemon/src/node.rs
SanjoDeundiak: Why don't we want to depend on cfg_if?

implementations/rust/kex/src/x3dh.rs
SanjoDeundiak: Needs some changes in the places of usage, I'll do it in further PRs

implementations/rust/vault/src/c/mod.rs
SanjoDeundiak: Agree

implementations/rust/vault/src/c/mod.rs
SanjoDeundiak: Yes, we have Vault::deinit instead. But I'm not sure we use it...

implementations/rust/vault/src/c/mod.rs
SanjoDeundiak: Requires changes to C library

implementations/rust/c/bindings/src/helpers.rs
SanjoDeundiak: The reason to that is that type cannot be used as a property name in rust, right? But this code is bindings from c, and c doesn't have that restriction. So why would we change property name in c code?

implementations/rust/vault/src/software/mod.rs
SanjoDeundiak: I think it should be called "downcast_secret"

implementations/rust/vault/src/types.rs
SanjoDeundiak: This is workaround for atecc, because during epilogue of key agreement it uses different approach about which secrets go to which slots.

implementations/rust/vault/src/types.rs
SanjoDeundiak: To which category from existing ones does ChainKey goes then?

SanjoDeundiak:
> This is a massive PR. Is it possible to break this into smaller PRs to have a better understanding of what all the changes are for?

This is the plan

implementations/rust/vault/src/c/mod.rs
mikelodder7: Revisit later

PR #661, opened 2020-11-21
feat(rust): implement tcp message size, add tests to node crate

robinbudd:


robinbudd:
@mrinalwadhwa @mikelodder7 Review and merge please?

implementations/rust/transport/src/tcp.rs
```diff
@@ -151,11 +150,25 @@ impl TcpManager {
...
             for a in &self.addresses {
+                let mut l = 0;
                 match self.connections.get_mut(&a.to_string()) {
                     Some(t) => {
-                        if let Err(s) = t.receive_message() {
-                            println!("tcp receive failed {}", s);
-                            return false;
+                        loop {
+                            l += 1;
+                            match t.try_receive() {
```
mikelodder7, line 158: should we use a timeout rather than returning immediately?

robinbudd: We could, but I don't like blocking for any time in the case where we're running on a single-threaded system. We'll try again at the next polling interval.

PR #687, opened 2020-12-04
feat(rust): Queues and Topics

jared-s:
# Queue and Topics

Traits and in-memory implementation of basic Queues, Topics and "workers" for each.

An example Queue using Redis is under the `examples/redis/` directory.


jared-s:
Cleaning this up in another branch.

PR #689, opened 2020-12-06
refactor(rust): implement prototype of refactor for messaging platform

robinbudd:
refactor router -> message_router
refactor tcp_manager
implement worker_manager
refactor node binary & lib to lib crate
move node binary to integration_test
implement no_std_traits for traits that don't rely on std


implementations/rust/refactor_prototype/message_router/src/lib.rs
```diff
@@ -0,0 +1,79 @@
+#![no_std]
+extern crate alloc;
+use alloc::boxed::Box;
```
mikelodder7, line 3: Probably want all the `use alloc` statements in a cfg_if block since this is not needed for std mode.

robinbudd: Good point, there's a lot of work to be done when we configure for std/no_std. This is just a prototype.

mikelodder7 (approved):

PR #690, opened 2020-12-07
chore(rust): make secretkey and publickey more usable

SanjoDeundiak:

### DIFF
--- a/implementations/rust/kex/src/x3dh.rs
+++ b/implementations/rust/kex/src/x3dh.rs
@@ -53,10 +53,10 @@ impl PreKeyBundle {
     /// Convert the prekey bundle to a byte array
     pub fn to_bytes(&self) -> Vec<u8> {
         let mut output = Vec::new();
-        output.extend_from_slice(self.identity_key.0.as_slice());
-        output.extend_from_slice(self.signed_prekey.0.as_slice());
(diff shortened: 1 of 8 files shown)


implementations/rust/kex/src/x3dh.rs
```diff
@@ -413,9 +414,9 @@ impl KeyExchanger for X3dhInitiator {
 
                 // Check the prekey_bundle signature
                 vault.verify(
-                    prekey_bundle.signature_prekey.0,
-                    prekey_bundle.identity_key.0.as_slice(),
-                    prekey_bundle.signed_prekey.0.as_slice(),
+                    prekey_bundle.signature_prekey.as_ref().clone(), /* FIXME */
```
SanjoDeundiak, line 417: @mikelodder7 should verify take array reference instead of array?

mikelodder7: yes it should

mikelodder7 (changes requested):
just make the one change and it should be good.

implementations/rust/kex/src/x3dh.rs
SanjoDeundiak: @mikelodder7 addressed in #704

mikelodder7 (approved):

PR #725, opened 2020-12-22
feat(rust): replacing ffi_support

SanjoDeundiak:


implementations/rust/ffi/include/kex.h
```diff
@@ -0,0 +1,88 @@
+//
+// Created by Oleksandr Deundiak on 10.09.2020.
```
mikelodder7, line 2:
```suggestion
// Created by Ockam Developers
```

implementations/rust/ffi/src/kex.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+/// Create a new kex initiator and return it
+#[no_mangle]
+pub extern "C" fn ockam_kex_xx_initiator(
+    context: &mut FfiKexFatPointer,
+    vault: FfiVaultFatPointer,
+    identity_key: u64,
+) -> FfiOckamError {
+    let res = || {
```
mikelodder7, line 54: I don't understand this. Why are you creating an anonymous function then immediately calling it?
What's the purpose of the anonymous function?

SanjoDeundiak: The purpose is to allow using ? operator inside

mikelodder7 (approved):

implementations/rs/common/src/error.rs
```diff
@@ -19,7 +19,7 @@ impl OckamError {
         self.code
     }
     /// Error domain
-    pub fn domain(&self) -> &str {
+    pub fn domain(&self) -> &'static str {
```
mikelodder7, line 22: good change

mikelodder7 (approved):

PR #732, opened 2020-12-28
feat(rust): app and service apps

mikelodder7:
Updates the okta example to use two different processes: the service and the enroller.

### DIFF
--- a/implementations/rust/okta/management_app/src/main.rs
+++ b/implementations/rust/okta/management_app/src/main.rs
@@ -1,0 +1,269 @@
+#[macro_use]
+extern crate lazy_static;
+
+mod config;
+
+use ockam::{
+    vault::*,
+    vault::types::*,
+    vault_software::*,
+};
+use oktaplugin::{Messages, OckamMessages, OckamRole, KeyEstablishment, EstablishmentBundle};
+use rand::prelude::*;
+use std::{
+    fs,
+    io::{self, Write},
+    net::TcpStream,
+    path::Path,
+    sync::Mutex,
+};
+use structopt::StructOpt;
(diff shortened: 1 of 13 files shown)


implementations/rust/okta/Cargo.lock
```diff
@@ -275,6 +275,15 @@ version = "0.3.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f878195a49cee50e006b02b93cf7e0a95a38ac7b776b4c4d9cc1207cd20fcb3d"
 
+[[package]]
```
SanjoDeundiak, line 278: Do we still want to include Cargo.lock files in our repo?

implementations/rust/okta/management_app/src/main.rs
```diff
@@ -0,0 +1,204 @@
...
+const FILE_NAME: &str = ".env";
+
+lazy_static! {
+    static ref VAULT: Mutex<DefaultVault> = Default::default();
+}
+
+fn main() {
+    let cfg =
+        if Path::new(FILE_NAME).is_file() {
+            let contents = fs::read_to_string(FILE_NAME).unwrap();
```
SanjoDeundiak, line 35: Would be nice to remove unwraps

SanjoDeundiak (changes requested):

implementations/rust/okta/management_app/src/main.rs
mikelodder7: CLI's are okay to have unwraps. This is just a demo anyway

SanjoDeundiak (approved):

jared-s:
Closing  - extremely out of date, using CI resources

PR #739, opened 2021-01-04
feat(rust): ockam_macros proc macro crate

jared-s:
# ockam_macros crate

This crate holds macros for the Ockam core. One macro is defined, an async wrapper around the `executor` crate. When we add `tokio` support, we can add it similarly.

### DIFF
--- a/implementations/rust/ockam_macros/Cargo.toml
+++ b/implementations/rust/ockam_macros/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "ockam_macros"
+version = "0.1.0"
+authors = ["Ockam Developers"]
+edition = "2018"
+
+[lib]
+proc-macro = true
+
+[features]
+default = []
+async = []
+executor_async = ["async", "executor"]
+
+[dependencies]
+executor = {version = "*", optional = true}
+
--- a/implementations/rust/ockam_macros/src/lib.rs
+++ b/implementations/rust/ockam_macros/src/lib.rs
@@ -1,0 +1,79 @@
+extern crate proc_macro;
+
+use self::proc_macro::{Delimiter, TokenStream, TokenTree};
+use core::str::FromStr;
+
+#[proc_macro_attribute]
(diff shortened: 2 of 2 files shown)


robinbudd (approved):

implementations/rust/ockam_macros/src/lib.rs
```diff
@@ -0,0 +1,79 @@
+extern crate proc_macro;
+
+use self::proc_macro::{Delimiter, TokenStream, TokenTree};
+use std::str::FromStr;
```
mrinalwadhwa, line 4: Can we avoid this use of `std`?

implementations/rust/ockam_macros/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+[lib]
+proc-macro = true
+
+[features]
+default = []
+async = []
+executor_async = ["async", "executor"]
+
+[dependencies]
+executor = {version = "*", optional = true}
```
mrinalwadhwa, line 16: could you add a trailing empty line to the file
our `.editorconfig` file specifies that but we don't have a linter for it yet.

mrinalwadhwa (changes requested):

mrinalwadhwa:
@jared-s the commit linter doesn't allow capital letters in the commit message.

implementations/rust/ockam_macros/src/lib.rs
jared-s: Yep!

mrinalwadhwa:
Awesome!

The changes here look ready to merge, I think we can squash the 3 commits into one, under this original commit message:

```
feat(rust): ockam_macros proc macro crate
```

Once you push that squashed commit, and the build passes, I'll merge

mrinalwadhwa:
Fixes #741

mrinalwadhwa (approved):

mrinalwadhwa:
@jared-s the commit requires a signature before we merge.

PR #746, opened 2021-01-07
feat(rust): address structure and traits

jared-s:
Implements a basic, typeless Address with String internal storage. Several convenience traits are implemented.

1. Address internal storage is a String, for convenience and easy cloning.
2. Implements various traits for conversion to and from Address and String
3. Implements Hash so that it can be used in hash tables.

We should decide whether or not we want to introduce the notion of `Address Type`.


mrinalwadhwa (approved):

mrinalwadhwa:
Merged https://github.com/ockam-network/ockam/commit/41e4e214ddd12166761237853541bf3bde6ec4f9 with some minor changes to the commit message.

PR #748, opened 2021-01-07
feat(rust): queue trait

jared-s:
Provides the basic Queue trait.

PR #749, opened 2021-01-08
node attribute macro and node creates

mrinalwadhwa:
chore(rust): add gitignore
feat(rust): ockam node attribute macro
feat(rust): no_std node
feat(rust): tokio node

based off #748, should be merged after that.


jared-s (approved):
:tada:

implementations/rust/ockam/src/queue.rs
```diff
@@ -0,0 +1,44 @@
+use crate::address::Addressable;
+use alloc::collections::VecDeque;
+
+pub trait Queue<T> {
+    fn enqueue(&mut self, element: T) -> crate::Result<bool>;
+    fn dequeue(&mut self) -> Option<T>;
+    fn is_empty(&self) -> bool;
+}
```
robinbudd, line 8: I think all three of us (you, Jared and I) have all implemented some version of a queue!

robinbudd (approved):

implementations/rust/ockam/src/queue.rs
mrinalwadhwa: this code is in this PR only because my branch is based on Jared's queue PR  #748

mrinalwadhwa: so one less queue 😄

PR #750, opened 2021-01-08
feat(rust): address type enum and address member

jared-s:
This provides a minimal AddressType enum definition: Worker and Undefined. I held off including the transport specific address types in case we discover they should be modeled differently.

### DIFF
--- a/implementations/rust/ockam/src/address.rs
+++ b/implementations/rust/ockam/src/address.rs
@@ -4,15 +4,31 @@ use core::fmt::{Display, Formatter};
 use core::hash::{Hash, Hasher};
 
 #[derive(Eq, PartialEq, Clone, Debug)]
+pub enum AddressType {
+    Worker = 0,
+    Undefined = 255,
+}
+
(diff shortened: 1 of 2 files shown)


jared-s:
@robinbudd @mrinalwadhwa for exploring the Address Type question, here is a minimal def and branch

implementations/rust/ockam/src/address.rs
```diff
@@ -3,17 +3,33 @@ use alloc::string::{String, ToString};
 use core::fmt::{Display, Formatter};
 use core::hash::{Hash, Hasher};
 
+#[derive(Eq, PartialEq, Clone, Debug)]
+pub enum AddressType {
+    Worker = 0,
+    Undefined = 255,
+}
+
```
mrinalwadhwa, line 11: I think it would be nice if we can avoid a global list of address types.
Each transport knows the number (representing its address type) it has picked but there is no global list.

jared-s:
Including in larger PR

PR #763, opened 2021-01-13
rust impl structure

mrinalwadhwa:


mrinalwadhwa:
@robinbudd @jared-s 
- note the move of ockam crates from `implementations/rust/` to `implementations/rust/ockam/`
- and `implementations/rust/examples` to `implementations/rust/examples/workers` ... so we can have many example crates under `implementations/rust/examples`

mrinalwadhwa:
this makes the C, Rust and Elixir impl top level folders roughly identical.

jared-s (approved):
Cool! :+1:

PR #766, opened 2021-01-14
Sanjo/contact prototype

SanjoDeundiak:


implementations/rs/ockam/src/contact/contact.rs
```diff
@@ -0,0 +1,39 @@
...
+            vault,
+        }
+    }
+}
+
+impl Contact {
+    pub fn identifier(&self) -> &str {
+        &self.identifier
+    }
+    pub fn events(&self) -> &Vec<ContactEvent> {
```
mikelodder7, line 33:
```suggestion
    pub fn events(&self) -> &[ContactEvent] {
```

mikelodder7 (changes requested):
Just one small change. Lots of similarities between contact and profile. A subsequent PR could extract the common components.

implementations/rs/ockam/src/contact/contact.rs
SanjoDeundiak: @mikelodder7 I wonder why returning slice is better

SanjoDeundiak:
> Just one small change. Lots of similarities between contact and profile. A subsequent PR could extract the common components.

I agree, but I'm tolerant to copy&pasting code if logically two pisces should be separate

implementations/rs/ockam/src/contact/contact.rs
SanjoDeundiak: The only case coming up to my mind - it simplifies things when we want to use Vec vs something heapless depending on enabled features

mikelodder7 (approved):

PR #770, opened 2021-01-19
prototype of profile verification policies and purpose

mrinalwadhwa:
This starts with the key signature verification method and tries to structure it as a collection of verification policies.
Along the lines of https://w3c.github.io/did-spec-registries/#ed25519verificationkey2018

### DIFF
--- a/implementations/rust/ockam/ockam/src/entity/profile.rs
+++ b/implementations/rust/ockam/ockam/src/entity/profile.rs
@@ -4,10 +4,14 @@ pub use identifier::*;
 mod change;
 pub use change::*;
 
+mod verification;
+pub use verification::*;
+
 #[derive(Clone, Debug)]
 pub struct Profile {
     pub identifier: ProfileIdentifier,
     pub change_history: ProfileChangeHistory,
+    pub verification_policies: Vec<ProfileVerificationPolicy>,
 }
 
 impl Profile {
@@ -15,6 +19,7 @@ impl Profile {
         Profile {
             identifier: ProfileIdentifier::new(),
             change_history: ProfileChangeHistory::new(),
+            verification_policies: vec![],
         }
     }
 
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam/src/entity/profile/rule.rs
```diff
@@ -0,0 +1,19 @@
+use super::*;
+
+#[derive(Clone, Debug)]
+pub struct ProfileRuleEd25519Verification {
+    public_key: Vec<u8>,
```
SanjoDeundiak, line 5: Should we add id to each rule?

mrinalwadhwa: That's a good question, I'm not sure.
Maybe we can use the index in the vector that contains the policies authoritative on which policy is being updated. 🤔

mrinalwadhwa:
In a discussion with @mikelodder7 yesterday, we decided that policies was a better (more consistent) word than rules.

In the latest push I broke down VerificationPolicy, VerificationMethod, VerificationPurpose. A policy defines a verification method and a list of purposes it can be used for.

In the DID spec, DID documents, index by purpose - which is why DID docs have properties like `authentication` which is good when you're trying to authenticate and need to lookup the valid verification method for authentication.

From the perspective of our change_events having just one way of updating policies seems better. We can create an index by purpose on top of that, if we need the lookup optimization.

implementations/rust/ockam/ockam/src/entity/profile/rule.rs
SanjoDeundiak: @mrinal I think to simplify DID compliance we need to introduce string (human-readable or not) that would be id for VerificationMethod

mrinalwadhwa: I was thinking something like this ... when serializing to a DID, we could serialize verification policy at index `1` to did doc with verification method:

```
{
  id: "did:ockam:1234"
  "verificationMethod": [{
      "id": "did:ockam:1234#verification_policy_1"
      ...
  }
}
```

either way, let's tackle it when we get to DID serialization of a profile.
for the time being I'm recording it in an issue https://github.com/ockam-network/ockam/issues/778

mikelodder7 (approved):

PR #782, opened 2021-01-20
feature(rust): use async/await for worker creation

jared-s:
1. Remove started and stopped handler callbacks
2. Made `WorkerBuilder::start` async
3. Deleted closure related code.
4. Deleted old tests: Rust unit tests can't use async. We will need to rewrite tests using `tokio_test`.
5. All tests now use async


jared-s:
As @robinbudd has mentioned before, doing anything useful with `async` immediately required upgrading all heap wrappers to be `Arc<Mutex<T>>` and `Send` elsewhere. For now, I've just done that throughout this code. We should do another pass in a future PR that uses the switchable lock trait macro.

mrinalwadhwa (approved):
Exciting 👍

PR #786, opened 2021-01-25
Sanjo/new vault

SanjoDeundiak:


implementations/rust/ockam/ockam_vault/traits/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+license = "Apache-2.0"
+
+[lib]
+crate-type = ["rlib"]
+
+[dependencies]
+ockam_error = {path = "../../ockam_error", version = "*"}
+ockam_macro = {path = "../../ockam_macro", version = "*"}
+zeroize = { version = "1.1", features = ["zeroize_derive"] }
+cfg-if = "1.0"
```
mrinalwadhwa, line 15: do we really need this dependency?
could we use https://doc.rust-lang.org/1.49.0/std/macro.cfg.html instead?

implementations/rust/ockam/ockam_vault/traits/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+#[macro_use]
+extern crate cfg_if;
+#[macro_use]
+extern crate ockam_macro;
```
mrinalwadhwa, line 4: we can just `use macro` in Rust 2018
https://youtu.be/AN9FoZgLcFg?t=1387

mrinalwadhwa:
@SanjoDeundiak before we merge this, could you do some simple moves:

1. Move software vault to a crate called `ockam_vault`

2. Move traits to a create called `ockam_vault_core`

Both crates in `implementations/rust`

implementations/rust/ockam/ockam_vault/traits/Cargo.toml
SanjoDeundiak: @mrinalwadhwa these crates are designed for different purpose. cfg-if is works exactly like standard rust cfg, but allows you to have multiple statements under one cfg. It's just for convenience and it's very lightweight (the whole lib is one macro). std::cfg does not support conditional compilation

mrinalwadhwa: Great 👍

SanjoDeundiak:
@mrinalwadhwa could have been simple, if I didn't have a pile of changes on top of it :D Fixed, anyways

mrinalwadhwa:
Thank you. 🙏 
One more thing, could you stack this on top of #791 - will need changing `OckamResult<T>` to `Result<T, ockam_core::Error>`

mrinalwadhwa (approved):

PR #791, opened 2021-01-27
improve error

mrinalwadhwa:
- feat(rust): improve error and move it to ockam core


implementations/rust/ockam/ockam_core/src/error/no_std_error.rs
```diff
@@ -0,0 +1,14 @@
+#[derive(Debug)]
+pub struct Error {
+    code: u32,
```
SanjoDeundiak, line 3: What stops us from having string error domain without std?

implementations/rust/ockam/ockam_core/src/error/std_error.rs
```diff
@@ -1,35 +1,29 @@
-use std::fmt::{Display, Formatter};
+use std::fmt::{Display, Formatter, Result};
 
-/// Common error
 #[derive(Debug)]
-pub struct OckamError {
+pub struct Error {
```
SanjoDeundiak, line 4: I thought everybody was against renaming OckamError -> Error

implementations/rust/ockam/ockam/src/error.rs
```diff
@@ -0,0 +1,17 @@
+use ockam_core::Error;
+
+#[derive(Clone, Copy, Debug)]
+pub enum OckamError {
+    None,
+}
+
+impl OckamError {
+    pub const DOMAIN_CODE: u32 = 10_000;
```
SanjoDeundiak, line 9: Why do we need it?

mikelodder7: I agree. How is this better than what Sanjo had before?

implementations/rust/ockam/ockam_core/src/error/std_error.rs
mikelodder7: I am

mrinalwadhwa:
@SanjoDeundiak @mikelodder7 

Some context:

1. I would like the error exposed to applications to be called - `ockam::Error` so it is consistent with other types were exposing like `ockam::Node`, `ockam::Entity`, `ockam::Profile` ... `ockam::OckamError` is not consistent. 

2. Static strings cost code space in embedded settings. So we want no_std env's to have the option to not use them.
(we may want more feature flags around this)

This allows the code to be unique even when there is not string domain name
https://github.com/ockam-network/ockam/blob/4a3bb01f3f05bbb4a850139f60d2e48aa3d161cb/implementations/rust/ockam/ockam/src/error.rs#L8-L11

We'll have to manage DOMAIN_CODE to be unique across modules.

mrinalwadhwa:
I tried switching the Error::new constructor based on `std`or `no_std` for the OckamError above but since `ockam` crate doesn't yet compile in `no_std` because of Vec being used in some places.. we'll have to come back to that.

And to avoid repetition in all domain specific error types maybe have derive macro like this error

mikelodder7:
@SanjoDeundiak 's version uses `&'static str` which is no_std compliant and we had set the Vec to be no_std compliant as well. See #683 as an example.

mrinalwadhwa:
We decided to merge this for now, we'll revisit the discussion next week.
Thank you @mikelodder7 & @SanjoDeundiak, let's discuss more next week.

PR #792, opened 2021-01-27
Sanjo/new vault2

SanjoDeundiak:


implementations/rust/ockam/ockam_vault_core/src/open_close_vault.rs
```diff
@@ -2,6 +2,12 @@ use crate::vault::Vault;
 use ockam_error::OckamResult;
 
 pub trait OpenCloseVault: Sized {
+    type InnerVault;
+
+    fn get_data_mut(&mut self) -> &mut Self::InnerVault;
```
mikelodder7, line 7: I'm not understanding the purpose for these functions.

implementations/rust/ockam/ockam_vault/src/lib.rs
```diff
@@ -1,71 +1,10 @@
...
-        let vault = vault.open().unwrap();
-
-        let hash = vault.sha256(&[0u8; 32]).unwrap();
-        assert_eq!(
-            hex::encode(hash),
-            "66687aadf862bd776c8fc18b8e9f8e20089714856ee233b3902a591d0d5f2925"
-        );
-    }
-}
+pub extern crate ockam_vault_core;
```
mikelodder7, line 1: Do we want to alias this crate or perhaps only export a subset of the crate?

SanjoDeundiak: @mrinalwadhwa I think it's question to you

implementations/rust/ockam/ockam_vault_core/src/open_close_vault.rs
SanjoDeundiak: @mikelodder7 it allows to use vaults like this:
```rs
let mut vault = SoftwareVault::default();
let mut vault = vault.open().unwrap();
let mut attributes = SecretAttributes {
    stype: SecretType::Curve25519,
    persistence: SecretPersistence::Ephemeral,
    length: CURVE25519_SECRET_LENGTH,
};

let res = vault.secret_generate(attributes);
```

So, SoftwareVault implements OpenCloseVault trait, but none of other vault traits. InnerVault is SoftwareVaultImpl and implements all the other vault traits (like sign, verify, encrypt, etc.)

mikelodder7 (approved):

mrinalwadhwa (approved):

PR #796, opened 2021-01-28
Profile

SanjoDeundiak:


implementations/rust/ockam/ockam/src/entity/profile.rs
```diff
@@ -31,40 +69,258 @@ impl Profile {
...
     }
 }
 
 impl Profile {
-    pub fn apply(&mut self, change_event: ProfileChangeEvent) {
-        if !self.verify(&change_event) {
-            return; // TODO: Throw error
+    pub fn create_with_key(
+        label: String,
+        key_type: ProfileKeyType,
```
mikelodder7, line 86: can we collapse key_type and key_purpose into the key_attributes?

implementations/rust/ockam/ockam/src/entity/profile.rs
```diff
@@ -31,40 +69,258 @@ impl Profile {
...
+        key_purpose: ProfileKeyPurpose,
+        attributes: Option<ProfileEventAttributes>,
+        vault: Arc<Mutex<dyn ProfileVault>>,
+    ) -> Result<Self, ockam_core::Error> {
+        let attributes = attributes.unwrap_or(ProfileEventAttributes::new());
+
+        let mut v = vault.lock().unwrap();
+
+        // TODO: Should be customisable
+        let secret_attributes = SecretAttributes {
```
mikelodder7, line 96: Love this, we should emulate it for the caller into this.

implementations/rust/ockam/ockam/src/entity/profile.rs
```diff
@@ -31,40 +69,258 @@ impl Profile {
...
+            persistence: SecretPersistence::Persistent,
+            length: 0,
+        };
+
+        let private_key = v.secret_generate(secret_attributes)?;
+        let public_key = v.secret_public_key_get(&private_key)?;
+
+        let change =
+            CreateKeyChange::new(label, key_type, key_purpose, public_key.as_ref().to_vec());
+        let prev_id = v.sha256(&[])?;
```
mikelodder7, line 107: I think we should hash the public key vs an empty byte array, otherwise, it will be the same value for all profiles.

implementations/rust/ockam/ockam/src/entity/profile.rs
```diff
@@ -31,40 +69,258 @@ impl Profile {
...
+    }
+}
+
+impl Profile {
+    fn check_consistency(_change_event: &ProfileChangeEvent) -> bool {
+        // TODO: check event for consistency: e.g. you cannot rotate the same key twice during one event
+        true
+    }
+
+    fn apply_no_verification(&mut self, change_event: ProfileChangeEvent) -> Result<(), Error> {
```
mikelodder7, line 182:
```suggestion
    fn apply_no_verification(&self, change_event: ProfileChangeEvent) -> Result<(), Error> {
```

mikelodder7, line 182: Not sure why this is needed vs apply.

implementations/rust/ockam/ockam/src/entity/profile.rs
```diff
@@ -31,40 +69,258 @@ impl Profile {
...
+                    c.key_type(),
+                    c.key_purpose(),
+                )?;
+                let prev_key_change = Self::find_key_change_in_event(
+                    prev_key_event,
+                    c.label(),
+                    c.key_type(),
+                    c.key_purpose(),
+                )
+                .unwrap();
```
mikelodder7, line 255: Why the unwrap here?

implementations/rust/ockam/ockam/src/entity/profile.rs
```diff
@@ -31,40 +69,258 @@ impl Profile {
...
         }
+
+        self.apply_no_verification(change_event)
     }
 
-    pub fn verify(&self, _change_event: &ProfileChangeEvent) -> bool {
-        // loop over all proofs and verify them
-        unimplemented!()
+    // WARNING: Checks only one event, assumes all previous events are verified
+    pub fn verify(&self, change_event: &ProfileChangeEvent) -> Result<bool, ockam_core::Error> {
```
mikelodder7, line 201: I think this function can be changed to return an error vs false. It should simplify your logic.

implementations/rust/ockam/ockam/src/entity/profile.rs
```diff
@@ -31,40 +69,258 @@ impl Profile {
...
+        label: String,
+        key_type: ProfileKeyType,
+        key_purpose: ProfileKeyPurpose,
+        attributes: Option<ProfileEventAttributes>,
+    ) -> Result<(), ockam_core::Error> {
+        let event = self.rotate_key_event(label, key_type, key_purpose, attributes)?;
+        self.apply_no_verification(event)
+    }
+
+    pub fn get_private_key(
```
mikelodder7, line 159: Interesting idea to use event sourcing to retrieve the key. I like it.

implementations/rust/ockam/ockam/src/entity/profile.rs
```diff
@@ -31,40 +69,258 @@ impl Profile {
...
+
+    pub fn rotate_key(
+        &mut self,
+        label: String,
+        key_type: ProfileKeyType,
+        key_purpose: ProfileKeyPurpose,
+        attributes: Option<ProfileEventAttributes>,
+    ) -> Result<(), ockam_core::Error> {
+        let event = self.rotate_key_event(label, key_type, key_purpose, attributes)?;
+        self.apply_no_verification(event)
```
mikelodder7, line 156: Not sure why you wouldn't verify during a rotation. That seems counter-intuitive.

implementations/rust/ockam/ockam/src/entity/profile/change.rs
```diff
@@ -1,12 +1,9 @@
...
-#[derive(Clone, Debug)]
-pub enum ProfileChangeProof {}
+use crate::entity::profile::event_identifier::EventIdentifier;
+use crate::entity::profile::proof::ProfileChangeProof;
+use crate::entity::profile_change::ProfileChange;
+use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug)]
+#[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct Changes(Vec<ProfileChange>);
```
mikelodder7, line 7:
```suggestion
pub type Changes = Vec<ProfileChange>;
```

implementations/rust/ockam/ockam/src/entity/profile/profile_change.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, Eq, PartialEq)]
+pub enum ProfileKeyType {
+    Main,
+    Additional,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, Eq, PartialEq)]
+pub enum ProfileKeyPurpose {
+    Kex,
```
mikelodder7, line 22: What about Attestations?

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/create_key.rs
```diff
@@ -0,0 +1,100 @@
...
+use crate::OckamError;
+use ockam_vault_core::types::{SecretAttributes, SecretPersistence, SecretType};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct CreateKeyChange {
+    label: String,
+    key_type: ProfileKeyType,
+    key_purpose: ProfileKeyPurpose,
+    public_key: Vec<u8>,
```
mikelodder7, line 15: Do we want to combine key_type and public key into an enum that wraps the type and value together?

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/create_key.rs
```diff
@@ -0,0 +1,100 @@
...
+            public_key,
+        }
+    }
+}
+
+impl CreateKeyChange {
+    pub fn label(&self) -> &str {
+        &self.label
+    }
+    pub fn key_type(&self) -> ProfileKeyType {
```
mikelodder7, line 38: So in profile, you return references and here you return structs. We need to be consistent. I prefer without reference

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/create_key.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub fn key_purpose(&self) -> ProfileKeyPurpose {
+        self.key_purpose
+    }
+    pub fn public_key(&self) -> &[u8] {
+        &self.public_key
+    }
+}
+
+impl Profile {
+    pub(crate) fn create_key_event(
```
mikelodder7, line 50: Looks similar to create_profile. Should refactor when you can.

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/create_key.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        let prev_id = self.get_last_event_id()?;
+
+        let mut v = self.vault.lock().unwrap();
+
+        // TODO: Should be customisable
+        let secret_attributes = SecretAttributes {
+            stype: SecretType::Curve25519,
+            persistence: SecretPersistence::Persistent,
+            length: 0,
```
mikelodder7, line 75: length could be derivable from key type.

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/rotate_key.rs
```diff
@@ -0,0 +1,106 @@
...
+            length: 0,
+        };
+
+        let private_key = v.secret_generate(secret_attributes)?;
+        let public_key = v.secret_public_key_get(&private_key)?.as_ref().to_vec();
+
+        let event = RotateKeyChange::new(label, key_type, key_purpose, public_key);
+
+        let change = ProfileChange::new(
+            1,
```
mikelodder7, line 81: is this hardcoded to 1 or does it change according to how many events have already happened?

implementations/rust/ockam/ockam/src/entity/profile/profile_helpers.rs
```diff
@@ -0,0 +1,144 @@
...
+        if let Some(e) = self.change_history.as_ref().last() {
+            Ok(e.identifier().clone())
+        } else {
+            Err(OckamError::InvalidInternalState.into())
+        }
+    }
+
+    pub(crate) fn find_key_change_in_event<'a>(
+        event: &'a ProfileChangeEvent,
+        label: &str,
```
mikelodder7, line 22: Should we have the ability to search by event id?
Might be necessary because there might be an event with duplicate labels, type and purpose.

implementations/rust/ockam/ockam/src/entity/profile/profile_helpers.rs
```diff
@@ -0,0 +1,144 @@
...
+                  // }
+            })
+    }
+
+    pub(crate) fn find_last_key_event(
+        &self,
+        label: &str,
+        key_type: ProfileKeyType,
+        key_purpose: ProfileKeyPurpose,
+    ) -> Result<&ProfileChangeEvent, ockam_core::Error> {
```
mikelodder7, line 52: Again, inconsistency with references vs structs.

implementations/rust/ockam/ockam/src/entity/profile/profile_helpers.rs
```diff
@@ -0,0 +1,144 @@
...
+        self.change_history
+            .as_ref()
+            .iter()
+            .rev()
+            .find(|e| Self::find_key_change_in_event(e, label, key_type, key_purpose).is_some())
+            .ok_or(OckamError::InvalidInternalState.into())
+    }
+
+    pub(crate) fn find_key_event_before(
```
mikelodder7, line 61: I don't see any reason why we could just expose an array ref and let the caller do their own searching.

implementations/rust/ockam/ockam/src/entity/profile/profile_helpers.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub(crate) fn get_change_public_key(
+        change: &ProfileChange,
+    ) -> Result<&[u8], ockam_core::Error> {
+        match change.change_type() {
+            CreateKey(change) => Ok(change.public_key()),
+            RotateKey(change) => Ok(change.public_key()),
+        }
+    }
+
+    pub(crate) fn get_private_key_from_event(
```
mikelodder7, line 90:
```suggestion
    pub(crate) fn get_secret_key_from_event(
```

implementations/rust/ockam/ockam/src/entity/profile/proof.rs
```diff
@@ -0,0 +1,82 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, Eq, PartialEq)]
+pub enum SignatureType {
+    SelfSign,
+    Previous,
+}
+
+#[derive(Debug, Clone)]
+struct SignatureData([u8; 64]);
+
+impl Serialize for SignatureData {
```
mikelodder7, line 14: See https://crates.io/crates/serde-big-array

mikelodder7 (changes requested):
Left some comments.

implementations/rust/ockam/ockam/src/entity/profile.rs
SanjoDeundiak: This is constant that indicates that you're looking at the first event, it's the only purpose for it and I think it's fine it's the same for everybody. I could have removed it, but that will convert previous_id to Option<> which I wanted to avoid.

implementations/rust/ockam/ockam/src/entity/profile.rs
SanjoDeundiak: @mikelodder7 could you please elaborate on this?

implementations/rust/ockam/ockam/src/entity/profile.rs
SanjoDeundiak: @mikelodder7 definitely. Do you think label should also go there?

implementations/rust/ockam/ockam/src/entity/profile.rs
SanjoDeundiak: If you receive event from somewhere (e.g. network or maybe even local storage) you need to verify that. If you just created event yourself (e.g. called rotate key) you don't need to verify it

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/create_key.rs
SanjoDeundiak: @mikelodder7 yes, that's something we needed to do while working on vault. Still thinks this should be implemented inside vault crate

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/create_key.rs
SanjoDeundiak: for private keys yes, for buffers - no. that's why we decided to add length as separate field during last vault rework

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/rotate_key.rs
SanjoDeundiak: it's version number, still needs to be a named const...

implementations/rust/ockam/ockam/src/entity/profile/profile_helpers.rs
SanjoDeundiak: we haven't decided yet what is unique for specific key and what can be duplicated, that's why this was done based solely on my own vision

implementations/rust/ockam/ockam/src/entity/profile/profile_helpers.rs
SanjoDeundiak: We can, but this code is used from other places (verify function) + I think moving this code to separate function makes caller side look cleaner + there is probability we'll be reusing this function.

implementations/rust/ockam/ockam/src/entity/profile/proof.rs
SanjoDeundiak: tnx, I wish I knew that

implementations/rust/ockam/ockam/src/entity/profile.rs
mikelodder7: yes, I think it should. It's just another attribute of the key.

implementations/rust/ockam/ockam/src/entity/profile.rs
mikelodder7: Similar to the above discussion. The caller just selects a set of attributes for the profile key but we hide as much key specific detail as we can.

implementations/rust/ockam/ockam/src/entity/profile.rs
mikelodder7: Okay, let's add a domain separation parameter that we hardcode for this so it's not just the hash of empty. It's good practice to have.

implementations/rust/ockam/ockam/src/entity/profile.rs
mikelodder7: So it's an optimization. Got it.

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/create_key.rs
mikelodder7: agreed.

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/create_key.rs
mikelodder7: In truth, when do we ever use a buffer? Buffers are just like an opaque type that we store with keys. I'm not sure it makes sense to put them there. I think we should move this to the knox layer.

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/rotate_key.rs
mikelodder7: Thanks for the clarification.

implementations/rust/ockam/ockam/src/entity/profile/profile_helpers.rs
mikelodder7: Got it. Thoughts on just letting the caller search the events themselves?

implementations/rust/ockam/ockam/src/entity/profile.rs
SanjoDeundiak: agree

implementations/rust/ockam/ockam/src/entity/profile.rs
SanjoDeundiak: agree

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/create_key.rs
SanjoDeundiak: chain key is a good example of secret buffer

implementations/rust/ockam/ockam/src/entity/profile/profile_helpers.rs
SanjoDeundiak: I'm not sure it's helpful, we can add it anytime

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/create_key.rs
SanjoDeundiak: @mikelodder7 it depends on type. Small types like simple enums and integers should be returned by value instead of const ref.

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/create_key.rs
SanjoDeundiak: Agree

implementations/rust/ockam/ockam/src/entity/profile/profile_helpers.rs
SanjoDeundiak: @mikelodder7 do we have a place with decisions about such naming? If not, we should create one

implementations/rust/ockam/ockam/src/entity/profile/profile_change_type/create_key.rs
SanjoDeundiak: @mikelodder7 how about throwing kid inside that struct as well?

mikelodder7 (approved):

PR #857, opened 2021-02-10
fix(rust): clippy issues in ockam crate

jared-s:
Fixes some Clippy warnings. Requires a few code changes which I'll highlight inline. For the most part, I applied the suggestions from Clippy ouput, using CLion.

### DIFF
--- a/implementations/rust/ockam/ockam/src/profile.rs
+++ b/implementations/rust/ockam/ockam/src/profile.rs
@@ -105,13 +105,11 @@ impl Profile {
         change_events: Vec<ProfileChangeEvent>,
         vault: Arc<Mutex<dyn ProfileVault>>,
     ) -> Self {
-        let profile = Self {
+        Profile {
             identifier,
             change_history: ProfileChangeHistory::new(change_events),
             vault,
-        };
-
-        profile
+        }
     }
 }
 
@@ -139,7 +137,7 @@ impl Profile {
         )?;
 
         let change = ProfileChangeHistory::find_key_change_in_event(&change_event, &key_attributes)
-            .ok_or_else(|| OckamError::InvalidInternalState)?;
+            .ok_or(OckamError::InvalidInternalState)?;
         let public_key = ProfileChangeHistory::get_change_public_key(&change)?;
 
         let public_kid = v.compute_key_id_for_public_key(&public_key)?;
--- a/implementations/rust/ockam/ockam/src/profile.rs
+++ b/implementations/rust/ockam/ockam/src/profile.rs
@@ -23,7 +23,7 @@ pub use profile_change_type::*;
 pub mod profile_change_history;
 
 pub const OCKAM_NO_EVENT: &[u8] = "OCKAM_NO_EVENT".as_bytes();
-pub const PROFILE_ROOT_KEY_LABEL: &'static str = "OCKAM_PRK";
+pub const PROFILE_ROOT_KEY_LABEL: &str = "OCKAM_PRK";
 pub const OCKAM_PROFILE_VERSION: u8 = 1;
 pub const PROFILE_CHANGE_CURRENT_VERSION: u8 = 1;
 
--- a/implementations/rust/ockam/ockam/src/profile/profile_change_history.rs
+++ b/implementations/rust/ockam/ockam/src/profile/profile_change_history.rs
@@ -66,7 +66,7 @@ impl ProfileChangeHistory {
             .iter()
             .rev()
             .find(|e| Self::find_key_change_in_event(e, key_attributes).is_some())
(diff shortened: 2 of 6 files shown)


implementations/rust/ockam/ockam/src/profile.rs
```diff
@@ -274,11 +274,11 @@ impl Profile {
                         serde_bare::to_vec(c.data()).map_err(|_| OckamError::BareError)?;
                     let data_hash = vault.sha256(data_binary.as_slice())?;
 
-                    if !vault
+                    if vault
                         .verify(c.self_signature(), c.data().public_key(), &data_hash)
-                        .is_ok()
+                        .is_err()
                     {
-                        false;
```
jared-s, line 281: This one requires some eyeballs @SanjoDeundiak @mikelodder7. Previously it was falling through after executing the `false;` empty block. I assume if that validation fails we want to abort? Which is this change

implementations/rust/ockam/ockam/src/profile/profile_change_history.rs
```diff
@@ -89,7 +89,11 @@ impl ProfileChangeHistory {
             RotateKey(change) => change.data().public_key(),
         };
 
-        Ok(PublicKey::new(data.into()))
```
jared-s, line 92: Clippy complained that this function always succeeded, thus did not need to be wrapped in a Result. But all of our APIs should return Result, and we will probably find ways to make it fail in the future :) So I started with a trivial failure mode.

implementations/rust/ockam/ockam/src/profile.rs
```diff
@@ -145,7 +145,7 @@ impl Profile {
         )?;
 
         let change = ProfileChangeHistory::find_key_change_in_event(&change_event, &key_attributes)
-            .ok_or_else(|| OckamError::InvalidInternalState)?;
+            .ok_or(OckamError::InvalidInternalState)?;
```
SanjoDeundiak, line 148: This was clippy warning before

implementations/rust/ockam/ockam/src/profile.rs
SanjoDeundiak: This is very nice catch. I'll fix it in my Contact PR (because I moved this code there)

implementations/rust/ockam/ockam/src/profile.rs
```diff
@@ -388,7 +388,6 @@ mod test {
                 println!("{} is valid", id);
             } else {
                 println!("{} is not valid", id);
-                assert!(false);
```
SanjoDeundiak, line 391: I removed this code in my Contacts PR

implementations/rust/ockam/ockam/src/profile/profile_change_history.rs
SanjoDeundiak: Nice

implementations/rust/ockam/ockam/src/profile.rs
```diff
@@ -24,7 +24,7 @@ pub use profile_change_type::*;
 mod profile_change_history;
 
 pub const OCKAM_NO_EVENT: &[u8] = "OCKAM_NO_EVENT".as_bytes();
-pub const PROFILE_ROOT_KEY_LABEL: &'static str = "OCKAM_PRK";
+pub const PROFILE_ROOT_KEY_LABEL: &str = "OCKAM_PRK";
```
SanjoDeundiak, line 27: Nice

SanjoDeundiak (changes requested):
Let's discuss things here, and then I'll fix all the issues myself in my latest PR, because large part of the code been fixed here was moved, and merging this and my PR together will bring a pile of conflicts

implementations/rust/ockam/ockam_vault_core/src/types.rs
```diff
@@ -26,7 +26,7 @@ cfg_if! {
         pub type SmallBuffer<T> = heapless::Vec<T, U4>;
         /// Buffer for large binaries (e.g. encrypted data). Max size - 512
         pub type Buffer<T> = heapless::Vec<T, U512>;
-        pub type KeyId = heapless::String;
+        pub type KeyId = heapless::String<U32>;
```
jared-s, line 29: @mikelodder7 is `U32` the correct heapless generic here?

SanjoDeundiak: @jared-s It's 32 bytes hex-encoded, so U64 should be good

jared-s: ooops thanks Sanjo, meant to tag you

SanjoDeundiak:
Probably can be closed @jared-s

PR #863, opened 2021-02-11
feat(rust): add facade for various std no-std modes

mikelodder7:
This allows all other crates to no longer have to worry about the various imports for the three different std no-std modes.

Crate consumers just need to add

`use ockam_core::lib::*;`

which figures out the correct structs and traits to import.


jared-s:
Merged as https://github.com/ockam-network/ockam/pull/867

PR #872, opened 2021-02-12
feat(rust): migrate to use ockam_core facade for structs

mikelodder7:
This changes all instances where Vec and String are used to call upon the recently added facade in ockam_core.


mrinalwadhwa:
merged f1f00a43486a78c7ce55a170cfecea56cb7d0ca7

PR #893, opened 2021-02-16
Working on resilient message delivery.

hairyhum:
### Rationale

In order to provide reliable "at least once" delivery there should be a confirmation mechanism for consumers to inform the topic that the message was processed.
This PR concentrates on implementing the protocol for that and splitting out the storage API from consumer management.

### Proposed Changes

Change the API for message consumption and protocol requirements for consumers.

- Add an index to each message.
- Add a `confirm(topic_name, pid, index)` api to confirm received messages
- Allow consumers to set a limit of messages to receive before confirming (effectively messages "in flight")
- Remove the `process_queued_messages` api as it doesn't make much sense.

Change implementation of the topic storage:

- Store all messages to storage regardless of consumers existence
- Add consumer state storage to store current consumers indexes (last sent, last confirmed)

TODO:

- Monitor consumer PID
- Consumer ID is still a pid. If we need to carry over consumer ID to the pid restart it needs to be decoupled
- bump_consumers might need a rework for distributed storage. 

The `Ockam.Topics.Stram` modules are more of a demonstration of the alternative approach which might be taken here: consumers in that case are using `pull` model. Doesn't need to go with this PR, but there for demonstration reasons.

### DIFF
--- a/implementations/elixir/ockam/_ockam/lib/ockam/stream.ex
+++ b/implementations/elixir/ockam/_ockam/lib/ockam/stream.ex
@@ -1,0 +1,328 @@
+defmodule Ockam.Topics.Stream.Storage do
+  @type message() :: %{message: term(), index: integer()}
+  @type topic() :: atom()
+
+  @callback create_topic(topic()) :: {:ok, atom()} | {:error, term()}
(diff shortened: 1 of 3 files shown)


hairyhum:
Closed in favour of #1203

PR #894, opened 2021-02-16
docs(rust): update docs for ockam_ffi

SanjoDeundiak:


jared-s:
Merged as https://github.com/ockam-network/ockam/pull/913

PR #912, opened 2021-02-17
Ockam worker relay mailbox

spacekookie:
This PR implements a few new concepts, while also refactoring the API.

* Move API endpoints to `Context`
  * Remove `Node` concept entirely
* Implement a worker mailbox to buffer messages
* Allow workers to block for incoming messages
  * Messages that don't belong into that 'state' provide `.cancel()` and will be re-queued
* Move ockam worker core (previously node) to be a minimal ockam worker itself

### DIFF
--- a/implementations/rust/examples/node/examples/create_node_with_macro.rs
+++ b/implementations/rust/examples/node/examples/create_node_with_macro.rs
@@ -1,4 +1,4 @@
 #[ockam::node]
 async fn main(context: ockam::Context) {
-    context.node().stop().unwrap();
+    context.stop().unwrap();
 }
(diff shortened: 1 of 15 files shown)


spacekookie:
Oh, the `ockam_node_attribute` crate was still using the old API. Hopefully this fixes it

mrinalwadhwa:
the trybuild tests that try to use the node attribute are failing, will need an update to the new api
https://github.com/ockam-network/ockam/tree/develop/implementations/rust/ockam/ockam/tests/node_attribute

mrinalwadhwa:
The API is starting to look really good 🎉 
The code in this PR is merged. I had to update a couple of tests.

PR #920, opened 2021-02-18
ockam worker ergonomics changes

spacekookie:


jared-s:
Merged as https://github.com/ockam-network/ockam/pull/921

PR #922, opened 2021-02-18
Elixir Ockam TCP Transport Initial Implementation

mjc:
This adds the beginnings of a TCP transport to Ockam's Elixir codebase.
Additionally this switches to BARE wire encoding.

Currently every outgoing message opens a new TCP connection.
A follow-up must attach outgoing connections to `:ranch` and also implement `Ockam.Worker` on the ranch handler, so that return messages go to the correct location.


implementations/elixir/ockam/ockam/mix.exs
```diff
@@ -72,6 +74,7 @@ defmodule Ockam.MixProject do
 
   defp aliases do
     [
+      credo: "credo --strict",
```
mrinalwadhwa, line 77: there's another alias for this below `lint.credo`

mjc: Sure, I noticed that. However, I was running `mix credo` automatically in VSCode and did not look at the alias list.  It seems like switching the default to what we expect would keep others from falling into the same trap.  I'd argue that we should do that on all of our elixir packages, too - but didn't think about it til just now.

mrinalwadhwa: cool, makes sense 👍

mrinalwadhwa:
974551305c918836b60ce68de46ec607ce9410e8 - could you change this commit message to describe the changes in the commit.

mrinalwadhwa:
Awesome, looks ready to merge, one last thing ... the commit message got left mid sentence `test(elixir): add encoding/decoding tests for`

PR #925, opened 2021-02-18
feat(rust): update profile folder/module structure

SanjoDeundiak:


jared-s:
Merged as https://github.com/ockam-network/ockam/pull/931

PR #926, opened 2021-02-18
feat(rust): move credentials to main ockam crate

mikelodder7:


jared-s:
Merged as https://github.com/ockam-network/ockam/pull/929

PR #965, opened 2021-02-22
Moving xx and core to releasable state

SanjoDeundiak:


mikelodder7 (approved):

implementations/rust/ockam/ockam_key_exchange_core/src/lib.rs
```diff
@@ -32,14 +32,50 @@ pub trait NewKeyExchanger<E: KeyExchanger = Self, F: KeyExchanger = Self> {
 /// A Completed Key Exchange elements
 #[derive(Debug, Zeroize)]
 pub struct CompletedKeyExchange {
+    h: [u8; 32],
```
SanjoDeundiak, line 35: Making this fields private and adding getters may break some code @robinbudd

PR #968, opened 2021-02-22
Spacekookie/worker supervisors

spacekookie:
This PR depends on #967 and contains a basic worker-supervisor relationship mechanism

### DIFF
--- a/implementations/rust/examples/worker/examples/worker_supervisor.rs
+++ b/implementations/rust/examples/worker/examples/worker_supervisor.rs
@@ -1,0 +1,38 @@
+use ockam::{Context, Error, Result, Worker as WorkerTrait};
+
+struct Supervisor;
+
+impl WorkerTrait for Supervisor {
+    type Context = Context;
+    type Message = ();
+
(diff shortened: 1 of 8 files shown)


jared-s:
This PR touches a lot of error handling code in FFI as well as some changes in behavior for the relay service.

Two questions:

1. Are the error strategies introduce in this PR compatible with the latest error paradigms we have switched to in the past weeks?
2. Are these relay changes compatible with the current Hub implementations, and is there any conflict with the upcoming Stream implementation?

Are these necessary parts for the supervision system, or can we split these changes out?

spacekookie:
I'm not sure I'm aware of how the error handling paradigm has changed in the last few weeks. Fundamentally these changes are still applicable to the Relay, although I'll have to have a look at how to refactor them to match the new code structure.

I think the most important part is that this approach assumes that errors can be sent across the network, thus being able to report _any_ error suffered by any worker to their supervisor.

jared-s (approved):

jared-s:
@spacekookie makes sense. I could use your help with the merge conflicts in `context.rs`, quite a bit has changed in there. The others look pretty trivial. Thanks!

PR #969, opened 2021-02-22
feat(rust): add example of issue, prove, and verify a credential

mikelodder7:


jared-s:
Notes from call @mikelodder7 @mrinalwadhwa

- Use case should be tied to the idea of Attribute Based Access Control (ABAC)
- Rename Holder, Issuer, Verifier to have Credential prefixes: CredentialHolder etc
- Difference between Manifest and Schema? Rename PresentationManifest?
- Consider different GUID format instead of random hex: Meaningful names? Routing hierarchy?
- Example should use two attributes: one blind/hidden and one not
- Consider alternative words to `blinding` etc
- Truck Driver ID should be the hidden attribute.
- The capability / ability to drive flag should be revealed from the driver to the truck.
- Is "truck maintenance" a better scenario than "truck driving"?
- Break example down into three pieces: Issuer, Holder, Verifier binaries.
- They can share code such as `CredentialSchema`
- issuer: Acme Corp, holder / subject, Bob the Maintenance Guy, Verifier = Truck
- Something needs to happen before the Issuer offers a Credential: Holder needs to initiate the process.
- Use @SanjoDeundiak profile verification before Issuer offers credential
- Good point to tie in Credentials with Profile verification
- Attribute Schema needs to be updated to contain a bit indicating whether an attribute can be blinded / hidden
- Fragments instead of blinding?

mikelodder7:
So if I did all of that in a single PR it would be massive. Instead, that should be done with incremental PRs.
I say we merge this one and I'll apply the suggested changes iteratively.

mrinalwadhwa:
@mikelodder7 sounds good, let's merge this once the linter errors are resolved.

jared-s:
Rebasing and merging here https://github.com/ockam-network/ockam/pull/974

jared-s:
Merged https://github.com/ockam-network/ockam/pull/974

PR #971, opened 2021-02-23
feat(rust): add profile authentication

SanjoDeundiak:


implementations/rust/ockam/ockam/src/attestation.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+pub(crate) struct Attestation {}
+
+impl Attestation {
+    pub(crate) fn generate_attestation_request(
+        requester_profile_id: ProfileIdentifier,
+        _vault: &mut dyn ProfileVault,
+    ) -> ockam_core::Result<Vec<u8>> {
+        // FIXME: generate random nonce using vault
```
SanjoDeundiak, line 61: Should we add random trait to vault? Or should we just use whatever we have on this platform

SanjoDeundiak: @mikelodder7

mikelodder7: I think we should have the system specify the RNG. IoT usually have to seed this on boot.

SanjoDeundiak: @mikelodder7 OsRng? thread_rng()?

SanjoDeundiak:
The only one thing missing here is docs on functions. Will add that in another PR

implementations/rust/ockam/ockam/src/credential/holder.rs
```diff
@@ -107,7 +107,7 @@ impl CredentialHolder {
         proof_request_id: [u8; 32],
     ) -> Result<Vec<CredentialPresentation>, CredentialError> {
         // To prove the id-secret is the same across credentials we use a Schnorr proof
-        // which requires that the proof blinding factor be the same. If there's only one credential
+        // which requires that the proof blinding proof be the same. If there's only one credential
```
mikelodder7, line 110: This comment change is not correct. It was correct before. Is this accidental?

implementations/rust/ockam/ockam/src/profile.rs
```diff
@@ -1,22 +1,28 @@
-use crate::{Contact, OckamError};
+use crate::OckamError;
 use hashbrown::HashMap;
```
mikelodder7, line 2: switch to using ockam_core::lib::HashMap. That will use the correct one depending on no_std, alloc, or std

implementations/rust/ockam/ockam/src/profile.rs
```diff
@@ -1,22 +1,28 @@
...
 pub use change::*;
+use history::ProfileChangeHistory;
 
 pub trait ProfileVault: SecretVault + KeyIdVault + Hasher + Signer + Verifier {}
 
 impl<D> ProfileVault for D where D: SecretVault + KeyIdVault + Hasher + Signer + Verifier {}
 
 pub type ProfileEventAttributes = HashMap<String, String>;
+/// Contacts Database
+pub type ContactsDb = HashMap<ProfileIdentifier, Contact>;
```
mikelodder7, line 25: Should we consider a struct instead of type if we want to handle persistence.?
Doesn't need to be addressed in this PR, just food for thought.

mikelodder7 (commented):
Good code. Just address the one minor issue.

implementations/rust/ockam/ockam/src/credential/holder.rs
SanjoDeundiak: I was doing find&replace. Good catch. Will fix

implementations/rust/ockam/ockam/src/profile.rs
SanjoDeundiak: I'm sure persistence is something we'll be working on in the future. For now I hope that's fine

implementations/rust/ockam/ockam/src/profile.rs
SanjoDeundiak: Fixed

implementations/rust/ockam/ockam/src/credential/holder.rs
SanjoDeundiak: Fixed

mikelodder7 (approved):

PR #972, opened 2021-02-23
feat(rust): prefix terms with credential

mikelodder7:


jared-s:
Rebasing here https://github.com/ockam-network/ockam/pull/977

Issue #962, opened 2021-02-21
Hub - Forwarding Service


mrinalwadhwa:

mrinalwadhwa:
Required for #952

mrinalwadhwa:
Forwarding Service is merged into develop and works with the UDP transport. Here's a test:

Start `iex` in the hub app:
```bash
cd implementations/elixir/ockam/ockam_hub
iex -S mix
```

Start UDP transport:
```elixir
{:ok, address} = Ockam.Transport.UDP.create_listener [port: 5000, route_outgoing: true]
```

Start `iex` in the ockam app:
```bash
cd implementations/elixir/ockam/ockam
iex -S mix
```

```
mix run --no-halt b.exs
```

```elixir
# b.exs

alias Ockam.Transport.UDPAddress
{:ok, address} = Ockam.Transport.UDP.create_listener [port: 6000, route_outgoing: true]

defmodule EchoService do
  use Ockam.Worker

  alias Ockam.Message

  @impl true
  def setup(_options, state) do
    state = Map.put(state, :registered, false)
    {:ok, state}
  end

  @impl true
  def handle_message(message, %{ registered: false } = state) do
    forwading_address = Message.payload(message)
    IO.puts("FORWADING ADDRESS: #{inspect({forwading_address, state})}")
    state = Map.put(state, :registered, true)
    {:ok, state}
  end

  @impl true
  def handle_message(message, state) do
    reply = %{
      onward_route: Message.return_route(message),
      return_route: [state.address],
      payload: Message.payload(message)
    }

    IO.puts("MESSAGE: #{inspect(message)}\nREPLY: #{inspect(reply)}")
    Router.route(reply)

    {:ok, state}
  end
end

EchoService.create(address: "echo_service")

route_to_node_running_forwading_service = [%UDPAddress{ip: {127, 0, 0, 1}, port: 5000}]
forwarding_service_register_request = %{
  onward_route: route_to_node_running_forwading_service ++ ["forwarding_service"],
  return_route: ["echo_service"],
  payload: "register"
}

Ockam.Router.route(forwarding_service_register_request)
```

This will print a forwarding address, copy it and pass it as first argument fo `a.exs`
```
mix run --no-halt a.exs 30a9f31d
```

```elixir
# a.exs

[forwading_address_of_echo_service_on_b | _] = System.argv()

alias Ockam.Transport.UDPAddress
[...]

PR #980, opened 2021-02-23
forwarder

mrinalwadhwa:
- fix(elixir): failing node send if the address in unknown
- feat(elxir): use message protocol in echo service
- feat(elixir): add forwarding service to hub


mrinalwadhwa:
Addresses #962

PR #991, opened 2021-02-26
feat(rust): support multiple addresses per worker

spacekookie:
### Proposed Changes

* Each worker has a primary address (the first one)
* Only primary addresses can shut down the worker
* Message handle provides address of the message


jared-s:
Merged as https://github.com/ockam-network/ockam/pull/1019

PR #992, opened 2021-02-26
chore(rust): change worker initialize to be async

spacekookie:
Change the worker initialisation to use an async function.


jared-s:
Rebasing now.

jared-s:
Merged as https://github.com/ockam-network/ockam/pull/993

PR #997, opened 2021-03-01
build: cla bot checks does user sign agreement

sobi3ch:


.github/workflows/cla.yml
```diff
@@ -0,0 +1,74 @@
...
+jobs:
+  check-submitter-for-cla:
+    name: Check submitter for CLA
+    env:
+      MESSAGE: "Thank you <USER> for creating this <EVENT>. Before you continue you will need to sign our CLA.
+To do it please create a new PR to the https://github.com/ockam-network/contributors repo adding yourself to the CSV file."
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone CONTRIBUTORS
+        uses: actions/checkout@v2.3.4
```
mrinalwadhwa, line 16: As recommended by Action security guidelines we want to pin actions to exact version hashes.
For example https://github.com/ockam-network/ockam/blob/develop/.github/workflows/ci.yml#L13

metaclips: resolved in https://github.com/ockam-network/ockam/pull/997/commits/cf8977be977863ee2fa366a151f95c570fbb8c75 thanks

PR #1000, opened 2021-03-01
feat(rust): establish ockam_examples crate

jared-s:
# Ockam Examples

The `ockam_examples` crates contains the source code and Markdown files for Ockam examples and guides.


implementations/rust/ockam/ockam_examples/examples/node_ping_pong.rs
```diff
@@ -0,0 +1,86 @@
+//! Spawn to workers that play some ping-pong
```
vlisivka, line 1: Typo: to -> two

jared-s:
For now, this duplicates some examples that are in the `rust/examples` directory. That directory will be removed eventually, however some folks are still actively working in there. I will be syncing examples to this crate through the week, and then removing the directory.

PR #1001, opened 2021-03-02
feat(rust): additional cli args for issuer

mikelodder7:


implementations/rust/examples/credentials/src/issuer.rs
```diff
@@ -2,32 +2,57 @@ mod util;
...
+    };
     let pk = issuer.get_public_key();
     let pop = issuer.create_proof_of_possession();
     let schema = example_schema();
     let mut pending_offers = BTreeSet::new();
 
     //TODO: publish public key and proof of possession for verifiers
-
-    let listener = UnixListener::bind("/tmp/issuer.socket").unwrap();
+    let listener = Listener::bind(args.socket, args.port).unwrap();
```
vlisivka, line 48: It better to use .expect("message") than .unwrap(), because it much easier to reason about error with explanation.

mikelodder7: The Result Err is a String. It prints either way.

vlisivka: unwrap() will print error message, but it will not explain reason. Usually, in this case I use something like 

.expect("Cannot bind socket to address. Maybe another daemon is already running at this ip:port? Try to close daemon, or use another ip:port combination, e.g. 127.0.0.2:9090.");

It serves both as the comment in code and as debugging message for a system administrator.

mikelodder7: The example code underneath is already doing that. Hence the unwrap

jared-s:
Merged https://github.com/ockam-network/ockam/pull/1025

PR #1036, opened 2021-03-05
Modular router system

spacekookie:
This PR provides the basics for a more modular router system.

Changes so far:

- Implement address type in `Address`
  - Provide convenience methods to create Addresses from Strings
  - The functionality of `AddressSet` is not impacted by this change 
- Implement Route and RouteBuilder abstractions over the ockam_core addresses

## Update (2021-03-09)

- Improve the route builder to allow in-place editing of route types via the same builder API
- Add a mechanism by which routers can register themselves via the `Context` API
- Wrap messages in ockam_node so that router implementations are forced to handle `RouterMessage` types
  - Log and handle errors to comply with this requirement gracefully
- Add a `Context` API endpoint to "forward" a message from a router to a "child worker"
- There is a minimal example of this system in use called `worker_simple_router`
  - `Consumer` could potentially be a `TcpConnectionWorker`
  - The address space prefix (`proxy_me_...`) was chosen to make the difference between the two address spaces clear, but fundamentally the mapped address space could be anything.

### Context API additions

**Why add `forward_message(TransportMessage)`?**

By letting the router re-enter its re-addressed message through the same routing stack, a lot of logic can be re-used, instead of pushing this responsibility onto the external router author.
Currently this only works if the router forwards the message to a non-router worker.

**Why add a `register(...)` API instead of using the existing message sending system?**

The internal router is _not_ implemented as a worker, because this would cause a bootstrap problem.
I initially tried to implement it this way, but ran into a lot of problems and complexities that can be solved by using the command channel message types used by the `Context` API to request per-worker senders, or to shut down a node.

### What is missing?

Currently a user-worker has no access to the route information of the message that it received.

Routers must accept `RouterMessage`, which is a slim wrapper around a `TransportMessage`, which contains all of this state. To give userspace workers access to the same information I envision a conposition type similar to `Cancel<_, M>`, which implements `Deref<M>`, and provides an API to access route information to make it easy for users to send replies to workers based entirely on the message route.

### Data flow chart

Following is a breakdown of the data flow through this API.

Routing


spacekookie:
## Update 2021-03-15

There has been a lot of movement on this PR since I started it. I rebased it against develop this morning to make sure we can merge it.

Following is a breakdown of changes since the last update.  These are primarily around the TCP transport crate.

### Changes to `ockam_node`

* Introduce `Routed<M>` which provides access to the route information of a message
  * When wanting to move the message, call `.take()` to drop the composite type
  * Also provides `.sender()` and `.reply()` to quickly construct a reply route

### Changes to `ockam_transport_tcp`

* Refactor `tcp_transport` internals to use the new worker and routing system
  * A TCP connection worker is implemented as two workers: one to receive, and one to send
  * Uses tokio's `into_split()` mechanism to separate ownership of input and output streams
  * The receiving worker is implemented as a custom loop in `initialize()`
    * We don't currently have a mechanism to have access to a worker `Context`, without receiving messages
    * The receive worker uses an `AtomicBool` to kill its receive loop when stopping the worker pair 
    * There may be better designs available to solve this issue, but they will require some research
* Retire the `ockam_router` crate, as it is now fully implemented in `ockam_node`
* Rewrite `ockam/examples/tcp` crate with the new architecture
* The new `TcpRouter` can either act as a client router, or bind a TcpListener to a socket to accept incoming connections
  * Provide a function `register(...)` on `RouterHandle` to allow connection workers to register themselves with the router
  * When calling `TcpRouter::bind()` another worker is started: `TcpIncomingWorker`, which is shut down when the router is

Following is an overview of the **TCP specific parts** of the routing architecture.

tcp-routing

### What's missing

Most of the design is complete and only lacking in some implementation specifics. However, there are some small design questions to be resolved.

**Connection workers won't shut down immediately**

In the current design the `TcpRecvWorker` and `TcpListenWorker` are implemented as custom loops in the `initialize()` function. The run loop is gated via an `Arc<AtomicBool>` which is switched off when the worker is meant to be shut down.

This means that a worker might idle for longer than it is meant to, because it is blocking on a tokio i/o recv call! This concept needs to be changed to expose a Future API, which then allows the worker to `select(...)` between the two futures.

**Examples**

Getting the previous examples to work with the new code base would have been much harder than starting from scratch, which means that currently there are no examples of this API working

spacekookie:
## Update

This PR has examples now and should be ready to merge @jared-s (if I get past CI :P)

spacekookie:
Actually, the rebase seems to have added a bunch of files that I previously deleted so I have to have another look at cleaning up the history before merging!

spacekookie:
Okay, looks good now. Not sure why the tests fail, but that doesn't feel like something that this PR touches

jared-s:
Awesome thanks @spacekookie , I'll investigate the failures and take it from here!

spacekookie:
@jared-s Oh, I fixed it. I'm pushing one last thing to also enable some lints I had forgotten about.

jared-s (approved):

jared-s:
Merged as https://github.com/ockam-network/ockam/pull/1070 !

PR #1063, opened 2021-03-15
chore(elixir): bump test duration to see if ci is really that slow

mjc:
Bump the timeout for assert_receive() again to see if it solves an intermittent CI failure.  Why is talking to localhost so slow on CI??


mjc:
this wasn't the problem

PR #1074, opened 2021-03-17
fix[elixir]: reuse tcp connection when sending messages back

hairyhum:
### Proposed Changes

FIX: add server ip/port to return route on TCP transport listener
FIX: route messages received from the connection on TCP transport client

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/client.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/client.ex
@@ -2,6 +2,8 @@ defmodule Ockam.Transport.TCP.Client do
   @moduledoc false
   use GenServer
 
+  @wire_encoder_decoder Ockam.Wire.Binary.V2
+
   @impl true
   def init(%{ip: ip, port: port} = state) do
     # TODO: connect/3 and controlling_process/2 should be in a callback.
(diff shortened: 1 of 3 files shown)


hairyhum:
There is an idea to use client connections as registered workers as well, but it would require more work to make client connections support the worker API, hence out of scope for this one.

hairyhum:
Tested with ockam_hub with elixir client
Tested with ockam_hub with rust examples with a changed echo service name as following:
```
diff --git a/implementations/rust/examples/tcp/examples/network_echo_client.rs b/implementations/rust/examples/tcp/examples/network_echo_client.rs
index f15313c5..0effe945 100644
--- a/implementations/rust/examples/tcp/examples/network_echo_client.rs
+++ b/implementations/rust/examples/tcp/examples/network_echo_client.rs
@@ -36,7 +36,7 @@ async fn main(mut ctx: Context) -> Result<()> {
     ctx.send_message(
         Route::new()
             .append(format!("1#{}", peer_addr))
-            .append("simple.responder"),
+            .append("echo_service"),
         String::from("Hello you over there!"),
     )
     .await?;
```

hairyhum:
@mjc ready to review/merge (the tests are flaky in CI though)

@spacekookie maybe we should use `echo_service` name in the Rust examples to have them work identical to the hub?

mjc (approved):

spacekookie:
> @spacekookie maybe we should use `echo_service` name in the Rust examples to have them work identical to the hub?

I had that thought last night as well. I'll make the change today along with the actual example guide

mrinalwadhwa (approved):

PR #1075, opened 2021-03-17
feat[elixir]: register tcp client as worker to route back to it without reconnecting

hairyhum:
### Proposed Changes
Move code for sending messages from tcp listener to tcp client
Establish connections by creating new clients as workers
Make clients add its own identity instead of ip/port to return route

This allows the ping-pong example when two workers on different machines can talk back and forth using the same connection.

Depends on #1074

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/client.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/client.ex
@@ -2,6 +2,8 @@ defmodule Ockam.Transport.TCP.Client do
   @moduledoc false
   use GenServer
 
+  @wire_encoder_decoder Ockam.Wire.Binary.V2
+
   @impl true
   def init(%{ip: ip, port: port} = state) do
     # TODO: connect/3 and controlling_process/2 should be in a callback.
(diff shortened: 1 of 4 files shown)


implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/client.ex
```diff
@@ -1,44 +1,82 @@
...
+  end
+
+  defp send_over_tcp(data, %{socket: socket}) do
+    :gen_tcp.send(socket, data)
   end
 
-  @spec send(atom | pid | {atom, any} | {:via, atom, any}, any) :: any
-  def send(pid, data) do
-    GenServer.call(pid, {:send, data})
+  defp update_return_route(message, %{address: address}) do
```
mjc, line 78: I like this name much better than `set_`

implementations/elixir/ockam/ockam/test/ockam/router_test.exs
```diff
@@ -9,19 +9,129 @@ defmodule Ockam.Router.Tests.Printer do
...
+
+  def next(n) do
+    int_n = n |> String.trim() |> String.to_integer()
+    to_string(int_n + 1)
+  end
+end
+
 defmodule Ockam.Router.Tests do
-  use ExUnit.Case, async: true
+  use ExUnit.Case, async: false
```
mjc, line 112: good thinking, I forgot about this flag entirely.

mjc (approved):

mrinalwadhwa (approved):

PR #1082, opened 2021-03-22
chore(elixir): improve Wire.Decoder error handling

feniks65:
### Proposed Changes

Make Wire.decode errors raisable -   #[888](https://github.com/ockam-network/ockam/issues/888)

@mjc @mrinalwadhwa
Have you thought of something different like inspecting binary data in Ockam.Wire.Binary.V2.decode? 

@mrinalwadhwa you pointed out Inspect.Opts module in this case. Did you think of something like piping the binary data through 
`|> Inspect.Algebra.to_doc(%Inspect.Opts{})`
with specific options in Inspect.Opts ?

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/wire.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/wire.ex
@@ -71,19 +71,27 @@ defmodule Ockam.Wire do
   @spec decode(decoder :: atom, encoded :: binary) ::
           {:ok, message :: Message.t()} | {:error, error :: DecodeError.t()}
 
+  def decode!(decoder, encoded) do
+    case decode(decoder, encoded) do
+      {:ok, message} -> {:ok, message}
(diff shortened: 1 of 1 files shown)


mrinalwadhwa:
@feniks65 thank you for sending this PR!

I think it may be better if we follow the standard library's convention of raising from a function with a `!` .. so `decode` would return `{:ok, message} | {:error, error}` while `decode!` would delegate to `decode` and raise if the return value is not `{:ok, message}`

What do you think?

feniks65:
How about now?
I have introduced an additional function `decode!` as you have suggested which should distinguish the result of the call to the basic `decode()`.
If an error will be returned from decode() then an exception will be raised.

mrinalwadhwa:
Hi @feniks65 sorry for losing track of this PR.
The changes look great! 
I rebased and merged 652ee29217b5994f44501be71bad8fcbcab964a1

Thank you 🙏

PR #1105, opened 2021-03-24
fix(elixir): use return route instead of payload to return forwarding address

hairyhum:
### Proposed Changes

Instead of using payload to communicate the forwadring mailbox address, use the return route.

Payload encoding might be different in different implementations
Return route sounds like a reasonable way to communicate changes in the
routing


mjc (approved):

jared-s:
Merged as https://github.com/ockam-network/ockam/pull/1106

PR #1108, opened 2021-03-25
Forwarder example

spacekookie:
Currently the example uses a patch from @hairyhum to make the `forwarder_service` use the return route of the message its sending back to communicate the forwarding address.

This is required because of incompatible payload encodings between the Rust code and Elixir code.


jared-s:
Merged as https://github.com/ockam-network/ockam/pull/1109

PR #1110, opened 2021-03-25
feat(rust): helper trait for peer address formatting

jared-s:
What do you all think of this idea?

This is a small helper trait to hide the repeated formatting of address types in peer addresses.

The `PeerAddressFormatter` could be moved to `ockam_core` and then each distinct transport can implement it.

Also added a local const for Address Type 1 (TCP)

### DIFF
--- a/implementations/rust/ockam/ockam_transport_tcp/src/router.rs
+++ b/implementations/rust/ockam/ockam_transport_tcp/src/router.rs
@@ -6,6 +6,7 @@ use crate::{
 use ockam::{async_worker, Address, Context, Result, Routed, RouterMessage, Worker};
 use std::{collections::BTreeMap, net::SocketAddr};
 
+const ADDRESS_TYPE_TCP: u8 = 1;
 const DEFAULT_ADDRESS: &'static str = "io.ockam.router.tcp";
 
 /// A TCP address router and connection listener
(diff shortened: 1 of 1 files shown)


implementations/rust/ockam/ockam_transport_tcp/src/router.rs
```diff
@@ -136,3 +136,21 @@ impl TcpRouter {
...
+pub trait PeerAddressFormatter {
+    fn peer_addr<S: ToString>(_: S) -> Address;
+}
+
+/// Address format for TCP peers
+impl PeerAddressFormatter for TcpRouter {
+    fn peer_addr<S: ToString>(s: S) -> Address {
+        format!("{}#{}", ADDRESS_TYPE_TCP, s.to_string()).into()
+    }
+}
```
spacekookie, line 150: I like the idea of having a common interface to construct addresses of a certain type, but I'm not so sure about associating the function with the `TcpRouter` (or actually any particular type).

It might not be immediately obvious to a user why they should expect this function on the Router, vs (for example) the connection worker handle.

I think we might be able to get away with just having a convention that transport crates expose a function in their root called `address` (and enforce this with some tests maybe), so that users of the crate can simply call `tcp::address("127.0.0.1:1234")` (provided they `use ockam_transport_tcp as tcp` first).

jared-s: That sounds like a good idea.

jared-s:
Can be addressed later with other `address_t` refactors

PR #1112, opened 2021-03-26
docs: add adr for routing protocol formats

hairyhum:


jared-s:
Merged as https://github.com/ockam-network/ockam/pull/1167

PR #1124, opened 2021-03-29
feat(rust): node macro uses ockam result

jared-s:
Modifies the `#[ockam::node]` macro to synthesize a `use ockam::Result` statement. This allows smoother usage of the APIs.


jared-s:
Not a good solution, there are additional complexities here. Can easily wind up with redundant imports which trigger lint errors

PR #1148, opened 2021-04-01
feat(rust): return error from process function after kex is completed

SanjoDeundiak:


jared-s:
Merged as https://github.com/ockam-network/ockam/pull/1150

PR #1149, opened 2021-04-01
Sanjo/channel routing fixes

SanjoDeundiak:

### DIFF
--- a/implementations/rust/ockam/ockam_core/src/message.rs
+++ b/implementations/rust/ockam/ockam_core/src/message.rs
@@ -43,25 +43,36 @@ impl From<serde_bare::Error> for crate::Error {
 /// without requiring changes in the user message types.
 pub struct Routed<M: Message> {
     inner: M,
-    route: Route,
+    return_: Route,
+    onward: Route,
 }
 
 impl<M: Message> Routed<M> {
     /// Create a new Routed message wrapper
-    pub fn new(inner: M, route: Route) -> Self {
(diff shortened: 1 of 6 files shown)


SanjoDeundiak:
Changes by @spacekookie that make Channels feasible

spacekookie (approved):
Looks good!

implementations/rust/ockam/ockam_core/src/message.rs
```diff
@@ -43,25 +43,36 @@ impl From<serde_bare::Error> for crate::Error {
 /// without requiring changes in the user message types.
 pub struct Routed<M: Message> {
     inner: M,
-    route: Route,
+    return_: Route,
+    onward: Route,
```
mrinalwadhwa, line 47: could we make these fields `return_route` and `onward_route`?

SanjoDeundiak: I think we should do that in a separate PR. @spacekookie what do you think?

mrinalwadhwa: Different PR works 👍

spacekookie: Sorry, I didn't see the followup messages before doing a push. I reverted the commit. Will open a PR in a sec

jared-s:
Merged https://github.com/ockam-network/ockam/pull/1152

PR #1158, opened 2021-04-02
feat(elixir): helper module to send telemetry to a remote node

hairyhum:
### Proposed Changes

Add a module to use as telemetry message handler to send telemetry to remote nodes like `ockam-hub-ui`


jared-s:
Merged https://github.com/ockam-network/ockam/pull/1165

PR #1164, opened 2021-04-05
feat: send messages to UI to aid end-user development

mjc:
This attaches telemetry to our UI so that users can see the messages they send on  their node.


hairyhum (commented):
This makes sense, although let's rename `send_to_ui` to something implying it's setting up the sending handler, maybe "setup_send_to_ui" or "attach_send_to_ui".

hairyhum (approved):

PR #1168, opened 2021-04-05
Jared/ockam signature core

jared-s:
# v0.1.0 - 2021-04-05

- Initial release.


jared-s:
After https://github.com/ockam-network/ockam/pull/1166

PR #1170, opened 2021-04-05
feat(rust): add bbs+ signature scheme

mikelodder7:


jared-s:
Rebasing in https://github.com/ockam-network/ockam/pull/1171

PR #1179, opened 2021-04-06
ockam_node v0.5.0

jared-s:
### Added
- Expose onward route information to user workers.
- Make context receive retry receiving if wrong type.
- Handle message payloads without inner length.
- Make consuming cancel wrapper yield routed wrapper
- Use special message parsing for context receive.

### Changed
- Updated dependencies.
- Log worker errors instead of panicking.


jared-s:
After https://github.com/ockam-network/ockam/pull/1178

PR #1181, opened 2021-04-06
feat(rust): reject to spawn worker with colliding addresses

spacekookie:
This commit adds a check to not initialise & spawn workers that are being registered for addresses that are already in-use on the system.

Not all spawn errors are being caught yet, and the error types returned by `ockam_node` should be looked at again to make sure we can properly communicate to the user when things go wrong.


jared-s:
Merged as https://github.com/ockam-network/ockam/pull/1210

PR #1184, opened 2021-04-06
feat(rust): simplify tcp worker api for most common usecase

spacekookie:
\


jared-s:
Merged as https://github.com/ockam-network/ockam/pull/1212

PR #1209, opened 2021-04-08
feat(rust): allow duplicating context outside of worker lifecycle

spacekookie:
This change allows the node Context to be used outside of the regular worker lifecycle, the same way that the root worker (so far called `App`, now renamed to `NullWorker`) has previously done.

The `Context` still has an address, and can wait for messages with `ctx.receive()`, but no longer runs its lifecycle loop (accepting messages) on its own.

This is a feature needed by @SanjoDeundiak to more easily implement wrapper types for the vault APIs. It also allows us to refactor the `ockam_transport_tcp` internals to be simplified.

### DIFF
--- a/implementations/rust/ockam/ockam_examples/Cargo.lock
+++ b/implementations/rust/ockam/ockam_examples/Cargo.lock
@@ -2629,6 +2629,7 @@ dependencies = [
  "ockam 0.4.2",
  "ockam_node 0.5.0",
  "serde",
+ "tokio",
  "tracing",
 ]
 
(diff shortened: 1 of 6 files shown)


spacekookie:
Before we merge this PR, make sure to port the address collision checking from #1181 to this API

jared-s:
https://github.com/ockam-network/ockam/pull/1181 has been merged, however there are some changes in https://github.com/ockam-network/ockam/pull/1199 that we should get sorted first

spacekookie:
I updated this PR with the mechanism added in ec8a4fcfe55c9cbbc8509695c1b184b1fdeea3c2

jared-s:
Merged as https://github.com/ockam-network/ockam/pull/1216

PR #1217, opened 2021-04-09
chore(rust): rename passthrough message type to any

spacekookie:


jared-s:
Merged https://github.com/ockam-network/ockam/pull/1218

PR #1227, opened 2021-04-12
Node Context API refactor

spacekookie:
@jared-s I'll let you know when this PR has all required changes outlined in the MIro board

### DIFF
--- a/implementations/rust/ockam/ockam/src/remote_mailbox.rs
+++ b/implementations/rust/ockam/ockam/src/remote_mailbox.rs
@@ -61,7 +61,7 @@ impl<T: Message> RemoteMailbox<T> {
         let worker_address: Address = random();
         ctx.start_worker(worker_address, remote_mailbox).await?;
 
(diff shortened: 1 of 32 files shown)


jared-s:
Sounds good, thank you!

jared-s:
For the rename of `Message::reply` to `return_route`, I think we probably want to keep `reply`with a:

```rust
#[deprecated(since = "0.8.0", note = "Please use the return_route function instead")]
```

For at least a few versions, since we have a lot of folks out there at the moment who have been going through the guide recently. After we update the guide source, we can rip it out.

jared-s:
Eh, I'm sure it will be fine without deprecation at this point

mrinalwadhwa:
Yes I think it fine without depreciation notice for our current preview releases

spacekookie:
`5c84c27` should fix the issue where node panics when the main context goes out of scope. Could you test this @SanjoDeundiak ?

jared-s:
Confirmed that the panic is fixed in my test scenario

spacekookie:
:partying_face:

PR #1243, opened 2021-04-13
feat: use env var to get node fqdn

mjc:
Setting an env var is the fastest way for a node to know its FQDN and external IP.


hairyhum:
Clean up is being done in #1281

PR #1244, opened 2021-04-13
feat: use env var to get node fqdn

jared-s:
Rebase of https://github.com/ockam-network/ockam/pull/1243


jared-s:
Holding on this for now

PR #1250, opened 2021-04-13
fix(rust): fix flakey encryption test

jared-s:
Fixes a condition which caused spurious assertion failures in CI

### DIFF
--- a/implementations/rust/ockam/ockam_vault/src/symmetric_impl.rs
+++ b/implementations/rust/ockam/ockam_vault/src/symmetric_impl.rs
@@ -109,7 +109,8 @@ mod tests {
         assert!(res.is_ok());
         let plaintext = res.unwrap();
         assert_eq!(plaintext, message.to_vec());
-        ciphertext[0] ^= ciphertext[1];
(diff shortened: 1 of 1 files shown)


implementations/rust/ockam/ockam_vault/src/symmetric_impl.rs
```diff
@@ -109,7 +109,8 @@ mod tests {
         assert!(res.is_ok());
         let plaintext = res.unwrap();
         assert_eq!(plaintext, message.to_vec());
-        ciphertext[0] ^= ciphertext[1];
```
jared-s, line 112: The spirit of this statement is to corrupt the ciphertext and show that decryption fails. In CI, sometimes we hit the case where one or both are zero, the ciphertext will remain unchanged, decryption will succeed, and the assertion fails.

This change corrupts the ciphertext in two places with constants, which will always fail

jared-s: @SanjoDeundiak you might find this interesting

PR #1255, opened 2021-04-14
Getting tcp_transport working with multi-hop routes

spacekookie:
To run the example:

1. cargo run --example 08-routing-over-transport-many-hops-responder
2. cargo run --example 08-routing-over-transport-many-hops-middle
3. cargo run --example 08-routing-over-transport-many-hops-initiator

To run this in a single binary the TcpTransport API needs to change slightly.

### DIFF
--- a/examples/rust/get_started/examples/08-routing-over-transport-many-hops-initiator.rs
+++ b/examples/rust/get_started/examples/08-routing-over-transport-many-hops-initiator.rs
@@ -1,0 +1,25 @@
+use ockam::{Context, Result, Route};
+use ockam_transport_tcp::{TcpTransport, TCP};
+
+#[ockam::node]
+async fn main(mut ctx: Context) -> Result<()> {
+    TcpTransport::create(&ctx, "127.0.0.1:4000").await?;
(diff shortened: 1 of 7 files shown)


spacekookie:
I'm wondering what you think about this API:

```rust
#[ockam::node]
    // Creates a new router and spawns a listening worker on that port
    let tcp_a = TcpTransport::listen(&ctx, "127.0.0.1:5000").await?;

    // Creates a new router, listens, and then also creates a connection worker on the same router
    let tcp_b = TcpTransport::listen(&ctx, "127.0.0.1:4000").await?;
    tcp_b.connect(&ctx, "127.0.0.1:4000").await?;

    // Creates a new router and creates a connection worker on it
    let tcp_c = TcpTransport::connect(&ctx, "127.0.0.1:4000").await?;
}
```

Keeping the function names as `create` and `create_listener` would also work, but I think the API would be less clear in that case.

mrinalwadhwa:
I like that API, let's change it. I'm merging this PR, let's make that change is separate one.

Issue #1161, opened 2021-04-02
Rust - Implement a higher level TCP registration and initialization process


jared-s:
Establishing a TCP connection to a peer currently requires a workflow [such as](https://github.com/ockam-network/ockam/blob/develop/implementations/rust/ockam/ockam_examples/examples/step-4-client.rs#L11-L15):

```rust
// Create and register a connection
let router = TcpRouter::register(&ctx).await?;
let connection = tcp::start_tcp_worker(&ctx, remote_node.parse::<SocketAddr>().unwrap()).await?;
router.register(&connection).await?;
```

We would like to simplify this process as much as possible, ideally to a single call.

jared-s:
@spacekookie another issue related to TCP refactors

spacekookie:
We can probably close this issue now, since the TCP transport has been refactored.

PR #1256, opened 2021-04-14
chore(rust): simplify tcp init process

feniks65:
### Proposed Changes

Rust - Implement a higher level TCP registration and initialization process

In this PR I wanted to address [#1161](https://github.com/ockam-network/ockam/issues/1161).

@jared-s 
What do you think of such a function?
Is there ockam_transport_tcp lib.rs correct place to put such a function?
Were you thinking of something different than just wrapping initialization and registration code into a public function?
In here, I am just returning `Context` and then `send_message()` can be called on this context.

### DIFF
--- a/implementations/rust/ockam/ockam_transport_tcp/src/lib.rs
+++ b/implementations/rust/ockam/ockam_transport_tcp/src/lib.rs
@@ -61,3 +61,13 @@ impl TcpTransport {
         TcpRouter::bind(ctx, socket_addr).await
     }
 }
+
+pub async fn establish_tcp_connection(
+    ctx: &mut Context,
+    remote_node: String,
+) -> Result<&mut Context> {
+    let router = TcpRouter::register(ctx).await?;
+    let connection = TcpTransport::create(ctx, remote_node.parse::<SocketAddr>().unwrap()).await?;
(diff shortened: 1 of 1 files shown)


jared-s:
Hey @feniks65!

We just improved this API, in the latest `ockam_transport_tcp` crate release (and on develop). Now, all that is necessary is a call to `TcpTransport::create`. 

```rust
    // Get our peer address
    let peer_addr = get_peer_addr();

    // Initialize the TCP stack by opening a connection to a the remote
    TcpTransport::create(&ctx, &peer_addr).await?;

    // Send a message to the remote
    ctx.send(
        Route::new()
            .append_t(TCP, &peer_addr)
            .append("echo_service"),
        String::from("Hello you over there!"),
    )
    .await?;
````

It's a lot better. please give it a shot!

jared-s:
The guides will be updated to reflect these new changes today.

feniks65:
@jared-s 
Sure, I should have noticed that function `create()` is already in there.
So there is no need to simplifying anything now. `TcpTransport::create()` does the thing.

jared-s:
Thanks for the suggestion though! Keep 'em coming!

PR #1265, opened 2021-04-15
fix(rust): revert tcp router address to default static

jared-s:
Temporary fix for https://github.com/ockam-network/ockam/issues/1259


jared-s:
Not a fix

PR #1280, opened 2021-04-16
feat(rust): implement ps signature and refactor structures

mikelodder7:
Implements Pointcheval Saunders signatures
Refactors some code from BBS+ into signature_core.


jared-s:
Merged as https://github.com/ockam-network/ockam/pull/1298

Issue #1296, opened 2021-04-21
Kafka integration for hub stream service


hairyhum:
Stream service is supposed to work with different storage backends. 
Kafka integration can be implemented as a storage backend.

Open questions:
- Start it as a separate hub service or an option to stream service
- Using kafka partitions: adding partition abstraction to the stream service or implement with multiple streams
- Use kafka offset storage to implement index service or keep it separate
- Use kafka stream service as a gateway between ockam system and outside infra talking with kafka directly

hairyhum:
Kafka offset storage as a backend for index service: https://github.com/ockam-network/ockam/pull/1297/files#diff-fa62920271210d858f7069553704c409aad53fbbadd370849c03f1c1dc66623e

Kafka stream is saving plain messages. It can be used to convert ockam messages to kafka messages and back without additional code using https://github.com/ockam-network/ockam/pull/1297/files#diff-3f10cd6d6f76de096062edf898f45c2ee6c5937539e08b24946a1a650b84bba9

Kafka stream storage and offset storage can be set up as parameters when creating a service: https://github.com/ockam-network/ockam/pull/1297/files#diff-1219400148749e39771fe8a5309354c90869b187b55fbbcdea815ed986a653ffR42

hairyhum:
Addressed in #1297

PR #1297, opened 2021-04-21
feat(elixir): kafka storage integration for stream service

hairyhum:
### Proposed Changes

Implementing #1296


hairyhum:
`kafka_ex` does not support Kafka SASL protocol.
To make things like Confluent cloud work with Ockam, the library needs to be replaced with [kafka_protocol](https://hex.pm/packages/kafka_protocol) for example

hairyhum:
Using `brod` library for kafka integration

PR #1320, opened 2021-04-23
feat(rust): support modular protocol parsing in ockam crate

spacekookie:
The general design of this protocol parsing approach is this:

- A worker that wants to parse many different protocols creates a `ProtocolParser<W>`. This type  is generic for the worker it is being implemented for
- Each protocol has a type that implements the `ParserFragment<W>` trait
  - This trait is used to let the main protocol parser know which protocol IDs it can handle
- There is a `UserParser<W, T>` closure type which MUST be used by any parser implementing `ParserFragment<W>`
  - `W` is the worker for which this parser type is implemented, and `T` is the response type given to the user to map to worker state
  - **Example**: `stream::ResponseParser` is generic over `W` but constrains `T: stream::responses::Response`
- Initialising the `ProtocolParser` SHOULD happen in the worker's `initialize` function.
  - Use `.attach(...)` to register a type implementing `ParserFragment<W>` with the `ProtocolParser`
- Parsing can then be handled by the `ProtocolParser` in the worker's `handle_message` function
  - Due to borrowing constraints a user must first call `.prepare()` on the parser
  - The parser then takes a mutable borrow of the worker state and the message being parsed
- The `UserParser<W, T>` closure gets access to the mutable worker state
  - This means no closure environment is being captured --> the closure is `Send` and `Sync` 
  - Each protocol-specific `UserParser` can then match on the incoming message type

## Example

There is a full example worker using this mechanism in the `ockam_examples` crate.

```rust
#[derive(Default)]
struct MyWorker {
    parser: ProtocolParser<MyWorker>,
    stream: Option<String>,
}

#[async_worker]
impl Worker for MyWorker {
    type Context = Context;
    type Message = Any;

    async fn initialize(&mut self, _: &mut Context) -> Result<()> {
        self.parser
            .attach(ResponseParser::new(|worker: &mut Self, msg| match msg {
                Response::Init(Init { stream_name }) => worker.stream = Some(stream_name),
                _ => { /* ignore messages */ }
            }));

        Ok(())
    }

    async fn handle_message(&mut self, ctx: &mut Context, msg: Routed<Any>) -> Result<()> {
        self.parser.prepare().parse(self, msg)?;

        println!("Stream name is now: `{:?}`", self.stream);
        Ok(())
    }
}
```

### DIFF
--- a/implementations/rust/ockam/ockam/src/lib.rs
+++ b/implementations/rust/ockam/ockam/src/lib.rs
@@ -16,6 +16,9 @@
 #[macro_use]
 extern crate serde_big_array;
 
+#[macro_use]
+extern crate tracing;
+
 big_array! { BigArray; 96 }
 
 // ---
@@ -46,9 +49,11 @@ mod remote_forwarder;
 pub use remote_forwarder::*;
 
 pub use ockam_core::worker;
+pub mod protocols;
+
 pub use ockam_core::{
-    Address, Any, Encoded, Error, Message, Result, Route, Routed, RouterMessage, TransportMessage,
-    Worker,
+    Address, Any, Encoded, Error, Message, ProtocolId, Result, Route, Routed, RouterMessage,
+    TransportMessage, Worker,
 };
 
 pub use ockam_channel::SecureChannel;
--- a/implementations/rust/ockam/ockam/src/protocols/parser.rs
+++ b/implementations/rust/ockam/ockam/src/protocols/parser.rs
@@ -1,0 +1,123 @@
+use crate::{
+    error::OckamError, protocols::ProtocolPayload, Any, Message, ProtocolId, Result, Routed, Worker,
+};
+use std::{collections::BTreeMap, marker::PhantomData, ops::Deref, sync::Arc, sync::RwLock};
+
+/// A parser for a protocol fragment
+///
+/// **If you are not a protocol author, you may want to use
+/// [`UserParser`](UserParser) instead!**
+///
+/// Protocols are implemented as separate structures, wrapped in a
+/// carrier type.  Because Rust can't have a function return different
+/// types from a function, each protocol message (here called
+/// "Fragment") needs to be handled by a separate parser.
+pub trait ParserFragment<W>
+where
+    W: Worker,
+{
+    /// Return the set of `ProtocolID`s this parser can handle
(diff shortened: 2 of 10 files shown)


implementations/rust/ockam/ockam/src/lib.rs
```diff
@@ -6,6 +6,9 @@
 #[macro_use]
 extern crate serde_big_array;
 
+#[macro_use]
+extern crate tracing;
```
SanjoDeundiak, line 10: This is not required

spacekookie: There is a `trace!` line in the protocol parser and I think we will want to start logging more in the ockam crate, which is made much more easier by including the macros in the root of the ockam crate

implementations/rust/ockam/ockam/src/protocols/parser.rs
```diff
@@ -0,0 +1,123 @@
...
+    fn ids(&self) -> Vec<ProtocolId>;
+
+    /// Parse an incoming message for a particular worker
+    fn parse(&self, _state: &mut W, _msg: ProtocolPayload) -> Result<()> {
+        Ok(())
+    }
+}
+
+/// A user-closure to map protocol return values to worker state
+pub type UserParser<W, T> = Box<dyn Fn(&mut W, T) + Send + Sync + 'static>;
```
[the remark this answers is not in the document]
spacekookie: The change ended up being much simpler than I thought it would be. It uses proper generics now.

Thanks for the feedback!

implementations/rust/ockam/ockam/src/lib.rs
SanjoDeundiak: I thought we're moving towards latest rust syntax https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html
Do you think #[macro_use] is more convenient?

spacekookie: For stuff like the log macros I think `#[macro_use]` is more convenient, because otherwise you have to import all different log levels you want to use in a file, so `use tracing::{info, warn}`, and say you want to change a `warn` to `error`, then you need to add the import, etc

jared-s (approved):

jared-s:
Really cool stuff. Merged

PR #1322, opened 2021-04-25
feat(elixir): Token Lease Manager Prototype 

mendrugory:
### Proposed Changes

The motivation of this PR is the possibility of including a Flexible Token Lease Manager which will interact with different Cloud APIs. This PR also includes an implementation for InfluxDB.

Token Lease Manager has been implemented as Ockam Worker, although it has been tested during development as GenServer.

Token Lease Manager will depend on two implementations:
* Cloud system: Where the actual tokens will be created and revoked
* Storage system: Where the Lease will be stored due to some lack of required information in the Cloud API, like expiration time or time-to-live.

Storage system has not been implemented in this PR.

Cloud system has been thought as a Behaviour and this one is like a CRUD (Creation, Read, Update and Deletion). Then, the implementation of this behaviour will depend on the actual Cloud API although there are several approaches. 
With the first approach, the process of the Manager will directly interact with the Cloud API through its implementation. Other approach could be to run the cloud system in a different process, which implies to communicate token manager and cloud system processes. The first approach, implies to store some configuration or state about the cloud system in the manager. The second approach implies to have another process and extra-comunication between processes.
The PR has been implemented with the first approach.

Token Lease Manager will choose its cloud system implementation using a factory pattern. 

Influxdb Cloud system calls have been implemented using http requests, thanks to httpoison library, instead of any influxdb client for elixir. The configuration/state of the cloud api implementation can be passed or it can be also generated "on the fly" if it is not, although the performance is worse.

Another two approaches came to my mind to revoke tokens. The first one was to use a cron-like system, but the storage system would receive too many requests if we want to have a decent accuracy. Other approach would be to take advantage of Elixir message system and send a delayed message to the Token Lease Manager. This approach is going to be very accurate and although the amount of messages would be proportional to the number of tokens, they should not consume too much memory. The main concern about this approach is that the expiration message will be in memory so, if the BEAM is restarted, the messages will be lost. In order to solve this issue, the manager should retrieve all the tokens from the storage system (maybe also from the cloud system and merge them) and create the expiration messages again at start time.

Examples of calls:
```elixir
# Token Lease Manager Creation
iex> send(manager, {:create, %{"expiration" => 120, "orgID" => "2dd6f18583c359d5","permissions" => [%{"action" => "read", "resource" => %{"type" => "authorizations"}}]}})
# Token Lease Manager Deletion
iex> send(manager, {:revoke, "076b91cb6b409000"})
```

### DIFF
--- a/implementations/elixir/ockam/ockam/mix.exs
+++ b/implementations/elixir/ockam/ockam/mix.exs
@@ -47,7 +47,6 @@ defmodule Ockam.MixProject do
     [
       {:bare, "~> 0.1.1"},
       {:gen_state_machine, "~> 3.0"},
-      {:httpoison, "~> 1.8"},
       {:ockam_vault_software, path: "../ockam_vault_software", optional: true},
       {:telemetry, "~> 0.4.2", optional: true},
       {:ranch, "~> 2.0", optional: true},
--- a/implementations/elixir/ockam/ockam/mix.exs
+++ b/implementations/elixir/ockam/ockam/mix.exs
@@ -47,6 +47,7 @@ defmodule Ockam.MixProject do
     [
       {:bare, "~> 0.1.1"},
       {:gen_state_machine, "~> 3.0"},
+      {:httpoison, "~> 1.8"},
       {:ockam_vault_software, path: "../ockam_vault_software", optional: true},
       {:telemetry, "~> 0.4.2", optional: true},
       {:ranch, "~> 2.0", optional: true},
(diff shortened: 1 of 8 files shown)


mrinalwadhwa:
Thank you for sending the PR ❤️

Before we can merge, could you accept our CLA

Read the CLA at
https://www.ockam.io/learn/how-to-guides/contributing/cla/

To accept the CLA please send a different PR to our contributors repo indicating that you accept the CLA by adding your Git/Github details in a row at the end of the CONTRIBUTORS.csv file:
https://github.com/ockam-network/contributors/blob/master/CONTRIBUTORS.csv

Looking forward to merging your first contribution!!

hairyhum:
@mendrugory I'm going to rework that a bit to make it a GenServer, because it's not implementing the Ockam.Worker API yet, for consistency so we can merge it.

mendrugory:
Go ahead

hairyhum:
Moving to https://github.com/ockam-network/ockam/pull/1401 on order to merge

PR #1324, opened 2021-04-26
refactor(elixir): enforce use of typed struct for message

hairyhum:
### Proposed Changes

In order to improve usability of `Ockam.Worker`, make the message type explicit struct and match handle_message on the message type only.
Also match the message type in `Ockam.Router.route`


hairyhum:
Maybe we could handle any type of messages in `Ockam.Router.route/1` and pass properly typed messages to the handlers, but if the code has to use payload, onward_route and return_route it might as well use the struct when creating a message.
There are helper functions in the `Ockam.Message` module to mimic those in Rust API, which may be used more for that purpose.

hairyhum:
API changes in this PR:

1. `Ockam.Worker.handle_message/2` is now can only called with `Ockam.Message` as a first argument
This allows to use `GenServer.handle_info/2` in workers to handle other types of erlang messages
2. `Ockam.Router.route/1` still supports plain maps, but now `return_route` is optional there (default is `[]`)
3. Added `Ockam.Router.route/(2,3)` for convenience
To route a simple message you no longer need to create a message structure, instead you can run `Ockam.Router.route("something", ["somewhere"])`
5. Added message forwarding helpers to forward messages and trace return routes

`Message.forward(message)` - drop the first address in onward route
`Message.forward(message, route)` - replace onward route with `route`
`Message.trace_address(message, address)` - add `address` to the return route
`Message.forward_trace(message)` - forward and trace first address in onward route
`Message.forward_trace(message, address)` - forward and trace `address`
`Message.forward_trace(message, address, route)` - forward to `route` and trace `address`

PR #1333, opened 2021-04-27
Rust cli github authentication

piiih:
### Proposed Changes
Issue: https://github.com/ockam-network/ockam/issues/1068

Basically you just need to run the following in `tools/ockam-hub-cli` directory:
```sh
cargo run -- auth
```
PS: It doesn't show the ockam-hub's address, because I didn't have api instructions. If I have, I can implement it :)


tools/ockam-hub-cli/src/auth/github.rs
```diff
@@ -0,0 +1,119 @@
+use colored::*;
+use std::collections::HashMap;
+use std::future::Future;
+use std::pin::Pin;
+use webbrowser;
+
+const GITHUB_CLIENT_ID: &'static str = "609323e5ba23958cb2f5";
```
jared-s, line 7: Can we externalize this as an environment variable?

tools/ockam-hub-cli/src/auth/github.rs
```diff
@@ -0,0 +1,119 @@
...
+            payload,
+        )
+        .await?;
+
+        if let Some(error) = response.get("error") {
+            if error == "authorization_pending" {
+                let duration = std::time::Duration::from_secs(time_to_retry);
+                std::thread::sleep(duration);
+                return try_access_token(client, payload, time_to_retry).await;
+            }
```
jared-s, line 43: Can we add an `else` here that will panic on any other kind of error?

tools/ockam-hub-cli/src/auth/github.rs
jared-s: Or, if this value is a stable identifier that will absolutely never change, let's add documentation to say what the value is, how to obtain it. etc

jared-s (changes requested):
A few questions/changes - thanks!

tools/ockam-hub-cli/src/auth/github.rs
piiih: This value is a fixed value that represents the `ockam` app on github, it won't change.

piiih: What do you think that is the best approch @jared-s ?

jared-s: Thanks for the clarification, It's okay for now.

jared-s:
Merged the changes in https://github.com/ockam-network/ockam/pull/1403 - thanks!

tools/ockam-hub-cli/src/auth/github.rs
piiih: Sry for the long time to fix, I was busy this week :/

PR #1363, opened 2021-05-04
Chore: add cargo audit to rust's lint task

adrianbenavides:
## Changes
* Run `cargo audit` in the rust's lint tasks
* Add a new gradle task to run `cargo audit fix` automatically for all rust subcrates (not used in CI). Let me know if you think this can be helpful

## Related issues
* https://github.com/ockam-network/ockam/issues/1048

## Other
Note that no `Cargo.toml` changes are needed right now, as all the `cargo audit` warns/errors that it was complaining about a couple of weeks ago have been fixed in the past few PR's.

Also, since the `lint` stage is not required, is it okay to let it fail when `cargo audit` returns errors?

### DIFF
--- a/implementations/rust/build.gradle
+++ b/implementations/rust/build.gradle
@@ -31,6 +31,9 @@ commands {
       'cargo clippy --all --all-targets -- --cap-lints=warn',
       'cargo audit',
     ],
+    audit_fix: [
+      'cargo audit fix',
+    ],
   ]
 
   directories = crates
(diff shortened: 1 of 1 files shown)


mrinalwadhwa:
@adrianbenavides thank you for contributing this 🙏 

* I rebased and merged the changes in https://github.com/ockam-network/ockam/pull/1363/commits/a13cdfd4406c2cec22829311a8cec685bcf9e6ab

* https://github.com/ockam-network/ockam/pull/1363/commits/3350f62bc94836fc17073999c30f0749c410d8a2 I feel we should leave out for now, most people will work on individual crates and they can call `cargo audit fix` manually in that crate. The blanket gradle task I feel is not needed. In case we discover that enough devs need it, I'll come back an apply the above patch.

Thank you for contributing ❤️

mrinalwadhwa:
Merged https://github.com/ockam-network/ockam/commit/6a09b756477585f1b9620a67be9b45087eb37c46

PR #1367, opened 2021-05-05
feat(rust): initial profile traits

jared-s:
@SanjoDeundiak what do you think about something like this?

### DIFF
--- a/implementations/rust/ockam/ockam/src/profile.rs
+++ b/implementations/rust/ockam/ockam/src/profile.rs
@@ -39,7 +39,7 @@ pub type ContactsDb = HashMap<ProfileIdentifier, Contact>;
 ///
 /// ```
 /// # use ockam_vault::SoftwareVault;
-/// # use ockam::{Profile, KeyAttributes, Vault};
+/// # use ockam::{Profile, ProfileChanges, ProfileSecrets, KeyAttributes, Vault};
 /// # fn main() -> ockam_core::Result<()> {
 /// # let (mut ctx, mut executor) = ockam_node::start_node();
 /// # executor.execute(async move {
@@ -76,7 +76,7 @@ pub type ContactsDb = HashMap<ProfileIdentifier, Contact>;
 ///
 /// ```
 /// # use ockam_vault::SoftwareVault;
-/// # use ockam::{Profile, Vault};
+/// # use ockam::{Profile, ProfileAuth, ProfileContacts, ProfileSecrets, Vault};
 /// fn alice_main() -> ockam_core::Result<()> {
 ///     # let (mut ctx, mut executor) = ockam_node::start_node();
 ///     # executor.execute(async move {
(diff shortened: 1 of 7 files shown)


implementations/rust/ockam/ockam/src/profile.rs
```diff
@@ -191,18 +191,98 @@ impl Profile {
...
     /// Return unique [`Profile`] identifier, which is equal to sha256 of the root public key
-    pub fn identifier(&self) -> &ProfileIdentifier {
+    fn identifier(&self) -> &ProfileIdentifier;
+}
+
+pub trait ProfileChanges {
+    /// Return change history chain
+    fn change_events(&self) -> &[ProfileChangeEvent];
+    /// Add a change event.
+    fn update_no_verification(
```
jared-s, line 203: I feel like this should be renamed, the current name here is unclear. `add_change_event`? `add_change_no_verify`? How critical is the 'no verification' aspect? I don't see an alternative which updates and also verifies..

implementations/rust/ockam/ockam/src/profile.rs
```diff
@@ -191,18 +191,98 @@ impl Profile {
...
+        &mut self,
+        change_event: ProfileChangeEvent,
+    ) -> ockam_core::Result<()>;
+
+    /// Verify the whole change event chain
+    fn verify(&mut self) -> ockam_core::Result<()>;
```
jared-s, line 209: This could also use a more descriptive name. `verify_changes` or `verify_profile`?

implementations/rust/ockam/ockam/src/profile.rs
```diff
@@ -191,18 +191,98 @@ impl Profile {
...
+        profile_id: &ProfileIdentifier,
+        change_events: Vec<ProfileChangeEvent>,
+    ) -> ockam_core::Result<()>;
+}
+
+pub trait ProfileAuth {
+    fn generate_authentication_proof(
+        &mut self,
+        channel_state: &[u8],
+    ) -> ockam_core::Result<Vec<u8>>;
```
jared-s, line 237: We should introduce a high level type for this proof `Vec<u8>`

implementations/rust/ockam/ockam/src/profile/channel.rs
```diff
@@ -2,6 +2,8 @@ use ockam_core::{Address, Result, Route};
 use ockam_node::Context;
 
 mod listener;
+use crate::profile::ProfileAuth;
+use crate::profile::ProfileContacts;
```
jared-s, line 6: How should we handle trait explosion like this?

- Per crate preludes? `use crate::profile::prelude::*"?
- Just accept it and import what's necessary?
- ?
-

mrinalwadhwa: Internally, I think we should import what's necessary and keep dependencies explicit.
We can explore prelude::* for user facing api incase it becomes necessary ... but I suggest we avoid it as long as we can.

SanjoDeundiak (approved):

PR #1370, opened 2021-05-06
refactor(rust): rename async_worker to worker

### LINKED ISSUE #1220: rename #[async_worker] to #[ockam::worker]
1. This export would change
https://github.com/ockam-network/ockam/blob/develop/implementations/rust/ockam/ockam/src/lib.rs#L40

to:
```
... as worker;
```

2. we'll also have to find and replace all use of `#[async_worker]` with `#[ockam::worker]`
3. remove imports of  `async_worker`, #[ockam::worker] will work if `ockam` is a dependency.

spacekookie:
Fixes #1220


jared-s:
Rebased and merged

PR #1380, opened 2021-05-11
Implement sync and async node API split

spacekookie:
This commit is mostly an experiment to investigate how we can split
the sync and async variants of functions with the least amount of code
duplication.  The approach taken in this commit comes with a bit of
overhead in terms of documentation, because functions SHOULD be
documented for both their sync and async variants.

It does come with the advantage that both sync and async APIs can be
made public via compile-options.


spacekookie:
While I won't have the chance to finish this conversion in the next 2 days, I changed my mind about the validity of this change: I think it will make things a lot easier when building services in the `ockam` crate, where we use closures to call parsers for specific protocols. Because Rust doesn't have async closures (in `stable`) yet this will make it much easier to call `Context` functions.

Issue #1371, opened 2021-05-06
Elixir: Creating new workers from Ockam.Worker.setup hangs 


hairyhum:
If a worker wants to create another worker in its `setup` function, this will hang because of the gen-call loop with worker supervisor.
Currently to start an additional worker it should use `handle_` functions in some way.

`GenServer` supports delayed init, which can be used to run `setup` function for a worker, instead of `GenServer.init`, OR delayed init should be exposed to the `Ockam.Worker` API to allow creating additional workers on init.

PR #1391, opened 2021-05-12
chore(elixir): add delayed init to ockam.worker

feniks65:
### Proposed Changes

Add delayed init for Ockam.Worker

@hairyhum
As it was requested by you, I wanted to add delayed init for `Ockam.Worker` as solution for [#1371 ](https://github.com/ockam-network/ockam/issues/1371).

My idea behind this is to immediately get back from the `init()` call. Then I have moved all the code from `init()` to the specific async call handle_cast(). 
Since casts calls are handled asynchronously, the code from the `init()` should not block creating additional worker in the `setup()` call.

Is it something that you had in your mind?

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/worker.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/worker.ex
@@ -37,8 +37,12 @@ defmodule Ockam.Worker do
         options = Keyword.put_new_lazy(options, :address, &Node.get_random_unregistered_address/0)
 
         case Node.start_supervised(__MODULE__, options) do
-          {:ok, _pid, worker} -> {:ok, worker}
-          error -> error
+          {:ok, pid, worker} ->
+            :sys.get_state(pid)
+            {:ok, worker}
+
+          error ->
+            error
         end
       end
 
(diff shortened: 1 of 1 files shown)


hairyhum:
@feniks65 I was thinking of using the `{:ok, state, {:continue, ...}}` reply: https://hexdocs.pm/elixir/GenServer.html#c:init/1
The `handle_continue` callback is called after init replies to the `start_link`, so it's not blocking the supervisor. At the same time it's protected from any stray messages coming in because it's not receiving yet: https://github.com/erlang/otp/blob/master/lib/stdlib/src/gen_server.erl#L394

feniks65:
@hairyhum 
Ok. How about now?
All the work will be done in `handle_continue()` call.

feniks65:
@hairyhum 
There is one question. It seems like my change made router tests fail from time to time.
I was experimenting with the timeout in `assert_receive()` call but still, from time to time those tests are failing.
It seems like since `handle_continue()` is being called asynchronously now those tests need to be changed.

For example if I would increase timeout in `assert_receive()` by 10 x times like that:
`      assert_receive(
        {
          :trace,
          ^ping_pong_server,
          :receive,
          %{
            onward_route: [
              "ping_pong_server"
            ],
            return_route: [registered_server_tcp, "ping_pong_client"],
            payload: "ping 1"
          }
        },
        10_000
      )`
it seems to work most of the time.

What is your suggestion, how should I deal with it?

hairyhum:
I think increasing timeouts in tests is fine in general because test runners might get slow.
Also you can add something like `:sys.get_state(pid)`, which is a synchronous call, after the setup in the test in order to sync with the worker, this should reduce the flakyness of the timeouts.

hairyhum:
@feniks65 hi, the tests are still failing. I have another idea: what if we call `:sys.get_state` in the `Ockam.Worker.create` after `start_supervised`? This way we would be blocking on initialization, but it would block the user code instead of the supervisor.
We can also add `create_async` to not block there. WDYT?

feniks65:
@hairyhum 

Thank you for yet another suggestion. Adding `:sys_get()` in create() call for Ockam.Worker is the thing that I needed.
At least in my docker tests are passing.
I was experimenting for the last few days with RouterTests but with no luck. It seemed to me that whenever Echo test was started as the first test from the RouterTest batch, then the Simple TCP test and Ping Pong test worked fine.
In other cases, one of those two tests was failing.

hairyhum (approved):

hairyhum:
Merged in https://github.com/ockam-network/ockam/pull/1442

PR #1392, opened 2021-05-12
Initial Entity implementation

jared-s:
This creates a delegating Entity. 

- Entity has a default Profile, and delegates all operations to it
- The Entity type implements `ProfileTrait` 
- ProfileTrait updated to add accessor for a Vault, as well as Clone, Send and static lifetime
- SecureChannel creation refactored to use generic Profile traits instead of `ProfileImpl`. This could move into `Entity` eventually but for now i broke it out into a `SecureChannel` builder. Having Profiles construct the Channels themselves was problematic to the design.

### DIFF
--- a/implementations/rust/ockam/ockam_entity/src/entity.rs
+++ b/implementations/rust/ockam/ockam_entity/src/entity.rs
@@ -1,0 +1,190 @@
+use crate::{ProfileIdentifier, ProfileTrait};
+
+pub mod authentication;
+pub use authentication::*;
+pub mod change;
+pub use change::*;
+pub mod contacts;
+pub use contacts::*;
+pub mod identifiers;
+pub use identifiers::*;
+pub mod secrets;
+pub use secrets::*;
+
+/// An Entity represents an identity in various authentication contexts.
+#[derive(Clone)]
+pub struct Entity<P: ProfileTrait> {
+    default_profile_identifier: ProfileIdentifier,
+    profiles: Vec<P>,
+}
+
+impl<P: ProfileTrait> Entity<P> {
+    /// Create a new Entity with the given default profile.
+    pub fn new(default_profile: P) -> Self {
+        let idref = default_profile.identifier().unwrap();
+        let default_profile_identifier = ProfileIdentifier::from_key_id(idref.key_id().clone());
+        let profiles = vec![default_profile];
+        Entity {
+            default_profile_identifier,
+            profiles,
+        }
+    }
+
+    fn default_profile(&self) -> Option<&P> {
+        self.profiles
+            .iter()
+            .find(|profile| self.default_profile_identifier == profile.identifier().unwrap())
+    }
+
+    fn _add_profile(&mut self, profile: P) {
+        self.profiles.push(profile);
+    }
+}
+
+#[cfg(test)]
+#[allow(unreachable_code, unused_variables)]
+mod test {
+    use crate::{
+        Entity, KeyAttributes, Profile, ProfileAuth, ProfileContacts, ProfileIdentity,
+        ProfileSecrets, ProfileSync, ProfileTrait,
+    };
+    use ockam_node::Context;
+    use ockam_vault_sync_core::Vault;
+
+    async fn new_entity(ctx: &Context) -> ockam_core::Result<Entity<ProfileSync>> {
+        let vault = Vault::create(ctx)?;
+
+        let profile = Profile::create(&ctx, &vault).await;
+        assert!(profile.is_ok());
+
+        let profile = profile.unwrap();
+        Ok(Entity::new(profile))
+    }
+
+    #[test]
+    fn test_new_entity() {
+        let (mut ctx, mut executor) = ockam_node::start_node();
+        executor
+            .execute(async move {
+                let e = new_entity(&ctx).await.unwrap();
+                assert!(!e
+                    .default_profile_identifier
+                    .to_string_representation()
+                    .is_empty());
+                assert!(!e.profiles.is_empty());
+
+                let default = e.default_profile();
+
+                assert!(default.is_some());
+                ctx.stop().await.unwrap();
+            })
+            .unwrap();
+    }
+
+    fn entity_auth_tests<P: ProfileTrait>(mut e: Entity<P>) -> ockam_core::Result<()> {
+        let channel_state = "test".as_bytes();
+        let proof = e.generate_authentication_proof(channel_state);
+        assert!(proof.is_ok());
+
+        let proof = proof.unwrap();
+
+        // TODO WIP: Need Contacts for this test to be successful. This tests the delegation but not correct operation currently.
+        let default_id = e.default_profile_identifier.clone();
+        let valid = e.verify_authentication_proof(channel_state, &default_id, proof.as_slice());
+        // assert!(valid.is_ok());
+        Ok(())
+    }
+
+    fn entity_change_tests<P: ProfileTrait>(e: Entity<P>) -> ockam_core::Result<()> {
+        // TODO WIP: Need key ops and other event generating APIs to easily test this
+        // change_events update_no_verification verify
+        Ok(())
+    }
+
+    async fn entity_contacts_tests<P: ProfileTrait>(
+        ctx: &Context,
+        mut e: Entity<P>,
+    ) -> ockam_core::Result<()> {
+        //    verify_and_update_contact
+
+        let alice = new_entity(&ctx).await.unwrap();
+        let alice_id = alice.identifier()?.clone();
+
+        let alice_contact = alice.serialize_to_contact()?;
+        let alice_contact = Profile::deserialize_contact(alice_contact.as_slice())?;
+
+        let to_alice_contact = alice.to_contact()?;
+        assert_eq!(alice_contact.identifier(), to_alice_contact.identifier());
+
+        e.verify_contact(&alice_contact)?;
+
+        e.verify_and_add_contact(alice_contact)?;
+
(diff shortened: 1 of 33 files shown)


implementations/rust/ockam/ockam_entity/src/imp.rs
```diff
@@ -207,6 +207,10 @@ impl<V: ProfileVault> ProfileContacts for ProfileImpl<V> {
     }
 
     fn verify_and_add_contact(&mut self, contact: Contact) -> ockam_core::Result<()> {
+        // Don't allow a profile to be a contact of itself.
+        if &self.identifier == contact.identifier() {
```
jared-s, line 211: @SanjoDeundiak what do you think of this check? I thought of the case when writing tests and saw I was able to add a profile as a contact of itself. Seems like it could cause some problems in the future.

implementations/rust/ockam/ockam_entity/src/channel/listener.rs
```diff
@@ -1,28 +1,18 @@
...
 use rand::random;
 
-pub(crate) struct ProfileChannelListener<V: ProfileVault> {
-    profile: ProfileImpl<V>, // TODO: Avoid copying profile
-    vault: V,
-    listener_address: Option<Address>,
-}
-
-impl<V: ProfileVault> ProfileChannelListener<V> {
-    pub fn new(profile: ProfileImpl<V>, vault: V) -> Self {
```
jared-s, line 15: This was making things complicated when using the generics, and constructing it inline was easier. Can't remember which type was giving trouble.

implementations/rust/ockam/ockam_entity/src/entity.rs
```diff
@@ -0,0 +1,363 @@
...
+};
+
+use crate::EntityError::ProfileNotFound;
+use ockam_core::Result;
+use ockam_vault::ockam_vault_core::{PublicKey, Secret};
+
+/// An Entity represents an identity in various authentication contexts.
+#[derive(Clone)]
+pub struct Entity<V: ProfileVault> {
+    default_profile_identifier: ProfileIdentifier,
```
jared-s, line 14: We're storing a copy of the profile identifier, instead of a ref to one in the set. A ref would require introducing a lifetime which we don't want to do. This also allows the underlying Profile to be lazily loaded and otherwise decoupled from the profile impl storage. That being said, A profile with this identifier must be in the `profiles` set. Since a default profile is required on `Entity` construction, we can guarantee this if we forbid removal of the default profile, when we add profile removal.

implementations/rust/ockam/ockam_entity/src/entity.rs
```diff
@@ -0,0 +1,363 @@
...
+    profiles: Vec<ProfileImpl<V>>,
+}
+
+impl<V: ProfileVault> Entity<V> {
+    #[allow(dead_code)]
+    pub fn new(default_profile: ProfileImpl<V>) -> Self {
+        let idref = default_profile.identifier();
+        let default_profile_identifier = ProfileIdentifier::from_key_id(idref.key_id().clone());
```
jared-s, line 22: Thinking of adding a new function on ProfileIdentifier that makes a copy using something like this

implementations/rust/ockam/ockam_entity/src/entity.rs
```diff
@@ -0,0 +1,363 @@
...
+        let profiles = vec![default_profile];
+        Entity {
+            default_profile_identifier,
+            profiles,
+        }
+    }
+
+    fn default_profile(&self) -> Option<&ProfileImpl<V>> {
```
jared-s, line 30: This way of getting the default profile can be used when only immutable access is needed

implementations/rust/ockam/ockam_entity/src/entity.rs
```diff
@@ -0,0 +1,363 @@
...
+        if let Some(profile) = self.default_profile() {
+            profile.change_events()
+        } else {
+            &[]
+        }
+    }
+
+    fn update_no_verification(&mut self, change_event: ProfileChangeEvent) -> Result<()> {
+        for profile in &mut self.profiles {
+            if &self.default_profile_identifier == profile.identifier() {
```
jared-s, line 73: For operations that require mutable access to the default profile, each delegating method iterates the profiles instead of using `default_profile())`

implementations/rust/ockam/ockam_entity/src/entity.rs
```diff
@@ -0,0 +1,363 @@
...
+        if let Some(profile) = self.default_profile() {
+            profile.clone().verify()
+        } else {
+            Err(ProfileNotFound.into())
+        }
+    }
+}
+
+impl<V: ProfileVault> ProfileContacts for Entity<V> {
+    fn contacts(&self) -> &ContactsDb {
```
jared-s, line 90: Wondering if we should change this signature to be wrapped in a Result. It also tightly couples the trait to a data type (a hefty one, a hash map). Might want to be made generic.

implementations/rust/ockam/ockam_entity/src/entity.rs
```diff
@@ -0,0 +1,363 @@
...
+        if let Some(profile) = self.default_profile() {
+            profile.contacts()
+        } else {
+            /* Since the return value is a reference, and we have no default profile, panic.
+            We could potentially store an empty contacts db somewhere and return it here, but that
+            seems like a waste. */
+            panic!("Entity has no default profile")
```
jared-s, line 97: Since this and a few other APIs don't return Result, we have to panic

implementations/rust/ockam/ockam_entity/src/traits.rs
```diff
@@ -9,6 +9,10 @@ pub trait ProfileIdentity {
     fn identifier(&self) -> &ProfileIdentifier;
 }
 
+pub trait ProfileVaultAccess<V: ProfileVault> {
+    fn vault(&mut self) -> V;
```
jared-s, line 13: Required by channels

implementations/rust/ockam/ockam_entity/src/imp.rs
SanjoDeundiak: @jared-s I would say self contact should in contacts list by default. Therefore, adding it will cause duplicate error

implementations/rust/ockam/ockam_entity/src/entity.rs
SanjoDeundiak: @jared-s they will be returning Result after my PR.

implementations/rust/ockam/ockam_entity/src/entity.rs
SanjoDeundiak: We can return Vec instead.

implementations/rust/ockam/ockam_entity/src/channel.rs
```diff
@@ -11,32 +11,51 @@ pub(crate) use listener::*;
...
+        Initiator::create(ctx, route, profile).await
     }
 
     /// Create mutually authenticated secure channel listener
-    pub async fn create_secure_channel_listener<A: Into<Address>>(
-        &mut self,
+    #[allow(dead_code)]
+    pub async fn create_secure_channel_listener<A, V, P>(
         ctx: &Context,
+        mut profile: P,
```
SanjoDeundiak, line 37: Let's add Clone to ProfileTrait and replace this with reference

implementations/rust/ockam/ockam_entity/src/channel.rs
```diff
@@ -11,32 +11,51 @@ pub(crate) use listener::*;
...
+        R: Into<Route>,
+        V: ProfileVault,
+        P: ProfileTrait<V>,
+    {
+        Initiator::create(ctx, route, profile).await
     }
 
     /// Create mutually authenticated secure channel listener
-    pub async fn create_secure_channel_listener<A: Into<Address>>(
-        &mut self,
```
SanjoDeundiak, line 26: What is the reason why it's no longer a Profile method?

implementations/rust/ockam/ockam_entity/src/channel.rs
```diff
@@ -11,32 +11,51 @@ pub(crate) use listener::*;
...
+    #[allow(dead_code)]
```
SanjoDeundiak, line 34: It looks like this function doesn't belong to crate's public API, otherwise it wouldn't be dead_code

implementations/rust/ockam/ockam_entity/src/channel.rs
jared-s: Because it can't be expressed as a trait. I tried using async trait but there were some other constraints in the signature.

jared-s: This is functionality that Entity needs to proxy to Profile somehow, so we either need to refactor the signature to be friendlier to a trait that both can implement, or manually add the methods on Entity.

SanjoDeundiak: Interesting, don't see anything obvious why this can't be a trait

implementations/rust/ockam/ockam_entity/src/channel.rs
jared-s: Let's pair after demos tomorrow, I'll go through what I did and see where I got stuck

implementations/rust/ockam/ockam_entity/src/channel.rs
jared-s: Let's pair after demos tomorrow, we can go through and see where I got stuck

PR #1397, opened 2021-05-13
docs: tweak file header for hop.rs

sobi3ch:
### Proposed Changes
Small docs tweak. Adding file header for the hop code for 03-routing


jared-s:
Merged, thanks!

PR #1398, opened 2021-05-13
chore(rust): turn worker shutdown into async function

spacekookie:
@SanjoDeundiak that should fix the issue you were having


jared-s:
Merged

PR #1399, opened 2021-05-13
feat(rust): add profile worker

SanjoDeundiak:


SanjoDeundiak:
Should be merger after #1411

PR #1400, opened 2021-05-14
Bi-directional streaming workers in Rust

spacekookie:
This PR adds bi-directional stream workers in the `ockam` crate, along with some other utilities, such as the ability to send delayed events to workers (even self).

## Proposed API

The API doesn't exist yet, but I imagine it will look somewhat like this.

```rust
#[ockam::node]
    let tcp = // ... setup tcp transport

    // tx is the sending worker
    // rx is a handle that a user can block on to read messages from, like a channel
    let (tx, rx) = ockam::stream::connect(&ctx, Route::new().append(hub).append("stream_service"))?;

    ctx.send(Route::new().append(tx).append("my friend"), "Hello!".to_string()).await?;

    Ok(())
}
```


spacekookie:
The current API doesn't work 100% yet but this is generally how it will be used:

```rust
use ockam::{stream::Stream, Context, Result, Route};
use ockam_transport_tcp::{TcpTransport, TCP};

#[ockam::node]
    // ... get hub_addr somehow

    tcp.connect(hub_addr).await?;


    // Stream is a builder to start and configure streams 
    let (tx, rx) = Stream::new(&ctx)?
        // Start stream workers for a route (to stream_service) and name
        // You can also call `.recipient(...)` on `Stream` to make the consumer forward messages automatically
        .connect(
            Route::new()
                .append_t(TCP, hub_addr)
                .append("stream_service"),
            "test-stream".to_string(),
        )
        .await?;

    // You can use the `tx` address like any other address
    ctx.send(tx, "Hello world!".to_string()).await?;

    // But you can do special things with the `rx` address
    let msg: String = rx.next().await?;
    println!("Reply received via stream '{}'!", msg);

    ctx.stop().await
}


```

implementations/rust/ockam/ockam/src/delay.rs
```diff
@@ -0,0 +1,77 @@
...
+        Ok(Self {
+            route,
+            ctx,
+            d: Duration::from_millis(100),
+            msg,
+        })
+    }
+
+    /// Adjust the delay time with a [`Duration`](std::time::Duration)
+    pub fn duration(self, d: Duration) -> Self {
```
SanjoDeundiak, line 30: Let's stick to rust idiomatic naming and start setters with set_

SanjoDeundiak: Or call it something like "with_duration", because "duration" is what getter would be called.

implementations/rust/ockam/ockam/src/protocols/stream/requests.rs
```diff
@@ -2,65 +2,70 @@
 
 use crate::protocols::ProtocolPayload;
 use serde::{Deserialize, Serialize};
+use serde_bare::Uint;
 
 /// Request a new mailbox to be created
+#[repr(C)]
```
SanjoDeundiak, line 8: Is that required for bare serialization to work correctly?

SanjoDeundiak:
> The current API doesn't work 100% yet but this is generally how it will be used:
> 
> ```rust
> use ockam::{stream::Stream, Context, Result, Route};
> use ockam_transport_tcp::{TcpTransport, TCP};
> 
> #[ockam::node]
> async fn main(mut ctx: Context) -> Result<()> {
>     // ... get hub_addr somehow
> 
>     let tcp = TcpTransport::create(&ctx).await?;
>     tcp.connect(hub_addr).await?;
> 
> 
>     // Stream is a builder to start and configure streams 
>     let (tx, rx) = Stream::new(&ctx)?
>         // Start stream workers for a route (to stream_service) and name
>         // You can also call `.recipient(...)` on `Stream` to make the consumer forward messages automatically
>         .connect(
>             Route::new()
>                 .append_t(TCP, hub_addr)
>                 .append("stream_service"),
>             "test-stream".to_string(),
>         )
>         .await?;
> 
>     // You can use the `tx` address like any other address
>     ctx.send(tx, "Hello world!".to_string()).await?;
> 
>     // But you can do special things with the `rx` address
>     let msg: String = rx.next().await?;
>     println!("Reply received via stream '{}'!", msg);
> 
>     ctx.stop().await
> }
> ```

LGTM!

implementations/rust/ockam/ockam/src/protocols/stream/requests.rs
spacekookie: Turns out no. I added those to debug a problem I was having, but it turned out to be unrelated. Thanks for pointing this out, I'll remove those again!

spacekookie:
The API has changed slightly, and this may be something we want to change in the future. Currently this implementation does not use a specific encoding for associating input and output streams, which means that users need to provide this information.

```rust
    let (tx, _) = Stream::new(&ctx)?
        .connect(
            Route::new().append_t(TCP, "127.0.0.1:4000"),
            // Stream name from THIS node to the OTHER node
            "test-a-b",
            // Stream name from OTHER to THIS
            "test-b-a",
        )
        .await?;

    ctx.send(tx.to_route().append("printer"), "Hello World!".to_string())
        .await?;
```

On the remote side (so where the consumer is created) the message will be automatically forwarded to the "printer" worker. If a message is only addressed to the `tx` worker it must be pulled manually on the receiving end.

spacekookie:
@jared-s @mrinalwadhwa this should be ready to merge now :)

spacekookie:
Updated the PR to have two endpoints to configure the stream and index service names on the shared remote.

If we want to support multiple remotes for these services we should probably just remove the shared route and replace those two functions with some that take a route each

antoinevg:
I've rebased this and added the changes needed to bring it into line with the latest `develop` branch.

Can we please merge it?

mrinalwadhwa:
@antoinevg thank you for getting this ready for merge.

In one of the original commits this empty file seems to have snuck into this PR. 
examples/rust/get_started/.projectile

Could you remove that please? then I'll merge.

antoinevg:
Done, thank you @mrinalwadhwa !

PR #1417, opened 2021-05-21
Sanjo/kex update

SanjoDeundiak:


implementations/rust/ockam/ockam_key_exchange_x3dh/src/initiator.rs
```diff
@@ -138,54 +154,22 @@ impl<V: X3dhVault> KeyExchanger for Initiator<V> {
...
-                );
-
-                let skb = if let Some(ik) = &self.identity_key {
-                    ik.clone()
-                } else {
-                    self.vault.secret_generate(atts)?
-                };
-                let ikb = self.vault.secret_public_key_get(&skb)?;
-
-                let mut plaintext = ikb.as_ref().to_vec();
```
SanjoDeundiak, line 166: This doesn't belong to x3dh itself, and I'm not aware of why we had this here. Therefore, I'm removing everything that doesn't belong to x3dh. Pretty sure we'll revisit these choices soon.

PR #1418, opened 2021-05-21
Sanjo/transport fixes

SanjoDeundiak:


spacekookie (approved):
In my PR #1400 I'm adding a function `Address::new(tt, inner)` that we should use instead of `format!(..).into()` but we can change that later.

Otherwise LGTM!

jared-s:
Rebased and merged

SanjoDeundiak:
FYI @adrianbenavides

PR #1421, opened 2021-05-25
Entity Feature

jared-s:
# Entity Feature Branch - Coordination PR

- Updates Examples 5, 6, 9, 12 to use Entities.
- Improves/experiments with Route building by use of `vec!`
- Secure Channel creation is now done through Entity
- Entity can list its active secure channels


implementations/rust/ockam/ockam_core/src/routing/route.rs
```diff
@@ -19,6 +19,15 @@ impl Route {
         RouteBuilder::new()
     }
 
+    /// Create a route from a Vec of addresses
+    pub fn create<T: Into<Address>>(vt: Vec<T>) -> Self {
```
SanjoDeundiak, line 23: Can we replace vector with slice?

implementations/rust/ockam/ockam/src/entity/remote.rs
```diff
@@ -0,0 +1,24 @@
+use crate::Route;
+
+use ockam_core::lib::Display;
+use serde::{Deserialize, Serialize};
+use std::fmt::{Formatter, Result};
+
+#[derive(Serialize, Deserialize, Debug, Clone, Hash, Ord, PartialOrd, Eq, PartialEq)]
+pub struct RemoteEntity {
+    pub route: Route,
```
SanjoDeundiak, line 9: Let's avoid using pub fields if possible

implementations/rust/ockam/ockam/src/entity/remote.rs
```diff
@@ -0,0 +1,24 @@
...
+}
+
+impl Display for RemoteEntity {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+        self.route.fmt(f)
+    }
+}
+
+impl RemoteEntity {
+    pub fn create<R: Into<Route>>(route: R) -> Self {
```
SanjoDeundiak, line 19: I'm sure we inconsistently use new vs create around the codebase, but don't remember what is the right choice here

implementations/rust/ockam/ockam/src/entity/local.rs
```diff
@@ -0,0 +1,219 @@
...
+};
+use ockam_core::{Address, AddressSet, Message, Result, Worker};
+use ockam_node::{Cancel, Context};
+use ockam_vault_core::{PublicKey, Secret};
+use ockam_vault_sync_core::Vault;
+
+use ockam_core::hashbrown::HashMap;
+
+pub struct LocalEntity {
+    pub ctx: Context,
```
SanjoDeundiak, line 15: I think it can be private

implementations/rust/ockam/ockam/src/entity/local.rs
```diff
@@ -0,0 +1,219 @@
...
+    vault: Address,
+    entity: Entity<ProfileSync>,
```
SanjoDeundiak, line 17: Do we want to rename ProfileSync -> EntitySync?

implementations/rust/ockam/ockam/src/entity/local.rs
```diff
@@ -0,0 +1,219 @@
...
+    secure_channels: HashMap<Address, Route>,
+}
+
+impl LocalEntity {
+    pub async fn create<A: Into<Address>>(node_ctx: &Context, address: A) -> Result<LocalEntity> {
+        let ctx = node_ctx.new_context(address).await?;
```
SanjoDeundiak, line 23: Can we just generate random address here?

implementations/rust/ockam/ockam/src/entity/local.rs
```diff
@@ -0,0 +1,219 @@
...
+use ockam_vault_core::{PublicKey, Secret};
+use ockam_vault_sync_core::Vault;
+
+use ockam_core::hashbrown::HashMap;
+
+pub struct LocalEntity {
+    pub ctx: Context,
+    vault: Address,
+    entity: Entity<ProfileSync>,
+    secure_channels: HashMap<Address, Route>,
```
SanjoDeundiak, line 18: It would be great to add some additional data to each channel in the future. And I'm not sure Route that is stored here is useful

implementations/rust/ockam/ockam/src/entity/local.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub fn list_secure_channels(&self) -> Result<Vec<(&Address, &Route)>> {
+        Ok(self.secure_channels.iter().collect())
+    }
+
+    pub async fn forward<A: Into<Address>>(
+        &mut self,
+        remote_entity: RemoteEntity,
+        service_address: A,
+    ) -> Result<RemoteForwarderInfo> {
+        let address = remote_entity.route.next().unwrap().to_string();
```
SanjoDeundiak, line 98: Too many assumptions that are not obvious from function API

implementations/rust/ockam/ockam/src/entity/local.rs
```diff
@@ -0,0 +1,219 @@
...
+        let address = address.strip_prefix("0#").unwrap(); // TODO how can we clean this up?
+        RemoteForwarder::create(&self.ctx, address, service_address.into()).await
+    }
+
+    pub async fn send<R, M>(&self, route: R, msg: M) -> Result<()>
```
SanjoDeundiak, line 103: Not sure why we need send&receive here

examples/rust/get_started/examples/12-secure-channel-over-cloud-node-forwarding-responder.rs
```diff
@@ -1,10 +1,10 @@
-use ockam::{Context, RemoteForwarder, Result, SecureChannel, TcpTransport, Vault};
+use ockam::{Context, LocalEntity, RemoteForwarder, Result, TcpTransport};
 use ockam_get_started::Echoer;
 
 #[ockam::node]
-async fn main(mut ctx: Context) -> Result<()> {
+async fn main(ctx: Context) -> Result<()> {
     // Create a cloud node by going to https://hub.ockam.network
-    let cloud_node_tcp_address = "Paste the tcp address of your cloud node here.";
+    let cloud_node_tcp_address = "40.78.99.34:4000"; //"Paste the tcp address of your cloud node here.";
```
SanjoDeundiak, line 7: Let's not forget to remove that

implementations/rust/ockam/ockam/src/entity/issuer.rs
```diff
@@ -0,0 +1,123 @@
+use crate::{
+    Address, CredentialAttribute, CredentialFragment2, CredentialIssuer, CredentialOffer,
+    CredentialRequest, CredentialSchema, OckamError, OfferIdBytes, ProfileTrait, ProofBytes,
+    PublicKeyBytes, Result, Route,
+};
+
+pub struct LocalIssuingEntity<E: ProfileTrait> {
```
SanjoDeundiak, line 7: Maybe we're ready to rename ProfileTrait -> EntityTrait

implementations/rust/ockam/ockam/src/entity/issuer.rs
```diff
@@ -0,0 +1,123 @@
...
+    address: Address,
+    credential_issuer: CredentialIssuer,
+    credential_schema: CredentialSchema,
+    _entity: E,
+    public_key: PublicKeyBytes,
```
SanjoDeundiak, line 12: How is PublicKeyBytes different from PublicKey?

examples/rust/get_started/examples/05-secure-channel.rs
```diff
@@ -1,29 +1,31 @@
...
+use ockam::{Address, Context, LocalEntity, Result, Route, SecureChannel, Vault};
 use ockam_get_started::Echoer;
 
 #[ockam::node]
 async fn main(mut ctx: Context) -> Result<()> {
     // Start an Echoer worker at address "echoer"
-    ctx.start_worker("echoer", Echoer).await?;
-
-    let vault = Vault::create(&ctx)?;
+    let mut local = LocalEntity::create_with_worker(&ctx, "echoer", Echoer).await?;
```
SanjoDeundiak, line 9: Why do we associate Echoer with entity?

SanjoDeundiak, line 9: Let's have consistent names, like initiator&responder/alice&bob/etc.

implementations/rust/ockam/ockam/src/entity/local.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub async fn secure_channel_listen_on_address<A: Into<Address>>(
+        &mut self,
+        address: A,
+    ) -> Result<()> {
+        self.entity
+            .create_secure_channel_listener(&self.ctx, address.into(), &self.vault)
+            .await
+    }
+
+    pub fn secure_channel_address(&self) -> String {
```
SanjoDeundiak, line 65: Is it used?

examples/rust/get_started/examples/09-secure-channel-over-many-transport-hops-initiator.rs
```diff
@@ -11,26 +11,18 @@ async fn main(mut ctx: Context) -> Result<()> {
     // Create a TCP connection.
     tcp.connect("127.0.0.1:3000").await?;
 
-    let vault = Vault::create(&ctx)?;
+    let mut initiator = LocalEntity::create(&ctx, "initiator").await?;
+    let middle: Address = (TCP, "127.0.0.1:3000").into();
+    let responder: Address = (TCP, "127.0.0.1:4000").into();
+    let route: Route = vec![middle, responder, "secure_channel_listener".into()].into();
```
SanjoDeundiak, line 17: Love that syntax. Would be even better if we could do that with slice (not sure it's possible)

implementations/rust/ockam/ockam_core/src/routing/route.rs
jared-s: Hmmm, I'll experiment. That would be cool.

implementations/rust/ockam/ockam/src/entity/remote.rs
jared-s: This one should probably be `new`. I think the distinction we are using is `new` are always simple constructors and `create` can take high level actions such as starting workers. Which this is not

implementations/rust/ockam/ockam/src/entity/local.rs
jared-s: @mrinalwadhwa ?

implementations/rust/ockam/ockam/src/entity/local.rs
jared-s: Yes, I suppose when/if they need to be named explicitly we can add it then.

implementations/rust/ockam/ockam/src/entity/local.rs
jared-s: This is pretty much only here for the proof of concept `list_secure_channels` API that we mentioned in the design call. Mrinal do you think this is useful at this point? Should we create some higher level type that encapsulates Route/whatever metadata?

implementations/rust/ockam/ockam/src/entity/local.rs
jared-s: You mean in regarding the entire `forward` function or this address parsing?  This is definitely a hack, mostly to work around our weird non-symmetric Address <-> String conversion

implementations/rust/ockam/ockam/src/entity/local.rs
jared-s: I'll remove - this was from before when moving the app Context instead of using `new_context`

implementations/rust/ockam/ockam/src/entity/issuer.rs
jared-s: @mrinalwadhwa do you think it's time?

implementations/rust/ockam/ockam/src/entity/issuer.rs
jared-s: `PublicKeyBytes` is a type for the raw slice key material. I added it a few months ago to begin cleaning up bare usages of arrays in return signatures like `[u8 : 96]`. It looks like `PublicKey` is similar but a heapless vec instead of a slice. The differences are originating from the credentials API, so we may need to work `PublicKey` into there more.

examples/rust/get_started/examples/05-secure-channel.rs
jared-s: To allow the entity to automatically create and start the worker. Experimental

jared-s: @mrinalwadhwa what do you think of this coupling?

jared-s: This was the original word from the guide. I agree we should be consistent but didn't want to change the original too much.

implementations/rust/ockam/ockam/src/entity/local.rs
jared-s: Nope old, removing thx

examples/rust/get_started/examples/05-secure-channel.rs
```diff
@@ -1,29 +1,37 @@
...
 
     // Send a message to the echoer worker, via the secure channel.
-    ctx.send(
-        // route to the "echoer" worker via the secure channel.
-        Route::new().append(channel.address()).append("echoer"),
-        // the message you want echo-ed back
-        "Hello Ockam!".to_string(),
-    )
-    .await?;
+    ctx.send(route, "Hello Ockam!".to_string()).await?;
```
jared-s, line 34: Can't put the Route vec in line here without some changes to `send`, due to the way the generics are handled. When one tries, the compiler says: `cannot infer type for type parameter R`. That is the `R: Into<Route>` trait bound. Seems like there should be enough compile time information here.

implementations/rust/ockam/ockam_core/src/routing/route.rs
jared-s: Couldn't figure out a way. I also couldn't get it working even with a static constructor, in a meaningful way. I can declare `fn create<T: Into<Address>>(slice: &[T])` but when one iterates over the slice it yields `&T` instead of `T` itself, so can't call '`into()`. Maybe could figure out something with more playing.

implementations/rust/ockam/ockam/src/entity.rs
```diff
@@ -1,6 +1,178 @@
...
+        &mut self,
+        address: A,
+        trust_policy: T,
+    ) -> Result<()> {
+        self.entity
+            .create_secure_channel_listener(&self.ctx, address.into(), trust_policy, &self.vault)
+            .await
+    }
+
+    pub async fn create_secure_channel_listener<T: TrustPolicy>(
```
SanjoDeundiak, line 45: Isn't it the same as function above?

implementations/rust/ockam/ockam/src/entity.rs
```diff
@@ -1,6 +1,178 @@
...
+        self.secure_channels.insert(channel.clone(), route);
+
+        Ok(channel)
+    }
+
+    pub fn list_secure_channels(&self) -> Result<Vec<(&Address, &Route)>> {
+        Ok(self.secure_channels.iter().collect())
+    }
+
+    pub async fn stop(&mut self) -> Result<()> {
```
SanjoDeundiak, line 74: How about moving it to Drop? What do you think?

implementations/rust/ockam/ockam/src/protocols/mod.rs
```diff
@@ -23,6 +23,7 @@ impl ProtocolPayload {
     /// In order to easily decode incoming `ProtocolPayload`s, it is
     /// recommended to use the `ProtocolParser` abstraction, which handles
     /// matching between different decoders based on the protocol ID.
+    #[allow(dead_code)]
```
SanjoDeundiak, line 26: I think protocol-related code shouldn't have landed in this PR

implementations/rust/ockam/ockam_channel/src/secure_channel_worker.rs
```diff
@@ -223,10 +223,10 @@ impl<V: SecureChannelVault, K: KeyExchanger + Send + 'static> SecureChannelWorke
         } else {
             return Err(SecureChannelError::InvalidInternalState.into());
         }
-        let _ = key_exchanger.process(payload.as_slice())?;
+        let _ = key_exchanger.handle_response(payload.as_slice())?;
```
SanjoDeundiak, line 226: This also should be merged to develop not using this PR

implementations/rust/ockam/ockam_entity/src/profile_set.rs
```diff
@@ -11,62 +11,125 @@ pub use identifiers::*;
 pub mod secrets;
 pub use secrets::*;
 
+use crate::EntityError::{InvalidInternalState, InvalidParameter};
+use ockam_core::hashbrown::hash_map::HashMap;
+
 /// An Entity represents an identity in various authentication contexts.
 #[derive(Clone)]
-pub struct Entity<P: ProfileTrait> {
+pub struct ProfileSet<P: ProfileTrait> {
```
SanjoDeundiak, line 19: Can we just replace generic with Profile type? If there are no other cases that would simplify things a lot

implementations/rust/ockam/ockam_entity/src/profile_set.rs
```diff
@@ -11,62 +11,125 @@ pub use identifiers::*;
...
+pub trait ProfileRetrieve<P: ProfileTrait> {
+    fn profile(&self, profile_identifier: &ProfileIdentifier) -> Option<&P>;
+}
+
+pub trait ProfileAdd<P: ProfileTrait> {
+    fn add_profile(&mut self, profile: P) -> Result<()>;
+}
+
+pub trait ProfileUpdate<P: ProfileTrait> {
+    fn update_profile(&mut self, old_profile_id: &ProfileIdentifier, profile: P) -> Result<()>;
```
SanjoDeundiak, line 33: 1. Profile identifier doesn't change (that's why I think there is no such thing as "old_profile_id")
2. I think we only update profile with new set of change events

implementations/rust/ockam/ockam_entity/src/profile_set.rs
```diff
@@ -11,62 +11,125 @@ pub use identifiers::*;
...
+}
+
+impl<P: ProfileTrait> ProfileAdd<P> for ProfileSet<P> {
+    fn add_profile(&mut self, profile: P) -> Result<()> {
+        if let Ok(id) = profile.identifier() {
+            if let Some(_) = self.profiles.insert(id, profile) {
+                return Ok(());
+            }
+        }
+        Err(InvalidInternalState.into())
```
SanjoDeundiak, line 77: Would be great to replace with something more user friendly.

implementations/rust/ockam/ockam_entity/src/profile_set.rs
```diff
@@ -11,62 +11,125 @@ pub use identifiers::*;
...
 #[derive(Clone)]
-pub struct Entity<P: ProfileTrait> {
+pub struct ProfileSet<P: ProfileTrait> {
     default_profile_identifier: ProfileIdentifier,
-    profiles: Vec<P>,
+    profiles: HashMap<ProfileIdentifier, P>,
+}
+
+pub trait ProfileRetrieve<P: ProfileTrait> {
+    fn profile(&self, profile_identifier: &ProfileIdentifier) -> Option<&P>;
```
SanjoDeundiak, line 25: I think because we'll be adding that to EntityWorker, return type should be Option<P> (or maybe Result<P>, or Result<Option<P>>)

implementations/rust/ockam/ockam_entity/src/profile_set.rs
```diff
@@ -11,62 +11,125 @@ pub use identifiers::*;
...
+            Ok(())
+        } else {
+            Err(InvalidInternalState.into())
+        }
+    }
+}
+
+impl<P: ProfileTrait> ProfileUpdate<P> for ProfileSet<P> {
+    fn update_profile(&mut self, old_profile_id: &ProfileIdentifier, profile: P) -> Result<()> {
+        self.remove_profile(&old_profile_id)?;
```
SanjoDeundiak, line 102: That should be definitely improved in the future by applying new changes to existing profile. But I think it's good for now, but let's have an issue or at least comment in the code

implementations/rust/ockam/ockam/src/protocols/mod.rs
jared-s: This is from develop rebase

implementations/rust/ockam/ockam_channel/src/secure_channel_worker.rs
jared-s: Also rebase

implementations/rust/ockam/ockam_entity/src/profile_set.rs
jared-s: I think so - trying today

implementations/rust/ockam/ockam_entity/src/profile_set.rs
jared-s: I was wondering if update was necessary at all - removing for now

implementations/rust/ockam/ockam_entity/src/profile_set.rs
jared-s: Ehh, given your other comments regarding change events, I think this function is premature. Going to remove.

implementations/rust/ockam/ockam_entity/src/profile_set.rs
jared-s: We don't want to move the `Profile` out of the `ProfileSet` (soon to be in Entity itself), do we? The profiles are stored in a `HashMap` by identifier. Typically we would `get` a reference to a contained value. Would we need to pull it out mutably? Since it's a worker, do we need to?

jared-s: I found use cases for both, but the signatures need more work.

PR #1423, opened 2021-05-25
feat(rust): entity refactor

jared-s:
- Roll back cloud_node_tcp_address copy
- Lower ctx viz in LocalEntity 
- Remove send and receive from LocalEntity
- Remove local entity secure_channel_address
- Remove forwarding from LocalEntity
- Remove RemoteEntity. Its only current use was a small forwarding example. Not fully baked, I don't think.
- LocalEntity randomize address for new context.
- Entity no longer starts a worker.
- `LocalEntity`  renamed to `Entity`.
- Example variables named `local` have been renamed to `alice` and `bob` as appropriate.
- Example secure channel name changed to `bob_secure_channel_listener`


jared-s:
@SanjoDeundiak FYI this is my first set of changes today

PR #1425, opened 2021-05-26
feat(rust): add route init using tuples

SanjoDeundiak:


examples/rust/get_started/examples/05-secure-channel.rs
```diff
@@ -27,11 +27,8 @@ async fn main(mut ctx: Context) -> Result<()> {
         .create_secure_channel("bob_secure_channel_listener", alice_trust_policy)
         .await?;
 
-    let echoer: Address = "echoer".into();
-    let route: Route = vec![channel_to_bob, echoer].into();
-
     // Send a message to the echoer worker, via the secure channel.
-    ctx.send(route, "Hello Ockam!".to_string()).await?;
+    ctx.send((channel_to_bob, "echoer"), "Hello Ockam!".to_string()).await?;
```
SanjoDeundiak, line 31: @mrinalwadhwa @jared-s thoughts?

implementations/rust/ockam/ockam_core/src/routing/route.rs
```diff
@@ -13,6 +13,30 @@ pub struct Route {
     inner: VecDeque<Address>,
 }
 
+impl<A: Into<Address>, B: Into<Address>> From<(A, B)> for Route {
```
SanjoDeundiak, line 16: It's ugly, but pretty common in such cases

mrinalwadhwa: Hmm, so as long as there are up to 5 hops in a route .. it can be a tuple? 
It looks really nice ... but feels too hacky

spacekookie: I think we can already create routes via `vec![...]`, because `Vec<Address>` implements `Into<Route>`, right? Otherwise we could have a `route!` macro which handles this conversion too (or then gives users the option to have something like this:

```rust
route![(TCP, "127.0.0.1:4000"), "my-friend"]
```

So tuples for (type, payload), or just strings if they're local addresses.

SanjoDeundiak: @spacekookie creation using vec is still there, but it's sometimes not convenient. Macro would be interesting though

jared-s: I've also started thinking a `route!` macro would probably be the cleanest for end users

jared-s: So what do we think about merging this feature? The explicit unrolling of each potential size is a bit hacky, but a pretty common occurrence in rust. It does improve the end user experience in my opinion.

SanjoDeundiak: @jared-s hold on, I'll try to work on macro soon

PR #1427, opened 2021-05-26
feat(rust): move and refactor entity

jared-s:
- Move `Entity` to ockam_entity (had been in ockam due to old dependencies)
- Entity `drop` now calls `stop` on the Entity Context.
- Removed `ProfileUpdate` trait - unclear API and need.
- De-generalized ProfileSet to use `ProfileSync` directly instead of `ProfileTrait`
- Merged `ProfileSet` into Entity.
- Reverted Example 5 to original copy, pre-Entity.
- Introduced Example 5a with Entities and mutual authentication.


jared-s:
cc @SanjoDeundiak @mrinalwadhwa

PR #1456, opened 2021-06-04
unit tests to increase test cover

mendrugory:
### Proposed Changes

Unit tests for Ockam.Hub.Service.Alias, Ockam.Hub.Service.Echo, Ockam.Hub.KafkaStreamHandler and Ockam.Hub.Web.
Related to this [issue](https://github.com/ockam-network/ockam/issues/1451).

### DIFF
--- a/implementations/elixir/ockam/ockam_hub/test/hub/service/alias_test.exs
+++ b/implementations/elixir/ockam/ockam_hub/test/hub/service/alias_test.exs
@@ -1,0 +1,85 @@
+defmodule Test.Hub.Service.AliasTest do
+  use ExUnit.Case
+
+  alias Ockam.Hub.Service.Alias, as: AliasService
+  alias Ockam.Hub.Service.Echo, as: EchoService
+  alias Ockam.Router
+  alias Test.Utils
+
+  test "alias test" do
+    {:ok, _worker, worker_address} =
+      Test.Hub.Service.AliasTestWorker.start_link(address: "alias_test_address")
+
+    {:ok, _alias, _alias_address} = AliasService.start_link(address: "alias_address")
+    {:ok, _echo, _echo_address} = EchoService.start_link(address: "echo_address")
+
+    msg = %{onward_route: [worker_address], return_route: [], payload: Utils.pid_to_string()}
+    Router.route(msg)
+
+    assert_receive(:ok, 5_000)
+  end
+end
+
+defmodule Test.Hub.Service.AliasTestWorker do
+  @moduledoc false
+
+  use Ockam.Worker
+
+  alias Ockam.Message
+  alias Ockam.Router
+  alias Test.Utils
+
+  @alias_address "alias_address"
+
+  @impl true
+  def handle_message(message, state) do
+    case Message.return_route(message) do
+      [] -> registration(message, state)
+      _other -> process(message, state)
+    end
+  end
+
+  defp registration(message, state) do
+    msg = %{
+      onward_route: [@alias_address],
+      return_route: [state.address]
+    }
+
+    Router.route(msg)
+
+    new_state =
+      state
+      |> Map.put(:test_process, Message.payload(message))
+      |> Map.put(:status, :registered)
+
+    {:ok, new_state}
+  end
+
+  defp process(message, state) when state.status == :registered do
+    msg = %{
+      onward_route: Message.return_route(message),
+      return_route: [state.address],
+      payload: "hello"
+    }
+
+    Router.route(msg)
+
+    new_state = Map.put(state, :status, :messaging)
+    {:ok, new_state}
+  end
+
+  defp process(message, state) when state.status == :messaging do
+    result =
+      case Message.payload(message) do
+        "hello" -> :ok
+        _other -> :error
+      end
+
+    state
+    |> Map.get(:test_process)
+    |> Utils.string_to_pid()
+    |> send(result)
+
+    {:ok, state}
+  end
+end
--- a/implementations/elixir/ockam/ockam_hub/test/hub/web/web_test.exs
+++ b/implementations/elixir/ockam/ockam_hub/test/hub/web/web_test.exs
@@ -1,0 +1,54 @@
+defmodule Test.Hub.Web.WebTest do
+  use ExUnit.Case
+  use Plug.Test
+
+  alias Ockam.Hub.Web.Router
+
+  @options Router.init([])
+
+  @http_status_bad_request 400
+  @http_status_unauthorized 401
+  @http_status_forbidden 403
+
+  test "create stream without authorization" do
+    conn =
+      :post
+      |> conn("/streams", %{"stream_prefix" => "prefix"})
+      |> Router.call(@options)
+
+    assert conn.status == @http_status_unauthorized
+  end
+
+  test "create stream with prefix" do
+    conn =
+      :post
+      |> conn(create_url("/streams"), %{"stream_prefix" => "prefix"})
+      |> Router.call(@options)
+
+    assert conn.status == @http_status_forbidden
+    assert conn.resp_body == "Kafka integration disabled"
+  end
+
+  test "create stream with wrong prefix" do
+    conn =
+      :post
+      |> conn(create_url("/streams"), %{"stream_pref" => "prefix"})
+      |> Router.call(@options)
+
+    assert conn.status == @http_status_bad_request
+  end
+
+  test "create stream without prefix" do
+    conn =
+      :post
+      |> conn(create_url("/streams"), %{})
+      |> Router.call(@options)
+
+    assert conn.status == @http_status_bad_request
+  end
+
+  defp create_url(endpoint) do
+    secret = Application.get_env(:ockam_hub, :auth_message)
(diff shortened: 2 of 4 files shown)


implementations/elixir/ockam/ockam_hub/test/hub/web/web_test.exs
```diff
@@ -0,0 +1,41 @@
...
+  test "create stream without authorization" do
+    conn = Router.call(conn(:post, "/streams", %{"stream_prefix" => "prefix"}), @options)
+    assert conn.status == @http_status_unauthorized
+  end
+
+  test "create stream with prefix" do
+    conn =
+      Router.call(conn(:post, create_url("/streams"), %{"stream_prefix" => "prefix"}), @options)
+
+    assert conn.status == @http_status_ok
```
hairyhum, line 22: Can you add response body check as well?

implementations/elixir/ockam/ockam_hub/test/hub/service/alias_test.exs
```diff
@@ -0,0 +1,71 @@
...
+  use ExUnit.Case
+
+  alias Ockam.Hub.Service.Alias, as: AliasService
+  alias Ockam.Hub.Service.Echo, as: EchoService
+
+  alias Ockam.Router
+
+  test "alias test" do
+    {:ok, _worker, worker_address} =
+      Test.Hub.Service.AliasTestWorker.start_link(address: {192, 168, 2, 1})
```
hairyhum, line 11: We usually use binaries for local addresses. Although it's possible to use tuples, this test might be confusing.

implementations/elixir/ockam/ockam_hub/test/hub/service/alias_test.exs
```diff
@@ -0,0 +1,71 @@
...
+
+    {:ok, _alias, _alias_address} = AliasService.start_link(address: {192, 168, 2, 2})
+    {:ok, _echo, _echo_address} = EchoService.start_link(address: {192, 168, 2, 3})
+
+    msg = %{onward_route: [worker_address], return_route: [], payload: self()}
```
hairyhum, line 16: Payload is expected to be binary, although it will not break the local routing if you try to use this code with transports, it will break.

hairyhum (changes requested):
Data types we were using for addresses and payload are not currently enforced, but they are implied in other code.
We didn't have a proper take on the data types in elixir so we're currently sticking with binaries for payloads and local addresses.

implementations/elixir/ockam/ockam_hub/test/hub/web/web_test.exs
mendrugory: The current response body is empty, because there is no kafka running. I do not know if for unit test we should have a kafka up and running. That seems to be more related to the integration tests.

Currently the returned status code is 200, if the action is creation, I think that we should change to a 201. What do you think?

hairyhum: Fair enough. Let's shift it to the integration test. Also the http codes are not ideal, so here is the update: https://github.com/ockam-network/ockam/commit/96971795d3cd280f57fbac2fc848a3a488e6a86d

implementations/elixir/ockam/ockam_hub/test/hub/service/alias_test.exs
```diff
@@ -0,0 +1,65 @@
...
+      [] -> registration(message, state)
+      _other -> process(message, state)
+    end
+  end
+
+  defp registration(message, state) do
+    msg = %{
+      onward_route: [@alias_address],
+      return_route: [state.address],
+      payload: %{registration_payload: Message.payload(message), forward_route: @echo_address}
```
hairyhum, line 47: What's the purpose of the `forward_route` here?

implementations/elixir/ockam/ockam_hub/test/hub/service/alias_test.exs
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    Router.route(msg)
+
+    new_state = Map.put(state, :status, :registered)
+    {:ok, new_state}
+  end
+
+  defp process(message, state) when state.status == :registered do
```
hairyhum, line 56: The forwarding should also have the third step where you send a message to an alias.

hairyhum (commented):
I'm not sure I can follow the logic of the alias test.

implementations/elixir/ockam/ockam_hub/test/hub/service/alias_test.exs
mendrugory: According to the docs, `:registration_payload` will be send after registration to `:forward_route`. This behaviour is check out when the status of the alias test is is `:registered`. I have also added a `:messaging` status in order to send message after registration.

hairyhum: `registration_payload` and `forward_route` are not in the message payload. They supposed to be in the options then the `Ockam.Hub.Service.Alias.Forwarder` worker is created by the alias service worker. It's an internal API for the alias service.
`registration_payload` is the payload of the registration message, while `forward_route` is the return route of the message. https://github.com/ockam-network/ockam/blob/develop/implementations/elixir/ockam/ockam_hub/lib/hub/service/alias.ex#L27

A worker cannot register a different worker in the alias service ATM, it can only register itself and it receives a registration response message from the forwarder worker. The return route of this response message can be used to send messages to the registered worker through the forwarder.

We can think of the better API for that when we do a proper "registry service", but right now that's how it works.

What do you think would be the best format to document that? Function docs? Process diagram in the moduledoc? Any other ideas on how to make it clear?
Also this test would be one of the reference points to that.

mendrugory: Thank you for the clarification. Let me continue with this test.

Although function docs are encouraged by a lot of people, I personally prefer to have as much information as possible in the documentation, so `moduledoc` would be better from my point of view.

implementations/elixir/ockam/ockam_hub/test/hub/service/alias_test.exs
```diff
@@ -0,0 +1,87 @@
...
+  @moduledoc false
+
+  use Ockam.Worker
+
+  alias Ockam.Message
+  alias Ockam.Router
+  alias Test.Utils
+
+  @echo_address "ddf373h47asdf"
+  @alias_address "cdf373h48asdf"
```
hairyhum, line 33: We can also use meaningful names in tests, like "alias_address" or "alias_test_service_address"
Also is `@echo_address` still needed? It doesn't seem to be doing anything.

implementations/elixir/ockam/ockam_hub/test/hub/service/alias_test.exs
```diff
@@ -0,0 +1,87 @@
...
+  defp process(message, state) when state.status == :messaging do
+    result =
+      case Message.payload(message) do
+        "hello" -> :ok
+        _other -> :error
+      end
+
+    state
+    |> Map.get(:test_process)
+    |> Message.payload()
```
hairyhum, line 81: Is this necessary?

implementations/elixir/ockam/ockam_hub/test/hub/service/alias_test.exs
```diff
@@ -0,0 +1,87 @@
...
+  alias Ockam.Hub.Service.Echo, as: EchoService
+  alias Ockam.Router
+  alias Test.Utils
+
+  test "alias test" do
+    {:ok, _worker, worker_address} =
+      Test.Hub.Service.AliasTestWorker.start_link(address: "bdf373h28asdf")
+
+    {:ok, _alias, _alias_address} = AliasService.start_link(address: "cdf373h48asdf")
+    {:ok, _echo, _echo_address} = EchoService.start_link(address: "ddf373h47asdf")
```
hairyhum, line 14: i don't remember, but does the test terminate linked processes? Maybe we need an explicit cleanup?

implementations/elixir/ockam/ockam_hub/test/hub/service/alias_test.exs
mendrugory: This step is to send the `:ok` to the actual test.

implementations/elixir/ockam/ockam_hub/test/hub/service/alias_test.exs
mendrugory: As the process is linked to the test process, it should be terminated.

implementations/elixir/ockam/ockam_hub/test/hub/service/alias_test.exs
hairyhum: But shouldn't `:test_process` key already contain the payload? On line 52: `Map.put(:test_process, Message.payload(message))`

mendrugory: It was a leftover of the previous version

hairyhum (approved):

PR #1473, opened 2021-06-09
feat(elixir): a declarative services creation on hub node

hairyhum:
Goal:
Experimenting on defining a set of configured services to start when a node starts.
Trying to make it easier for the infra code to deal with add-ons and their configuration.

Changes:
Make it possible to specify in the application config which services
to start with which parameters.

Allow hub addons to provide their services via Ockam.Hub.Service.Provider behaviour


implementations/elixir/ockam/ockam_hub/config/runtime.exs
```diff
@@ -155,3 +155,27 @@ config :ockam_kafka,
     client_config: kafka_client_config,
     replication_factor: kafka_replication_factor
   ]
+
+## Services config
+
+## This may be default of on-demand nodes
+kafka_service_prefix = System.get_env("KAFKA_SERVICE_PREFIX") || ""
+
+config :ockam_hub,
```
hairyhum, line 164: The idea here is that `service_providers` are defined per add-on artifact (e.g. ockam-node-kafka docker image)
while `services` is provided by the user (via hub-console) and declaratively defines the services (with arguments) to run on the node.

implementations/elixir/ockam/ockam_hub/config/runtime.exs
```diff
@@ -155,3 +155,27 @@ config :ockam_kafka,
...
+    Ockam.Kafka.Hub.Service.Provider
+  ],
+  services: [
+    :echo,
+    :forwarding,
+    :stream,
+    ## same as `:stream_index`
+    {:stream_index, []},
+    {:stream_kafka, [prefix: kafka_service_prefix, user: kafka_user]},
+    {:stream_index_kafka, [prefix: kafka_service_prefix, user: kafka_user]}
```
hairyhum, line 180: `kafka_user` here might be used to pass the user credentials from the console

implementations/elixir/ockam/ockam_hub/config/runtime.exs
mendrugory: How would be the service definition procedure from ockam hub node perspective?

hairyhum: The options I though were environemnt variable or a file in some format (e.g. json). 
This may be accessed from `Ockam.Hub.Service.Provider`, does not have to be from `runtime.exs`.
A file makes it easier to deal with secrets since we can encrypt it.

mendrugory:
From my point of view, using `SERVICES_CONFIG_SOURCE=file` is a cleaner approach when the configuration is complex. That involves two steps, setting up a file with the necessary information and point it using the env variable, instead of using just the env variables, but I think that it is better.
I would just use env variables if I can identify the inner info of the json file with final env variables (i.e. `SERVICE_USERNAME`, `STREAM_NAMES` separated by commas, ...)

hairyhum:
@mendrugory can you elaborate what do you mean by "I would just use env variables if I can identify the inner info of the json file with final env variables (i.e. SERVICE_USERNAME, STREAM_NAMES separated by commas, ...)"

mendrugory:
I just meant about "I would just use env variables if I can identify the inner info of the json file with final env variables (i.e. SERVICE_USERNAME, STREAM_NAMES separated by commas, ...)" that I like to use env variables when the variable is just a value (name, integer) but I do not personally like to use them, if I can avoid it, to use it for complex data structure (lists, maps, ...).

hairyhum:
Added a new config source `list`, which is passed as `SERVICES_LIST` variable. It can be used to specify comma separated string of services to run assuming default configuration for all of them:
```
SERVICES_LIST="echo,forwarding,stream"
## the same as
SERVICES_JSON='{"echo":{}, "forwarding":{}, "stream": {}}'
```

Whitespace between service names is ignored: `SERVICES_LIST="echo, forwarding, stream"` is valid

mendrugory (approved):

PR #1478, opened 2021-06-14
Credentials for Entity Worker and Profile

jared-s:
# Changes

## Credential code move
- Moved credential code from `ockam` to `ockam_entity`
	- We want a `Profile` to be able to assume any of the 3 credential roles (`Issuer`, `Holder`, and `Verifier`)
	- We need to add some state in order to achieve this, not just implementing trait behavior.
		- Signing public and private keys (field scalars), issued credentials, offer info
- The alternative seemed to be yet another crate such as `ockam_credential` which has a dependency on `signature_core` and at least one `signature_*`, as well as the credential 'role' implementations for issuer, holder and verifier. This might still be the cleanest approach.
	

## Holder, Issuer, Verifier 
	
- Traits defined for these roles
- Implemented trait for ProfileImpl, ProfileWorker, Entity

# Current Issues

- BARE related encoding crash in `Holder::accept_credential_offer` that I'm tracking down.
- Working through validating other API calls.

# Thoughts
- The signature and credential types are different than our current Vault types
	- Different kinds of public and private keys
	- Lots of smaller material: identifiers, offers, credentials, fragments. These will vary depending on signature scheme.
- Are we going to launch with Credentials tied to a specific signature crate?
	- Current implementation is tightly coupled to BBS+
	- How can we generalize between them? e.g. "Fragment 1 and 2" are algo specific.


jared-s:
@mrinalwadhwa @SanjoDeundiak FYI the state of Credentials integration to Profiles. Implementation is complete, but there are some serialization bugs to work out now, I think between serde_bare and the BigArray custom serializer. Working through that next.

implementations/rust/ockam/ockam_entity/src/worker/profile_sync.rs
```diff
@@ -11,7 +11,7 @@ use tracing::debug;
 
 /// Synchronous worker wrapper around a Profile
 pub struct ProfileSync {
-    ctx: Context,
+    pub(crate) ctx: Context,
```
SanjoDeundiak, line 14: Can we have getter instead?

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
```diff
@@ -0,0 +1,98 @@
...
+        presentation_manifests: &[PresentationManifest],
+        proof_request_id: ProofRequestId,
+        rng: impl RngCore + CryptoRng,
+    ) -> Result<Vec<CredentialPresentation>>;
+}
+
+/// Credential Verifier
+pub trait Verifier {
+    /// Create a unique proof request id so the holder must create a fresh proof
+    fn create_proof_request_id(&mut self, rng: impl RngCore) -> Result<ProofRequestId>;
```
SanjoDeundiak, line 82: Should it be + CryptoRng?

implementations/rust/ockam/ockam_entity/src/entity.rs
```diff
@@ -105,7 +110,7 @@ impl Entity {
 impl ProfileAdd for Entity {
     fn add_profile(&mut self, profile: ProfileSync) -> Result<()> {
         if let Ok(id) = profile.identifier() {
-            if let Some(_) = self.profiles.insert(id, profile) {
+            if self.profiles.insert(id, profile).is_some() {
```
SanjoDeundiak, line 113: Did you consider writing such things like this?
```rs
let id = profile.identifier().or(InvalidInternalState.into())?;
let _ = self.profiles.insert(id, profile).ok_or_else(InvalidInternalState.into())?;

Ok(())
```
Do you think it looks cleaner?

SanjoDeundiak: The downside is duplicating Errors, but you reduce code nesting, which is nice

implementations/rust/ockam/ockam_entity/src/entity.rs
```diff
@@ -306,3 +311,159 @@ impl ProfileAuth for Entity {
         }
     }
 }
+
+impl Issuer for Entity {
+    fn get_signing_key(&mut self) -> Result<SigningKeyBytes> {
+        if let Some(profile) = self.default_profile_mut() {
```
SanjoDeundiak, line 317: Same here, I would prefer ok_or way to do that to have less nesting, but wonder what do you think about that

implementations/rust/ockam/ockam_entity/src/entity.rs
jared-s: Yeah I think I like this more, the nesting is pretty annoying

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
jared-s: Maybe - I'll research, this is from Mike's original implementation

implementations/rust/ockam/ockam_entity/src/worker/profile_sync.rs
jared-s: Done. It would be cool if we could somehow generalize / reuse the pattern of 

```rust
block_future(&self.ctx().runtime(), async move {...})
```

To something where we can just pass a closure, like on_runtime(async move {})...
That's tricky for several reasons, I played with it a bit and will have to wait for a later time

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
jared-s: This dependency is trickling up from [ff::Field::random](https://docs.rs/ff/0.10.0/ff/trait.Field.html#tymethod.random) which is used by BLS Scalar class. `fill_bytes` is the only API used

SanjoDeundiak: CryptoRng is an empty marker trait, used to indicate that underlying implementation is safe to use for cryptography. I would say we should not even bother checking when it's safe to use Rng without CryptoRng constraint, since that's tricky. Let's require it everywhere

jared-s: SGTM

implementations/rust/ockam/ockam_entity/src/worker/profile_sync.rs
SanjoDeundiak: I think I tried to do that too, and it was tricky

jared-s:
# Update - BARE serialization issues

I spent today tracking down problems with serialization of the data structures related to credentials. 

- Currently, serialization blows up in Message decoding due to an expectation mismatch in types
- The heart of the problem is that the `signature_core` and `signature_bbs_plus` crates are `no_std` and use things like `heapless::Vec`. This has pushed these heapless structs up to the higher levels.
- There exists some thin serde `Serialize` and `Deserialize` implementations for these types, but these serializes are not compatible with how we are doing BARE encoding.

Through process of elimination, I've narrowed down the conflicts to:
- the `bls12_381_plus::G1Projective` type being exposed all of the way up. This has a custom `Serialize` and `Deserialize` implementation, but also appears to be incompatible with our encoding scheme and crashes. I've temporarily hacked around this by externalizing it to a byte array, and then reconstructing post serialization.
- custom serializers used in the signature code such as `VecSerializer`. I'm still working on this one.

jared-s:
# It's alive!

Working example with the current API [is here](https://github.com/ockam-network/ockam/blob/jared/credential_traits/implementations/rust/ockam/ockam_examples/example_projects/credentials/src/main.rs).

- Still very rough, lots to rename and clean up. 
- A number of types involved in Credentials currently have temporary, externalized types defined.
- These new types begin with `Ext` and then the name of its dual internal type.
- `From` is defined both ways between the `Ext` and internal type, to allow easy conversion between the two.
- The `Ext` types are used for message serialization in worker request/response types.
- The `Ext` types are then converted back to their internal types at point of use.

The conflict with serialization and the credential types mostly stems from the `heapless` types, and the existing custom `Serialize` and `Deserialize` implementations in the `signature_*` crates _as well as_ types from lower level BLS related crates.

The real solution is probably an abstraction layer that prevents these externals from being exposed, combined with a higher level marshaling strategy that can deal with `heapless` types in a simpler fashion.

jared-s:
@SanjoDeundiak this PR now has a full implementation of Issuer, Holder and Verifier using the new entity system. Still using the `Ext` type hacks, and in general more to do in the serialized credential types, but I didn't want this to grow any bigger.

implementations/rust/ockam/ockam_entity/src/profile_state.rs
```diff
@@ -21,6 +33,8 @@ pub struct ProfileState {
     contacts: Contacts,
     vault: VaultSync,
     key_attribs: KeyAttributes,
+    signing_key: BbsSecretKey,
```
SanjoDeundiak, line 36: This is what I'm trying to get_rid in my PR. Should we merge this first and then mine, or just remove this temporary so I could revive it properly?

implementations/rust/ockam/ockam_vault_core/src/types.rs
```diff
@@ -116,6 +116,16 @@ pub struct SecretAttributes {
     length: usize,
 }
 
+impl Default for SecretAttributes {
```
SanjoDeundiak, line 119: Some kind of merge problem here?

SanjoDeundiak (approved):

PR #1493, opened 2021-06-21
feat(rust): standardize all crates to tokio 1.7

jared-s:


jared-s:
Since CI isn't configured to kick off with just these toml changes, I'm going to address this upgrade as a part of the release cycle later today.

mrinalwadhwa:
it should've triggered
https://github.com/ockam-network/ockam/blob/develop/.github/workflows/rust_build_ockam_ockam_node.yml#L12

I noticed some delay in actions triggering over the weekend.

jared-s:
Oh interesting - thanks - I'll keep an eye out

PR #1505, opened 2021-06-25
feat(rust): entity worker

jared-s:
# Entity / Profile Refactor


examples/rust/get_started/examples/06-secure-channel-many-hops.rs
```diff
@@ -8,22 +8,23 @@ async fn main(mut ctx: Context) -> Result<()> {
...
-    let mut bob = Entity::create(&ctx).await?;
+    let mut bob = Entity::create(&ctx)?;
 
     // Start 3 hop workers at addresses "h1", "h2" and "h3".
     ctx.start_worker("h1", Hop).await?;
     ctx.start_worker("h2", Hop).await?;
     ctx.start_worker("h3", Hop).await?;
 
-    // Create a secure channel listener at address "secure_channel_listener"
+    // TODO JDS Secure Channel integration point
```
SanjoDeundiak, line 18: Do we want to fix that before merging to develop?

implementations/rust/ockam/ockam/src/protocols/stream/requests.rs
```diff
@@ -10,7 +10,8 @@ pub struct CreateStreamRequest {
 }
 
 impl CreateStreamRequest {
-    #[allow(clippy::new_ret_no_self)]
+    //noinspection ALL
+    #[allow(dead_code, clippy::new_ret_no_self)]
```
SanjoDeundiak, line 14: I guess this function is not exposed from the library. Should we fix that?

implementations/rust/ockam/ockam_vault_core/src/types.rs
```diff
@@ -110,6 +116,16 @@ pub struct SecretAttributes {
     length: usize,
 }
 
+impl Default for SecretAttributes {
```
SanjoDeundiak, line 119: I'm not sure we should ever let somebody create such structs without explicitly choosing all parameters

implementations/rust/ockam/ockam_node/src/node.rs
```diff
@@ -10,7 +10,7 @@ pub struct NullWorker;
 
 impl NullWorker {
     /// Create and register a new NullWorker context
-    pub(crate) fn new(rt: Arc<Runtime>, addr: &Address, tx: Sender<NodeMessage>) -> Context {
+    pub fn new(rt: Arc<Runtime>, addr: &Address, tx: Sender<NodeMessage>) -> Context {
```
SanjoDeundiak, line 13: Is there purpose for such worker in public API?

implementations/rust/ockam/ockam_entity/src/profile.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub const PROFILE_UPDATE: &'static str = "OCKAM_PUK";
+    /// Label for key used to issue credentials
+    pub const CREDENTIALS_ISSUE: &'static str = "OCKAM_CIK";
+    /// Current version of change structure
+    pub const CURRENT_CHANGE_VERSION: u8 = 1;
+}
+
+impl Identity for Profile {
+    fn identifier(&mut self) -> Result<ProfileIdentifier> {
+        self.entity().identifier()
```
SanjoDeundiak, line 61: Don't really understand why would Profile delegate to Entity and not vice versa

implementations/rust/ockam/ockam_entity/src/traits.rs
```diff
@@ -1,140 +1,65 @@
...
-/// Profile identity.
-pub trait ProfileIdentity {
+pub type Proof = Vec<u8>;
+
+/// Identity
+pub trait Identity: Send + 'static {
     /// Return unique [`Profile`] identifier, which is equal to sha256 of the root public key
-    fn identifier(&self) -> Result<ProfileIdentifier>;
-}
+    fn identifier(&mut self) -> Result<ProfileIdentifier>;
```
SanjoDeundiak, line 10: I think part of the function should take immutable self, including this fn

implementations/rust/ockam/ockam_entity/src/worker/create_key.rs
```diff
@@ -69,23 +70,19 @@ impl CreateKeyChange {
...
-        let attributes = attributes.unwrap_or_default();
-
-        // TODO: Should be customisable
-        let secret_attributes = SecretAttributes::new(
-            SecretType::Curve25519,
-            SecretPersistence::Persistent,
-            CURVE25519_SECRET_LENGTH,
-        );
+        let secret_attributes = match key_attributes.meta() {
+            MetaKeyAttributes::SecretAttributes(secret_attributes) => *secret_attributes,
```
SanjoDeundiak, line 83: I'm not sure this the mechanism we use for adjusting secret type

implementations/rust/ockam/ockam_entity/src/worker/create_key.rs
```diff
@@ -139,17 +136,23 @@ impl<V: ProfileVault> ProfileImpl<V> {
         )
         .is_ok()
         {
-            return Err(EntityError::InvalidInternalState.into());
+            panic!("attempted to recreate the same key")
```
SanjoDeundiak, line 139: Why panic?

implementations/rust/ockam/ockam_entity/src/worker/entity_worker.rs
```diff
@@ -0,0 +1,164 @@
...
+use ockam_core::{
+    lib::{result::Result::Ok, HashMap},
+    Result, Routed, Worker,
+};
+use ockam_node::Context;
+use ockam_vault_sync_core::{Vault, VaultSync};
+
+#[derive(Default)]
+pub struct EntityWorker {
+    profiles: HashMap<ProfileIdentifier, ProfileState>,
```
SanjoDeundiak, line 16: Why do we need ProfileState? Why not just Profile?

implementations/rust/ockam/ockam_entity/src/lib.rs
```diff
@@ -8,39 +8,86 @@
...
+pub use worker::*;
+
+use crate::EntityError;
+
+pub struct ContextAddress {
+    ctx: Context,
+    address: Address,
+}
+
+pub type Handle = Arc<Mutex<ContextAddress>>;
```
SanjoDeundiak, line 38: Not sure why we should use Arc and/or Mutex here

implementations/rust/ockam/ockam_entity/src/lib.rs
```diff
@@ -8,39 +8,86 @@
...
+            &self.ctx.runtime(),
+            async move { self.async_cast(msg).await },
+        )
+    }
+
+    pub async fn async_call<I: Message + Send + 'static, O: Message + Send + 'static>(
+        &mut self,
+        msg: I,
+    ) -> Result<O> {
+        self.ctx.send(self.address.clone(), msg).await?;
```
SanjoDeundiak, line 60: There is probability we need to spawn new ctx here each time

implementations/rust/ockam/ockam_entity/src/lib.rs
```diff
@@ -8,39 +8,86 @@
...
+        let msg = self.ctx.receive::<O>().await?;
+        Ok(msg.take().body())
+    }
+
+    pub fn call<I: Message + Send + 'static, O: Message + Send + 'static>(
```
SanjoDeundiak, line 65: API looks great

implementations/rust/ockam/ockam_entity/src/profile.rs
```diff
@@ -0,0 +1,135 @@
...
+/// her key and sends corresponding [`Profile`] changes to Bob.
+/// TODO
+///
+use crate::{
+    Changes, Contact, Entity, Handle, Identity, ProfileChangeEvent, ProfileIdentifier, Proof,
+};
+use ockam_core::Result;
+use ockam_vault::{PublicKey, Secret};
+
+pub struct Profile {
```
SanjoDeundiak, line 26: I'm not sure we need such struct

implementations/rust/ockam/ockam_entity/src/entity.rs
```diff
@@ -1,308 +1,236 @@
...
     }
+}
 
-    fn update_no_verification(&mut self, change_event: ProfileChangeEvent) -> Result<()> {
-        if let Some(profile) = self.default_profile_mut() {
-            profile.update_no_verification(change_event)
-        } else {
-            Err(ProfileNotFound.into())
-        }
+impl Identity for Entity {
```
SanjoDeundiak, line 91: Sounds like a slogan. "Identity for Entity"

SanjoDeundiak (changes requested):
A lot of great changes. Need to settle up some details

examples/rust/get_started/examples/06-secure-channel-many-hops.rs
jared-s: Definitely, I'll go through and fix all `TODO JDS` first

implementations/rust/ockam/ockam/src/protocols/stream/requests.rs
jared-s: Yes, will expose

implementations/rust/ockam/ockam_vault_core/src/types.rs
jared-s: Fair enough, yeah. Will remove Default impl

implementations/rust/ockam/ockam_node/src/node.rs
jared-s: Reverting this, was an experiment

implementations/rust/ockam/ockam_entity/src/profile.rs
jared-s: Let's go over the full architecture with Mrinal soon

implementations/rust/ockam/ockam_entity/src/traits.rs
jared-s: I'll work on this

implementations/rust/ockam/ockam_entity/src/worker/create_key.rs
jared-s: Notes from our conversation: Talking about using labels and different kinds of secrets. Users will not not when they will need to create things like BLS secret vs other secret. Consider APIs using predefined label such as "signing key" and API maybe like `create_signing_key` etc for all of the various kinds of secrets.

implementations/rust/ockam/ockam_entity/src/worker/create_key.rs
jared-s: Just debugging, will revert

implementations/rust/ockam/ockam_entity/src/worker/entity_worker.rs
jared-s: We talked about this a bit..lowest level state vs high level handle, in this case. Maybe we can collapse it further

implementations/rust/ockam/ockam_entity/src/lib.rs
jared-s: This is to allow a Profile and an Entity to share the context. From call notes: Change to `clone`

implementations/rust/ockam/ockam_entity/src/lib.rs
jared-s: Interesting, will look at it

implementations/rust/ockam/ockam_entity/src/lib.rs
jared-s: Was able to clean this up with your new_context suggestion:

```rust
pub struct Handle {
    ctx: Context,
    address: Address,
}

impl Clone for Handle {
    fn clone(&self) -> Self {
        block_future(&self.ctx.runtime(), async move {
            Handle {
                ctx: self
                    .ctx
                    .new_context(Address::random(0))
                    .await
                    .expect("new_context failed"),
                address: self.address.clone(),
            }
        })
    }
}```

File: implementations/rust/ockam/ockam_entity/src/profile.rs (hunk 5 above)
Comment (jared-s) [reply implementations/rust/ockam/ockam_entity/src/profile.rs:63]: Discussion is larger than this PR since we wanna merge it in the morning. Noting to talk about it later

File: implementations/rust/ockam/ockam_entity/src/traits.rs (hunk 6 above)
Comment (jared-s) [reply implementations/rust/ockam/ockam_entity/src/traits.rs]: Reduced mutability everywhere possible. The remaining ones are due to trickling up mutable vault requirements. However, I'm thinking it might be an anti-pattern to have the same traits at the Public API and internal APIs. It is hamstringing us in a lot of situations I think..So for example, the `Handler` APIs should always be able to be immutable since they are just all messaging.

File: implementations/rust/ockam/ockam_entity/src/lib.rs (hunk 11 above)
Comment (jared-s) [reply implementations/rust/ockam/ockam_entity/src/lib.rs]: Implemented

**Comment by jared-s** [comment]
@SanjoDeundiak ready for another look. To finish the secure channel integration, I created sync wrappers, simpler public API and request/response model for the creation of listener and channel.

File: examples/rust/get_started/examples/12-secure-channel-over-cloud-node-forwarding-initiator.rs [hunk 15]
```diff
@@ -33,6 +31,6 @@ async fn main(mut ctx: Context) -> Result<()> {
     let reply = ctx.receive::<String>().await?;
     println!("App Received: {}", reply); // should print "Hello Ockam!"
 
-    // Stop all workers, stop the node, cleanup and return.
+    // Stop all workers, stop the node, cleanup and return.*/
```
SanjoDeundiak, line 34: remove trailing */

examples/rust/get_started/examples/09-secure-channel-over-many-transport-hops-initiator.rs
```diff
@@ -26,8 +26,7 @@ async fn main(mut ctx: Context) -> Result<()> {
 
     // Wait to receive a reply and print it.
     let reply = ctx.receive::<String>().await?;
-    println!("App Received: {}", reply); // should print "Hello Ockam!"
-
-    // Stop all workers, stop the node, cleanup and return.
+    println!("App Received: {}", reply); // should print "Hello Ockam!"*/
```
SanjoDeundiak, line 29: remove trailing */

examples/rust/get_started/examples/06-secure-channel-many-hops.rs
```diff
@@ -42,6 +41,6 @@ async fn main(mut ctx: Context) -> Result<()> {
     let reply = ctx.receive::<String>().await?;
     println!("App Received: {}", reply); // should print "Hello Ockam!"
 
-    // Stop all workers, stop the node, cleanup and return.
+    // Stop all workers, stop the node, cleanup and return.*/
```
SanjoDeundiak, line 44: +1

implementations/rust/ockam/ockam_entity/src/lib.rs
```diff
@@ -8,39 +8,102 @@
...
+use ockam_core::{lib::HashMap, Address, Message, Result};
+use ockam_node::{block_future, Context};
+use ockam_vault::{Hasher, KeyIdVault, SecretVault, Signer, Verifier};
+pub use profile_state::*;
+pub use traits::*;
+pub use worker::*;
+
+use crate::EntityError;
+
+pub struct Handle {
```
SanjoDeundiak, line 31: Consider moving this to ockam_node (in separate PR maybe). It would be ideally to place it in ockam_core, but I believe it's not possible until we have Context/Node traits

SanjoDeundiak (approved):

PR #1507, opened 2021-06-28
fix(elixir): remove deduplication from stream client

hairyhum:
Deduplication in the stream client is not reliable enough.
Since secure channel do not require deduplication, it's not critically needed

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/stream/client/bi_directional.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/stream/client/bi_directional.ex
@@ -42,8 +42,8 @@ defmodule Ockam.Stream.Client.BiDirectional do
   Creates a return stream publisher id it doesn't exist
   Routes the message locally with return publisher address in return route
   """
-  def handle_message(data, consumer_stream, subscription_id, stream_options, state) do
-    with {:ok, %{return_stream: publisher_stream, message: message, message_id: message_id}} <-
+  def handle_message(data, consumer_stream, subscription_id, stream_options, _state) do
+    with {:ok, %{return_stream: publisher_stream, message: message}} <-
            decode_message(data) do
       {:ok, publisher_address} =
         ensure_publisher(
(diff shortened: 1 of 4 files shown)


implementations/elixir/ockam/ockam/lib/ockam/stream/client/bi_directional.ex
```diff
@@ -99,22 +75,22 @@ defmodule Ockam.Stream.Client.BiDirectional do
...
-      %{return_stream: stream, message: wire_message, message_id: message_id},
+      %{return_stream: stream, message: wire_message},
       @bare_message
     )
   end
 
   def decode_message(data) do
     case :bare.decode(data, @bare_message) do
-      {:ok, %{return_stream: stream, message: wire_message, message_id: message_id}, ""} ->
+      {:ok, %{return_stream: stream, message: wire_message}, ""} ->
```
mendrugory, line 91: I think that it could be good to use with statement just to re-act to non-expected values.

implementations/elixir/ockam/ockam/lib/ockam/stream/client/bi_directional.ex
```diff
@@ -99,22 +75,22 @@ defmodule Ockam.Stream.Client.BiDirectional do
...
-  @bare_message {:struct, [return_stream: :string, message: :data, message_id: :uint]}
+  @bare_message {:struct, [return_stream: :string, message: :data]}
 
-  def encode_message(%{return_stream: stream, message: message, message_id: message_id}) do
+  def encode_message(%{return_stream: stream, message: message}) do
     {:ok, wire_message} = Ockam.Wire.encode(@transport_message_encoder, message)
 
     :bare.encode(
-      %{return_stream: stream, message: wire_message, message_id: message_id},
+      %{return_stream: stream, message: wire_message},
```
mendrugory, line 84: I think that it could be good to use with statement just to re-act to non-expected values.

hairyhum: I'm not sure I understand what you mean. Like using the function argument here to make it crash on encode instead of function clause?

mendrugory: I was just thinking about something like:

```elixir
with {:ok, wire_message} <- Ockam.Wire.encode(@transport_message_encoder, message) do
    :bare.encode(
      %{return_stream: stream, message: wire_message, message_id: message_id}
    ...
```

or maybe a `case`
just to ensure the values.

mendrugory (approved):

PR #1521, opened 2021-07-02
WIP feat(rust): add no_std + alloc support to enable embedded device support for ockam

antoinevg:
### Proposed Changes

* rework `ockam::lib` to `ockam::compat`
  - mirror `std::` and `alloc::` module structures
  - add `no_std` versions of missing `rand::` functionality
  - remove references to items that live in `core::`
* update all ockam modules to use `ockam::compat` instead of `ockam::lib`
* add `no_std` and `alloc` features to all ockam crates
  - populate all crate dependencies to `std`, `no_std` and `alloc`
  - update all crate `#[cfg(feature=...)]` attributes
* implementa `no_std` version of `ockam_node_attribute`
* add placeholder semihosting logging macros for embedded targets
* add placeholder `no_std` embedded `Mutex` implementation for `ockam_vault_sync`
* implement `no_std` versions of tokio functionality used by `ockam_node`:
  - `tokio::sync::mpsc`
  - `tokio::runtime`
  - `tokio::task`
  - `tokio::time`

### Notes

With this commit the first four examples from the rust getting started guide should compile and run on the Microchip ATSAME54-XPRO board and under emulation with `qemu-system-arm`.

1. **My handling of `rand::` is a bit clumsy:**
  * There are issues with multiple included versions of the `rand` crate conflicting with each other.
  * I've special cased the inclusion of `no_std` implementations of functionality missing from the `rand` crate.

2. **Tests are disabled for `ockam_node_no_std`**
  * If we don't set an embedded target the crate can't build.
  * If we set an embedded target the tests can't run.
  * In the medium term I want to set the tests up to be able to run under `qemu-system-arch` which will resolve this.

3. **Some of the crypto modules are currently broken with `no_std`**
  * I see there's been some major refactoring happening here over the last few days.
  * I think it's probably better if I hold off on fixing this again until things have stabilized.

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/lib.rs
+++ b/implementations/rust/ockam/ockam_node/src/lib.rs
@@ -1,4 +1,6 @@
 //! ockam_node - Ockam Node API
+
+#![cfg_attr(not(feature = "std"), no_std)]
 #![deny(
     missing_docs,
     dead_code,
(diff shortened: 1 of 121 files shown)


implementations/rust/ockam/ockam_node/src/lib.rs
```diff
@@ -9,9 +11,71 @@
...
+extern crate alloc;
+
+#[cfg(feature = "std")]
 #[macro_use]
 extern crate tracing;
 
+#[cfg(feature = "std")]
+pub use tokio;
+#[cfg(feature = "no_std")]
+pub use ockam_node_no_std::tokio;
```
spacekookie, line 40: Not a real review but I love how you implement the embedded shim in this. I was wondering while writing the default node whether a real embedded port would even be necessary. Very happy to see that an async shim to map onto the context_m async stuff is enough 🥳

jared-s:
Wahoo! @mrinalwadhwa any idea why CI jobs are not kicking off..is it because it's a PR from fork not branch?

PR #1529, opened 2021-07-07
feat(rust): add bls signature stub

SanjoDeundiak:


SanjoDeundiak:
@mrinalwadhwa @jared-s we should investigate bls signatures in term of public key self signature, and either add proper self signature for those keys, or disable self-sign for them

jared-s:
@SanjoDeundiak did you intend for the target branch of this PR to be `sanjo/from_external` instead of `develop`?

jared-s:
Commits are on develop

PR #1530, opened 2021-07-07
feat(rust): move signing key to change events

SanjoDeundiak:


jared-s:
https://github.com/ockam-network/ockam/pull/1537

PR #1539, opened 2021-07-08
feat(elixir): support hostnames in tcp addresses

hairyhum:
Implement serialization and connection using string hostnames in `TCPAddress`

The schema for address serialization if following:
```
type TCPAddress {
     host: ({:data, 4} | {:data, 16} | string) // ipv4, ipv6 or host
     port: :u16
}
```

This schema is compatible with the current format for TCP address serialization when using IP addresses.


hairyhum:
This needs more rework for compatibility with Rust TCP addresses #1540

hairyhum:
Closed in favour of https://github.com/ockam-network/ockam/pull/1789
For now. We may return to that later.

PR #1544, opened 2021-07-13
Sanjo/sc trust

SanjoDeundiak:


examples/rust/get_started/examples/06-secure-channel-many-hops.rs
```diff
@@ -16,15 +16,15 @@ async fn main(mut ctx: Context) -> Result<()> {
     ctx.start_worker("h3", Hop).await?;
 
     // Create a secure channel listener at address "secure_channel_listener"
-    bob.create_secure_channel_listener("secure_channel_listener")?;
+    bob.create_secure_channel_listener("secure_channel_listener", NoOpTrustPolicy)?;
```
SanjoDeundiak, line 19: @jared-s Should we use IdentifierTrustPolicy here?

jared-s: Probably so, I'll add a note on Miro that we will want to switch to that. Nop is fine for this PR, we're going to redo all of the examples this week anway.

PR #1547, opened 2021-07-15
Sanjo/credentials example

SanjoDeundiak:


implementations/rust/ockam/ockam_entity/src/credential/macros.rs
```diff
@@ -12,8 +12,12 @@ macro_rules! credential_type {
...
         use $crate::CredentialAttributeType::{Number, Utf8String, Blob};
 
-        #[allow(unused_mut)]
-        let mut attributes = vec![];
+        let mut attributes = vec![CredentialAttributeSchema {
+            label: $crate::SECRET_ID.into(),
+            description: "".to_string(),
+            attribute_type: Blob,
+            unknown: true,
+        }]; // FIXME
```
jared-s, line 20: For FIXMEs would you mind also adding a very short description, for clarity? In this case it is that the schema should not be constructed but passed in from somewhere?

implementations/rust/ockam/ockam_entity/src/credential/messages.rs
```diff
@@ -0,0 +1,15 @@
...
+
+#[derive(Serialize, Deserialize)]
+pub enum CredentialProtocolMessage {
+    IssueOfferRequest(String),
+    IssueOffer(crate::CredentialOffer),
+    IssueRequest(crate::CredentialRequest, Vec<crate::CredentialAttribute>),
+    IssueResponse(crate::CredentialFragment2),
+    PresentationOffer,
+    PresentationRequest(crate::ProofRequestId),
+    PresentationResponse(crate::CredentialPresentation),
```
jared-s, line 11: Can we `use` instead of prefix, or is there an advantage here?

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
```diff
@@ -24,7 +35,7 @@ pub trait CredentialIssuer {
         &mut self,
         schema: &CredentialSchema,
         attributes: &[CredentialAttribute],
-    ) -> Result<Credential>;
+    ) -> Result<BbsCredential>;
```
jared-s, line 38: This seems like a big choice since it is exposed to the public API. We likely won't always be tied to BBS+, do we want to expose that detail?

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
```diff
@@ -89,3 +100,151 @@ pub trait CredentialVerifier {
...
+    fn try_from(value: &str) -> Result<Self> {
+        if let Some(str) = value.strip_prefix(Self::PREFIX) {
+            Ok(Self(str.into()))
+        } else {
+            Err(EntityError::InvalidProfileId.into())
+        }
+    }
+}
+
+impl TryFrom<String> for CredentialIdentifier {
```
jared-s, line 143: These are nice, we should do similar trait impls for all of our identifiers

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
```diff
@@ -89,3 +100,151 @@ pub trait CredentialVerifier {
...
+    fn try_from(value: String) -> Result<Self> {
+        Self::try_from(value.as_str())
+    }
+}
+
+#[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
+pub struct Credential {
+    id: CredentialIdentifier,
+    issuer_id: EntityIdentifier,
+    type_id: String,
```
jared-s, line 155: Not much information is conveyed in the name `type_id` - this is really the identifier of the schema yeah?

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
```diff
@@ -89,3 +100,151 @@ pub trait CredentialVerifier {
...
+        &self.type_id
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct EntityCredential {
+    credential: Credential,
+    bbs_credential: BbsCredential,
+    #[serde(with = "BigArray")]
+    issuer_pubkey: [u8; 96],
```
jared-s, line 185: Please use high level type aliases instead of raw arrays, in this case `SigningPublicKey`. Open to re-name of the type.

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
```diff
@@ -89,3 +100,151 @@ pub trait CredentialVerifier {
...
+            issuer_pubkey,
+            schema,
+        }
+    }
+}
+
+pub trait CredentialProtocol {
+    fn create_credential_issuance_listener(
+        &mut self,
+        address: impl Into<Address> + Send,
```
jared-s, line 223: We have other traits that use the generic form `<A: Into<Address> + Send>` on the function - Is there any idiomatic reasoning to using `impl` for these instead? Just curious. I think we should choose one way and stick with it, if possible.

implementations/rust/ockam/ockam_entity/src/credential/workers/holder.rs
```diff
@@ -0,0 +1,128 @@
...
+};
+use async_trait::async_trait;
+use ockam_core::lib::convert::TryInto;
+use ockam_core::{Address, Result, Route, Routed, Worker};
+use ockam_node::Context;
+use rand::random;
+
+enum State {
+    One,
+    Two([u8; 96], CredentialFragment1),
```
jared-s, line 14: Is this a `SigningPublicKey`?

implementations/rust/ockam/ockam_entity/src/credential/workers/holder.rs
```diff
@@ -0,0 +1,128 @@
...
+    CredentialProtocolMessage, CredentialSchema, EntityCredential, EntityError, Holder, Identity,
+    Profile, ProfileIdentifier,
+};
+use async_trait::async_trait;
+use ockam_core::lib::convert::TryInto;
+use ockam_core::{Address, Result, Route, Routed, Worker};
+use ockam_node::Context;
+use rand::random;
+
+enum State {
```
jared-s, line 12: Please rename the states to meaningful process steps

implementations/rust/ockam/ockam_entity/src/credential/workers/holder.rs
```diff
@@ -0,0 +1,128 @@
...
+    type Message = CredentialProtocolMessage;
+
+    async fn initialize(&mut self, ctx: &mut Self::Context) -> Result<()> {
+        ctx.send(
+            self.issuer_route.clone(),
+            CredentialProtocolMessage::IssueOfferRequest(self.schema.id.clone()),
+        )
+        .await?;
+
+        Ok(())
```
jared-s, line 61:
```suggestion
        .await
```

implementations/rust/ockam/ockam_entity/src/credential/workers/issuer.rs
```diff
@@ -0,0 +1,107 @@
...
+    check_message_origin, CredentialAttribute, CredentialProtocolMessage, CredentialSchema, Issuer,
+    Profile, ProfileIdentifier,
+};
+use async_trait::async_trait;
+use ockam_core::{Result, Route, Routed, Worker};
+use ockam_node::Context;
+
+enum State {
+    Offer(Route),
+    Issue([u8; 32]),
```
jared-s, line 11: `ProofRequestId`? or `OfferId` ?

implementations/rust/ockam/ockam_entity/src/credential/workers/listener.rs
```diff
@@ -0,0 +1,63 @@
+use crate::{
+    get_secure_channel_participant_id, CredentialProtocolMessage, CredentialSchema, IssuerWorker,
+    Profile, SecureChannelTrustInfo, TrustPolicy, TrustPolicyImpl,
+};
+use async_trait::async_trait;
+use ockam_core::{Address, Result, Routed, Worker};
+use ockam_node::Context;
+
+pub struct ListenerWorker {
```
jared-s, line 9: `CredentialProtocolListener`? Name is too generic

implementations/rust/ockam/ockam_entity/src/credential/workers/presenter.rs
```diff
@@ -0,0 +1,116 @@
+use crate::{
+    CredentialProtocolMessage, EntityCredential, Holder, PresentationFinishedMessage,
+    PresentationManifest, Profile,
+};
+use async_trait::async_trait;
+use ockam_core::{Address, Result, Route, Routed, Worker};
+use ockam_node::Context;
+
+enum State {
```
jared-s, line 9: Semantic state names please

implementations/rust/ockam/ockam_entity/src/credential/workers/verifier.rs
```diff
@@ -0,0 +1,120 @@
...
+use crate::{
+    check_message_origin, get_secure_channel_participant_id, CredentialAttribute,
+    CredentialProtocolMessage, CredentialSchema, PresentationManifest, Profile, ProfileIdentifier,
+    ProofRequestId, SigningPublicKey,
+};
+use async_trait::async_trait;
+use ockam_core::{Address, Result, Routed, Worker};
+use ockam_node::Context;
+
+enum State {
```
jared-s, line 11: Semantic state names here

jared-s (changes requested):
Looking good! Lots of good functionality. Some questions and suggestions

implementations/rust/ockam/ockam_entity/src/credential/messages.rs
SanjoDeundiak: There were duplicates in names, but I'll check again

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
SanjoDeundiak: Since we now use Credential for a type that is used in user code, we need another name for this type. Suggestions?

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
SanjoDeundiak: I think we want to stop exposing word "schema"... TypedId is what we ended up in the last meeting, suggestions are welcome of course

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
SanjoDeundiak: I think impl version is more elegant, especially when you have multiple generics and when you have 2 arguments of the same type (e.g. Into<Address>, but they cannot share the same generic because one of them may be String and the other &str.
The only constraint is that you cannot use impl in struct definition

implementations/rust/ockam/ockam_entity/src/credential/workers/holder.rs
SanjoDeundiak: I actually prefer returning Ok(()), because it's easier to modify such functions

implementations/rust/ockam/ockam_entity/src/credential/workers/listener.rs
SanjoDeundiak: I think we should consider making this type private, the name is then not a concern anymore, right?

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
jared-s: Let's rename later, good for now

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
jared-s: I read a bit more and I agree it's more elegant, I will switch Identity trait to this style

jared-s (approved):
Thanks for all the changes, looks great

PR #1553, opened 2021-07-19
Select a version for `rand` in the examples.

thomcc:
### Proposed Changes

Explicitly specify the version of `rand` in the example crate.

---

Without this, when building the workspace I get the following error:

```
error: failed to select a version for `rand`.
    ... required by package `ockam_get_started v0.1.0 (/Users/thom/src/ockam/examples/rust/get_started)`
versions that meet the requirements `=0.7.3` are: 0.7.3

the package `ockam_get_started` depends on `rand`, with features: `std_rng` but `rand` does not have these features.

failed to select a version for `rand` which could resolve this conflict
```

Specifying the rand version fixes the error, and makes it clear which version of `rand` this is coded against. It also future-proofs it, in the case where `rand` has breaking changes in the 0.9 release (or some later semver-major release).

### DIFF
--- a/examples/rust/get_started/Cargo.toml
+++ b/examples/rust/get_started/Cargo.toml
@@ -8,5 +8,5 @@ edition = "2018"
 ockam = {path = "../../../implementations/rust/ockam/ockam", features = ["ockam_transport_tcp", "ockam_vault"]}
 
 [dev-dependencies]
-rand = { version = "*", features = ["std_rng"] }
+rand = { version = "0.8.4", features = ["std_rng"] }
 


jared-s:
Great find. A couple of things before we can merge this:

- Could you head over to [our contributors repo](https://github.com/ockam-network/contributors) and PR a row for yourself in CONTRIBUTORS.csv?
- We also require signed commits for all commits, with the same key that signed the CLA

jared-s:
Great, thanks! One very small detail left. We have fairly strict lint rules for commit messages. Would you mind revising the message to be something like `fix(rust): select a version for rand in the examples`?

thomcc:
Is that good? I can also delete the description text as well, if you meant for me to *only* include that text in the message.

jared-s:
Perfect, thank you!

thomcc:
I've deleted the other parts of the commit message, as reproducing the error text exceeded 100 columns, which caused your commitlint to fail.

Wrapping the line feels weird since it's an exact quote of the error text, so I've opted just to remove it.

It was identical to the PR submission comment anyway.

jared-s:
I went ahead and merged it in [this PR](https://github.com/ockam-network/ockam/pull/1554) after doing one more fixup that I forgot to mention earlier (100 character limit on messages).

Thanks again!

thomcc:
Ah, sorry that was necessary!

PR #1561, opened 2021-07-19
fix(rust): include stddef.h in ockam_ffi/include/vault.h

thomcc:
### Proposed Changes

Add `#include <stddef.h>` to `ockam_ffi/include/vault.h`.

This is a very minor issue I noticed when flipping through files. In order to get `size_t`, which is used all over the place in that code, you should import `stddef.h`. On some platforms, `stdint.h` will itself include `stddef.h`, which means the include probably works on many targets.

This isn't guaranteed though, and at least on my machine (on macOS), it doesn't appear to be the case.

### DIFF
--- a/implementations/rust/ockam/ockam_ffi/include/vault.h
+++ b/implementations/rust/ockam/ockam_ffi/include/vault.h
@@ -3,6 +3,7 @@
 #ifndef RUST_VAULT_H
 #define RUST_VAULT_H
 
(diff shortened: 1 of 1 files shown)


SanjoDeundiak (approved):

jared-s:
Rebased and merged as https://github.com/ockam-network/ockam/pull/1569 Thanks!

PR #1581, opened 2021-07-22
Sanjo/mitm worker

SanjoDeundiak:

### DIFF
--- a/examples/rust/get_started/src/lib.rs
+++ b/examples/rust/get_started/src/lib.rs
@@ -3,3 +3,6 @@ pub use echoer::*;
 
 mod hop;
 pub use hop::*;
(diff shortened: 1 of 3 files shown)


jared-s (approved):

mrinalwadhwa:
@SanjoDeundiak @jared-s The name for something like this this in messaging literature is Wire Tap
https://www.enterpriseintegrationpatterns.com/patterns/messaging/WireTap.html

SanjoDeundiak:
@mrinalwadhwa I imagine this could be pretty neat with WireTap Worker + LoggerWorker that doesn't forward, just prints to the console. Not sure we need this complication for get started though

PR #1590, opened 2021-07-23
Use case guides

hairyhum:


hairyhum:
Improving the guide to make it shorter:

- Instead of using echo, just send a message one way and print it
- Remove `tcp.connect` (https://github.com/ockam-network/ockam/pull/1574)

jared-s:
Merged into `guides_draft`

Issue #1374, opened 2021-05-08
Vault - MPC - Secure Multiparty Computation / Threshold Signatures


mrinalwadhwa:
This looks like a useful starting point.
https://github.com/ing-bank/threshold-signatures

L3pereira:
Hello, 
Can I try solve this issue?

L3pereira:
> This looks like a useful starting point.
> https://github.com/ing-bank/threshold-signatures

the package you mentioned (ecdsa-mpc) , has some problems, the guys developing zeroize lib, moved all versions bellow 1.0.0.
ecdsa-mpc not only uses zeroize 0.10.1, but also uses curv-kzen that in itself uses also zeroize 0.10.

mrinalwadhwa:
@L3pereira thank you for picking this up! 

We now have a BLS implementation in Ockam
https://github.com/ockam-network/ockam/tree/develop/implementations/rust/ockam/signature_bls

@mikelodder7 may be able to share some advise on using it in threshold scheme.

L3pereira:
So, if I understand, what you want is an implementation of Shamir's secret sharing with BLS.
For example, you have one node with 3 workers, each worker will sign the message with their private key, then they generate a polynomial where the signature is the free coefficient f(0) and the other points are assigned and sent to the respective workers, once the coefficients are collected (the threshold), the final signature can be reconstructed through Lagrange interpolation.

mikelodder7:
@L3pereira there's the `vsss-rs` crate that can be used to do secret sharing with ockam's bls signature crate. I've used it myself for that purpose. If you want to do threshold signing that's a little more work and would be happy to help you with it.

L3pereira:
> @L3pereira there's the `vsss-rs` crate that can be used to do secret sharing with ockam's bls signature crate. I've used it myself for that purpose. If you want to do threshold signing that's a little more work and would be happy to help you with it.


Ok, I will advance with your crate, 
thank you for your help

L3pereira:
@mikelodder7, I'm trying to create shares from a BLS signature, but  the signature is a G1Projective and I need a PrimeField
for example, 

```
let hash_msg: G1Projective  = G1Projective::hash::<ExpandMsgXmd<sha2::Sha256>>(msg.as_ref(), DST);
let signature : G1Projective = hash_msg * sk.0;
let res = Shamir::<2, 3>::split_secret::<G1Projective , MockRng, 33>(signature , &mut rng);
```

[...]

PR #1591, opened 2021-07-23
feat(rust): add threshold bls signing

mikelodder7:
This PR adds threshold BLS signatures. 
Threshold BLS works by using key shares to sign instead of secret keys which yield partial signatures which cannot be verified by itself. Partial signatures must be combined with a minimum threshold of partial signatures to produce a complete signature which can be verified.

### DIFF
--- a/implementations/rust/ockam/ockam/Cargo.lock
+++ b/implementations/rust/ockam/ockam/Cargo.lock
@@ -344,9 +344,9 @@ checksum = "1847abb9cb65d566acd5942e94aea9c8f547ad02c98e1649326fc0e8910b8b1e"
 
 [[package]]
 name = "futures"
-version = "0.3.15"
+version = "0.3.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0e7e43a803dae2fa37c1f6a8fe121e1f7bf9548b4dfc0522a42f34145dadfc27"
+checksum = "1adc00f486adfc9ce99f77d717836f0c5aa84965eb0b4f051f4e83f7cab53f8b"
 dependencies = [
  "futures-channel",
  "futures-core",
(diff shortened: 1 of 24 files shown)


mrinalwadhwa:
@mikelodder7 thank you for sending this 🙏  🎉 
This looks like a great start to a multi party vault, that @L3pereira is also exploring in #1374

mrinalwadhwa:
I updated the remaining locks in examples and merged https://github.com/ockam-network/ockam/commit/52dc91acf33f0e3ef6c7a58bb80ee78221687a04

Thank you @mikelodder7

L3pereira:
@mikelodder7 in your library you have this example https://github.com/mikelodder7/vsss-rs/blob/main/src/tests/bls12_381_tests.rs

and you create signatures over two G1 and G2 
```
        let h1 = G1Projective::hash::<ExpandMsgXmd<sha2::Sha256>>(msg, dst);
        let h2 = G2Projective::hash::<ExpandMsgXmd<sha2::Sha256>>(msg, dst);

        let s1 = h1 * sk;
        let s2 = h2 * sk;
```
and also call combine_shares_group for G1 and G2, then you run two times the miller loop to sum the pairings.
```
let res1 = Shamir::<3, 5>::combine_shares_group::<Scalar, G1Projective, 49>(&sig_shares1);
let res2 = Shamir::<3, 5>::combine_shares_group::<Scalar, G2Projective, 97>(&sig_shares2);
```
But in this commit, the function "from_partials" (signature.rs) combines only on G1.

`        let point =
            Shamir::<T, N>::combine_shares_group::<Scalar, G1Projective, PARTIAL_SIGNATURE_BYTES>(
                &pp,
            )?;`

Could you please explain this difference?

mikelodder7:
In vsss-rs I’m demonstrating how to combine in either group. In this repo the signature is only in one group and thus you only combine in one. The loop is to test that any combination of threshold values should work so it tries them all.

PR #1592, opened 2021-07-24
chore(rust): add examples for websocket transport

adrianbenavides:
## Resolves 

https://github.com/ockam-network/ockam/issues/1589

## Main changes

- Refactor websocket transport to improve error handling + add more logs where needed
- Add variants of examples 7, 8 and 9 using the websocket transport

## Notes

- I had to add the `ockam_transport_websocket` as a dependency and feature of the `ockam` crate. Let me know if I didn't do this right.
- Examples 8 and 9 are pretty much identical to TCP-only examples, with the only difference that one of the hops is using the websocket transport
- Example 7 is a little bit different. Instead of just copy-pasting the tcp example, I moved the example out from the `ockam_transport_websocket` crate (with some improvements in the error handling side), to show a practical usage of this transport.


mrinalwadhwa:
Hey @adrianbenavides thank you for sending this 🙏 

Some suggestions:

1. I think it's best not to have websocket transport be a dependency of the ockam crate. Our goal is to show that someone could write and use a transport without any special support from the ockam crate. The websocket transport can be the first example of this. We only included the tcp transport behind a feature flag so there is some default transport for new users. For these examples lets add `ockam_transport_websocket` as a direct dependency to the [get_started crate](https://github.com/ockam-network/ockam/blob/develop/examples/rust/get_started/Cargo.toml)

2. Examples 8 looks great 👍 

3. Example 9 is really interesting and points to an interesting api design / modularity challenge. The only reason the initiator has a dependency on the websocket transport crate is because [it needs the WS constant.](https://github.com/ockam-network/ockam/blob/bba8fff28bd80ce0ac31cbefdbe7358663650cca/examples/rust/get_started/examples/09-secure-channel-over-many-transports-many-hops-initiator.rs#L20). We would like to find a better solution for this. In the meantime I think arranging Example 9 like Example 8 would be better - websocket in the first hop, tcp in the second hop.

4. Example 7 - our goal is to show that an application doesn't need to change when the underlying Ockam Node implementations are switched. PR #1521 has an example of a new node implementation. Because of this we want to avoid surfacing tokio in our examples - there will be node implementations that won't depend on tokio but would still be able to support websockets (this may need a different websocket implementation)

adrianbenavides:
Thanks for the review @mrinalwadhwa. I've applied all the suggestions you mentioned.

Regarding what you said about examples 7 and 9, now that I'm slightly more familiar with the transport layer, I see now the improvement potential both crates have in terms of design to make them runtime agnostic and reactive to events.

Pretty excited to keep collaborating and see how the ockam ecosystem evolves!

mrinalwadhwa:
Awesome!! Looking forward to your ideas and collaborating.

mrinalwadhwa:
Thank you @adrianbenavides 🙏 
I made some minor changes to commit messages and merged 

cdc5580ccba3b6415f5b30262b167fa0337d83ec
bce00fcc2dfc9f6912f886f629f284e332434d23
4959022eb914ede6e5550247384b87e24328d4b0

PR #1593, opened 2021-07-25
chore: Add KeyId computation while importing secret in ockam_vault

devalain:
### Proposed change
Simply add a private function to `SoftwareVault` that would be called from `secret_import`.

### DIFF
--- a/implementations/rust/ockam/ockam_vault/src/secret_impl.rs
+++ b/implementations/rust/ockam/ockam_vault/src/secret_impl.rs
@@ -47,18 +47,14 @@ impl SecretVault for SoftwareVault {
     /// Generate fresh secret. Only Curve25519 and Buffer types are supported
     fn secret_generate(&mut self, attributes: SecretAttributes) -> ockam_core::Result<Secret> {
         let mut rng = thread_rng(); // FIXME
-        let (key, key_id) = match attributes.stype() {
+        let key = match attributes.stype() {
             SecretType::Curve25519 => {
                 // FIXME
                 let mut bytes = [0u8; 32];
                 rng.fill_bytes(&mut bytes);
                 let sk = x25519_dalek::StaticSecret::from(bytes);
-                let public = x25519_dalek::PublicKey::from(&sk);
-                let private = SecretKey::new(sk.to_bytes().to_vec());
-                let key_id = self
-                    .compute_key_id_for_public_key(&PublicKey::new(public.as_bytes().to_vec()))?;
 
-                (private, Some(key_id))
+                SecretKey::new(sk.to_bytes().to_vec())
             }
             SecretType::Buffer => {
                 if attributes.persistence() != SecretPersistence::Ephemeral {
@@ -66,7 +62,7 @@ impl SecretVault for SoftwareVault {
                 };
                 let mut key = vec![0u8; attributes.length()];
                 rng.fill_bytes(key.as_mut_slice());
-                (SecretKey::new(key), None)
+                SecretKey::new(key)
             }
             SecretType::Aes => {
                 if attributes.length() != AES256_SECRET_LENGTH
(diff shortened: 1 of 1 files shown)


mrinalwadhwa:
@devalain Thank you for sending the PR and picking Ockam to make you first open source contribution ❤️ 🙏 

While we wait for @SanjoDeundiak's review, could you accept our CLA ...

Read the CLA at
https://www.ockam.io/learn/how-to-guides/contributing/cla/

To accept the CLA please send a different PR to our contributors repo indicating that you accept the CLA by adding your Git/Github details in a row at the end of the CONTRIBUTORS.csv file:
https://github.com/ockam-network/contributors/blob/master/CONTRIBUTORS.csv

Looking forward to merging your first contribution!!

implementations/rust/ockam/ockam_vault/src/secret_impl.rs
```diff
@@ -3,14 +3,46 @@ use crate::VaultError;
...
+                    ))?,
+                )
+            }
+            SecretType::Bls => {
+                let bls_secret_key = BlsSecretKey::from_bytes(secret.try_into().unwrap()).unwrap();
+                let public_key =
+                    PublicKey::new(BlsPublicKey::from(&bls_secret_key).to_bytes().into());
+                Some(self.compute_key_id_for_public_key(&public_key)?)
+            }
+            _ => None,
```
SanjoDeundiak, line 41: Can we explicitly enumerate all members here, so that whenever we add new secret type we won't be able to compile library until all affected places are updated?

SanjoDeundiak (changes requested):
Hey @devalain !  Thanks for your contribution. There are few things we need to fix before merging your PR
1. Could you please [update](https://docs.github.com/en/github/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message#rewriting-the-most-recent-commit-message) your commit message to match our [guide](https://www.ockam.io/learn/how-to-guides/contributing/CONTRIBUTING/#commit-messages)?
2. Could you please add simple tests to check your fix? Easiest case would be checking that after both key generation and key import you can extract public key using secret_public_key_get, get key id using compute_key_id_for_public_key and obtain the secret from vault using get_secret_by_key_id?
3. Ideal case is to also add simple test vector with at least 1 predefined BLS and 1 predefined curve25519 key for which key id is computed and compared to predefined values. That's optional
4. After you add tests from p.2 you'll notice that secret_import still always uses None as key id value.

devalain:
Hey @SanjoDeundiak, I added some tests as you suggested in p.1 and I will do the p.2. I did not notice p.3, could you give me more detail about it please ?

SanjoDeundiak:
@devalain I think it's github's UI tricked me into thinking that you replaced key id computation in secret_generate with function call and did not add it to secret_import, but it's vice versa. What do you think about replacing key id computation in secret_generate with a call of your newly added function?
Other than that looks great to me. Waiting for the last test to be added.
P.S. Could you please also run cargo fmt on your code?

devalain:
> What do you think about replacing key id computation in secret_generate with a call of your newly added function?
I thought I would not because then a call to `secret_generate` would compute public keys 2 times. Maybe there is a good reason I didn't think of ?

> Other than that looks great to me. Waiting for the last test to be added.
> P.S. Could you please also run cargo fmt on your code?
Just did it :)

SanjoDeundiak:
@devalain I think secret_generate computes public key only for further computation of key id. So there shouldn't be double computation

implementations/rust/ockam/ockam_vault/src/secret_impl.rs
```diff
@@ -155,4 +188,81 @@ mod tests {
...
+        let public_key = vault.secret_public_key_get(&sec_idx).unwrap();
+        vault.compute_key_id_for_public_key(&public_key).unwrap()
+    }
+
+    #[test]
+    fn secret_import_compute_key_id_predefined() {
+        let tmp = "∇∙E=0,∇∙B=0,∇⨯E=-∂B/∂t,∇⨯B=(1/c²) ∂E/∂t".as_bytes();
+        // This works for BLS only by chance ! Use `signature_bbs_plus::SecretKey::hash` to
+        // generate a secret key for an arbitrary string.
+        let (bytes_c25519, bytes_bls): (&[u8], &[u8]) = (&tmp[0..32], &tmp[32..]);
```
SanjoDeundiak, line 250: I don't think these are valid keys, so as we add key validation during import this test will fail. Let's generate some random keys for this test.
P.S. I hope Maxwell pardons me

SanjoDeundiak: > Ok, I put some random (and boring ;)) keys.
> 
> Joke aside, how do I validate secret keys in ockam ?

I was trying to say that not every 32 byte binary is a valid curve25519 secret key, and I think this is true for any asymmetric cryptography (not 100% sure). Depending on curve there are some arithmetic checks needed to be done to ensure key is valid. We currently don't have them.

SanjoDeundiak:
@devalain merging as #1610

implementations/rust/ockam/ockam_vault/src/secret_impl.rs
devalain: > We currently don't have them.

Wouldn't that be an interesting feature to add to ockam ? I'm not sure in which crate exactly.

mrinalwadhwa: @devalain yes that kind of validation would be a great addition.

SanjoDeundiak: > > We currently don't have them.
> 
> Wouldn't that be an interesting feature to add to ockam ? I'm not sure in which crate exactly.

I think it goes straight to to secret_import in ockam_vault

PR #1578, opened 2021-07-21
feat(elixir): add tracing service to trace payloads

hairyhum:
### Proposed Changes

Add a new service to trace messages going through the Hub.

A tracing address is registered same way as forwarding address
Registered tracing address worker copies payloads to the registered route

PR #1598, opened 2021-07-26
feat(rust): remote_tracer client of tracing service

hairyhum:
Similar API to RemoteForwarder.

Using #1578

### DIFF
--- a/examples/rust/get_started/examples/xx-message-tracing.rs
+++ b/examples/rust/get_started/examples/xx-message-tracing.rs
@@ -1,0 +1,28 @@
+use ockam::{route, TCP, Context, RemoteTracer, Result, TcpTransport};
+
+#[ockam::node]
+async fn main(mut ctx: Context) -> Result<()> {
+    // Create a cloud node by going to https://hub.ockam.network
+    let cloud_node_tcp_address = "127.0.0.1:4000";
+
+    // Initialize the TCP Transport.
+    let tcp = TcpTransport::create(&ctx).await?;
+
+    // Create a TCP connection to your cloud node.
+    tcp.connect(cloud_node_tcp_address).await?;
+
+    // Create a tracer to trace messages to the app address
+    let tracer =
+        RemoteTracer::create(&ctx, route![(TCP, cloud_node_tcp_address)], ctx.address()).await?;
+
+    println!(
+        "Use this address in the Hub route to trace messages: {}",
+        tracer.remote_address()
+    );
+
+    let reply = ctx.receive_timeout::<String>(1000).await?;
+
+    println!("Traced payload: {}", reply);
+
+    ctx.stop().await
+}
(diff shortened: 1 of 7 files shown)


SanjoDeundiak:
@hairyhum so what is the purpose of this service?

hairyhum:
@SanjoDeundiak To create wiretap workers on the Hub to see what a certain step in the Hub routing sees.

SanjoDeundiak:
> @SanjoDeundiak To create wiretap workers on the Hub to see what a certain step in the Hub routing sees.

I think the trick with end-to-end encryption is that you don't depend on what's going on on server-side and you can effectively demonstrate security features on client side. Using other words: the server may be malicious, this is why we need end-to-encryption, therefore, demonstrating that server doesn't see the data by asking it, doesn't prove anything. Showing what we send to the server is the right way.

hairyhum:
I think you're coming from the point of view of proof, but my intention was to make a demo. And malicious servers are completely outside of the scope of the demo.
Showing "what the middle sees" is more snappy as a demo.

Also the Rust application is not always the termination point for the secure channel. If we establish a channel with the Hub to get something from another service. And secure channel is not the only use for wiretap.

The main motivation here was that getting logs from server is hard, especially in demo scenarios.

SanjoDeundiak:
> I think you're coming from the point of view of proof, but my intention was to make a demo. And malicious servers are completely outside of the scope here.
> Showing "what the middle sees" is more snappy in a demo context.
> 
> Also the Rust application is not always the termination point for the secure channel. If we establish a channel with the Hub to get something from another service. And secure channel is not the only use for wiretap.
> 
> The main motivation here was that getting logs from server is hard, especially in demo scenarios.
> 
> It's fine with me if you make existing demos work without it, but in the future demos I'll need that.

I know we want demo, but the idea of demo is to show end-to-end encrypted channels, and showing that we don't trust server and use it solely for transportation needs is a good way of demonstrating our "Zero Trust Architecture".
I think yesterday @mrinalwadhwa mentioned good approach of showing what server sees - adding local tcp hop which prints payloads.

antoinevg:
I don't understand Ockam well enough yet to comment on the value of this as a demo/example but would like to comment on the utility as a debugging aid.

Specifically in my case it would have been very useful to see the routing from the perspective of the server while I was trying to understand why messages in the Kafka streams examples weren't being forwarded to the echoer.

That said @SanjoDeundiak, is your main point that this functionality is useless for demonstrating end-to-end encrypted channels?

If so, I agree this is a valid concern. 

Do you have any ideas on how we could avoid this while still being able to have this (or something else?) functionality to get debug visibility on server-side communication?

Otherwise, what other negative outcomes do we need to guard against if we go down this road?

etorreborre:
@hairyhum should we close this PR now?

PR #1618, opened 2021-07-27
feat: vault keys validation on import

devalain:
#### Validating `Curve25519` secret keys

Every constructor in `x25519_dalek` seems to use the following function on its input:
```rust
fn clamp_scalar(mut scalar: [u8; 32]) -> Scalar {
    scalar[0] &= 248;
    scalar[31] &= 127;
    scalar[31] |= 64;

    Scalar::from_bits(scalar)
}
```

and there is no validation so, I think a good way to validate these secrets is to do something like the following:

```rust
fn is_valid_curve25519_secret(scalar: [u8;32]) -> bool {
    scalar[0] & !248 == 0
        && scalar[31] & !127 == 0
        && scalar[31] & 64 == 64
}
```

#### Validating `Bls` secret keys

I choosed to use `signature_bbs_plus::SecretKey::from_bytes` and then see if it is `Some`.

### DIFF
--- a/implementations/rust/ockam/ockam_vault/src/secret_impl.rs
+++ b/implementations/rust/ockam/ockam_vault/src/secret_impl.rs
@@ -41,6 +41,33 @@ impl SoftwareVault {
             SecretType::Buffer | SecretType::Aes | SecretType::P256 => None,
         })
     }
+
+    /// Validate secret key.
+    pub fn check_secret(
+        &mut self,
+        secret: &[u8],
+        attributes: &SecretAttributes,
+    ) -> ockam_core::Result<()> {
+        match attributes.stype() {
+            SecretType::Curve25519 => {
+                if secret.len() != CURVE25519_SECRET_LENGTH {
+                    return Err(VaultError::InvalidCurve25519SecretLength.into());
+                }
+                let is_clamped =
+                    secret[0] & !248 == 0 && secret[31] & !127 == 0 && secret[31] & 64 == 64;
+                if !is_clamped {
+                    return Err(VaultError::InvalidCurve25519Secret.into());
+                }
+            }
+            SecretType::Bls => {
+                if secret.len() != BlsSecretKey::BYTES {
+                    return Err(VaultError::InvalidBlsSecretLength.into());
+                }
+            }
+            SecretType::Buffer | SecretType::Aes | SecretType::P256 => {}
+        }
+        Ok(())
+    }
 }
 
 impl SecretVault for SoftwareVault {
@@ -99,7 +126,7 @@ impl SecretVault for SoftwareVault {
         secret: &[u8],
         attributes: SecretAttributes,
     ) -> ockam_core::Result<Secret> {
-        // FIXME: Should we check secrets here?
+        self.check_secret(secret, &attributes)?;
         let key_id_opt = self.compute_key_id(secret, &attributes)?;
         self.next_id += 1;
         self.entries.insert(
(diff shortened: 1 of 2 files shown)


SanjoDeundiak:
> ### Proposed Changes
> #### Validating `Curve25519` secret keys
> Every constructor in `x25519_dalek` seems to use the following function on its input:
> 
> ```rust
> fn clamp_scalar(mut scalar: [u8; 32]) -> Scalar {
>     scalar[0] &= 248;
>     scalar[31] &= 127;
>     scalar[31] |= 64;
> 
>     Scalar::from_bits(scalar)
> }
> ```
> 
> and there is no validation so, I think a good way to validate these secrets is to do something like the following:
> 
> ```rust
> fn is_valid_curve25519_secret(scalar: [u8;32]) -> bool {
>     scalar[0] & !248 == 0
>         && scalar[31] & !127 == 0
>         && scalar[31] & 64 == 64
> }
> ```
> 
> #### Validating `Bls` secret keys
> I choosed to use `signature_bbs_plus::SecretKey::from_bytes` and then see if it is `Some`.

Hey @devalain . I would say it's better to avoid pulling this low level arithmetics to our code, if possible. What if we check that scalar == clamp_scalar(scalar)?

devalain:
> Hey @devalain . I would say it's better to avoid pulling this low level arithmetics to our code, if possible. What if we check that scalar == clamp_scalar(scalar)?

That would be better indeed but looking at the [code](https://github.com/dalek-cryptography/x25519-dalek/blob/3bc0d07a095eb88ab50b243cab8e4118ff740f36/src/x25519.rs#L184-L190) of the `x25519-dalek` crate, the function is private. I could use `From<[u8;32]> for StaticSecret`.

But then I would have to clone the secret array right ?

implementations/rust/ockam/ockam_vault/src/secret_impl.rs
```diff
@@ -41,6 +41,41 @@ impl SoftwareVault {
...
+        secret: &[u8],
+        attributes: &SecretAttributes,
+    ) -> ockam_core::Result<()> {
+        match attributes.stype() {
+            SecretType::Curve25519 => {
+                if secret.len() != CURVE25519_SECRET_LENGTH {
+                    return Err(VaultError::InvalidCurve25519SecretLength.into());
+                }
+                let clamped = x25519_dalek::StaticSecret::from(
+                    TryInto::<[u8; CURVE25519_SECRET_LENGTH]>::try_into(secret).unwrap(),
```
SanjoDeundiak, line 57: If this conversion succeeded, then secret.len() == CURVE25519_SECRET_LENGTH. So, instead of unwrap we can return InvalidCurve25519SecretLength if conversion failed, and remove the length check above

implementations/rust/ockam/ockam_vault/src/secret_impl.rs
```diff
@@ -41,6 +41,41 @@ impl SoftwareVault {
...
+                .to_bytes();
+                if secret != &clamped[..] {
+                    return Err(VaultError::InvalidCurve25519Secret.into());
+                }
+            }
+            SecretType::Bls => {
+                if secret.len() != BlsSecretKey::BYTES {
+                    return Err(VaultError::InvalidBlsSecretLength.into());
+                }
+                if BlsSecretKey::from_bytes(secret.try_into().unwrap())
```
SanjoDeundiak, line 68: Same as above

SanjoDeundiak:
@devalain let's remove unwraps and merge it

devalain:
> @devalain let's remove unwraps and merge it

done.

SanjoDeundiak:
@devalain what do you think about mapping error using `map_err` and using `?` operator?

devalain:
> @devalain what do you think about mapping error using `map_err` and using `?` operator? That should simplify code and reduce nesting

I think I like to contribute because in return I learn something :+1:

SanjoDeundiak:
Moved to #1626 1626

PR #1619, opened 2021-07-28
fix(elixir): cleanup config to disable node ip resolution on start

hairyhum:
Node IP is used by TelemetryForwarder, which is a left over from
the previous integraiton with the Hub Console.

Remove IP resolution and default host from the config.

Disable TelemetryForwarder.


mrinalwadhwa (approved):
Looks good 👍

PR #1622, opened 2021-07-28
Updates to hub guides

hairyhum:


jared-s:
Copied content over to `guides_draft` manually since the parent directory structure changed, causing rebase grief. Merged

PR #1628, opened 2021-07-29
feat(rust): add simple entity builder

SanjoDeundiak:


implementations/rust/ockam/ockam_entity/src/entity.rs
```diff
@@ -71,6 +61,10 @@ fn err<T>() -> Result<T> {
 impl Entity {
     pub fn create_profile(&mut self, vault_address: &Address) -> Result<Profile> {
         if let Res::CreateProfile(id) = self.call(CreateProfile(vault_address.clone()))? {
+            // Set current_profile_id, if it's first profile
```
SanjoDeundiak, line 64: @jared-s what do you think?

jared-s: I think it should be ok, it will require explicit calls to create_profile before an entity can use a profile right? so basically we are taking away the default profile and making it lazily created. which i think is fine

jared-s (approved):

jared-s:
@SanjoDeundiak ockam_entity tests are hanging - there may be a failure?

jared-s:
Merged https://github.com/ockam-network/ockam/pull/1670

PR #1645, opened 2021-07-30
chore: update of the invalid tests vectors

devalain:
### Proposed Changes
Updated the keys so that [this PR](https://github.com/ockam-network/ockam/pull/1626) can be merged.


SanjoDeundiak:
#1695

mrinalwadhwa:
@devalain

mrinalwadhwa:
@devalain Before we merge this, could you share the source of these vectors?
I think we should also have a test with something from the noise wiki https://github.com/noiseprotocol/noise_wiki/wiki/Test-vectors

devalain:
> @devalain Before we merge this, could you share the source of these vectors?
> I think we should also have a test with something from the noise wiki https://github.com/noiseprotocol/noise_wiki/wiki/Test-vectors

I answered [there](https://github.com/ockam-network/ockam/pull/1695#issuecomment-892873874). I'm reading the noise wiki now.

devalain:
I checked the [cacophony](https://raw.githubusercontent.com/centromere/cacophony/master/vectors/cacophony.txt) document for a test vectors source and then I write a small [crate](https://github.com/devalain/xx-noise-tests) to verify the `Curve25519` secrets and I found that no key is clamped.

A `Curve25519` secret key is probably not required to be clamped to be valid but I am sure it will be clamped before use (see [there](https://github.com/dalek-cryptography/x25519-dalek/blob/3bc0d07a095eb88ab50b243cab8e4118ff740f36/src/x25519.rs#L81), [there](https://github.com/dalek-cryptography/x25519-dalek/blob/3bc0d07a095eb88ab50b243cab8e4118ff740f36/src/x25519.rs#L131) and [there](https://github.com/dalek-cryptography/x25519-dalek/blob/3bc0d07a095eb88ab50b243cab8e4118ff740f36/src/x25519.rs#L143) for example).

mrinalwadhwa:
@devalain thank you for exploring. We were busy release a bunch of docs yesterday so I couldn't respond sooner.

Test vectors from [this approach](https://github.com/ockam-network/ockam/pull/1695#issuecomment-892873874) protect us against regression in our code. So that is great.

I wanted to include vectors from the Noise wiki as a sanity check that we are and remain compatible with other implementations.

Interesting 🤔
> A Curve25519 secret key is probably not required to be clamped to be valid but I am sure it will be clamped before use (see there, there and there for example).

devalain:
> Test vectors from [this approach](https://github.com/ockam-network/ockam/pull/1695#issuecomment-892873874) protect us against regression in our code. So that is great.

If so what needs to be done yet for merging ?

SanjoDeundiak:
See #1695

PR #1646, opened 2021-07-30
feat(rust): add a simple generator for unique, human-readable identifiers

antoinevg:
### Proposed Changes

A new module, living in the `ockam` crate that makes it possible to easily generate unique identifiers that are suitable for use in distributed systems.

Examples:

```rust
use ockam::Unique;

// looks like: responder-to-initiator_07b11da9
let sender_name = Unique::with_prefix("responder-to-initiator");
```

### DIFF
--- a/implementations/rust/ockam/ockam/src/lib.rs
+++ b/implementations/rust/ockam/ockam/src/lib.rs
@@ -39,6 +39,7 @@ mod lease;
 mod monotonic;
 mod protocols;
 mod remote_forwarder;
+mod unique;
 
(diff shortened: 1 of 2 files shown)


antoinevg:
Modified the code to stick with 32-bit id's for now.

In the longer term we should be looking at basing these off the nodes' Ethernet MAC addresses.

hairyhum (approved):

jared-s:
Merged https://github.com/ockam-network/ockam/pull/1669 !

PR #1663, opened 2021-08-02
refactor: separate websocket transport code into common + implementation

adrianbenavides:
Hi, first off: this is a draft MR and its only purpose is to give the necessary context to discuss the following proposal.

The changes are pretty much straightforward. The code in the `common` directory is the code that could be extracted into a separated crate (`ockam_transport`), which includes the duplicated code between the tcp and websocket transports (router and errors), and a bunch of traits to implement the specific transports.

I did these changes yesterday before pulling the latest router changes from last week, so there is still work to do integrating those changes in this MR. Also, you will note that I've removed the `AtomicBool` usage because I wanted to experiment with a different approach, but I can put it back if that makes sense (see the question below).

But the point of this MR is not the code details, but rather the higher-level transport design. So, before continuing refining the code, I wanted to show you the current status so you guys can tell me what you think about it and if you think it's a good idea to keep forward with this MR.

### General structure

All the code in the `common` module has two dependencies: ockam_core and tracing, and defines the traits that the transport must implement and the logic of the router, which should be the same for all the transports.

The code in the root module, that includes the implementation of the traits for the websocket transport. I've also refactored the code structure to make it easier to navigate, similar to what has been done recently in the tcp transport.

This structure will also enable the creation of a `mock` transport for tests.

### Atomic bool (question)

I have a question regarding the atomic bool. If the router handler had the `ListenerWorker`'s addresses stored, couldn't we just stop them using the handler's `context`? Is this an option that has been already discussed?

### Related issues

https://github.com/ockam-network/ockam/issues/1580


implementations/rust/ockam/ockam_transport_websocket/src/common/router.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+impl RouterHandler {
+    fn new(ctx: Context, addr: Address) -> Result<Self> {
+        debug!("Initializing RouterHandler instance");
+        Ok(Self { ctx, addr })
+    }
+
+    pub async fn shutdown(&self) -> Result<()> {
+        self.ctx.stop_worker(self.addr.clone()).await
```
SanjoDeundiak, line 165: There may be many RouterHandler instances that use the same Router. We definitely don't want to stop Router when one of the Handlers stops.

implementations/rust/ockam/ockam_transport_websocket/src/common/router.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+#[worker]
+impl Worker for Router {
+    type Message = RouterMessage;
+    type Context = Context;
+
+    async fn shutdown(&mut self, ctx: &mut Self::Context) -> Result<()> {
+        for (_accepts, self_addr) in self.map.iter() {
+            ctx.stop_worker(self_addr.clone()).await?;
```
SanjoDeundiak, line 123: We probably should ignore errors here. We should keep in mind that any of the workers could be shut down by the node

SanjoDeundiak:
Hey @adrianbenavides , tnx for this PR! I think we're on the same page in terms of understanding what is the final goal with transport - there should be common crate which defines common functionality that would help avoid duplication and ease implementation of new transports. I think introducing traits that people can implement (like TransportNode, Transport), so that common crate could take those implementations and help to build a complete transport implementation, is a right approach. However, reviewing this piece of work is pretty hard, because of amount of changes and because it's very hard to make sure everything is still working as it was.
My suggestion is let's start with creating ockam_transport crate. Empty crate would be perfect for the first PR! After that, let's start replicating structure we have in tcp implementation (I think tcp is looking pretty good at the moment), but trying to abstract away tcp specific things, transforming them into something that can be used for any transport. Simultaneously, let's start using those things in tcp (ideally ws as well). Errors is ideal first candidate for moving into ockam_transport crate.
Also, if you could add tests for things related to what you're adding to ockam_tranposrt, that would simplify review process very much!
What do you think?

adrianbenavides:
Hi @SanjoDeundiak, I totally agree that this draft is hard to review because it's all new code. I just wanted to show you a high-level design of how I see this refactor before submitting an actual PR. How could I do this the next time?

Back to your reply, I can create a first PR with the new ockam_transport crate with only the errors module and build up from there. Thank you very much for your time :+1: !

PR #1678, opened 2021-08-02
Fix Stream polling and timeouts, add wip stream examples

jared-s:
- feat(rust): reimplement entity secure channels using workers
- doc: add secure channel via streams example wip


jared-s:
Merging in different PR after other patches

Issue #1571, opened 2021-07-21
Implement BLS signatures


SanjoDeundiak:
For Ockam Credentials we use the BLS curve for BBS+ signatures.

In order to use BBS+ signatures, each Entity must generate a BLS key pair.

All the public keys in Ockam are stored inside Entity using linked-list-like data structure, all the updates in this structure are signed by both 1) dedicated Entity Update Key and 2) Self-signed by the key that is added.

Currently BLS signatures are non-op [here](https://github.com/ockam-network/ockam/blob/21b47e742de04c439947e50a46b4e0870d15094a/implementations/rust/ockam/ockam_vault/src/signer_impl.rs#L31) and [here](https://github.com/ockam-network/ockam/blob/21b47e742de04c439947e50a46b4e0870d15094a/implementations/rust/ockam/ockam_vault/src/verifier_impl.rs#L24), which doesn't allow to check self-signature in Entities.

We need to fix this.
The signature structure likely needs to change to support that .. right now it's just an array of fixed size.

---
_If you would like someone to help / mentor / pair program with you as you work on your first Ockam issue, please leave a comment on [this discussion.](https://github.com/ockam-network/ockam/discussions/1081)_

SanjoDeundiak:
Done

PR #1690, opened 2021-08-03
chore(rust): implement BLS signature

feniks65:
### Proposed Changes

Add BBS+ signatures based on BLS curve.

In this pull request, I wanted to address the following issue #1571 

@SanjoDeundiak 
I am not sure if you were thinking about such kind of an implementation, but still, I wanted to prepare something to be able to discuss it further.
Especially my code is far from ideal when it comes to optimizations - in plenty of places I copy the vector with signature bytes. This can be optimized.
Also I am not sure if I did the right thing changing the declarations of sign() and verify() calls in the whole compilation codebase..

Can share your thoughts on those changes?

### DIFF
--- a/implementations/rust/ockam/ockam_key_exchange_x3dh/src/responder.rs
+++ b/implementations/rust/ockam/ockam_key_exchange_x3dh/src/responder.rs
@@ -100,10 +100,14 @@ impl<V: X3dhVault> KeyExchanger for Responder<V> {
                     .sign(identity_secret_key, signed_prekey_pub.as_ref())?;
                 let identity_key = self.vault.secret_public_key_get(identity_secret_key)?;
                 let one_time_prekey_pub = self.vault.secret_public_key_get(one_time_prekey)?;
+                if signature.as_ref().len() != 64 {
+                    return Err(X3DHError::SignatureLenMismatch.into());
+                }
+                let signature_array = array_ref![signature.as_ref(), 0, 64]; //check it against panic
                 let bundle = PreKeyBundle {
                     identity_key,
                     signed_prekey: signed_prekey_pub,
-                    signature_prekey: Signature(signature),
+                    signature_prekey: Signature(*signature_array),
                     one_time_prekey: one_time_prekey_pub,
                 };
                 self.state = ResponderState::Done;
--- a/implementations/rust/ockam/ockam_vault/Cargo.toml
+++ b/implementations/rust/ockam/ockam_vault/Cargo.toml
@@ -33,7 +33,10 @@ sha2 = "0.9"
 x25519-dalek = "1.1"
 zeroize = { version = "1.1", features = ["zeroize_derive"] }
 tracing = "0.1.26"
+signature_core = { version = "0.13.0"                   , path = "../signature_core" }
 signature_bbs_plus = {path = "../signature_bbs_plus", version = "0.13.0"       }
+signature_bls = { version = "0.11.0"                   , path = "../signature_bls" }
+signature_ps = { version = "0.11.0"          , default-features = false }
 
 [dev-dependencies]
 ockam_vault_test_suite = { path = "../ockam_vault_test_suite", version = "0.10.0"                    }
--- a/implementations/rust/ockam/ockam_vault/src/signer_impl.rs
+++ b/implementations/rust/ockam/ockam_vault/src/signer_impl.rs
@@ -2,12 +2,15 @@ use crate::software_vault::SoftwareVault;
 use crate::xeddsa::XEddsaSigner;
 use crate::VaultError;
 use arrayref::array_ref;
-use ockam_vault_core::{Secret, SecretType, Signer, CURVE25519_SECRET_LENGTH};
+use ockam_vault_core::{Secret, SecretType, Signature, Signer, CURVE25519_SECRET_LENGTH};
 use rand::{thread_rng, RngCore};
+use signature_bbs_plus::{Issuer, MessageGenerators};
+use signature_bls::SecretKey;
+use signature_core::lib::Message;
 
 impl Signer for SoftwareVault {
     /// Sign data with xeddsa algorithm. Only curve25519 is supported.
-    fn sign(&mut self, secret_key: &Secret, data: &[u8]) -> ockam_core::Result<[u8; 64]> {
+    fn sign(&mut self, secret_key: &Secret, data: &[u8]) -> ockam_core::Result<Signature> {
         let entry = self.get_entry(secret_key)?;
         let key = entry.key().as_ref();
         match entry.key_attributes().stype() {
@@ -22,13 +25,21 @@ impl Signer for SoftwareVault {
                         CURVE25519_SECRET_LENGTH
                     ))
                     .sign(data.as_ref(), &nonce);
-                    Ok(sig)
+                    Ok(Signature::new(sig.to_vec()))
                 } else {
                     Err(VaultError::InvalidKeyType.into())
                 }
             }
             SecretType::Bls => {
-                Ok([0u8; 64]) // FIXME
+                if key.len() == 32 {
+                    let bls_secret_key = SecretKey::from_bytes(array_ref!(key, 0, 32)).unwrap();
+                    let generators = MessageGenerators::from_secret_key(&bls_secret_key, 1);
+                    let messages = [Message::hash(data)];
+                    let sig = Issuer::sign(&bls_secret_key, &generators, &messages).unwrap();
+                    Ok(Signature::new(sig.to_bytes().to_vec()))
+                } else {
+                    Err(VaultError::InvalidKeyType.into())
+                }
             }
             SecretType::Buffer | SecretType::Aes | SecretType::P256 => {
                 Err(VaultError::InvalidKeyType.into())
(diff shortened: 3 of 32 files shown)


implementations/rust/ockam/ockam_key_exchange_x3dh/src/responder.rs
```diff
@@ -100,10 +100,11 @@ impl<V: X3dhVault> KeyExchanger for Responder<V> {
                     .sign(identity_secret_key, signed_prekey_pub.as_ref())?;
                 let identity_key = self.vault.secret_public_key_get(identity_secret_key)?;
                 let one_time_prekey_pub = self.vault.secret_public_key_get(one_time_prekey)?;
+                let signature_array = *array_ref![signature, 0, 64];
```
SanjoDeundiak, line 103: Let's check the length of signature, we don't want to panic in case signature of invalid size was passed

implementations/rust/ockam/ockam_vault_core/Cargo.toml
```diff
@@ -23,4 +23,5 @@ heapless = { version = "0.7", features = ["serde"], optional = true }
 serde = {version = "1.0", features = ["derive"]}
 serde-big-array = "0.3"
 zeroize = { version = "1.1", features = ["zeroize_derive"] }
+either = "1.6.1"
```
SanjoDeundiak, line 26: Is it used?

implementations/rust/ockam/ockam_vault/Cargo.toml
```diff
@@ -33,7 +33,11 @@ sha2 = "0.9"
 x25519-dalek = "1.1"
 zeroize = { version = "1.1", features = ["zeroize_derive"] }
 tracing = "0.1.26"
+either = "1.6.1"
```
SanjoDeundiak, line 36: Is it used?

implementations/rust/ockam/ockam_vault_test_suite/src/signer_impl.rs
```diff
@@ -14,7 +15,9 @@ pub fn sign(vault: &mut (impl Signer + Verifier + SecretVault)) {
     let res = vault.sign(&secret, b"hello world!");
     assert!(res.is_ok());
     let pubkey = vault.secret_public_key_get(&secret).unwrap();
+    //*array_ref![digest, 0, 32]
     let signature = res.unwrap();
-    let res = vault.verify(&signature, &pubkey, b"hello world!").unwrap();
+    let signature_array = *array_ref![signature, 0, 64];
```
SanjoDeundiak, line 20: Why do we need this? Is there now difference in types that sign produces and verify accepts?

implementations/rust/ockam/ockam_vault_test_suite/src/signer_impl.rs
```diff
@@ -14,7 +15,9 @@ pub fn sign(vault: &mut (impl Signer + Verifier + SecretVault)) {
     let res = vault.sign(&secret, b"hello world!");
     assert!(res.is_ok());
     let pubkey = vault.secret_public_key_get(&secret).unwrap();
+    //*array_ref![digest, 0, 32]
```
SanjoDeundiak, line 18: Let's remove it

implementations/rust/ockam/ockam_vault/src/verifier_impl.rs
```diff
@@ -2,26 +2,37 @@ use crate::software_vault::SoftwareVault;
...
     fn verify(
         &mut self,
-        signature: &[u8; 64],
+        signature: Buffer<u8>,
         public_key: &PublicKey,
         data: &[u8],
     ) -> ockam_core::Result<bool> {
         // TODO: Add public key type
         if public_key.as_ref().len() == CURVE25519_PUBLIC_LENGTH {
+            let signature_array = array_ref!(signature, 0, 64);
```
SanjoDeundiak, line 20: Should not panic if length is invalid

implementations/rust/ockam/ockam_vault/src/signer_impl.rs
```diff
@@ -22,13 +26,23 @@ impl Signer for SoftwareVault {
...
-                    Ok(sig)
+                    Ok(sig.into())
                 } else {
                     Err(VaultError::InvalidKeyType.into())
                 }
             }
             SecretType::Bls => {
-                Ok([0u8; 64]) // FIXME
+                if key.len() == 32 {
+                    let bls_secret_key = SecretKey::from_bytes(array_ref!(key, 0, 32)).unwrap();
```
SanjoDeundiak, line 36: @feniks65 @mikelodder7 is this the proper way of using BLS for simple signing? Could we use ECDSA? Would it be simpler/more effective?

implementations/rust/ockam/ockam_vault/src/signer_impl.rs
```diff
@@ -2,12 +2,16 @@ use crate::software_vault::SoftwareVault;
 use crate::xeddsa::XEddsaSigner;
 use crate::VaultError;
 use arrayref::array_ref;
-use ockam_vault_core::{Secret, SecretType, Signer, CURVE25519_SECRET_LENGTH};
+use ockam_vault_core::{Buffer, Secret, SecretType, Signer, CURVE25519_SECRET_LENGTH};
 use rand::{thread_rng, RngCore};
+use signature_bls::SecretKey;
+use signature_bbs_plus::{Issuer, MessageGenerators};
+use signature_core::lib::*;
```
SanjoDeundiak, line 9: Let's avoid importing the whole module

SanjoDeundiak:
@feniks65 tnx for the PR! I left some comments

implementations/rust/ockam/ockam_vault_core/Cargo.toml
feniks65: It is not. Thanks for pointing this out. My first idea was to user Either type to store two types of signature arrays - one 64 bit long and the second one for BBS+

implementations/rust/ockam/ockam_key_exchange_x3dh/src/responder.rs
feniks65: Okey. Now I have added checking signature length in those places where I copy inner bytes of a signature.
So it should be solved now.

implementations/rust/ockam/ockam_vault/Cargo.toml
feniks65: No, it is leftover from my experiments. Removed.

implementations/rust/ockam/ockam_vault_core/src/verifier.rs
```diff
@@ -1,10 +1,10 @@
-use crate::PublicKey;
+use crate::{Buffer, PublicKey};
 use ockam_core::Result;
 use zeroize::Zeroize;
 
 /// Signature verification functionality
 pub trait Verifier: Zeroize {
     /// Verify a signature for given data using given public key
-    fn verify(&mut self, signature: &[u8; 64], public_key: &PublicKey, data: &[u8])
+    fn verify(&mut self, signature: Buffer<u8>, public_key: &PublicKey, data: &[u8])
```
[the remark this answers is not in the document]
feniks65: Okay. Thanks for pointing me out how to solve this. I have create Signature type in ockam_vault_core/src/types.rs and I use it instead of Buffer<u8>.

implementations/rust/ockam/ockam_vault_test_suite/src/signer_impl.rs
feniks65: Yeah. I have mixed it but now, since Signature type is there as you've suggested it should be fine for testing. The same on the sign() side and verify() side.

implementations/rust/ockam/ockam_vault_test_suite/src/signer_impl.rs
feniks65: Yes. Removed.

implementations/rust/ockam/ockam_vault/src/verifier_impl.rs
feniks65: Ok. Now I check if signature length equals to 64 bit. Should not panic now.

implementations/rust/ockam/ockam_vault/src/signer_impl.rs
feniks65: @SanjoDeundiak @mikelodder7 
I have used an example from the following file implementations/rust/ockam/signature_bbs_plus/README.md
thinking that this is how I should proceed with BBS+ signature signing.

implementations/rust/ockam/ockam_vault/src/signer_impl.rs
feniks65: Sure. Now I have just imported Message type.

implementations/rust/ockam/ockam_vault/src/signer_impl.rs
SanjoDeundiak: @feniks65 I think BBS+ is a little bit overkill in this use case, since we don't use its full potential, but since this is the only thing we have currently implemented for BLS, let's merge it and revisit the decision in the future. Nice job!

SanjoDeundiak:
LGTM! Moved to #1738

PR #1691, opened 2021-08-03
feat(elixir): sharded access to stream index storage

hairyhum:
Since index service is global in the hub, using some index management
integrations like kafka offset management can create a bottleneck for
all clients using the service.

Create a separate processes to perform the storage operations and only wait
for response if one required by the protocol.

TODO: further improvement would be to send the get response from the shard
as well avoiding any blocking on the service side.


mrinalwadhwa (approved):
Look great 👍

mendrugory (approved):

PR #1695, opened 2021-08-04
feat(rust): add key validation during secret_import in ockam_vault

SanjoDeundiak:


SanjoDeundiak:
#1645

mrinalwadhwa:
@jared-s Please hold on merging this until we know the source of the test vectors

devalain:
> @jared-s Please hold on merging this until we know the source of the test vectors

I generated the test vectors with something like
```rust
let attrs = SecretAttributes::new(
    SecretType::Curve25519,
    SecretPersistence::Ephemeral,
    CURVE25519_SECRET_LENGTH,
);
let mut vault = SoftwareVault::default();
let sec_idx = vault.secret_generate(attrs).unwrap();
let secret = vault.secret_export(&sec_idx).unwrap();
println!("{}", hex::encode(&secret));
```
Is that the information you where looking for ?

jared-s (commented):
Would it be possible to integrate well-known test vectors, such as from [RFC 7748](https://datatracker.ietf.org/doc/html/rfc7748#section-5.2)? We need to validate against an external source of truth.

devalain:
> Would it be possible to integrate well-known test vectors, such as from [RFC 7748](https://datatracker.ietf.org/doc/html/rfc7748#section-5.2)? We need to validate against an external source of truth.

It seems these test vectors are not clamped (try example 2 of [this crate](https://github.com/devalain/xx-noise-tests)).

So I think we should change the private key test so that any random 32 bytes array is a valid `curve25519` private key.

SanjoDeundiak:
@devalain After some investigation I agree with you, nice catch. Therefore, I fixed curve25519 key generation, so now clamping is only happens before usage, and also removed curve25519 check during import. XX verified test vectors is great thing to add, but goes out of scope of current issue. But if you want to work on it - that would be great!
@mrinalwadhwa @jared-s should be fine to merge now

jared-s:
Thanks all - Merged as https://github.com/ockam-network/ockam/pull/1750

PR #1706, opened 2021-08-04
chore(elixir): some helpers to clean up workers

hairyhum:
Track last message timestamp coming through a worker.
Make stream service handle stream instance terminations.
Add a helper module to filter workers and stop them
using module name and callback on state.

### DIFF
--- a/implementations/elixir/ockam/ockam_hub/lib/hub/worker_cleanup.ex
+++ b/implementations/elixir/ockam/ockam_hub/lib/hub/worker_cleanup.ex
@@ -1,0 +1,83 @@
+defmodule Ockam.Hub.WorkerCleanup do
+  @moduledoc """
+  Helper functions to clean up workers by condition
+  """
+
+  def cleanup_idle_workers(module, idle_time) do
+    workers = find_idle_workers(module, idle_time)
+    stop_workers(workers)
+  end
+
+  def find_idle_workers(module, idle_time) when is_integer(idle_time) do
+    now = System.os_time(:millisecond)
+
+    find_workers(module, fn {_name, state} ->
+      idle_worker?(state, now, idle_time)
+    end)
+  end
+
+  def find_orphan_forwarders() do
+    module = Ockam.Hub.Service.Forwarding.Forwarder
+
+    filter_fun = fn {_name, state} ->
+      case Map.get(state, :forward_route) do
+        [first_address | _] ->
+          not address_live?(first_address)
+
+        _other ->
+          true
+      end
+    end
+
+    find_workers(module, filter_fun)
+  end
+
+  def address_live?(address) do
+    case Ockam.Node.whereis(address) do
+      pid when is_pid(pid) ->
+        Process.alive?(pid)
+
+      _other ->
+        false
+    end
+  end
+
+  def cleanup_workers(module, filter_fun) do
+    workers = find_workers(module, filter_fun)
+    stop_workers(workers)
+  end
+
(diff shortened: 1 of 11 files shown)


implementations/elixir/ockam/ockam_hub/lib/hub/worker_cleanup.ex
```diff
@@ -0,0 +1,89 @@
...
+    end)
+  end
+
+  def idle_worker?(state, now, idle_time) do
+    case Map.get(state, :last_message_ts) do
+      nil ->
+        false
+
+      val when is_integer(val) ->
+        case now - val do
```
mendrugory, line 74: could we substitute the `case` statement by only `now - val > idle_time` ?

```elixir
val when is_integer(val) ->
  now - val > idle_time
```


It is just a matter of readability, in case it covers the necessary cases, and I know that it is totally subjective.

implementations/elixir/ockam/ockam_kafka/lib/kafka.ex
```diff
@@ -87,6 +89,64 @@ defmodule Ockam.Kafka do
...
+  def fetch(topic, partition, offset, options) do
+    brod_options = brod_options(options)
+
+    :brod.fetch(
+      brod_options.endpoints,
+      topic,
+      partition,
+      offset,
+      20,
+      1,
```
mendrugory, line 127: I would prefer to have the numbers used into functions defined as constant within the module otherwise it is difficult to understand why these numbers are used

hairyhum:
## How to use the cleanup:
- Filter workers by module:
```
Ockam.Hub.WorkerCleanup.find_module_workers(module_name)
# Filter forwarders:
Ockam.Hub.WorkerCleanup.find_module_workers(Ockam.Hub.Service.Forwarding.Forwarder)
# Filter stream instances
Ockam.Hub.WorkerCleanup.find_module_workers(Ockam.Stream.Workers.Stream)
# Filter stream index shards
Ockam.Hub.WorkerCleanup.find_module_workers(Ockam.Stream.Index.Shard)
```

- Find forwarders with broken routes:
```
Ockam.Hub.WorkerCleanup.find_orphan_forwarders()
```

- Find workers which didn't have messages for a while:
```
Ockam.Hub.WorkerCleanup.find_idle_workers(module_name, idle_time_ms)
```

- Find workers with custom function on state:
```
Ockam.Hub.WorkerCleanup.find_workers(module_name, fn({name, state}) -> ... end)
```

- Cleanup workers with custom function on state:
```
Ockam.Hub.WorkerCleanup.cleanup_workers(module_name, fn({name, state}) -> ... end)
```

- Cleanup idle workers:
```
Ockam.Hub.WorkerCleanup.cleanup_idle_workers(module_name, idle_time_ms)
```

- Cleanup forwarders with broken routes:
```
Ockam.Hub.WorkerCleanup.find_orphan_forwarders() |> Ockam.Hub.WorkerCleanup.stop_workers()
```

## How to clean up topics:

- Find idle topics (with no messages for the last `idle_time_ms` milliseconds)
```
Ockam.Hub.Kafka.TopicCleanup.find_idle_topics(idle_time_ms, options)
```

- Delete those topics
```
Ockam.Hub.Kafka.TopicCleanup.cleanup_topics(topics, options)
```

Where `options` can be `[]` when kafka is configured for a node, or be like `:ockam_kafka` configuration in https://github.com/ockam-network/ockam/blob/develop/implementations/elixir/ockam/ockam_hub/config/runtime.exs#L133

PR #1729, opened 2021-08-10
ci: create documentation example cron job

metaclips:
### Proposed Changes
Test documentation examples hourly

### DIFF
--- a/.github/workflows/documentation_examples_check.yml
+++ b/.github/workflows/documentation_examples_check.yml
@@ -1,5 +1,11 @@
 name: documentation_examples_check
-on: push
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 */2 * * *'
+  push:
+    branches:
+      - '*'
 
 jobs:
     guide_examples_check:


.github/workflows/documentation_examples_check.yml
```diff
@@ -1,5 +1,14 @@
 name: documentation_examples_check
-on: push
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 1 * * *'
```
mrinalwadhwa, line 5: Lets change this to [every two hours](https://cron.help/#0_*/2_*_*_*)


```
0 */2 * * *
```

.github/workflows/documentation_examples_check.yml
```diff
@@ -1,5 +1,14 @@
...
+  push:
+    branches:
+      - '*'
+  pull_request:
+    branches:
+      - '*'
```
mrinalwadhwa, line 11: I think - on push should be sufficient, let's remove the pull request trigger. In the past when I've tried keeping both triggers we get unnecessary runs

mrinalwadhwa (changes requested):
Thank you @metaclips. I left a couple of small change requests.

.github/workflows/documentation_examples_check.yml
metaclips: Thanks

.github/workflows/documentation_examples_check.yml
metaclips: Fixed

jared-s:
Merged as https://github.com/ockam-network/ockam/pull/1730

PR #1736, opened 2021-08-12
chore(rust): replace std:: modules with core:: and alternate implementations

antoinevg:
### Proposed Changes

* refactor `ockam::lib` into `ockam::compat`
  - mirror `std::` and `alloc::` module structures
  - add no_std versions of missing `rand::` functionality
  - remove references to any items that are available in `core::`
* update ockam crates to use `ockam::compat` instead of `ockam::lib`
* add no_std and alloc features to all ockam crates
  - populate dependency features for std, no_std and alloc
  - add crate `#[cfg(feature=...)]` attributes

### Notes

The biggest impact of these changes is on how standard library imports look moving forward.

When importing from the standard library the new rules are:

1. always prefer `core::<mod>` over `std::<mod>` where it's available. (e.g. `std::fmt::Result` -> `core::fmt::Result`)
2. use `ockam_core::compat::<mod>` equivalents where possible. (e.g. `std::sync::Arc -> ockam_core::compat::sync::Arc`)
3. if you need to add new items to compat, follow the originating namespace. (e.g. `compat::vec::Vec` and not `compat::Vec`)


implementations/rust/ockam/ockam/Cargo.toml
```diff
@@ -18,13 +18,22 @@ version = "0.24.0-dev"
...
+
+# Option: "alloc" enables support for heap allocation on "no_std"
+# platforms, requires nightly.
+alloc = ["ockam_core/alloc", "ockam_node/alloc", "serde/alloc", "ockam_node_attribute/no_std"]
 
 [dependencies]
-ockam_core = { path = "../ockam_core", version = "0.25.0-dev"         }
-ockam_node = { path = "../ockam_node", version = "0.23.0-dev"        , optional = true }
+ockam_core = { path = "../ockam_core", version = "0.25.0-dev", default-features = false }
+ockam_node = { path = "../ockam_node", version = "0.23.0-dev", default-features = false, optional = true }
```
SanjoDeundiak, line 36: I know it's not introduced by this PR, but why is it optional?

implementations/rust/ockam/ockam/Cargo.toml
```diff
@@ -18,13 +18,22 @@ version = "0.24.0-dev"
 default = ["std", "ockam_transport_tcp", "software_vault", "noise_xx"]
 software_vault = ["ockam_vault", "ockam_vault_sync_core", "ockam_vault_sync_core/software_vault", "ockam_channel/software_vault"]
 noise_xx = ["ockam_key_exchange_xx", "ockam_channel/noise_xx"]
-std = ["ockam_node", "serde/std"]
-alloc = ["ockam_core/alloc", "serde/alloc"]
-no_std = ["ockam_core/no_std", "serde"]
+
+# Option (enabled by default): "std" enables functionality expected to
```
SanjoDeundiak, line 22: Should we add some check to produce meaningful error if chosen feature set is incompatible? E.g. std&no-std enabled simultaneously makes no sense, right?

implementations/rust/ockam/ockam/Cargo.toml
antoinevg: Probably because the original assumption was that there would be a `ockam_node` and a `ockam_node_no_std`. 

We can probably drop this.

implementations/rust/ockam/ockam/Cargo.toml
antoinevg: Compile-time feature checks live in `ockam_core`:

https://github.com/ockam-network/ockam/blob/e7b7d4b2f74b487ba9a8e05a9a1a526a8fcf6079/implementations/rust/ockam/ockam_core/src/lib.rs#L19

implementations/rust/ockam/ockam/src/remote_forwarder.rs
```diff
@@ -1,7 +1,10 @@
 #![deny(missing_docs)]
 
 use crate::{route, Context, OckamError};
+#[cfg(not(feature = "std"))]
+use ockam_core::compat::rand::random;
```
SanjoDeundiak, line 5: Can we simplify this somehow? This looks to me like we need to use this complicated 4-lines random import all the time, am I wrong?

implementations/rust/ockam/ockam_core/src/compat.rs
```diff
@@ -0,0 +1,178 @@
...
+        pub use super::not_random::OsRng;
+    }
+
+    /// Placeholders for various features from 'rand' that are not
+    /// supported on no_std targets.
+    ///
+    /// WARNING: Thess implementations are NOT random, please do not
+    /// try to use these in production!
+    #[cfg(not(feature = "std"))]
+    mod not_random {
```
SanjoDeundiak, line 98: Why is it needed?

implementations/rust/ockam/ockam_core/src/compat.rs
```diff
@@ -0,0 +1,178 @@
...
+        use super::*;
+
+        /// rand::thread_rng()
+        /// WARNING: This implementation is neither random nor thread-local.
+        pub fn thread_rng() -> rand_pcg::Lcg64Xsh32 {
+            use rand::SeedableRng;
+            rand_pcg::Pcg32::seed_from_u64(1234)
```
SanjoDeundiak, line 105: TODO: Replace with true random number by rolling a dice (Kidding)

implementations/rust/ockam/ockam/src/remote_forwarder.rs
antoinevg: `rand` is somewhat of a special case. 

What I originally wanted was to just move it into `compat::` for both `std` and`no_std` builds.

Unfortunately I ran into an issue where different versions of the `rand` crate are being pulled in by the various crates. Specifically there were some breaking changes when I tried to unify the versions.

So this is very much a workaround for now until we can unify the different `rand` versions across the crates. :-P

implementations/rust/ockam/ockam_core/src/compat.rs
```diff
@@ -0,0 +1,178 @@
+//! A facade around the various collections and primitives needed
+//! when using "std", "no_std + alloc" or "no_std" targets.
+//!
+//! When importing from the standard library:
+//!
+//!   1. always prefer core::<mod> over std::<mod> where it's
```
SanjoDeundiak, line 6: Would be great if we could create automated check for this. @mrinalwadhwa @metaclips

implementations/rust/ockam/ockam_core/src/compat.rs
antoinevg: `no_std` lacks some of the operating system based functions. So those are placeholders so we can carry on developing until we have implementations for the various hardware we'll be supporting.

implementations/rust/ockam/ockam_core/src/routing/address.rs
```diff
@@ -119,7 +123,10 @@ impl Address {
 
     /// Generate a random address with a specific type
     pub fn random(tt: u8) -> Self {
-        Self { tt, ..random() }
```
SanjoDeundiak, line 122: How is new implementation different from old one?

implementations/rust/ockam/ockam_entity/src/credential/util.rs
```diff
@@ -75,7 +75,7 @@ where
             let _l = if let Some(l) = s.size_hint() { l } else { 0 };
             let mut buf = Vec::new();
             while let Some(a) = s.next_element()? {
-                #[cfg(all(feature = "no_std", not(feature = "std")))]
+                #[cfg(all(not(feature = "std"), not(feature = "std")))]
```
SanjoDeundiak, line 78: That looks strange to me. Could we put actual push operation outside of cfg attribute, and only handle error inside attribute scope?

implementations/rust/ockam/ockam_key_exchange_core/Cargo.toml
```diff
@@ -12,7 +12,19 @@ keywords = ["ockam", "crypto", "kex", "cryptography", "encryption"]
...
+no_std = ["ockam_core/no_std", "ockam_vault_core/no_std"]
+
+# Option: "alloc" enables support for heap allocation on "no_std" platforms, requires nightly.
+alloc = ["ockam_core/alloc", "ockam_vault_core/alloc"]
+
 [dependencies]
-ockam_core = { path = "../ockam_core", version = "0.25.0-dev"                           }
-ockam_vault_core = { path = "../ockam_vault_core" , version = "0.19.0-dev"                           }
+ockam_core = { path = "../ockam_core", version = "0.25.0-dev", default_features = false }
+ockam_vault_core = { path = "../ockam_vault_core", version = "0.19.0-dev", default_features = false, optional = true }
```
SanjoDeundiak, line 29: Why is it optional?

implementations/rust/ockam/ockam_key_exchange_xx/Cargo.toml
```diff
@@ -12,10 +12,25 @@ keywords = ["ockam", "crypto", "xx", "cryptography", "encryption"]
...
+# Option: "alloc" enables support for heap allocation on "no_std"
+# platforms, requires nightly.
+alloc = ["ockam_core/alloc", "ockam_key_exchange_core/alloc", "ockam_vault_core/alloc"]
+
 [dependencies]
-ockam_core = { path = "../ockam_core", version = "0.25.0-dev"                           }
-ockam_vault_core = { path = "../ockam_vault_core", version = "0.19.0-dev"                           }
-ockam_key_exchange_core = { path = "../ockam_key_exchange_core", version = "0.17.0-dev"                           }
+ockam_core = { path = "../ockam_core", version = "0.25.0-dev", default_features = false }
+ockam_vault_core = { path = "../ockam_vault_core", version = "0.19.0-dev", default_features = false, optional = true }
```
SanjoDeundiak, line 32: Here as well. Why should it be optional?

implementations/rust/ockam/ockam_node/Cargo.toml
```diff
@@ -10,11 +10,31 @@ license = "Apache-2.0"
...
+# Option: "no_std" enables functionality required for platforms
+# without the standard library.
+no_std = ["ockam_core/no_std", "ockam_node_no_std/no_std", "heapless"]
+
+# Option: "alloc" enables support for heap allocation on "no_std"
+# platforms, requires nightly.
+alloc = ["ockam_core/alloc", "ockam_node_no_std/alloc"]
+
+# Option: "dump_internals" when set, will dump the internal state of workers at startup via the trace! macro.
+dump_internals = []
```
SanjoDeundiak, line 29: I think we don't use it anymore

implementations/rust/ockam/ockam_node/src/node.rs
```diff
@@ -31,6 +31,7 @@ impl ockam_core::Worker for NullWorker {
 
 /// Start a node
 pub fn start_node() -> (Context, Executor) {
+    #[cfg(feature = "std")]
```
SanjoDeundiak, line 34: IMHO let's move this to setup_tracing function and not expose those implementation details

implementations/rust/ockam/ockam_core/src/routing/address.rs
antoinevg: Nice catch - I wrote that before I implemented `random()` for `no_std` - I'll revert it.

implementations/rust/ockam/ockam_node/Cargo.toml
SanjoDeundiak: I don't remember why I removed it, but if it works, we can keep it

implementations/rust/ockam/ockam_vault_sync_core/src/vault_mutex/hasher.rs
```diff
@@ -4,7 +4,17 @@ use ockam_vault_core::{Hasher, Secret, SecretAttributes, SmallBuffer};
 
 impl<V: Hasher> Hasher for VaultMutex<V> {
     fn sha256(&mut self, data: &[u8]) -> Result<[u8; 32]> {
-        self.0.lock().unwrap().sha256(data)
+        #[cfg(feature = "std")]
```
SanjoDeundiak, line 7: That looks very copy&pasty. If we can't make it better, how about disabling VaultMutex for no_std? I think we only use it for testing purposes when we don't want to a start node.

implementations/rust/ockam/ockam/Cargo.toml
SanjoDeundiak: I think ockam crate makes no sense without any node as a dependency, so unless it's temporary convenient for your work, l vote drop.

implementations/rust/ockam/ockam_entity/src/credential/util.rs
antoinevg: Heh, busted. Over-zealous search/replace! 

I didn't write this but I'm assuming this was a test for a `heapless::vec::Vec` overflow. I also don't think you can have an attribute inside an expression. 

I'm just going to remove the duplicate cfg check.

implementations/rust/ockam/ockam_key_exchange_core/Cargo.toml
antoinevg: I have no idea why I did that. 😅

implementations/rust/ockam/ockam_key_exchange_xx/Cargo.toml
antoinevg: See above. 😂

implementations/rust/ockam/ockam_node/Cargo.toml
antoinevg: It's still in `ockam_node/src/router.rs` afaik.

implementations/rust/ockam/ockam/src/remote_forwarder.rs
SanjoDeundiak: Hm. I think unifying rand version across dependencies may be impossible, unless we want to fork and maintain all of them) There is no guarantee authors will update their crates' rand dependencies in reasonable time, which will make us use the old version of rand, which is IMHO pretty bad. Any other thoughts how we could fix this issue?

implementations/rust/ockam/ockam_core/src/compat.rs
SanjoDeundiak: Let's make sure we do everything to prevent this implementation from been used. If compilation error here is stopping you from continuing your work, how about runtime panic? Or at least warning in the log?

implementations/rust/ockam/ockam_vault_sync_core/src/vault_mutex/hasher.rs
antoinevg: Can we please keep those for now, t

antoinevg: Can we please keep those for now? I'm busy with a `no_std` Mutex implementation with the same syntax as `std::sync::Mutex` so those are a placeholder until we can move all Mutex usages to `compat::sync::Mutex`

implementations/rust/ockam/ockam_core/src/compat.rs
antoinevg: `panic!` is no good because I do need to actually run them. Logging also will be very verbose. 

That said, these currently only run for `no_std` and we'll be replacing them with their hardware-specific equivalents before we release the dev preview.

Maybe hide them behind a feature for now like `unsafe_random`?

implementations/rust/ockam/ockam_entity/src/credential/util.rs
SanjoDeundiak: I don't think you can have attribute inside an expression, but could that look something like this?
```
let r = buf.push(a);
#[cfg(all(feature = "no_std", not(feature = "std")))]
{
    r.map_err(|_| DError::invalid_length(_l, &self))?;
}
```

implementations/rust/ockam/ockam_node/Cargo.toml
SanjoDeundiak: I think I removed it during my latest changes to node. But if it works, we can keep it

implementations/rust/ockam/ockam_vault_sync_core/src/vault_mutex/hasher.rs
SanjoDeundiak: Sure

implementations/rust/ockam/ockam_entity/src/credential/util.rs
antoinevg: *doh* - that makes sense!

implementations/rust/ockam/ockam_core/src/compat.rs
SanjoDeundiak: Yeah, would be great. Maybe compiler warning also works

SanjoDeundiak (approved):
Nice!

implementations/rust/ockam/ockam_core/src/compat.rs
antoinevg: Cool, I've added the feature, it's disabled by default. Also looks like `compile_warning!` is still not a thing. :-P

jared-s:
Landed as https://github.com/ockam-network/ockam/pull/1754 - :tada:

PR #1739, opened 2021-08-13
docs(rust): update adr 0003 crate feature flags

antoinevg:
### Proposed Changes

Update ADR documentation for `std`, `no_std` and `no_std+alloc` feature flags.


jared-s:
Merged as https://github.com/ockam-network/ockam/pull/1747

PR #1742, opened 2021-08-16
feat(rust): create ockam_transport_core crate for generic transport code

SanjoDeundiak:


SanjoDeundiak:
@adrianbenavides what do you think is the best next step? It would be great if refactor is as granular as possible, but I know it can be challenging. How about moving TcpRouter to ockam_transport_core? It doesn't have direct Tcp dependencies, which should simplify things

adrianbenavides:
> @adrianbenavides what do you think is the best next step? It would be great if refactor is as granular as possible, but I know it can be challenging. How about moving TcpRouter to ockam_transport_core? It doesn't have direct Tcp dependencies, which should simplify things

Yeah, moving the Router should be pretty straightforward (and the last step before abstracting the transport logic :tada:). Another thing that could be moved is the `parse_socket_addr`.

jared-s:
Sorry for the delay, this has been merged as https://github.com/ockam-network/ockam/pull/1755 and we'll publish the new crate shortly! Thank you!

PR #1763, opened 2021-08-18
feat(elixir): token lease manager

mendrugory:
### Proposed Changes

This PR will try to implement a Token Lease Manager which will be able to work with different Token Cloud providers. The first one, which is implemented in this PR, is InfluxDB.


implementations/elixir/ockam/ockam_hub/lib/hub/service/provider/token_lease_manager.ex
```diff
@@ -0,0 +1,33 @@
+## TODO: this needs to be moved to ockam_kafka after we create ockam_hub_core
+defmodule Ockam.TokenLeaseManager.Hub.Service.Provider do
+  @moduledoc false
+  @behaviour Ockam.Hub.Service.Provider
+
+  @services [:influxdb_token_manager]
+  @address "token_lease_service"
```
hairyhum, line 7: Is that generic or specific to influxdb? 
Is there some options to pass to the lease manager to make it influxdb specific?

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager.ex
```diff
@@ -1,130 +1,242 @@
...
+    {:ok, state}
   end
 
-  defp get_cloud_service(name) do
-    case name do
-      :influxdb -> Ockam.TokenLeaseManager.CloudServiceInfluxdb
-      _other -> Ockam.TokenLeaseManager.CloudServiceInfluxdb
-    end
+  defp process_request(received_payload, state) do
+    {_prefix, payload} = get_payload_prefix(received_payload)
```
hairyhum, line 81: Is the purpose of this function to get the prefix, or to trim the prefix?
Is the prefix indended to be used somewhere?

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager.ex
```diff
@@ -1,130 +1,242 @@
...
+    length =
+      case String.length(response) do
+        l when l > 128 and l < 255 ->
+          :binary.encode_unsigned(l) <> <<1>>
+
+        l ->
+          :binary.encode_unsigned(l)
+      end
+
+    length <> response
```
hairyhum, line 94: I suspect that's a replacement for `:bare.encode(string, :string)`, is that so?

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager/storage_service.ex
```diff
@@ -4,9 +4,46 @@ defmodule Ockam.TokenLeaseManager.StorageService do
   @type reason :: any()
   @type lease_id :: String.t()
   @type options :: Keyword.t()
+  @type storage_conf :: map()
 
-  @callback save(lease) :: :ok | {:error, reason}
-  @callback get(lease_id) :: {:ok, lease} | {:error, reason}
-  @callback remove(lease_id) :: :ok | {:error, reason}
-  @callback get_all(lease) :: {:ok, [lease]} | {:error, reason}
+  @callback do_init(options) :: {:ok, pid()} | {:error, reason}
```
hairyhum, line 9: I personally prefer to use `do_` prefix for internal function and not callbacks.
If that's a callback module for a gen-server, maybe we can use the `handle_` prefix instead?
Same for the cloud_service

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager/cloud_service.ex
```diff
@@ -4,11 +4,50 @@ defmodule Ockam.TokenLeaseManager.CloudService do
...
-  @callback get_all() :: {:ok, [lease]} | {:error, reason}
+  @callback do_init(options) :: {:ok, cloud_configuration} | {:error, reason}
+  @callback do_create(cloud_configuration, creation_options) :: {:ok, lease} | {:error, reason}
+  @callback do_revoke(cloud_configuration, token_id) :: :ok | {:error, reason}
+  @callback do_renew(cloud_configuration, token_id) :: :ok | {:error, reason}
+  @callback do_get(cloud_configuration, token_id) :: {:ok, lease} | :not_found | {:error, reason}
+  @callback do_get_all(cloud_configuration) :: {:ok, [lease]} | {:error, reason}
+  @callback do_get_address(cloud_configuration) :: {:ok, binary()} | {:error, reason}
+
+  defmacro __using__(_opts) do
```
hairyhum, line 19: This macro might have been a module. It would be easier to debug. But that's my personal preference.

implementations/elixir/ockam/ockam_hub/lib/hub.ex
```diff
@@ -47,6 +47,16 @@ defmodule Ockam.Hub do
       ] ++
         services_specs
 
+    children =
```
hairyhum, line 50: Let's move that to a service provider as well to keep this module clean.

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager.ex
```diff
@@ -1,130 +1,242 @@
...
+        "get" ->
+          fn -> do_get(state[:cloud_service], payload["token_id"]) end
+      end
+
+    result = fun.()
+
+    encoding_result(result)
+  end
+
+  defp encoding_result(:ok) do
```
hairyhum, line 209: Can we use verbs for actions like that (e.g. `encode/decode_result`). `encoding_result` sounds a lot like a noun to me.

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager.ex
```diff
@@ -1,130 +1,242 @@
...
+         {:ok, storage_service} <-
+           get_storage_service(storage_service_name, cloud_service_name, cloud_service_address) do
+      # Initializing all saved ttls
+      send(self(), :set_all_ttl)
+
+      %{
+        cloud_service: cloud_service,
+        storage_service: storage_service
+      }
+    else
```
hairyhum, line 31: Should we move that branch to `setup`? I suspect it would crash on this error otherwise.

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager.ex
```diff
@@ -1,130 +1,242 @@
...
+      {:ok, mod}
+    else
+      error -> error
+    end
   end
 
-  defp get_all_leases() do
-    # TODO
-    []
+  defp get_cloud_service_module("INFLUXDB"),
```
hairyhum, line 125: Since we have the pluggable behaviour, I think having a strict set of possible modules is not the best idea. 
Can we move the string to module name association out of this module (maybe to the service provider) or just use module names and not strings?

hairyhum (commented):
A few comments mostly about configuration and code style.

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager.ex
mendrugory: At the beginning was to get it, but it is not useful anymore so I could change it to `trim_payload_prefix`

implementations/elixir/ockam/ockam_hub/lib/hub/service/provider/token_lease_manager.ex
mendrugory: It is generic but, instead of using `@services [:token_manager]` I wanted to list all the possible "implementations" of the token manager, depending on the cloud provider.

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager.ex
mendrugory: I do not think so.

I need to encode in 16 bytes the length of the message which is an integer. Let me know if you see anything weird.

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager/storage_service.ex
mendrugory: sure

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager/cloud_service.ex
mendrugory: It was a module, but I just wanted to make it a macro to make it closer to the storage one and to be prepared for future cloud providers that maybe need a long-term connection.

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager.ex
mendrugory: sure

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager.ex
mendrugory: sure

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager.ex
mendrugory: sure

implementations/elixir/ockam/ockam_hub/lib/hub/service/provider/token_lease_manager.ex
hairyhum: I's just confusing that the service config is called `influxdb_token_manager`, while the service address is `token_lease_service`, I think address should be more specific then config, because you can extend config with args, but you don't expand address.

mendrugory: let's call it just `token_lease_manager`

implementations/elixir/ockam/ockam_hub/lib/hub/service/provider.ex
```diff
@@ -73,7 +73,7 @@ defmodule Ockam.Hub.Service.Provider do
     service_providers_map = get_service_providers_map(providers)
 
     Enum.map(services_config, fn service_config ->
-      do_get_service_child_spec(service_config, service_providers_map)
+      handle_get_service_child_spec(service_config, service_providers_map)
```
hairyhum, line 76: Is this change on purpose or just a part of some rename?
Seems like some `do_` were auto-renamed to `handle_`

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager.ex
hairyhum: There is one more left: `decoding_input`

implementations/elixir/ockam/ockam_hub/lib/hub/service/provider.ex
mendrugory: my bad

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager/cloud_service.ex
hairyhum: In that case do we want to return potentially updated state from the `handle` functions? Or they're not supposed to change the state?

implementations/elixir/ockam/ockam_hub/lib/hub/service/token_lease_manager.ex
hairyhum: You're trying to decode/encode a string compatible with Rust. Rust does BARE encoding with a `string` type. So let's use BARE in elixir as well.

hairyhum (approved):

PR #1789, opened 2021-08-25
chore(elixir): change internal transport address format to serialized

hairyhum:
Instead of using protocol for Ockam.Address only allow strings, tuples and
`%Ockam.Address` struct to be handled like addresses.

Change serialization for TCP and UDP addresses to use string IP and hostnames
for compatibility with Rust

Cleanup legacy code around address serialization/parsing

Support hostnames in TCP addresses


hairyhum:
Aiming to address https://github.com/ockam-network/ockam/issues/1540

implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/address.ex
```diff
@@ -1,87 +1,60 @@
...
-  @ipv6 1
+          _other ->
+            {:error, {:invalid_port, port_str}}
+        end
 
-  def serialize(%TCPAddress{ip: ip, port: port}) do
-    with {:ok, serialized_ip} <- serialize_ip(ip),
-         {:ok, serialized_port} <- serialize_port(port) do
-      %{type: @tcp, value: :binary.list_to_bin([serialized_ip, serialized_port])}
+      _other ->
```
mendrugory, line 35: Do we always expect an explicit port or, could it happen that the user assumes a default port?

implementations/elixir/ockam/ockam/lib/ockam/transport/udp/address.ex
```diff
@@ -1,80 +1,69 @@
...
+    case Regex.split(~r/(:)\d*$/, value, include_captures: true, on: :all_but_first, trim: true) do
+      [ip_str, ":", port_str] ->
+        with {:ok, port} <- parse_port(port_str),
+             {:ok, ip} <- parse_ip(ip_str) do
+          {:ok, {ip, port}}
+        else
+          error -> error
+        end
+
+      _other ->
```
mendrugory, line 48: same question than before

PR #1790, opened 2021-08-25
feat(rust): port forwarding service from elixir to rust

### LINKED ISSUE #1761: Forwarder in Rust
We have a Forwarding service in Elixir, it would be neat to have this ported Rust
https://github.com/ockam-network/ockam/blob/develop/implementations/elixir/ockam/ockam_hub/lib/hub/service/forwarding.ex

There is a client `RemoteForwarder` for this in Rust
https://github.com/ockam-network/ockam/blob/develop/implementations/rust/ockam/ockam/src/remote_forwarder.rs

This guide has an example of how a `RemoteForwarder` is used.
https://github.com/ockam-network/ockam/tree/develop/documentation/use-cases/end-to-end-encryption-with-rust#bob

---
_If you would like someone to help / mentor / pair program with you as you work on your first Ockam issue, please leave a comment on [this discussion.](https://github.com/ockam-network/ockam/discussions/1081)_

Stupremee:
This pull request adds the `ForwardingService` to the `ockam` crate, which is a port of the Elixir forwarder service.

Resolves #1761


Stupremee:
I wasn't sure where to put the service, so I added it to the `ockam` crate, but it may be better elsewhere.

mrinalwadhwa:
Awesome, thank you for sending this PR 🎉 
The `ockam` crate is a good place for the Forwarder.

implementations/rust/ockam/ockam/src/forwarder.rs
```diff
@@ -0,0 +1,96 @@
...
+    type Context = Context;
+    type Message = Any;
+
+    async fn handle_message(
+        &mut self,
+        ctx: &mut Self::Context,
+        msg: Routed<Self::Message>,
+    ) -> Result<()> {
+        let forward_route = msg.return_route();
+        let payload = msg.into_transport_message().payload;
```
SanjoDeundiak, line 32: What is expected to be in payload here?

SanjoDeundiak:
Hey @Stupremee ! Thanks for the contribution. Could you please add few simple tests for newly added functionality?

implementations/rust/ockam/ockam/src/forwarder.rs
Stupremee: The client expects to receive the exact same payload it sends.

https://github.com/ockam-network/ockam/blob/ffadf0633adcae4d0a061fa322fb17dd15afe6e6/implementations/rust/ockam/ockam/src/remote_forwarder.rs#L87-L92

implementations/rust/ockam/ockam/src/forwarder.rs
```diff
@@ -0,0 +1,96 @@
+use crate::Context;
+use ockam_core::{Address, Any, LocalMessage, Result, Route, Routed, TransportMessage, Worker};
+use tracing::info;
+
+/// Alias worker to register remote workers under local names.
+///
+/// To talk with this worker, you can use the [`RemoteForwarder`](crate::RemoteForwarder)
+/// which is a compatible client for this server.
+#[non_exhaustive]
+pub struct ForwardingService {}
```
thomcc, line 10: micronit: Since it's non-exhaustive, it can be a unit struct, e.g. `pub struct ForwardingService;`, which has a bit less syntactic noise.

implementations/rust/ockam/ockam/src/forwarder.rs
```diff
@@ -0,0 +1,96 @@
...
+
+impl ForwardingService {
+    /// Start a forwarding service. The address of the forwarding service will be
+    /// `forwarding_service`.
```
thomcc, line 14:
```suggestion
    /// `"forwarding_service"`.
```

(Just to make extra-clear it's not referring to an identifier, or something)

implementations/rust/ockam/ockam/src/forwarder.rs
```diff
@@ -0,0 +1,96 @@
...
+        let payload = msg.into_transport_message().payload;
+        Forwarder::create(ctx, forward_route, payload).await?;
+
+        Ok(())
+    }
+}
+
+struct Forwarder {
+    forward_route: Route,
+    payload: Option<Vec<u8>>,
```
thomcc, line 41: I think a comment about why this is `Option`/when this is `None` might be worth it.

It's so that this extra information can be passed from the `create` function into `initialize`, right?

(I think this is probably a bit of an API wart with Workers work that we' like to address eventually, but for now just a comment would help)

thomcc (commented):
Thanks a ton for the PR, this looks great. I have a couple minor nits that would be good to address (in addition to changes requested by others).

implementations/rust/ockam/ockam/src/forwarder.rs
Stupremee: > It's so that this extra information can be passed from the create function into initialize, right?

Exactly.

PR #1801, opened 2021-08-27
fix(rust): change `size_t` to `uint32_t` in C header file

Hraesvel:
### Proposed Changes

- fix  ockam-network/ockam#1791

### DIFF
--- a/implementations/rust/ockam/ockam_ffi/include/vault.h
+++ b/implementations/rust/ockam/ockam_ffi/include/vault.h
@@ -69,7 +69,7 @@ ockam_vault_extern_error_t ockam_vault_default_init(ockam_vault_t* vault);
  */
 ockam_vault_extern_error_t ockam_vault_sha256(ockam_vault_t  vault,
                                               const uint8_t* input,
-                                              size_t         input_length,
+                                              uint32_t       input_length,
                                               uint8_t*       digest);
 
 /**
@@ -98,7 +98,7 @@ ockam_vault_extern_error_t ockam_vault_secret_import(ockam_vault_t              
                                                      ockam_vault_secret_t*           secret,
                                                      ockam_vault_secret_attributes_t attributes,
                                                      const uint8_t*                  input,
-                                                     size_t                          input_length);
+                                                     uint32_t                        input_length);
 
 /**
  * @brief   Export data from an ockam vault secret into the supplied output buffer.
@@ -112,7 +112,7 @@ ockam_vault_extern_error_t ockam_vault_secret_import(ockam_vault_t              
 ockam_vault_extern_error_t ockam_vault_secret_export(ockam_vault_t        vault,
                                                      ockam_vault_secret_t secret,
                                                      uint8_t*             output_buffer,
-                                                     size_t               output_buffer_size,
+                                                     uint32_t             output_buffer_size,
                                                      size_t*              output_buffer_length);
 
 /**
@@ -127,7 +127,7 @@ ockam_vault_extern_error_t ockam_vault_secret_export(ockam_vault_t        vault,
 ockam_vault_extern_error_t ockam_vault_secret_publickey_get(ockam_vault_t        vault,
                                                             ockam_vault_secret_t secret,
                                                             uint8_t*             output_buffer,
-                                                            size_t               output_buffer_size,
+                                                            uint32_t             output_buffer_size,
                                                             size_t*              output_buffer_length);
 
 /**
@@ -162,7 +162,7 @@ ockam_vault_extern_error_t ockam_vault_secret_destroy(ockam_vault_t vault, ockam
 ockam_vault_extern_error_t ockam_vault_ecdh(ockam_vault_t         vault,
                                             ockam_vault_secret_t  privatekey,
                                             const uint8_t*        peer_publickey,
-                                            size_t                peer_publickey_length,
+                                            uint32_t              peer_publickey_length,
                                             ockam_vault_secret_t* shared_secret);
 
 /**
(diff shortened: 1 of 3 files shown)


thomcc (commented):
Looks great.

There are two areas that I missed before when filing the issue, that I see now, and that it would be nice if you could get. If not, we can address them in a follow up.

1. Can you get the places where we pass `size_t*` and expect it to come through as `&mut u32`? (This one's actually pretty concerning, and the only one of these that I'm not exactly sure how it works "in practice" on 64 bit machines (I guess the full size_t's happen to be zero-initialized first. Lucky...) An example of this is line 233, although there's at least another place where it happens.

    The fix for this should be to pass these as `uint32_t*` as well. I suspect this may require changes to the C-elixir interface that calls it, for example, changing the size_t here to use uint32_t as well. https://github.com/ockam-network/ockam/blob/develop/implementations/elixir/ockam/ockam_vault_software/native/vault/software/vault.c#L266

2. On the rust equivalent code, there are a few places where we have a `uint8_t*` parameter, which represents the pointer to a start of a buffer, but that comes through as a `&mut u8` (for example https://github.com/ockam-network/ockam/blob/develop/implementations/rust/ockam/ockam_ffi/src/vault.rs#L376) and then we read "past the end" of that reference. This is currently UB under stacked borrows, although it's not exploited currently, and may be relaxed in the future (see https://github.com/rust-lang/unsafe-code-guidelines/issues/256, and https://github.com/rust-lang/unsafe-code-guidelines/issues/134 for some discussion of this issue).

    The fix for this should be to make these `*mut u8`, although you may also need to fix some compile errors in the Rust side that happens in a result.

Sorry about this, I should have done a more thorough investigation when filing the issue 😓.

The changes you've made look great, and if the additions I've requested are a bit involved, we can address them in a follow-up.

Hraesvel:
I don't mind making these additional changes. Are there any unit tests that can be used to debug the FFI?

From my understanding change the signature from `&mut u32` to `*mut u32` will involve using some `unsafe {}` blocks.
This is what I would imagine the snippet to look like.
```rust

    let plaintext_length: &mut u32 = unsafe { &mut *plaintext_length };
    *plaintext_length = 0u32;

```

As for `&mut u8` to `*mut u8`  swapping the signatures didn't cause any compile errors.

thomcc:
> From my understanding change the signature from &mut u32 to *mut u32

Ah, my bad, I meant you needed to change the C side to be `uint32_t*` rather than `size_t*. I totally misspoke, and have fixed the comment. Does that make sense though?

> As for &mut u8 to *mut u8 swapping the signatures didn't cause any compile errors.

👍 

> Are there any unit tests that can be used to debug the FFI?

The best way to do this is to run the `ockam_vault_software` tests, which is an elixir library that uses the rust code over the FFI.

You can either install llvm, cmake, rust, erlang, elixir, and some version of java (this may not be a complete list) and run

```
$ cd implementations/elixir/ockam/ockam_vault_software && ../../../../gradlew test
```

or do it via docker, e.g.

```
$ docker run --rm -it -e HOST_USER_ID=$(id -u) --volume $(pwd):/work ockam/builder:latest bash
$ cd implementations/elixir/ockam/ockam_vault_software && ../../../../gradlew test
```

If you have issues here, someone else (possibly @mrinalwadhwa) might need to point you in the right direction, since I'm new enough with ockam to not be super familiar with the ways in which this can go wrong 😅

implementations/rust/ockam/ockam_ffi/include/vault.h
```diff
@@ -113,7 +113,7 @@ ockam_vault_extern_error_t ockam_vault_secret_export(ockam_vault_t        vault,
                                                      ockam_vault_secret_t secret,
                                                      uint8_t*             output_buffer,
                                                      uint32_t             output_buffer_size,
-                                                     size_t*              output_buffer_length);
+                                                     uint32_t&            output_buffer_length);
```
thomcc, line 116: Hmm, this should be `uint32_t*`, not `uint32_t&` (`uint32_t&` is a C++-only type of pointer).

thomcc (commented):
So, there's a bit more that's needed:

There are a few corresponding changes you need to make in implementations/elixir/ockam/ockam_vault_software/native/vault/software/vault.c, in addition.

For example, since you're changing `ockam_vault_secret_export` so that it now takes a `uint32_t*` (a pointer to a `uint32_t`), over here in `ockam_vault_software`, we call it passing in a pointer to a `size_t`:

https://github.com/ockam-network/ockam/blob/ec02e1779e045b8c4ac8f6326d01aff2460d03b0/implementations/elixir/ockam/ockam_vault_software/native/vault/software/vault.c#L266-L268

The fix there is to change the `size_t length = 0;` to `uint32_t length = 0;`.

This also needs to be done in other places that call functions where you changed the `size_t*` parameter to `uint32_t*`.

If this is a bit too involved, then just let me know and I can help you out with that part.

---

P.S. Sorry if this is just an in-progress commit, and you just haven't gotten to the next part yet. 

P.P.S. Also, a minor nit: the commit message should be `fix(rust)` not `fxi(rust)` 🙂

implementations/rust/ockam/ockam_ffi/include/vault.h
Hraesvel: Thanks for bringing that to my attention. dealing with FFI can really get my wires crossed.

metaclips:
> Fixed my commit titles to have the appropriate prefix tags. 
> 
> 
> 
> Also want to note that I used the Docker approach to build and test `ockam_vault_software`. However, I'm working on a Windows machine and I should note that even with using the WSL2 backend for Docker I was sometimes getting failures on builds, related to os permissions...
> 
> 
> 
> With these recent changes, I was able to build and then just run `mix test` directly in `ockam_vault_software` with all tests passing.
> 
> 

Thanks @Hraesvel can you please create an issue with your docker logs. Would love to look into this. Thanks.

thomcc (approved):
Looks great! Thanks for the patch, and for bearing with all the change requests 😀.

Hraesvel:
> 
> 
> Looks great! Thanks for the patch, and for bearing with all the change requests 😀.

Np, thank you for your guidance and review.

look forward to contributing more to the project.

thomcc:
This is ready to be merged! (Sorry, didn't realize I was supposed to do this)

jared-s (approved):

jared-s:
Rebased and merged as https://github.com/ockam-network/ockam/pull/1861 - thank you!

PR #1816, opened 2021-09-01
Use Unified Cargo Workspace (very WIP)

thomcc:
This went much smother than expected.

Discussed in https://github.com/ockam-network/ockam/issues/1815.


implementations/rust/ockam/ockam_examples/example_projects/channel/Cargo.toml
```diff
@@ -3,9 +3,8 @@ name = "channel_examples"
 version = "0.0.0"
 authors = ["Ockam Developers"]
 edition = "2018"
-
-[lib]
-crate-type = ["rlib"]
```
antoinevg, line 8: I didn't know about autobins, nice!

implementations/rust/ockam/ockam_ffi/Cargo.toml
```diff
@@ -15,9 +15,6 @@ description = """FFI layer for ockam_vault.
 [lib]
 crate-type = ["staticlib", "rlib", "cdylib"]
```
antoinevg, line 17: It would be great if Cargo had a way to select crate-type per target so we can get rid of:

```
warning: dropping unsupported crate type `cdylib` for target `thumbv7em-none-eabihf`

warning: `ockam_vault` (lib) generated 1 warning
```

thomcc: In `.cargo/config.toml` you can put
```
[target.<triple>]
rustflags = ["--crate-type", "staticlib"]
```
But I don't know if cargo will still pass it's --crate-type flags in addition.

PR #1828, opened 2021-09-04
Cargo Release

metaclips:
### Proposed Changes
This PR adds automatic continues delivery of crates
We should also support automatic `cargo publish` using [cargo release](https://crates.io/crates/cargo-release). This will require https://github.com/ockam-network/ockam/pull/1816


jared-s:
Is this ready for develop @metaclips  ? It looks close but I see a deploy failure above

PR #1843, opened 2021-09-08
fix(rust): pin vsss

jared-s:


jared-s:
The breaking version of this crate has been yanked. Unnecessary

PR #1844, opened 2021-09-09
ci: perform `cargo update and check` on examples

metaclips:
### Proposed Changes
This pull request creates a cron job that performs cargo update and checks for failing dependency patches


jared-s:
Merged

PR #1883, opened 2021-09-20
docs: update crate-feature-flags decision

thomcc:
This updates the crate-feature-flags doc, about how we use different cargo feature flags, specifically the alloc/std/no_std feature flags.

This change will make us more compatible with what the ecosystem in general does, and easier for us to move to a workspace (this change somewhat blocks work on that front, although I've sort of assumed *something* like this would be accepted.).

It includes some background on why it's organized like this, the status quo in the stdlib (and the ecosystem too), and the differences from that with what we're doing.

Obviously, I'm open to changes/suggestions/improvements here.

(Apologies if I left something in that should have been changed, or in the case of any typos/incomplete sentences... I wrote much of this while on cold medicine, and there was a bit of a gap between starting it and completing it, so it's *entirely possible* I messed up somewhere)

### DIFF
--- a/documentation/architecture/decisions/0003-crate-feature-flags.md
+++ b/documentation/architecture/decisions/0003-crate-feature-flags.md
@@ -1,6 +1,6 @@
 # 3. Crate Feature Flags
 
-Date: 2021-02-10
+Date: 2021-09-20
 
 ## Status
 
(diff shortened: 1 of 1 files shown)


thomcc:
Also, let me know if this should instead be submitted in some other way.

documentation/architecture/decisions/0003-crate-feature-flags.md
```diff
@@ -80,17 +116,24 @@ no\_std without alloc:
 
 To effect conditional compilation on `std`, `no_std+alloc` and `no_std` builds the following attributes can be used:
 
-    // only compile for std
-    #[cfg(feature = "std")]
+```rust
```
thomcc, line 119: This is a lot of changes from how we do things now, but *hopefully* it should make the common cases easier.

documentation/architecture/decisions/0003-crate-feature-flags.md
```diff
@@ -104,3 +147,5 @@ The rules of use are as follows:
 2. otherwise, use `ockam_core::compat::<mod>` equivalents where available. (e.g. `std::sync::Arc -> ockam_core::compat::sync::Arc`)
 3. if you need to add new items to compat, follow the originating namespace. (e.g. `compat::vec::Vec` and not `compat::Vec`)
 4. if none of the above apply use the conditional compilation attributes as documented above
+
+In the future, we may move `ockam_core::compat` into a dedicated crate in order to reduce feature fragility, however prior to general availability, this is probably fine).
```
thomcc, line 151: Thinking about this more, it wouldn't really solve the problem I'm concerned with. Essentially, having pre-1.0 public dependencies prevents us from becoming 1.0 in a sense, since we have a different stability promise/commitment than our public dependencies. (see https://rust-lang.github.io/api-guidelines/necessities.html#c-stable for some rationalization here)

documentation/architecture/decisions/0003-crate-feature-flags.md
```diff
@@ -1,24 +1,62 @@
...
+As you may note, these are analagous to the three configurations we wish to support, but there is an important difference:
+
+We should do not adopt the restriction that `libcore` and `liballoc` has around target-specific checks and `cfg`s and such. We favor practicality over purity, and without this, we would be completely unable to provide useful functionality for embedded environments.
 
 ## Decision
 
+To reflect these three scenarios, we will use three cargo features:
+1. `feature = "std"`, which allows full use of `libstd`. Mirroring the Rust stdlib, it implies `feature = "alloc"`.
+2. `feature = "alloc"`, which allows use of an allocator. As  `feature = "std"`.
+3. `feature = "no_std"` is used to add dependencies which are needed only in `no_std` configurations.
```
thomcc, line 57: This (`feature = "no_std"`) isn't common in the ecosystem from what I've seen, but it's also hard to imagine avoiding it.

That said, in general it's consider very bad to have two dependencies that error when they're both on, but it's not considered nearly as bad to have a feature you must turn on if another feature is disabled. This is largely because the former breaks library users and the latter does not.

documentation/architecture/decisions/0003-crate-feature-flags.md
```diff
@@ -1,24 +1,62 @@
...
 
-1. `no_std` Requires that memory be allocated statically at compile time. Some features might be enabled in this mode for stack based collection usage like `heapless`. This is stack only mode.
-1. `no_std+alloc` Standard collections provided by `alloc` can be used but they do not use the standard allocator. This allows the use of heap allocation via custom allocator(s).
-1. `std` normal Rust. This is standard mode.
+1. `no_std`: Requires that memory be allocated statically at compile time, and no global allocator is available. Some features might be enabled in this mode for stack based collection usage like `heapless`. This is stack only mode.
+1. `no_std+alloc`: Standard collections provided by `alloc` can be used but they do not use the standard allocator. This allows the use of heap allocation via custom allocator(s).
+1. `std`: normal Rust. This is the default mode.
 
-This document describes the various scenarios that can happen and how to address them. It also describes what is needed for end users of the system to understand what they get and ensure consistency across ockam crates.
+### Background: The Rust Standard Libraries
```
thomcc, line 17: This is possibly a bit more in-depth than it needs to be, but hopefully it explains background and rationale.

documentation/architecture/decisions/0003-crate-feature-flags.md
```diff
@@ -1,24 +1,62 @@
...
 
+The public API of Rust's standard library is split into 3 parts.
```
antoinevg, line 19: This is really great background information to have here. Even if you have prior background it's not always clear what is what and why.

For instance, it hadn't ever occurred to me to think of `libstd` as a superset of both `libaloc` and `libcore`. For some reason I'd always categorized `liballoc` into a corner by itself.

documentation/architecture/decisions/0003-crate-feature-flags.md
```diff
@@ -80,17 +116,24 @@ no\_std without alloc:
...
-    // only compile for no_std+alloc
-    #[cfg(feature = "alloc")]
+// compile for `no_std` (without `alloc`) only.
+#[cfg(not(feature = "alloc"))]
 
-    // compile for no_std and no_std_+alloc
-    #[cfg(any(not(feature = "std"), feature = "alloc"))]
+// Compile for no_std+alloc only.
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+```
```
antoinevg, line 136: Much simpler. Thank you. +1

antoinevg (approved):

jared-s (approved):

jared-s:
Rebased and merged as https://github.com/ockam-network/ockam/pull/1909 !

PR #1912, opened 2021-09-28
chore(rust): replace git dependencies

antoinevg:
### Proposed Changes

This pull request removes all dependencies that are referenced as git repositories and replaces them with published crates.

The affected dependencies are:

1. `core2` - upstream pushed our needed changes today.
2. `ockam_executor` - has been moved into the main Ockam repository.
3. `serde_bare` - IN PROGRESS


SanjoDeundiak (approved):

implementations/rust/ockam/ockam_executor/src/channel.rs
```diff
@@ -0,0 +1,151 @@
...
+impl<T: core::fmt::Debug> Receiver<T> {
+    pub async fn recv(&mut self) -> Option<T> {
+        let result = poll_fn(|context| match self.0.queue.dequeue() {
+            Some(value) => {
+                self.0.wake_sender.wake();
+                Poll::Ready(Some(value))
+            }
+            None => {
+                self.0.wake_receiver.register(&context.waker());
+                if self.0.is_sender_closed.load(Ordering::Relaxed) {
```
thomcc, line 122: Does this need to be Acquire to synchronize with the writes in the sender's Drop / ensure prior stores are visible? (or does it not matter here)

implementations/rust/ockam/ockam_executor/src/channel.rs
```diff
@@ -0,0 +1,151 @@
...
+            }
+        })
+        .await;
+        result
+    }
+}
+
+impl<T> core::fmt::Debug for Receiver<T> {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(f, "[Receiver]")
```
thomcc, line 136: If we're concerned about code size, `f.write_str` might be better.

It might not matter without format arguments though, I don't remember.

thomcc (commented):
Just a couple driveby comments for the ockam_executor code. Doesn't need to block landing, although if the ordering issue is actually an issue we should address it.

jared-s (approved):

implementations/rust/ockam/ockam_executor/src/channel.rs
antoinevg: Acquire would be better yes. That said, worst case the executor will do another round of polling so it doesn't matter too much here. Oh and we're (currently) running only in a single-threaded / interrupt-free context (for now).

implementations/rust/ockam/ockam_executor/src/channel.rs
antoinevg: Good to know!

PR #1916, opened 2021-09-29
feat(rust): make credentials optional (disabled by deafult)

SanjoDeundiak:


thomcc (commented):
This looks fine to me, but I probably am not the person who should review this (I'm not very familiar with this code yet).

jared-s (approved):

Issue #1848, opened 2021-09-09
Switch to using Handle everywhere we interact with Workers


SanjoDeundiak:
[Handle](https://github.com/ockam-network/ockam/tree/ockam_v0.27.0/implementations/rust/ockam/ockam_entity/src/lib.rs#L41) should be moved to ockam_node and used everywhere we interact with Workers in request-response manner (e.g. VaultSync)

Hraesvel:
I've been working on this for a bit. [hraesvel/tranpose_handle](https://github.com/ockam-network/ockam/compare/develop...Hraesvel:hraesvel/transpose_handle?expand=1) some of the highlights are listed below.

- I've moved Handle into `ockam_node::handle` bacea17a5280e529b82750da15f31f3d48b7da04
- Added a few methods to expose the fields e5093a2df2aae770b543fcd1bd7b4083b007d13a  &  e5093a2df2aae770b543fcd1bd7b4083b007d13a
- fix code that broke due to the move.

Also I've started to switch `VaultSync` over to `Handle` [hraesvel/impl_handle_for_vault_sync](https://github.com/Hraesvel/ockam/compare/hraesvel/transpose_handle...Hraesvel:hraesvel/impl_handle_for_vault_sync?expand=1).

- convert `VaultSync` 548e3b21e6c27996a757361077fa375a05f98a78, as for label the methods to `depricated & allow unused` I figured that was the best way to bother appease the `#[deny(...)]` and not break code. 
  -  I actually commented `deprecated` attrs out in a later commit. would like your input. Eventually, once all the functions are converted these will be labeled as unused. 
- convert `hasher.rs` to use `call()` methods 4a39c39fc0a31d3348c133807bb500101bc19fa9
- remainder is a bunch of temporary fixes.

want to talk about the conversion that was done to `hasher.rs` I found 2 variations, the difference is minor but it really comes down to use or not use turbo fishing.
[turbo fish](https://github.com/Hraesvel/ockam/blob/4a39c39fc0a31d3348c133807bb500101bc19fa9/implementations/rust/ockam/ockam_vault_sync_core/src/vault_sync/hasher.rs#L8)
[define variable types](https://github.com/Hraesvel/ockam/blob/4a39c39fc0a31d3348c133807bb500101bc19fa9/implementations/rust/ockam/ockam_vault_sync_core/src/vault_sync/hasher.rs#L21)

[...]

Issue #1853, opened 2021-09-09
Fix possible Worker deadlock


SanjoDeundiak:
We have deadlock [here](https://github.com/ockam-network/ockam/tree/ockam_v0.27.0/implementations/rust/ockam/ockam_entity/src/worker/entity_worker.rs#L190) if spawn call is removed. The problem is that called function require interaction with the same worker that it's called from to succeed, but worker is busy and cannot respond, since it waits for that function to finish.

mrinalwadhwa:
@SanjoDeundiak is this still relevant?

PR #1924, opened 2021-10-05
chore(rust): transpose `Handle` component and implementation for `VaultSync`

Hraesvel:
This PR is for issue  #1848

`Handle` component moved into `ockam_node::handle`  in an addition `ockam_vault_sync_core::VaultSync` 
has been converted to use the `Handle` component.

These two parts are separate branches on my fork so if it's preferable I can opt to make two PR. 

@SanjoDeundiak if you can verify that this is to spec and that the way I converted `VaultSync` is suitable

### DIFF
--- a/implementations/rust/ockam/ockam/Cargo.lock
+++ b/implementations/rust/ockam/ockam/Cargo.lock
@@ -210,10 +210,27 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f29fb62df98b3994d31d6ebb80607e75f784bc814bb801b4c75d6a595fcead26"
 dependencies = [
  "digest",
- "ff",
- "group",
+ "ff 0.10.1",
+ "group 0.10.0",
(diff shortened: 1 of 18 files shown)


Hraesvel:
For `rust_test_ockam_ockam_transport_tcp` sometimes it passes and other times it fails.
I'm having trouble debugging the root cause and not sure if it's due to the changes I've made or something else.

Is this related to #1853?

SanjoDeundiak:
Hey @Hraesvel , thanks for submitting this PR! I'll review it soon.
I think your PR has nothing to do with tcp tests failing, and it's not connected to #1853 either. We'll need to investigate and fix the issue

Hraesvel:
@SanjoDeundiak Thanks for the review, I've made the patches.

Also don't mind that force push, that was just me experimenting with squashing commits. something I think I'll be adopting more of in the future. My commit list is kind of long IMO.

PR #1935, opened 2021-10-05
impl core::str::FromStr for Address

### LINKED ISSUE #1926: Implement `FromStr` for `Address` (and implement the existing `Address::from_string` in terms of it)
`Address::from_string` takes `Into<String>` and either returns an `Address` or panics: https://github.com/ockam-network/ockam/blob/6a3d1e715d564fe2a65136e8ea71367b7f752b6f/implementations/rust/ockam/ockam_core/src/routing/address.rs#L94

This is is okay... but is a bit inflexible (no way to handle the error), and not super idiomatic and means people can't access `FromStr` (or use `s.parse()`). Ideally, we should *at least* also implement `FromStr` https://doc.rust-lang.org/nightly/std/str/trait.FromStr.html

I think this should:

1. Move the parsing logic from `Address::from_string` into a new `impl core::str::FromStr for Address` (this should not panic on errors — it should return them).
2. The `Address::from_string` can then become `s.into().parse::<Address>().expect("Invalid address string")`, or something similar.
3. If possible, change `Into<String>` to `AsRef<str>` in that param, but it may be challenging depending on trait bounds.

This is a good issue, but requires a bit of rust experience.

plaxi0s:
### Proposed Changes

Fixes #1926
1. Moved parsing logic from ```Address::from_string``` to ```impl core::str::FromStr for Address```.
2. Created an error type ```AddressParseError``` which is returned from ```core::str::FromStr::from_str``` impl for Address when parsing fails.
3. Modified ```Address::from_string``` to use the above change.

### DIFF
--- a/implementations/rust/ockam/ockam_core/src/routing/address.rs
+++ b/implementations/rust/ockam/ockam_core/src/routing/address.rs
@@ -78,7 +78,46 @@ pub struct Address {
     pub tt: u8,
     inner: Vec<u8>,
 }
+/// An error which is returned when address parsing from string fails
+#[derive(Debug)]
+pub struct AddressParseError {
+    kind: AddressParseErrorKind,
+}
+/// Enum to store the cause of address parsing failure
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AddressParseErrorKind {
+    /// Unable to parse address num in the address string
+    InvalidAddr(std::num::ParseIntError),
+    /// Address string has more than one '#' separator
+    MultipleSep,
+}
 
+impl AddressParseError {
+    /// Create new instance
+    pub fn new(kind: AddressParseErrorKind) -> Self {
+        Self { kind }
+    }
+    /// Gets the cause of address parsing failure.
+    pub fn kind(&self) -> &AddressParseErrorKind {
+        &self.kind
+    }
+}
+impl Display for AddressParseError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &self.kind {
+            AddressParseErrorKind::InvalidAddr(e) => {
+                write!(f, "Failed to parse address type: '{}'", e)
+            }
+            AddressParseErrorKind::MultipleSep => {
+                write!(
+                    f,
+                    "Invalid address string: more than one '#' separator found"
+                )
+            }
+        }
+    }
+}
+impl std::error::Error for AddressParseError {}
 impl Address {
     /// Create a new address from separate type and data parts
     pub fn new<S: Into<String>>(tt: u8, inner: S) -> Self {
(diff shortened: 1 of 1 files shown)


implementations/rust/ockam/ockam_core/src/routing/address.rs
```diff
@@ -78,7 +78,46 @@ pub struct Address {
...
     inner: Vec<u8>,
 }
+/// An error which is returned when address parsing from string fails
+#[derive(Debug)]
+pub struct AddressParseError {
+    kind: AddressParseErrorKind,
+}
+/// Enum to store the cause of address parsing failure
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AddressParseErrorKind {
```
thomcc, line 88:
```suggestion
#[non_exhaustive]
pub enum AddressParseErrorKind {
```

implementations/rust/ockam/ockam_core/src/routing/address.rs
```diff
@@ -92,38 +131,52 @@ impl Address {
...
-                }
-            };
-
-            (tt, vec.remove(0).as_bytes().to_vec())
+            match str::parse(vec.remove(0)) {
+                Ok(tt) => Ok(Address {
+                    tt,
+                    inner: vec.remove(0).as_bytes().to_vec(),
+                }),
+                Err(e) => Err(AddressParseError::new(AddressParseErrorKind::InvalidAddr(
```
thomcc, line 171: (part of my `s/InvalidAddr/InvalidType/` nit)
```suggestion
                Err(e) => Err(AddressParseError::new(AddressParseErrorKind::InvalidType(
```

implementations/rust/ockam/ockam_core/src/routing/address.rs
```diff
@@ -78,7 +78,46 @@ pub struct Address {
...
+    }
+    /// Gets the cause of address parsing failure.
+    pub fn kind(&self) -> &AddressParseErrorKind {
+        &self.kind
+    }
+}
+impl Display for AddressParseError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &self.kind {
+            AddressParseErrorKind::InvalidAddr(e) => {
```
thomcc, line 108: (`s/InvalidAddr/InvalidType/g` continues...)
```suggestion
            AddressParseErrorKind::InvalidType(e) => {
```

thomcc (approved):
Thanks, wow!

I actually had expected you to use ockam_core::Error, but actually, I think you nailed it — Address is fundamental enough for us to be worth being explicit[^1].

Anyway, there's a lot of small tweaks I've suggested, but they're very minor, this is great (and I can handle it if you don't end up getting back)

[^1]: Mostly explict, at least... I'll still need more thoughts for how to move away from the surprise `panic!` without upsetting unbalancing ergonomics too badly...

jared-s:
Rebased and merged as https://github.com/ockam-network/ockam/pull/1955 - Thank you!

Issue #1944, opened 2021-10-06
Split Context methods accepting `Into<T>` generics up to improve compile time and binary size


thomcc:
`Context` has several functions which are generic over `Into<T>` and such, so that the user doesn't have to call `into()` themselves, and we can do it for them. This is great, but depending on how it's done, it can harm compile times, and needlessly increase binary size.

There's a way to avoid these downsides and keep the nice API, which is to separate
the public API (with nice auto-conversions and such) from the implementation.

It would be nice to convert these the methods in Context to follow this pattern.

---

For example, `new_context` is defined as so:

```rs
    pub async fn new_context<S: Into<Address>>(&self, addr: S) -> Result<Context> {
        let addr = addr.into();
        // actual work here...
    }
```

would be transformed this version this one:

```rs
    pub async fn new_context<S: Into<Address>>(&self, addr: S) -> Result<Context> {
        self.new_context_impl(addr.into()).await
    }
    async fn new_context_impl(&self, addr: Address) -> Result<()> {
        // actual work here...
    }
```

PR #1952, opened 2021-10-07
Extract private implementations and `Into<T>` wrappers

giladwo:
### Proposed Changes
This addresses #1944 by splitting methods with `Into<T>` parameters into thin public wrappers and private monomorphic implementations.


giladwo:
@jared-s Could you please mark this PR as `hacktoberfest-accepted`?

PR #1956, opened 2021-10-07
Prebuilt libs for ockam_vault_software

hairyhum:
Pre-built `.so` files for MacOS (universal) and Linux (x86_64)

This way Elixir Ockam implementation can be used without a need to build the libs in Rust.

Custom build libs take precedence over pre-built ones.

There is no automatic re-build though. The pre-built libs need to be re-built when `ockam_vault` or `ockam_ffi` is updated


mrinalwadhwa:
I rebased and merged this in https://github.com/ockam-network/ockam/pull/1980

PR #1972, opened 2021-10-09
Replace `Into<String>` with `AsRef<str>` in `ockam_transport_websocket`

### LINKED ISSUE #1970: Use `AsRef<str>` rather than `Into<String>` in `ockam_transport_websocket` if only a `&str` is needed.
This is equivalent to https://github.com/ockam-network/ockam/issues/1945, but for `ockam_transport_websocket`. I haven't yet checked how many of these there are, but certainly there are a couple, such as https://github.com/ockam-network/ockam/blob/develop/implementations/rust/ockam/ockam_transport_websocket/src/lib.rs#L119

fisherdarling:
### Proposed Changes

This PR replaces `Into<String>` bounds `AsRef<str>` for functions that just need a &str.

closes #1970

### DIFF
--- a/implementations/rust/ockam/ockam_transport_websocket/src/lib.rs
+++ b/implementations/rust/ockam/ockam_transport_websocket/src/lib.rs
@@ -91,8 +91,8 @@ pub struct WebSocketTransport {
 /// WebSocket address type constant
 pub const WS: u8 = 2;
 
-fn parse_socket_addr<S: Into<String>>(s: S) -> Result<SocketAddr> {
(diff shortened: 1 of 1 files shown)


thomcc:
Ah, can you accept our CLA at https://ockam.io/learn/how-to-guides/contributing/cla? It just takes filing a PR adding your info to https://github.com/ockam-network/contributors/blob/master/CONTRIBUTORS.csv.

Sorry for the trouble — we're really looking forward to merging your contribution!

Issue #1969, opened 2021-10-08
Fix broken rustdoc links across ockam crates


thomcc:
If you cd into some of our crates and run `cargo doc`, you'll get warnings (or errors) that the documentation uses links that are no longer valid. For example

```
$ cd implementations/rust/ockam/ockam
$ cargo doc

... <lots of build output> ...

warning: unresolved link to `ockem_core::RouteBuilder`
   --> /Users/thom/src/ockam/implementations/rust/ockam/ockam_node/src/context.rs:255:27
    |
255 |     /// [`RouteBuilder`]: ockem_core::RouteBuilder
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^ no item named `ockem_core` in scope

...

error: unresolved link to `UserParser`
  --> src/protocols/parser.rs:16:20
   |
16 | /// [`UserParser`](UserParser) instead!**
   |                    ^^^^^^^^^^ no item named `UserParser` in scope
   |
```

Pretty much all of these are fair game. If it isn't clear what the documentation should reference instead, it's acceptable to make it no longer a link, as after this I'll be filing another issue for fixing stale references like `Foo` in docs (with more of a guide for how to figure that out).

PR #1974, opened 2021-10-09
Remove outdated documentation about UserParser

bin-wang:
This resolves one of the errors mentioned in #1969. Currently when I run `cargo doc` for the rust implementation there is this error:
```shell
error: unresolved link to `UserParser`
  --> src/protocols/parser.rs:16:20
   |
16 | /// [`UserParser`](UserParser) instead!**
   |                    ^^^^^^^^^^ no item named `UserParser` in scope
   |
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
```

The reason for this error is that the type `UserParser` has been removed in 552d079e2b9bb672f5415769ce59331228355940. I believe there is no equivalent type in the code base, therefore it makes no sense to mention it in the documentation.

### DIFF
--- a/implementations/rust/ockam/ockam/src/protocols/parser.rs
+++ b/implementations/rust/ockam/ockam/src/protocols/parser.rs
@@ -12,9 +12,6 @@ use ockam_core::Decodable;
 
 /// A parser for a protocol fragment
 ///
-/// **If you are not a protocol author, you may want to use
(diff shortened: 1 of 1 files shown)


wkk:
@thomcc Of course. I should have configured my git to sign automatically. Should be verified now. CLA accepted at  ockam-network/contributors#42

Issue #1975, opened 2021-10-09
Run action when workflow files are updated


metaclips:
Currently, GitHub Actions only run when crates or gradle files are updated e.g. https://github.com/ockam-network/ockam/blob/73ff992de50d59967baca88ce5fcfae22ae7a5a0/.github/workflows/rust_build_ockam_ockam_core.yml#L10-L14. 
We should also run workflow actions when workflow files are updated, for [example](https://github.com/ockam-network/ockam/commit/cbb6ecf4d2ebcc9809cb6e053540c258da16f60e)

wkk:
Hi there. I believe I fixed `CommandsPlugin.groovy` but I'm struggling to regenerate the workflow files. Any advice?

mrinalwadhwa:
@wkk thank you so much for the work you put into this.

With the recent change of us using Rust Cargo workspace .. we ended up removing the CommandPlugin completely in #2308 ... It made understanding what is going on the gradle build simpler.

Thank you for taking time to contribute 🙏
Let me know if you want to work on a different feature and I would be happy to help.

PR #1979, opened 2021-10-10
[WIP] ci: run action when workflow files are updated

wkk:
As discussed #1975, I believe that these are the changes needed to add the workflow file itself to the paths to watch for. 

However I'm not sure how I should update the workflow files. I tried `./gradlew :rust:delete_workflows && ./gradlew :rust:generate_workflows`, however it generated many workflow files that weren't tracked. Even in the tracked workflow files there are some unexpected changes, most notably the `Generate random key` and `Cache Gradle` steps are lost and I'm not sure how they were added in the first place.


mrinalwadhwa:
@wkk thank you so much for the work you put into this.

With the recent change of us using Rust Cargo workspace .. we ended up removing the CommandPlugin completely in https://github.com/ockam-network/ockam/pull/2308 ... It made understanding what is going on the gradle build simpler. 

Thank you for taking time to contribute 🙏 
Let me know if you want to work on a different feature and I would be happy to help.

PR #1991, opened 2021-10-12
Sanjo/hraesvel/switch handle/handle move+convert vaultsync

SanjoDeundiak:


jared-s:
Rebased and merged as https://github.com/ockam-network/ockam/pull/1996

PR #2007, opened 2021-10-14
Implement graceful node shutdown mechanism

spacekookie:
This PR does two things:

* Unify worker and processor logic in the router
* Implement a more robust and verifiable shutdown mechanism


spacekookie:
Things left to do documented in #2088

hairyhum:
Regarding multi-hop reliability. 
I don't really see how is that achievable in general. Every forwarding worker in the route may delay, drop, duplicate or modify the messages. Unless we know exactly how the worker behaves there is no way to tell if the message ever to be routed. 
For some workers we can assume they are sources or sinks of other workers, but can we really tell? Am I missing something?

Although I agree that starting/stopping workers in groups (stages, clusters) can be generally useful

In some cases I can imagine you want to shutdown TCP first if connections are a source of messages in the node and not the sink.

spacekookie:
I think the focus is not so much on providing actual messaging reliability on layer 1, especially because that always pulls in a bunch of resend/ control logic with it. I guess the title of the section in my first comment is a bit misleading in that regard.

The changes I was talking about yesterday are in regards to a very simple race condition that we are observing in our examples.

In actuality yes, the runtime can't really guarantee what a worker will do with a message. So this is by no means a replacement for higher level protocols.
Instead I would consider it some additional state the router has to take into account when shutting down a set of workers or the whole node.

implementations/rust/ockam/ockam_core/src/routing/address.rs
```diff
@@ -159,6 +159,13 @@ impl Address {
     pub fn random(tt: u8) -> Self {
         Self { tt, ..random() }
     }
+
+    /// Create a new address with a suffix
+    pub fn suffix<S: Into<String>>(&self, s: S) -> Self {
+        let mut inner = self.inner.clone();
+        inner.append(&mut s.into().as_bytes().to_vec());
+        Self { inner, tt: self.tt }
+    }
```
spacekookie, line 168: This isn't used anymore but I don't know if I should remove it again. Is this conceivably in some other setting?

implementations/rust/ockam/ockam_node/src/relay/processor_relay.rs
```diff
@@ -54,92 +48,36 @@ where
...
-            RxError(ockam_core::Error),
-        }
+        // This future resolves when a stop control signal is received
+        let shutdown_signal = async { ctrl_rx.recv().await };
 
-        #[cfg(feature = "std")]
-        let stop_reason;
+        // Then select over the two futures
         #[cfg(feature = "std")]
         tokio::select! {
```
spacekookie, line 56: @antoinevg especially for this (and the same code basically in `worker_relay.rs` I would like your input

implementations/rust/ockam/ockam_core/src/routing/address.rs
SanjoDeundiak: I heard @antoinevg wanted suffixes/prefixes for addresses. I personally think addresses should be random where possible.

antoinevg: I don't think that was addresses - it was a workaround to deal with the fact that we couldn't block on the return of stream names from elixir:

https://github.com/ockam-network/ockam/commit/931343473032a16a572ea35e0475af48ae175821#diff-408f7d99d304bd5499de15a5e8c41f54eb5cfa3e6b507a20b055764c44066f61R12-R14

The point is probably moot now anyway so don't hold onto that function for my sake!

implementations/rust/ockam/ockam_node/src/relay/processor_relay.rs
antoinevg: The right thing to do here would be to implement `select!` for `no_std` so we don't need to handle the two paths differently.

I'm not proud of the hack I did to work around things in the meantime:

https://github.com/ockam-network/ockam/blob/906ebab4bf2ed18191c966e050fb178298d14f0f/implementations/rust/ockam/ockam_node/src/relay/processor_relay.rs#L66-L89

implementations/rust/ockam/ockam_core/src/routing/address.rs
spacekookie: Alright, i'll remove it again then

implementations/rust/ockam/ockam_node/src/router/utils.rs
```diff
@@ -0,0 +1,74 @@
...
+
+        return Ok(());
+    }
+
+    match router.map.internal.get(&primary_address) {
+        Some(record) => reply.send(NodeReply::sender(addr.clone(), record.sender(), wrap)),
+        None => reply.send(NodeReply::no_such_worker(addr.clone())),
+    }
+    .await
+    .expect("Ockam node internal I/O failed!");
```
SanjoDeundiak, line 41: Should it fail silently or return error instead?

implementations/rust/ockam/ockam_node/src/router/utils.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use ockam_core::{Address, AddressSet, Result};
+
+/// Receive an address and resolve it to a sender
+///
+/// This function only applies to local address types, and will
+/// fail to resolve a correct address if it given a remote
+/// address.
+pub(super) async fn resolve(
+    router: &mut Router,
```
SanjoDeundiak, line 17: Why not make it Router func?

implementations/rust/ockam/ockam_node/src/router/utils.rs
```diff
@@ -0,0 +1,74 @@
...
+pub(super) fn router_addr(router: &mut Router, tt: u8) -> Result<Address> {
+    router
+        .external
+        .get(&tt)
+        .cloned()
+        .ok_or_else(|| Error::InternalIOFailure.into())
+}
+
+/// Check if an address is already in-use by another worker
+pub(super) async fn check_addr_collisions(
```
SanjoDeundiak, line 55: Is there a reason why we would use it? Because of async nature of runtime is there a guarantee that when you checked for collisions there won't be any collision by the time you want to perform an operation (e.g. start worker at that address?)

implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -36,32 +39,45 @@ pub struct Context {
...
         address: AddressSet,
-        mailbox: Mailbox,
-    ) -> Self {
-        Self {
-            rt,
-            sender,
-            address,
-            mailbox,
-        }
+    ) -> (Self, SenderPair, Receiver<CtrlSignal>) {
```
SanjoDeundiak, line 65: I wonder why that doesn't make clippy mad

implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -86,11 +102,15 @@ impl Context {
     }
 
     async fn new_context_impl(&self, addr: Address) -> Result<Context> {
-        let (ctx, tx) = NullWorker::create(Arc::clone(&self.rt), &addr, self.sender.clone());
+        // Create a new context and get access to the mailbox senders
+        let (ctx, sender, _) = Self::new(
```
SanjoDeundiak, line 106: Maybe let's not create excess channel here?

implementations/rust/ockam/ockam_node/src/messages.rs
```diff
@@ -124,6 +136,59 @@ pub enum NodeError {
...
+    /// * Wait for shutdown ACK hooks from worker set
+    /// * Signal worker clusters in reverse-creation order to stop
+    /// * Wait for shutdown ACK hooks from each cluster before moving onto the
+    ///   next
+    /// * All shutdown-signalled workers may process their entire mailbox,
+    ///   while not allowing new messages to be queued
+    ///
+    /// Graceful shutdown procedure will be pre-maturely terminated
+    /// when reaching the timeout (failover into `Immediate`
+    /// strategy).  **A given timeout of `0` will wait forever!**
```
hairyhum, line 173: Is that a standard thing in Rust to mean that `0` is `forever` and not `instant`?

hairyhum:
Is there an API for worker implementations when they get the shutdown signal, or it's supposed to be all internal?

implementations/rust/ockam/ockam_node/src/messages.rs
```diff
@@ -124,6 +136,59 @@ pub enum NodeError {
...
+    /// Rejected because the worker is currently shutting down
+    WorkerShutdown,
+}
+
+/// Specify the type of node shutdown
+///
+/// For most users `ShutdownType::Graceful()` is recommended.  The
+/// `Default` implementation uses a 1 second timeout.
+#[derive(Debug, Copy, Clone)]
+pub enum ShutdownType {
```
thomcc, line 157: nit: Should this be `#[non_exhaustive]`?

thomcc (commented):
This is great! A few drive-by nits, but very exciting progress. I think we'll probably want to figure out something more typesafe than raw strings for the cluster labels, but I could be wrong, and I certainly don't thing that needs to block anything.

Also, it won't let me comment on this file since it's empty, but I think you should delete `ockam_node/src/router/resolve.rs`, which is empty.

implementations/rust/ockam/ockam_node/src/router/stop_worker.rs
```diff
@@ -0,0 +1,53 @@
...
+};
+
+use ockam_core::{Address, Result};
+
+pub(super) async fn exec(
+    router: &mut Router,
+    addr: &Address,
+    reply: &Sender<NodeReplyResult>,
+) -> Result<()> {
+    trace!("Stopping worker '{}'", addr);
```
SanjoDeundiak, line 16: There is control channel for worker too, is it used?

SanjoDeundiak (changes requested):
Tnx for all the effort here so far! I think this PR should add a lot of tests, since it changes very critical parts of our codebase, potentially breaking many things. Particularly:
- Processor shutdown
- Worker shutdown
- Node shutdown (graceful, stop now, timeout shutdown), including different scenarios (like workers and processors that hang and don't wont to stop)
- Clusters (creating clusters, shutting down clusters)
And I think there are potentially a lot more

P.S. I haven't yet reviewed the PR fully, I'll be glad to come back when tests are in place, and potential bugs found by those tests are fixed

implementations/rust/ockam/ockam_node/src/router/start_worker.rs
```diff
@@ -0,0 +1,78 @@
...
+    router
+        .map
+        .internal
+        .insert(primary_addr.clone(), address_record);
+
+    #[cfg(feature = "std")]
+    if std::env::var("OCKAM_DUMP_INTERNALS").is_ok() {
+        trace!("{:#?}", router.map.internal);
+    }
+    #[cfg(all(not(feature = "std"), feature = "dump_internals"))]
```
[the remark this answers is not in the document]
antoinevg: `no_std` does not have env vars. Any reason we can't drop the env var and just use the feature? 

All that said, while dumping internals is handy for debugging maybe it should just be a `trace!` macro ?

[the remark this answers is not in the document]
spacekookie: Tbh I like the env var. That way you don't have to rebuild the binary. In some settings you may not be _able_ to rebuild the binary easily if you're trying to debug a fault in production.

But: I've been thinking about creating a simple log macro which hides this complexity for us

implementations/rust/ockam/ockam_node/src/messages.rs
spacekookie: The reason I didn't choose `core::time::Duration` for this is that in the public API I really wanted to avoid pushing that type on users. But we can have `Option<core::time::Duration>` in our internal representation and then have the API only take values in seconds or milliseconds or something

implementations/rust/ockam/ockam_core/src/routing/address.rs
```diff
@@ -159,6 +159,13 @@ impl Address {
     pub fn random(tt: u8) -> Self {
         Self { tt, ..random() }
     }
+
+    /// Create a new address with a suffix
+    pub fn suffix<S: Into<String>>(&self, s: S) -> Self {
```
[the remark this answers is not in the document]
spacekookie: Good to know. Although this code is gonna get removed again. I added it for an earlier prototype in this PR but ended up not needing it.

implementations/rust/ockam/ockam_node/src/router/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+                    if let Some(sender) = self.state.stop_reply() {
+                        sender.send(NodeReply::ok()).await;
+                        self.map.internal.clear();
+                        break;
+                    }
+                }
+
+                StopAck(addr) => {
+                    if shutdown::ack(self, addr).await {
+                        info!("No more workers left.  Goodbye!");
```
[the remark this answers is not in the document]
spacekookie: Why is this a log line or why does the node shut down here?

implementations/rust/ockam/ockam_node/src/router/stop_worker.rs
spacekookie: It is used, but only for aborts (i.e. if a worker hasn't shut down yet by the end of the node shutdown timeout).

spacekookie:
> @hairyhum  Is there an API for worker implementations when they get the shutdown signal, or it's supposed to be all internal?

Yes and no. The `Worker` trait has a `shutdown()` function which can be used to handle any de-allocation tasks and can also send messages to other workers (that have not started shutting down yet!).

When a worker gets interrupted there's no way for workers to run user-code.

implementations/rust/ockam/ockam_node/src/router/start_processor.rs
```diff
@@ -0,0 +1,72 @@
...
+    reply: &Sender<NodeReplyResult>,
+) -> Result<()> {
+    match router.state.node_state() {
+        NodeState::Running => start(router, addrs, senders, reply).await,
+        NodeState::Stopping(_) => reject(addrs, reply).await,
+    }?;
+    Ok(())
+}
+
+async fn start(
```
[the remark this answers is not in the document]
spacekookie: No, it's basically the same function but gated behind a match to reject the command if the executor is shutting down

implementations/rust/ockam/ockam_node/src/node.rs
```diff
@@ -37,14 +18,11 @@ pub fn start_node() -> (Context, Executor) {
...
-    let addr = "app".into();
+    let addr: Address = "app".into();
 
     // The root application worker needs a mailbox and relay to accept
     // messages from workers, and to buffer incoming transcoded data.
-    let (ctx, mb_tx) = NullWorker::create(exe.runtime(), &addr, exe.sender());
-
-    // Build a mailbox worker to buffer messages
-    let sender = WorkerRelay::<NullWorker, _>::build_root(&exe.runtime(), mb_tx);
+    let (ctx, sender, _) = Context::new(exe.runtime(), exe.sender(), addr.into());
```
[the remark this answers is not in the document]
spacekookie: The control channel is only used for workers/ processors that have a relay. The Nullworker doesn't, so in the new terminology is a "bare worker". I'll make sure to make that distinction more clear in the comments. And also there probably should be a `Context::new_bare` function which doesn't allocate a control channel at all

implementations/rust/ockam/ockam_node/src/router/utils.rs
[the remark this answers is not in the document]
spacekookie: That's a good observation. I checked and this code isn't used or exposed via our API at all. I think it's a remnant of the first prototype API. I'll remove this and the associated messages types :+1:

implementations/rust/ockam/ockam_transport_tcp/src/workers/listener.rs
```diff
@@ -38,6 +38,10 @@ impl TcpListenProcessor {
 impl Processor for TcpListenProcessor {
     type Context = Context;
 
+    async fn initialize(&mut self, ctx: &mut Context) -> Result<()> {
+        ctx.set_cluster(crate::CLUSTER_NAME).await
```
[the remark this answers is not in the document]
spacekookie: There might be a race condition here when you create the TcpTransport and immediately shut down the node. Then the worker won't have initialised before it's cleaned up.
I'm not sure how much of a concert this is but we have two options I think:

1. Make `start_worker` wait for worker initialisation to complete
2. Pass cluster name via Context API (for example `start_cluster_worker(addr, cluster_name, worker)`

The first one obviously has other implications. The second one might be a bit verbose but probably otherwise safer.

implementations/rust/ockam/ockam_node/src/router/utils.rs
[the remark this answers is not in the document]
SanjoDeundiak: I'm sure there was such check before that PR, so please make sure there is nothing left from it

implementations/rust/ockam/ockam_transport_tcp/src/workers/listener.rs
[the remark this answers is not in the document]
SanjoDeundiak: The first one will potentially hurt performance
The second one sounds good, I'll work on passing options to start_worker (it's needed by what I'm doing), maybe we'll be able to setup cluster as well

implementations/rust/ockam/ockam_node/src/router/mod.rs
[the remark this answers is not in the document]
SanjoDeundiak: It looks like this message is received when individual worker (or processor?) is stopped. So why is it saying there are no workers left?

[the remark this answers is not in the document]
spacekookie: No, the message is only printed if shutdown::ack() returns true, which is only the case when there are no more workers left to shut down (or no more clusters).

But this is maybe a bit confusing so maybe I'll encode it in a Result or proper enum if there's other variants I need to consider

mrinalwadhwa:
@spacekookie when trying example 9, did you replace these two lines with `1.node.ockam.network:4000`? 

https://github.com/ockam-network/ockam/blob/c3409f5971d588d7742a73656a51b42e33b4fb00/examples/rust/get_started/examples/09-streams-responder.rs#L9
https://github.com/ockam-network/ockam/blob/c3409f5971d588d7742a73656a51b42e33b4fb00/examples/rust/get_started/examples/09-streams-initiator.rs#L8

spacekookie:
@mrinalwadhwa I had changed those but to the IP address of a custom node I created on `hub.ockam.network` which just hung both programs. Putting in `1.node.ockam.network:4000` works.

## Updates

* `43e0bfe` changes spawn logging from info to debug
* `c5a4cd8` was needed to make the stream consumer gracefully handle node shutdown (fetch_interval will be rejected when the node is stopping because it's a spawn action)

implementations/rust/ockam/ockam_node/src/router/record.rs
```diff
@@ -0,0 +1,187 @@
...
+            ctrl_tx,
+            state: AddressState::Running,
+            meta,
+        }
+    }
+
+    /// Signal this worker to stop -- it will no longer be able to receive messages
+    pub async fn stop(&mut self) {
+        if self.meta.processor {
+            self.ctrl_tx.send(CtrlSignal::InterruptStop).await;
```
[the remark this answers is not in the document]
SanjoDeundiak: @spacekookie this test works on develop and doesn't work on your branch. Could you please recheck? If it's true, that means your changes break stop_processor
```rs
#[async_trait]
impl Processor for WaitingProcessor {
    type Context = Context;

    async fn initialize(&mut self, _context: &mut Self::Context) -> Result<()> {
        self.initialize_was_called.store(true, Ordering::Relaxed);
        assert!(self.initialize_was_called.load(Ordering::Relaxed));

        Ok(())
    }

    async fn shutdown(&mut self, _context: &mut Self::Context) -> Result<()> {
        self.shutdown_was_called.store(true, Ordering::Relaxed);
        assert!(self.initialize_was_called.load(Ordering::Relaxed));
        assert!(self.shutdown_was_called.load(Ordering::Relaxed));

        Ok(())
    }

    async fn process(&mut self, _ctx: &mut Self::Context) -> Result<bool> {
        sleep(Duration::new(1, 0)).await;

        Ok(true)
    }
}

#[allow(non_snake_case)]
#[test]
fn waiting_processor__shutdown__should_be_interrupted() {
    let initialize_was_called = Arc::new(AtomicBool::new(false));
    let shutdown_was_called = Arc::new(AtomicBool::new(false));

    let initialize_was_called_clone = initialize_was_called.clone();
    let shutdown_was_called_clone = shutdown_was_called.clone();

    let (mut ctx, mut executor) = start_node();
    executor
        .execute(async move {
            let processor = WaitingProcessor {
                initialize_was_called: initialize_was_called_clone,
                shutdown_was_called: shutdown_was_called_clone,
            };

            ctx.start_processor("waiting_processor", processor)
                .await
                .unwrap();
            sleep(Duration::new(1, 0)).await;

            ctx.stop_processor("waiting_processor").await.unwrap();

            assert!(initialize_was_called.load(Ordering::Relaxed));
            assert!(shutdown_was_called.load(Ordering::Relaxed));

            ctx.stop().await.unwrap();
        })
        .unwrap();

    // Wait till tokio Runtime is shut down
    std::thread::sleep(Duration::new(1, 0));
}
```

PR #2061, opened 2021-10-20
jared/adrian/node test proc macro

jdspdx:
Rebase of https://github.com/ockam-network/ockam/pull/1904#pullrequestreview-772885355


SanjoDeundiak:
@adrianbenavides I think the next logical step would be to use new macro for all tests and examples, then move macro to ```ockam_node_attribute```. What do you think? Do you want to work on it?

PR #2068, opened 2021-10-21
build(elixir): move default config to mix.exs file

hairyhum:
Config values from `config/config.exs` are only loaded when building
from the app directory.
If the app is built as a dependency, configuration should be specified
in the mix.exs file.


hairyhum:
To test these changes impact, use the branch name `hairyhum/ockam-app-default-config` in `examples/elixir/get_started/setup.exs` and run an example

PR #2131, opened 2021-10-30
Fix various issues in `ockam_ffi`.

### LINKED ISSUE #1562: `ockam_vault_extern_error_t`'s definition/lack of documentation might encourage memory leaks.
When filing #1561, I noticed an issue which is plausibly a bit more serious. Specifically,  `ockam_vault_extern_error_t::domain` is allocated by `CString::into_raw` (which notes that it's C's responsibility to free https://github.com/ockam-network/ockam/blob/develop/implementations/rust/ockam/ockam_ffi/src/error.rs#L37), but no API to free resources managed by the `ockam_vault_extern_error_t` is provided, nor is the fact that it's needed documented.

Moreover, this field being a `const char*` parameter implies that it'd be invalid for the caller to free it. If I saw this in a C api, I'd probably expect that it's a pointer to a static error description string.

That said, fixing this is kind of a pain, and maybe it doesn't matter in practice, since errors should be rare — I'm only careful about noticing it since my background includes a lot of FFI work, including a [very similar](https://docs.rs/ffi-support/0.4.3/ffi_support/struct.ExternError.html) type.

thomcc:
I had some computer issues this morning that I fixed by updating the OS, which meant that I wasn't on my primary dev machine for the day (I forgot to switch back after the update completed).

As a result of this, I decided to do my audit of the FFI code, which I had been meaning to do for a while (since at Mozilla, this kind of FFI code was my responsibility, so I know some non-obvious pitfalls to watch out for, and the solutions to them).

Fixes #1562, as well as the fact that we unwind from `extern "C"` fns, which is UB (fixing this robustly is a bit tricky, but not *too* bad — doing this correctly is a bit involved, but the implementation here should be robust.

I also cleaned up a use of `#[cfg(feature = "bls")]` when there's no `bls` feature, which is not that important, but caused a false positive error in my editor.

### DIFF
--- a/implementations/rust/ockam/ockam_ffi/src/vault.rs
+++ b/implementations/rust/ockam/ockam_ffi/src/vault.rs
@@ -60,17 +60,18 @@ async fn get_vault(context: FfiVaultFatPointer) -> Result<VaultMutex<SoftwareVau
 /// Create and return a default Ockam Vault.
 #[no_mangle]
 pub extern "C" fn ockam_vault_default_init(context: &mut FfiVaultFatPointer) -> FfiOckamError {
-    // TODO: handle logging
+    handle_panics(|| {
+        // TODO: handle logging
+        let handle = block_future(async move {
+            let mut write_lock = SOFTWARE_VAULTS.write().await;
+            write_lock.push(VaultMutex::create(SoftwareVault::default()));
+            write_lock.len() - 1
+        });
 
-    let handle = block_future(async move {
-        let mut write_lock = SOFTWARE_VAULTS.write().await;
-        write_lock.push(VaultMutex::create(SoftwareVault::default()));
-        write_lock.len() - 1
-    });
+        *context = FfiVaultFatPointer::new(handle as u64, FfiVaultType::Software);
 
-    *context = FfiVaultFatPointer::new(handle as u64, FfiVaultType::Software);
-
-    FfiOckamError::none()
+        Ok(())
+    })
 }
 
 /// Compute the SHA-256 hash on `input` and put the result in `digest`.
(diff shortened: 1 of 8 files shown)


implementations/rust/ockam/ockam_ffi/src/vault.rs
```diff
@@ -423,53 +420,96 @@ pub extern "C" fn ockam_vault_aead_aes_gcm_decrypt(
...
+        })?;
+        Ok(())
     })
 }
+
+fn handle_panics<F>(f: F) -> FfiOckamError
+where
+    F: FnOnce() -> Result<(), FfiOckamError>,
+{
+    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(f));
```
thomcc, line 482: AssertUnwindSafe here doesn't really matter — `UnwindSafe` is dubious at best, is unrelated to safety, and we aren't in a position to properly represent it anyway (as propagating the unwinding would be UB).

implementations/rust/ockam/ockam_ffi/src/vault.rs
```diff
@@ -423,53 +420,96 @@ pub extern "C" fn ockam_vault_aead_aes_gcm_decrypt(
...
+}
+
+/// Aborts on drop, used to guard against panics in a section of code.
+///
+/// Correct usage should `mem::forget` this struct after the non-panicking
+/// section.
+struct AbortOnDrop;
+impl Drop for AbortOnDrop {
+    fn drop(&mut self) {
+        eprintln!("Panic from error drop, aborting!");
```
thomcc, line 512: In principal, this `eprintln!` could panic (if stderr is closed, for example), but this Drop can only be called in response to a panic, and Rust calls `core::intrinsics::abort` if it runs a `Drop` that panics while it's already panicking.

That is to say, we could implement this `Drop` as `panic!("Panic from error drop, aborting!")`, but this has notably better user experience, as `core::intrinsics::abort` is a SIGILL (on x86), not a SIGABRT.

jared-s:
Merged as #2138 - thanks!

PR #2153, opened 2021-11-02
ci: release process document

jared-s:
Creates [RELEASE.md](https://github.com/ockam-network/ockam/blob/e49e6613ce3e99d027f60b8963fcfc808e829d77/tools/scripts/release/RELEASE.md) which documents our existing release process, proposed new process, and existing concerns.

- [Scripts directory](https://github.com/ockam-network/ockam/tree/e49e6613ce3e99d027f60b8963fcfc808e829d77/tools/scripts/release)
- Old script: [release.sh](https://github.com/ockam-network/ockam/blob/e49e6613ce3e99d027f60b8963fcfc808e829d77/tools/scripts/release/release.sh)
- Old way of finding changes: [crate-changes.sh](https://github.com/ockam-network/ockam/blob/e49e6613ce3e99d027f60b8963fcfc808e829d77/tools/scripts/release/crate-changes.sh)
- New script: [crate-publisher.sh](https://github.com/ockam-network/ockam/blob/e49e6613ce3e99d027f60b8963fcfc808e829d77/tools/scripts/release/crate-publisher.sh)

### DIFF
--- a/tools/scripts/release/RELEASE.md
+++ b/tools/scripts/release/RELEASE.md
@@ -1,0 +1,175 @@
+
+# Current Process
+
+Steps:
+
+- Create a release branch
+- Enumerate per-crate incoming Changelog entries
+- Create a release plan (list of crates to be published, in dependency order)
+- Increment version number for changed crates and crates with changed dependencies.
+- Update READMEs
+- Update dependency versions in Cargo.toml
+- Integrate incoming Changelog entries into existing Changelogs
+- Merge to develop
+- Merge to main
+- `cargo publish` crates in dependency order
+- `gh release` crates. Release `ockam` crate last.
+
+Details:
+
+### Incoming Changes
+
+To find what has changed between releases for each crate, I use a script that parses `git log` output.
+
+New files named `Changelog-INCOMING.md` are created for each crate. This file has a header that has the correct version, date, and header structure. Below the header is a list of commit messages that affected the crate.
+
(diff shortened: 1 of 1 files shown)


tools/scripts/release/RELEASE.md
```diff
@@ -0,0 +1,175 @@
...
+  - One script to prepare the release branch
+  - One script to publish to GitHub and crates.io
+
+## Release branch script
+- Remove all git operations other than those used to build Changelogs.
+- Remove ordered list of crates and use the directory structure of the repo to determine crates. For this script, dependency order is not important.
+- Use existing `crate-changes.sh` script to automatically populate Changelogs
+- Increment version numbers
+- Update READMEs and dependant `Cargo.toml`.
+- The end result should be changes staged to a release branch that includes all version increment work.
```
metaclips, line 161: 👍 Thanks Jared. We can start with this.

metaclips (commented):
Thanks for the thorough review. Looking at the current `crate-publisher.sh` we should be able to further reduce commands.
As suggested, we can reduce the script to only update the readme and cargo version and deps, since those are manually done currently.
We should also further investigate how we can use `cargo release workspace` feature to fully automate our release process and also integrate it with our Changelog generation.
With the ability to `exclude` crates we don't want to release, automatic `crate publish` order and also incoming feature to [exclude crates that don't have changes comparing it to tags](https://github.com/crate-ci/cargo-release/pull/296), we should be able to create a leaner script that does the following with few commands.

- Generate Changelog with `git commit`
- Call `cargo release` to update version, readme, and dependencies of all crates.

PR #2156, opened 2021-11-03
chore: add "no_main" feature to "node" macro of ockam_node_test_attribute crate

adrianbenavides:
Add changes from 184d74049b9ddc92c8d2c523fc44be9601617036 to the `ockam_node_test_attribute` crate.

Refactor `node` macro so it can infer the `Context` and `Result` type names. Now the functions using this macro can be defined with/without a `Result` type and with any alias (`ockam::Result` or just `Result`). Also, the `Context` can be defined as just `Context` or `ockam::Context`. The current `ockam::node` macro doesn't need this because it copies/pastes the type paths as arguments of the `trampoline` function. In this case, we need to define them explicitly when creating the variables.

Finally, I've done a refactor in the `hygiene` module, separating input checks and ctx checks into their own structs to make the code easier to navigate. Some tests have slightly changed due to the checks ordering, but overall they are more explicit in what they test now.

I have uncommitted changes using the `ockam_node_test_attribute::node` macro on all get_started examples and it works as expected, I can create a PR with those changes after these 2 PR's get merged. As far as I know, the `ockam::node` macro is not used internally anywhere right? 

## Related
* Follow up of https://github.com/ockam-network/ockam/pull/2155


implementations/rust/ockam/ockam_node_test_attribute/src/parser.rs
```diff
@@ -12,36 +12,55 @@ pub(crate) fn node(
...
-            executor.execute(async move {
-                #body
-            })
+        fn ockam_async_main() #ret_type {
+            let (mut #ctx_ident, mut executor) = ockam_node::start_node() as (#ctx_type, ockam::Executor);
+            executor.execute(async move #body)#err_handling
         }
-        main().unwrap();
+        // TODO: safe way to print the error before panicking?
+        ockam_async_main().unwrap();
```
adrianbenavides, line 63: Any ideas here?

SanjoDeundiak (approved):
Nice!

jared-s:
Merged https://github.com/ockam-network/ockam/pull/2175 - thanks!

PR #2158, opened 2021-11-03
chore(rust): replace `colored` with `owo-colors` in `ockam-hub-cli`

thomcc:
Specifically `colored`, which is used by `ockam-hub-cli` (and some largely-bitrotted example code that we need to move to a branch).

I replaced it with `owo-colors` (MIT licensed), which is almost a drop-in replacement and is probably the most popular lib for this for new code. That said, I don't love that library (it spews garbage on certain terminals), but it's fairly widely used and is fine for now[^1]

[^1]: Terminal coloring (particularly doing it in a way that doesn't have a terrible UX for an unsupported terminal) is actually an area that's very [near](https://shift.click/dump/ansi-palette) and [dear](https://gitlab.com/thomcc/near-ansi) to my heart, and I have a WIP project I'll try and finish up this week (or this evening?) that should be usable as a longer term replacement.

This also makes a couple changes to `ockam-hub-cli`'s Cargo.toml so that we successfully pass `cargo deny check licenses` when run on the whole workspace. I'll leave it to @metaclips to integrate that into our CI.

Oh, and the update to the NOTICE file was done by hand. I have no clue how it was generated, TBH.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -286,17 +286,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "colored"
-version = "2.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b3616f750b84d8f0de8a58bda93e08e2a81ad3f523089b05f1dffecab48c6cbd"
-dependencies = [
- "atty",
- "lazy_static",
- "winapi",
-]
-
-[[package]]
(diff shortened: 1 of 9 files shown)


jared-s:
Thanks, merged as https://github.com/ockam-network/ockam/pull/2159

PR #2181, opened 2021-11-09
fix(rust): break circular dependency between ockam_macro and ockam_core in dev deps

jared-s:


thomcc:
Hm, that's kinda unfortunate. I actually had thought that making this a dev-dep wasn't allowed either? Probably we should eventually put them in `ockam_node`, or in a new `ockam_test` crate.

TBH, I'm not 100% convinced the node attribute's error messages (and thus the trybuild tests they require) are worth the trouble — when I've seen them it felt worse than what I suspect rustc would give us, but I guess I might be wrong.

PR #2196, opened 2021-11-11
refactor: merge macro crates

adrianbenavides:
* https://github.com/ockam-network/ockam/issues/2186

As described in the above issue, the different macro subcrates have been merged in three new crates:
- ockam_macros: async_try_clone_derive, message_derive, node_attribute
- ockam_test_macros: node_test_attribute
- ockam_test_macros_internal: vault_attribute

Also, replaces old crates names by new ones in github workflows: 
- ockam_macro, ockam_message_derive and ockam_node_attribute  -> ockam_macros and delete duplicates
- ockam_node_test_attribute -> ockam_test_macros
- ockam_vault_test_attribute -> ockam_test_macros_internal


implementations/rust/ockam/ockam_examples/example_projects/tcp/Cargo.lock
```diff
@@ -0,0 +1,1458 @@
+# This file is automatically @generated by Cargo.
```
SanjoDeundiak, line 1: Not sure why this lock file appeared here. Could you please remove it?

implementations/rust/ockam/ockam_core/Cargo.toml
```diff
@@ -32,7 +32,7 @@ no_std = ["core2", "hex", "rand_pcg", "spin"]
 alloc = ["core2/alloc", "heapless", "hex/alloc", "serde/alloc", "serde_bare/alloc"]
 
 [dependencies]
-ockam_message_derive = { path = "../ockam_message_derive", version = "0.1.0"     , default_features = false }
+ockam_macros = { path = "../ockam_macros", version = "0.1.0"     , default_features = false }
```
SanjoDeundiak, line 35: Isn't there a circular dependency?

SanjoDeundiak (changes requested):
Tnx for the PR. Looks great, only few things to fix

implementations/rust/ockam/ockam_examples/example_projects/tcp/Cargo.lock
adrianbenavides: Oops, so many files :laughing: I'll remove it

implementations/rust/ockam/ockam_core/Cargo.toml
adrianbenavides: I created the PR to run the github workflows and see if everything compiles. I tried compiling a few crates on my computer, but I haven't had the time yet to run a full test. I also want to add some context to the issue description. I'll try to do it as soon as possible. Thanks sanjo for your time!

adrianbenavides:
I have a question about github workflows. The `node` attribute tests from the `ockam_macros` are only executed if the `no_main` flag is disabled but they are not currently run. How are feature flags configured in github workflows?

implementations/rust/ockam/ockam_core/Cargo.toml
adrianbenavides: `ockam_macros` depend on `ockam` on `dev_dependencies`, which, as far as I know, is not a problem as the tests are run against the compiled version without the dev deps. I'm probably missing something.

thomcc: Hm, I think this can still cause issues during publish. I'd prefer you move the tests into the ockam crate, breaking the cycle.

thomcc (commented):
This looks good to me with the cycles broken. I think Jared should take a look just to avoid any issues that will cause trouble during release time.

Also, I think we eventually should move no_main from a feature into an argument, e.g. `#[ockam::node(no_main)]`, but that shouldn't be done as part of this PR.

implementations/rust/ockam/ockam_core/Cargo.toml
adrianbenavides: OK done! Now none of the macros crates has a dependency on ockam/core/node.

thomcc (approved):
This looks good to me, thanks for the updates!

jared-s (approved):

thomcc:
Hm, after some discussion internally, I think we might want to make a few changes here. Sorry for changing our mind on this, since it's slightly different than what you've done. Concretely, we should:

1. Merge the ockam_test_macros and ockam_macros, putting the test macros behind a `#[cfg(feature = "test")]` (note that this couldn't be `#[cfg(test)]`, which is only set when compiling the crate being tested, and thus not our macro crate).

2. Rename the `ockam_test_macros_internal` to `ockam_macros_internal` (in practice it should only contain test code, and this shouldn't[^1] be published)

Does that make sense? Again, apologies for going back on what we said earlier.

[^1]: We should probably do something in the future to prevent us from publishing ockam_macros_internal, if possible (`publish = false` in the cargo.toml would do this, but it might interact poorly with our scripts, and in the short term it being in the dev deps might mean we have to publish — I don't remember).

thomcc (changes requested):
See comment above, which gives a thorough explanation of the changes we want.

adrianbenavides:
> See comment above, which gives a thorough explanation of the changes we want.

No problem @thomcc, that's what we are for!

> Rename the ockam_test_macros_internal to ockam_macros_internal (in practice it should only contain test code, and this shouldn't be published)

To keep it consistent with `ockam_macros` I'd also include the `test` feature here. I think it gives the necessary context to understand where the macros will be used without having to dig into the code. If I were new to the codebase, it would be confusing the see a macros crate with test attributes behind a feature and another macros crate, also with test attributes, without it. Let me know what you think!

mrinalwadhwa:
Merged

Issue #2088, opened 2021-10-25
Improve node shutdown


spacekookie:
Current PR #2007.

Currently we are experiencing a lot of race conditions during node shutdown. The impact of this might be mild in production systems but is creating a lot of noise in our CI.

This is only part 1 of a series of node improvements.

## Update on #2007 

- [x] Graceful shutdown mechanism implemented in `ockam_node`
- [x] Decoupling "node shutdown" and "single worker shutdown" mechanisms
- [x] Verify all examples still work 
- [x] Minor code cleanups

PR #2197, opened 2021-11-11
Implement Rust pipes

spacekookie:
This is an initial implementation of pipe workers in Rust.  It consists of three main parts.

1. Basic `PipeSender` and `PipeReceiver` worker pair and an API to create them
2. A function composition behaviour API to stack different pipe behaviours on top of each other (each for internal and external message types)
3. Basic resend/ ack and strict ordering enforcement logic

Two unrelated changes

* `656db3244adc24043170b024aa09f45eca7bcb8e` adds a trace-level log line when a worker relay receives a message to make dead-locks easier to spot in logs
* `b60fb4386c30a5f2fd3a65f0fa731e493deb23f2` fixes the illusive bug where the router would sometimes crash on shutdown. The need for an immediate shutdown was fixed by #2088, but this fixes the issue when calling `ctx.stop_now().await`


jared-s:
> This PR adds a new dependency on [`dyn-clone`](https://github.com/dtolnay/dyn-clone) btw! From what I can tell it shouldn't impact our `no_std` implementations but I thought I'd more publicly document it anyway! (ping @antoinevg @jared-s)
> 
> Otherwise, if the CI gods be willing, I think this code is ready to be merged

Ran the no_std examples. all good!

PR #2207, opened 2021-11-15
fix(rust): fix no_std builds

antoinevg:
## Proposed Changes

This PR fixes various things that are currently breaking `no_std` builds:

* `ockam_executor` deadlocks inside the `ProcessorRelay` run loop.
* `ockam_executor` is missing an implementation of the `warn!` logging macro
* we don't (yet) have an implementation of `select!` for `ockam_executor`
* adding `futures-util` with default features enabled breaks `no_std` builds

## Checks

```
cd implementations/rust/ockam/ockam

cargo build --no-default-features --features "no_std alloc software_vault"
```

For bonus points arm cortex-m builds can be tested with:

```
cargo +nightly build --target thumbv7em-none-eabihf --no-default-features --features "no_std alloc software_vault"
```


SanjoDeundiak (approved):

spacekookie (approved):
LGTM

Issue #2112, opened 2021-10-26
Worker tracing tools for elixir nodes


hairyhum:
Currently elixir workers report events using `telemetry` which are used in metrics collection for ockam hub nodes.

It would be useful for debugging to be able to enable some sort of logging mechanism for those events. 
This mechanism could use `telemetry` because it's already used for reporting.

It would be convenient for this mechanism to be controlled from the runtime.

PR #2212, opened 2021-11-16
chore(elixir): add logging telemetry events

feniks65:
## Current Behaviour

Currently elixir telemetry module is not logging events that are going through it.

## Proposed Changes

Introduce logging of events that go through telemetry module.

@hairyhum 
While reading the description of #2112 I thought 'hey, would it be  sufficient to just use Logger in this case?'.
And I have added put them before calling telemetry.execute() in all places.

The example of the print from Logger looks as follows:
```
runner@7b3926dbdbc7:/work/implementations/elixir/ockam/ockam$ iex -S mix
Erlang/OTP 23 [erts-11.1.5] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [hipe]


23:54:38.567 module=Ockam.Telemetry line=54 pid=<0.245.0> [debug] emit_event() event_name: [Ockam.Router, :start_link], measurements: %{}, metadata: %{options: [], return_value: {:ok, #PID<0.246.0>}}
```

However, I am not sure about enabling and disabling it in the runtime. For sure severity of logs can be tuned, but I don't know if this would be sufficient in this case?

## Checks

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam.ex
@@ -15,6 +15,15 @@ defmodule Ockam do
   def start(_type, _args) do
     if Code.ensure_loaded?(:telemetry) do
       {:ok, _apps} = Application.ensure_all_started(:telemetry)
+      events = [
+        [:ockam, Ockam.Router, :start_link],
+        [:ockam, :decode_and_send_to_router],
+        [:ockam, :encode_and_send_over_udp],
+        [:ockam, :init],
+        [:ockam, :handle_info]
+      ]
+      :telemetry.attach_many("logger", events, &Ockam.TelemetryLogger.handle_event/4, nil)
+      {:ok, _apps} = Application.ensure_all_started(:telemetry)
     end
 
     # Specifications of child processes that will be started and supervised.
--- a/implementations/elixir/ockam/ockam/lib/ockam/telemetry.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/telemetry.ex
@@ -1,5 +1,6 @@
 if Code.ensure_loaded?(:telemetry) do
   defmodule Ockam.Telemetry do
+
     @moduledoc """
     Provides functions to emit `:telemetry` events.
     """
--- a/implementations/elixir/ockam/ockam/lib/ockam/telemetry_logger.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/telemetry_logger.ex
@@ -1,0 +1,11 @@
+if Code.ensure_loaded?(:telemetry) do
+  defmodule Ockam.TelemetryLogger do
+    require Logger
+
+    def handle_event(event, measurements, metadata, _config) do
+      Logger.info(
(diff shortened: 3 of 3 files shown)


mrinalwadhwa:
@feniks65 thank you for send the PR!

I wonder if a more flexible approach would be to create a module that would make it easy to attach a handler to a predefined set of telemetry events.

Something like this:

```elixir
handler = fn event, measurements, metadata, _conf ->
  IO.puts("\n\n===> \n#{inspect(event)}, \n#{inspect(measurements)}, \n#{inspect(metadata)}")
end

events = [
  [:ockam, Ockam.Router, :route, :start],
  [:ockam, Ockam.Router, :route, :stop]
]

:ok = :telemetry.attach_many("logger", events, handler, nil)
```

What do you think? @feniks65 @hairyhum

hairyhum:
@feniks65 @mrinalwadhwa I was also thinking of something more like telemetry handler, like mentioned here: https://github.com/beam-telemetry/telemetry, so we can enable and disable this feature.

This would allow us apart from enable/disable also control which messages do we want to log.

feniks65:
@hairyhum @mrinalwadhwa 

How about now?
I have create additional `TelemetryLogger` module with `handle_event()` function.
Also in ockam.ex I put there `:telemetry.attach_many()` with the following events to be intercepted:
`       [:ockam, Ockam.Router, :start_link],
        [:ockam, :decode_and_send_to_router],
        [:ockam, :encode_and_send_over_udp],
        [:ockam, :init],
        [:ockam, :handle_info]`
but I am wondering if these are all events that should be logged?
The list was prepared based on my the output from the grep call which listed all calls to 'Telemetry.emit_event...'.

hairyhum:
@feniks65 that's a good start.
I think we're mostly interested in `[:ockam, Ockam.Router, :route, :stop]`, `[:ockam, <worker_module>, :init, :stop]` and `[:ockam, <worker_module>, :handle_message, :stop]` events.

Each worker module has its own event name, so I believe you have to add them all.

You can get all worker module names using something like:
```
Application.loaded_applications() # from all loaded applications
  |> Enum.flat_map(fn({app, _, _}) -> Application.spec(app, :modules) end) # list all modules
  |> Enum.filter(fn(module) -> String.starts_with?(to_string(module), "Elixir.Ockam") end) # only take those starting with Ockam.
  |> Enum.filter(fn(module) ->
    Enum.member?(
      module.module_info(:attributes) # from module attributes
        |> Keyword.get_values(:behaviour) # get all behaviours
        |> List.flatten,
      Ockam.Worker) # contains Ockam.Worker
  end)
```

This will take all the Elixir modules starting with `Ockam` and check if they implement `Ockam.Worker`

That's not too pretty and we can figure out a better way to do that, but for now this should do.

Can you also move attaching of this event to a function, maybe in `TelemetryLogger` module and add a function to detach it as well, so it's easier to do that in runtime?

Also maybe add some sort of log message to indicate that this is a telemetry event instead of `===> `

hairyhum:
Merged in #2675

Issue #1804, opened 2021-08-30
Ockam TCP Transport using smoltcp


mrinalwadhwa:
https://github.com/smoltcp-rs/smoltcp

---
_If you would like someone to help / mentor / pair program with you as you work on your first Ockam issue, please leave a comment on [this discussion.](https://github.com/ockam-network/ockam/discussions/1081)_

nihalpasham:
Indeed, this a great starting point. I guess, the next step is how do we layer `ockam-transport` on top of such an example? Do we have any examples for porting ockam onto a different transport layer?

mrinalwadhwa:
We'll need to write a new transport crate - maybe we call it `ockam_transport_tcp_ smoltcp`

[`ockam_transport_tcp`](https://github.com/ockam-network/ockam/tree/develop/implementations/rust/ockam/ockam_transport_tcp) is a good reference for a transport implementation.

All the crates are [here](https://github.com/ockam-network/ockam/tree/develop/implementations/rust/ockam), `ockam_transport_websocket` is a good reference too.

nihalpasham:
Okay, I finally managed to get down to cloning the repo, today. I'm still trying to wrap my head around a couple of things, so I could be wrong here but here's what I've gathered. 

To build the `ockam-transport-smoltcp` impl while retaining the existing api (more or less), we'll need 3 things

- `no_std ockam-node impl:` 
	- guess, we have a barebones impl for this i.e. a global executor to poll and execute futures. 
	- For the absolute simplest impl, I guess we can simply collect all futures into a single `async` block and pass them to the executor. 
	- And then use another (maybe interrupt driven) mechanism to register with a `waker` for notifications. 
- `high-level ockam-transport layer wrapper: `
	- This layer just spawns 3 ockam-workers - TCPRouter, TcpSendWorker, TcpRecvProcessor
	- I presume this layer doesn’t need to change much i.e. we'll just need new names for the types (maybe)
	- Question here - what are `inlets` and `portals` actually used for i.e. are they required to establish TCP connections (i.e. connect, listen, send, receive)? 
- `replace tokio async-wrappers used to instantiate tcpsockets/streams with a smoltcp abstraction:`
	- the high-level ockam-transport layer calls into tokio to get a TCPstream and send/receive data. 
	- we'll need a similar api around `smoltcp`

Does this make sense i.e. is my understanding correct or am I missing something? 

[...]

PR #2213, opened 2021-11-16
Add smoltcp TCP transport 

conectado:
# Implementing feature discussed in #1804 

## Description

**This PR includes**:
* An implementation of the tcp transport protocol using `smoltcp`.
* Abstract common behavior between the `smoltcp` and `std` implementation into `ockam_transport_core`.

**This PR does NOT include**:
* A portal and inlet/outlet implementation for `smoltcp`.
* A `no_std` implementation of the `ockam_transport_smoltcp` traits needed to run an example in a `no_std` device. (Although it does compile in `no_std`).

## Implementation notes

* [`ockam_transport_core::tcp`](https://github.com/conectado/ockam/tree/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_core/src/tcp) includes the extracted behavior from `ockam_transport_tcp`.
  * [`worker`](https://github.com/conectado/ockam/tree/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_core/src/tcp/workers) are the same workers as they were in the old `ockam_transport_tcp` crate but using the new traits to work in `no_std` with smoltcp.
  * [`router`](https://github.com/conectado/ockam/tree/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_core/src/tcp/router) includes the old definition of `router` and `handle` as in the old `ockam_transport_tcp`, save for the methods used by portals and inlets/outlets, those were [extended by a newtype pattern](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_tcp/src/handle.rs#L10) in the `ockam_transport_tcp` crate.
  * [`traits`](https://github.com/conectado/ockam/tree/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_core/src/tcp/traits) Includes the definition of the behavior needed from tokio's socket to implement the workers(and `endpoint_resolver` to resolve endpoints with the `router`). Additionally:
    * [`io`](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_core/src/tcp/traits/io.rs)  Includes a defintion of the async `io` traits, copied from `futures`, that is compatible with `no_std`(plus the necessary extensions), this is needed because neither `tokio` nor `futures` have a `no_std` compatible `AsyncRead` and `AsyncWrite` due to their use of [`std::io::Error`](https://doc.rust-lang.org/std/io/struct.Error.html). 
* [`ockam_transport_smoltcp`](https://github.com/conectado/ockam/tree/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp) includes the implementation of the protocol using smoltcp. The surface structure is basically the same as `ockam_transport_tcp` using a [`SmolTcpTransport`](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/transport.rs#L114-L117) instead of `TcpTransport` with some differences needed by additional configuration for smoltcp and the new method [`get_stack`](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/transport.rs#L267) to obtain the stack for polling manually.
  * [`port_provider`](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/port_provider.rs) is needed to abstract the behavior of obtaining a port that can be pretty different in each implementation(For now I decided to make the `PortProvider` trait static to be able to hide it inside the smoltcp's [`EndpointResolver`](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/transport.rs#L272) and then an implementation can use an static global if it needs to keep track of some state, this way the `PortProvider` doesn't exist for the normal tcp implementation).
  * [`net`](https://github.com/conectado/ockam/tree/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net) here is most of the interesting things for this crate, providing a wrapper for `smoltcp` related things. (A lot of the implementation was based on [embassy](https://github.com/embassy-rs/embassy))
    * [`stack`](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net/stack.rs) provides a tcp stack. 
      * [`StackFacade`](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net/stack.rs#L124) provides type safety to make sure the stack is initialized before being used. (Note: I only managed to [panic](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net/stack.rs#L169-L173) to prevent double initialization any idea to give type safety to prevent this is welcomed!)
    * [`tcp`](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net/tcp.rs) Provides an interface similar to tcp's sockets.
    * [`timer`](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net/timer/mod.rs) provide a monotonic clock trait that is needed by the stack(More discussion about this in this PR comments!).
    * [`device`](https://github.com/conectado/ockam/tree/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net/device) Provides a `Device` wrapper around [smoltcp's `Device`](https://docs.rs/smoltcp/0.8.0/smoltcp/phy/trait.Device.html) to handle explicitly the wakers for async. (This module also includes the [tuntap](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net/device/tuntap.rs) implementation)
* I tried not to use `unsafe` unless absolutely needed(only for the tuntap interface), even if it meant having mutexes wrapping other mutexes, so that those can be added in a separate PR.
* I tried to prevent allocations to make it easier to stop depending on `alloc` in the future except when the implementation would be much harder or would require `unsafe`.(I figured it'd be okay requiring `alloc` since the whole project requires `alloc` anyways) I tried to leave a `todo` comment over each allocation.
* I didn't include an integration test like in [the tcp implementation](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net/device/tuntap.rs) because that would either require creating programatically a tuntap interface or mocking the [`Device`](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net/device/tuntap.rs). Leaving this for a future PR.

## Test the PR

* To test that `ockam_transport_smoltcp` compiles in `no_std`, even if there is no example doing this:
  `cd implementaitons/rust/ockam/ockam_transport_smoltcp && cargo +nightly check --target thumbv7em-none-eabihf --no-default-features --features="no_std, alloc, pool-32"`  
* To test the PR functioning, I added an example with a [README](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net/device/tuntap.rs) on how to run it.

## Next steps
* Add at least one `no_std` implementation of:
  * [`clock`](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net/device/tuntap.rs)
  * [`Device`](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net/device/tuntap.rs). Leaving this for a future PR.
  * [`PortProvider`](https://github.com/conectado/ockam/blob/conectado/tcp-transport-using-smoltcp/implementations/rust/ockam/ockam_transport_smoltcp/src/net/device/tuntap.rs)
* Use [`embedded_nal`](https://docs.rs/embedded-nal/latest/embedded_nal/) as suggested [here](https://github.com/ockam-network/ockam/pull/2213#issuecomment-992253547), this probably would let us replace some of the `traits` in `ockam_tcp_core`.
* Extract portal and inlet/outlet behavior into `ockam_tcp_core` and add an implementation for smoltcp. 
* Add the integration test mentioned before.
* Add `unsafe` sections to prevent wrapping mutexes within mutexes and allocations.

## Checks


thomcc:
This looks great so far, please don't hesitate to ask if you have any questions!

mrinalwadhwa:
@conectado thank you for continuing to work on this.
Let us know if you need help or have any questions along the way.

conectado:
> @conectado thank you for continuing to work on this. Let us know if you need help or have any questions along the way.

Thanks 😄 For now things are going quite smoothly, I will start working on making it work with `no-std`. Will let you know if I've any difficulties.

conectado:
@mrinalwadhwa @thomcc I'm in the middle of changing the implementation to support `no-std` (I'm starting with a version that needs `alloc`) and when [polling the ethernet interface](https://github.com/ockam-network/ockam/blob/60630dbc92ec93b508d5ed14ae070dea44b4c254/implementations/rust/ockam/ockam_transport_smoltcp/src/net/stack.rs#L126) for new packages I need to pass [the current `Instant`](https://github.com/ockam-network/ockam/blob/60630dbc92ec93b508d5ed14ae070dea44b4c254/implementations/rust/ockam/ockam_transport_smoltcp/src/net/stack.rs#L123) which is the number of milliseconds that passed since startup. However, the method `Instant::now` provided by smoltcp only works in `std`, **I was wondering if there's something already in the repo to deal with time in `no-std` enviroment**.

If there's not I think there are a couple of options on how to advance with this:
* We could use [`embassy-rs`](https://github.com/embassy-rs/embassy) on which I based most of the async implementation for smoltcp and [it provides a `timer` implementation that we could use](https://github.com/embassy-rs/embassy/blob/master/embassy/src/time/instant.rs#L20-L24). The disadvantages of using embassy are:
  * It doesn't have a release in crates.io yet so we have to depend on an specific commit
  * It only works on nightly (But it seems that the `no_std` examples in ockam only works on nightly too, so that might not be a problem)
* We could leverage the types and traits provided by [embedded-time](https://github.com/FluenTech/embedded-time) this could be beneficial also because there seems to be an on-going effort to include it into the embedded-hal crate(see rust-embedded/embedded-hal#211)
* We could create our own traits and types as simple as possible for this case. 

For now I'll try going with the third option(at least to understand the problem better and decide later) but any kind of tip or insight is appreciated as I don't have any experience with this topic 🙏

thomcc:
@antoinevg might have some insight onto the 2nd option, but in general the third option sounds best for now, if possible, since I believe we'd like not to only be able to use this in embedded (even if that is the primary use case).

conectado:
> @antoinevg might have some insight onto the 2nd option, but in general the third option sounds best for now, if possible, since I believe we'd like not to only be able to use this in embedded (even if that is the primary use case).

I think we could use smoltcp's `Instant::now` when `std` is on and use one of the other options when that's not the case.

thomcc:
Ah, that would work.

antoinevg:
> when [polling the ethernet interface](https://github.com/ockam-network/ockam/blob/60630dbc92ec93b508d5ed14ae070dea44b4c254/implementations/rust/ockam/ockam_transport_smoltcp/src/net/stack.rs#L126) for new packages I need to pass [the current `Instant`](https://github.com/ockam-network/ockam/blob/60630dbc92ec93b508d5ed14ae070dea44b4c254/implementations/rust/ockam/ockam_transport_smoltcp/src/net/stack.rs#L123) which is the number of milliseconds that passed since startup. However, the method `Instant::now` provided by smoltcp only works in `std`

Everyone will tend to have their own way of handling system time for their particular project so I'd probably try to steer away from making that decision for them.

For myself, I usually handle this by defining an atomic integer and then setting a `SysTick` exception or similar to occur every 1ms to increment it.

See, for example:

https://github.com/antoinevg/nucleo-h7xx/blob/1e3aa9bae284da845c5f7a76c2ce192cf91d91ef/examples/ethernet_hal.rs#L195-L222
 
I'd be perfectly happy if you just passed smoltcp's timestamp requirement on to the developer. 

i.e. requiring developers to pass it in via your `Stack` poll method as `poll_iface(self: &Arc<Self>,  cx: &mut Context, timestamp: Instant)` or somesuch.

conectado:
> > when [polling the ethernet interface](https://github.com/ockam-network/ockam/blob/60630dbc92ec93b508d5ed14ae070dea44b4c254/implementations/rust/ockam/ockam_transport_smoltcp/src/net/stack.rs#L126) for new packages I need to pass [the current `Instant`](https://github.com/ockam-network/ockam/blob/60630dbc92ec93b508d5ed14ae070dea44b4c254/implementations/rust/ockam/ockam_transport_smoltcp/src/net/stack.rs#L123) which is the number of milliseconds that passed since startup. However, the method `Instant::now` provided by smoltcp only works in `std`

> I'd be perfectly happy if you just passed smoltcp's timestamp requirement on to the developer.
> 
> i.e. requiring developers to pass it in via your `Stack` poll method as `poll_iface(self: &Arc<Self>, cx: &mut Context, timestamp: Instant)` or somesuch.

Since I  [spawn the task to poll the interface](https://github.com/ockam-network/ockam/blob/60630dbc92ec93b508d5ed14ae070dea44b4c254/implementations/rust/ockam/ockam_transport_smoltcp/src/net/stack.rs#L147-L156) myself I could ask for the developer a struct when starting the `Stack` (`Stack::run(&self, clock: impl Clock)`) implementing a simple trait:

```rs
type MIlliseconds = i64;

trait Clock {
  type Instant: Into<Milliseconds>
  fn now(&self) -> Self::Instant;
}
```
Or do you think it is better if I let starting the task for polling to the developer too?

antoinevg:
> Or do you think it is better if I let starting the task for polling to the developer too?

It may also be useful to give the developer the option to do even the polling themselves.

e.g. On embedded it's best to poll `smoltcp` in the `[ethernet]` interrupt handler and that can be set up very differently depending on how the app is structured.

In the longer run maybe we should even be looking at something like:

https://crates.io/crates/embedded-nal

All that said, there are so many open questions around async/embedded that I think almost any approach that results in running code is valuable right now because it helps us all understand the space better!

conectado:
> > Or do you think it is better if I let starting the task for polling to the developer too?

> All that said, there are so many open questions around async/embedded that I think almost any approach that results in running code is valuable right now because it helps us all understand the space better!

Alright! Then, I will advance with my current approach, I think adding the possibility for the developer to `poll` the interface themselves is easy so I will add that too for now and later on revisit to use `embedded-nal` if possible.

mrinalwadhwa:
@conectado awesome work! Thank you ❤️ 
Looking forward to merging this after we get some reviews.

Cargo.lock
```diff
@@ -204,9 +244,9 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "1.1.0"
+version = "1.0.1"
```
mrinalwadhwa, line 247: We may need to update the Cargo.lock before merging since this diff bumps several dependency versions down from what is in develop.

conectado: Fixed [here](https://github.com/ockam-network/ockam/pull/2213/commits/9cd88d23eff0baa7285c270eba52d8d9f8f081f9)!

implementations/rust/ockam/ockam_transport_core/src/lib.rs
```diff
@@ -2,6 +2,16 @@
...
+
 mod error;
 #[cfg(test)]
 mod error_test;
+pub mod tcp;
+
+/// TCP address type constant
+pub const TCP: u8 = 1;
+
+pub(crate) const CLUSTER_NAME: &str = "_internals.transport.tcp";
```
spacekookie, line 17: Minor note: maybe we want to rename this cluster to `smoltcp` so that shutdowns between the two tcp stacks (if running in parallel) are independent.

implementations/rust/ockam/ockam_transport_core/src/tcp/workers/receiver.rs
```diff
@@ -1,51 +1,53 @@
...
 ///
 /// This half of the worker is created when spawning a new connection
 /// worker pair, and listens for incoming TCP packets, to relay into
 /// the node message system.
-pub(crate) struct TcpRecvProcessor {
-    rx: OwnedReadHalf,
+pub(crate) struct TcpRecvProcessor<R> {
     peer_addr: Address,
+    tcp_stream: R,
+    cluster_name: &'static str,
```
spacekookie, line 21: Is there a reason to store the `cluster_name` again here?

spacekookie (commented):
Hey!

Thank you for all the effort you already put into this PR. I left you some comments with minor things I'd like to see changed in the API.

I haven't read _all_ the code yet (it's quite a lot :sweat_smile:) but this looks very good so far :slightly_smiling_face:

thomcc:
I'll finish reviewing later today, but the length of the lines that have comments is a bit much. I think we try to keep that to around 100 characters.

If you use vscode, I use <https://marketplace.visualstudio.com/items?itemName=stkb.rewrap> to help with this.

implementations/rust/ockam/ockam_examples/example_projects/smoltcp/examples/network_echo_server.rs
```diff
@@ -0,0 +1,69 @@
...
+        _ => panic!("Cannot parse address"),
+    };
+
+    let configuration = InterfaceConfiguration::<_, Routes<'static>>::new(
+        [0x02, 0x03, 0x04, 0x05, 0x06, 0x07],
+        [IpCidr::new(IpAddress::from_str(bind_ip_addr).unwrap(), 24)],
+        &*DEVICE,
+    );
+    let tcp = SmolTcpTransport::<ThreadLocalPortProvider>::create(&ctx, configuration, Some(StdClock))
+        .await?;
```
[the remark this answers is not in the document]
conectado: I think we could either create a type alias:

```rs
type TcpTransport = SmolTcpTransport<ThreadLocalPortProvider>
```
So most users would use `TcpTransport` and those who care about the port provider use `SmolTcpTransport<P>`.

Or we could add a default generic parameter:

```rs
struct SmolTcpTransport<P = ThreadLocalPortProvider> { ... }
```

But that has the problem that users that wanted to use the default would need to do `<SmolTcpTransport>::create(...)`.

However, I think that most users that will use this crate would do it in a `no_std` context, meaning they will need to provide the port provider. I think that using this in a `std` context would be mostly done for testing and examples.

implementations/rust/ockam/ockam_transport_core/src/lib.rs
conectado: Done [here](https://github.com/ockam-network/ockam/pull/2213/commits/87cb060fe0e3ee2d5f7e2750ea7e4ac9f378cfb0)!

implementations/rust/ockam/ockam_transport_core/src/tcp/workers/receiver.rs
conectado: It was a vestigial change when I took the `cluster_name` as a parameter from outside the crate and I needed to store it to use it in the worker's `initialize`. But since I [changed it back](https://github.com/ockam-network/ockam/pull/2213#discussion_r812353796) to taking the `cluster_name` from outside I need to keep it.

antoinevg (changes requested):
Thank you. This is a stunning PR, really good work.

I can't wait to try this out on embedded!

implementations/rust/ockam/ockam_transport_smoltcp/src/net/device/packet_pool.rs
```diff
@@ -0,0 +1,77 @@
+// This is copied verbatim(Some stuff we don't need removed) from [embassy](https://github.com/embassy-rs/embassy/blob/7561fa19348530ce85e2645e0be8801b9b2bbe13/embassy-net/src/packet_pool.rs)
+// One advantage of using atomic_pool instead of heapless it doesn't seem to have the same kind of ABA problem.
```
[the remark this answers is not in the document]
conectado: The **main** reason to use `atomic_pool` was because `embassy` was doing that.

I'm not sure the advantages of `heapless`'s Treiber stack over `atomic_pool`'s  method of keeping track of available pool using a bitset.

While the advantage of `atomic_pool`'s method is clear, it is `Sync` for all architectures, I'm not sure if it's because `atomic_pool` didn't take the same considerations and edge cases into account or the implementation is sound. 

Also, `atomic_pool` doesn't mention the ABA soundness in x64 problem and by superficially reviewing it I came to the conclusion that it doesn't seem to have that problem since that seems to arise from the Treiber stack using a linked list but I can't confirm it %100.

I will read more on this and get back to you! (Atomics are really complicated 😵‍💫)

implementations/rust/ockam/ockam_transport_core/Cargo.toml
```diff
@@ -14,22 +14,35 @@ Generic Transport primitives.
...
 # Feature (enabled by default): "std" enables functionality expected to
 # be available on a standard platform.
-std = ["ockam_core/std"]
+std = [
+    "ockam_core/std",
+    "ockam_node/std",
+    "tokio",
+    "tokio/io-util",
+    "tokio/net",
+]
```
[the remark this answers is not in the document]
conectado: 1. I think this is not a problem since `tokio` is an optional dependency that's only included with `std` enabled. I only use `tokio` here for the implementations of the `AsyncRead` and `AsyncWrite` traits on `std`. I could completely remove the dependency by using the newtype pattern to do the implementation in `ockam_transport_tcp` if that's preferable.
2.  I figured that we can start extracting the behavior as the weirdness appears, that way we maximize the shared behavior.
3. Although not particularly a big change while working on this the `heartbeat` was added and by solving conflicts it was automatically enjoyed by the smoltcp crate.

All in all I thought it was a good idea as the behavior they share don't seem coincidental and maybe in the future more protocols could use this worker/router implementation.

Anyways, I understand the points and also the generic implementation along with all the trait bounds make the code much harder to parse so if you still think is better to have each crate with their own worker/router implementation I will go back to that!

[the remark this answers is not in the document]
antoinevg: Valid points all but I do feel like the end result is a bigger increase in overall code complexity than what we're gaining through de-duplication here.

Thoughts @thomcc @SanjoDeundiak @spacekookie ?

etorreborre:
Hi @conectado, do you plan to resume your work at some stage on this PR? Otherwise, since it has derived quite a bit from the current `develop` I propose to close it and hopefully someone else will be able to use it to restart the work at some stage.

etorreborre:
Since this branch has diverged we are going to close this PR for now. If anyone wants to resume the work on the smoltcp transport, the branch is available in the ockam repository at `conectado/tcp-transport-using-smoltcp`.

mrinalwadhwa:
@conectado Thank you for all the time you put into this 🙏 
Hopefully someone else will come along and take your great work further from this branch
https://github.com/build-trust/ockam/tree/conectado/tcp-transport-using-smoltcp

PR #2224, opened 2021-11-17
Elixir implementations: Add custom options on Worker Genserver

mvisentin88:
## Checks

- [ ] All commits in this Pull Request follow the Ockam [commit message convention](https://www.ockam.io/learn/how-to-guides/contributing/CONTRIBUTING#commit-messages).
- [ ] I accept the Ockam Community [Code of Conduct](https://www.ockam.io/learn/how-to-guides/high-performance-team/conduct/).
- [ ] I have accepted the Ockam [Contributor Licence Agreement](https://www.ockam.io/learn/how-to-guides/contributing/cla/) by adding my Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/ockam-network/contributors/blob/master/CONTRIBUTORS.csv) file in a separate pull request to the [ockam-network/contributors](https://github.com/ockam-network/contributors) repository.


mrinalwadhwa:
@mvisentin88 thank you for sending this PR! and welcome to the Ockam community!

Could help me understand more about your usecase for this change?

One approach for custom options is already supported by the create function ... here's a quick example:

```elixir
defmodule W do
  use Ockam.Worker

  @impl true
  def setup(options, state) do
    IO.puts("Setup - Options: #{inspect(options)}\t State: #{inspect(state)}")
    {:ok, state}
  end

  @impl true
  def handle_message(message, %{address: address} = state) do
    IO.puts("Address: #{address}\t Received: #{inspect(message)}")
    {:ok, state}
  end
end


{:ok, _random_address} = W.create(my_option: 100)
{:ok, "my_worker"} = W.create(address: "my_worker", my_option: 100)
```

mvisentin88:
@mrinalwadhwa  your are welcome is a pleasure. 

I'm only think that's maybe is better to set as default behaviour.

hairyhum:
@mvisentin88 as I understand that would make workers which don't have `setup/2` function implemented have all the options in the state. Makes sense.

Although I don't think we want to override what we have in the state already at the point of `setup`, so we should merge in another direction.

Also for the code cleanup, can you please remove the dependencies update from this PR and make it a single commit?

mrinalwadhwa:
Closing this, since it has been open for a long time.

@mvisentin88 if you would like to send a new PR with the following adjustments that @hairyhum suggested above, we would love to include it. Thank you.

> ... as I understand that would make workers which don't have `setup/2` function implemented have all the options in the state. Makes sense.
> 
> Although I don't think we want to override what we have in the state already at the point of `setup/2`, so it should merge in another direction.
> 
> Also for the code cleanup, can you please remove the dependencies update from this PR and make it a single commit?

PR #2229, opened 2021-11-18
feat(rust): decouple `Context` from `Worker`

thomcc:
This does the following things, more or less (it's a fairly large refactor):
- Introduces a new `NodeContext` trait.
- Makes Worker/Processor generic over `<C: NodeContext>`
- Changes all the workers/processors to behave this way
- Changes all the structs that need to reference Contexts to be generic over the Context (for example, `Handle<C: NodeContext>` rather than Handle)
- Moves Handle/Cancel and such into ockam_core
- Updates examples, etc (fewer changes here than expected).

Things I'm not happy with (none of these need to be addressed before landing, I think):
- I think the api of `NodeContext` is a bit janky. It's not really minimal. Much of this needs the (forthcoming) generic async work before it can be improved, but after that should be improvable.
- Several things that used to take `impl Into<Route>` or `impl Into<String>`, can't anymore, at the moment, but this should be fixable in the future (although, it might take true async traits, but we don't need dynamic dispatch so this might be relatively soon...). This mostly impacts internals, rather than the public API used by examples, since `Context` still defines these as inherent methods.
- AsyncTryClone derive needs to be adapted so that it can handle the generic type.


spacekookie:
I'm wondering if we can abstract the underlying runtime the same way we currently do with `ockam_node` in the `ockam` crate, i.e. by creating the API we need from a runtime (probably `spawn`, `block_on`, `channels`, ...?) in our own crate, which can then include either `tokio` or `async-std` or whatever via a feature flag.

There's no reason the async runtime needs to be a generic type on Context (or for Context to even understand that there can be different runtimes)

implementations/rust/ockam/ockam_channel/src/secure_channel.rs
```diff
@@ -52,24 +52,25 @@ impl SecureChannel {
...
+        ctx: &C,
         address: A,
         new_key_exchanger: N,
         vault: V,
     ) -> Result<()> {
         let address = address.into();
         let channel_listener = SecureChannelListener::new(new_key_exchanger, vault);
         info!("Starting SecureChannel listener at {}", &address);
-        ctx.start_worker(address, channel_listener).await?;
+        ctx.start_worker(address.into(), channel_listener).await?;
```
spacekookie, line 65: Is this a new necessity of the `Context` trait (i.e. will users have to call `.into()` manually on types that they pass to the new context API?

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
```diff
@@ -159,6 +159,8 @@ pub struct Credential {
     type_id: String,
 }
 
+impl ockam_core::Message for Credential {}
+
```
spacekookie, line 163: This should probably use the `#[derive(Message)]` macro

spacekookie (commented):
Thank you for all the work you've put into this already. Generally this PR looks very good.

Two things I noticed: 

1. having to call `.into()` on any Context function that takes a generic `Into<T>` parameter seems like a pretty big drawback in terms of developer UX. Is that actually the case or am I misunderstanding some of the sections of this PR?
2. there are a bunch of style nitpicks in this PR which should probably be pulled out into it's own set of changes. This change set is already pretty huge and some things seem to be changed for style reasons more than functionality.

implementations/rust/ockam/ockam_channel/src/secure_channel.rs
thomcc: The issue is that you end up needing `Into<Route> + Send + 'static`, or the resulting future is fairly restricted due to the way `async_trait` desugars. This ends up being pretty restrictive and leading to confusing lifetime errors for users, but perhaps it's worth it?

Note that we still have the inherent methods, so it doesn't impact the examples, mostly the internals. This also might be alleviated with proper (non-`dyn`) async trait support.

thomcc:
> there are a bunch of style nitpicks in this PR which should probably be pulled out into it's own set of changes. This change set is already pretty huge and some things seem to be changed for style reasons more than functionality.

Ah yeah this is fair, I started by trying to clean up the API to something we could stabilize, but then decided to do that later once the scope of this was more clear.

The Into stuff is unfortunate, I agree. I have some other thoughts about how we can do this (a wrapper around `impl NodeContext` (or `&mut impl NodeContext`) that provides the intos prior to entering async) — the issue is that we see this as just an convenience helper on the arguments, but rustc believes that the inputs have to potentially live for the entire length of the async computation, and potentially need to be sent to other threads.

This is unfortunate, and is made worse by the async_trait macro, so I'm hopeful it's not a permanent situation — async trait support is fairly far along, as long as you don't need `dyn trait` support (which is very unclear when it will happen).

implementations/rust/ockam/ockam_entity/src/credential/traits.rs
thomcc: For a while during this PR I was bit by https://github.com/rust-analyzer/rust-analyzer/issues/10540, which made me think that wasn't working for some reason.

thomcc:
Oh, the other category of "stylistic changes" is "things that fail linting from `../../gradlew lint`" (which I had been using as "approximately what CI does") Many of these were things I had thought I had (somehow) caused, but it seems that `../../gradlew lint` just fails on `develop` 🥲

implementations/rust/ockam/ockam_core/src/cancel.rs
```diff
@@ -0,0 +1,67 @@
...
+            inner,
+            local_msg,
+            addr,
+            ctx,
+        }
+    }
+
+    /// Cancel this message
+    pub async fn cancel(self) -> Result<()> {
+        self.ctx.forward(self.local_msg).await
```
SanjoDeundiak, line 30: Off-topic: if cancelling a message became that simple, do we even need this structure?

implementations/rust/ockam/ockam_core/src/message.rs
```diff
@@ -72,6 +72,7 @@ pub trait Message: Encodable + Decodable + Send + 'static {}
 impl Message for () {}
```
SanjoDeundiak, line 72: Off-topic: should we remove those implementations to discourage users from sending such basic types to workers? I think it's better to wrap even such simple data into dedicated structs. Even though we don't use that approach in our examples

implementations/rust/ockam/ockam_entity/src/worker/trust_policy_worker.rs
```diff
@@ -1,46 +1,52 @@
...
 use ockam_core::compat::boxed::Box;
 use ockam_core::{
-    async_trait::async_trait, Address, AsyncTryClone, Message, Result, Routed, Worker,
+    async_trait::async_trait, Address, AsyncTryClone, Handle, Message, NodeContext, Result, Routed,
+    Worker,
 };
 use ockam_message_derive::Message;
-use ockam_node::{Context, Handle};
 use serde::{Deserialize, Serialize};
-#[derive(AsyncTryClone)]
```
SanjoDeundiak, line 9: So, does it work?

implementations/rust/ockam/ockam_node/Cargo.toml
```diff
@@ -36,6 +36,7 @@ tracing-subscriber = { version = "0.3", features = ["fmt", "env-filter"], option
 heapless = { version = "0.7", features = [ "mpmc_large" ], optional = true }
 ockam_node_no_std = { path = "../ockam_node_no_std", version = "0.12.0" , default-features = false, optional = true }
 ockam_executor = { path = "../ockam_executor", version = "^0.8.0", default-features = false, optional = true }
+async-trait = "0.1.42"
```
SanjoDeundiak, line 39: ockam_core exports that dependency

thomcc: Hm, it shouldn't. We really should not be reexporting dependencies, it's a bad practice. But I can use it for now.

implementations/rust/ockam/ockam_core/src/message.rs
thomcc: We should either do that or implement it for all serializable/deserializable stdlib types.

implementations/rust/ockam/ockam_node/Cargo.toml
SanjoDeundiak: @thomcc Could you please elaborate on why it's a bad practice?

SanjoDeundiak (approved):

implementations/rust/ockam/ockam_node/Cargo.toml
thomcc: It increases the surface of our public APIs to include every API in the re-exported library. This means that if they have a major breaking change, we are forced to follow as well, or violate semver (in a very real way — note that we had an issue around another library doing this very thing in the past).

This is not worth the small convenience it adds.

thomcc: Note that it's okay to reexport dependencies when you control both. It can also be okay if you are okay with that tradeoff, but for this I don't think it's worth it. Ideally, we would even prevent `#[ockam::worker]` from being applied to non-worker implementations.

But it also means that you shouldn't release a version 1.0 stable without all of your public dependencies being there too.

implementations/rust/ockam/ockam_core/src/message.rs
thomcc: Thinking a bit, of these, I think I'm in favor of implementing it for everything, tbh. More convenient and less of a roadblock than it would be to tell people "Hey, you need to define a newtype wrapper here". It will keep both our examples simple, as well as requiring less work for users.

implementations/rust/ockam/ockam_node/Cargo.toml
SanjoDeundiak: @thomcc we can create another typedef for async_trait to have more control over it (to be able to redefine it in the future if needed). Other than that, I don't understand what are the concerns about public API surface. async_trait directly affects all of our public traits (I think all of them are public). Correct me if I'm wrong, let's imagine user created their own implementation of our trait in their crate and declared async trait dependency directly, now if there's breaking change in async trait crate and ockam and user's crates have different version of it - it won't work. While if user used reexported async_trait dependency on their implementation - it would still work (worst case - there will be 2 async_trait crates in the final binary/library)

thomcc: > Correct me if I'm wrong, let's imagine user created their own implementation of our trait in their crate and declared async trait dependency directly, now if there's breaking change in async trait crate and ockam and user's crates have different version of it - it won't work

It depends on what the change is. In general, I suspect most changes to async_trait would still work here, but yes, it's possible that some could exist where it would not.

A bigger issue is users who use ockam without needing to write any custom workers, or otherwise interact with our async_trait export. Now because we reexport it, if it has a breaking change (say 0.3 to 0.4 is done to change), Ockam is forced to take a breaking change as well — this could be as significant as 1.0 to 2.0, which sounds major, but in reality would just because a minor part of our API surface had a breaking change. It would be unfortunate.

It's worth noting that the author of async_trait is generally quite good about marking crates that won't suffer any more breaking changes as 1.0, so I would not be that confident about async_trait being good for public usage here.

SanjoDeundiak: @thomcc I don't think it's possible to write anything meaningful with ockam without creating your own workers... And making users add async_trait to their Cargo.toml along with ockam is not something we want as well

thomcc: > I don't think it's possible to write anything meaningful with ockam without creating your own workers

That's definitely distinct from what @mrinalwadhwa has indicated to me, but it does seem true at the moment.

Anyway, my long term plan for solving this issue with the `#[worker]` macro is to move it into our code, and make it specific to defining workers. As it is, someone can do:

```rs
#[ockam::worker]
trait SomeAsyncTraitThatHasNothingToDoWithWorkers { ... }

// or

#[ockam::worker]
impl SomeOtherAsyncTrait for MyType { ... }
```

and have it work. That's not something we want — you should only be allowed to use that macro with workers — most ways we might want to evolve the `worker` macro would be incompatble with it (for example, I've wanted it to automatically insert `tracing` calls — perhaps behind a feature), and would make adding such functionality into a breaking change when it doesn't need to be.

> making users add async_trait to their Cargo.toml along with ockam is not something we want as well

While I think where possible we should avoid having multiple Cargo.toml lines for most usage, honestly, that's a smaller concern than having API stability, IMO.

You'll note that even with `tokio`, their stability commitment has forced them to pull `tokio-streams` out, becuase they expect it to break sooner, and users of tokio that need stream functionality are forced to import both crates separately.

I don't think we can realistically be in the position where people using our APIs don't have to pull in any other libraries. We should just move it off of the critical path (as the note above about the `#[worker]` trait likely addresses it for defining worker).

implementations/rust/ockam/ockam_core/src/message.rs
spacekookie: Yea I'd second that. At times I've wanted to just send an `i64` in a demo which used to work and now doesn't.
In reality primitive type messages are probably less useful outside of a quick prototype but if that's all the user wants they shouldn't have to wrap the type in a struct that does nothing else

implementations/rust/ockam/ockam_core/src/cancel.rs
spacekookie: I think it's still valuable. Not because of the complexity involved in this function, but the complexity on the user's part to understand the internals of the runtime enough to connect that `forward(...)` re-queues the message.

SanjoDeundiak: @spacekookie how about moving that to Routed<>?

SanjoDeundiak:
Would be great to revisit in the future

PR #2233, opened 2021-11-18
fix(rust): pin ed25519 version due to deprecation changes

jared-s:


thomcc:
It's probably better to just `#[allow(deprectated)]`. Pinning has consequences downstream, allow(deprecated) doesn't.

PR #2240, opened 2021-11-19
chore(rust): deny warnings in ci, not local development

thomcc:
Right now in the rust code, if there's a compiler warning during local development, the build will fail to complete. This is inconvenient, since *really*, we only care about this stuff when *landing* code in the repo. As a result, this PR moves these checks entirely into CI (where they already existed, for the most part, although I firmed that up).

The exception to this the `unsafe_code` lint, which is reported by various tooling, and isn't *really* in the same category as other lints. This one is kept as deny in the crates that can use it (although perhaps it should be increased to `#![forbid]` in the future).

## Checks

### DIFF
--- a/.github/workflows/cargo_update_check.yml
+++ b/.github/workflows/cargo_update_check.yml
@@ -16,6 +16,8 @@ jobs:
 
       - name: Update and Check Ockam Example
         working-directory: examples/rust/get_started
+        env:
+          RUSTFLAGS: -Dwarnings
         run: |
           rm -rf Cargo.lock
           cargo update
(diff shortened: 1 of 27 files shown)


.github/workflows/cargo_update_check.yml
```diff
@@ -16,6 +16,8 @@ jobs:
 
       - name: Update and Check Ockam Example
         working-directory: examples/rust/get_started
+        env:
```
thomcc, line 19: This is done in the env because `cargo check` doesn't accept `-Dwarnings` as an argument — cargo clippy does, which this doesn't use)

thomcc:
hm, I guess I turned on additional linting for some of this. I'll fix the warnings rather than disabling it though.

PR #2244, opened 2021-11-20
chore(rust): improve performance of gradle lint task

thomcc:
By not generating documentation for dependencies. Ideally we'd use `RUSTDOCFLAGS="--check"` but it isn't stable yet, and I don't know how much more it would speed things up.

Times, according to gradle:

Before (already built): `9m 2s`
Before (full rebuild): I don't know, presumably longer.

After (already built): `2m 23s`
After (full rebuild): `5m 1s`


mrinalwadhwa (approved):
Awesome 🚀

PR #2254, opened 2021-11-23
fix(rust): change context import from ockam_node to crate

spacekookie:
A small fix related to #2197


mrinalwadhwa:
rebased and merged in #2266

PR #2269, opened 2021-11-26
chore: remove symlinks to `DEVELOP.md` and `LICENSE`

mrinalwadhwa:


thomcc:
It's worth noting that we'll have to re-add the license if we eventually need to get packaged on Linux (if we want the ockam command installable from repos like apt/pacman/etc — admittedly this is something of a ways off).

mrinalwadhwa:
Let's add them back when we reach that point, for now the Cargo.toml `license` field is sufficient.

PR #2284, opened 2021-11-30
Rust bi-directional messaging channels

spacekookie:
The initial implementation of bi-directional messaging channels, built on top of uni-directional messaging pipes.

During the handshake creation two pipe workers are spawned on each end of the channel (Sender and Receiver), plus an additional channel worker that is responsible for doing route backtracing.

Currently this code does not allow for additional behaviour to be attached to a channel worker (for example to implement secure channels)


jared-s (approved):
Very cool! :tada:

PR #2289, opened 2021-11-30
feat(rust): add worker ref example for `SecureChannel`

SanjoDeundiak:


implementations/rust/ockam/ockam_entity/src/channel/secure_channel_worker.rs
```diff
@@ -507,6 +522,38 @@ impl<I: Identity, T: TrustPolicy> SecureChannelWorker<I, T> {
...
+
+    async fn handle_api(
+        &mut self,
+        ctx: &mut Context,
+        msg: Routed<<Self as Worker>::Message>,
+        s: State<I, T>,
+    ) -> Result<()> {
+        let return_route = msg.return_route();
+
+        let msg = SecureChannelApiRequest::decode(msg.payload())?;
```
spacekookie, line 534: This is something that I do a lot in the `ockam` crate as well. I'm wondering if maybe we could just add a function to `TransportMessage` that will decode the payload and return `T`

SanjoDeundiak: @spacekookie nice idea, added as issue #2298

Issue #2283, opened 2021-11-30
Worker relay ctrl channel is triggered during regular Worker stop


SanjoDeundiak:
[This](https://github.com/ockam-network/ockam/blob/0d94740199209147509591fec703cbb9ab96b298/implementations/rust/ockam/ockam_node/src/relay/worker_relay.rs#L143) control channel was designed to be used during node shutdown, however during Worker stop the other side of control channel gets dropped by Router, which will make .recv() function return with `None`. Sometimes, recv_message() will return first, because messages channel is also dropped, so it's a race condition. I don't think it causes any real-world problem, but doesn't work as intended as well.

spacekookie:
I think this issue was resolved in #2667. Closing for now.

PR #2295, opened 2021-12-01
Sanjo/improvements

SanjoDeundiak:


implementations/rust/ockam/ockam_node/src/relay/worker_relay.rs
```diff
@@ -136,13 +136,17 @@ where
...
                     }
                 },
-                _ = ctrl_rx.recv() => {
-                    debug!("Relay received shutdown signal, terminating!");
-                    break;
+                result = ctrl_rx.recv() => {
+                    if result.is_some() {
+                        debug!("Relay received shutdown signal, terminating!");
+                        break;
+                    }
```
spacekookie, line 147: I assume this gates the `break` condition from the case where the control channel is dropped and returns `None`?

spacekookie (approved):
LGTM

implementations/rust/ockam/ockam_node/src/relay/worker_relay.rs
SanjoDeundiak: @spacekookie yes, I don't like that solution, there is race condition described in #2283. Would be great to solve that properly. For now I just thought reasonable to distinguish between dropping a tx side and receiving a signal, because at least debug message was misleading - it printed that we got shutdown signal even if we didn't

mrinalwadhwa:
@SanjoDeundiak small typo here:
```
chore(rust): move and improve ockam_node tets
```

SanjoDeundiak:
> @SanjoDeundiak small typo here:
> 
> ```
> chore(rust): move and improve ockam_node tets
> ```

@mrinalwadhwa Fixed

PR #2301, opened 2021-12-03
build: bump gradle version to 7.3.1

antoinevg:
## Current Behaviour

On newer JDK versions (`>= openjdk16`)  all gradle operations  fail with the message:

```
Could not open settings generic class cache for settings file ...
```

## Proposed Changes

Updating gradle to the latest version (`7.3.1`) resolves this issue.


metaclips (approved):
Seems this breakage is ARM specific. Thanks @antoinevg

PR #2308, opened 2021-12-06
build: improve gradle and github workflows for rust and elixir

mrinalwadhwa:
# Debug

**Rust**
```
cd implementations/rust
../../gradlew build
``` 

**Elixir**

Depends on Rust

```
cd implementations/elixir
../../gradlew build
``` 

# Test

**Rust**
```
cd implementations/rust
../../gradlew test
``` 

**Elixir**
```
cd implementations/elixir
../../gradlew test
``` 

# Release

**Rust**
```
cd implementations/rust
../../gradlew build -Pmode=release
``` 

**Elixir**
```
cd implementations/elixir
../../gradlew build -Pmode=release
``` 

# Clean

**Rust**
```
cd implementations/rust
../../gradlew clean
``` 

**Elixir**
```
cd implementations/elixir
../../gradlew clean
``` 

# VeryClean

**Rust**
```
cd implementations/rust
../../gradlew veryClean
``` 

**Elixir**
```
cd implementations/elixir
../../gradlew veryClean
``` 

# From the repo root

```
./gradlew build
```

```
./gradlew test
```

```
./gradlew clean
```

```
./gradlew veryClean
```


mrinalwadhwa:
@thomcc @metaclips @hairyhum this PR is work in progress ... could I get a prelim review of it.

I'm not happy with the 8minute and 16minute build times for Rust and Elixir workflows in Github Runners. I'm hoping some caching will tackle that better than going back to lots of parallel workflows

hairyhum:
Build time for elixir is affected by the dialyzer, I believe. Unfortunately dialyxr doesn't to a best job verifying the cached PLT files and uses old files even when the code change (for dependencies), this means we need to rebuild them more often than actually needed.

mrinalwadhwa:
@jared-s @thomcc @metaclips @hairyhum @SanjoDeundiak could you try the changes from this PR and if no one notices any major issues with this beyond the speed in CI, let's rebase and merge this. We can improve CI speed in later increments.

thomcc:
The rust tests and lint in gradle are very thorough but certainly do *tons* of redundant work for various reasons.

I've been meaning to get to it but haven't found time. I'll take a peek on Friday. The problem is that gradle still performs the tasks as if it's not in a workspace, and so we lose most of of the benefits of running in a workspace.

For example, by cding into each directory and performing the build, the feature selection algorithm is rerun from that crate — this means we end up changing the feature set of e.g. tokio (or something else used by everything), and then all dependents get rebuilt. This happens over and over again; An acquaintance of mine published  https://crates.io/crates/cargo-hakari to help solve this problem, which I'll look into for us.

In the mean time, I suspect for CI we should just `cargo test --workspace` and such, which will be much faster, should hit the same test (AFAIK?) and should also hit the GHA cache far more reliably.

thomcc (approved):
Build seems to work locally, and I also can follow what the gradle does much more clearly. 🚢

PR #2326, opened 2021-12-08
fix(rust): fix `tcp_inlet_and_outlet` example dockerfile

SanjoDeundiak:


jared-s:
Confirmed locally, I believe the old Dockerfile was for pre-workspace migration

jared-s:
Merged https://github.com/ockam-network/ockam/pull/2332

PR #2341, opened 2021-12-10
ci: fix nightly github workflow

mrinalwadhwa:


mrinalwadhwa:
fixed in https://github.com/ockam-network/ockam/pull/2345

PR #2345, opened 2021-12-11
ci: actions and workflows

mrinalwadhwa:


thomcc:
Hmm, my understanding of how some of this stuff works is that caching e.g. the target folder on all jobs would be a problem, because not all jobs perform builds (of rust code) — this means a cache entry with a fully populated target folder could be wiped out by one with a missing or mostly empty target folder.

mrinalwadhwa:
@thomcc you're probably right, I'm still experimenting with actions/cache.
I've removed that change form this PR, will tackle it in a different PR.

PR #2356, opened 2021-12-14
release script update

metaclips:
## Current Behaviour
To perform a manual release, we run the scripts in the following order

- Generate Changelog. Updates the changelog files with commit messages.
- Bump crates using `cargo-release` which should also indicate updated crate version in changelog and readme files.
- Publish Crates to crates.io
- Bump Crates To -Dev Version.
- Tag Crates using commit SHA in step 3

The current order ensures we only update crates that were published to -dev ( we track crates to be bumped to -dev using diff of last git-tag ). This gives wrong `git tag` names on release as we get the tag details from Cargo.toml and bumping to `-dev` before tagging allocates the tag name as name_version-dev (version of the bumped -dev, instead of that of the SHA).
https://github.com/ockam-network/ockam/blob/5a84ea59fd9d8a67b8e0585a6be312c6d54fc906/tools/scripts/release/tagging.sh#L14-L17

This issue only happens when running tagging and publishing scripts manually as CI does a `checkout` to the commit SHA before tagging or publishing is done.
## Proposed Changes

Proposed order of running script both manually and in CI is

- Generate Changelog, which should update the changelog files with commits.
- Bump crates using `cargo-release` which should also indicate updated crate version in changelog and readme files.
- Publish Crates to crates.io
- Tag Crates. We can tag crates using the HEAD commit since it's that of the bumped versions.
- Bump Crates To -Dev Version. Bumping crates to -dev version doesn't keep track of git diff anymore as we only perform a textual search of Cargo.toml if the version contains a -dev substring https://github.com/ockam-network/ockam/blob/be6c633d513179a6d373d0648693984a8ead48cd/tools/scripts/release/dev-bump.sh#L5-L7

Also, this PR ensures
- Chore commits are skipped in changelog generator
- We have a separate script that bumps crates to -dev version which doesn't need to track last git tags.

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -30,6 +30,7 @@ commit_parsers = [
     { message = ".*crate release", skip = true},
     { message = ".*update crates changelog", skip = true},
     { message = "^ci", skip = true},
+    { message = "^chore", skip = true},
     { message = "^.*: add", group = "Added"},
     { message = "^.*: support", group = "Added"},
     { message = "^.*: remove", group = "Removed"},
--- a/release.toml
+++ b/release.toml
@@ -1,12 +1,12 @@
 dev-version = true
 dev-version-ext = "dev"
-post-release-commit-message = "ci: release {{crate_name}} dev version {{version}}"
+post-release-commit-message = "chore(rust): release {{crate_name}} dev version {{version}}"
 push = false
 tag = false
 verify = true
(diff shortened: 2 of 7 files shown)


jared-s (approved):

jared-s:
Merged #2424

PR #2359, opened 2021-12-14
refactor(rust): delete the ockam_node_no_std crate

### LINKED ISSUE #2287: Delete the `ockam_node_no_std` crate.
The [`ockam_node_no_std`](https://github.com/ockam-network/ockam/tree/develop/implementations/rust/ockam/ockam_node_no_std) crate was mostly around to remind us that we needed to support no_std and alternative nodes. The support for no_std landed in ockam_node and ockam_executor, and the alternative node work is actively being worked on, but won't involve this crate.

As a result, this crate should just be removed. Note that you should also delete some references to it from elsewhere, so you may need to do a search. For example, inside CI (`.github/workflows`), in any Cargo.toml files of other crates, etc.

totsteps:
closes #2287


mrinalwadhwa (approved):
Thank you, looks great 👍

PR #2363, opened 2021-12-15
ci: cancel github workflows if there is a new trigger

mrinalwadhwa:
Use Concurrency Groups to cancel github workflows if there is a new trigger.

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -16,6 +16,10 @@ on:
   schedule:
     - cron:  '0 1 * * *' # Run every day at 1am UTC https://cron.help/#0_1_*_*_*
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     name: Build
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -5,6 +5,10 @@ permissions:
 
 on: [pull_request, push, workflow_dispatch]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   lint:
     name: Lint
--- a/.github/workflows/editorconfig.yml
+++ b/.github/workflows/editorconfig.yml
@@ -5,6 +5,10 @@ permissions:
 
 on: [pull_request, push, workflow_dispatch]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   lint:
     name: Lint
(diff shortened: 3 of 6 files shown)


thomcc:
Hm, doesn't this already happen?

mrinalwadhwa:
No, from what I can tell, this is a good watch https://youtu.be/JMHs5lYpvAM?t=314
> Hm, doesn't this already happen?

mrinalwadhwa:
Just confirmed it. For example, this workflow was canceled while it was running because of a new push to the PR branch
https://github.com/ockam-network/ockam/actions/runs/1581076168

thomcc:
Hm, I must be misunderstanding — that's what I thought "cancel github workflows if there is a new trigger" meant. I'll watch the video.

thomcc:
Oh, I see, that's from this branch, hmm...

mrinalwadhwa:
What I meant to say was that it doesn't happen if you don't have a concurrency group defined. New pushes continue to trigger more runs. This PR adds concurrency group definitions that are unique to a workflow+branch.
https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency

mrinalwadhwa:
I also tried 

```
    if: |
      github.event_name != "pull_request" ||
        github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name
```

based on this article (it has a good description of the tradeoffs)
https://wildwolf.name/github-actions-how-to-avoid-running-the-same-workflow-multiple-times/ 

but it didn't quite work on my first try .. will try on a later PR.

thomcc:
I see. I'm fairly certain that I've seen that behavior on PRs already (that is, when you update a PR branch, typically the pending jobs all fail — I've never dug into why, nor do I have a branch on hand that did it the other way).

I guess the default is slightly different though https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#examples-using-concurrency-and-the-default-behavior-1, and doesn't seem to have the flag to cancel in progress... Hrm, strange.

Well, either way I agree this behavior sounds desirable.

thomcc:
Hm, that blog post helps some, the description of the problem here:

> Therefore we have four wasted jobs running: the jobs running for `pull_request` are the same as the ones running for `push`.

is something I've seen before, although I've followed the suggestion here:

> The easiest workaround is to restrict the `push` event only to the main repository branch, like this: *\<snip\>*

(You may note that [my blog on using Github Actions with Rust](https://shift.click/blog/github-actions-rust/) doesn't mention it (I don't think it does anyway, it's been a while since I wrote it), but does manage to suggest doing the right thing for this — because it's a problem I'm familiar with)

I'm not exactly how `concurrency` ties in here, but will do some more reading.

PR #2367, opened 2021-12-15
remove workflows that require repo dispatch

metaclips:
Remove docker_hub workflow as it's no longer needed


mrinalwadhwa:
merged in https://github.com/ockam-network/ockam/pull/2368

Issue #1947, opened 2021-10-07
Possible deadlock under high load, due to fixed-capacity channels?


thomcc:
https://github.com/ockam-network/ockam/blob/8bf1502c8a105c874e8d493097c6c57948183a09/implementations/rust/ockam/ockam_node/src/context.rs#L129 (and some others) use a fixed-capacity channel (which if full, blocks the sender until it isn't).

This is surprising to me since there's a pretty classic deadlock that can happen because of these in some use cases like ours... Or at least it seems like it could — there's enough going on in `ockam_node` that I can't be sure if this is handled — it could be.

---
Specifically, the situation where I worry a deadlock happens is:

1. If you have two workers, W1 and W2, where W1 and W2 which may message each-other at least some of the time. (I think they need to be local workers too, but maybe I'm just lacking imagination).

2. Both W1 and W2 are very nearly at their channels capacity. (For example, if they're unable to process messages fast enough, and the backlog fills all the way up.

3. Now, W1 sends a message to W2. Because W2s channel is full, the `send(msg).await` wont resolve until W2 pops the next item of its channel and W1's message finds space.

4. Before W2 gets to do that (while W1 is still blocked), W2 sends W1 a message.

5. Sadly, W1's channel is also at capacity, so W2's `send(msg).await` won't wont resolve until W1 pops the next item of its channel.

That is — W1 is blocked until W2 pops an item off its queue... and W2 is *also* blocked until W1 pops an item off its queue.

---

That said, I don't know for sure this is actually possible — message sending in ockam is involved enough that it might not be, and might just *look* possible for ockam. (Concretely, I'm assuming `ctx.send_message` bottoms out on something like a channel::send to these channels)

That said, I have see this type of deadlock before in other projeces, so even though it sounds unlikely, it definitely happens under high load (especially if you arent on full release build).

PR #2371, opened 2021-12-15
feat(rust): various ockam_executor updates

antoinevg:
# Proposed Changes

This PR resolves a number of outstanding TODO's related to the `ockam_executor` crate:

## fix(rust): protect processor relays against accidental async executor deadlock

There exists a footgun in `ockam_executor` whereby `processor_relay`'s run loop can deadlock if an implementer of the `Processor` trait were to e.g. tightly poll a method without yielding back to the executor.

This fix simply adds a call to `crate::tokio::task::yield_now().await;` in the run loop to ensure the executor can still service other tasks.

## chore(rust): clean up ockam_executor code

A general clean up of `ockam_executor` code: enabling warnings, fixing lints, removing unused code etc.

## feat(rust): use the tracing crate for logging on no_std

Up till now we've been using a simple set of placeholder log macros (`info!`, `debug!` etc.) in `ockam_executor` to add minimal logging support on embedded.

These log macros have been implemented using the `ockam_core::println!` macro which, on embedded, is a simple wrapper around the Cortex semihosting logging facility.

While it's been good enough for demos there are a couple of issues with this:

* It needs to pull architecture specific crates into `ockam_node`, namely `cortex-m` and `cortex-m-semihosting`
* There are a lot of logging facilities out there and Ockam, itself, should not have to know anything about them nor have to implement them.

So the `tracing` crate is a little more heavy-weight than I'd like for embedded usage but, at least in the short term, we need _something_ and there aren't a lot of options out there that will play nicely with both `std` and `no_std`.


implementations/rust/ockam/ockam_executor/src/channel.rs
```diff
@@ -50,28 +57,21 @@ struct Inner<T> {
     is_sender_closed: AtomicBool,
 }
 
+impl<T> Inner<T> {
+    fn _len(&self) -> usize {
```
thomcc, line 61: Nit: Both are private and underscore-prefixed, so is there a reason to have this rather than just making the caller read `_length` directly?

implementations/rust/ockam/ockam_executor/src/channel.rs
```diff
@@ -50,28 +57,21 @@ struct Inner<T> {
...
+        self._length
+    }
+}
+
+/// Sender
 pub struct Sender<T>(Arc<Inner<T>>);
 
 impl<T: core::fmt::Debug> Sender<T> {
     pub async fn send(&self, value: T) -> Result<(), error::SendError<T>> {
```
thomcc, line 70: Since `SendFuture` is public, this should probably be a non-async fn that returns `SendFuture<'_, T>` directly. That is, this is equivalent but would be more flexible:

```rs
pub fn send(&self, value: T) -> SendFuture<'_, T> {
    Self {
        inner: &self.0,
        value: Some(value),
    }
}
```

That way, the user could put the `SendFuture` in a struct or something (and it also with gives a reason for having it as a separate type)

implementations/rust/ockam/ockam_executor/src/channel.rs
```diff
@@ -104,39 +104,99 @@ impl<T> Drop for Sender<T> {
...
     }
 }
 
+impl<'a, T> Unpin for SendFuture<'a, T> {}
+
+/// Receiver
 pub struct Receiver<T>(Arc<Inner<T>>);
 
 impl<T: core::fmt::Debug> Receiver<T> {
     pub async fn recv(&mut self) -> Option<T> {
```
thomcc, line 155: Same thing I said about `SendFuture` applies here too — should return `RecieveFuture` directly rather than waiting on it internally.

implementations/rust/ockam/ockam_executor/src/channel.rs
```diff
@@ -104,39 +104,99 @@ impl<T> Drop for Sender<T> {
...
+        }
     }
 }
 
+impl<'a, T> Unpin for SendFuture<'a, T> {}
+
+/// Receiver
 pub struct Receiver<T>(Arc<Inner<T>>);
 
 impl<T: core::fmt::Debug> Receiver<T> {
```
thomcc, line 154: Why is there a `Debug` bound here?

implementations/rust/ockam/ockam_executor/src/channel.rs
```diff
@@ -104,39 +104,99 @@ impl<T> Drop for Sender<T> {
...
+    }
+}
+
+pub struct ReceiveFuture<'a, T> {
+    inner: &'a Inner<T>,
+}
+
+impl<'a, T> Future for ReceiveFuture<'a, T>
+where
+    T: core::fmt::Debug,
```
thomcc, line 176: Ditto about this (why `Debug` bound?)

implementations/rust/ockam/ockam_executor/src/executor.rs
```diff
@@ -177,6 +214,19 @@ impl Task {
     fn allocate(future: impl Future + 'static) -> Box<Task> {
         Box::new(Node {
             id: TaskId::new(),
+            _name: "Task",
+            future: UnsafeCell::new(async {
+                // task terminating
+                future.await;
+            }),
+            // TODO future: Box::pin(future),
```
thomcc, line 222: In the mean time, is there a reason to not change the arg to `impl Future + Unpin + 'static`?

thomcc (commented):
Seems fine, with a couple nits.

Also we're really going to need to go over all the `ockam_executor` code at some point in the future to audit for soundness and thread-safety (obviously there are TODOs along these lines already)

implementations/rust/ockam/ockam_executor/src/channel.rs
antoinevg: I agree! 

The reason I'm being weird here is because this a drop-in for `tokio::sync::mpsc::channel` which means our code is expecting a `Result` and not the future. 

Having it as a separate type was an artifact of making it easier to try different approaches to implementing `poll()`

implementations/rust/ockam/ockam_executor/src/channel.rs
antoinevg: See above.

implementations/rust/ockam/ockam_executor/src/channel.rs
antoinevg: Whoops. That's left over from the early days to make it easier to debug the `no_std` implementation of `ockam::node`. I've removed them.

implementations/rust/ockam/ockam_executor/src/channel.rs
antoinevg: As above.

implementations/rust/ockam/ockam_executor/src/executor.rs
antoinevg: I did try this now but the `Unpin` trait bound ends up bubbling into `ockam_node` with sundry wreckage ensuing. 😕 

i.e. When we introduce:

```rust
impl Task {
    fn allocate(future: impl Future + Unpin + 'static) -> Box<Task> {]
   ...
```
…we end up with things like:

```rust
   Compiling ockam_node v0.41.1-dev (/Users/antoine/Ockam/ockam.git/implementations/rust/ockam/ockam_node)
error[E0277]: `F` cannot be unpinned
  --> /Users/antoine/Ockam/ockam.git/implementations/rust/ockam/ockam_node/src/executor.rs:78:30
   |
78 |         let _join = rt.spawn(future);
   |                        ----- ^^^^^^ the trait `Unpin` is not implemented for `F`
   |                        |
   |                        required by a bound introduced by this call
   |
   = note: consider using `Box::pin`
note: required by a bound in `Runtime::spawn`
  --> /Users/antoine/Ockam/ockam.git/implementations/rust/ockam/ockam_executor/src/runtime.rs:55:21
   |
55 |         F: Future + Unpin + Send + 'static,
   |                     ^^^^^ required by this bound in `Runtime::spawn`
help: consider further restricting this bound
   |
74 |         F: Future + Send + 'static + core::marker::Unpin,
   |                                    +++++++++++++++++++++
```

Ideally I shouldn't have to rewrite `ockam_node` methods to make this work so I still have some way to go to attain  full grokking of the many nuances here! ☺️

implementations/rust/ockam/ockam_executor/src/channel.rs
antoinevg: Ugh, this is indeed quite awful in a number of ways:

* The fundamental mismatch here is that `ockam_executor::channel` is a drop-in replacement for [`tokio::sync::mpsc::channel`](https://docs.rs/tokio/latest/tokio/sync/mpsc/fn.channel.html)
* Tokio takes a `length` parameter for its channels and does a dynamic allocation.
* `ockam_executor::channel`, on the other hand, is a thin wrapper around the `heapless` queue which does static allocation.
* So `length` is actually never used and (🙈) on `no_std` all queues are just allocated with a fixed size of 16.

At one point I was messing around with tracking queue lengths and item counts at runtime but gave that up in favor of tackling the underlying mismatch. (Maybe replacing the heapless queue with a dynamically allocated one, maybe switching tokio::sync::mpsc::channel for a statically allocated one, maybe something else?)

Why I had a `len()` method I really can't remember! 😂

Either way, I've cut these out for now and made a short doc note about what's going on here.

implementations/rust/ockam/ockam_executor/src/executor.rs
thomcc: Ah, oof. My concern with ignoring it is that it's often a real crash or UB — same as using memory from a functions stack frame after returning from it...

We have some other dodgy unpins here, and this crate is still WIP. Needs to be revisited later, tho.

implementations/rust/ockam/ockam_executor/src/channel.rs
thomcc: Ah, hmmmm. Alright then.

implementations/rust/ockam/ockam_executor/src/channel.rs
thomcc: I see. Wrapping the subset of tokio's channel that we need (so that it takes `const N: usize` that it just passes to the ctor) wouldnt be too bad I suspect, since we never use actually-dynamic sizes really. If it's not causing problems now lets wait til we need it, though.

(I suspect we'll have to move some of them to unbounded eventually anyway, to avoid deadlocks like #1947. I do have another solution in a comment there for how to handle it properly, but this kind of "gridlock" deadlock is common enough that I feel like we'll hit it at some point... But maybe not on embedded)

thomcc (commented):
LGTM

jared-s (approved):

PR #2420, opened 2022-01-05
Avoid using Workers for Vault and Identity

SanjoDeundiak:


mrinalwadhwa:
Since this changes the examples in guides. Let's discuss alternate names for Profile before we merge this.
Need to think about this more but some candidates - IdentityProfile, Identity or Identifier.

thomcc (approved):
This looks fine to me, assuming the renames are desired (probably ideally they should be in a different PR, but I don't think switching at this point would help anything). A lot what I might have to say was covered already.

thomcc:
Landed in #2476

PR #2428, opened 2022-01-10
add nightly check

metaclips:
This PR adds nightly check to our Ockam repo. Workflow dispatch can also be started using a reference `commit sha` which we can run all Rust tests upon.


jared-s:
Merged https://github.com/ockam-network/ockam/pull/2429

PR #2453, opened 2022-01-20
Design prototypes for low-overhead worker composition

spacekookie:
The problem at hand is that nesting workers for advanced messaging structures brings a lot of overhead with it. This PR (as well as the previous pipe and channel PRs) attempt to find a solution to this problem in a way that is both fun to use and performant.

This PR takes the approach of embedding a small actor system within a worker, each with their own address and state required for their function to be executed, but while sharing the worker relay, mailbox, and routing mechanisms.


antoinevg:
As Alan Kay et al put it, it's [turtles all the way down](https://www.informit.com/articles/article.aspx?p=1671638&seqNum=2)! 🐢

We can talk about the details but, in principle, I think this is the way to go. 

It also gives us a system which is tractable to  being reasoned about a'la [CSP](https://en.wikipedia.org/wiki/Communicating_sequential_processes).

Also probably the most powerful foundational abstraction we're going to get that won't  adding in proper types and re-inventing [Occam-π](https://en.wikipedia.org/wiki/Occam-%CF%80).

thomcc:
This looks great, and is a very interesting approach. I can't wait to dig into it when I get back.

implementations/rust/ockam/ockam/src/system/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+    where
+        A: Into<Address>,
+        F: Fn(WorkerSystem<W>) -> W,
+    {
+        let worker = fn_(self.sys);
+        self.ctx.start_worker(addr.into(), worker).await
+    }
+}
+
+/// A componasble worker system type
```
SanjoDeundiak, line 45: typo

implementations/rust/ockam/ockam/src/system/handler.rs
```diff
@@ -0,0 +1,23 @@
...
+where
+    C: Send + 'static,
+    M: Message,
+{
+    /// Run an initialise hook
+    ///
+    /// This gives this SystemHandler the chance to setup any state on
+    /// the worker or send messages to other workers (and hooks).
+    /// Finally it must return the address it wants to respond to
+    async fn initialize(&mut self, ctx: &mut C) -> Result<Address>;
```
SanjoDeundiak, line 19: We decide what is the correct address for a particular Worker when we start it, is there a reason why for SystemHandler we delegate that responsibility to the SystemHandler instance itself?

SanjoDeundiak (commented):
I think I got the overall idea of this PR, but I can't think of real world examples of how we could use this. Any suggestions?
Usually I needed multiple addresses per worker to handle different Message types while having shared Worker's state, it looks like this PR doesn't help with multiple Message types while allowing to split Worker's state.

spacekookie:
@SanjoDeundiak the idea behind this feature is to replace the [`BehaviorHook`](https://github.com/ockam-network/ockam/blob/develop/implementations/rust/ockam/ockam/src/pipe/behavior/mod.rs#L22) system currently used to attach additional behaviour to a pipe or channel.

The problem with the existing system is that different behaviours may conflict with each other (or require special interactions) that are hard to express. Generally I got the feeling that this approach would get messy pretty quickly.

I'm also still working on this, ideally every `SystemHandler` would be able to accept it's own message type (not just the parent type) -- even if it's just a small shim that decodes payloads on the fly.

implementations/rust/ockam/ockam/src/system/handler.rs
spacekookie: The reason why I left this up to the SystemHandler instance at the moment is that potentially a worker could add new SystemHandlers during runtime in response to external commands. But I do also think that this makes initialisation easier (you just have to pass a SystemHandler istance, not that _and_ and address you can potentially typo.

spacekookie:
@thomcc @mrinalwadhwa i think this is ready to merge today

implementations/rust/ockam/ockam/src/metadata.rs
```diff
@@ -0,0 +1,159 @@
...
+        if self.ext.is_none() {
+            self.ext = Some(Metadata(BTreeMap::new()));
+        }
+
+        self.ext.as_mut().unwrap().insert(key.into(), val);
+        self
+    }
+
+    /// Get the current scope metadata
+    pub fn get_scope(&self) -> &Vec<Vec<u8>> {
```
thomcc, line 77: Nit: Given that these all are public, I'm not sure having getters is worthwhile. (And if we do want them, probably this one should be `&[Vec<u8>]` too, and they shouldn't use the `get_` prefix, since that's not generally considered good style in rust)

thomcc (approved):
This looks fine to me. I have some very minor API design nits. Sorry for not getting to it sooner.

spacekookie:
@thomcc done

thomcc:
Landed in #2477.

PR #2457, opened 2022-01-25
feat(rust): ssh secure channel echoer cli

jared-s:
Integrates ssh public and private key reading and echoer example into cli. This iteration is based on the `ssh_example` project in the `ockam_examples` folder.

There are a few issues of note:
- CLI specific logging cannot be initialzied, because the Ockam library itself is initializing a logger.
- The existing logger cuts off lines after a certain width. Some of the error messages are being truncated.
- Only Ed25519 keys are supported. Create a test keypair with `ssh-keygen -t ed25519`
- The secure channel listener (`channel-listen`) and client (`channel`) are hard coded to initialize an `Echoer` worker as in the example code. A method of selecting which workers are registered needs to be developed.


implementations/rust/ockam/ockam_command/src/command/channel.rs
```diff
@@ -0,0 +1,57 @@
...
+        channel_name: &str,
+        message: &str,
+    ) -> Result<(), AppError> {
+        let _tcp = TcpTransport::create(ctx).await?;
+
+        let mut vault = VaultSync::create(ctx, SoftwareVault::default()).await?;
+        let vault_address = vault.address();
+        let mut alice = Entity::create(ctx, &vault_address).await?;
+
+        let secret_key = std::fs::read_to_string(secret_key_path).unwrap();
```
thomcc, line 25: We should probably try to return an error if the path provided doesn't exist, rather than panicking.

implementations/rust/ockam/ockam_command/src/command/channel.rs
```diff
@@ -0,0 +1,57 @@
...
+
+        let secret_key = ssh_key::PrivateKey::from_openssh(&secret_key)
```
thomcc, line 27: So the unwraps here are only concerning since they happen on e.g. rsa keys, which is a pretty easy mistake to make.

implementations/rust/ockam/ockam_command/src/config.rs
```diff
@@ -1,26 +1,44 @@
...
+        name: String,
+    },
+    Channel {
+        #[clap(short, long)]
+        channel: String,
+
+        #[clap(short, long, default_value = "secure_channel")]
+        name: String,
+
+        #[clap(short, long, default_value = "Hello, Ockam!")]
```
thomcc, line 39: It's not clear to me this should have a default value.

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -13,6 +15,12 @@ pub enum AppError {
     #[error("invalid argument")]
     InvalidArgument,
 
+    #[error("i/o error")]
+    IOError(std::io::Error),
```
thomcc, line 19: In Rust it's considered better style to camel-case abbreviations (avoids `XMLHttpRequest`-style situations)
```suggestion
    IoError(std::io::Error),
```

thomcc: More broadly we'll probably want to use something like `anyhow` for the CLI, but that can be addressed later.

thomcc (approved):
This is probably fine.

The error handling needs some work, but part of that is on me (although I suspect the error requirements for the library and CLI will be different — as they often are in Rust code). This can be addressed in a followup, though.

implementations/rust/ockam/ockam_node/Cargo.toml
```diff
@@ -45,6 +45,7 @@ dump_internals = []
 [dependencies]
 ockam_core = { path = "../ockam_core", version = "^0.43.1-dev", default_features = false }
 ockam_macros = { path = "../ockam_macros", version = "^0.4.1-dev"}
+thread_local = ">=1.1.4"
```
metaclips, line 48: We don't need to add this dep to fix the [rust sec issue](https://github.com/ockam-network/ockam/runs/4930245013?check_suite_focus=true#step:7:22). A `cargo update` should update the dependency.

jared-s:
Thanks guys, fixed those things up.

metaclips (approved):
🚀

PR #2471, opened 2022-01-28
docs(rust): readmes for example tools

jared-s:
READMEs that describe the example tools. @thomcc please let me know if this makes sense


thomcc (approved):
Yep, makes complete sense.

PR #2478, opened 2022-02-01
release script update

### LINKED ISSUE #2462: Update release script to work with `git-cliff` v0.5.0+
The current version of git-cliff is 0.5.0, and it seems it no longer works with `tools/scripts/release/changelog.sh`. Instead, it issues the following error many times:
```
error: Found argument '--commit-path' which wasn't expected, or isn't valid in this context

USAGE:
    git-cliff --unreleased
```

Downgrading to v0.4.2 worked, so this isn't a blocker, but is something we should fix.

CC @metaclips

### LINKED ISSUE #2463: `crate-bump.sh` script uses bash features not supported in stock macOS.
By default, macOS comes with a rather old installation of `bash`:

```sh
$ bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin21)
Copyright (C) 2007 Free Software Foundation, Inc.
```

This version doesn't seem to support `declare -A` which we use in `tools/scripts/release/crate-bump.sh`. Instead, I get an error like the following:

```sh
tools/scripts/release/crate-bump.sh: line 19: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
```

After this, it appears to work (I guess this version of bash does support associative arrays, but just doesn't support the `declare -A` flag), so I suspect we can just delete that line? Not sure.

CC @metaclips

### LINKED ISSUE #2464: Enable "exit on error" (`set -e`) on in shell scripts
In #2462 and #2463, there were errors in the release scripts, and the scripts continued silently ignoring them. While no damage was done in these cases, in general the result of this could be pretty bad. Bash supports a mode for aborting the script in this case, which can be enabled by `set -e`. I believe it also could be enabled by changing the shebang to `#!/usr/bin/env bash -e`, but that needs some testing.

See https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html for documentation on `set -e`.

Anyway, we should probably do this to avoid anything like this happening in a case that actually causes issues.

CC @metaclips

### LINKED ISSUE #2466: Remove the "dev bump" part of the release workflow (e.g. don't use `-dev` versions in between publishes)
We've adopted a dev-bump workflow after releasing, meaning we bump all our crates to -dev versions after the release. There are a lot of downsides to this in Rust development — for example, it breaks `cargo`'s `[patch.crates-io]` feature.

I argued heavily (and convincingly enough for them to follow) to change the default behavior of this in cargo-release, because of these downsides — see my comments in this thread, starting at https://github.com/crate-ci/cargo-release/issues/82#issuecomment-933100970 for my arguments. So, as you can imagine, I don't think we should do it here.

It also makes our rebases more difficult, although we already have a problem here regardless of this.

metaclips:
- Update release scripts to exit on error. 
- We can tag a single crates with this update. 
- Update git cliff to version 0.5.0+

fixes #2466 
fixes #2464 
fixes #2463 
fixes #2462


tools/scripts/release/tagging.sh
```diff
@@ -37,4 +37,14 @@ for crate in ${updated_crates[@]}; do
 * [CHANGELOG](https://github.com/ockam-network/ockam/blob/${name}_v$version/implementations/rust/ockam/$name/CHANGELOG.md)";
 
     gh release create --draft --notes "$text" -t "$name v${version} (rust crate)" "$tag" --target $COMMIT_SHA
+}
+
+if [[ ! -z $TAG_SINGLE_CRATE ]]; then
+    echo "Tagging only $TAG_SINGLE_CRATE crate"
+    tag_crate $TAG_SINGLE_CRATE
```
thomcc, line 44: Hmm, I sort of wonder if tag / release should be different operations in the future, once we get to a "not every release is semver-breaking" situation.

That said, this is great for now.

thomcc (approved):
Wonderful. Thanks for keeping the documentation up to date too.

PR #2482, opened 2022-02-03
Rust pipe worker re-implementation

spacekookie:
This PR re-implements Rust pipes with the new system worker composition feature.

So far the code focuses on gaining a better understanding of how these systems should be initialised, and the way users have to interact with the API.

To that end this PR adds two new builder types:

* `PipeBuilder`, a simple API for constructing pipe senders and receivers, as well as dynamic pipe handshakes
* `SystemBuilder`, an abstraction to create networks of SystemHandlers, each with a "next" address to re-route messages to.


implementations/rust/ockam/ockam/src/pipe2/mod.rs
```diff
@@ -0,0 +1,107 @@
+// If `pipe` is so great, why is there no `pipe2`?
```
thomcc, line 1: ... Well, it's sure hard to argue with that 😂. Surprised we don't get this from users, even.

(TBH, I do think pipe isn't an ideal name for either of them, since it's so close and easily confused with channel, but that for sure doesn't need to be solved here)

spacekookie: Yea, the terminology of "pipe", "channel", and "stream" is a bit overlapping and confusing. I'd be especially curious to hear from unix developers who have a very different understanding of what a pipe is as to how this lingo may be confusing.

It could be something we'll just have to accept and have good documentation (and visualisations) on and then we need to use those terms consistently.

implementations/rust/ockam/ockam/src/pipe2/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+/// A builder structure for pipes
+///
+/// The easiest way to create a pipe is via
+/// `builder.basic().static()`.
+pub struct PipeBuilder {
+    send_hooks: SystemBuilder<Context, OckamMessage>,
+    recv_hooks: SystemBuilder<Context, OckamMessage>,
+    /// The selected pipe initialisation mode
+    mode: Mode,
+    /// Peer information
```
hairyhum, line 42: Why does it have both `peer` and `recv`? Is it bidirectional?

implementations/rust/ockam/ockam/src/pipe2/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// initialisation request
+    pub fn receive<A: Into<Address>>(mut self, addr: A) -> Self {
+        self.recv = Some(addr.into());
+        self
+    }
+
+    /// Set this pipe to enforce the ordering of incoming messages
+    pub fn enforce_ordering(mut self) -> Self {
+        // A pipe can currently only have two types of hooks.
+        // "Delivery" and "Ordering".  Because we want "ordering" to
```
hairyhum, line 102: Minor note, but this sounds like something to be done on `build`

implementations/rust/ockam/ockam/src/pipe2/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+                "ordering",
+                hooks::SenderOrdering::default(),
+            )
+            .default(next_send_addr);
+        self
+    }
+
+    /// Enable the delivery guarantee behaviour on this pipe
+    ///
+    /// Additional behaviours can be added to compose a custom pipe
```
hairyhum, line 135: Is there a `add_cusom_behaviour` API somewhere? Do we want this pipe to have a custom behaviour?

hairyhum (commented):
Can you please add "usage" or "doc" section in the docs or at least PR description?

implementations/rust/ockam/ockam/src/pipe2/mod.rs
spacekookie: The builder is bi-directional, yea. So you use the same builder for both building the sender and receiver

implementations/rust/ockam/ockam/src/pipe2/mod.rs
spacekookie: So far there isn't, but fundamentally nothing is stopping us from allowing custom behaviour on pipes.

implementations/rust/ockam/ockam/src/pipe2/mod.rs
spacekookie: I was kinda on the fence about this. On one hand, yes, it feels like this is a bit too complex to be done in-line to these attribute functions. And moving this code to `build` will also simplify it.

On the other hand it does limit the behaviour that users can specify. But maybe that's okay. I opted for the slightly more flexible and complicated solution to start

implementations/rust/ockam/ockam/src/pipe2/mod.rs
hairyhum: Where would the custom behaviour go in the pipeline?
Since we control order of "ordering" and "redelivery" behaviours.

spacekookie: I changed the initialisation of the pipe to do all the work in `build` for now. But I suspect we will want custom hooks on the sender side to go last, and on the receiver side to be first.
We could also have two functions to add hooks (`at_start` and `at_end` or something). But it's not needed for no and I wanna get this PR merged so let's leave it out for now

implementations/rust/ockam/ockam/src/system/builder.rs
```diff
@@ -0,0 +1,296 @@
...
+        self.chain_for("default", addr, prev_id, id, handler)
+    }
+
+    /// Add a new handler to the builder, re-addressing the previous
+    /// handler's named route to this one
+    pub fn chain_for<A, H>(
+        &mut self,
+        rule: impl Into<String>,
+        addr: A,
+        prev_id: impl Into<String>,
```
thomcc, line 129: We could probably reduce this to AsRef<str>, not that it matters much.

implementations/rust/ockam/ockam/src/system/builder.rs
```diff
@@ -0,0 +1,296 @@
...
+    /// a second instance of every address to exist on the same node.
+    ///
+    /// This problem may be solved by having scoped addresses in the
+    /// future.
+    ///
+    /// This function assigns a new address to every handler and then
+    /// changes routes that point to the old addresses.  Any route
+    /// that is not in the internal set (i.e. the fin address routes)
+    /// are replaced with the externally provided fin-address.
+    pub fn readdress(&mut self, fin_addr: &Address) {
```
thomcc, line 159: Hmm, this functions name reads kind of odd to me readdress -> read-dress or re-address (obviously the latter in this case).

I wonder if we can come up with a better name for it... For now the documentation should be sufficient.

implementations/rust/ockam/ockam/src/system/builder.rs
```diff
@@ -0,0 +1,296 @@
...
+            system.attach_boxed(addr, inner);
+        }
+
+        Ok(system)
+    }
+}
+
+/// Builder API for a single SystemHandler
+///
+/// This type may panic during destruction if not properly initialised
```
thomcc, line 219: Not a fan of this in general, but the work to do in a typestatey way is not worth it, and having a `fn finish(self) -> Result<()>` is also not good (because its an API usage error).

So I suppose this is fine in this case.

implementations/rust/ockam/ockam/src/system/handler.rs
```diff
@@ -1,17 +1,47 @@
...
+use dyn_clone::DynClone;
+use ockam_core::compat::{boxed::Box, collections::BTreeMap, string::String};
+use ockam_core::{Address, Message, Result, Routed};
 
 /// Handle a single type of message for a worker system-address
 ///
 /// A handle may re-emit messages to the worker system, or to the
 /// Ockam runtime.  All state associated with a particular protocol
 /// must be contained in the type that implements this trait.
+#[clonable]
```
thomcc, line 11: I'm familiar with `dyn_clone` (and have no issues with our use of it), but not `dyn_clonable`.

What does this save us? If it's just nicer syntax than `dyn_clone::clone_trait_object!(SystemHandler);` or whatever, I'm not sure this is worth an additional dependency, let alone on a proc macro (which have a much large impact on build times than library crates).

If it's more significant than perhaps it's fine, though

implementations/rust/ockam/ockam/src/system/handler.rs
```diff
@@ -1,17 +1,47 @@
...
+    /// workers, and other internal handlers.  To allow workers to
+    /// create behaviour pipelines, we need to pre-define "routes" for
+    /// a SystemHandler (i.e. where is a message sent after it is done
+    /// processing it).
+    ///
+    /// In most cases this only requires a "default" route, but may
+    /// use different routing labels in more complicated setups (to
+    /// build processing graphs, instead of pipelines).
+    ///
+    /// It is highly recommended to use the
```
thomcc, line 31: Should this piece of documentation should go on the trait docs, rather than the function docs?

implementations/rust/ockam/ockam/src/system/hooks/delivery.rs
```diff
@@ -0,0 +1,170 @@
...
+};
+
+#[derive(Clone, Default)]
+pub struct SenderConfirm {
+    next: Option<Address>,
+    journal: BTreeMap<Address, OckamMessage>,
+}
+
+fn to_str(v: &[u8]) -> &str {
+    core::str::from_utf8(v).unwrap()
```
thomcc, line 15: We probably should handle this better, given that we don't require metadata values be UTF-8 (or strings at all)...

implementations/rust/ockam/ockam/src/system/hooks/delivery.rs
```diff
@@ -0,0 +1,170 @@
...
+                        "ockam.pipe.type",
+                        "ockam.pipe.resend_notify".as_bytes().to_vec(),
+                    )
+                })
+                .await?
+                .with_seconds(5)
+                .spawn();
+
+                // Forward the new message to the next address
+                ctx.send(self.next.as_ref().unwrap().clone(), outer_msg)
```
thomcc, line 74: This unwrap shouldn't trigger unless this is called prior to initialize (which is impossible[^1]), right?

[^1]: Short of someone manually calling it, which we don't/shouldn't support.

implementations/rust/ockam/ockam/src/system/hooks/delivery.rs
```diff
@@ -0,0 +1,170 @@
...
+                    .await?;
+            }
+
+            // For any "ACK" we receive we can delete the
+            // corresponding ACK id from the journal
+            Some(ref tt) if tt == &"ockam.pipe.ack" => {
```
thomcc, line 80: Surprised you can't just `Some("ockam.pipe.ack")` here (and below). It doesn't make much difference though.

implementations/rust/ockam/ockam/src/system/hooks/mod.rs
```diff
@@ -0,0 +1,11 @@
+#![allow(unused)] // FIXME
```
thomcc, line 1: FIXME 😉

thomcc (approved):
Incredible work, honestly. Wonderfully documented too. I have a few nits, but none of them are serious and it can probably land as-is if you'd rather.

implementations/rust/ockam/ockam/src/system/handler.rs
thomcc: This is probably the only thing I'd like to change if possible before landing.

implementations/rust/ockam/ockam/src/system/hooks/mod.rs
thomcc: (Oh, and this)

spacekookie:
@thomcc can we please just merge this and address the nitpicks in GA review?

implementations/rust/ockam/ockam/src/system/handler.rs
spacekookie: `dyn_clone::clone_trait_object!(SystemHandler); ` in this context is extremely hard to use because of the generic type bounds on `SystemHandler`

implementations/rust/ockam/ockam/src/system/hooks/delivery.rs
spacekookie: I don't really see the issue with this considering it's an internal implementation detail to the protocol between different pipe worker endpoints. In this case we _do_ enforce UTF-8 but this detail isn't exposed to the user at all.

implementations/rust/ockam/ockam/src/system/hooks/delivery.rs
spacekookie: Correct. `self.next` is set during worker initialisation. If someone were to create a worker instance and then manually call handle_message on it (with the correct message type and wrapping) _then_ it would fail

implementations/rust/ockam/ockam/src/system/hooks/mod.rs
spacekookie: woops :see_no_evil:

spacekookie: It's strange that my git-grep search for "unused" didn't show this line O.o

thomcc:
Uhh, this needs too many changes to rebase, and I'm guessing at a lot of them. You mind rebasing it for me? I'll land it tomorrow if you do.

thomcc:
Hm, sorry, thought this was ready.

Issue #2378, opened 2021-12-17
Integrate Tokio Console with Ockam Node


mrinalwadhwa:
https://tokio.rs/blog/2021-12-announcing-tokio-console
https://github.com/ockam-network/ockam/tree/develop/implementations/rust
https://github.com/ockam-network/ockam/tree/develop/implementations/rust/ockam/ockam_node

mattjperez:
@mrinalwadhwa Can I grab this?

mrinalwadhwa:
@mattjperez that would be awesome! ❤️ 
@thomcc & @spacekookie can help if you have questions along the way.

mattjperez:
@thomcc @spacekookie to use tokio-console, you currently need to use `RUSTFLAGS="--cfg tokio_unstable"`.

I've feature-gated the use of tokio-console behind `#[cfg(console)]`, but this would then require users to run `RUSTFLAGS="--cfg tokio_unstable --cfg console" cargo build` for their projects, which feels messy. However, once tokio-console is stabilized, it will go away naturally.

Thoughts?

mattjperez:
Also, this thing is pretty cool. 

I'm using the echoer and app examples from the [How-To Guides](https://www.ockam.io/learn/how-to-guides/rust/02-worker/) guides (looping 20 times) for testing and can see stuff like this.

![2022-02-06_20-46](https://user-images.githubusercontent.com/41358385/152712719-74286917-4382-4998-bd94-2448c3d1c261.png)
![2022-02-06_20-48](https://user-images.githubusercontent.com/41358385/152712754-7086c847-dfd0-4c59-b2e1-788c617fb9b3.png)

PR #2484, opened 2022-02-07
feat(rust): add tokio-console tracing to ockam_node

mattjperez:
#2378 

This commit:
1. adds the tokio/console-subscriber's tracing Layer to ockam_node's
`setup_tracing()` function.
2. adds the `ockam_node/console` feature for conditional compilation.

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/node.rs
+++ b/implementations/rust/ockam/ockam_node/src/node.rs
@@ -1,7 +1,8 @@
 use crate::{Context, Executor};
 use ockam_core::{Address, Passthrough};
+use tracing_subscriber::prelude::*;
 #[cfg(feature = "std")]
-use tracing_subscriber::{filter::LevelFilter, fmt, EnvFilter};
+use tracing_subscriber::{filter::LevelFilter, EnvFilter};
 
 /// A minimal worker implementation that does nothing
 pub struct NullWorker;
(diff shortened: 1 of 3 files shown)


implementations/rust/ockam/ockam_node/src/node.rs
```diff
@@ -31,16 +32,57 @@ pub fn start_node() -> (Context, Executor) {
...
+#[cfg(feature = "console")]
+/// Utility to setup tracing-subscriber from the environment.
+/// Enables usage with the `tokio-console` diagnostic toolkit.
+///
+/// # Usage
+/// The `tokio_unstable` cfg flag, which enables experimental APIs in Tokio, must
+/// be enabled. It can be enabled by setting the `RUSTFLAGS` environment variable
+/// at build-time:
+/// ```shell
+/// $ RUSTFLAGS="--cfg tokio_unstable" cargo build
```
thomcc, line 64: I'm not sure what this is referring to, since we don't have a `tokio_unstable` feature?

EDIT: Ah, This is cfg not features. Hmm...

implementations/rust/ockam/ockam_node/src/node.rs
```diff
@@ -31,16 +32,57 @@ pub fn start_node() -> (Context, Executor) {
...
+/// ```
+///
 fn setup_tracing() {
     #[cfg(feature = "std")]
     {
         let filter = EnvFilter::try_from_env("OCKAM_LOG").unwrap_or_else(|_| {
             EnvFilter::default()
                 .add_directive(LevelFilter::INFO.into())
                 .add_directive("ockam_node=info".parse().unwrap())
+                .add_directive("tokio=trace".parse().unwrap())
```
thomcc, line 74: I'm not sure these are the right level. That's quite verbose -- IIRC tokio is very noisy at this level.

implementations/rust/ockam/ockam_node/src/node.rs
```diff
@@ -31,16 +32,57 @@ pub fn start_node() -> (Context, Executor) {
 }
 
 /// Utility to setup tracing-subscriber from the environment
+#[cfg(not(feature = "console"))]
```
thomcc, line 35: I think we don't want to duplicate this entirely -- we should instead move the CFG into the body of `setup_tracing`. Otherwise I think these will drift.

implementations/rust/ockam/ockam_node/src/node.rs
mattjperez: Is there another level you think would be a better fit?

thomcc: Info, probably.

etorreborre:
Hi @mattjperez I am doing a pass on PRs which have been opened for a long time. Sorry for such a delay in coming back to it! Is it something that you plan to resume working on? Thanks

etorreborre:
@mattjperez Thank you for all the time you put into this 🙏. Since the branch has now diverged from `develop` I am going to close this PR for now. Hopefully someone else will come along and take your great work further from this branch
https://github.com/build-trust/ockam/tree/mattjperez/ockam_node_add-tokio-console

PR #2488, opened 2022-02-09
build(rust): update examples workspace setup

SanjoDeundiak:


thomcc:
Hm, thanks. One note: the `no_std` example shouldn't be in the archive subdir as that's the only thing that exercises that functionality at all.

Also, in case you haven't seen it, I have a more involved plan here https://docs.google.com/document/d/1VlXcaC57BpAAHH-LOuXYTDX8pgj7W3Vpvbs0Ahfg44E/edit. I've been vaguely meaning to convert to markdown somehow and make an issue too.

Anyway, what you've done in this PR is similar to the first part of the plan there, so I'm in favor.

PR #2492, opened 2022-02-10
Introduce replacement error type.

thomcc:
This introduces the new error handling type, which has been mercilessly ripped out from the other branch I was working on, but is a lot less of a "rewrite the world" kind of change, and something we can migrate to. For example, this doesn't migrate us to it at all. That's something we'll do as we clean up some of the code.

There are a couple pieces that aren't *quite* finished, such as backtrace/spantrace handling (for one, do we want/need both of these?), and I suspect the things I'm doing here are... suboptimal on embedded. I'll write down a concrete list after the demo meeting, but I don't think anything is a deal-breaker quite yet (unless no_std is broken, which should be an easy fix).

Anyway, there's quite a bit of documentation in the comments, so see that for more details!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -158,6 +158,7 @@ dependencies = [
  "miniz_oxide",
  "object",
  "rustc-demangle",
+ "serde 1.0.136",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


SanjoDeundiak:
Replaced by #2566

PR #2495, opened 2022-02-10
Hairyhum/discovery service

hairyhum:
## Proposed Changes

Adding `discovery_service` which supposed to keep track of services available on the cloud and make then accessible to the clients.

Currently using the registry model from #2467 

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2448,9 +2448,9 @@ dependencies = [
 
 [[package]]
 name = "regex"
-version = "1.5.4"
+version = "1.5.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d07a8629359eb56f1e2fb1652bb04212c072a87ba68546a04065d525673ac461"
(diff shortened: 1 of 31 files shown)


hairyhum:
In the current prototype services need to register with discovery on startup. For that purpose start function is extended by service provider to add registration after gen_server is started.
Downside of that approach is that the state is stored in the discovery service and it might restart leaving its state inconsistent with the actual service configuration.

Alternatively, service discovery may read current started services info from the supervisor state. This makes discovery service state more consistent with the actual services running on the node.
Downside is that services on other nodes would not be visible in that case, as a solution a forwarding worker may be started as a service on the discovery node and tunnel messages to the remote node in such way that for user it would appear as if they interact with a service on discovery node. 

Either way remote node services implementation would require a solution to transport and channel tunneling to remote nodes through or around the access node, which may have additional performance implications.

I'm leaning towards using the supervisor API for service discovery for now, because for the local services it's more consistent.
The client API would stay mostly unchanged and if needed we can switch back to the registration approach.

PR #2502, opened 2022-02-16
feat(rust): add worker ready status

spacekookie:
## Current Behaviour

A problem we repeatedly run into is that initialisation code of one
worker takes longer than another worker is waiting for.  This is a
common problem in multi-processor environments and one that is not
easy to solve.

## Proposed Changes

This commit adds a draft API that allows workers to wait for the
readiness of other addresses.  This is done by marking an address as
"ready" only after its initialization function has been called.

The way that this is currently implemented in the Router breaks a lot
of existing tests because we have implicit race conditions in the
existing examples and tests.

An alternative implementation of this feature doesn't use the router
state value and instead adds a separate 'ready' boolean that can only
be checked by this one test.

@SanjoDeundiak @antoinevg @thomcc thougths?

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/context.rs
+++ b/implementations/rust/ockam/ockam_node/src/context.rs
@@ -627,4 +627,27 @@ impl Context {
     pub async fn set_access_control(&mut self) -> Result<()> {
         unimplemented!()
     }
+
+    /// This function is called by Relay to indicate a worker is initialised
+    pub(crate) async fn set_ready(&mut self) -> Result<()> {
+        self.sender
+            .send(NodeMessage::set_ready(self.address()))
+            .await
+            .map_err(|_| Error::InternalIOFailure)?;
+        Ok(())
+    }
+
+    /// Wait for a particular address to become "ready"
+    pub async fn wait_for<A: Into<Address>>(&mut self, addr: A) -> Result<()> {
+        let (msg, mut reply) = NodeMessage::get_ready(addr.into());
+        self.sender
+            .send(msg)
+            .await
+            .map_err(|_| Error::InternalIOFailure)?;
+
+        // This call blocks until the address has become ready or is
+        // dropped
+        reply.recv().await.ok_or(Error::InternalIOFailure)??;
+        Ok(())
+    }
 }
--- a/implementations/rust/ockam/ockam_node/src/error.rs
+++ b/implementations/rust/ockam/ockam_node/src/error.rs
@@ -13,10 +13,8 @@ pub enum Error {
     FailedStartProcessor,
     /// Worker start failed because the address was already taken
     WorkerAddressTaken,
-    /// The requested worker address is unknown
-    UnknownWorker,
-    /// The requested processor address is unknown
-    UnknownProcessor,
+    /// The requested address is unknown
+    UnknownAddress,
     /// Unable to stop a worker
     FailedStopWorker,
     /// Unable to list available workers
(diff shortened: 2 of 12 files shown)


antoinevg:
I think this is the right shape for the solution.

It's orders better than a boolean flag, that approach always seems to result in race conditions between different parts of the system trying to modify the flag!

The only thoughts I have to add to this is that there's something to be said for separating "liveness" and "readiness" into separate state values.

This may make it easier to disambiguate supervisor checks because e.g. a system can be alive but not ready.

SanjoDeundiak:
@spacekookie  I'm fine with the fact, that after ctx.start_worker call succeeded - you can immediately send messages to it's mailbox (which is the main job of a Worker). While if there's some long setup process (e.g. waiting for Worker::initialize to finish in the simplest case, but also may be long and complicated process) - you can handle that in a separate function and use a callback address. I have been using that approach for some time, for example SecureChannel::create spawns workers and initiates process of creating the channel, then it waits for a message that indicates process completion, and then returns the address of newly crated channel, that can be used immediately. [here](https://github.com/ockam-network/ockam/blob/develop/implementations/rust/ockam/ockam_identity/src/channel/secure_channel_worker.rs#L141)

spacekookie:
I think being able to send messages to a worker before its fully initialised is _fine_ per say (and I'm currently changing the code in a way that makes that possible again).

But for calls like `TcpTransport::create` I think it's important that we can check _inside the function_ whether the system has actually initialised and is ready for users to use.

thomcc (dismissed):
Seems reasonable to me.

SanjoDeundiak:
@spacekookie I meant that the fact that you can send messages to the Worker immediately after it was started (but not initialized) is indeed correct behaviour. Did it stop working?

I think ```system has actually initialised and is ready for users to use``` can mean very different things, it doesn't necessary mean that ```Worker::initialize``` function has finished - and this PR won't help in such situations (and there are plenty of such situations throughout our code). If we're talking about ```TcpTransport::create``` bug - it's fixed much easier by moving register call which doesn't require this polling loop to the Router. Do you have examples where checking for readiness the way this PR does is the way to go?

spacekookie:
> I think system has actually initialised and is ready for users to use can mean very different things, it doesn't necessary mean that Worker::initialize function has finished, it can much more complicated

That's true in a way, we do have workers that rely on handshakes and delayed initialisation patterns. And indeed this PR doesn't help in any of those cases.

But I do feel like having an API that _gives you the option_ to wait for the initialisation of a worker to complete is a good pattern.

Let's look at the TcpTransport bug: yea you can move the `register` call outside of the worker, but now you're configuring a worker from outside of it, violating separation of concerns. I've written a lot of code that does this too, for example the pipe and channel modules in the `ockam` crate and I think it's a shame that we can't actually put any useful initialisation into the `initialize` function because it will always result in a race condition.

This feels like a foot gun that we need to make sure users don't run into or end up having to over-complicate their code in order to initialise their systems correctly.

mrinalwadhwa:
> But I do feel like having an API that gives you the option to wait for the initialisation of a worker to complete is a good pattern.

+1 I think this should be the default.

Erlang's GenServer is a good proven reference here. 

> gen_server:start_link is synchronous. It does not return until the gen_server has been initialized and is ready to receive requests.

https://www.erlang.org/doc/design_principles/gen_server_concepts.html#starting-a-gen_server 
https://www.erlang.org/doc/man/gen_server.html#start_link-4

SanjoDeundiak:
@spacekookie speaking of the implementation of that API. Are there any benefits of this polling the Router behaviour vs using callback address? Isn't that hurting performance a lot? It looks like one long Worker initialization and few places where you're waiting for it to be ready is going to kill the system

spacekookie:
> Are there any benefits of this polling the Router behaviour vs using callback address?

No, I can change that implementation detail. I opted for polling yesterday because it was a lot quicker to implement

thomcc (approved):
LGTM, going to merge this before the release today.

PR #2505, opened 2022-02-17
feat(rust): add `Echoer` worker to ockam crate

SanjoDeundiak:


implementations/rust/ockam/ockam/src/workers/echoer.rs
```diff
@@ -0,0 +1,19 @@
...
+
+pub struct Echoer;
+
+#[crate::worker]
+impl Worker for Echoer {
+    type Context = Context;
+    type Message = String;
+
+    async fn handle_message(&mut self, ctx: &mut Context, msg: Routed<String>) -> Result<()> {
+        println!("Address: {}, Received: {}", ctx.address(), msg);
```
thomcc, line 14: I'm not 100% sure this is appropriate to be in the `ockam` crate -- It's generally bad for library code to println without indication. Can you elaborate on the motivation here?

SanjoDeundiak: I'll replace it with log message

thomcc (approved):
I still think this is a little dubious, but can imagine a place where it fits in in the future.

PR #2509, opened 2022-02-17
Sanjo/heartbeat

SanjoDeundiak:


implementations/rust/ockam/ockam_node/Cargo.toml
```diff
@@ -52,6 +52,7 @@ tokio = { version = "1.8", default-features = false, optional = true, features =
     "rt-multi-thread",
     "macros",
 ] }
+futures = "0.3.10"
```
thomcc, line 55: Hmm, I think we may need to disable some features for no_std?

thomcc (commented):
Seems fine if the futures dep is adjusted to be okay on no_std -- It seems to pull in std by default, according to https://docs.rs/crate/futures/latest/features, unless I'm missing something.

spacekookie (commented):
This looks good, although with something called "heartbeat" I would expect the API to re-schedule events for me, which this doesn't appear to do.

The two choices I see here is either to make `Heartbeat` implicitly re-schedule sending events (i.e. you pick an interval and it sends a copy of the provided message every interval), _or_ we rename this type to be more in-line with the [`DelayedEvent`](https://github.com/ockam-network/ockam/blob/develop/implementations/rust/ockam/ockam/src/delay.rs) type we have in the ockam crate (which this can then replace)

implementations/rust/ockam/ockam_node/Cargo.toml
SanjoDeundiak: @thomcc interesting, CI should check that no_std build is working

thomcc (approved):

thomcc:
Landed in #2511

PR #2503, opened 2022-02-16
feat(elixir):add restart_type option for workers

calebbourg:
## Checks

- [ ] All commits in this Pull Request are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- [ ] All commits in this Pull Request follow the Ockam [commit message convention](https://www.ockam.io/learn/how-to-guides/contributing/CONTRIBUTING#commit-messages).
- [ ] I accept the Ockam Community [Code of Conduct](https://www.ockam.io/learn/how-to-guides/high-performance-team/conduct/).
- [ ] I have accepted the Ockam [Contributor Licence Agreement](https://www.ockam.io/learn/how-to-guides/contributing/cla/) by adding my Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/ockam-network/contributors/blob/master/CONTRIBUTORS.csv) file in a separate pull request to the [ockam-network/contributors](https://github.com/ockam-network/contributors) repository.

PR #2515, opened 2022-02-18
Rebased version of ockam_transport_ble

thomcc:
Closes #2433

PR #2516, opened 2022-02-18
feat(elixir):add restart_type option for workers

hairyhum:
Rebase of #2503


thomcc:
Ah, sorry, I think we had a mid-air collision with this and #2515

PR #2530, opened 2022-02-23
update rust version to `1.58.1`

metaclips:
Bump Rust version to 1.58.1


thomcc (approved):
Sure. I think theres another release tomorrow (https://www.whatrustisit.com), its a shame not to get that, but I gather we need this ASAP to fix bustage.

PR #2533, opened 2022-02-24
Reorganize and document `ockam` crate

thomcc:
This is the GA review and updates for the `ockam` crate. It's still a work in progress (I haven't had as productive a week as I had hoped, for personal reasons).

After I finish up, land these changes, etc, it should be handed to @spacekookie, who has more knowledge about this and has a better idea about how some of the APIs will evolve in the wake of the updates from [pipe2](https://github.com/ockam-network/ockam/pull/2482).

That said, as mentioned, not *quite* at that point yet.


implementations/rust/ockam/ockam/src/lib.rs
```diff
@@ -1,5 +1,7 @@
 //! Ockam is a library for building devices that communicate securely, privately
```
ashleygwilliams, line 1: We may want to update this line to be consistent with other taglines across Ockam's properties, e.g. "End-to-end encryption and mutual authentication for distributed applications." I think the tagline here is confusing because it speaks about "building devices".

implementations/rust/ockam/ockam/src/unique.rs
```diff
@@ -1,27 +1,23 @@
 use ockam_core::compat::rand::{thread_rng, Rng};
 use ockam_core::compat::string::String;
 
-/// A simple generator for unique, human-readable identifiers suitable
-/// for use in distributed systems.
-pub struct Unique;
+/// A simple generator for unique, human-readable identifiers suitable for use
```
ashleygwilliams, line 4: In general, using the word "simple" in docs is not ideal. Would it be possible to drop it? Or is it simple in relation to a more complex generator?

ashleygwilliams (commented):
A few comments, nothing major- in general I try to avoid words like "easy", "simple", "just" in documentation as it can be unnecessarily frustrating signal that folks who might run into issues while getting started. 

There are some TODO comments, I'm not sure what our policy is on those.

PR #2539, opened 2022-02-28
release script update

metaclips:
This PR fixes an issue where we don't release/bump crates whose inter-dependency is bumped.

### DIFF
--- a/tools/scripts/release/crate-bump.sh
+++ b/tools/scripts/release/crate-bump.sh
@@ -14,8 +14,6 @@ if [[ -z $RELEASE_VERSION ]]; then
     exit 1
 fi
 
-source tools/scripts/release/crates-to-publish.sh
-
 declare -A specified_crate_version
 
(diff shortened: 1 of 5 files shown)


tools/scripts/release/crate-bump.sh
```diff
@@ -26,14 +24,38 @@ for word in ${crate_array[@]}; do
...
+        echo y | cargo release $version --no-push --no-publish --no-tag --no-dev-version --package $name --execute
+    done
+}
+
+bump_crate
+
+# Bump crates that cargo release has modified/updated it's dependencies in `cargo.toml`.
+#
+# Get crates whose cargo.toml file has been updated omitting recently updated crates.
+bump_crate
```
metaclips, line 61: Line 56 bumps crates which has been updated since last tag and then we crates whose ockam dependencies are updated.

thomcc (approved):
LGTM

PR #2542, opened 2022-03-01
ignore `RUSTSEC-2020-0159` advisory

### LINKED ISSUE #2540: ci: Remove cargo audit ignore for `RUSTSEC-2022-0008` after `> btleplug 0.9.1` ships
Upstream pull request: 

* https://github.com/deviceplug/btleplug/pull/254

### LINKED ISSUE #2541: ci: Remove cargo audit ignore for `RUSTSEC-2020-0159`
The [RUSTSEC-2020-0159](https://rustsec.org/advisories/RUSTSEC-2020-0159.html) advisory identifies a vulnerability in the [`chrono`](https://crates.io/crates/chrono) crate due to a call to `localtime_r`.

Time to resolution is indefinite.

More information:

* https://github.com/chronotope/chrono/issues/499

metaclips:
- Ignore `RUSTSEC-2020-0159` advisory till fixed #2541 
- Ignore `RUSTSEC-2022-0008` advisory till fixed #2540
- Cargo deny now checks for all features in CI

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -20,3 +20,6 @@ allow = [
 unmaintained = "deny"
 vulnerability = "deny"
 yanked = "warn"
+ignore = [
+    "RUSTSEC-2020-0159"
+]
--- a/deny.toml
+++ b/deny.toml
@@ -21,5 +21,6 @@ unmaintained = "deny"
 vulnerability = "deny"
 yanked = "warn"
 ignore = [
-    "RUSTSEC-2020-0159"
+    "RUSTSEC-2020-0159",
+    "RUSTSEC-2022-0008"
 ]
--- a/implementations/rust/build.gradle
+++ b/implementations/rust/build.gradle
@@ -37,7 +37,7 @@ task lint_cargo_clippy {
 task lint_cargo_deny {
   doLast {
     exec { commandLine cargo('install', '--locked', '--version', '0.11.0', 'cargo-deny') }
-    exec { commandLine cargo('deny', '--manifest-path=../../Cargo.toml', 'check', 'licenses', 'advisories') }
+    exec { commandLine cargo('deny', '--all-features', '--manifest-path=../../Cargo.toml', 'check', 'licenses', 'advisories') }
   }
 }
 


antoinevg (approved):
Looks good.

thomcc:
Which crates use these? I recall this vuln does result in a segfault (and possible use-after-free) in impacted code, so I'm worried about just ignoring it.

metaclips:
> Which crates use these? I recall the chrono vuln does result in a segfault (and possible use-after-free) in impacted code, so I'm worried about just ignoring it.

`ockam_transport_ble` imports the `chrono` crates. We missed it recently as `cargo deny` was only checking default features. This has been updated.

antoinevg:
> 

I don't think we have much choice here. :-/ 

The `chrono` issue has been open for two years now and is only going to be fixed when (if?) `localtime_r` is fixed.

For the `windows` crate I've already submitted an upstream PR but they move slowly.

I've opened two issues locally to track these:

* https://github.com/ockam-network/ockam/issues/2541
* https://github.com/ockam-network/ockam/issues/2540

thomcc:
Can we move away from using `chrono`? There's tons of stuff in it completely broken under multithreaded use (https://github.com/chronotope/chrono/blob/3467172c31188006147585f6ed3727629d642fed/src/sys/unix.rs#L56-L78 for example).

I don't know of a great replacement, the new versions of the `time` crate seem to be improved, but not as fully featured.

antoinevg:
> Can we move away from using `chrono`?

Hrmmm… so it's an indirect dependency via:

`chrono -> rtcc -> stm32f4xx-hal -> ockam_transport_ble`

Which, funnily enough, means it's not affected by the advisory in this instance as it's running in a non-posix bare-metal environment.

So I think we're safe here.

thomcc:
Hm, yeah, it'd be nice to move on from it to avoid any risk of adding code that is vulnerable... But I suppose for now it's fine. It's worth keeping an eye on that we don't add new dependencies on chrono (or new direct usages) which are unsafe.

thomcc (approved):
Alright, seems fine then.

PR #2544, opened 2022-03-02
fix: remove tokio and ockam_node imports on ockam_macros::test.

adrianbenavides:
Now the macro's user is responsible for importing the appropriate modules that include the timeout and start_node methods.

From ockam_node, start_node is a local method and it can be used straight away. When using another internal crate from this repository, the user may import these methods from ockam or ockam_node. This also would allow us to use a non-tokio implementation for the timeout method.

Related issue: https://github.com/ockam-network/ockam/issues/2500


adrianbenavides:
I didn't have the gpg properly configured on my new PC so I've recreated this PR here with the signed commit: https://github.com/ockam-network/ockam/pull/2545

PR #2545, opened 2022-03-02
feat: add "crate" attribute to ockam_macros::test

adrianbenavides:
~~Now the macro's user is responsible for importing the appropriate modules that include the `timeout` and `start_node` methods.~~

~~From `ockam_node`, `start_node` is a local method and it can be used straight away. From another internal crate of this repository, the user may import these methods from `ockam` or `ockam_node`. This also would allow us to use a non-tokio implementation for the `timeout` method.~~

As discussed below, this PR adds a new attribute to the `ockam_macros::test` macro to let the user define from what crate the `timeout` and `start_node` functions should be imported.

The default value for this attribute is `ockam_node`.

Example of usage: `#[ockam_macros::test(crate = "ockam")]`

Related issue: https://github.com/ockam-network/ockam/issues/2500


implementations/rust/ockam/ockam_macros/src/node_test_attribute/parser.rs
```diff
@@ -50,9 +49,8 @@ fn output_node_test(
     input.block = parse2(quote_spanned! {last_stmt_end_span=>
         {
             use core::time::Duration;
-            use ockam_node::tokio::time::timeout;
 
-            let (mut #ctx_ident, mut executor) = ockam_node::start_node();
```
adrianbenavides, line 55: This is the main change to the macro.

implementations/rust/ockam/ockam_node/src/tests.rs
```diff
@@ -371,22 +371,20 @@ impl Worker for WaitForWorker {
...
+    info!("Done waiting :)");
 
-            let t2 = tokio::time::Instant::now();
-            assert!((t2 - t1) > Duration::from_secs(1));
+    let t2 = tokio::time::Instant::now();
+    assert!((t2 - t1) > Duration::from_secs(1));
 
-            ctx.stop().await.unwrap();
-        })
-        .unwrap();
```
adrianbenavides, line 391: Here we are now using the `test` macro from the `ockam_node` crate.

implementations/rust/ockam/ockam_macros/src/node_test_attribute/parser.rs
thomcc: Hm, so now it has to be explicitly imported? I don't think this is a great experience, to be honest, I suspect the first time use of this macro would always include a hard to understand error.

Another option here would be to accept a `#[ockam::node(crate = "ockam_node")]` or `#[ockam::node(crate = "ockam")]`[^1], or `#[ockam::node(crate = "crate")]`, similar to https://serde.rs/container-attrs.html#crate. If not provided, I think it could default to `ockam_node`.

So, I think I'd prefer this, if possible.

[^1]: I think we'd have to move the declaration of `start_node` to somewhere else for this to actually work, admittedly.

adrianbenavides: I have some changes ready but there are some details I wanted to clarify first:

1. The default crate shouldn't be `ockam` instead of `ockam_node`? If the entry point of the public API is the `ockam` crate, then every time a user wanted to write a test would have to set the `crate` attribute in the macro. Most of our internal tests will have to be modified to set the `crate` attribute though.
2. I've been researching about not using a string in the attribute and I'm pretty much sure there is no other way to do it. Do you have any examples of this in mind? The thing is that an attribute is a `MetaNameValue` and the value is a `Lit` which apparently can only be a string, an integer or a boolean (see https://docs.rs/syn/latest/syn/struct.MetaNameValue.html)

adrianbenavides: As discussed internally:

1. We want to default to `ockam_node`
2. We will consider modifying how the `crate` attribute is parsed using the [`MetaList`](https://docs.rs/syn/latest/syn/struct.MetaList.html) type, which would allow us to write the value as a path instead of as a string (e.g. `#[ockam::test(crate = path::to::crate)]`

thomcc (dismissed):
Seems great! Thanks!

thomcc:
Ah, this has a merge commit, which we don't allow. Looks like we aren't catching that in CI the way we hoped. I'll DM you in slack to help work that out, but we should figure out why its not detected by our CI checks.

thomcc (approved):

thomcc:
Ah, one last thing, can you change the commit message to `feat(rust):` rather than `chore:`? This is actually a new feature, after all.

thomcc (approved):

PR #2546, opened 2022-03-03
WIP: GA Review for `ockam_core` crate

antoinevg:
### Checklist

* [x] 🤔 Should this file exist?
* [x] ❤️ Is the Public API friendly? 
* [x] 📚 Are all Public functions documented? 
* [x] 🪂 Do public functions have rust doc examples? 
* [x] 🆗 Do all functions return the correct errors?
* [x] 🛃 Are all error results returned inside the implementation being handled correctly?
* [x] ✅ Do all modules have sufficient test coverage?
* [x] 🪵 Is there sufficient logging and are the correct levels being used?
* [x] 🛌 Are all modules ready for use on embedded? 
* [x] 📦 Are we using the correct Cargo features in all files and `Cargo.toml`?
* [x] ☣️ Has all `unsafe` code been audited? 

### Remaining Issues

* A: ❌ Random number generation for embedded needs to be defined.
    - #2555 
* A: ❌ `ProtocolId` has been supplanted by the new metadata message types `ockam::OckamMessage`, K: ❌ (move src/message.rs to src/message/mod.rs)
* K: ❌ (we still need a separate trait for "processors" and "workers" but we should keep them both in the worker tree and rename things to be less confusing)
* A: ❌ Route::recipient() -> For API consistency `Address` should probably return a `Result<&Address>` rather than a `panic!` or `clone()`
 
### Refactored

* 🚮 Remove unused type: `ockam_core::ResultMessage`
* ♻️ Refactor names in: `ockam_core::access_control`

    | From | To |
    | :---- | :- |
    | `Passthrough`  | `AllowAll` |
    | `NoAccess` | `DenyAll` |
    | `AccessControl::msg_is_authorized` | `AccessControl::is_authorized` |

* ♻️ Refactor function argument names in: `ockam_core::routing::address`

    | From | To |
    | :---- | :- |
    | `tt` | `transport` |
    | `inner`  | `data` |
    | `addr` | `address` |

* ♻️ Refactor function names in: `ockam_core::vault::key_id_vault`

    | From | To |
    | :---- | :- |
    | `get_secret_by_key_id` | `secret_by_key_id` |

* ♻️ Move `traits` module into its own file.

* ♻️ Move `ockam_core::println_no_std` to `ockam_core::compat::println`


thomcc (approved):
As a cursory review, this looks fine.

I'll do a more thorough review later, there are definite changes I'd like to make, but they're largely unrelated to the thing's you've changed (for example, some functions return &String, and Address::tt shouldn't be public IMO, and should be a newtype wrapper around u8 rather than a u8 directly, etc).

That said, for all I know, those might be in @spacekookie's ockam_core workover, so I'll probably just do a review pass closer at the end.

PR #2549, opened 2022-03-08
script further fix

metaclips:
This fixes the issue below
crateA -> crateB -> crateC -> crateD
Where -> means "is a dependency of", we need to still bump crates whose cyclic dependencies are updated. 
To do this, we `recursively check for crates that are modified`, `bump their versions` and then `store bumped crates names to a state`. On the next iteration we check for modified crates and `only bump crates whose version has not been bumped yet`, recursively checking and bumping till `old state is same as new state` (there are no newly bumped cyclic-deps whose version hasn't been bumped yet).

We can also specify versions that cyclic dependent crates will be bumped to by defining var `BUMPED_DEP_CRATES_VERSION` if not defined all cyclic dependent crate that their source code were not modified will be bumped as a `minor` version.

### DIFF
--- a/tools/scripts/release/README.md
+++ b/tools/scripts/release/README.md
@@ -37,6 +37,12 @@ MODIFIED_RELEASE="signature_core:patch ockam_entity:major" RELEASE_VERSION=relea
 ```
 only signature_core and ockam_entity crates are bumped.
 
+Crates whose transitive dependencies were `only` bumped can be version-bumped with a specified version using the `BUMPED_DEP_CRATES_VERSION` definition so as to follow a different release version
+```bash
+BUMPED_DEP_CRATES_VERSION=patch RELEASE_VERSION=minor tools/scripts/release/crate-bump.sh
+```
+If `BUMPED_DEP_CRATES_VERSION` is not defined then transitive dependent crates are bumped as `minor`.
+
 
 ## Crate Publish
 
--- a/tools/scripts/release/crate-bump.sh
+++ b/tools/scripts/release/crate-bump.sh
@@ -14,6 +14,11 @@ if [[ -z $RELEASE_VERSION ]]; then
     exit 1
 fi
 
+if [[ -z $BUMPED_DEP_CRATES_VERSION ]]; then
+    echo "Version of bumped transitive dependencies set to minor"
+    BUMPED_DEP_CRATES_VERSION="minor"
+fi
+
 declare -A specified_crate_version
 
 crate_array=($MODIFIED_RELEASE)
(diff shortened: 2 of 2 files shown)


tools/scripts/release/crate-bump.sh
```diff
@@ -14,6 +14,11 @@ if [[ -z $RELEASE_VERSION ]]; then
     exit 1
 fi
 
+if [[ -z $BUMPED_DEP_CRATES_VERSION ]]; then
+    echo "Version of bumped circular dependencies set to minor"
```
thomcc, line 18:
```suggestion
    echo "Version of bumped transitive dependencies set to minor"
```

tools/scripts/release/README.md
```diff
@@ -37,6 +37,12 @@ MODIFIED_RELEASE="signature_core:patch ockam_entity:major" RELEASE_VERSION=relea
 ```
 only signature_core and ockam_entity crates are bumped.
 
+Crates whose circular dependencies were `only` bumped can be version-bumped with a specified version using the `BUMPED_DEP_CRATES_VERSION` definition so as to follow a different release version
```
thomcc, line 40:
```suggestion
Crates whose transitive dependencies were `only` bumped can be version-bumped with a specified version using the `BUMPED_DEP_CRATES_VERSION` definition so as to follow a different release version
```

tools/scripts/release/crate-bump.sh
```diff
@@ -26,14 +31,49 @@ done
...
 declare -A bumped_crates
 
-bump_crate() {
-    source tools/scripts/release/crates-to-publish.sh
+recently_updated_crates=""
 
-    echo "Bumping crates with updated dependency. Note crates whose version has been updated recently will be omitted"
-    echo "$updated_crates"
+source tools/scripts/release/crates-to-publish.sh
+bumping_circular_deps=false
```
thomcc, line 37:
```suggestion
bumping_transitive_deps=false
```

tools/scripts/release/crate-bump.sh
```diff
@@ -26,14 +31,49 @@ done
...
+# been bumped for a release.
+while [[ $updated_crates != $recently_updated_crates ]]; do
     for crate in ${updated_crates[@]}; do
+        if [[ ! -z "${bumped_crates[$crate]}" ]]; then
+            echo "===> $crate has been bumped recently ignoring"
+            continue
+        fi
+
         version=$RELEASE_VERSION
+        if [[ $bumping_circular_deps == true ]]; then
```
thomcc, line 72:
```suggestion
        if [[ $bumping_transitive_deps == true ]]; then
```

tools/scripts/release/crate-bump.sh
```diff
@@ -26,14 +31,49 @@ done
...
+            version=$BUMPED_DEP_CRATES_VERSION
+            echo "Bumping circular dependent crate $crate version to $version"
```
thomcc, line 74:
```suggestion
            echo "Bumping transitive dependent crate $crate version to $version"
```

tools/scripts/release/crate-bump.sh
```diff
@@ -42,21 +82,17 @@ bump_crate() {
...
         bumped_crates[$crate]=true
 
         echo "Bumping $crate crate"
         echo y | cargo release $version --no-push --no-publish --no-tag --no-dev-version --package $name --execute
     done
-}
 
-bump_crate
+    recently_updated_crates=$updated_crates
+    bumping_circular_deps=true
```
thomcc, line 92:
```suggestion
    bumping_transitive_deps=true
```

tools/scripts/release/README.md
```diff
@@ -37,6 +37,12 @@ MODIFIED_RELEASE="signature_core:patch ockam_entity:major" RELEASE_VERSION=relea
...
+```bash
+BUMPED_DEP_CRATES_VERSION=patch RELEASE_VERSION=minor tools/scripts/release/crate-bump.sh
+```
+If `BUMPED_DEP_CRATES_VERSION` is not defined then circular dependent crates are bumped as `minor`.
```
thomcc, line 44:
```suggestion
If `BUMPED_DEP_CRATES_VERSION` is not defined then transitive dependent crates are bumped as `minor`.
```

thomcc (dismissed):
Seems about right, the only comment is that the word you're looking for is "transitive dependency" :).

Circular dependencies are a different thing (they're where crate A depends on crate B, and crate B depends on crate A). It's actually forbidden by cargo for you to have circular crate dependencies.

r=me with fixes applied (you can just do them in one commit, I'm not sure if the github "suggestion" stuff will actually work).

metaclips:
Updated @thomcc. Thanks

PR #2553, opened 2022-03-10
GA review `ockam_macros` - first steps

adrianbenavides:
GA `ockam_macros` tracking issue: https://github.com/ockam-network/ockam/issues/2552

### general

- doc examples for every macro
- remove `extra-traits` feature of `syn` dependency

### `node_test`

- tests: add `main` function to `pass` tests so that we can use positive asserts instead of using `compile_fail` and expect a specific error

### `message`

- tests
- minor refactor to handle errors and remove `unwrap`/`expect`

### `vault_test`

- tests
- minor refactor to handle errors and remove `unwrap`/`expect`
- split into two different modules for `vault_test` and `vault_test_sync` macros

### `async_try_clone`

- doc examples
- minor refactor to handle errors and remove `unwrap`/`expect`


thomcc (approved):
Looks great. I think eventually we should find a way to remove the `ignore` marks from the doctests/examples (I worry they'll end up becoming incorrect), but for now this is fine.

adrianbenavides:
> Looks great. I think eventually we should find a way to remove the `ignore` marks from the doctests/examples (I worry they'll end up becoming incorrect), but for now this is fine.

We could do that if we could add the `ockam` dependency as a dev-dependency in this crate, and I thought this could not be possible due to circular dependencies issues (I remember a discussion about this some months ago). If that's not an issue I could open a new PR with these changes.

PR #2559, opened 2022-03-14
Metaclips/git cliff fix

metaclips:
This is a temporary fix for changelog generation with `git-cliff` (update to v0.6.1). To run this script requires `zsh` terminal and script must be called after `crate-bump` script is called.


thomcc (approved):
Hm, so now we bump before changelog? That seems fine.

I'll merge this and file any issues I notice with its use as followups.

PR #2561, opened 2022-03-17
feat(elixir): Implement portals protocol

polvorin:
https://github.com/ockam-network/ockam/issues/1961

## Current Behaviour

Portals are available only on the rust implementation

## Proposed Changes

Implement the [portal protocol](https://github.com/ockam-network/proposals/tree/main/design/0012-portals
) on elixir. 

* Patch bare_extended to support variant types, compatible with how they are encoded on rust side.
* Inlet and Outlet workers
* A simple InletListener that listen for connections and spawns Inlet workers
* Some initial examples, following the pattern of the rust ones

rust and elixir version are interoperable as long as the communication is in plaintext.  But secure channels can't be established, needs more investigation.

Implementation is with simple `Ockam.Workers` processes.   Given the  ping -> pong  handshake,  I suspect the framework at https://github.com/ockam-network/ockam/tree/develop/implementations/elixir/ockam/ockam/lib/ockam/session might be a better approach?.  I didn't take that route so far because I also needed to receive async messages from  `gen_tcp` sockets, and looks like would need to adapt the session' code to support that. 

Feedback is welcome!

## Checks

- [X] All commits in this Pull Request follow the Ockam [commit message convention](https://www.ockam.io/learn/how-to-guides/contributing/CONTRIBUTING#commit-messages).
- [X] I accept the Ockam Community [Code of Conduct](https://www.ockam.io/learn/how-to-guides/high-performance-team/conduct/).
- [X] I have accepted the Ockam [Contributor Licence Agreement](https://www.ockam.io/learn/how-to-guides/contributing/cla/) by adding my Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/ockam-network/contributors/blob/master/CONTRIBUTORS.csv) file in a separate pull request to the [ockam-network/contributors](https://github.com/ockam-network/contributors) repository.

### DIFF
--- a/examples/elixir/tcp_inlet_and_outlet/02-inlet.exs
+++ b/examples/elixir/tcp_inlet_and_outlet/02-inlet.exs
@@ -1,0 +1,15 @@
+["setup.exs"] |> Enum.map(&Code.require_file/1)
+
+# Parse argument.   Usage:  elixir 02-inlet.exs inlet_listen_port
+[lport_s] = System.argv()
+{lport, ""} = Integer.parse(lport_s)
+
+Ockam.Transport.TCP.start()
+
+{:ok, _pid} =
+  Ockam.Transport.Portal.InletListener.start_link(
+    port: lport,
+    peer_route: [Ockam.Transport.TCPAddress.new("localhost", 4000), "outlet"]
+  )
+
+Process.sleep(:infinity)
--- a/implementations/elixir/ockam/ockam/lib/ockam/protocol/bare_extended.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/protocol/bare_extended.ex
@@ -1,12 +1,14 @@
-defmodule Ockam.Bare.Extended do
+defmodule Ockam.Bare.Union do
   @moduledoc """
-  Extension for BARE schema to support simple tags for union types.
+  Extension for BARE schema
 
+  Support simple tags for union types,
   Union type can be defined as [type1: schema, type2: schema]
   and can be encoded and decoded from/to {:type1, data} or {:type2, data}
+
   """
 
-  @type schema() :: any()
+  @type schema() :: :bare.spec()
   @type extended_schema() :: schema() | [{atom(), schema()}]
 
   @type match_error() :: {:error, {:unmatched_subtype, atom(), extended_schema()}}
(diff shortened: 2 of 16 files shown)


hairyhum:
Hi @polvorin,

- We could use the session framework, but as you said it might require some changes. But we currently don't have it used everywhere and might do that in the future, so we can deliver this change separately.
- Rust and Elixir secure channels have diverged because Rust implementation went ahead and I don't think they are compatible right now, so we can keep it out of scope.

implementations/elixir/ockam/ockam/lib/ockam/protocol/bare_extended.ex
```diff
@@ -71,3 +73,79 @@ defmodule Ockam.Bare.Extended do
     extended_schema
   end
 end
+
+
+defmodule Ockam.Bare.Variant do
+  @moduledoc """
+  Support variant types
+  Variant types are defined as {:variant, [atom() | {atom(), schema()]}
```
hairyhum, line 81: Is that how Rust implementation encodes it?
I don't see it in the BARE spec. It looks like a union with `:void` type.
This notation looks more clean, but is not documented in BARE.
We'll have to contribute it to upstream at some point, but for now it's ok.

implementations/elixir/ockam/ockam/lib/ockam/transport/portal/msgs.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Ockam.Transport.Portal.TunnelProtocol do
+    @typedoc """
+    Tunneling protocol messages
+    """
+    alias Ockam.Bare.Extended, as: BareExt
+    @type msg :: :ping | :pong | :disconnect | {:payload, binary()}
+
+    # a bit redundant, likely this can be inferred from the msg() type above via macro?
+    @schema  {:variant, [:ping, :pong, :disconnect, {:payload, :data}]}
```
hairyhum, line 9: Personally I prefer explicit schemas when dealing with cross-implementation protocols.
I like the approach of defining a module per protocol message though, we'll need to do that in other protocols as well.

implementations/elixir/ockam/ockam/lib/ockam/transport/portal/outlet.ex
```diff
@@ -0,0 +1,59 @@
...
+    alias Ockam.Transport.Portal.TunnelProtocol
+
+    require Logger
+
+    @impl true
+    def setup(options, state) do
+       msg = options[:init_message]
+       target_host = options[:target_host]
+       target_port = options[:target_port]
+       Logger.info("Starting outlet worker to #{target_host}:#{target_port}.  peer: #{inspect(msg.return_route)}")
```
hairyhum, line 13: I we should access message fields consistently at least within a single module. We either use `Message.return_route(msg)` here or `msg.payload` next line.
Just so we don't confuse people who read the code later.

implementations/elixir/ockam/ockam/lib/ockam/transport/portal/msgs.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Ockam.Transport.Portal.TunnelProtocol do
```
hairyhum, line 1: Can we keep the module name consistent with the file name? It's easier to navigate this way.

hairyhum (commented):
Looks good. There are some minor notes.

hairyhum:
BTW, you can format the code with `mix format`

examples/elixir/tcp_inlet_and_outlet/02-inlet.exs
```diff
@@ -0,0 +1,10 @@
+["setup.exs"] |> Enum.map(&Code.require_file/1)
+
+# Parse argument.   Usage:  elixir 02-inlet.exs inlet_listen_port
+[lport_s] = System.argv
+{lport, ""} = Integer.parse(lport_s)
+
+{:ok, _pid} = Ockam.Transport.Portal.InletListener.start_link(port: lport, peer_route: [Ockam.Transport.TCPAddress.new("localhost", 4000), "outlet"])
+Ockam.Transport.TCP.start()
```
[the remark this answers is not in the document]
polvorin: You are right!.   It worked when testing just because it took me some time to _connect_  a socket to the inlet side after starting the script,   so by the time the InletListener do receive a connection and attempts to handshake,  the TCP() transport is already up. 

Will fix

implementations/elixir/ockam/ockam/lib/ockam/transport/portal/inlet_listener.ex
```diff
@@ -0,0 +1,49 @@
...
+
+    defp accept(lsocket, peer_route) do
+      {:ok, socket} = :gen_tcp.accept(lsocket)
+      {:ok, worker} = Ockam.Transport.Portal.InletWorker.create([peer_route: peer_route])
+      case Ockam.Node.whereis(worker) do
+        nil ->
+          raise "Worker #{inspect(worker)} not found"
+        pid ->
+          :ok =:gen_tcp.controlling_process(socket, pid)
+          GenServer.cast(pid, {:takeover, socket})
```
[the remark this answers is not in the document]
polvorin: We can pass the socket when creating the InletWorker,  but we would need anyway some form of synchronization to tell the worker it's safe to set it to active mode.  We can't set it to active mode *until* we have transferred the ownership of the socket,  otherwise risking tcp' data to be delivered to the parent (listening) process instead of the worker.
Per [gen_tcp:controlling_process/2 ](https://www.erlang.org/doc/man/gen_tcp.html#controlling_process-2)docs: 

> For instance changing the sockets active mode before the transfer is complete may cause this.

Being realistic,  it's true true that it should be _very_ remote possibility of this to happens, particularly because the worker doesn't set the socket to active until receiving the :pong from the other side.

implementations/elixir/ockam/ockam/lib/ockam/transport/portal/outlet.ex
polvorin: Right, will fix this.  I was confused myself of what's more idiomatic on Elixir,   should I go with the `msg.payload`  way?

implementations/elixir/ockam/ockam/lib/ockam/protocol/bare_extended.ex
polvorin: Right, I didn't found it documented neither.  Learn it by looking at the data that was received from the rust side.

See here:
```
    let ping = PortalMessage::Ping;
    let pong = PortalMessage::Pong;
    let disc = PortalMessage::Disconnect;
    let payload = PortalMessage::Payload(vec![55, 55, 55, 55]);

    println!("Ping: {:?}", serde_bare::to_vec(&ping));
    println!("Pong: {:?}", serde_bare::to_vec(&pong));
    println!("Disc: {:?}", serde_bare::to_vec(&disc));
    println!("Payload: {:?}", serde_bare::to_vec(&payload));

Ping: Ok([0])
Pong: Ok([1])
Disc: Ok([2])
Payload: Ok([3, 4, 55, 55, 55, 55])
```
The implementation is more like a workaround that serve for the current message definitions we have, but it won't handle cases where the variant is not the top-most definition.   Not sure if upstreaming this is an option given that seems to be an addition/deviation of the rust implementation.   Perhaps we can get by with this till https://github.com/ockam-network/ockam/issues/2496 is defined.

examples/elixir/tcp_inlet_and_outlet/02-inlet.exs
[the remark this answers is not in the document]
hairyhum: This should be also addressed in other examples

hairyhum:
@polvorin can you please squash the commits so we can merge it?

polvorin:
> @polvorin can you please squash the commits so we can merge it?

commits squashed now :+1:

hairyhum:
Merged in #2613

PR #2563, opened 2022-03-17
GA review ockam_macros - refactor macros

adrianbenavides:
GA `ockam_macros` tracking issue: https://github.com/ockam-network/ockam/issues/2552

**Note**: once https://github.com/ockam-network/ockam/pull/2553 is merged, I'll rebase this branch and change the target branch to `develop`

## Refactors

- redesign `node`, `test` and `async_try_clone_derive` macros: follow a similar design to that from `serde` macros, which allows us to display all the errors that occurred in the macro expansion instead of just showing the first one. This design also reduces duplicated code and it's easier to understand, maintain and extend.

- document code: since macros can be a pain to debug, I've tried to add comments everywhere I could see so that anyone new to this code can understand it easily.

## Breaking (sort of) changes

- the `node` macro is now more permissive: to avoid frustration when using this macro, now it allows "bad usages" as long as the code still compiles. The only hard check is that the input function is called `main`. Not using the `ockam::Context`, not having an `ockam::Context` as a function argument, having a function without a body... doesn't allow its use.

- the `test` macro is as restrictive as before, nothing changed here. We need it this way so we can stop the test from running indefinitely.

## Remaining issues

* [x] refactor the `async_try_clone_derive` macro to follow a similar structure to that of the `node` macro


thomcc (approved):
This looks fine for now, I think I'll be doing a more thorough pass for most of the crates later anyway.

PR #2564, opened 2022-03-17
Sanjo/vault updates

SanjoDeundiak:


thomcc (approved):
I think we should seriously revisit the locking scheme here, but lgtm

PR #2566, opened 2022-03-18
The big error refactor

spacekookie:
_(original message)_ This PR is a current rebase and fix of #2492, while also applying its changes to how `ockam_node` uses error data.

## Current state

What a massive refactor this is turning out to be but I think we're almost there.

- [x] Replace all instances of `ockam_core::old_error` with `ockam_core::Error2`
- [x] Roughly map existing error codes and structures onto (Origin, Kind) tuple
- [x] Rename `Error2` to `Error`
- [x] Simplify `Error2` API to remove requirement for `ErrorCode` to be public
- [ ] Adjust `Kind` types for crate-local errors (currently I've mostly chosen `Misuse` and `Invalid`. This is not accurate for all wrapped error types and we should have a think about how we classify each category of error to find the most appropriate mapping.  More importantly: do we want to encourage users to handle different kinds of errors differently? Or should we do so internally? i.e. panic on some but not others?)
- [x] Fix `no_std` usage


thomcc:
Didn't get to this today, but will take a look tomorrow. Feel free to message me if there's anything else you need from me tomorrow (stuff to land aside from pipe2, I mean).

spacekookie:
## Progress update

I talked to @thomcc and I think we should push through with the error refactoring on this branch. It'll create too much breakage everywhere else (looking at CI...) if we do this gradually. So let's just pull off the bandaid.

Some thoughts on the new error system and how we should use it below

### Crate-local vs module-local error types

Currently we have crate-local error types for node, ockam, transport, etc. For some of these systems a crate-wide error type makes sens. But for other things they don't. Because the underlying error types are more flexible we can have more granular Error types. Any underlying failure information (for example from a bad serialisation, etc) should then be included in the local error type.

Following is an incomplete list of error types we need to re-introduce into this system:

- [ ] `ockam_core::routing`
- [ ] `ockam_core::messages` (encoding/ schema errors down the line)
- [ ] `ockam_node::context` (these should be _high level_ errors aimed to communicate as much as possible to users about how they misused the API)
- [ ] `ockam_node::internal` (currently all "internal io failure" errors)

... to be expanded


### Updated crates

- [x] `ockam_core`
- [x] `ockam_node`
- [ ] `ockam_transport_core`
- [ ] `ockam_transport_tcp`
- [ ] `ockam_channel`
- [ ] `ockam`

implementations/rust/ockam/ockam_core/src/access_control.rs
```diff
@@ -35,7 +35,7 @@ pub struct AllowAll;
 #[async_trait]
 impl AccessControl for AllowAll {
     async fn is_authorized(&mut self, _local_msg: &LocalMessage) -> Result<bool> {
-        crate::allow()
+        crate::error::allow()
```
antoinevg, line 38: These shouldn't be errors as they denote a state rather than an error condition.

implementations/rust/ockam/ockam_core/src/access_control.rs
```diff
@@ -45,15 +45,17 @@ pub struct DenyAll;
 #[async_trait]
 impl AccessControl for DenyAll {
     async fn is_authorized(&mut self, _local_msg: &LocalMessage) -> Result<bool> {
-        crate::deny()
+        crate::error::deny()
```
antoinevg, line 48: ditto

implementations/rust/ockam/ockam/src/error.rs
```diff
@@ -1,43 +1,68 @@
...
 #![allow(missing_docs)] // Contents are self describing for now.
-#[derive(Clone, Copy, Debug)]
+
+use ockam_core::{
+    errcode::{ErrorCode, Kind, Origin},
+    thiserror, Error2,
+};
+
+#[derive(Clone, Copy, Debug, thiserror::Error)]
 pub enum OckamError {
```
thomcc, line 10: I suspect the thiserror implementations like this would be better as
```rs
impl core::fmt::Display for OckamError {
    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
        core::fmt::Debug::fmt(self, f)
    }
}

impl ockam_core::compat::Error for OckamError {}
```

I believe this will fix the `no_std` issue. Note that std::error::Error is not available to no_std code currently, so we have a shim for it in the compat module. (I know I told you that thiserror is fine, but I think this probably means we can't use it).

On the bright side, this should be a lot less code.

implementations/rust/ockam/ockam_transport_core/src/error.rs
```diff
@@ -58,67 +85,67 @@ impl From<io::Error> for TransportError {
     }
 }
 
-#[cfg(test)]
-mod test {
-    use ockam_core::compat::collections::HashMap;
+// #[cfg(test)]
```
thomcc, line 88: Probably just delete.

thomcc (commented):
This largely looks good to me. Someone else should review ockam_core::error::* though, since I wrote that code.

thomcc:
@spacekookie
1. Ended up having to remove `thiserror` -- doesn't support no_std :(. I also think we'd probably like a macro of our
2. Removed NoneError too. Two reasons:
	- The design already supported unknown cause (hence the cause being an Option).
	- You were using it to indicate missing/unknown error cause, but the display impl said "Option<T> was None" (or something like this), which would be confusing in the future, and was probably a mistake easy for others to make

That said this all revealed some serious flaws in this design! In particular, the no_std error handling becomes *extremely painful*. Use of `Into<Box<dyn Error>>` the way we always do in libstd is... not viable here. I'll need to think more, and maybe ask for some advice.

thomcc:
@spacekookie Alright, I think that's all the CI failures, but you can take anything that's left (and the rebase).

spacekookie:
> @spacekookie
> 
>     1. Ended up having to remove `thiserror` -- doesn't support no_std :(. I also think we'd probably like a macro of our

Well that's disappointing. I thought it supported `no_std` :pensive: Oh well, manuall Display's it is.


>     2. Removed NoneError too. Two reasons:
>        
>        * The design already supported unknown cause (hence the cause being an Option).
>        * You were using it to indicate missing/unknown error cause, but the display impl said "Option was None" (or something like this), which would be confusing in the future, and was probably a mistake easy for others to make

Yea true, I first added the `NoneError` and _then_ added the `_without_cause()` functions which made it irrelevant again. I was kinda on the fence around keeping it in at this point so good riddance :grin: 

Anyway, thisk for pushing it across the finish line. I'll give it another round of review today and then I think we should merge this ASAP before something else needs to be re-based

spacekookie:
Interestingly enough there's a PR that adds `no_std` support but hasn't been touched since 2020 https://github.com/dtolnay/thiserror/pull/64

implementations/rust/ockam/ockam_core/src/error/inner/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+/// The internal error type, which [`super::NewError`] wraps.
+///
+/// When an allocator is available, this is `Box`ed for performance reasons,
+/// (although some of the fields here are a little goofy in low level contexts).
+#[derive(Serialize, Deserialize)]
+pub(super) struct ErrorData {
+    pub(super) code: ErrorCode,
+    // FIXME: `#[cfg()]`ing the fields out changes the serialization format. Not good.
+    #[cfg(feature = "std")]
+    payload: Vec<PayloadEntry>,
```
twittner, line 24: Could `payload` and `source_location` not be made available for `alloc`?

antoinevg: Generally I think we want to keep the Error type as light-weight as possible for embedded targets. Memory is tiiiight and we're already using too much in other areas of the codebase! 😅

implementations/rust/ockam/ockam/src/remote.rs
```diff
@@ -219,12 +219,10 @@ impl Worker for RemoteForwarder {
                 let route = msg.return_route();
 
                 info!("RemoteForwarder registered with route: {}", route);
-                let address;
-                if let Some(a) = route.clone().recipient().to_string().strip_prefix("0#") {
-                    address = a.to_string();
-                } else {
-                    return Err(OckamError::InvalidHubResponse.into());
-                }
+                let address = match route.clone().recipient().to_string().strip_prefix("0#") {
```
SanjoDeundiak, line 222: Recently I found out the most convenient (for me) way to write such things:
```
let address = route
                    .clone()
                    .recipient()
                    .to_string()
                    .strip_prefix("0#")
                    .ok_or(OckamError::InvalidHubResponse)?
                    .to_string();
```
In case anyone likes this as well

implementations/rust/ockam/ockam_core/src/error/code.rs
```diff
@@ -0,0 +1,419 @@
...
+    /// deserialization, have an unknown `Origin` (for now this is too
+    /// error-prone for various reasons).
+    Unknown = 0,
+    /// Reserved for errors emitted by applications using ockam.
+    Application = 1,
+    /// An error originating from the vault.
+    Vault = 2,
+    /// Errors emitted by the transport layer.
+    Transport = 3,
+    /// Errors from some part of the node implementation — the router or relay,
```
SanjoDeundiak, line 123: There is NBSP symbol after the hyphen o_O As well as on line 259

implementations/rust/ockam/ockam_core/src/error/code.rs
```diff
@@ -0,0 +1,419 @@
...
+    ///
+    /// Note: The actual source (file, line, ...) will (hopefully) be available
+    /// on the error itself, as one of the members of the payload.
+    Other = 12,
+    // This is a `#[non_exhaustive]` enum — we're free to add more variants
+    // here. Do not add any which contain payloads (it should stay a "C style
+    // enum"). Payload information should be added to the error itself.
+}
+
+/// Catgory indicates "what went wrong", in abstract terms.
```
SanjoDeundiak, line 151: typo

implementations/rust/ockam/ockam_core/src/error/code.rs
```diff
@@ -0,0 +1,419 @@
+//! Ockam error code enumerations.
```
SanjoDeundiak, line 1: I thought we don't want to introduce coupling. Are we fine with the fact that developers of transport/vault/etc. crates won't be able to introduce their own Kind and Origin values?

implementations/rust/ockam/ockam_core/src/error/code.rs
```diff
@@ -0,0 +1,419 @@
...
+    /// - Et cetera.
+    ///
+    /// But should generally not be used to hold non-identifying metadata, such
+    /// as the date, device IDs, as that information should be stored on the
+    /// payload itself.
+    ///
+    /// Concretely: two `ErrorCode` with different `extra` values should
+    /// identify types of errors.
+    // TODO: is 32 bits okay on embedded? This puts us to 64 bits for the
+    // structure in practice, but means that we'll alwasy be able to hold OS
```
SanjoDeundiak, line 47: typo

implementations/rust/ockam/ockam_core/src/error/code.rs
```diff
@@ -0,0 +1,419 @@
...
+    /// For example, this is appropriate for a component (for example, a custom
+    /// Vault) which do not implement the entire API surface.
+    Unsupported = 3,
+
+    /// Some referenced entity was not found.
+    ///
+    /// For example, this may be appropriate for:
+    ///
+    /// - A vault which does not recognize a `Secret` which it recieves.
+    /// - FFI that recieves an integer `handle` that does not belong to any
```
SanjoDeundiak, line 220: typo

implementations/rust/ockam/ockam_core/src/error/code.rs
```diff
@@ -0,0 +1,419 @@
...
+    ///
+    /// Information about what exactly it is that could not be located should be
+    /// available on the [`Error`](crate::Error) itself.
+    NotFound = 4,
+
+    /// The operation failed because
+    ///
+    /// For example, this may be appropriate for:
+    ///
+    /// - A vault which does not recognize a `Secret` which it recieves.
```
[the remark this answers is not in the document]
SanjoDeundiak: well, I stop checking for them here :D

implementations/rust/ockam/ockam_core/src/error/inner/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+mod trace_config;
+
+type BoxDynErr = Box<dyn ErrorTrait + Send + Sync + 'static>;
+
+/// The internal error type, which [`super::NewError`] wraps.
+///
+/// When an allocator is available, this is `Box`ed for performance reasons,
+/// (although some of the fields here are a little goofy in low level contexts).
+#[derive(Serialize, Deserialize)]
+pub(super) struct ErrorData {
```
SanjoDeundiak, line 20: It looks like for hypothetical no_std vault developer the only usable place to report information about the error is `extra: i32` field, which is probably worse than it was.
std situation is much improved though

implementations/rust/ockam/ockam_core/src/error/inner/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+    // example, or a human description of the cause.
+    Info(String, String),
+}
+
+// Information that is only relevant on this machine. May be logged, so should
+// not contain sensitive data, but is discarded when serializing.
+#[derive(Debug)]
+enum LocalPayloadEntry {
+    #[cfg(all(feature = "std", feature = "backtrace"))]
+    Backtrace(backtrace::Backtrace),
```
SanjoDeundiak, line 153: How is that different from backtrace inside `cause` field?

implementations/rust/ockam/ockam_core/src/error/inner/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+            source_loc: location.into(),
+            local: Some(local),
+            payload,
+        }
+    }
+
+    #[cold]
+    pub fn add_context(&mut self, _key: &str, _val: &dyn core::fmt::Display) {
+        #[cfg(feature = "std")]
+        self.payload.push(PayloadEntry::Info(
```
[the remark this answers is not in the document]
thomcc: Currently, yes. But it would make downcasting impossible, and loose Debug formatting, unlike use of the Error trait. I would like people who implement Worker to be able to get their errors out again on the other side, at least on std platforms.

The only difficulty here is due to a constraint that existed when this code was initially written (round-trip serialization of errors) that no longer exists.

implementations/rust/ockam/ockam_core/src/error/code.rs
thomcc: Yep, I think so.

implementations/rust/ockam/ockam_core/src/error/inner/mod.rs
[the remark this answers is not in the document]
thomcc: Ah, wait, `add_context` yes that's fine, but I have a further rework for this. (But it will happen as a follow-up, since this is already an enormous PR thats too fragile to keep in a branch for much longer).

implementations/rust/ockam/ockam_transport_tcp/tests/send_receive.rs
```diff
@@ -86,10 +86,12 @@ async fn tcp_lifecycle__reconnect__should_not_error(ctx: &mut Context) -> Result
     // TcpSender address should not exist
     let res = child_ctx.send(tx_address.clone(), "TEST".to_string()).await;
     assert!(res.is_err());
-    assert_eq!(
-        res.err().unwrap(),
-        ockam_node::error::Error::UnknownAddress.into()
-    );
+
+    // FIXME!
+    // assert_eq!(
```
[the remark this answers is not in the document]
thomcc: This is the point of error codes, actually, and once downcasting is supported (I describe how it will be done in the issue), you can get the exact error value back.

implementations/rust/ockam/ockam_core/src/error/inner/mod.rs
thomcc: You mean in the dyn Error? https://doc.rust-lang.org/nightly/std/error/trait.Error.html#method.backtrace just the interface by which you can return a backtrace. It is not automatically provided or anything like that, you still have to capture it.

implementations/rust/ockam/ockam_core/src/error/code.rs
spacekookie: What the ... no idea how that got there but good catch :sweat_smile:

implementations/rust/ockam/ockam_core/src/error/inner/mod.rs
thomcc: > One thing it might be quite useful for is to stick a 32-bit memory address in.

It's tricky because this part probably should be serializable. On embedded we could store a `Box<Box<dyn core2::error::Error>>` in the non-error code portion (or [`ThinBox<core2::error::Error>`](https://doc.rust-lang.org/nightly/std/boxed/struct.ThinBox.html)). Or a `NonNull<()>` that you take responsibility for wrangling (it would be leaked though, and wouldn't track lifetimes, so this would be quite unsafe).

thomcc (approved):
There's more to do here but can mostly be done in a follow up. Thanks! I'll be merging this.

implementations/rust/ockam/ockam_core/src/error/inner/mod.rs
antoinevg: `NonNull<()>` sounds about the closest to what I have in mind. The two different use-cases I thought of are:

1. pass a rusty pointer to [`defmt`](https://defmt.ferrous-systems.com/)  
2. pass a raw memory-address to a hardware debugger

What's the most rust-idiomatic solution we can come up with that fits both?

That said, this is not a particularly short-term design decision we need to make, just thinking about keeping escape hatches open that might be useful down the road?

Issue #2256, opened 2021-11-23
Replace TCP port 5000 in examples&guides


SanjoDeundiak:
Latests MacOS release uses port 5000 for some service, therefore guides that use that port won't work

thomcc:
Specifically, this refers to the examples in https://github.com/ockam-network/ockam/tree/develop/examples/rust and the guides in https://github.com/ockam-network/ockam/tree/develop/documentation/guides/rust and https://github.com/ockam-network/ockam/tree/develop/documentation/use-cases.

Note that you can check much of your work by doing something like this from the root of the repository. This both checks that the examples run and the. (That said, I don't know if it exhibits the issue described — I don't think so because I've been running this on up to date macOS without issues... It claims to be up to date, anyway).

```sh
$ export OCKAM_HOME=$PWD
$ tools/docs/check_documentation.sh
$ tools/docs/run_examples.sh
$ tools/docs/run_kafka_example.sh
```

Or, in one command: `export OCKAM_HOME=$PWD && tools/docs/check_documentation.sh && tools/docs/run_examples.sh && tools/docs/run_kafka_example.sh`

(@SanjoDeundiak / @mrinalwadhwa, is that missing anything?)

hseeberger:
There's an alternative: just toggle `System Preferences / Sharing / Airplay Receiver` off. Maybe the docs and examples could mention that?

neil2468:
Hi. I'd like to tackle this (https://github.com/ockam-network/ockam/discussions/1081#discussioncomment-2389817).

I've two questions...
- The port 5000 is hardcoded in some of the guide's images. How can I update those images?
- It seems there are no tests (like run_examples.sh) for the 'secure remote access tunnels' guide. Should I add some?

Thanks!

mrinalwadhwa:
@neil2468 Thank you for looking into this!

> The port 5000 is hardcoded in some of the guide's images. How can I update those images?

The diagrams.drawio file [here](https://github.com/ockam-network/ockam/tree/develop/documentation/use-cases/end-to-end-encrypt-all-application-layer-communication/diagrams
) should have all the images that have port numbers. If you find some images that are in there, let me know.

> It seems there are no tests (like run_examples.sh) for the 'secure remote access tunnels' guide. Should I add some?

That would be great!

adrianbenavides:
Looks like new examples (from late 2022 and 2023) are using port 5000 again. Should we close this for now @mrinalwadhwa ?

[...]

PR #2569, opened 2022-03-21
fix(rust): fix variable name in check_documentation.sh and update use case

neil2468:
While looking at #2256, found a variable was misnamed in tools/docs/check_documentation.sh.

1. Fixed variable in check_documenation.sh that was preventing checks on secure-remote-access-tunnels use case.
2. Updated example code in secure-remote-access-tunnels use case;  check_documentation detected it was out of date.


thomcc (approved):
Looks great!

PR #2577, opened 2022-03-22
Add suborbital demo

thomcc:
We're using this in "production" now, so realistically I think we should land it (with perhaps minor changes if needed), and address the larger issues (which exist) in follow-ups.

In particular, I didn't realize that I can get everything I need from the vault just from the existing public vault API, rather than adding this `software_vault_storage` feature, which was misguided at best. I think this should be addressed in a follow up, though.

## Checks


mattgreg:
Awesome!

implementations/rust/ockam/ockam_transport_websocket/src/error_test.rs
```diff
@@ -17,11 +17,11 @@ fn code_and_domain() {
         let err: ockam_core::Error = ws_err.into();
         match ws_err {
             WebSocketError::Transport(_) => {
-                assert_eq!(err.domain(), TransportError::DOMAIN_NAME);
+                // assert_eq!(err.domain(), TransportError::DOMAIN_NAME);
```
SanjoDeundiak, line 20: Let's fix these

implementations/rust/ockam/ockam_node/src/node.rs
```diff
@@ -30,18 +28,26 @@ pub fn start_node() -> (Context, Executor) {
...
+/// Does nothing if the `no_init_tracing` feature is enabled (for now -- this
+/// should be improved, though).
 fn setup_tracing() {
     #[cfg(feature = "std")]
-    {
-        let filter = EnvFilter::try_from_env("OCKAM_LOG").unwrap_or_else(|_| {
-            EnvFilter::default()
-                .add_directive(LevelFilter::INFO.into())
-                .add_directive("ockam_node=info".parse().unwrap())
+    if !cfg!(feature = "no_init_tracing") {
```
SanjoDeundiak, line 37: Could you please recheck if logging is still working after this change? (It wasn't during your demo)

implementations/rust/ockam/ockam_vault/Cargo.toml
```diff
@@ -79,7 +81,10 @@ rand_pcg = { version = "0.3.1", default-features = false, optional = true }
 sha2 = { version = "0.9", default-features = false }
 x25519-dalek = { version = "1.0", default_features = false }
 cfg-if = "1.0"
-tracing = { version = "0.1", default-features = false }
+tracing = { version = "0.1", default-features = false, features = ["attributes"] }
+# tokio = { version = "1.8", features = ["fs", "io-std", "io-util"], optional = true }
```
SanjoDeundiak, line 85: Remove?

implementations/rust/ockam/ockam_vault/src/secret_impl.rs
```diff
@@ -88,6 +88,7 @@ impl SoftwareVault {
 #[async_trait]
 impl SecretVault for SoftwareVault {
     /// Generate fresh secret. Only Curve25519 and Buffer types are supported
+    #[tracing::instrument(skip_all, err)]
```
SanjoDeundiak, line 91: Is this preventing from sensitive data leakage to tracing surface?

implementations/rust/ockam/ockam_identity/src/channel/secure_channel_worker.rs
```diff
@@ -150,7 +150,7 @@ impl<I: IdentityTrait, T: TrustPolicy> SecureChannelWorker<I, T> {
     pub(crate) async fn create_responder(
         ctx: &Context,
         identity: I,
-        trust_policy: T,
+        trust_policy: Arc<dyn TrustPolicy>,
```
SanjoDeundiak, line 153: This is different approach comparing to what we use everywhere else, why have you made that decision?

implementations/rust/ockam/ockam_command/src/storage.rs
```diff
@@ -0,0 +1,163 @@
...
+                    || format!("failed to update permissions for ockam data at `{path:?}`"),
+                )?;
+            }
+        }
+    }
+    Ok(path)
+}
+
+#[tracing::instrument(level = "debug", skip(data), err, fields(path = ?path))]
+pub fn write(path: &std::path::Path, data: &[u8]) -> anyhow::Result<()> {
```
SanjoDeundiak, line 120: We should consider adding some authenticated encryption to files written to the disk. Having list of trusted identities and secrets in the file in plain text is the most vulnerable place here. Having symmetric key passed as environment variable can strengthen things a lot.

implementations/rust/ockam/ockam_identity/src/channel/trust_policy/any_trust_policy.rs
```diff
@@ -17,14 +17,15 @@ impl<F: TrustPolicy, S: TrustPolicy> AnyTrustPolicy<F, S> {
 
 #[async_trait]
 impl<F: TrustPolicy, S: TrustPolicy> TrustPolicy for AnyTrustPolicy<F, S> {
-    async fn check(&mut self, trust_info: &SecureChannelTrustInfo) -> Result<bool> {
+    async fn check(&self, trust_info: &SecureChannelTrustInfo) -> Result<bool> {
+        // TODO: is the short circuit here a side channel?
```
SanjoDeundiak, line 21: I don't think it is, but const time processing would be great improvement anyways

SanjoDeundiak (changes requested):
Nice job. Left few comments
I think Vault serialization needs rework, if we could avoid merging this part until it's improved, that would be great.

implementations/rust/ockam/ockam_command/src/args.rs
```diff
@@ -0,0 +1,93 @@
...
+    ///   `-vv` => Debug level.
+    ///   `-vvv` => Trace level.
+    #[clap(long, short, parse(from_occurrences))]
+    pub verbose: u8,
+}
+
+#[derive(Clone, Debug, Args)]
+pub struct InletOpts {
+    /// Ockam's cloud node address
+    pub cloud_addr: String,
```
SanjoDeundiak, line 20: Are we using default saved Identity for SecureChannel while creating inlet/outlet? Did you consider using specifying identity_id here? At least optionally.

implementations/rust/ockam/ockam_command/src/storage.rs
thomcc: I agree.

thomcc:
>  if we could avoid merging this part until it's improved, that would be great.

I'm not sure I have time to fix this, and would like to get it in. I'll see what I can do today, though.

implementations/rust/ockam/ockam_transport_websocket/src/error_test.rs
thomcc: Hm, I think I'll just remove the commented out assert. It will be replaced by the better error patch soon anyway, and the way I wrote the current value I'm deliberately avoiding hardcoding it.

implementations/rust/ockam/ockam_vault/src/secret_impl.rs
thomcc: I don't think so, but I don't think it implements debug.

implementations/rust/ockam/ockam_command/src/args.rs
thomcc: I think we probably will want this later though, but for now it's fine.

implementations/rust/ockam/ockam_node/src/node.rs
thomcc: Hm, yes, this seems broken still, I'll figure out why.

implementations/rust/ockam/ockam_identity/src/channel/trust_policy/any_trust_policy.rs
thomcc: Yeah, I *think* it should be fine since sha256(k) doesnt reveal anything about k, which is public anyway. That said, it does seem like its worth doing just in case I'm missing something.

thomcc:
> I'm not sure I have time to fix this, and would like to get it in. I'll see what I can do today, though.

One of the issues here is that we still would need a migration from whatever this format is to the new format. So I think we should just do it in a follow up that takes these into account, even if it's not ideal.

That said I still have to fix the logging (If I don't get it done by the end of the day).

implementations/rust/ockam/ockam_vault/src/secret_impl.rs
SanjoDeundiak: @thomcc I'm just wondering what is the purpose of this here

implementations/rust/ockam/ockam_identity/src/channel/secure_channel_worker.rs
thomcc: Imagine a user needs to support multiple kinds of TrustPolicy at the same time. For example, if the CLI had an arg that was `--trust-everyone` that overrode the file choice.

This way, they can just do something 
```rs
fn get_trust_policy() -> Arc<dyn TrustPolicy> {
    // this wouldn't be how we actually do but you get the idea
    if given_arg("--trust-everyone") {
        Arc::new(TrustEveryonePolicy)
    } else {
        Arc::new(TrustMultiIdentifiersPolicy::new(get_identifiers()))
    }
}
```

The other option would be to force them to write an enum like
```rs
pub enum MyTrustPolicy {
    Everyone(TrustEveryonePolicy),
    List(TrustMultiIdentifiersPolicy),
    // ...
}
```

But this requires much more boilerplate, and is more tedious.

thomcc: I think this case is cut and dry enough to keep this way, TBH. I don't think we need to make a decision like this universally.

implementations/rust/ockam/ockam_vault/src/secret_impl.rs
thomcc: I can remove it if you care, but it was useful when debugging.

SanjoDeundiak: I mean, if we don't have clear reasons to keep it - let's remove it

SanjoDeundiak (dismissed):

PR #2604, opened 2022-03-29
chore(rust): remove the `dyn-clonable` dependency

thomcc:
Seems to break `no_std` and isn't really needed anyway.


spacekookie (approved):
LGTM!

(that is one galaxy-brain macro-syntax though :see_no_evil:)

PR #2606, opened 2022-03-29
Tcp disconnect

SanjoDeundiak:


implementations/rust/ockam/ockam_transport_tcp/src/router/mod.rs
```diff
@@ -51,10 +56,34 @@ impl TcpRouter {
         Ok(())
     }
 
-    async fn connect(&mut self, peer: String) -> Result<Address> {
+    async fn handle_unregister(&mut self, self_addr: Address) -> Result<()> {
+        trace!("TCP unregistration request: {}", &self_addr);
+        // TODO: Could have been much more efficient...
```
thomcc, line 61: I think this is just [`BTreeMap::retain`](https://doc.rust-lang.org/nightly/std/collections/struct.BTreeMap.html#method.retain), so could be `self.map.retain(|(_, v)| v != self_addr)`.

But it seems unlikely for the performance of this function to matter, unregister should be rare.

thomcc (dismissed):
This looks good. One nit that is not required to addres.

Biggest issue is that it has no tests, and is the kind of subtle code that could easily break without noticing. So it would be very good to test. This is fine for now, though.

implementations/rust/ockam/ockam_transport_tcp/src/router/mod.rs
SanjoDeundiak: @thomcc nice, I'll fix that

thomcc (approved):
Oh, thanks for adding tests!

PR #2609, opened 2022-03-30
fix(rust): use serde_bare to prepend length to payload when missing

### LINKED ISSUE #2236: Fix handling of larger messages in ockam_node
Currently, we don't handle messages above a certain size correctly in the rust https://github.com/ockam-network/ockam/blob/bbb2f85bbb561095497b701b50e8e005f301b518/implementations/rust/ockam/ockam_node/src/parser.rs#L17

This doesn't correctly implement the BARE variable length integer scheme described in https://www.ietf.org/archive/id/draft-devault-bare-01.txt:

> Each octet of the encoded value has the most-significant bit set, except for the last octet.  The remaining bits are the integer value in 7-bit groups, least-significant first.

(Note that this means the value we can't be higher than is 127, not 255 as mentioned in the linked source, because of how the encoding works — if it were 128 or larger, the top bit would be set, meaning we would only be supposed to use the bottom 7 from this octet, and check the next one for the rest.)

It's possible this is the cause of some of the errors that we're seeing in [#1624](https://github.com/ockam-network/ockam/issues/1624#issuecomment-972552973) (it seems somewhat surprising to me that we wouldn't even have "in-order for the most part in practice", but I could be mistaken or even misunderstanding what's being described). Either way, we should fix it.

It's possible this can be done by just using serde_bare. Not sure. If not, the varint scheme isn't that complex.

neil2468:
## Current Behaviour

Inside parser::message(), when serde_bare::from_slice() fails to decode a payload we prepend the length of the payload to the payload and try again.

The length is prepended as an unsigned byte (u8). That does not match the BARE spec but works for payloads below 127 bytes.

## Proposed Changes

Use serde_bare to prepend the length as a BARE Uint data type.

Add a test for parser::message() that fails with the old code and passes with the new. 

Fixes #2236 

## Checks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2003,6 +2003,7 @@ dependencies = [
  "ockam_executor",
  "ockam_macros",
  "serde 1.0.136",
+ "serde_bare",
  "tokio",
  "tracing",
  "tracing-subscriber",
--- a/implementations/rust/ockam/ockam_node/Cargo.toml
+++ b/implementations/rust/ockam/ockam_node/Cargo.toml
@@ -60,6 +60,7 @@ tracing-subscriber = { version = "0.3", features = [
 ], optional = true }
 heapless = { version = "0.7", features = ["mpmc_large"], optional = true }
 ockam_executor = { path = "../ockam_executor", version = "^0.18.0", default-features = false, optional = true }
+serde_bare = { version = "0.5.0", default-features = false }
 
 [dev-dependencies]
 serde = { version = "1.0", features = ["derive"] }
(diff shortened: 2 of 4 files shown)


thomcc (dismissed):
Thanks! Seems fine assuming green CI.

thomcc:
Hm, this patch should have changes to the `Cargo.lock` file. Can you push those up too?

neil2468:
> 

Sure.

thomcc (approved):
Looks great, thanks!

PR #2611, opened 2022-03-31
refactor(rust): ws - initialize stream on worker's initialize method

adrianbenavides:
As discussed here https://github.com/ockam-network/ockam/pull/2598, when a new client connects to the router, the websocket connection is now done in the `WebSocketSendWorker::initialize()` method to avoid blocking the router. Now the differences between the `tcp` and `ws` crates are minimum.

The proposed changes will help in the transition to a shared `TransportRouter`.


thomcc (approved):
LGTM

PR #2627, opened 2022-04-05
fix(rust): do not drop control channel sender when processing stopworker message

### LINKED ISSUE #2280: Stoping worker from its functions sometimes drops current branch of execution
In following example you'll see "Point A" printed exactly 100 times, when "Point B" is printed less than 100 times
```rs
use ockam::{Result, Context, route, Address, Worker, Routed};

struct Echoer;

#[ockam::worker]
impl Worker for Echoer {
    type Message = ();
    type Context = Context;

    async fn handle_message(&mut self, ctx: &mut Self::Context, _msg: Routed<Self::Message>) -> Result<()> {
        println!("Point A");
        ctx.stop_worker(ctx.address()).await?;
        println!("Point B");

        Ok(())
    }
}

#[ockam::node]
    for _ in 0..100 {
        let addr = Address::random(0);
        ctx.start_worker(addr.clone(), Echoer).await?;
        ctx.send(route![addr], ()).await?;
    }

    Ok(())
}
```

### LINKED ISSUE #2283: Worker relay ctrl channel is triggered during regular Worker stop

neil2468:
I hope this helps. If it does not, please feel free to ignore it. 

## Current Behaviour

As identified by @SanjoDeundiak in #2283...<code>WorkerRelay</code> uses <code>tokio::select!</code> to concurrently process messages via the mailbox and control channels. When the <code>Router</code> processes a <code>StopWorker</code> message, it drops its two <code>Sender</code>s for the worker. 

This closes the Workers two channels and a race condition results where, inside the <code>tokio::select!</code> branches, calls to <code>Receiver::recv()</code> return <code>None</code>. In issue #2280, sometimes the <code>tokio::select!</code> control channel branch completes first and causes the cancellation of the mailbox branch which is executing <code>Worker::handle_message()</code>. 

https://github.com/ockam-network/ockam/blob/1ebe144fbf3035a4bd68c80738fcbf42a5b8ee93/implementations/rust/ockam/ockam_node/src/relay/worker_relay.rs#L134

https://github.com/ockam-network/ockam/blob/1ebe144fbf3035a4bd68c80738fcbf42a5b8ee93/implementations/rust/ockam/ockam_node/src/router/stop_worker.rs#L27

## Proposed Changes

I think there are two approaches: (1) by re-architecting the <code>WorkerRelay</code>'s <code>tokio::select!</code> code to ignore when the control channel is closed, or (2) by stopping the <code>Router</code>'s early dropping of the control channel <code>Sender</code>.

I've chosen (2) as I think it requires less changes. 

With the changes, when <code>Router</code> processes a <code>StopWorker</code> message it no longer drops a <code>Worker</code>'s <code>AddressRecord</code> struct (which drops the two <code>Sender</code>s) and instead only drops the mailbox <code>Sender</code>. 

In this flow, the <code>Worker</code>'s <code>AddressRecord</code> is fully dropped later, as it has always been, when the <code>Router</code> processes a <code>StopAck</code> message. 

https://github.com/ockam-network/ockam/blob/1ebe144fbf3035a4bd68c80738fcbf42a5b8ee93/implementations/rust/ockam/ockam_node/src/router/mod.rs#L177

Added some test code for this issue. 
And some clippy issues were addressed.

Fixes #2280. 
Fixes #2283.

## Checks


spacekookie (dismissed):
Hey! Thank you for this patch.
I'm glad we're finally fixing this issue. The implementation looks good to me too :+1: 

(one minor alteration, maybe you could re-write the `if let Some(...) { ... } else { ... }` blocks as `match`es?
There's a big refactoring PR (#2566) which does this too and it'll cause less breakage when I rebase it later. But totally optional

neil2468:
> 

Thanks for reviewing and the feedback. 

I've changed the <code>if let Some(...) { ... } else { ... }</code> for <code>match</code>es. Hope that helps!

neil2468:
Sorry. I think I may have done something wrong. It seems this PR is still waiting on a review even though @spacekookie did give approval.

thomcc:
If you rebase I can merge tomorrow, sorry.

spacekookie (approved):

PR #2635, opened 2022-04-06
WIP chore(rust): update no_std example project

antoinevg:
This PR brings the `no_std` example project in `implementations/rust/ockam/ockam_examples/example_projects/no_std` up to date with the latest Ockam changes.

TODO

[x] 🚑 get project working again with latest `develop` branch
[ ] ♻️ refactor dependencies, Cargo features and tracing-subscriber after the `antoinevg/embedded-rust` branch has landed

### DIFF
--- a/implementations/rust/ockam/ockam_examples/example_projects/no_std/Cargo.toml
+++ b/implementations/rust/ockam/ockam_examples/example_projects/no_std/Cargo.toml
@@ -33,22 +33,30 @@ cortexm = [
     "cortex-m",
     "cortex-m-rt",
     "cortex-m-semihosting",
-    "panic-semihosting"
+    "panic-semihosting",
+    "log-semihosting",
+    "tracing",
+    "ockam_core/no_std"
 ]
 
 qemu = ["no_std", "alloc", "cortexm"]
 atsame54 = ["no_std", "alloc", "cortexm", "atsame54_xpro"]
 stm32f4 = ["no_std", "alloc", "cortexm", "stm32f4xx-hal"]
 
+log-itm = []
+log-semihosting = []
+log-uart = []
 
 [dependencies]
 ockam = { path = "../../../ockam", default_features = false, features = ["software_vault"] }
+ockam_core = { path = "../../../ockam_core", default_features = false, optional = true }
 
 alloc-cortex-m = { version = "0.4.1", optional = true }
 cortex-m = { version = "0.7.2", optional = true }
 cortex-m-rt = { version = "0.6.14", optional = true }
 cortex-m-semihosting = { version = "0.3.7", optional = true }
 panic-semihosting = { version = "0.5.6", optional = true }
+tracing = { version = "0.1", default-features = false, optional = true }
 
 atsame54_xpro = { version = "0.2.0", optional = true }
 stm32f4xx-hal = { version = "0.9.0", features = ["rt", "stm32f407"], optional = true }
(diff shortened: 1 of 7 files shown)


implementations/rust/ockam/ockam_examples/example_projects/no_std/Cargo.toml
```diff
@@ -33,22 +33,30 @@ cortexm = [
...
 atsame54 = ["no_std", "alloc", "cortexm", "atsame54_xpro"]
 stm32f4 = ["no_std", "alloc", "cortexm", "stm32f4xx-hal"]
 
+log-itm = []
+log-semihosting = []
+log-uart = []
 
 [dependencies]
 ockam = { path = "../../../ockam", default_features = false, features = ["software_vault"] }
+ockam_core = { path = "../../../ockam_core", default_features = false, optional = true }
```
SanjoDeundiak, line 52: Can we just reexport things you need from ockam_core to ockam? I think we want to avoid situations when you need to import other ockam crates that are already a dependency for `ockam` crate itself. Isn't it?

SanjoDeundiak (changes requested):
Have one question. Otherwise looks great

implementations/rust/ockam/ockam_examples/example_projects/no_std/Cargo.toml
antoinevg: Ah, nice catch - I switched the example over to use `tracing` for output instead of `println` - completely forgot to remove the dependency.

implementations/rust/ockam/ockam_examples/example_projects/no_std/src/tracing_subscriber.rs
```diff
@@ -0,0 +1,235 @@
...
+    let subscriber = EmbeddedSubscriber::new();
+    let dispatch = Dispatch::new(subscriber);
+    tracing::dispatcher::set_global_default(dispatch)
+        .expect("global default dispatcher for tracing is already set");
+    tracing::debug!("tracing_subscriber::register_with_uart initialized");
+}
+
+// - EmbeddedSubscriber -------------------------------------------------------
+
+// TODO add a SubscriberBuilder so we can set log levels etc.
```
thomcc, line 77: Do you want to file an issue for this?

thomcc (approved):

implementations/rust/ockam/ockam_examples/example_projects/no_std/src/tracing_subscriber.rs
thomcc: If it's not worth doing til we need it, that's fine too. But we could possibly get a contributor to take it.

antoinevg: I don't think we need to for this:

1. Any kind of EmbeddedSubscriber™ should never ship as part of an Ockam crate. The subscriber implementation for `tracing` depends on the specific board & other libs any given dev is targeting. So the `TODO` here is just a marker for anyone wanting to copy example code into their own project and wondering why their attempts to set log levels isn't working, it's not relevant to Ockam development.
2. This purpose of this commit is only to do the bare minimum to unblock @metaclips  last week so he can work on figuring out what we need to add testing against `qemu` to CI.
3. This example can be replaced soon anyway with a cleaner one I'm writing to test Embedded RNG stuff.

PR #2640, opened 2022-04-06
Only test macro error output against the stable version of rust (rather than both stable and nightly)

thomcc:
Closes #2639 
Because it includes it.

This has a long comment explaining the motivation.


metaclips (approved):
🚀

PR #2641, opened 2022-04-07
Session restart

SanjoDeundiak:


thomcc (approved):
Many of the changes in `ockam_command` belong somewhere else, since it should be a simple user of the library, and anything it needs to do we should assume others need too. If it is going to be in `ockam_command`, it needs some kinds of cleanup. 

This stuff can happen in a follow, though -- I'll file one and self-assign it for after the error work.

It seems to work for me in local testing, though.

implementations/rust/ockam/ockam_command/src/cmd/inlet.rs
```diff
@@ -1,5 +1,79 @@
...
+        });
+
+        let responder_route = route![channel, "session_responder"];
+
+        Ok(responder_route)
+    }
+
+    async fn stop_session(&mut self, _ctx: &Context) -> Result<()> {
+        if let Some(existing_session) = self.existing_session.take() {
+            // TODO: Stop SecureChannel
```
thomcc, line 70: That said, there are a number of todos still -- the code seems to work but I'm going to wait for you to get back to me on whether or not it's okay to land before they are addressed

SanjoDeundiak: @thomcc it's the only one todo and it's fine to merge

SanjoDeundiak:
@thomcc I think this part can be moved out of cli in the future https://github.com/ockam-network/ockam/pull/2641/commits/57df07b0bd89e49c017d7e6ac3d095435a9297d7
and it's written so that it should be very easy to move it. But I intentionally landed it into cli until we review and prove this approach is the way we want session management to work

thomcc:
> it's written so that it should be very easy to move it.

Yeah, that's a big part of why I don't mind landing it :)

implementations/rust/ockam/ockam_command/src/cmd/inlet.rs
thomcc: K

PR #2645, opened 2022-04-11
 fix(rust): update `compat::sync::Mutex` to return `Result` instead of `Option`

antoinevg:
## Current Behaviour

* The `std` library implementation of `std::sync::Mutex::lock()` returns a `Result<Guard, PoisonError<Guard>>`
* However, the `no_std` implementation in `ockam_core::compat::syncMutex::lock()` returns an `Option<spin::MutexGuard<'_, T>>`.

This, of course, leads to compilation problems when trying to use it in anger.

## Proposed Changes

This patch remedies the situation by modifying the `lock()` function to return a ` Result<spin::MutexGuard<'_, T>, ()>`.

(The reason for the `()` Error type is that the underlying spin-lock implementation will always either return immediately or spin until released. It has no error state. This is a whole other conversation though.)


implementations/rust/ockam/ockam_core/src/compat.rs
```diff
@@ -210,16 +210,16 @@ pub mod sync {
...
     pub struct Mutex<T>(spin::Mutex<T>);
     impl<T> Mutex<T> {
         /// Creates a new mutex in an unlocked state ready for use.
         pub fn new(value: T) -> Self {
             Mutex(spin::Mutex::new(value))
         }
         /// Acquires a mutex, blocking the current thread until it is able to do so.
-        pub fn lock(&self) -> Option<spin::MutexGuard<'_, T>> {
-            Some(self.0.lock())
+        pub fn lock(&self) -> Result<spin::MutexGuard<'_, T>, ()> {
```
twittner, line 221: How about `core::convert::Infallible` instead of `()`?

SanjoDeundiak (approved):
Cool

implementations/rust/ockam/ockam_core/src/compat.rs
antoinevg: Updated, tx! 😊

twittner (approved):

SanjoDeundiak (approved):

thomcc:
It's hard to say if this is the right call -- poisoning (the error returned after you panic while holding a mutex) is a highly contentious feature in std and its fairly likely that a mutex without poisoning will be added to the stdlib soon, and you can already extract the contained value easily. It's also pretty likely that we don't handle it correctly, and our programs would be permanently stuck in a half-working state after poisoning occurs.

So another option would be to have the compat case use a wrapped Mutex in `std` mode too, where we extract the value out of the Err case using [`PoisonError::into_inner`](https://doc.rust-lang.org/nightly/std/sync/struct.PoisonError.html#method.into_inner) -- something like:

```rs
pub use std::sync::MutexGuard;

pub struct Mutex<T>(std::sync::Mutex<T>);

impl<T> Mutex<T> {
    // ...
    pub fn lock(&self) -> MutexGuard<'_, T> {
        self.0.lock().unwrap_or_else(std::sync::PoisonError::into_inner)
    }
    // ...
}
```

Which means neither impl will return `Result` when locking. I think I'd prefer this (more robust, and if our programs don't need poisoning on embedded they probably don't need them with std either), but if there are objections I don't mind the current implementation.

antoinevg:
> Which means neither impl will return `Result` when locking. I think I'd prefer this (more robust, and if our programs don't need poisoning on embedded they probably don't need them with std either), but if there are objections I don't mind the current implementation.

Given that we've all agreed this is a placeholder implementation anyway and given that it is the function signature which is blocking development right now perhaps we can all agree to merge this PR so long today and then revisit the many outstanding issues once we've allocated actual time to work on `no_std` again? 🥺

thomcc (approved):

PR #2646, opened 2022-04-11
ockam_macros: docs revised + crate attribute to async_try_clone macro

adrianbenavides:


implementations/rust/ockam/ockam_macros/README.md
```diff
@@ -19,11 +19,82 @@ Add this to your `Cargo.toml`:
...
+#[ockam::test]
+async fn main(mut ctx: ockam::Context) -> ockam::Result<()> {
+    // ...
+}
+```
+
+To write vault-related tests:
+
+```rust
+use ockam_vault::SoftwareVault;
```
SanjoDeundiak, line 82: SoftwareVault was renamed to Vault

implementations/rust/ockam/ockam_macros/src/lib.rs
```diff
@@ -20,25 +21,33 @@ mod node_attribute;
...
+#[proc_macro_derive(AsyncTryClone, attributes(async_try_clone))]
 pub fn async_try_clone_derive(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
     async_try_clone_derive::expand(input)
         .unwrap_or_else(to_compile_errors)
         .into()
 }
 
-/// Implements Message trait for a type.
+/// Implements the [`Message`](https://docs.rs/ockam_core/latest/ockam_core/trait.Message.html) trait for a type.
```
SanjoDeundiak, line 50: Offtopic: I wonder if Message could/should automatically derive serde::{Serialize, Deserialize} traits?

SanjoDeundiak (approved):
Nice

implementations/rust/ockam/ockam_macros/src/lib.rs
SanjoDeundiak: What I noticed, is when I need to create Message struct, I need to add Serialize, Deserialize traits to this struct derivation list + often I need to add serde as a Cargo dependency, and often it's the only reason I should do that
@adrianbenavides @thomcc thoughts?

adrianbenavides: That would make sense to me, although I don't know if doing that would interfere with the serde derive. I can do some tests and see how it goes.

adrianbenavides: > What I noticed, is when I need to create Message struct, I need to add Serialize, Deserialize traits to this struct derivation list + often this leads to adding serde as a Cargo dependency (dependency that otherwise wouldn't be there) @adrianbenavides @thomcc thoughts?

I would say you can't avoid adding that dependency since you need to have it available in the crate you are using the macro.

SanjoDeundiak: I remember we had some discussion towards reexporting dependencies. Did we decide not to do that? @thomcc

thomcc: Yeah, re-exporting deps isn't good. If we want to auto-impl derive, I think a better way would be a blanket impl.

But doing it separately is more flexible, since it's not rare to need to manually write serialize/deserialize implementations.

PR #2648, opened 2022-04-11
fix(rust): fix clippy warnings.

twittner:
## Current Behaviour

Upgrading rustc to 1.60 produces some warnings from clippy.

## Proposed Changes

Fix clippy warnings introduced with rustc 1.60

## Checks


implementations/rust/ockam/ockam_core/src/access_control.rs
```diff
@@ -26,15 +26,15 @@ use crate::{LocalMessage, Result};
 #[async_trait]
 pub trait AccessControl: Send + Sync + 'static {
     /// Return true if the message is allowed to pass, and false if not.
-    async fn is_authorized(&mut self, local_msg: &LocalMessage) -> Result<bool>;
+    async fn is_authorized(&self, local_msg: &LocalMessage) -> Result<bool>;
```
twittner, line 29: This is debatable of course. Currently no impl requires unique access, but if the design is intended to do so, we may instead want to add `#[allow(clippy::wrong-self-convention)]`.

thomcc (approved):

implementations/rust/ockam/ockam_core/src/access_control.rs
thomcc: I think there used to be the perception in ockam's code that we should prefer `&mut self`, but I'm strongly in favor of not doing so, and places where we don't need it should remove it. Clippy is right here.

Anybody who needs mutability in this trait can hold onto a Mutex or something.

thomcc:
CC @antoinevg this touches ockam_transport_tcp, but your rebase hasn't happened yet (AFAICT), so I'm going to merge this regardless.

thomcc:
I mean, I'll merge it right away after the rebase. So feel free.

(FWIW I usually just squash stuff like that into a single commit so that rebasing doesnt end up needing to fix code that will get changed/moved/removed as part of a later patch in the rebase. I get not wanting to collapse work and loose history though, and there are other options too -- https://git-scm.com/docs/git-rerere sometimes helps as well)

PR #2652, opened 2022-04-12
chore(rust): improve logging of ockam_command's session keep-alive code

thomcc:
Requesed by @mrinalwadhwa -- only logical change is that I noticed we retry immediately on failure to start session. I added a small sleep, so that we don't behave badly in unexpected cases (we don't want to spam logs, saturate the network, etc).


mrinalwadhwa:
Looks great 👍

PR #2690, opened 2022-04-29
chore: update code reviewer rules

mrinalwadhwa:


.github/CODEOWNERS
```diff
@@ -1,14 +1,22 @@
...
-implementations/rust/* @jared-s @mrinalwadhwa
+# Request reviews from elixir-reviewers for any elixir implementation changes.
+*.ex *.exs               @ockam-network/elixir-reviewers
+/implementations/elixir/ @ockam-network/elixir-reviewers
 
-# Request reviews from Daniil and Mrinal when code in the Elixir implementation changes.
-implementations/elixir/* @hairyhum @mrinalwadhwa
+# Request reviews from rust-reviewers for any rust implementation changes.
+*.rs *.toml              @ockam-network/rust-reviewers
+/implementations/elixir/ @ockam-network/rust-reviewers
```
thomcc, line 10: I think this should be `/implementations/rust/`

thomcc:
I don't know if it needs to be formalized (I imagine anything nontrivial here would end up touching the other stuff anyway) but on the C, sh, and cmake parts of https://github.com/ockam-network/ockam/tree/develop/implementations/elixir/ockam/ockam_vault_software I would like to be a reviewer.

That said, I'd like to move[^1] most of this into Rust, and will at some point when I have time. So it might not matter, since that stuff will change anyway.

[^1]: While the previous version of this C code used Rust, I gather it brought in a bunch of heavyweight dependencies, which is not needed.

PR #2700, opened 2022-05-03
update tcp inlet outlet docs

metaclips:
This PR updates the tcp inlet and outlet doc. We also automate check in the `check documentation script`.
We should also automate checking all docs in this PR.


mrinalwadhwa:
We can close this for now.

PR #2706, opened 2022-05-04
WIP: Portals fix

SanjoDeundiak:


implementations/rust/ockam/ockam_transport_tcp/src/portal/portal_receiver.rs
```diff
@@ -62,6 +71,7 @@ impl Processor for TcpPortalRecvProcessor {
 
             // Let Sender forward payload to the other side
             ctx.send(route![self.sender_address.clone()], msg).await?;
+            ctx.sleep(Duration::from_millis(10)).await;
```
thomcc, line 74: Uh, this is pretty strange? What's the deal here?

thomcc (approved):
Unbounded is very concerning here, but we already have many issues of this sort already. It's very bad behavior for a library (or CLI tool) to have, though, since realistically it can exhaust system memory. (Erlang gets away with unbounded channels because crash and restart can be done by the VM in a more granular way than is availble to us)

This also will solve the issue we're seeing.

PR #2709, opened 2022-05-04
Cut release v0.56.0

thomcc:
This fix is intended to address the hangs caused by the TCP buffer size issues.

Ideally this would be a patch release, but this part of the script's functionality doesn't seem to actually work, so it's a minor release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1605,7 +1605,7 @@ dependencies = [
 
 [[package]]
 name = "ockam_command"
-version = "0.15.0"
+version = "0.16.0"
 dependencies = [
  "anyhow",
  "async-recursion",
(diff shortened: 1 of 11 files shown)


mrinalwadhwa (approved):

mrinalwadhwa:
It's weird that this release caused `ockam_transport_ble` to bump because it has a dev dependency on ockam @metaclips @antoinevg.

thomcc:
Yeah that's been an issue for a while. `ockam_transport_ble` should not depend on `ockam`, even as a dev-dep.

I get that that's annoying in practice, though.

antoinevg:
Cleaned up `ockam_transport_ble` dependencies in: #2710

PR #2710, opened 2022-05-05
chore(rust): clean ockam transport ble dependencies

antoinevg:
## Proposed Changes

This pull request removes the following dependencies from the `ockam_transport_ble` crate:

* `ockam`
* `ockam_executor`

### DIFF
--- a/implementations/rust/ockam/ockam_transport_ble/Cargo.toml
+++ b/implementations/rust/ockam/ockam_transport_ble/Cargo.toml
@@ -20,12 +20,11 @@ autoexamples = false
 publish = true
 
 [features]
-default = ["std", "use_btleplug", "ockam/software_vault"]
+default = ["std", "use_btleplug"]
 
 # Feature (enabled by default): "std" enables functionality expected to
 # be available on a standard platform.
 std = [
-    "ockam/std",
     "ockam_core/std",
     "ockam_node/std",
     "ockam_transport_core/std",
@@ -36,7 +35,6 @@ std = [
 # Feature: "no_std" enables functionality required for platforms
 # without the standard library.
 no_std = [
-    "ockam/no_std",
     "ockam_core/no_std",
     "ockam_node/no_std",
     "ockam_transport_core/no_std",
@@ -47,7 +45,6 @@ no_std = [
 # Feature: "alloc" enables support for heap allocation on "no_std"
 # platforms, requires nightly.
 alloc = [
-    "ockam/alloc",
     "ockam_core/alloc",
     "ockam_node/alloc",
     "ockam_transport_core/alloc",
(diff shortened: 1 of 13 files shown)


implementations/rust/ockam/ockam_transport_ble/Cargo.toml
```diff
@@ -128,6 +123,10 @@ cortex-m = "0.7.3"
 [target.'cfg(any(target_arch = "riscv32", target_arch = "riscv64"))'.dependencies]
 riscv = "0.7.0"
 
+[dev-dependencies]
+ockam_identity = { path = "../ockam_identity", version = "^0.44.0", optional = false }
```
SanjoDeundiak, line 127: Isn't false a default value for option flag?

implementations/rust/ockam/ockam_transport_ble/examples/04-routing-over-ble-transport-initiator.rs
```diff
@@ -1,12 +1,21 @@
...
+fn main() -> Result<()> {
+    let (ctx, mut exe) = ockam_node::start_node();
+    exe.execute(async move {
+        async_main(ctx).await.unwrap();
+    })
+    .unwrap();
+    Ok(())
+}
 
-#[ockam::node]
```
SanjoDeundiak, line 8: Can we make it work without ockam dependency? @adrianbenavides 
Should not prevent this PR from landing though

implementations/rust/ockam/ockam_transport_ble/src/driver/bluetooth_hci/mod.rs
```diff
@@ -92,7 +92,7 @@ where
         }
     }
 
-    pub fn reset<T, Time>(&mut self, timer: &mut T, time: Time) -> ockam::Result<()>
+    pub fn reset<T, Time>(&mut self, timer: &mut T, time: Time) -> ockam_core::Result<()>
```
SanjoDeundiak, line 95: Is there a reason why we can't make `use ockam_core::Result`?

SanjoDeundiak (approved):

implementations/rust/ockam/ockam_transport_ble/src/driver/bluetooth_hci/mod.rs
antoinevg: Ah yes, at one point there were a lot more `Result` types scattered through the code before I broke the various bluetooth bits into their own drivers and cleaned things up. 

No point in naming them explicitly anymore now though so I've gone and cleaned this up as well now.

implementations/rust/ockam/ockam_transport_ble/examples/04-routing-over-ble-transport-initiator.rs
antoinevg: That's been dropped here - I just start up the context and executor manually rather than pulling all the macro machinery in:

https://github.com/ockam-network/ockam/blob/615851692f4c61f910a030dc04421a86c56b312d/implementations/rust/ockam/ockam_transport_ble/examples/04-routing-over-ble-transport-initiator.rs#L9-L20

adrianbenavides: Yes, we have two options:
1. `pub use` the node macro at [ockam_core](https://github.com/ockam-network/ockam/blob/develop/implementations/rust/ockam/ockam_core/src/lib.rs#L47) and use the macro as `#[ockam_core::node(crate = "ockam_node")]` or similar
2. explicitly depend on `ockam_macros` and use the macro as `#[ockam_macros::node(crate = "ockam_node")]`

SanjoDeundiak (approved):

thomcc (approved):

PR #2716, opened 2022-05-06
fix(rust): only schedule heartbeat after elapsed time

twittner:
A `TcpSendWorker` currently schedules a new heartbeat for each incoming message. This results in a new context creation every time which consumes a lot of memory because each new context will allocate an mpsc channel. This pressure was alleviated somewhat with commits ac3b8901f4801261dc6b833ac157cfea239e9753 (use 10kb buffer for tcp portal) and 1c6fa8735cb7ceeee73b67a85eb1493782f3a36b (add delay to tcp portal) which resulted in a reduced rate of `TcpSendWorker::handle_message` calls and hence fewer heartbeat schedules.

This commit keeps track of the elapsed time between `handle_message` calls and only attempts to schedule a new heartbeat after the interval duration has been reached.

### DIFF
--- a/implementations/rust/ockam/ockam_transport_tcp/src/workers/sender.rs
+++ b/implementations/rust/ockam/ockam_transport_tcp/src/workers/sender.rs
@@ -6,6 +6,7 @@ use ockam_node::{Context, DelayedEvent};
 use ockam_transport_core::TransportError;
 use serde::{Deserialize, Serialize};
 use std::net::SocketAddr;
+use std::time::Instant;
 use tokio::io::AsyncWriteExt;
 use tokio::net::tcp::{OwnedReadHalf, OwnedWriteHalf};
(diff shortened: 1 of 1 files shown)


SanjoDeundiak:
But it's not called if we don't get messages (e.g. tcp connection is alive but not used)

twittner:
I did not mean to say that idle connections would cause high memory usage, maybe I did not express that well. I observed this with traffic. Here is what my profiler tells me about `develop` (584ca362a0dbe1795a7b7de4240c1f6516d4309b):

problem

SanjoDeundiak:
Well, I don't get what is your suggestion in this PR. We use heartbeats when tcp connection is not actively used to
  1. Keep the connection alive
  1. Detect dropped connections

Your PR makes heartbeats useless, if I read the code right

PR #2718, opened 2022-05-06
feat(elixir): request-response api behaviour, discovery and forwarding apis

hairyhum:
## Current Behaviour

Services all use different message formats but usually implement some sort of
request-response APIs.

## Proposed Changes

Add standartised request-response API to use in services and service clients.
Switch discovery service and static forwarding service to the new API format.

## Checks

- [ ] All commits in this Pull Request follow the Ockam [commit message convention](https://www.ockam.io/learn/how-to-guides/contributing/CONTRIBUTING#commit-messages).
- [ ] I accept the Ockam Community [Code of Conduct](https://www.ockam.io/learn/how-to-guides/high-performance-team/conduct/).
- [ ] I have accepted the Ockam [Contributor Licence Agreement](https://www.ockam.io/learn/how-to-guides/contributing/cla/) by adding my Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/ockam-network/contributors/blob/master/CONTRIBUTORS.csv) file in a separate pull request to the [ockam-network/contributors](https://github.com/ockam-network/contributors) repository.


mrinalwadhwa (dismissed):
This is awesome and looks great!!

implementations/elixir/ockam/ockam/lib/ockam/api/response.ex
```diff
@@ -0,0 +1,85 @@
...
+
+  @max_id 65_534
+
+  @schema {:map, [:id, :request_id, :status, :has_body]}
+
+  def encode(response) when is_map(response) do
+    body =
+      case response.body do
+        binary when is_binary(binary) -> binary
+        ## TODO fail instead?
```
polvorin: agree, +1 to specify   response.body  as `nil | binary()`   and fail otherwise.    Makes it simpler to reason about it  (what can/can't be put as body).   Plus when decoding back with `decode/1` ,  the  body is not decoded. 
 `x ==  decode(encode(x))` is great to maintain

implementations/elixir/ockam/ockam/lib/ockam/api/client.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Ockam.API.Client do
+  @moduledoc """
+  Okcam request-response API client helper
```
mrinalwadhwa: Small typo:

```suggestion
  Ockam request-response API client helper
```

mrinalwadhwa (approved):
Love it!

implementations/elixir/ockam/ockam/lib/ockam/api/response.ex
hairyhum: How do we treat the "" in that case? Do we say that `has_body=true` and it's empty?
@twittner some feedback from the Rust side is also useful from that point. Do we want to have `have_body=true` and empty bodies?

hairyhum: For now let's set the body to `nil` if `has_body=false` and to `""` if `has_body=true` and there is no body.

PR #2725, opened 2022-05-09
fix(rust): share context in scheduled hearbeats

twittner:
As mentioned in #2716, every `DelayedEvent::schedule` creates a new context which is registered with the router. Contrary to what I wrote in #2716 it is not the context creation as such that uses a lot of memory, but the fact that ― AFAICS ― the routing table entries are not cleaned up when the context is no longer used, creating a memory leak. To mitigate this (as a lot of scheduling takes place ), this PR re-uses a single context instead.

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/delayed.rs
+++ b/implementations/rust/ockam/ockam_node/src/delayed.rs
@@ -1,25 +1,26 @@
+use crate::compat::tokio::task::JoinHandle;
 use crate::Context;
 use core::time::Duration;
-use futures::future::{AbortHandle, Abortable};
+use ockam_core::compat::sync::Arc;
 use ockam_core::{Address, Message, Result};
 
 /// Allow to send message to destination address periodically after some delay
 /// Only one scheduled heartbeat allowed at a time
 /// Dropping this handle cancels scheduled heartbeat
-pub struct DelayedEvent<M: Message + Clone> {
-    ctx: Context,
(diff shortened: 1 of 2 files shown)


thomcc (dismissed):
Thanks! We do need to revisit the design here, but this is a good fix for now.

thomcc (dismissed):

twittner:
AFAICT, the CI failure is unrelated to the changes in this PR. I have opened issue https://github.com/build-trust/ockam/issues/2730 to track it.

twittner:
https://github.com/build-trust/ockam/issues/2730 seems to pop up more often on CI with this change. I suspect the change in relative timings (perhaps due to faster TCP worker pair creation) has an impact on the probability.

SanjoDeundiak:
Good change, but let's hold on merging it if it triggers other failure

thomcc (dismissed):

spacekookie:
I've been doing some investigation into this issue and we may not need this fix. I documented my testing here https://miro.com/app/board/uXjVO0PpZLg=/

In short: properly clearing "bare" context addresses (i.e. a context that is created with `ctx.new_context` without a full worker relay behind it) in the router cuts the memory usage from 160M peak to 6.5M :sparkles: 

Fixes available here: https://github.com/build-trust/ockam/pull/2762

twittner:
Closing in favour of #2762.

PR #2726, opened 2022-05-09
tests(rust): fix flaky transport tests

### LINKED ISSUE #2720: Failing test in websocket transport
This test fails occasionally:
https://github.com/build-trust/ockam/runs/6336949886?check_suite_focus=true

adrianbenavides:
Transport tests that were using a random port to create the "listener" sides were failing when the selected port was already used by another process. These changes add a loop to retry binding a transport until it succeeds.

It also adds a small change to the `test` macro so that we don't have to explicitly stop the context at the end of every test.

Fixes https://github.com/build-trust/ockam/issues/2720

### DIFF
--- a/implementations/rust/ockam/ockam_transport_tcp/src/transport.rs
+++ b/implementations/rust/ockam/ockam_transport_tcp/src/transport.rs
@@ -1,7 +1,10 @@
-use crate::{parse_socket_addr, TcpOutletListenWorker, TcpRouter, TcpRouterHandle};
+use std::net::SocketAddr;
+
 use ockam_core::compat::boxed::Box;
 use ockam_core::{Address, AsyncTryClone, Result, Route};
 use ockam_node::Context;
+
+use crate::{parse_socket_addr, TcpOutletListenWorker, TcpRouter, TcpRouterHandle};
 
 /// High level management interface for TCP transports
 ///
@@ -90,6 +93,12 @@ impl TcpTransport {
     }
 
     /// Start listening to incoming connections on an existing transport
+    ///
+    /// Returns the local address that this transport is bound to.
+    ///
+    /// This can be useful, for example, when binding to port 0 to figure out
+    /// which port was actually bound.
+    ///
     /// ```rust
     /// use ockam_transport_tcp::TcpTransport;
     /// # use ockam_node::Context;
@@ -98,10 +107,9 @@ impl TcpTransport {
     /// let tcp = TcpTransport::create(&ctx).await?;
     /// tcp.listen("127.0.0.1:8000").await?;
     /// # Ok(()) }
-    pub async fn listen<S: AsRef<str>>(&self, bind_addr: S) -> Result<()> {
+    pub async fn listen<S: AsRef<str>>(&self, bind_addr: S) -> Result<SocketAddr> {
         let bind_addr = parse_socket_addr(bind_addr.as_ref())?;
-        self.router_handle.bind(bind_addr).await?;
-        Ok(())
+        self.router_handle.bind(bind_addr).await
     }
 }
 
@@ -128,14 +136,9 @@ impl TcpTransport {
         &self,
         bind_addr: impl Into<String>,
         outlet_route: impl Into<Route>,
-    ) -> Result<Address> {
+    ) -> Result<(Address, SocketAddr)> {
         let bind_addr = parse_socket_addr(bind_addr.into())?;
-        let addr = self
-            .router_handle
-            .bind_inlet(outlet_route, bind_addr)
-            .await?;
-
-        Ok(addr)
+        self.router_handle.bind_inlet(outlet_route, bind_addr).await
     }
 
     /// Stop inlet at addr
--- a/implementations/rust/ockam/ockam_transport_tcp/src/transport.rs
+++ b/implementations/rust/ockam/ockam_transport_tcp/src/transport.rs
@@ -1,6 +1,4 @@
-use std::net::SocketAddr;
-
-use ockam_core::compat::boxed::Box;
+use ockam_core::compat::{boxed::Box, net::SocketAddr};
 use ockam_core::{Address, AsyncTryClone, Result, Route};
 use ockam_node::Context;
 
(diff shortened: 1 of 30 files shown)


twittner:
Can we not have the operating system assign us a port number by using 0?

adrianbenavides:
> Can we not have the operating system assign us a port number by using 0?

We should be able to by modifying the `listen` and `create_outlet` methods on the transports crates to return the assigned address. That would definitely be an improvement. Maybe we could leave this for a separate PR? Just to ensure that the tests are stable now without changing any production code.

Thanks for the tip @twittner!

adrianbenavides:
I've added the changes suggested by @twittner to remove the "random port" logic altogether.

SanjoDeundiak:
How do we manage to collide picking random port from 10000 to 65535?
P.S. Can we reduce the scope of this PR to fixing tests only? Would be glad to discuss macro update in separate PR

adrianbenavides:
> How do we manage to collide picking random port from 10000 to 65535?

In my case, I had some os processes running on ports 19455, 42849 and 50586. I don't know if a similar case could happen in CI, but knowing what ports are available is out of our control I guess, so it should be best if we can assign a port automatically by the os. Also if tests were to be run in parallel, chances would definitely be there, although this fix won't solve the problem toralf outlined here: https://github.com/build-trust/ockam/issues/2730.

> P.S. Can we reduce the scope of this PR to fixing tests only? Would be glad to discuss macro update in separate PR

Sure, done!

implementations/rust/ockam/ockam_transport_tcp/tests/send_receive.rs
```diff
@@ -1,33 +1,27 @@
...
-    let bind_address = format!("127.0.0.1:{}", rand_port);
-    let bind_address = bind_address.as_str();
+    let listener_address;
 
     let _listener = {
         let transport = TcpTransport::create(ctx).await?;
-        transport.listen(bind_address).await?;
+        listener_address = transport.listen("127.0.0.1:0").await?;
         ctx.start_worker("echoer", Echoer).await?;
     };
```
twittner: How about removing the `let _listener` block and initialising `listener_address` directly? E.g.

```rust
let transport = TcpTransport::create(ctx).await?;
let listener_address = transport.listen("127.0.0.1:0").await?;
ctx.start_worker("echoer", Echoer).await?;
```

implementations/rust/ockam/ockam_transport_tcp/tests/send_receive.rs
```diff
@@ -55,33 +50,27 @@ impl Worker for Echoer {
...
+        .sample_iter(&rand::distributions::Alphanumeric)
+        .take(256)
+        .map(char::from)
+        .collect();
 
-    let tx_address = transport.connect(bind_address).await?;
+    let tx_address = transport.connect(&listener_address).await?;
 
-    let r = route![(TCP, format!("localhost:{}", rand_port)), "echoer"];
+    let r = route![(TCP, listener_address.clone()), "echoer"];
```
twittner, line 63: `SocketAddr` implements `Copy`, so in theory the clone call is not needed.

twittner (dismissed):
:+1:

spacekookie (dismissed):

implementations/rust/ockam/ockam_transport_tcp/src/transport.rs
```diff
@@ -1,8 +1,11 @@
-use crate::{parse_socket_addr, TcpOutletListenWorker, TcpRouter, TcpRouterHandle};
+use std::net::SocketAddr;
```
SanjoDeundiak: The right import would be 
```
use ockam_core::compat::net::SocketAddr;
```
@antoinevg could you please share your recommendations on imports?

SanjoDeundiak (dismissed):
Looks great, one fix is required though

implementations/rust/ockam/ockam_transport_tcp/tests/send_receive.rs
adrianbenavides: It expects a `String` there, and also a few lines above that (when calling `transport.connect(&listener_address)`). It looks like cloning that string is the less painful option.

implementations/rust/ockam/ockam_transport_tcp/tests/send_receive.rs
adrianbenavides: Absolutely, thanks!

implementations/rust/ockam/ockam_transport_tcp/tests/send_receive.rs
twittner: Ah sorry! You are right of course, I missed that you had it converted to a string again in line 56.

implementations/rust/ockam/ockam_transport_tcp/src/transport.rs
adrianbenavides: I've replaced the `std` imports with the `ockam_core::compat` ones.

twittner (dismissed):

mrinalwadhwa (changes requested):
@adrianbenavides It looks like `commitlint` [ignores](https://github.com/conventional-changelog/commitlint/blob/914782aad70d353baf4d9fbbf9824c0211241484/%40commitlint/is-ignored/src/is-ignored.test.ts#L96) commits of the form:

```
Revert "test(rust): fix flaky transport tests"
```

That was surprising to me. We should add some overriding lints tests for it.

Since in out workflow commits get rebased and moved around, its better to avoid referring to older commits and instead just describe the change that is in a patch/commit.

adrianbenavides:
> @adrianbenavides It looks like `commitlint` [ignores](https://github.com/conventional-changelog/commitlint/blob/914782aad70d353baf4d9fbbf9824c0211241484/%40commitlint/is-ignored/src/is-ignored.test.ts#L96) commits of the form:
> 
> ```
> Revert "test(rust): fix flaky transport tests"
> ```
> 
> That was surprising to me. We should add some overriding lints tests for it.
> 
> Since in out workflow commits get rebased and moved around, its better to avoid referring to older commits and instead just describe the change that is in a patch/commit.

Yeah, I didn't think about changing the auto-generated message after the revert. I've changed it now!

mrinalwadhwa (approved):

PR #2732, opened 2022-05-11
Vault persistence

SanjoDeundiak:


implementations/rust/ockam/ockam_ffi/src/vault.rs
```diff
@@ -2,22 +2,67 @@ use crate::vault_types::{FfiSecretAttributes, SecretKeyHandle};
...
+
+#[derive(Clone, Default)]
+struct VaultEntry {
+    vault: Vault,
+    secrets_mapping: Arc<RwLock<SecretsMapping>>,
+}
+
+impl VaultEntry {
+    async fn insert(&self, key_id: KeyId) -> u64 {
+        self.secrets_mapping.write().await.insert(key_id)
```
twittner, line 52: Why do we require an async `RwLock` here instead of the one from `ockam_core`?

implementations/rust/ockam/ockam_vault/src/vault.rs
```diff
@@ -36,86 +35,62 @@ use ockam_node::compat::asynchronous::RwLock;
...
-}
 
-#[derive(Debug, Eq, PartialEq, Clone)]
-#[cfg_attr(feature = "storage", derive(serde::Serialize, serde::Deserialize))]
-pub(crate) struct VaultEntry {
-    key_id: Option<String>,
-    key_attributes: SecretAttributes,
-    key: SecretKey,
-}
+    pub(crate) async fn preload_from_storage(&self, key_id: &KeyId) -> Result<()> {
```
twittner: This method never returns an error. Could it return unit instead of a `Result`? Could you document the method?

implementations/rust/ockam/ockam_vault/src/vault.rs
```diff
@@ -36,86 +35,62 @@ use ockam_node::compat::asynchronous::RwLock;
...
-impl VaultEntry {
-    pub fn new(key_id: Option<String>, key_attributes: SecretAttributes, key: SecretKey) -> Self {
-        VaultEntry {
-            key_id,
-            key_attributes,
-            key,
-        }
+        let entry = match storage.load(key_id).await {
+            Ok(e) => e,
+            Err(_) => return Ok(()),
```
twittner: Why is the error ignored?

implementations/rust/ockam/ockam_vault/src/vault.rs
```diff
@@ -36,86 +35,62 @@ use ockam_node::compat::asynchronous::RwLock;
...
+        };
+
+        // Possible error due to concurrency, ignore it
+        let _ = self
+            .data
+            .entries
+            .write()
+            .await
+            .insert(key_id.clone(), entry);
```
twittner:
```suggestion
        self.data
            .entries
            .write()
            .await
            .insert(key_id.clone(), entry);
```
Maybe in the doc comment of `preload_from_storage` one could mention that concurrent invocations may overwrite each other's entry?

implementations/rust/ockam/ockam_vault/src/vault.rs
```diff
@@ -36,86 +35,62 @@ use ockam_node::compat::asynchronous::RwLock;
...
 
-#[derive(Debug, Eq, PartialEq, Clone)]
-#[cfg_attr(feature = "storage", derive(serde::Serialize, serde::Deserialize))]
-pub(crate) struct VaultEntry {
-    key_id: Option<String>,
-    key_attributes: SecretAttributes,
-    key: SecretKey,
-}
+    pub(crate) async fn preload_from_storage(&self, key_id: &KeyId) -> Result<()> {
+        if self.data.entries.read().await.contains_key(key_id) {
```
twittner: The storage is mutable and storage entries for a given key may change in between invocations of `preload_from_storage`. Is it acceptable that `preload_from_storage` does not always reflect the most current storage value?

implementations/rust/ockam/ockam_vault/src/storage/file_storage.rs
```diff
@@ -0,0 +1,235 @@
...
+            let vault_bytes = std::fs::read(path).map_err(|_| VaultError::StorageError)?;
+            Self::deserialize(&vault_bytes).await?
+        };
+
+        Ok(())
+    }
+
+    /// Constructor.
+    /// NOTE: Doesn't initialize the storage. Call [`FileStorage::init()`] or use [`FileStorage::create()`]
+    pub fn new(path: &Path) -> Self {
```
twittner: Why offer both `FileStorage::create` and `FileStorage::new` + `FileStorage::init`?

implementations/rust/ockam/ockam_vault/src/storage/file_storage.rs
```diff
@@ -0,0 +1,235 @@
...
+        let mut file = std::fs::OpenOptions::new()
+            .write(true)
+            .read(true)
+            // `create_new` means we error if it exists. This ensures the mode we
+            // provide is respect (the `mode(0o600)` is only used if creating the
+            // file)
+            .create(true)
+            .truncate(true)
+            .mode(0o600) // TODO: not portable, what about windows?
+            .open(path)
```
twittner: Rather than overwriting the file here, how about writing to a temporary file and [renaming](https://doc.rust-lang.org/std/fs/fn.rename.html) it to `path`?

implementations/rust/ockam/ockam_command/src/cmd/identity.rs
```diff
@@ -1,33 +1,31 @@
...
         if !args.overwrite {
             anyhow::bail!(
                 "An identity or vault already exists in {:?}. Pass `--overwrite` to continue anyway",
                 ockam_dir
             );
         }
-        if vault_path.exists() {
-            std::fs::remove_file(&vault_path)
-                .with_context(|| format!("Failed to remove {:?}", vault_path))?;
-        }
```
spacekookie, line 19: Is this directory cleared somewhere else or was this functionality just removed?

spacekookie (commented):
Looks good to me

implementations/rust/ockam/ockam_vault/src/storage/file_storage.rs
SanjoDeundiak: @thomcc could you please help with this? As the original author of the code

implementations/rust/ockam/ockam_vault/src/storage/file_storage.rs
SanjoDeundiak: @twittner you may want to create storage without initializing it. For example if there is some corrupt data (therefore init fails), but you'll be able to call clear() on that Storage

implementations/rust/ockam/ockam_ffi/src/vault.rs
SanjoDeundiak: This doesn't really matter, since I'm going to delete the whole ffi crate very soon (and replace it with tcp side-car), and I wanted to spend minimum time on it just to keep it working. But what are the benefits of having sync RwLock here?

implementations/rust/ockam/ockam_vault/src/vault.rs
SanjoDeundiak: Sure

implementations/rust/ockam/ockam_command/src/cmd/identity.rs
SanjoDeundiak: We now have Vault Storage that is responsible for that

implementations/rust/ockam/ockam_vault/src/vault.rs
SanjoDeundiak: We don't have a way to update entries in a Storage, so proper Storage trait implementation should either return the same entry each time for the given key_id, or return an error if there is no such Entry.
Therefore, it's fine if preload_from_storage was called multiple times concurrently for the same key_id. 1 of the calls succeeds, others fail silently and we should end up with correct state

SanjoDeundiak: Other option could be acquiring write lock at the beginning of this function, and only then call the storage, but that should be less performant I imagine

implementations/rust/ockam/ockam_vault/src/vault.rs
SanjoDeundiak: The strategy is following:
We call Vault function (e.g. sign) with some key_id. We either:
1) Have corresponding key loaded into memory (Vault::data field)
2) Don't have key in the memory, but have it in storage
3) Don't have key at all

So, we always call preload_from_storage before working with the secret and the corresponding behaviour should be following:
1) Do nothing
2) Load key from the storage to the memory
3) Try to load key from the storage, fail silently

The calling function after calling preload_from_storage will try to get secret from the memory, and if it still not there - return an error, because it's 3)

implementations/rust/ockam/ockam_vault/src/vault.rs
SanjoDeundiak: A Storage is meant to be mutable (allow adding and removing values), but values themselves should be immutable.
But I imagine current secret_destroy implementation may have data races, I'll update it

implementations/rust/ockam/ockam_ffi/src/vault.rs
twittner: > This doesn't really matter, since I'm going to delete the whole ffi crate very soon [...]

:ok: 

> But what are the benefits of having sync RwLock here?

Usually async locks are necessary only if the lock guard needs to be used across await points, or if there is very high contention and no viable sharding strategy exists. OTOH regular locks are much faster.

implementations/rust/ockam/ockam_vault/src/vault.rs
twittner: Maybe worth mentioning that in the method's doc comment.

implementations/rust/ockam/ockam_vault/src/vault.rs
twittner: AFAICS only in-memory map entries are replaced and it may be just fine. Reading the comment made me think that this is considered an error but the map does not return an error we need to handle.

implementations/rust/ockam/ockam_vault/src/vault.rs
twittner: What I had in mind is that nothing stops clients from overwriting the value associated to some key ID in storage with another value and subsequent calls to `preload_from_storage` would not reflect this change. Likewise even removed entries will still be accessible.

SanjoDeundiak: Yes, there are dozen of things user can do to mess up the storage. But I'm not sure we should react to that.

twittner (commented):
Other than overwriting the file in-place it looks good to me.

implementations/rust/ockam/ockam_vault/src/storage/file_storage.rs
```diff
@@ -0,0 +1,249 @@
...
+        };
+
+        let _ = std::fs::remove_file(&self.temp_path);
+
+        Ok(())
+    }
+
+    /// Constructor.
+    /// NOTE: Doesn't initialize the storage. Call [`FileStorage::init()`] or use [`FileStorage::create()`]
+    pub fn new(path: &Path, temp_path: &Path) -> Self {
```
twittner, line 140: I would construct the temporary file name automatically, using a random suffix, similar to what `mktemp` does. Then no second parameter is needed and the temporary file is less predicatable.

twittner (approved):

implementations/rust/ockam/ockam_vault/src/storage/file_storage.rs
SanjoDeundiak: In case anything goes sideways, we will be dumping our secrets to random temporary files.

spacekookie (approved):

Issue #2730, opened 2022-05-10
Race condition in TCP router registration


twittner:
In rare circumstances the test `tcp_lifecycle__reconnect__should_not_error` fails with the following error:

```
ERROR ockam_transport_tcp::router::tcp_router: TCP registration request failed, this address is already connected: 1#127.0.0.1:42706
```

This happens, when two concurrent registrations overlap, which in turn can happen, when `TcpRouter::allow_auto_connection` is enabled (the default) and the TCP receiver has read a message and tries to resolve the route before the worker pair created when accepting a connection has managed to register itself for the address. The following log (augmented with additional log messages) shows such a failing test:

```
running 1 test
2022-05-10T12:12:34.414252Z  INFO ockam_node::node: Initializing ockam node
2022-05-10T12:12:34.416223Z DEBUG ockam_transport_tcp::router::tcp_router: Initialising new TcpRouter main=0#1ff21c7310104d72597823e7fc8e206b api=0#ab871583e5d0cda557e4617a182e9485
2022-05-10T12:12:34.416337Z TRACE ockam_transport_tcp::router::tcp_router: creating handle to self addr=0#09366f8a784629c9126acf57762d45ca
2022-05-10T12:12:34.416467Z TRACE ockam_transport_tcp::router::tcp_router: Registering TCP router for type = 1
2022-05-10T12:12:34.416642Z DEBUG ockam_transport_tcp::workers::listener: Binding TcpListener to 127.0.0.1:0
2022-05-10T12:12:34.416713Z TRACE ockam_transport_tcp::workers::listener: starting listen processor addr=0#44e70d50160165b2c4a4bc9187572dfb
2022-05-10T12:12:34.416916Z DEBUG ockam_transport_tcp::workers::listener: Waiting for incoming TCP connection...
2022-05-10T12:12:34.417187Z TRACE ockam_transport_tcp::router::tcp_router: creating handle to self addr=0#dabb18aa7392d3a79f18194266f8efbc
2022-05-10T12:12:34.417274Z TRACE ockam_transport_tcp::router::tcp_router: starting worker pair for connected peer peer=127.0.0.1:34661
2022-05-10T12:12:34.417306Z TRACE ockam_transport_tcp::workers::sender: Creating new TCP worker pair tx_addr=0#bcebce73cc3b20fc45e43185297d0915 internal_addr=0#26998cf66747a8d96ddd93690062a09f
[...]

PR #2733, opened 2022-05-11
fix(rust): fix tcp router race condition

twittner:
As described in issue #2730, when a new TCP connection is accepted and the new worker pair started it may happen that before the subsequent router registration finishes the worker pair may have already read the first message from the socket and when resolving the route, the router may in turn auto-connect to the address and will then make another registration attempt, ultimately causing an error as there can be only one mapping for a given address. See issue #2730 for details.

This commit fixes the issue by deferring the spawning of the worker pair until the registration was completed.

### DIFF
--- a/implementations/rust/ockam/ockam_transport_tcp/src/workers/listener.rs
+++ b/implementations/rust/ockam/ockam_transport_tcp/src/workers/listener.rs
@@ -4,7 +4,7 @@ use ockam_core::{Address, Processor, Result};
 use ockam_node::Context;
 use ockam_transport_core::TransportError;
 use tokio::net::TcpListener;
-use tracing::debug;
+use tracing::{debug, trace};
 
 /// A TCP Listen processor
 ///
@@ -54,12 +54,23 @@ impl Processor for TcpListenProcessor {
         debug!("TCP connection accepted");
 
         let handle_clone = self.router_handle.async_try_clone().await?;
-        // And spawn a connection worker for it
-        let pair = TcpSendWorker::start_pair(ctx, handle_clone, Some(stream), peer, vec![]).await?;
+        // And create a connection worker for it
+        let (worker, pair) =
+            TcpSendWorker::new_pair(ctx, handle_clone, Some(stream), peer, Vec::new()).await?;
 
         // Register the connection with the local TcpRouter
         self.router_handle.register(&pair).await?;
-        debug!("TCP connection registered");
+        debug!(%peer, "TCP connection registered");
+
+        trace! {
+            peer = %peer,
+            tx_addr = %pair.tx_addr(),
+            int_addr = %worker.internal_addr(),
+            "starting tcp connection worker"
+        };
+
+        ctx.start_worker(vec![pair.tx_addr(), worker.internal_addr().clone()], worker)
+            .await?;
 
         Ok(true)
     }
--- a/implementations/rust/ockam/ockam_transport_tcp/src/workers/sender.rs
+++ b/implementations/rust/ockam/ockam_transport_tcp/src/workers/sender.rs
@@ -89,6 +89,37 @@ impl TcpSendWorker {
         }
     }
 
+    pub(crate) fn internal_addr(&self) -> &Address {
+        &self.internal_addr
+    }
+
+    /// Create a `(TcpSendWorker, WorkerPair)` without spawning the worker.
+    pub(crate) async fn new_pair(
+        ctx: &Context,
+        router_handle: TcpRouterHandle,
+        stream: Option<TcpStream>,
+        peer: SocketAddr,
+        hostnames: Vec<String>,
+    ) -> Result<(Self, WorkerPair)> {
+        let tx_addr = Address::random_local();
+        let int_addr = Address::random_local();
+        let sender = TcpSendWorker::new(
+            router_handle,
+            stream,
+            peer,
+            int_addr.clone(),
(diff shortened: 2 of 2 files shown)


thomcc (dismissed):
This looks good to me.

implementations/rust/ockam/ockam_transport_tcp/src/workers/listener.rs
```diff
@@ -54,12 +54,23 @@ impl Processor for TcpListenProcessor {
...
 
         let handle_clone = self.router_handle.async_try_clone().await?;
-        // And spawn a connection worker for it
-        let pair = TcpSendWorker::start_pair(ctx, handle_clone, Some(stream), peer, vec![]).await?;
+        // And create a connection worker for it
+        let (worker, pair) =
+            TcpSendWorker::new_pair(ctx, handle_clone, Some(stream), peer, Vec::new()).await?;
 
         // Register the connection with the local TcpRouter
         self.router_handle.register(&pair).await?;
```
SanjoDeundiak, line 62: Doesn't that change bring a new race condition? We registered the fact that our worker pair accepts messages for certain TCP addresses, but no mailbox has yet been created for them, so if something comes for that TCP address in that moment - it'll get lost. Although, I agree that this race condition may be less likely to happen and less harmful as the one fixed, but would be great to fix both

SanjoDeundiak (changes requested):
Very nice catch, but the problem may be harder to fix...

implementations/rust/ockam/ockam_transport_tcp/src/workers/listener.rs
SanjoDeundiak: Merging this PR and creating issue for a new race condition is an option

twittner: Who could know about this socket address and send a message to it? The socket has just been accepted and the socket address contains an ephemeral port number.

spacekookie: This might be a race condition, but it _should_ be fine. From how I understand the issue, the series of events is as follows:

1. Worker `A` sends a message to a TCP address
2. TCP router is set to "allow_auto_connect" and creates a new pair of workers
3. TCP router registers these workers as handling the respective TCP address
4. Messages routed to this address won't be delivered because the workers haven't been started yet
5. The connection workers are started.

But, because this logic happens in the message accept loop of the router, it won't get to any new message that it has to route, before the connection workers are (hopefully) initialised.

Alternatively we should be using the `context.wait_for` function mechanism instead to check that the connection workers are actually ready to receive and handle connections.

twittner: In the circumstances that @SanjoDeundiak described, a worker could only send a message to the socket address by chance, i.e. by trying random socket addresses as only the TCP router knows about this address at this point. Unless I am missing something this does not seem to be relevant. NB that the address we are looking at is the remote client address, not the remote listener address.

SanjoDeundiak: @twittner yes, you're right. This newly accepted TCP address is in fact unknown and highly unlikely that someone will send a message to it, but it's still possible. 
I think the right approach is to avoid sharing the address, until there is a mailbox for it. Once you break that rule, it's tricky to verify that you don't have race conditions.
My suggestion is let's merge this PR and create an issue for the race condition, since it may take quite an effort to fix

SanjoDeundiak: > This might be a race condition, but it _should_ be fine. From how I understand the issue, the series of events is as follows:
> 
> 1. Worker `A` sends a message to a TCP address
> 2. TCP router is set to "allow_auto_connect" and creates a new pair of workers
> 3. TCP router registers these workers as handling the respective TCP address
> 4. Messages routed to this address won't be delivered because the workers haven't been started yet
> 5. The connection workers are started.
> 
> But, because this logic happens in the message accept loop of the router, it won't get to any new message that it has to route, before the connection workers are (hopefully) initialised.
> 
> Alternatively we should be using the `context.wait_for` function mechanism instead to check that the connection workers are actually ready to receive and handle connections.

@spacekookie 
I think intended order of things is following:
1. Listener accepts the connection 
2. TCP worker pair is started
3. Worker pair is registered for the given TCP address
4. TCP receiver receives TCP traffic and forwards message to the node
5. Some worker replies to that message using return_route

While race condition happens if 3. is postponed due to async execution to be the last instruction, and in this case 5. fails (fails even more dramatically if auto_connect is enabled)

twittner: > This newly accepted TCP address is in fact unknown and highly unlikely that someone will send a message to it, but it's still possible. I think the right approach is to avoid sharing the address, until there is a mailbox for it.

My point is that the address is in fact not shared but only part of the router's mapping.   This is not a race condition but more an issue of access control. Everyone can attempt to send something to any TCP address and potentially disrupt the protocol. Dropping such adversarial messages could even be considered a feature.

SanjoDeundiak: Yes, but safer way is to keep address secret, until there is a mailbox ready receiving messages

SanjoDeundiak (approved):

PR #2734, opened 2022-05-12
WIP: add `enroll`, `space` and `proejct` subcommands

adrianbenavides:
- https://github.com/build-trust/codebase/issues/235

## Structure

The code is structured in 3 layers.

### 1. CLI Command

An entry function `run` receives the arguments the user passed as part of the command.

### 2. Cloud API

Since the cloud API can be accessed through at least two ways (node messaging and HTTP), there is a simple trait with one method to send requests. The arguments of this method define how to build the request and the type of the expected response.

The command creates an instance that implements this trait (now it's hardcoded to the node messaging) and calls the trait's method to send the request, waits for the result, and does something with it.

```rust
pub trait CloudApi {
    async fn send<Params: 'static, Payload: 'static, Response: 'static>(
        &mut self,
        method: RequestMethod,
        params: Params,
        payload: Payload,
    ) -> ockam::Result<Option<Response>>
    where
        Params: Send,
        Payload: Message + serde::Serialize + serde::de::DeserializeOwned,
        Response: serde::de::DeserializeOwned;
}
```
- The `RequestMethod` will be replaced by the `Method` described in the [CBOR schema](https://github.com/build-trust/ockam/pull/2767/files#diff-ab4fbfc6a56e72f5884b2c866992a2ec5f32c8a84a98b69aa10f1e915f16ee23R10).
- The `Params` object can hold information about how to build the request (e.g. API endpoint) but that it's not part of the payload.
- The `Payload` is the message that will be sent by the nodes or as the body of an HTTP request.

### 3. Command-specific services

The `enroll` command accesses a third-party API to retrieve the Auth0 tokens. To facilitate tests, this layer is also modeled using a trait that can be mocked.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -117,6 +117,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "assert_cmd"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "93ae1ddd39efd67689deb1979d80bad3bf7f2b09c6e6117c8d1f2443b5e2f83e"
+dependencies = [
+ "bstr",
+ "doc-comment",
+ "predicates",
+ "predicates-core",
+ "predicates-tree",
+ "wait-timeout",
+]
+
+[[package]]
 name = "async-recursion"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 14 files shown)


adrianbenavides:
Closed in favor of https://github.com/build-trust/ockam/pull/2797

PR #2753, opened 2022-05-15
ci: ignore github verified check is pr sender is on the ockam team

mrinalwadhwa:


mrinalwadhwa:
This doesn't work because team member api need `read:org` scope
https://docs.github.com/en/rest/teams/members

PR #2762, opened 2022-05-17
Ockam node memory usage fixes

spacekookie:
This PR attempts to fix the memory leaks present in the current Inlet/Outlet demo.

Testing data is available here (maybe if you're logged in?): https://miro.com/app/board/uXjVO0PpZLg=/

## Rationale

We create "bare contexts" all over the codebase to send messages, wait for specific replies, etc. These don't have a worker relay behind them and thus are outside of the regular shutdown mechanism. When transferring a 132MB file across Ockam the Outlet router allocated up to 14 THOUSAND addresses that were never cleared.

This change:

1. adds some debug logging so that we can have a better understanding of how many addresses are registered on a system.  In the future we may also want to extend this metrics reporting to numbers of clusters, channel load, etc
2. side-steps the normal shutdown mechanism for bare context records, which don't have a relay and can't interact with the "new" shutdown ACK mechanism.

This seems to fix the memory leak present in the demo.

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/context.rs
+++ b/implementations/rust/ockam/ockam_node/src/context.rs
@@ -1,4 +1,4 @@
-use crate::block_future;
+use crate::async_drop::AsyncDrop;
 use crate::relay::RelayPayload;
 use crate::tokio::{
     self,
@@ -38,25 +38,26 @@ impl AddressType {
     }
 }
 
+type AsyncDropSender = crate::tokio::sync::oneshot::Sender<Address>;
+
 /// Context contains Node state and references to the runtime.
 pub struct Context {
     address: AddressSet,
     sender: Sender<NodeMessage>,
     rt: Arc<Runtime>,
+    async_drop_sender: Option<AsyncDropSender>,
     mailbox: UnboundedReceiver<RelayMessage>,
     access_control: Box<dyn AccessControl>,
 }
 
 impl Drop for Context {
     fn drop(&mut self) {
-        block_future(&self.rt, async {
-            if let Ok(()) = self
-                .stop_address(self.address(), AddressType::Worker, true)
-                .await
-            {
-                debug!("De-allocated bare context {}", self.address());
+        if let Some(sender) = self.async_drop_sender.take() {
+            debug!("De-allocated bare context {}", self.address());
+            if let Err(e) = sender.send(self.address()) {
+                warn!("Encountered error while dropping bare context: {}", e);
             }
-        });
+        }
     }
 }
 
@@ -99,10 +100,14 @@ impl Context {
     ///
     /// This function returns a new instance of Context, the relay
     /// sender pair, and relay control signal receiver.
+    ///
+    /// `async_drop_sender` must be provided when creating a bare
+    /// Context type (i.e. not backed by a worker relay).
     pub(crate) fn new(
         rt: Arc<Runtime>,
         sender: Sender<NodeMessage>,
         address: AddressSet,
+        async_drop_sender: Option<AsyncDropSender>,
         access_control: impl AccessControl,
     ) -> (Self, SenderPair, Receiver<CtrlSignal>) {
         let (mailbox_tx, mailbox) = unbounded_channel();
@@ -113,6 +118,7 @@ impl Context {
                 sender,
                 address,
                 mailbox,
+                async_drop_sender,
                 access_control: Box::new(access_control),
             },
             SenderPair {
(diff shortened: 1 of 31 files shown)


implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -46,6 +47,19 @@ pub struct Context {
     access_control: Box<dyn AccessControl>,
 }
 
+impl Drop for Context {
+    fn drop(&mut self) {
+        block_future(&self.rt, async {
```
SanjoDeundiak: Shouldn't this call be ignored for non-bare Context?

spacekookie:
Okay, after some of the tests didn't like that I used a `block_future` call in the Drop handler I implemented an `AsyncDrop` mechanism, which seems to fix that issue, while also fixing the memory problem.

From the table I linked in the Miro board:

- Transferring a 132M video file at ~850 Kb/s (or KB? I don't know what curl uses as a metric)
- Outlet memory usage peaks at 9.5M. Importantly, sending multiple files doesn't seem to increase this number
- Inlet memory useage peaks at 1.7M (again, doesn't seem to increase with repeated transfers)

implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -37,15 +39,29 @@ impl AddressType {
...
     sender: Sender<NodeMessage>,
     rt: Arc<Runtime>,
+    async_drop_sender: Option<AsyncDropSender>,
     mailbox: UnboundedReceiver<RelayMessage>,
     access_control: Box<dyn AccessControl>,
 }
 
+impl Drop for Context {
+    fn drop(&mut self) {
+        if let Some(sender) = core::mem::replace(&mut self.async_drop_sender, None) {
```
twittner:
```suggestion
        if let Some(sender) = self.async_drop_sender.take() {
```

twittner (dismissed):

implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -1,3 +1,5 @@
+use crate::async_drop::AsyncDrop;
+// use crate::block_future;
```
SanjoDeundiak: Please remove

implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -136,13 +153,24 @@ impl Context {
 
     async fn new_context_impl(&self, addr: Address) -> Result<Context> {
```
SanjoDeundiak: Because lifecycle of bare vs worker Context differs, it would be great to have easily distinguished functions to create each type. AFAIK, for worker, Context::new is used, while Context::new_context_impl is only used for bare variant. But this is very hard to notice without deep investigation + it may change in the future which will lead to bugs.

implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -99,6 +115,7 @@ impl Context {
                 sender,
                 address,
                 mailbox,
+                async_drop_sender: None,
```
SanjoDeundiak: Same comment: not obvious that we only create worker's Context here, so that drop handle is None

implementations/rust/ockam/ockam_node/src/router/stop_worker.rs
```diff
@@ -9,10 +9,12 @@ use ockam_core::{Address, Result};
 pub(super) async fn exec(
     router: &mut Router,
     addr: &Address,
+    bare: bool,
```
SanjoDeundiak: Hm... Any other options for the name? Detached? Autonomous?

SanjoDeundiak (changes requested):
Great idea!

implementations/rust/ockam/ockam_node/src/context.rs
spacekookie: Hmm, yea. I agree that this is confusing. Although having had a look at the way we construct `Context` I didn't see a super obvious alternative.

We could have `Context::new` and `Context::new_bare` but the functions would be almost identical (except for the async_drop field). Maybe it would be enough to pass in a value for `async_drop_sender` to `Context::new` so that it becomes more obvious how the two types of Context are different?

implementations/rust/ockam/ockam_node/src/router/stop_worker.rs
spacekookie: I created a thread on slack to find a better name. I do agree that "bare context" is very ambiguous :sweat_smile:

implementations/rust/ockam/ockam_node/src/context.rs
spacekookie: Oh yea okay. I keep forgetting about `Option::take` :joy:

spacekookie: Oh yea okay. I keep forgetting about `Option::take` :joy:

SanjoDeundiak (approved):

spacekookie:
Woops I had forgotten to add the label. Does the bot still pick up the PR? :thinking:

PR #2764, opened 2022-05-18
Remove messaging cycles. Switch back to bounded channels

SanjoDeundiak:
Depends-On: #2732


implementations/rust/ockam/ockam_node/src/channel_types.rs
```diff
@@ -0,0 +1,31 @@
...
+pub type RouterSender<T> = tokio::sync::mpsc::Sender<T>;
+/// Router receiver
+pub type RouterReceiver<T> = tokio::sync::mpsc::Receiver<T>;
+
+/// Create router channel
+pub fn router_channel<T>() -> (RouterSender<T>, RouterReceiver<T>) {
+    tokio::sync::mpsc::channel(64)
+}
+
+// TODO: Consider replacing with oneshot
```
antoinevg, line 21: I _really_ like Oneshot! 

We're starting to see specialization of channel types so it works to start naming them by function (like e.g. MessageSender, MessageReceiver).

This also relates a little to @spacekookie's recent renaming for detached contexts and your use of a `Mailbox` abstraction in your access control PR. 

Basically, anything we can do that moves us further along the lines of generalizing the many different situations where we're using `mpsc` queues to communicate between ockam subsystems into a shared channel abstraction with common logic is a massive win IMO.

Erlang here we come!

antoinevg (dismissed):
I'm a bit out of the loop on how to test this under load but the code looks sane and I really like the direction you're going with this!

implementations/rust/ockam/ockam_channel/src/secure_channel_decryptor.rs
```diff
@@ -0,0 +1,255 @@
...
+struct DecryptorReadyState {
+    keys: ChannelKeys,
+    encryptor_address: Address,
+}
+
+/// Secure Channel Decryptor
+pub struct SecureChannelDecryptor<V: SecureChannelVault, K: SecureChannelKeyExchanger> {
+    role: Role,
+    key_exchanger: Option<K>,
+    // Optional address to which message is sent after SecureChannel is created
```
spacekookie: I would make this a `///` comment so it shows up in docs

implementations/rust/ockam/ockam_channel/src/secure_channel_encryptor.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    pub(crate) fn convert_nonce_from_u64(nonce: u64) -> ([u8; 8], [u8; 12]) {
+        let mut n: [u8; 12] = [0; 12];
+        let b: [u8; 8] = nonce.to_be_bytes();
+
+        n[4..].copy_from_slice(&b);
+
+        (b, n)
+    }
```
spacekookie, line 33: A short comment explaining what this function does might be useful for the future

implementations/rust/ockam/ockam_channel/src/secure_channel_encryptor.rs
```diff
@@ -0,0 +1,88 @@
...
+
+        let msg = TransportMessage::v1(onward_route, reply, payload.to_vec());
+        let payload = msg.encode()?;
+
+        let payload = {
+            let nonce = self.keys.nonce;
+
+            if nonce == u64::MAX {
+                return Err(SecureChannelError::InvalidNonce.into());
+            }
```
spacekookie, line 57: Is `0` a valid nonce?

implementations/rust/ockam/ockam_identity/src/channel/decryptor.rs
```diff
@@ -425,46 +446,6 @@ impl<I: IdentityTrait> SecureChannelWorker<I> {
         }
     }
```
spacekookie, line 447: Would returning a `Result<&mut State<I>>` work here instead?

spacekookie (commented):
This looks good so far. I left you a few documentation nitpicks but nothing major :+1:

implementations/rust/ockam/ockam_identity/src/channel/decryptor.rs
spacekookie: Oh wow github what the hell :raised_eyebrow: 

The comment is about using `take()` to return the State and having to put it back after using. Feels like `&mut State<I>` could make this harder to misuse

implementations/rust/ockam/ockam_channel/src/secure_channel_encryptor.rs
SanjoDeundiak: Any nonce value is valid, but it's important that unique (key, nonce) pair is used for encryption only once

implementations/rust/ockam/ockam_identity/src/channel/decryptor.rs
SanjoDeundiak: I wish it could work, but that leads to multiple mut references. So long term solution could be:
1. Reorganize what's included in state to avoid multiple mut references
2. Clone state before working on it
3. Handle errors carefully to always put new state after taking old one

Either way it's out of scope of this PR

antoinevg (approved):

implementations/rust/ockam/ockam_identity/src/channel/decryptor.rs
spacekookie: OK, thanks for the clarification!

implementations/rust/ockam/ockam_channel/src/secure_channel_encryptor.rs
spacekookie: Yes, that makes sense. I'm just confused why the explicit check for u64::MAX

mrinalwadhwa:
@SanjoDeundiak @antoinevg with the above change landed, can this be re-enabled?
https://github.com/build-trust/ockam/commit/f3f453f0e815e5600df991bb0ed10c3852fc5170#diff-73e17259d77e5fbef83b2bdbbe4dc40a912f807472287f7f45b77e0cbf78792dL159

SanjoDeundiak:
> @SanjoDeundiak @antoinevg with the above change landed, can this be re-enabled?
> 
> https://github.com/build-trust/ockam/commit/f3f453f0e815e5600df991bb0ed10c3852fc5170#diff-73e17259d77e5fbef83b2bdbbe4dc40a912f807472287f7f45b77e0cbf78792dL159

I think so. Few small fixes may be required

antoinevg:
I'm seeing a couple `no_std` imports that are breaking the build so far… busy looking into it!

antoinevg:
Fixed in: https://github.com/build-trust/ockam/pull/2813

Could you please give it the approve @SanjoDeundiak ?

You should be able to re-enable the `no_std` build for CI again once this has landed @metaclips

antoinevg:
Landed :-)

PR #2767, opened 2022-05-20
feat(rust): add ockam_api

twittner:
Add definitions of core request-response types to be shared with API implementations.

The schema includes optional type tags which are put into the resulting CBOR if the cargo feature "tag" is given. This should help catching type errors across implementations. If the feature is disabled (the default), no tag will be produced any if any are present when decoding CBOR bytes, they are ignored. If not desired I can remove this functionality.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/lib.rs
+++ b/implementations/rust/ockam/ockam_api/src/lib.rs
@@ -1,0 +1,324 @@
+use core::fmt;
+use minicbor::decode::{self, Decoder};
+use minicbor::encode::{self, Encoder, Write};
+use minicbor::{Decode, Encode};
+use ockam_core::compat::borrow::Cow;
+use ockam_core::compat::rand;
+use tinyvec::ArrayVec;
+
+/// CDDL schema or request and response headers as well as errors.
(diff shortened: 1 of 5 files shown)


implementations/rust/ockam/ockam_api/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+#[cbor(map)]
+pub struct Response {
+    /// Nominal type tag.
+    ///
+    /// If the "tag" feature is enabled, the resulting CBOR will contain a
+    /// unique numeric value that identifies this type to help catching type
+    /// errors. Otherwise this tag will not be produced and is ignored during
+    /// decoding if present.
+    #[cfg(feature = "tag")]
+    #[n(0)] tag: TypeTag<9750358>,
```
thomcc, line 71: Are the constants here arbitrary?

thomcc (approved):

implementations/rust/ockam/ockam_api/src/lib.rs
twittner: Yeah, I just generated some randomly.

PR #2768, opened 2022-05-20
feat(rust): add ockam_api_nodes

twittner:
On top of #2767.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1666,8 +1666,11 @@ dependencies = [
  "minicbor",
  "ockam_api",
  "ockam_core",
+ "ockam_macros",
+ "ockam_node",
  "quickcheck",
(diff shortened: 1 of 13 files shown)


mrinalwadhwa:
I'm not sure we should have a separate crate for `ockam_api_nodes`, could that be a module instead?

twittner:
It is possible to fold `ockam_api_nodes` into `ockam_api`. I thought that we might want to share the core type definitions with other crates, so I wanted to keep the number of dependencies minimal, but we can instead also have one api crate that contains all implementations if that is preferred.

twittner:
Includes https://github.com/build-trust/ockam/pull/2778.

mrinalwadhwa (approved):

PR #2778, opened 2022-05-24
feat(rust): add builders to ockam_api

twittner:
To enable a more fluent construction of requests and responses and their optional bodies.


twittner:
Superseded by https://github.com/build-trust/ockam/pull/2768

Issue #2682, opened 2022-04-26
Secure channel compatibility between Elixir and Rust implementations


hairyhum:
Currently Rust implementation of secure channels is ahead of Elixir and supports more functionality.

This makes it currently not possible to establish secure channels between Rust and Elixir nodes.

Elixir secure channels need to be reviewed and made compatible with the current Rust protocols.

hairyhum:
The scope of this issue is to ensure that the base encryption layer of secure channels is compatible, identity layer would be described in #2686

hairyhum:
Implemented in #2789

mrinalwadhwa:
Did we test this? We should have CI check this.

PR #2789, opened 2022-05-25
fix(elixir): secure channel compatibility between rust and elixir

hairyhum:
Rust encodes payloads as bare `data` type (adding size at the begining of the payload)
Fixed in elixir on the worker level, but it's weird considering we're exchanging raw binary data anyway.

Maybe in future Rust can work on `TransportMessage` level instead, because it's already doing double-encoding: https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_channel/src/secure_channel_worker.rs#L176

Rust encodes first request (message1 initiator->responder) as `CreateResponderChannelMessage`, which has an optional `extra_payload` (in newer versions) or `completed_callback_address` (in current version)
Added `InitHandshake` structure to handle that, currently decoding the extra field as `:data`, which might break with the current version.

Nonce size was different for elixir (16bit) and rust (64bit) - changed to 64

Also in the encrypted communication there is no hash (it's empty).

## Checks


hairyhum:
To test with Rust:

Secure channel from Rust to Elixir:
```
use ockam::{route, vault::Vault, Context, Result, TcpTransport, TCP, channel::SecureChannel};

#[ockam::node]
    let _tcp = TcpTransport::create(&ctx).await?;
    let vault = Vault::create();
    // Use "secure_channel" listener on the elixir node
    let r = route![(TCP, "localhost:4000"), "secure_channel"];
    let channel = SecureChannel::create(&ctx, r, &vault).await?;
    let channel_address = channel.address();

    // Send a message to the "echo" worker via the channel.
    ctx.send(route![channel_address, "echo"], "Hello Ockam!".to_string()).await?;

    // Wait to receive a reply and print it.
    let reply = ctx.receive::<String>().await?;
    println!("App Received: {}", reply); // should print "Hello Ockam!"

    // Stop all workers, stop the node, cleanup and return.
    ctx.stop().await
}
```

To test secure channel from Elixir to Rust:

Start Rust node
```
use ockam::{vault::Vault, Context, Result, TcpTransport, TCP, channel::SecureChannel};
use hello_ockam::Echoer;

#[ockam::node]
    let vault = Vault::create();

    // Listen on port 5000
    tcp.listen("127.0.0.1:5000").await?;
    // Create "echo" worker
    ctx.start_worker("echo", Echoer).await?;
    // Create "secure_channel" listener
    SecureChannel::create_listener(&ctx, "secure_channel", &vault).await?;

    // Wait for a stop message (which will never come)
    let _reply = ctx.receive::<String>().await?;

    // Stop all workers, stop the node, cleanup and return.
    ctx.stop().await
}
```
On Elixir node:
```
{:ok, vault} = Ockam.Vault.Software.init()
{:ok, identity} = Ockam.Vault.secret_generate(vault, type: :curve25519)
alias Ockam.Transport.TCPAddress
r = [TCPAddress.new("localhost", 5000), "secure_channel"]
{:ok, channel} = Ockam.SecureChannel.create(route: r, vault: vault, identity_keypair: identity)

Ockam.Workers.Call.call_on_current_process("HI!", [channel, "echo"])
```
**NOTE:** echoer in rust is not actually returning the same payload, but encodes it as BARE `string`.

hairyhum:
NOTE: This PR updates the prebuilt binary for vault for darwin. Needs to be tested on MacOS x64 and arm64

hairyhum:
Addresses https://github.com/build-trust/ockam/issues/2682

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #2792, opened 2022-05-26
Ockam Node diagnostics module

spacekookie:
Some simple utilities to capture diagnostics information from ockam_node. The UX side of this module is yet to be determined. But I suggest we merge ASAP and then use ockam_command to expose different sets of metrics and diagnostics on the terminal

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/context.rs
+++ b/implementations/rust/ockam/ockam_node/src/context.rs
@@ -53,7 +53,7 @@ pub struct Context {
     async_drop_sender: Option<AsyncDropSender>,
     mailbox: MessageReceiver<RelayMessage>,
     access_control: Box<dyn AccessControl>,
-    metrics: Arc<AtomicUsize>,
+    mailbox_count: Arc<AtomicUsize>,
 }
 
 impl Drop for Context {
@@ -86,7 +86,7 @@ impl Context {
                 trace!("{}: received new message!", self.address());
 
                 // First we update the mailbox fill metrics
-                self.metrics.fetch_sub(1, Ordering::Acquire);
+                self.mailbox_count.fetch_sub(1, Ordering::Acquire);
 
                 msg
             } else {
@@ -130,7 +130,7 @@ impl Context {
                 mailbox,
                 async_drop_sender,
                 access_control: Box::new(access_control),
-                metrics: Arc::new(0.into()),
+                mailbox_count: Arc::new(0.into()),
             },
             SenderPair {
                 msgs: mailbox_tx,
@@ -188,7 +188,7 @@ impl Context {
 
         // Create a "detached relay" and register it with the router
         let (msg, mut rx) =
-            NodeMessage::start_worker(addr.into(), sender, true, Arc::clone(&self.metrics));
+            NodeMessage::start_worker(addr.into(), sender, true, Arc::clone(&self.mailbox_count));
         self.sender
             .send(msg)
             .await
@@ -284,7 +284,7 @@ impl Context {
 
         // Send start request to router
         let (msg, mut rx) =
-            NodeMessage::start_worker(address, sender, false, Arc::clone(&self.metrics));
+            NodeMessage::start_worker(address, sender, false, Arc::clone(&self.mailbox_count));
         self.sender
             .send(msg)
             .await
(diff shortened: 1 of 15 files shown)


mrinalwadhwa:
@spacekookie some tests are failing.

spacekookie:
@mrinalwadhwa some of the CI failures looked unrelated which is a bit confusing (all the node ones that match on a specific error output). At least I don't _think_ anything I did in this PR should affect those :thinking:

mrinalwadhwa:
Note RUSTFLAGs are set in gradle as well https://github.com/build-trust/ockam/blob/develop/implementations/rust/build.gradle

spacekookie:
Ah thanks! I gave up on trying to inherit the rust flags from the environment and am just setting the appropriate `--cfg` in all the right places. Fingers crossed that it's less sad now

implementations/rust/ockam/ockam_node/src/executor.rs
```diff
@@ -71,11 +86,25 @@ impl Executor {
...
+        let alive = Arc::new(AtomicBool::from(true));
+        #[cfg(feature = "metrics")]
+        self.rt
+            .spawn(Arc::clone(&self.metrics).run(Arc::clone(&alive)));
+
+        // Spawn user code second
         let rt = Arc::clone(&self.rt);
         let join_body = rt.spawn(future);
 
+        // Then blocke on the execution of the router
```
SanjoDeundiak: Typo

implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -50,6 +53,7 @@ pub struct Context {
     async_drop_sender: Option<AsyncDropSender>,
     mailbox: MessageReceiver<RelayMessage>,
     access_control: Box<dyn AccessControl>,
+    metrics: Arc<AtomicUsize>,
```
SanjoDeundiak: Can this have more meaningful name please?

implementations/rust/ockam/ockam_node/src/router/mod.rs
```diff
@@ -134,8 +148,9 @@ impl Router {
                 addrs,
                 senders,
                 detached,
+                metrics,
                 ref reply,
-            } => start_worker::exec(self, addrs, senders, detached, reply).await?,
+            } => start_worker::exec(self, addrs, senders, detached, metrics, reply).await?,
```
SanjoDeundiak: Should processors also have these?

implementations/rust/ockam/ockam_node/src/router/start_processor.rs
```diff
@@ -34,6 +34,7 @@ async fn start(
         addr.clone().into(),
         msgs,
         ctrl,
+        Arc::new(0.into()), // don't track metrics
```
SanjoDeundiak: Why not?

implementations/rust/ockam/ockam_node/src/router/utils.rs
```diff
@@ -34,6 +34,7 @@ pub(super) async fn resolve(
     match router.map.internal.get(&primary_address) {
         Some(record) if record.check() => {
             trace!("{} OK", base);
+            record.increment_msg_count();
```
SanjoDeundiak, line 37: I guess technically it's correct for now, but I'm surprised that 1 address resolution == 1 message sent

SanjoDeundiak (changes requested):
Hm, not aware of consequences of enabling tokio_unstable. Have you looked into it?

implementations/rust/ockam/ockam_node/src/router/start_processor.rs
spacekookie: Because currently processors have to meaningful parameters to keep track of (for example incoming message queue)

spacekookie:
From everything I've seen it just enables some unstable features in the API. I don't really understand why this isn't just a crate feature flag but apparently it has to be set via `RUSTFLAGS` (maybe the feature needs to be checked in many different crates or some technical reason like this).

As for the 1 address resolution = 1 message: I agree that it's an assumption that is currently being made but until we change that behaviour I think implementing it like this is fine for now

implementations/rust/ockam/ockam_node/src/router/start_processor.rs
SanjoDeundiak: Processors have mailbox, not sure we use it though

SanjoDeundiak (dismissed):

SanjoDeundiak (dismissed):

mrinalwadhwa (dismissed):

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #2794, opened 2022-05-27
feat(rust): remove `AsRef` from `PublicKey` to avoid confusion

SanjoDeundiak:
Remove public_key.as_ref(), since PublicKey struct also has type information, and it was not obvious that as_ref() returns only binary part


twittner (approved):

spacekookie (approved):
Looks good to me

PR #2798, opened 2022-05-29
refactor(rust): rename dry_run command argument to test_argument_parser

mrinalwadhwa:
renaming dry_run command argument to test_argument_parser, we may need `--dry-run` for api dry runs later.cc: @adrianbenavides


adrianbenavides:
Thanks!

PR #2799, opened 2022-05-30
mrinal/command

mrinalwadhwa:
- refactor(rust): enroll, project and space commands
- feat(rust): improve ockam command help
- feat(rust): improve ockam node command help


implementations/rust/ockam/ockam_command/src/cloud/mod.rs
```diff
@@ -1,41 +1,40 @@
...
+    /// Enroll with Ockam Cloud
+    #[clap(display_order = 900)]
+    Enroll(EnrollCommand),
+
+    /// Create, update and delete projects in Ockam Cloud
+    #[clap(display_order = 900)]
+    Project(ProjectCommand),
+
+    /// Create, update and delete spaces in Ockam Cloud
+    #[clap(display_order = 900)]
```
adrianbenavides, line 28: Shouldn't the `display_order` be incremental? Or does it behave in a certain way if you assign the same order to multiple commands?

adrianbenavides (commented):
Love it! I think it all looks more homogeneous.

implementations/rust/ockam/ockam_command/src/cloud/mod.rs
mrinalwadhwa: Things with the same `display_order` are displayed alphabetically.
The default `display_order` value is 999, so the `help` subcommand is at 999 ... I chose 900 as our default so our subcommands show alphabetically before the `help` command.
https://docs.rs/clap/latest/clap/struct.Arg.html#method.display_order

adrianbenavides: Ohh, good to know, thanks!

PR #2803, opened 2022-05-31
feat(rust): add support for message flow authorization within ockam nodes

### LINKED ISSUE #2474: Worker Message Flow Authorization
### Related Issues

* #27 
* #1852 
* #2439 
* #2467 
* https://github.com/ockam-network/proposals/pull/88

---

# Worker Message Flow Authorization

> *"There are essentially three architectures used to secure computer systems: the out-of-band security architecture, the principal security architecture, and the capability security architecture.* <br/>
>  *These architectures take shape in non-computer security systems as well."*
>
> Kragen Sitaker, The Three Security Architectures [[Sitaker2000]]

## If Worker Message Flow Authorization is the answer, what's the question?

When we talk about the security of a given system what we are really talking about is our ability to enforce the following three constraints upon that system: [[Frantz]]

1. **Actors are secret:** An actor cannot be accessed by other actors.
2. **Actors are immutable:** An actor cannot be modified by other actors.
3. **Actors are trustworthy:** An actor cannot be misused by other actors.

As every new generation of security professional finally comes to accept, any distributed system that interacts with the Public Realm [[Slominski2013]] can only exercise programmatic control over two of these constraints:

### 1.Actors are secret

One way we can prevent access to an actor is by requiring explicit authorization from a 3rd actor before allowing access by another actor. Access Control Lists (ACLs) are an example of using a 3rd actor to withhold access.

Another way is to confine the ability of other actors to access the actor. Sandboxing (aka Capabilities) is an example of confining evaluation.

### 2. Actors are immutable

Again, actors can be protected from modification by using an ACL to withhold write access from unauthorized actors.

Again, we could also confine the ability of other actors to modify the actor using a Sandbox / Capabilities.

### 3. Actors are trustworthy

Where the wheels come off is when we have an actor with legitimate access to another actor.

Given the a-priori trust relationship, there is nothing stopping that actor from mis-using that access in a way which was not intended.

Instead, what we see again and again is how every privilege escalation attack begins with an actor using authorized access or capabilities in ways that were neither intended nor foreseen.

The truth is that even the most carefully configured ACL or the most heavily Sandboxed runtime remains vulnerable to configuration or programming mistakes.

The reason for this is that we need to be correct every time we make a change to the system but any malicious actor only has to wait for us to make a single mistake to be able to take advantage of the trust relationship.

### Then Why Bother At All?

Security is not a binary all or nothing proposition.

Instead, we do better to think of security as the probability that a given attacker with a given set of resources can compromise a given system.

We still lock our front door even if we sometimes forget to close the kitchen window.

The reason this works is that burglars also make mistakes.

Sometimes they give up if the front door is closed. Other times, when they check the kitchen window it's on a day that the window is closed.

On the other hand, if we were to give up and just leave the door and window open all the time we are almost guaranteed to be robbed!

To put it more formally, one way we can measure the vulnerability of a given system is *the product of statistical variation in vulnerability between dependent actors within the system.* [[Goldratt1984]]

This measure has implications for how practical a particular authorization mechanism is for any given realm of trust.

e.g. If you're managing hundreds of thousands of IOT devices on the open Internet you will have many regrets should you choose to deploy an ACL system consisting of a Radius server with LDAP backend!

We will look at this in more detail later on.

## Capabilities

> *"When speaking of object computation, all too much emphasis is often placed on the objects themselves. The fabric of an object system is the dynamic reference graph."*
>
> Alan Kay
>
> *"A capability combines Designation with Authority"*
>
> Norm Hardy

### The Confused Deputy

The [[Confused deputy problem]] is when a legitimate, more privileged actor is tricked by another actor to misuse its authority on the system.

For example:

* Bob is not authorized to access Carol directly.
* Bob has authorization to access Alice directly.
* Bob also has limited authorization to access Carol via Alice.
* Alice has unlimited authorization to access Carol directly.
* Bob asks Alice to ask Carol to do something that only Alice is authorized to do.

Bob's access succeeds because Carol has no way of knowing whether the request came from Bob or from Alice.

The simplest way to solve the confused deputy problem is to combine the reference to an Actor with the authorization to access that Actor.

For example:

* Bob does not have a reference to Carol.
* Bob has a reference to Alice.
* Alice has a reference to Carol.
* Bob asks Alice for a reference to Carol and gets back an BobCarol capability.

Now when Carol receives a request it can tell the difference between a BobCarol capability and an AliceCarol capability and won't act for one on behalf of another.

This is the essence of the definition of a capability:

> *An actors reference to itself combined with a reference to another actor.*

Or reference to the method of another actor if you like, it's actors all the way down either way!

## Patterns of Capability Security

Building a secure distributed capability system we can easily reason about requires us to place a number of constraints upon our Actor Model and Implementation. [[Miller2006]]

### Constraint #1: Only connectivity begets connectivity.

In many Actor Implementations actor references are usually globally available to any other actor.

However, in the original Actor Model [[Hewitt1977]], actors can only know about other actors that have been explicitly provided to them.

This means that the only way to send a message to another actor is if you are already holding a reference to it.

So, from the confused deputy example, there are only four ways Bob could have obtained a reference to Carol:

* Connectivity by Introduction: Alice sends Bob a reference to Carol
* Connectivity by Parenthood: Bob creates Carol, then he has the only reference to Carol
* Connectivity by Endowment: Bob is created by someone with a reference to Carol.
* Connectivity by Initial Conditions: In some initial state of the system Bob and Carol already exist with Bob holding a reference to Carol. (You cannot bootstrap any capability system without this!)

Our ability to enforce this constraint is only possible if we also add the next constraint:

### Constraint #2: Actor references are unforgeable and unique.

Without this constraint we are unable to prevent actors guessing other actor references or capability collisions.

(Realms simplify the implementation of this constraint, see below)

### Constraint #3: Absolute Encapsulation

From outside an actor boundary, there should be no access to actor internals without actor consent. Even if one has a reference to the actor.  Even if multiple actors are executing on the same physical hardware.

### Constraint #4: All Authority Accessed Only by References

[Kahn88, Rees96]

The authority an actor has to affect or be affected by actors outside of itself should be exactly represented by the references it holds.

All primitives for interacting with other actors must be embodied by primitive objects, and one must obtain references to these primitive objects in order to exercise the associated primitive functions.

### Selective Revocation

* Alice gives Carol a fingerprint of every capability reference sent out.
* Alice can now revoke any capability by telling Carol to discard the reference.
* Carol checks any incoming message capabilities against her list of fingerprints and discards all those with no entry.

TODO

### Rights Amplification

Sometimes the authorization from bringing two references together can exceed the sum of authorities provided by each individual.

e.g. Can + Can Opener = Oh Hai Tuna!

TODO

## Building Blocks: Actor Realms

> *"Capabilities are object references that can span machines and persist, while still being unforgeable."* <br/>
>
> Mark Miller, Capability Computation [[MillerCapComp]]

Capability implementations should be efficient.

If an implementation can be made simpler and faster then the implementation should be made simpler and faster!

Implementing our second constraint, "Actor references are unforgeable and unique", is particularly challenging.

Tristan Slominski proposes an elegant optimization strategy that overlaps quite neatly with Ockam's Actor implementation. [[Slominski2013]]

He divides the Actor "address space" into three realms:

### Realm #1: Memory Realm

* Multiple Actor configurations (workers) each execute in a unique (hopefully protected) memory space.
* Since we assume that two actors cannot occupy the same space in memory we can satisfy both the uniqueness and unforgeability constraints.  (CPU side-channel and DRAM exploits non-withstanding…)

### Realm #2: Trusted Realm

* Once the scale of the actor system spans beyond a single machine, the mapping from an actor to a unique memory address is lost. Now, the same memory address on one machine could map to a completely different actor (or nothing at all) on another machine.
* The simplifying assumption of the trusted realm is trust. The machines involved in the actor system are trusted and the communication between them is trusted. (e.g. Ockam Secure Channels)
* This satisfies the constraint of unforgeability because we can also assume that  this is guaranteed by a well behaved execution environment which is trusted to be implemented correctly as in the memory realm.
* This leaves multiple methods that can satisfy the distributed global uniqueness constraint on addresses in a trusted execution environment:
    - SECURE random selection from a large address space (e.g. 128 bit GUID, 160 bit SHA-1 of a random digest - I don't think any of these are secure) We also need some proxy mechanism for mapping between memory realm address and a trusted realm address.
    - Pair a host identifier (where uniqueness is enforced in the trusted realm) with memory address from the memory realm. (I think this is what Ockam does?)
    - Any others?

**tl;dr:** Basically, if we have a method (one presented here or otherwise) to select unique addresses in a Trusted Realm, which encompasses multiple Memory Realms, it is possible to enforce unique and unforgeable actor addresses.

### Realm #3: Public Realm

The Public Realm is the untrusted network and untrusted machines between Trusted Realms.

The known solution to communication over an untrusted medium is point-to-point encryption, in our case the public key variant of point-to-point encryption.

BUSY

## Building Blocks: Actor Channels

 aka Worker Routes

## Building Blocks: Actor Capabilities

aka Worker Method Addresses / References

Also see: cap-std

## Ockam Implementation

It's a nice model you have there.

It would be a shame if someone had to… implement it…

TODO

## Formal Analysis

Applied Pi Calculus [Abadi, Fournet] has a sufficiently rich notation and semantics to express the capability security model in a way that can be easily reasoned about.

e.g.

### Terms

    M, N ::= x               variables
           | n               names
           | f(M1, …, Mk)    function application

### Simple type system for terms

    Integer
    Key
    Channel<Integer>
    Channel<Key>

### Standard functions

    pair    :: x -> y -> (x, y)
    first   :: (x, y) -> x
    second  :: (x, y) -> y
    encrypt :: ? -> ?
    hash    :: ? -> ?
    symenc  :: (x, k) -> y     symmetric-key encryption
    symdec  :: (y, k) -> x     symmetric-key decryption
    pk      :: x -> p          public key from random seed x
    sk      :: x -> s          secret key from random seed x
    penc    :: (y, p) -> c     encrypt y with pubkey p to give ciphertext c
    pdec    :: (c, s) -> y     decrypt c with seckey s to give cleartext y
    sign    :: (x, s) -> y     sign x with seckey s to give sig y
    verify  :: (x, y, p) -> x  verify x with sig y and pubkey p is equal to x
    hash    :: x -> y          applies a one-way function to term M
    ...

### Processes

    Let u range over names and variables.

    P, Q ::= nil
           | ū<N>.P         send term N on channel u
           | u(x).P         receive term from channel P and assign to variable x
           | !P             replicate process P
           | P|Q            run processes P and Q in parallel
           | (vn)P          restrict name n to process P
           | if M = N
             then P else Q  conditional

### Examples

Probabilistic encryption using a random salt z for semantic security:

    pdec(penc(y, pk(x), z), sk(x)) = y

The Confused Deputy

    TODO

The Grant Matcher Puzzle

    TODO

## References

[Miller2006]: http://www.erights.org/talks/thesis/markm-thesis.pdf "Mark Miller, Robust Composition: Towards a Unified Approach to Access Control and Concurrency Control"

[MillerCapComp]: http://www.erights.org/elib/capability/index.html "Capability Computation"

[Stiegler]: http://www.skyhunter.com/marcs/capabilityIntro/index.html "Mark Stiegler, A Gentle Introduction to Capability Based Security"

[Sitaker2000]: http://www.canonical.org/%7Ekragen/3-sec-arch.html "Kragen Sitaker, The Three Security Architectures"

[Frantz]: http://www.erights.org/elib/capability/3parts.html "Bill Frantz, The Three Parts of Security"

[Slominski2013]: http://www.dalnefre.com/wp/2013/10/towards-a-universal-implementation-of-unforgeable-actor-addresses/#memory-realm "Tristan Slominski, Towards a Universal Implementation of Unforgeable Actor Addresses"

[Goldratt1984]: https://fredlybrand.com/2018/09/13/the-goal-chapter-11-dependent-events-statistical-fluctuations/ "Eli Goldratt, The Goal"

[Confused deputy problem]: https://en.wikipedia.org/wiki/Confused_deputy_problem "Confused deputy problem"

[Hewitt1977]: https://dspace.mit.edu/handle/1721.1/6272 "Carl Hewitt, Viewing Control Structures as Patterns of Passing Messages"

[Kahn88]: https://research.google.com/intl/en/pubs/author35958.html "Kenneth Kahn, and Mark S. Miller, Language Design and Open Systems, in, Bernardo Huberman (ed.), Ecology of Computation (Elsevier Science Publishers/North-Holland, 1988)"

[Rees96]: http://mumble.net/jar/pubs/secureos/ "Jonathan Rees, A Security Kernel Based on the Lambda-Calculus, (MIT, Cambridge, MA, 1996) MIT AI Memo No. 1564."

antoinevg:
## Summary

This PR provides initial support for message flow authorization within
Ockam nodes.

All incoming messages for an Ockam `Address` are now subject to an
authorization check implemented via the `AccessControl` trait before
it will be accepted for further processing.

Major new api's introduced are:

* `Context::start_worker_with_access_control(Address, Worker, AccessControl)`
* `Context::new_repeater(AccessControl) -> RepeaterContext`
* `Worker::is_authorized(&mut Self, Self::Context, Routed<Self::Message>)`
* A new `access_control` attribute for the `#[ockam::node]` macro

### `DetachedContext`, `RepeaterContext`

A `Context` with no associated worker or processor that can still send and receive messages.

A `DetachedContext` inherits the access control of the parent `Context` which created it.
A `RepeaterContext` requires that you specify access control for it on creation.

### `Mailbox`

A `Mailbox` is a single `Address` combined with `AccessControl`.

Previously a `Context` had a single `AccessControl` which was shared with all `Address`es.

### `Mailboxes`

Replaces `Context::address` with `Context::mailboxes`

A collection of `Mailbox`es (`Address` + `AccessControl`)

## Breaking changes

`Worker::handle_message` implementations that may have used a pattern like:

    let msg = LocalMessage::decode(msg.payload())?;
    self.handle_route(ctx, msg).await?;

Should now use:

    self.handle_route(ctx, msg.into_local_message()).await?;

## A word about defaults

At present message flow authorization defaults to an `AllowAll` policy.

This is set by:

* The `#[ockam::node]` macro creates the node's app `Context` with an
  `AllowAll` policy when one has not been specified.
* The `ockam::node::start_node()` function starts new nodes with an
  `AllowAll` policy. You can use
  `ockam::node::start_node_with_access_control()` to specify an
  alternate policy.

It is expected that this will migrate to a default of
`LocalOriginOnly` as we update the rest of the codebase to use message
flow authorization.

Another worthy goal may be to think about ways where we could set it
to `DenyAll` by default without complectifying the ockam api learning
process.

## The new `access_control` attribute for the `#[ockam::node]` macro

By default all nodes are started with an access control of `AllowAll`
for the `"app"` address.

While we could have used the more restrictive `LocalOriginOnly` by
default this would make our "getting started" examples more complex.

An example of overriding the default "app" access control is as
follows:

    use hello_ockam::Echoer;
    use ockam::access_control::{AllowedTransport, LocalOriginOnly};
    use ockam::{Context, Result, TcpTransport, TCP};
    use std::sync::Arc;

    #[ockam::node(access_control = "LocalOriginOnly")]

        tcp.listen("127.0.0.1:4000").await?;

        // Create an echoer worker
        ctx.start_worker_with_access_control(
            "echoer",
            Echoer,
            Arc::new(AllowedTransport::single(TCP))
        )
        .await?;

        // Don't call ctx.stop() here so this node runs forever.
        Ok(())
    }

## Custom worker authorization handling with `Worker::is_authorized`

The worker trait has been extended with an optional `is_authorized` method:

    async fn is_authorized(
        &mut self,
        _context: &mut Self::Context,
        _msg: Routed<Self::Message>,
    ) -> Result<bool> {
        crate::allow()
    }

This method will be invoked after any authorization performed on the
`Address` of the worker and before the message is forwarded to the
worker's `handle_message()` method is called.

The rationale for adding this trait instead of simply performing any
extended authorization within the `Worker::handle_message()` call is
twofold:

1. At some future point we may want to implement functionality to
   optionally disable authorization checks for trouble-shooting
   purposes.
1. Explicit often works out better in the long run than implicit.

### DIFF
--- a/implementations/rust/ockam/ockam_core/src/compat.rs
+++ b/implementations/rust/ockam/ockam_core/src/compat.rs
@@ -273,3 +273,54 @@ pub mod fmt {
     #[cfg(not(feature = "alloc"))]
     pub type Vec<T> = heapless::Vec<T, 64>;
 }
+
+/// Provides `future::poll_once`
+pub mod future {
+    use crate::{
+        errcode::{Kind, Origin},
+        Error, Result,
+    };
+    use futures_util::future::{Future, FutureExt};
+
+    /// Polls a future just once and returns the Result
+    ///
+    /// This is only used for some tests and it is hoped that we can
+    /// remove it if, at some point, this makes it into `core::future`
+    pub fn poll_once<'a, F, T>(future: F) -> Result<T>
+    where
+        F: Future<Output = Result<T>> + Send + 'a,
+    {
+        use core::task::{Context, Poll};
+        use core::task::{RawWaker, RawWakerVTable, Waker};
+
+        fn dummy_raw_waker() -> RawWaker {
+            fn no_op(_: *const ()) {}
+            fn clone(_: *const ()) -> RawWaker {
+                dummy_raw_waker()
+            }
+            let vtable = &RawWakerVTable::new(clone, no_op, no_op, no_op);
+            RawWaker::new(core::ptr::null(), vtable)
+        }
+
+        fn dummy_waker() -> Waker {
+            // The RawWaker's vtable only contains safe no-op
+            // functions which do not refer to the data field.
+            #[allow(unsafe_code)]
+            unsafe {
+                Waker::from_raw(dummy_raw_waker())
+            }
+        }
+
+        let waker = dummy_waker();
+        let mut context = Context::from_waker(&waker);
+        let result = future.boxed().poll_unpin(&mut context);
+        assert!(
+            result.is_ready(),
+            "poll_once() only accepts futures that resolve after being polled once"
+        );
+        match result {
+            Poll::Ready(value) => value,
+            Poll::Pending => Err(Error::new_without_cause(Origin::Core, Kind::Invalid)),
+        }
+    }
+}
--- a/implementations/rust/ockam/ockam_node/src/context.rs
+++ b/implementations/rust/ockam/ockam_node/src/context.rs
@@ -1,8 +1,5 @@
 use crate::async_drop::AsyncDrop;
-use crate::channel_types::{
-    message_channel, small_channel, MessageReceiver, SmallReceiver, SmallSender,
-};
-use crate::relay::RelayPayload;
+use crate::channel_types::{message_channel, small_channel, SmallReceiver, SmallSender};
 use crate::tokio::{self, runtime::Runtime, time::timeout};
 use crate::{
     error::*,
@@ -18,7 +15,7 @@ use core::{
 use ockam_core::compat::{boxed::Box, string::String, sync::Arc, vec::Vec};
 use ockam_core::{
     errcode::{Kind, Origin},
-    AccessControl, Address, AddressSet, AllowAll, AsyncTryClone, Error, LocalMessage, Message,
+    Address, AddressSet, AllowAll, AsyncTryClone, Error, LocalMessage, Mailbox, Mailboxes, Message,
     Processor, Result, Route, TransportMessage, TransportType, Worker,
 };
 
@@ -47,12 +44,11 @@ pub type DetachedContext = Context;
 
 /// Context contains Node state and references to the runtime.
 pub struct Context {
-    address: AddressSet,
+    mailboxes: Mailboxes,
     sender: SmallSender<NodeMessage>,
     rt: Arc<Runtime>,
+    receiver: SmallReceiver<RelayMessage>,
     async_drop_sender: Option<AsyncDropSender>,
-    mailbox: MessageReceiver<RelayMessage>,
-    access_control: Box<dyn AccessControl>,
     mailbox_count: Arc<AtomicUsize>,
 }
 
@@ -80,24 +76,28 @@ impl Context {
         self.rt.clone()
     }
     /// Wait for the next message from the mailbox
-    pub(crate) async fn mailbox_next(&mut self) -> Result<Option<RelayMessage>> {
+    pub(crate) async fn receiver_next(&mut self) -> Result<Option<RelayMessage>> {
         loop {
-            let relay_msg = if let Some(msg) = self.mailbox.recv().await {
+            let relay_msg = if let Some(msg) = self.receiver.recv().await.map(|msg| {
                 trace!("{}: received new message!", self.address());
 
                 // First we update the mailbox fill metrics
                 self.mailbox_count.fetch_sub(1, Ordering::Acquire);
 
                 msg
+            }) {
+                msg
             } else {
                 return Ok(None);
             };
 
-            if let RelayPayload::Direct(local_msg) = &relay_msg.data {
-                if !self.access_control.is_authorized(local_msg).await? {
-                    warn!("Message for {} did not pass access control", relay_msg.addr);
-                    continue;
-                }
+            if !self
+                .mailboxes
+                .is_authorized(&relay_msg.addr, &relay_msg.local_msg)
+                .await?
+            {
+                warn!("Message for {} did not pass access control", relay_msg.addr);
+                continue;
             }
 
             return Ok(Some(relay_msg));
@@ -116,20 +116,18 @@ impl Context {
     pub(crate) fn new(
         rt: Arc<Runtime>,
         sender: SmallSender<NodeMessage>,
-        address: AddressSet,
+        mailboxes: Mailboxes,
         async_drop_sender: Option<AsyncDropSender>,
-        access_control: impl AccessControl,
     ) -> (Self, SenderPair, SmallReceiver<CtrlSignal>) {
-        let (mailbox_tx, mailbox) = message_channel();
+        let (mailbox_tx, receiver) = message_channel();
         let (ctrl_tx, ctrl_rx) = small_channel();
         (
             Self {
                 rt,
                 sender,
-                address,
-                mailbox,
+                mailboxes,
+                receiver,
                 async_drop_sender,
-                access_control: Box::new(access_control),
                 mailbox_count: Arc::new(0.into()),
             },
             SenderPair {
@@ -142,12 +140,12 @@ impl Context {
 
     /// Return the primary address of the current worker
     pub fn address(&self) -> Address {
-        self.address.first()
+        self.mailboxes.main_address()
     }
 
     /// Return all addresses of the current worker
     pub fn aliases(&self) -> AddressSet {
-        self.address.clone().into_iter().skip(1).collect()
+        self.mailboxes.aliases()
     }
 
     /// Utility function to sleep tasks from other crates
@@ -161,11 +159,20 @@ impl Context {
     /// Note: this function is very low-level.  For most users
     /// [`start_worker()`](Self::start_worker) is the recommended to
     /// way to create a new worker context.
-    pub async fn new_detached<S: Into<Address>>(&self, addr: S) -> Result<DetachedContext> {
-        self.new_detached_impl(addr.into()).await
+    pub async fn new_detached<S: Into<AddressSet>>(
+        &self,
+        address_set: S,
+    ) -> Result<DetachedContext> {
+        // Inherit access control
+        let access_control = self.mailboxes.main_mailbox().access_control().clone();
+
+        let mailboxes = Mailboxes::from_address_set(address_set.into(), access_control);
+
+        self.new_detached_impl(mailboxes).await
     }
 
-    async fn new_detached_impl(&self, addr: Address) -> Result<DetachedContext> {
+    /// TODO FIXME this needs to be made private again after the examples have been refactored
+    pub async fn new_detached_impl(&self, mailboxes: Mailboxes) -> Result<DetachedContext> {
         // A detached Context exists without a worker relay, which
         // requires special shutdown handling.  To allow the Drop
         // handler to interact with the Node runtime, we use an
@@ -178,17 +185,17 @@ impl Context {
         async_drop.spawn(&self.rt);
 
         // Create a new context and get access to the mailbox senders
+        let addresses = mailboxes.addresses();
         let (ctx, sender, _) = Self::new(
             Arc::clone(&self.rt),
             self.sender.clone(),
-            addr.clone().into(),
+            mailboxes,
             Some(drop_sender),
-            AllowAll,
         );
 
         // Create a "detached relay" and register it with the router
         let (msg, mut rx) =
-            NodeMessage::start_worker(addr.into(), sender, true, Arc::clone(&self.mailbox_count));
+            NodeMessage::start_worker(addresses, sender, true, Arc::clone(&self.mailbox_count));
         self.sender
             .send(msg)
             .await
@@ -196,6 +203,7 @@ impl Context {
         rx.recv()
             .await
             .ok_or_else(|| NodeError::NodeState(NodeReason::Unknown).internal())??;
+
         Ok(ctx)
     }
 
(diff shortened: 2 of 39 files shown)


implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -156,16 +166,40 @@ impl Context {
         tokio::time::sleep(dur).await;
     }
 
-    /// Create a new detached context without spawning a full worker
+    /// Create a new detached `Context` that will apply the given
+    /// [`AccessControl`] to any incoming messages it receives
+    pub async fn new_repeater<AC>(&self, access_control: AC) -> Result<RepeaterContext>
```
SanjoDeundiak, line 171: I think it's the same as DetachedContext, we just need function that would allow passing AccessControl

implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -156,16 +166,40 @@ impl Context {
...
     /// [`start_worker()`](Self::start_worker) is the recommended to
     /// way to create a new worker context.
-    pub async fn new_detached<S: Into<Address>>(&self, addr: S) -> Result<DetachedContext> {
-        self.new_detached_impl(addr.into()).await
+    pub async fn new_detached<S: Into<AddressSet>>(
+        &self,
+        address_set: S,
+    ) -> Result<DetachedContext> {
+        // Inherit access control
+        let access_control = self.mailboxes.main_mailbox().access_control().clone();
```
SanjoDeundiak, line 194: Possible improvement: replace clone() with inherit() with optional message

implementations/rust/ockam/ockam_node/src/context.rs
antoinevg: Yup, It's also just a type alias to `Context` for now but I'd like to start typing them according to function so long in anticipation of future refactors.

implementations/rust/ockam/ockam_core/src/compat.rs
```diff
@@ -273,3 +273,54 @@ pub mod fmt {
     #[cfg(not(feature = "alloc"))]
     pub type Vec<T> = heapless::Vec<T, 64>;
 }
+
+/// Provides `future::poll_once`
+pub mod future {
```
SanjoDeundiak, line 278: Does it belong to this PR?

implementations/rust/ockam/ockam_node/src/local_info.rs
```diff
@@ -0,0 +1,74 @@
+use ockam_core::{
+    errcode::{Kind, Origin},
+    Decodable, Encodable, Error, LocalInfo, LocalMessage, Result, TransportType,
+};
+use serde::{Deserialize, Serialize};
+
+/// Node LocalInfo unique Identifier
+pub const EXTERNAL_IDENTIFIER: &str = "EXTERNAL_IDENTIFIER";
```
SanjoDeundiak, line 8: Maybe replace with CBOR TypeTag in the future

SanjoDeundiak (approved):

implementations/rust/ockam/ockam_core/src/compat.rs
antoinevg: To avoid cloning it for all the various AccessControl implementor tests yup :)

implementations/rust/ockam/ockam_node/src/context.rs
antoinevg: Good call, I'll implement this in the ABAC PR.

implementations/rust/ockam/ockam_node/src/local_info.rs
antoinevg: Check, I'll implement this in the ABAC PR.

PR #2591, opened 2022-03-24
Context impl debug

Jackbaude:
## Current Behaviour
Context does not impl std::fmt::Debug

## Proposed Changes
impl `std::fmt::Debug` for `Context`
https://github.com/ockam-network/ockam/issues/2576

## Checks

- [x] All commits in this Pull Request are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- [x] All commits in this Pull Request follow the Ockam [commit message convention](https://www.ockam.io/learn/how-to-guides/contributing/CONTRIBUTING#commit-messages).
- [x] I accept the Ockam Community [Code of Conduct](https://www.ockam.io/learn/how-to-guides/high-performance-team/conduct/).
- [x] I have accepted the Ockam [Contributor Licence Agreement](https://www.ockam.io/learn/how-to-guides/contributing/cla/) by adding my Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/ockam-network/contributors/blob/master/CONTRIBUTORS.csv) file in a separate pull request to the [ockam-network/contributors](https://github.com/ockam-network/contributors) repository.

PR #2804, opened 2022-06-01
feat(rust): add auth api

twittner:


implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -52,6 +53,12 @@ pub struct Context {
     access_control: Box<dyn AccessControl>,
 }
 
+impl fmt::Debug for Context {
```
SanjoDeundiak: Is that overlapping with #2591 ?

implementations/rust/ockam/ockam_api/src/lib.rs
```diff
@@ -469,3 +472,105 @@ impl<T: Encode<()>> ResponseBuilder<T> {
...
 }
+
+/// A newtype around `Cow<'_, str>` that borrows from input.
+///
+/// Contrary to `Cow<_, str>` the `Decode` impl for this type will always borrow
+/// from input so using it in types like `Option`, `Vec<_>` etc will not produce
+/// owned element values.
+#[derive(Debug, Clone, Encode, Decode, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cbor(transparent)]
+pub struct CowStr<'a>(#[b(0)] pub Cow<'a, str>);
```
SanjoDeundiak: What about putting this to a separate file, or `ockam_core`, or even minicbor? And this is definitely separate commit

implementations/rust/ockam/ockam_api/src/auth.rs
```diff
@@ -0,0 +1,317 @@
...
+use types::{Attribute, Attributes};
+
+/// Auth API server.
+#[derive(Debug)]
+pub struct Server<S> {
+    store: S,
+}
+
+#[ockam_core::worker]
+impl<S: Storage + Send + Sync + 'static> Worker for Server<S> {
```
SanjoDeundiak, line 25: Pros vs Cons adding `Send + Sync + 'static` as a requirement for the Storage trait itself? Not sure we ever use it without this requirement

SanjoDeundiak (dismissed):
Nice. Few questions
||
v

implementations/rust/ockam/ockam_node/src/context.rs
twittner: Indeed it is.

implementations/rust/ockam/ockam_api/src/lib.rs
twittner: I will move this into a separate PR.

implementations/rust/ockam/ockam_api/src/auth.rs
twittner: Still more general to constrain it on call site I think.

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #2807, opened 2022-06-01
feat(rust): add cowbytes and cowstr

twittner:
The `Cow<'_, T>` impl of `Decode` produces owned values. Types that derive `Decode` can declare fields as borrowing from input. This activates some special handling of `Cow`s, i.e. if they hold string or byte slices, the derived code will produce borrowed `Cow`s. This special handling does not kick in for types that do not derive `Decode`, e.g. `Vec`, `Option`, `BTreeMap` etc. `Cow`s as element types of such types will therefore not borrow from input data. To enable borrowing `Cow`s for such types, this PR adds two newtypes for `Cow<'_, str>` and `Cow<'_, [u8]>` that will always borrow from input, hence they are intended for use with `Vec`tors, `Option`s etc for cases where the usual `Cow` semantics are desired.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1747,18 +1747,18 @@ checksum = "2a60c7ce501c71e03a9c9c0d35b861413ae925bd979cc7a4e30d060069aaac8d"
 
 [[package]]
 name = "minicbor"
-version = "0.16.0"
+version = "0.17.1"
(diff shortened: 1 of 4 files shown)


SanjoDeundiak (dismissed):
Cool

SanjoDeundiak (approved):

PR #2835, opened 2022-06-07
Identity service

SanjoDeundiak:


implementations/rust/ockam/ockam_identity/src/service/models.rs
```diff
@@ -0,0 +1,241 @@
...
+use minicbor::{Decode, Encode};
+use ockam_api::{CowBytes, CowStr};
+
+#[cfg(feature = "tag")]
+use ockam_core::TypeTag;
+
+#[derive(Debug, Clone, Encode, Decode)]
+#[rustfmt::skip]
+#[cbor(map)]
+pub struct CreateResponse<'a> {
```
hairyhum, line 12: Is there `CreateRequest`?
In the API doc there is a `name` field.

SanjoDeundiak: What API doc? We don't have names for Identities

SanjoDeundiak: In the future we definitely want to at least support providing `key_id`, for now it should be fine without it

implementations/rust/ockam/ockam_identity/src/service.rs
```diff
@@ -0,0 +1,229 @@
...
+
+        Ok(res.verified())
+    }
+
+    #[ockam_macros::test]
+    async fn full_flow(ctx: &mut Context) -> Result<()> {
+        let vault1 = Vault::create();
+        let vault2 = Vault::create();
+        // Start services
+        IdentityService::create(ctx, "01", vault1.async_try_clone().await?).await?;
```
hairyhum, line 192: I don't really get why create multiple services here.

SanjoDeundiak: Just to simulate restarts and make sure everything is still working. It could have been 1 service

spacekookie (approved):

PR #2838, opened 2022-06-07
feat(elixir): add is_authorized to the worker behaviour

hairyhum:
Workers now can specify `is_authorized(message, state) :: :ok | {:error, reason}`
function to allow or deny some messages to be handled.

Added `Ockam.Worker.Authorization` module to provide some basic authorization helpers:

`allow_all()` - handle all messages
`deny_all()` - refuse all messages
`from_addresses(message, addresses)` - only handle messages with return route starting with one of the `addresses`
`to_addresses(message, addresses)` - only handle messages with onward route starting with one of the `addresses`
`to_my_address(message, state)` - only handle messages with onward route starting with one of the addresses from the state key `:all_addresses`
`is_secure(message)` - only handle message for which local metadata key `:transport` is set to `:secure_channel` TODO: add this key when message exiting secure channel
`from_identity(message, identity)` - only handle messages with local metadata key `:identity` set to `identity`

Authorization helpers can be chained with the `|>` pipe operator:

```
Authorization.from_addresses(message, ["one", "two"])
|> Authorization.to_my_addresses(message, state)
```

**NOTE**
Workers when registering additional address now have to use `Ockam.Worker.register_extra_addresses/3` or `Ockam.Worker.register_random_extra_address/2` to update `:all_addresses` in the worker state.

**BREAKING CHANGE**
Make all workers use `to_my_address` by default for `is_authorized/2`, which means that workers only handle messages sent to `:address` or addresses registered with `Ockam.Worker.register_extra_addresses/3`

Refactored `Ockam.Worker` `__using__` macro to delegate to `Ockam.Worker` functions.
This will help with debug to locate the right line and also allow to create workers without `use Ockam.Worker` e.g. from Erlang.

Removed `get_from_options/2` function

Added optional `timeout` to `Worker.create` function to allow creating process to wait longer for the worker to start

**BREAKING CHANGE**
Removed aliases from the `Ockam.Worker` `__using__` macro. Not `Node` and `Router` aliases will not be automatically created for the workers.

Added `Message.local_metadata/1` `Message.local_metadata_value/2` getters to fetch message metadata

Make `Ockam.Node.register_address/1,2` return `:ok` if the address is already registered for the current process.

Make `Ockam.Node.register_random_address/0,1,2,3` only retry registering if the error is `already_registered`

## Checks

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/worker/authorization.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/worker/authorization.ex
@@ -1,0 +1,129 @@
+defmodule Ockam.Worker.Authorization do
+  @moduledoc """
+  Helper functions to configure worker message authorization.
+
+  Usage:
+  ```
+  def is_authorized(message, _state)
+    Authorization.from_addresses(message, ["one", "two"])
+  end
+  ```
+
+  Pipelining helpers:
+  ```
+  def is_authorized(message, state)
+    Authorization.from_addresses(message, ["one", "two"])
+    |> Authorization.to_my_address(message, state)
+  end
+  ```
+  """
+  alias Ockam.Message
+
+  @doc """
+  Allow any messages to be handled by the worker
+  """
+  def allow_all(prev \\ :ok) do
+    chain(prev, fn -> :ok end)
+  end
+
+  @doc """
+  Deny all messages from being handled by the worker
+  """
+  def deny_all(prev \\ :ok) do
+    chain(prev, fn -> {:error, :deny_all} end)
+  end
+
+  @doc """
+  Only allow messages from a set of addresses to be handled by the worker.
+
+  Message address is taken from return_route in that case
+  """
+  def from_addresses(prev, message, addresses) when is_list(addresses) do
+    chain(prev, fn ->
+      from_addresses(message, addresses)
+    end)
+  end
+
+  def from_addresses(message, addresses) do
+    case Message.return_route(message) do
+      [return_address | _rest] ->
+        case Enum.member?(addresses, return_address) do
+          true -> :ok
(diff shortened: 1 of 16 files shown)


implementations/elixir/ockam/ockam/lib/ockam/worker/authorization.ex
```diff
@@ -0,0 +1,129 @@
...
+  """
+  def to_addresses(prev, message, addresses) when is_list(addresses) do
+    chain(prev, fn ->
+      to_addresses(message, addresses)
+    end)
+  end
+
+  def to_addresses(message, addresses) do
+    case Message.onward_route(message) do
+      [onward_address | _rest] ->
```
polvorin, line 73: would it make sense to enforce that this is the _final_ step in the route?

polvorin (dismissed):
LGTM,  left some comment regarding verification of target addresses. Lets say the worker address is `A`  ,  and specify `to_my_address` as authorization.   I'd have expected it to accept messges with route [`A`], but reject messages with route [`A`,  `B`]  (could perhaps help catch errors?)

implementations/elixir/ockam/ockam/lib/ockam/worker/authorization.ex
hairyhum: The intention for this verification was that the message is sent to this worker address. And messages get sent to the first address in the onward route. What happens to the rest of the route is up to the worker to decide, whether it forwards it or not.
We might add another authorization rule that the worker address is the only one left in the onward route, but I'm not sure we need that right now. If we have workers implementing that we can generalize that as a rule later.

mrinalwadhwa (dismissed):

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #2839, opened 2022-06-08
Identity service

SanjoDeundiak:


implementations/rust/ockam/ockam_identity/src/service/schema.cddl
```diff
@@ -0,0 +1,58 @@
+create_response = {
+    ?0: 3500430,
+     1: identity: CowBytes<'a,
```
twittner:
```suggestion
     1: bytes,
```
The key is `1`. Adding `identity: CowBytes<'a>` as invalid syntax and CDDL does not know about this type. You can add a comment:

```
     1: bytes, ;; identity
```

or write

```
     1: identity,
     2: identity_id
}
```

where `identity` and `identity_id` are already defined at the end of the file.

antoinevg (approved):
I can't speak to the crypto but the rest of the PR looks sane.

PR #2841, opened 2022-06-08
feat(rust): extend contact api for identity service

SanjoDeundiak:


implementations/rust/ockam/ockam_identity/src/service/schema.cddl
```diff
@@ -56,3 +57,4 @@ state            = bytes
 proof            = bytes
 peer_identity_id = text
 verified         = bool
+contact_id       = text
```
spacekookie, line 60: We should probably consolidate a lot of these types and schema updates in the `ockam_api` crate but this is fine for now (since we already landed some of these changes anyway).

spacekookie (approved):

implementations/rust/ockam/ockam_identity/src/service/schema.cddl
SanjoDeundiak: I think we're in the process of moving services into ockam_api, which will allow that

PR #2848, opened 2022-06-10
fix(rust): listen on `0.0.0.0` instead of `127.0.0.1`

mrinalwadhwa:
When a node is created using ockam command, make it listen
on `0.0.0.0` instead of `127.0.0.1.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -76,7 +76,7 @@ impl CreateCommand {
 
 async fn setup(ctx: Context, c: CreateCommand) -> anyhow::Result<()> {
     let tcp = TcpTransport::create(&ctx).await?;
-    tcp.listen(format!("127.0.0.1:{}", c.port)).await?;
+    tcp.listen(format!("0.0.0.0:{}", c.port)).await?;
 
     let s = auth::store::mem::Store::new();
     ctx.start_worker("authenticated", auth::Server::new(s))


implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -76,7 +76,7 @@ impl CreateCommand {
 
 async fn setup(ctx: Context, c: CreateCommand) -> anyhow::Result<()> {
     let tcp = TcpTransport::create(&ctx).await?;
-    tcp.listen(format!("127.0.0.1:{}", c.port)).await?;
+    tcp.listen(format!("0.0.0.0:{}", c.port)).await?;
```
antoinevg, line 79: This may earn us a security advisory! 😄

That said, do we have a `--listen` command line parameter so we can specify the interfaces to listen on?

I'm also curious if we can come up with something that moves the conversation further along than _"localhost should always be the default even when it makes things harder for new users"_.

One idea could be to…

1) output a debug something like:

        Node started. Accepting connections on: 127.0.0.1:4000, 192.168.20.226:4000, 192.168.20.225:4000

2) reference this message in both the getting started and deployment guides along with instructions on how to bind to a specific interface

antoinevg (approved):

antoinevg:
Argh, bitten by the auto merge! Forgot to check if the label was applied before hitting approve - would be nice to hear other comments too.

I've opened an issue here: https://github.com/build-trust/ockam/issues/2849

PR #2855, opened 2022-06-14
refactor(rust): use Attributes type on RequestEnrollmentToken

adrianbenavides:
Replace `Vec<TokenAttribute<'a>>` by `Attributes<'a>` to define the attributes of the `RequestEnrollmentToken` request structure. It also removes the `TokenAttribute` type since it's no longer used.

This will need changes on the elixir side, so I'll keep it as a draft while @polvorin tests these changes against the elixir side.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cloud/enroll.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/enroll.rs
@@ -137,30 +137,30 @@ pub mod auth0 {
 }
 
 pub mod enrollment_token {
+    use crate::auth::types::Attributes;
+
     use super::*;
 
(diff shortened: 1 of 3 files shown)


polvorin:
@adrianbenavides  it's ok to merge :+1: .   Once we have identity channel working,  it would be simpler to pass  `Attributes` directly instead of a  `RequestEnrollmentToken`  containing an `Attributes` field  (right now that would be cumbersome and require more changes as authenticated and token creation commands use different ways to pass the `identity`, so your approach is a good compromise)

polvorin (dismissed):

mrinalwadhwa (approved):

PR #2859, opened 2022-06-14
ockam CLI portal command

spacekookie:
This PR adds the ability to create portal inlets and outlets in ockam command.

It also re-structures the transports command to use sub-commands for tcp-listener and tcp-connector creation. In the future we may want to re-implement these commands as "Addon Commands" (the basic structure is also included in this PR).

However to support both add-ons, and meaningful help text generation via clap, we will need to restructure the way we build the CLI fundamentally (i.e. use the builder patterns, not the magic derives)

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/iolets.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/iolets.rs
@@ -6,40 +6,70 @@ use ockam_core::compat::borrow::Cow;
 #[cfg(feature = "tag")]
 use ockam_core::TypeTag;
 
-/// Request body to create an inlet
-#[derive(Debug, Clone, Decode, Encode)]
+/// Distinguish between inlet and outlet portals in the API
+#[derive(Copy, Clone, Debug, Decode, Encode)]
+#[rustfmt::skip]
+#[cbor(index_only)]
+pub enum IoletType {
+    #[n(0)] Inlet,
+    #[n(1)] Outlet,
+}
+
+/// Request body to create an inlet or outlet
+#[derive(Clone, Debug, Decode, Encode)]
 #[rustfmt::skip]
 #[cbor(map)]
-pub struct CreateInlet<'a> {
+pub struct CreateIolet<'a> {
     #[cfg(feature = "tag")]
     #[n(0)]
     tag: TypeTag<1407961>,
-    #[n(1)] pub addr: Cow<'a, str>,
+    #[n(1)] pub tt: IoletType,
+    #[n(2)] pub addr: Cow<'a, str>,
+    #[n(3)] pub alias: Option<Cow<'a, str>>,
 }
 
-/// Request body to create an outlet
-#[derive(Debug, Clone, Decode, Encode)]
-#[rustfmt::skip]
-#[cbor(map)]
-pub struct CreateOutlet<'a> {
-    #[cfg(feature = "tag")]
-    #[n(0)]
-    tag: TypeTag<1407961>,
-    #[n(1)] pub addr: Cow<'a, str>,
-}
-
-/// Respons body when interacting with a transport
-#[derive(Debug, Clone, Decode, Encode)]
+/// Respons body when interacting with an inlet or outlet
+#[derive(Clone, Debug, Decode, Encode)]
 #[rustfmt::skip]
 #[cbor(map)]
 pub struct IoletStatus<'a> {
     #[cfg(feature = "tag")]
     #[n(0)]
     tag: TypeTag<1581592>,
-    #[n(1)] pub payload: Cow<'a, str>,
-    /// Iolet ID inside the node manager
-    ///
-    /// We use this as a kind of URI to be able to address a transport
-    /// by a unique value for specific updates and deletion events.
-    #[n(2)] pub ioid: Cow<'a, str>,
+    #[n(1)] pub tt: IoletType,
+    #[n(2)] pub addr: Cow<'a, str>,
+    #[n(3)] pub alias: Cow<'a, str>,
+}
+
+impl<'a> IoletStatus<'a> {
+    pub fn new<S: Into<Cow<'a, str>>>(tt: IoletType, addr: S, alias: S) -> Self {
+        Self {
+            #[cfg(feature = "tag")]
+            tag: TypeTag,
+            tt,
+            addr: addr.into(),
+            alias: alias.into(),
+        }
+    }
+}
+
+/// Responsebody when returning a list of Iolets
+#[derive(Debug, Clone, Decode, Encode)]
+#[rustfmt::skip]
+#[cbor(map)]
+pub struct IoletList<'a> {
+    #[cfg(feature = "tag")]
+    #[n(0)]
+    tag: TypeTag<8401504>,
+    #[n(1)] pub list: Vec<IoletStatus<'a>>
+}
+
+impl<'a> IoletList<'a> {
+    pub fn new(list: Vec<IoletStatus<'a>>) -> Self {
+        Self {
+            #[cfg(feature = "tag")]
+            tag: TypeTag,
+            list,
+        }
+    }
 }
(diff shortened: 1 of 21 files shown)


spacekookie:
I'm having some issues with my yubikey at the moment which is why the commits are not signed. I'll rebase with a different computer

implementations/rust/ockam/ockam_api/src/nodes/iolets.rs
```diff
@@ -6,40 +6,106 @@ use ockam_core::compat::borrow::Cow;
...
-pub struct CreateInlet<'a> {
+pub struct CreateIolet<'a> {
     #[cfg(feature = "tag")]
     #[n(0)]
     tag: TypeTag<1407961>,
-    #[n(1)] pub addr: Cow<'a, str>,
+    /// The type of portal endpoint to create
+    #[n(1)] pub tt: IoletType,
+    /// The address the portal should connect or bind to
+    #[n(2)] pub addr: Cow<'a, str>,
```
SanjoDeundiak: Shouldn't you use `#[b(2)]` ?

spacekookie: Oh yea probably. I've just been using `#[n(_)]` for all the fields at the moment

implementations/rust/ockam/ockam_api/src/nodes/iolets.rs
```diff
@@ -6,40 +6,106 @@ use ockam_core::compat::borrow::Cow;
...
+            tag: TypeTag,
+            tt,
+            addr: addr.into(),
+            fwd: fwd.into(),
+            alias: alias.into(),
+        }
+    }
+}
+
+/// Respons body when interacting with an inlet or outlet
```
SanjoDeundiak: typo

implementations/rust/ockam/ockam_command/src/transport/create.rs
```diff
@@ -9,20 +9,23 @@ pub struct CreateCommand {
...
-    pub address: String,
+#[derive(Clone, Debug, Subcommand)]
+pub enum CreateTypeCommand {
+    /// Create a TCP listener transport
+    TcpListener {
+        /// Transport connection or bind address
+        bind: String,
+    },
+    /// Create a TCP connector transport
+    TcpConnector {
```
SanjoDeundiak, line 34: What is a connector?

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -131,6 +144,74 @@ impl NodeMan {
...
+            tt,
+            ..
+        } = dec.decode()?;
+        let addr = addr.to_string();
+        let alias = alias.map(|a| a.into()).unwrap_or_else(random_alias);
+
+        let res = match tt {
+            IoletType::Inlet => {
+                info!("Handling request to create inlet portal");
+                let outlet_route = Route::parse(fwd.unwrap()).unwrap();
```
SanjoDeundiak: Please remove unwrap here or in following PR

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -168,14 +251,22 @@ impl NodeMan {
...
                 .body(TransportList::new(self.get_transports()))
                 .encode(enc)?,
-            // TODO: Get all transports
-            // == Create a new transport
             (Post, "/node/transport") => self.add_transport(req, dec).await?.encode(enc)?,
             (Delete, "/node/transport") => self.delete_transport(req, dec).await?.encode(enc)?,
+
+            // ==*== Inlets & Outlets ==*==
+            (Get, "/node/portal") => self.get_iolets(req).encode(enc)?,
+            (Post, "/node/portal") => self.create_iolet(ctx, req, dec).await?.encode(enc)?,
```
SanjoDeundiak, line 406: I think having 2 separate endpoints would be convenient, since arguments are different
Also, we should stick to either portal or iolet term

implementations/rust/ockam/ockam_api/src/nodes/iolets.rs
```diff
@@ -6,40 +6,106 @@ use ockam_core::compat::borrow::Cow;
...
+    #[n(1)] pub tt: IoletType,
+    /// The address the portal should connect or bind to
+    #[b(2)] pub addr: Cow<'a, str>,
+    /// The forwarding address (must be ockam routing address)
+    ///
+    /// This field will be disregarded for portal outlets.  Portal
+    /// inlets MUST set this value to configure their forwarding
+    /// behaviour.  This can either be the address of an already
+    /// created outlet, or a forwarding mechanism via ockam cloud.
+    #[b(3)] pub fwd: Option<Cow<'a, str>>,
```
SanjoDeundiak: This name was misleading for me, as we have forwarding addresses. Could this have different name?
Also, what is the format for it? Do we want to use multiaddr for any user-facing Addresses/Routes?

implementations/rust/ockam/ockam_api/src/nodes/transport.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+}
+
+/// Respons body when interacting with a transport
+#[derive(Debug, Clone, Decode, Encode)]
+#[rustfmt::skip]
+#[cbor(map)]
+pub struct TransportList<'a> {
+    #[cfg(feature = "tag")]
+    #[n(0)]
```
SanjoDeundiak: Unneeded line break here I think. This rustfmt::skip has its flaws

implementations/rust/ockam/ockam_command/src/portal/create.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+pub async fn create_portal(
+    ctx: Context,
+    cmd: CreateCommand,
+    mut base_route: Route,
+) -> anyhow::Result<()> {
+    let resp: Vec<u8> = ctx
+        .send_and_receive(
+            base_route.modify().append("_internal.nodeman"),
```
SanjoDeundiak: "_internal.nodeman" deserves a const

implementations/rust/ockam/ockam_command/src/portal/create.rs
```diff
@@ -0,0 +1,97 @@
...
+            api::create_portal(&cmd)?,
+        )
+        .await
+        .unwrap();
```
SanjoDeundiak: I'm concerned we can't use `?` here

implementations/rust/ockam/ockam_command/src/portal/mod.rs
```diff
@@ -0,0 +1,31 @@
+mod create;
+// mod delete;
```
SanjoDeundiak: Are we going to bring it back?

SanjoDeundiak (changes requested):

implementations/rust/ockam/ockam_api/src/nodes/service.rs
spacekookie: The "portal" term has superseded the "iolets" term so I'll go change the references to iolets in the code. I don't think these should be different endpoints, that's why they're different methods. Different HTTP methods would have different payloads, so it's the same here

implementations/rust/ockam/ockam_api/src/nodes/iolets.rs
spacekookie: I renamed it to `peer`. Currently this is not a multiaddr but I'll change the whole CLI to interact with multiaddrs in a different PR

implementations/rust/ockam/ockam_command/src/portal/create.rs
spacekookie: Oh yes we can just use `?`. My bad :)

implementations/rust/ockam/ockam_command/src/portal/mod.rs
spacekookie: Yes we should. I deleted the commented out modules for now though and added a `TODO`

implementations/rust/ockam/ockam_command/src/transport/create.rs
spacekookie: I've added a better comment to the type which will be displayed in the help text. Not sure why github isn't picking up on the change

SanjoDeundiak (approved):

PR #2867, opened 2022-06-15
feat(rust): enroll adjustments

adrianbenavides:
* Change some messages shown to the user.
* Refactor how we read the user input. Doing it manually with `std::io` is error prone and for some reason both stdout and stderr were printed out even if no error occurred.
* Also, I've ignored a flaky test on the websocket crate. I'll look into that later on, but at least we won't have to retry jobs so often

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -634,6 +634,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "console"
+version = "0.15.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


mrinalwadhwa:
This should be a separate commit.
> Also, I've ignored a flaky test on the websocket crate. I'll look into that later on, but at least we won't have to retry jobs so often

adrianbenavides:
> This should be a separate commit.
> 
> > Also, I've ignored a flaky test on the websocket crate. I'll look into that later on, but at least we won't have to retry jobs so often

True, it's ready now!

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #2869, opened 2022-06-15
feat(rust): create worker builder for cleaner worker init

antoinevg:
This commit removes the following public methods and replaces them with the worker builder:

    Context::start_worker_with_access_control
    Context::start_worker_with_mailboxes

For example:

    ctx.start_worker_with_access_control("echoer", Echoer, AllowedTransport::single(TCP))

Becomes

    Builder::with_access_control(AllowedTransport::single(TCP), "echoer", Echoer) .start(&ctx)

### DIFF
--- a/examples/rust/get_started/examples/06-access-control-over-transport-responder.rs
+++ b/examples/rust/get_started/examples/06-access-control-over-transport-responder.rs
@@ -3,7 +3,7 @@
 
 use hello_ockam::Echoer;
 use ockam::access_control::{AllowedTransport, LocalOriginOnly};
-use ockam::{Context, Result, TcpTransport, TCP};
+use ockam::{Context, Result, TcpTransport, WorkerBuilder, TCP};
 
 #[ockam::node(access_control = "LocalOriginOnly")]
 async fn main(ctx: Context) -> Result<()> {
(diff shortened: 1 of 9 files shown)


spacekookie (dismissed):
We can probably bikeshed the builder API a bit more in the future. Currently each of the options are very comprehensive and not composable. That might be something we'll want to look into. But that can also be future-us' problem :)

antoinevg:
> We can probably bikeshed the builder API a bit more in the future. Currently each of the options are very comprehensive and not composable. That might be something we'll want to look into. But that can also be future-us' problem :)

Yeah, there are only mutually-exclusive mandatory options at the moment so it comes across somewhat clunky… I must confess just how tempted I am to turn this into a 'TypeState'-ed builder. (e.g. https://doc.rust-lang.org/stable/embedded-book/design-patterns/hal/gpio.html)

We should definitely revisit when we're off the critical path again!

spacekookie (approved):
Looks good to me.

An API idea that I just had (feel free to ignore as bike-shedding :grin:) would be to add a `worker()` function on context, which returns a `WorkerBuilder` with a reference to `ctx` already filled in internally.

So the API would be:

```rust
ctx
    .worker("echoer", Echoer)
    .access_control(AllowedTransport::single(TCP))
    .start()
    .await?
```

Thoughts?

PR #2872, opened 2022-06-15
Identity updates

SanjoDeundiak:
- Move `Storage` (`AuthenticatedStorage`)
 - Get rid of `Contact`
 - Get rid of `IdentityTrait`
 - Update `Idenity` API
 - Update `Idenity` Service API


implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -88,7 +88,7 @@ async fn setup(ctx: Context, c: CreateCommand) -> anyhow::Result<()> {
     let bind = format!("0.0.0.0:{}", c.port);
     tcp.listen(&bind).await?;
 
-    let s = auth::store::mem::Store::new();
+    let s = ockam::identity::authenticated_table::mem::Store::new();
```
mrinalwadhwa: Help me understand the thought behind this? The authenticated table belongs to the authenticated service and not to one identity.

SanjoDeundiak: So, I moved this Storage trait to the ockam_identity, because during Identity Secure Channel, participated Identity needs access to the Authenticated Table. Implementations of that trait may be located anywhere, but moving them along is easier because we can use them in Identity tests

SanjoDeundiak: Also, possibly this can be moved down the dependency tree, e.g. to `ockam_core`

implementations/rust/ockam/ockam_identity/src/authenticated_table.rs
```diff
@@ -3,12 +3,11 @@ pub mod lmdb;
...
+use ockam_core::{AsyncTryClone, Result};
 
 #[async_trait]
-pub trait Storage {
-    type Error: ockam_core::compat::error::Error + Send + Sync + 'static;
-
-    async fn get(&self, id: &str, key: &str) -> Result<Option<Vec<u8>>, Self::Error>;
-    async fn set(&self, id: &str, key: String, val: Vec<u8>) -> Result<(), Self::Error>;
-    async fn del(&self, id: &str, key: &str) -> Result<(), Self::Error>;
+pub trait Storage: AsyncTryClone + Send + Sync + 'static {
```
twittner: Nothing in `Storage` seems to require these supertraits, so I recommend removing them.

implementations/rust/ockam/ockam_identity/src/change_history.rs
```diff
@@ -1,25 +1,68 @@
...
+use minicbor::{Decode, Encode};
 use ockam_core::compat::vec::Vec;
 use ockam_core::{allow, deny, Encodable, Result};
 use ockam_vault::PublicKey;
 use serde::{Deserialize, Serialize};
 
+#[derive(Debug, Clone, Encode, Decode, PartialEq)]
+#[cbor(index_only)]
+pub enum IdentityHistoryComparison {
+    #[n(1)]
```
hairyhum, line 17: Do we always start enums with `1`?

SanjoDeundiak: I would start with 1

hairyhum (dismissed):

spacekookie (approved):

PR #2874, opened 2022-06-15
feat(elixir): minicbor compatibility and helpers

polvorin:
Implements a [minicbor](https://crates.io/crates/minicbor)' compatible encoding/decoding layer on top of CBOR library.

On the encoding side, it works by pre-processing elixir data_before_ encoding them with CBOR.
On the decoding side,  it post-process CBOR' result into the expected elixir data type.

Accepted types:
```
cbor_schema =  :integer
              |:boolean
              |:string
              |:binary
              |{:enum, ..}
              |{:list, element_schema :: cbor_schema()}
              |{:map,
                key_schema :: cbor_schema(),
                value_schema :: cbor_schema()}
              |{:struct, mod, ...}
              |{:struct, ...}
```
correct typing is checked at both encoding and decoding steps.
`:string` differs from `:binary` in that the former is a proper unicode string, while the latter is a raw binary (this is enforced, and error will be throw if trying to encode (or decode) as :string something that doesn't satisfy String.valid?/1 .

Examples:
```
      iex> Ockam.TypedCBOR.to_cbor_term(:binary, <<255>>)
      %CBOR.Tag{tag: :bytes, value: <<255>>}

      iex> Ockam.TypedCBOR.from_cbor_term(:binary, Ockam.TypedCBOR.to_cbor_term(:binary, <<255>>))
      <<255>>

      iex> Ockam.TypedCBOR.to_cbor_term({:enum, [a: 0, b: 1]}, :a)
      0

      iex> Ockam.TypedCBOR.from_cbor_term({:enum, [a: 0, b: 1]}, Ockam.TypedCBOR.to_cbor_term({:enum, [a: 0, b: 1]}, :b))
      :b

      iex> Ockam.TypedCBOR.to_cbor_term({:struct, %{a1: %{key: 1, schema: :string, required: true},
      ...>                                                                           a2: %{key: 2, schema: {:enum, [a: 0, b: 1]}}}},
      ...>                    %{a1: "aa", a2: :a})
      %{1 => "aa", 2 => 0}

      iex> Ockam.TypedCBOR.from_cbor_term({:struct, %{a1: %{key: 1, schema: :string, required: true},
      ...>                                                                               a2: %{key: 2, schema: {:enum, [a: 0, b: 1]}}}},
      ...>                                                                Ockam.TypedCBOR.to_cbor_term({:struct, %{a1: %{key: 1, schema: :string, required: true},
      ...>                                         a2: %{key: 2, schema: {:enum, [a: 0, b: 1]}}}},
      ...>                              %{a1: "aa", a2: :a}))
      %{a1: "aa", a2: :a}

```
     
But the preferred way to use the library is through a [TypedStruct](https://hexdocs.pm/typed_struct/readme.html) plugin that  helps defining structs and encoder/decoder this way.   

```
          defmodule Test.TestStruct do
            use TypedStruct

            typedstruct do
              plugin Ockam.TypedCBOR.Plugin
              field :a1, String.t(), minicbor: [key: 1]
              field :a2, :a | :b, , minicbor: [key: 1, schema: {:enum, [a: 0, b:1]}]
              field :a3, list(binary()) | nil, minicbor: [key: 3])
            end
          end
```

Notes:
* :a1 is :required, as its type is `String.t()`, to make it optional it should be defined as `String.t() | nil`  (see :a3)
* enums (like :a2) must be given an explicit schema

The above definition generates a `Test.TestStruct.t()` type, a `Test.TestStruct` struct,  and the following methods:
```
        Test.TestStruct.encode(%Test.TestStruct{}) -> iodata()
        Test.TestStruct.encode_list(list(%Test.TestStruct{})) -> iodata()
        Test.TestStruct.decode(iodata()) -> %Test.TestStruct{}
        Test.TestStruct.decode_list(iodata()) -> list(%Test.TestStruct{})
        Test.TestStruct.minicbor_schema() -> the minicbor schema
```

Accept nested structs:
```
    defmodule Test.Address do
      use TypedStruct

      typedstruct do
        plugin(Ockam.TypedCBOR.Plugin)
        [...]
      end
    end

    defmodule Test.Person do
      use TypedStruct

      typedstruct do
        plugin(Ockam.TypedCBOR.Plugin)
        field(:name, String.t(), minicbor: [key: 1])

        [...]

        field(:addresses, list(Test.Address.t()),
          minicbor: [key: 2, schema: {:list, Test.Address.minicbor_schema()}]
        )
        [...]
      end
    end
```

```
    p = %Test.Person{name: "Test", age: 23, gender: :male, addresses: [%Test.Address{city: "NY"}]}
    {:ok, ^p, ""} = Test.Person.decode(Test.Person.encode(p))

```

## Checks

- [X] All commits in this Pull Request are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- [X] All commits in this Pull Request follow the Ockam [commit message convention](https://www.ockam.io/learn/how-to-guides/contributing/CONTRIBUTING#commit-messages).
- [X] I accept the Ockam Community [Code of Conduct](https://www.ockam.io/learn/how-to-guides/high-performance-team/conduct/).
- [X] I have accepted the Ockam [Contributor License Agreement](https://www.ockam.io/learn/how-to-guides/contributing/cla/) by adding my Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/build-trust/contributors/blob/main/CONTRIBUTORS.csv) file in a separate pull request to the [build-trust/contributors](https://github.com/build-trust/contributors) repository.


implementations/elixir/ockam/tcbor/lib/tcbor/encoder.ex
```diff
@@ -0,0 +1,169 @@
...
+
+      iex> TCBOR.Encoder.prep({:struct, %{a1: %{key: 1, schema: :string, required: true},
+      ...>                               a2: %{key: 2, schema: {:enum, [a: 0, b: 1]}}}},
+      ...>                    %{a2: :b})
+      ** (TCBOR.Exception) field a1 is required
+  """
+
+  alias TCBOR.Exception
+
+  def prep(:integer, val) when is_integer(val), do: val
```
hairyhum: What does `prep` and `post` suppose to mean?

implementations/elixir/ockam/tcbor/lib/tcbor/encoder.ex
```diff
@@ -0,0 +1,169 @@
+defmodule TCBOR.Encoder do
```
hairyhum: Encoder and Decoder don't really encode or decode things right? Should we rename them to something more clear?

implementations/elixir/ockam/tcbor/lib/tcbor/plugin.ex
```diff
@@ -0,0 +1,96 @@
...
+  def extract_option(val), do: type_to_spec(val)
+  def extract_options({:|, _, [opt1, opts2]}), do: [extract_option(opt1) | extract_options(opts2)]
+  def extract_options(val), do: [extract_option(val)]
+
+  @impl true
+  def after_definition(_) do
+    quote do
+      def minicbor_schema(), do: {:struct, __MODULE__, @tt_fields |> Enum.into(%{})}
+
+      def encode(%__MODULE__{} = d), do: CBOR.encode(TCBOR.Encoder.prep(minicbor_schema(), d))
```
hairyhum: This looks like it can be a separate thing from the TypedStruct plugin.

implementations/elixir/ockam/tcbor/lib/tcbor/plugin.ex
```diff
@@ -0,0 +1,96 @@
...
+        with {:ok, map, rest} <- CBOR.decode(data) do
+          {:ok, TCBOR.Decoder.post(minicbor_schema(), map), rest}
+        end
+      end
+
+      def encode_list(l), do: CBOR.encode(TCBOR.Encoder.prep({:list, minicbor_schema()}, l))
+
+      def decode_list(data) do
+        with {:ok, l, rest} <- CBOR.decode(data) do
+          {:ok, TCBOR.Decoder.post({:list, minicbor_schema()}, l), rest}
```
hairyhum: We probably don't want to crash on decode. Should we catch here? It's should be easier than returning errors in `post`.

hairyhum (changes requested):
That looks good.
Some minor notes in the review comments.

implementations/elixir/ockam/tcbor/lib/tcbor/plugin.ex
hairyhum: Maybe we don't want to crash on encode either and have separate `encode!` and `decode!` functions, which will crash.

implementations/elixir/ockam/tcbor/lib/tcbor/encoder.ex
polvorin: prep**are** and post**process**   :upside_down_face:  .  

What they do actually is

Encoder: convert from  schema + erlang term ->   an erlang term that the CBOR  library will encode according to the schema we provided,  and perform type-checks (for example,  if we said that something is a integer on the schema, we can't pass a string) 
Decoder: convert from a schema + a term as used by the CBOR library ->  an erlang term  (this for example checks that if something was defined as binary(),  we found a proper `CBOR.Tag{tag: :bytes , ..}`  on that position when decoding, and replace that with the actual binary.

I'll rename these  from_cbor_term/2   to_cbor_term/2  to make it clearer?

implementations/elixir/ockam/tcbor/lib/tcbor/encoder.ex
polvorin: right now they use the underling `CBOR` elixir library representation as intermediate.   So they don't encode to a binary in cbor format,   they encode to an erlang term that the `CBOR` library can interpret.
The same when decoding,  it doesn't decode a binary in cbor format,  rather it takes what the `CBOR` library already produced and decode and type-check it

implementations/elixir/ockam/tcbor/lib/tcbor/encoder.ex
hairyhum: That makes sense

implementations/elixir/ockam/tcbor/lib/tcbor/encoder.ex
hairyhum: If we use words like `from_cbor_term` and `to_cbor_term`, maybe we can find some better names fro the modules as well.

implementations/elixir/ockam/tcbor/lib/tcbor/plugin.ex
polvorin: @hairyhum which part? 

would it be a blocker, or could we merge this and get to that latter?

hairyhum: We could.
I was thinking of something like `minicbor_schema` in this module and `encode` and `decode` in another module.
This would let us define `minicbor_schema` manually without TypedStruct and still have `encode` and `decode` since TypedStruct is not a requirement for this TCBOR to work.

hairyhum (approved):

implementations/elixir/ockam/tcbor/lib/tcbor/plugin.ex
polvorin: that's a good idea :+1:   will tackle that on a future PR  (need to add typetag at some point too)

PR #2876, opened 2022-06-15
feat(rust): add command to create forwarders

adrianbenavides:
https://github.com/build-trust/codebase/issues/202

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -1,17 +1,10 @@
 //! Node Manager (Node Man, the superhero that we deserve)
 
-use super::types::{CreateTransport, DeleteTransport};
-use crate::error::ApiError;
-use crate::nodes::secure_channel::{
-    CreateSecureChannelListenerRequest, CreateSecureChannelResponse,
-};
-use crate::nodes::types::CreateSecureChannelRequest;
-use crate::{
-    nodes::types::{NodeStatus, TransportList, TransportMode, TransportStatus, TransportType},
-    Method, Request, Response, ResponseBuilder,
-};
 use core::convert::Infallible;
+
 use minicbor::{encode::Write, Decoder};
+
+use ockam::remote::RemoteForwarder;
 use ockam::{Address, Context, Result, Routed, TcpTransport, Worker};
 use ockam_core::compat::{boxed::Box, collections::BTreeMap, string::String};
 use ockam_core::errcode::{Kind, Origin};
@@ -19,6 +12,11 @@ use ockam_identity::authenticated_storage::mem::InMemoryStorage;
 use ockam_identity::{Identity, TrustEveryonePolicy};
 use ockam_multiaddr::MultiAddr;
 use ockam_vault::Vault;
+
+use crate::error::ApiError;
+use crate::{nodes::types::*, Method, Request, Response, ResponseBuilder, Status};
+
+use super::types::{CreateTransport, DeleteTransport};
 
 // TODO: Move to multiaddr implementation
 fn invalid_multiaddr_error() -> ockam_core::Error {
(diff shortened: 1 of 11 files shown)


spacekookie (changes requested):
This PR looks pretty good so far. One thing I noticed is the use of `embedded_node`, instead of `connect_to`.

The way I understand it the forwarder would be created in an already spawned node, not the user-facing temporary node, right?

So the flow should be:

- User spawns some node `foobar`
- User creates a forwarder on node `foobar` for route `R`

Currently the forwarder would be created in a node that is terminated again once the CLI terminates.

adrianbenavides:
Yes, that's right. I'll be pushing some changes doing that in a little while. Almost there!

adrianbenavides:
OK, now it's in a more decent state @spacekookie. I haven't had time to test it properly (will do later), but at least we can start discussing the code.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -176,6 +209,7 @@ impl NodeMan {
             // == Create a new transport
             (Post, "/node/transport") => self.add_transport(req, dec).await?.encode(enc)?,
             (Delete, "/node/transport") => self.delete_transport(req, dec).await?.encode(enc)?,
+            (Post, "/node/forwarder") => self.create_forwarder(ctx, req, dec, enc).await?,
```
spacekookie, line 302: Nitpick: please also add a `// == Title` comment before the forwarder section to keep this match block nice and clean

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -119,6 +121,9 @@ pub enum OckamSubcommand {
     #[clap(display_order = 900, help_template = HELP_TEMPLATE, name = "token")]
     GenerateEnrollmentToken(GenerateEnrollmentTokenCommand),
 
+    /// Create forwarders
+    Forwarder(ForwarderCommand),
```
spacekookie:
```suggestion
    /// Create forwarders
    #[clap(display_order = 900, help_template = HELP_TEMPLATE)]
    Forwarder(ForwarderCommand),
```

You forgot the `clap` attribute invocation here. I _think_ it's important anyway :sweat_smile:

implementations/rust/ockam/ockam_command/src/forwarder/create.rs
```diff
@@ -0,0 +1,82 @@
...
+    #[clap(default_value_t = String::from("forwarder"))]
+    alias: String,
+}
+
+impl CreateCommand {
+    pub fn run(cfg: &mut OckamConfig, cmd: CreateCommand) {
+        if !cfg.has_node(&cmd.node_name) {
+            let node_cmd = node::CreateCommand::new(&cmd.node_name, true, cmd.port);
+            node::CreateCommand::run(cfg, node_cmd);
+        }
```
spacekookie: The current pattern for this in other commands is to use `cfg.select_node(...)`, which also allows for command-local overrides of which node should be targeted. Would this work here?

Furthermore, I think we should avoid having any magic fail-overs into other commands (i.e. going to create a node if it doesn't exist)

spacekookie (commented):
Excellent, thank you! I left you some feedback (mostly nitpicks lol). Almost there I'd say :rocket:

implementations/rust/ockam/ockam_command/src/forwarder/create.rs
adrianbenavides: OK got it, makes more sense. We'll now have every command following the same strategy of `try find node or fail`.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
adrianbenavides: Done

implementations/rust/ockam/ockam_command/src/lib.rs
adrianbenavides: Yeah, now that you mention it I think it renders the help template without the attribute. I'll open a separate PR removing them from subcommands because I'm pretty sure it's inherited from the top-level command.

In any case, I've added the attribute for this PR.

mrinalwadhwa:
@adrianbenavides could you squash the two commits in this PR so we don't have a commit with message `feat(rust): nitpicks` ... these messages end up in our release changelogs and this message would not make any sense in the change log https://github.com/build-trust/ockam/releases/tag/ockam_v0.62.0

adrianbenavides:
> @adrianbenavides could you squash the two commits in this PR so we don't have a commit with message `feat(rust): nitpicks` ... these messages end up in our release changelogs and this message would not make any sense in the change log https://github.com/build-trust/ockam/releases/tag/ockam_v0.62.0

I didn't have that in mind, sorry. It's fixed now!

adrianbenavides:
A couple of TODO's left to fix before leaving the draft status:
- Ephemeral vs Static forwarder: new command argument to define whether to use a static or dynamic forwarder
- trace logs should follow the structure from `implementations/rust/ockam/ockam_api/src/nodes/service.rs:169` as
```
trace! {
    target: "ockam::nodeman::service",
    id     = %req.id(),
    method = ?req.method(),
    path   = %req.path(),
    body   = %req.has_body(),
    "request"
}
```

adrianbenavides:
Ready for the final review @spacekookie @mrinalwadhwa

spacekookie (dismissed):

antoinevg (approved):

PR #2877, opened 2022-06-15
feat(elixir): authorization control for services

hairyhum:
To control message authorization for services on start, added `with_config` authorization
rule, which will check the `:authorization` option in the worker state to get the
rules.

This lets us do something like: `MyWorker.create(authorization: :is_local)`

Added `Ockam.Services.AuthorizationConfig` with `secure_channel/0` and `local/0` predefined
authorization configs to use in services

## Checks

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm.ex
@@ -69,8 +69,10 @@ defmodule Ockam.SecureChannel.EncryptedTransportProtocol.AeadAesGcm do
     {:ok, encrypted, ""} = :bare.decode(payload, :data)
 
     with {:ok, decrypted, data} <- decrypt(encrypted, data),
-         {:ok, decoded} <- Wire.decode(decrypted) do
-      message = Message.trace(decoded, data.plaintext_address)
+         {:ok, decoded} <- Wire.decode(decrypted, :secure_channel) do
+      message =
+        decoded
+        |> Message.trace(data.plaintext_address)
 
       Router.route(message)
 
--- a/implementations/elixir/ockam/ockam_services/lib/services/authorization_config.ex
+++ b/implementations/elixir/ockam/ockam_services/lib/services/authorization_config.ex
@@ -1,0 +1,12 @@
+defmodule Ockam.Services.AuthorizationConfig do
+  @moduledoc """
+  Predefined configs for services authorization
+  """
+  def secure_channel() do
+    [:to_my_address, :is_secure]
+  end
+
+  def local() do
+    [:to_my_address, :is_local]
+  end
+end
--- a/implementations/elixir/ockam/ockam/lib/ockam/message.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/message.ex
@@ -79,6 +79,10 @@ defmodule Ockam.Message do
     %{message | payload: payload}
   end
 
+  def set_local_metadata(%Message{} = message, metadata) when is_map(metadata) do
+    %{message | local_metadata: metadata}
+  end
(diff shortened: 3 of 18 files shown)


hairyhum:
Follow up to #2838

hairyhum:
Example usage can be found in https://github.com/build-trust/ockam/pull/2877/files#diff-32f8de46331f9713fa74a422982b2c7e51fc344ccd44b13c732027585c5425f8L53

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm.ex
```diff
@@ -70,7 +70,10 @@ defmodule Ockam.SecureChannel.EncryptedTransportProtocol.AeadAesGcm do
 
     with {:ok, decrypted, data} <- decrypt(encrypted, data),
          {:ok, decoded} <- Wire.decode(decrypted) do
-      message = Message.trace(decoded, data.plaintext_address)
+      message =
+        decoded
+        |> Message.trace(data.plaintext_address)
+        |> Message.put_local_metadata(:transport, :secure_channel)
```
polvorin: this looks a bit different than the other  `:transport`  values  (in the sense that `:tcp` and `:udp` do seems similar,  but `:secure_channel` is something that happens after that?). 
Would it be better if instead of overriding the `:transport`  it set another attribute like `:identity`? .  And the authorization that enforces messages  come from a secure channel then check the presence of this attribute   (latter can check for specific identities as well)

hairyhum: This field is used to specify the place where message was unwrapped. Because wrapping a message destroys the local metadata, every time message is wrapped and unwrapped this field should be added to specify the entry point to the local scope.
For example in pipe wrappers: https://github.com/build-trust/ockam/pull/2877/files#diff-9fa81d2fe4269b5dcba95fcff4b59b4323e92e9e55022d53fbf6efba839449c4R25

Maybe we could use a different name for that or even have a special value for "local" messages (instead of no value). I was thinking of maybe using `:channel` or something like that. WDYT?

polvorin: :ok_hand:  fair enough
for me `transport` is a bit unclear as  a secure channel (I _think_ ?) is ultimately a worker, while a tcp `transport` is not?   But also I'm not too familiar with the nomenclature,  I think we can get back to this latter if needed for clarity

polvorin (dismissed):

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm.ex
hairyhum: Make it so Ockam.Wire always adds `:channel` metadata to messages,
while Router adds `source: :local` to messages without `:source` and `:channel` set.
This means that all messages not decoded from a channel or transport will have `source: local`
and all messages decoded will have `source: :channel, channel: type`

polvorin (dismissed):

implementations/elixir/ockam/ockam_services/lib/services/authorization_config.ex
```diff
@@ -0,0 +1,12 @@
+defmodule Ockam.Services.AuthorizationConfig do
```
mrinalwadhwa, line 1: This can possibly the beginning of a policies layer and maybe this module could be called `AuthorizationPolicy`

mrinalwadhwa (approved):

mrinalwadhwa (approved):

mrinalwadhwa:
@Mergifyio refresh

PR #2879, opened 2022-06-16
feat(rust): implement attribute-based access control for message-flow authorization

antoinevg:
This PR extends Ockam message flow authorization to support Attribute Based Access Control (ABAC).

The implementation currently lives in the `ockam` crate under the [`abac`](https://github.com/build-trust/ockam/tree/antoinevg/abac/implementations/rust/ockam/ockam/src/abac) module.

## PR Scope

* This PR contains core ABAC API's required for clients making ABAC authorization requests.
* This PR contains core ABAC API's required to implement ABAC services making authorization decisions.
* This PR does not contain a production-ready implementation of an ABAC service for making authorization decisions. Currently the [`abac::mem::Memory`](https://github.com/build-trust/ockam/blob/antoinevg/abac/implementations/rust/ockam/ockam/src/abac/mem.rs) implementation can be used for the purposes of client development and testing.
* This PR does not contain the final implementation of the ABAC policy definition DSL. We can discuss this topic in the related GitHub issue.

## Discussion

Where should the `abac` module live?

Two thoughts:

1. Move `ockam::abac` module into an `ockam_abac` crate:
   - Depends on `ockam_identity::Identifier`, can we move this to `ockam_core` as well?

OR:

2. Split the `ockam::abac` module up a little to support useage in crates that do not have the `ockam` module as a dependency:
    - Move `abac::traits` to `ockam_core::abac::traits`
    - Move `abac::types` to `ockam_core::abac::types`
    - Depends on `ockam_identity::Identifier`, can we move this to `ockam_core` as well?

### Making ABAC authorization requests

Three things are required to use ABAC at the authorization request point:

1. An identity identifier that represents the identity requesting access to a given resource and action (e.g. `ockam_identity::IdentityIdentifer`)
1. A request triple: `Subject(identity_identifier)`, `Resource(path)`, `Action(method)`
1. A reference (e.g. Rust object reference, worker address) to some ABAC implementation supporting the `AbacAuthorization::is_authorized(subject, resource, action)` trait.

For example:

https://github.com/build-trust/ockam/blob/128cc2fcd99036923c2921ec1ea48a1289a004c7/implementations/rust/ockam/ockam_examples/example_projects/access_control/examples/02-abac-in-place.rs#L53-L90

### Implementing ABAC services

ABAC service architecture decomposes into three traits that can be implemented as a single monolithic service or as separate, loosely coupled services that may also wrap 3rd-party sources:

- `AbacAttributeStorage` - an interface for retrieving attributes from the authenticated table or other attribute source.
- `AbacPolicyStorage` - an interface for retrieving ABAC policies from our own local policy storage or some other source.
- `AbacAuthorization` - the "decision point" interface: `is_authorized(subject, resource, action)`

For example:

https://github.com/build-trust/ockam/blob/128cc2fcd99036923c2921ec1ea48a1289a004c7/implementations/rust/ockam/ockam_examples/example_projects/access_control/examples/02-abac-in-place.rs#L96-L152

### Examples

* Implement ABAC service as Rust objects: [02-abac-in-place.rs](https://github.com/build-trust/ockam/blob/antoinevg/abac/implementations/rust/ockam/ockam_examples/example_projects/access_control/examples/02-abac-in-place.rs)
* Implement ABAC service as Ockam workers: [03-abac-workers.rs](https://github.com/build-trust/ockam/blob/antoinevg/abac/implementations/rust/ockam/ockam_examples/example_projects/access_control/examples/03-abac-workers.rs)

## Subjects, Resources and Actions

An ABAC authorization request consists of three parameters:

1. `Subject`: The entity making the request. Expressed as a user id such as `ockam_identity::IdentityIdentifier`
2. `Resource`: The name of the resource that the subject wishes to access. Expressed as an api path such as: `"/nodes"`
3. `Action`: The action the subject wishes to perform on the resource. Expressed as a HTTP method such as `Method::Get` or `Method::Post`.

## Policy definition

The abac module provides an implementation (thank you @twittner!) of a simple DSL for defining ABAC policies.

### Attributes

Attributes are associated with a `Subject`, `Resource` or `Action`.

An attribute consists of a simple tuple expressing a name/value pair:

```rust
pub type Attribute = (String, Value);
```

For example:

```rust
[
    ("space", abac::string("green")),
    ("enrollment_token", abac::int(0x1234)),
]
```

Defines two attributes:

* an attribute named `"space"` with a Value of `"green"`
* an attribute named `"enrollment_token"` with a Value of `0x1234`.

Attributes can be associated with a subject, resource or action by reference:

```
    let mut resource = Resource::from("/projects");
    resource.extend([("space", abac::string("some_customer_space"))]);
```

or by value:

```
    let subject = Subject::from(identity_identifier).with_attributes([("has_mustache", abac::bool(true))]);
```

### Policies

Policies are associated with a `Resource` and consist of an `Action` and a `Conditional`.

The `Conditional` defines the rules to be evaluated before the given `Action` will be authorized on the associated `Resource`.

For example:

```rust
let policy = abac::eq("has_mustache", abac::bool(true));
```

This defines a new policy Conditional which will check a subject's attributes for one with the name "has_mustache" and the value true.

Conditional operators may also be chained and nested to create new conditional operators.

For example:

```rust
let eq_space_green = abac::eq("space", abac::string("green"));
let has_enrollment_token = abac::eq("enrollment_token", abac::bool(true));

let policy = eq_space_green.and(has_enrollment_token);
```

This defines a new policy condition which will only authorize a subject with a set of attributes that include both a `"space : green"` attribute and an `"enrollment_token : true"` attribute.

Finally, policies are associated with `Resources` and `Actions` by making a call to an ABAC service implementing `AbacPolicyStorage` using:

```rust
abac_backend.set_policy(
    Resource::from("/projects").with_attributes(["space", abac::string("green")]),
    Action::from(Method::Post),
    eq_space_green.and(has_enrollment_token),
)
.await?;
```

## Schema reference

The ABAC logical schema has the form:

* Attributes: `Subject`  ||--|{ `Attribute` ||--|| (`key`, `Value`)
* Policies: `Resource` ||--|{ `policy` ||--|| (`Action`, `Conditional`)

([schema syntax](https://mermaid-js.github.io/mermaid/#/entityRelationshipDiagram?id=syntax))


implementations/rust/ockam/ockam/src/abac/mem.rs
```diff
@@ -0,0 +1,131 @@
+//! In-memory implementation of the `Abac type`.
```
antoinevg: FIXME: ```//! In-memory implementation of the [`Abac`] trait.```

implementations/rust/ockam/ockam/src/abac/mem.rs
```diff
@@ -0,0 +1,131 @@
...
+impl Memory {
+    /// Create a new `Memory` ABAC backend
+    pub fn new() -> Self {
+        Self {
+            inner: Arc::new(RwLock::new(Inner::new())),
+        }
+    }
+}
+
+/// Inner implements the [`Abac`] trait for the [`Memory`] ABAC backend.
```
antoinevg: FIXME: ```/// `Inner` implements the [`Abac`] trait for the [`Memory`] ABAC backend.```

implementations/rust/ockam/ockam/src/abac/metadata.rs
```diff
@@ -0,0 +1,49 @@
+use super::{error::AbacError, AbacLocalInfo, Parameters};
+use crate::OckamMessage;
+use ockam_core::{Decodable, Encodable, Message, Result};
+
+use serde::{Deserialize, Serialize};
+
+/// Abac Metadata unique identifier.
+pub const ABAC_METADATA_IDENTIFIER: &str = "ABAC_METADATA_IDENTIFIER";
+
+/// Abac Metadata used for LocalMessage
```
antoinevg: FIXME: ```// Abac [`Metadata`] used for [`OckamMessage`]```

implementations/rust/ockam/ockam/src/abac/access_control.rs
```diff
@@ -0,0 +1,60 @@
...
+        write!(f, "AttributeBasedAccessControl")
+    }
+}
+
+impl<A> AttributeBasedAccessControl<A> {
+    /// Create a new `AttributeBasedAccessControl` with the given
+    /// [`AbacAuthorization`] backend.
+    pub fn new(backend: A) -> Self
+    where
+        A: AbacAuthorization + Debug + Send + Sync + 'static,
```
SanjoDeundiak: I think `AbacAuthorization` implies `Send + Sync + 'static`

implementations/rust/ockam/ockam/src/abac/access_control.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+#[async_trait]
+impl<A> AccessControl for AttributeBasedAccessControl<A>
+where
+    A: AbacAuthorization + Debug + Send + Sync + 'static,
+{
+    async fn is_authorized(&self, local_msg: &LocalMessage) -> Result<bool> {
+        // pull request triple out of LocalMessage's LocalInfo
+        let local_info = AbacLocalInfo::find_info(local_msg)?;
```
SanjoDeundiak: Didn't we agree to not include any ABAC-related data to LocalInfo?

implementations/rust/ockam/ockam/src/abac/metadata.rs
```diff
@@ -0,0 +1,55 @@
...
+use ockam_core::{Decodable, Encodable, Message, Result};
+
+use serde::{Deserialize, Serialize};
+
+/// ABAC [`Metadata`] unique identifier.
+pub const ABAC_METADATA_IDENTIFIER: &str = "ABAC_METADATA_IDENTIFIER";
+
+/// ABAC [`Metadata`] used for [`OckamMessage`]
+#[derive(Debug, Serialize, Deserialize)]
+pub struct AbacMetadata {
```
SanjoDeundiak: What is the difference between `AbacMetadata` and `AbacLocalInfo`?

implementations/rust/ockam/ockam/src/abac/mod.rs
```diff
@@ -0,0 +1,21 @@
+//! Attribute Based Access Control
+
+pub mod access_control;
```
SanjoDeundiak: Maybe ABAC already deserves a separate crate

implementations/rust/ockam/ockam/src/abac/workers.rs
```diff
@@ -0,0 +1,100 @@
...
+    abac::{AbacLocalInfo, AbacMetadata, ABAC_LOCAL_INFO_IDENTIFIER},
+    Any, Context, OckamMessage, Result, Route, Routed, Worker,
+};
+use ockam_core::compat::boxed::Box;
+use ockam_core::{Decodable, LocalMessage};
+
+/// An `AbacUnwrapperWorker` wraps any [`AbacLocalInfo`] attached to a
+/// [`LocalMessage`] in `OckamMessage` [`Metadata`] so that it can be
+/// recovered after being sent to another Ockam node.
+pub struct AbacWrapperWorker {
```
SanjoDeundiak: I thought we agreed to remove it

implementations/rust/ockam/ockam_examples/example_projects/access_control/.gitignore
```diff
@@ -0,0 +1,17 @@
+**/*.rs.bk
```
SanjoDeundiak: Isn't root `.gitignore` good enough?

implementations/rust/ockam/ockam_examples/example_projects/access_control/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+license = "Apache-2.0"
+publish = false
+rust-version = "1.56.0"
+
+[features]
+default = ["std"]
+
+# Feature (enabled by default): "std" enables functionality expected to
+# be available on a standard platform.
+std = ["ockam/default"]
```
SanjoDeundiak, line 15: Should it be `ockam/std`?

implementations/rust/ockam/ockam_examples/example_projects/access_control/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+
+# Feature: "alloc" enables support for heap allocation on "no_std"
+# platforms, requires nightly.
+alloc = ["ockam/alloc"]
+
+[dependencies]
+ockam = { path = "../../../ockam", default_features = false, features = [
+    "software_vault",
+] }
+ockam_core = { path = "../../../ockam_core" }
```
SanjoDeundiak, line 29: Ideally we want to avoid such dependency, what is needed in this example that couldn't be imported through `ockam`?

implementations/rust/ockam/ockam_examples/example_projects/access_control/examples/02-abac-in-place.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    async fn is_authorized(
+        &mut self,
+        _ctx: &mut Self::Context,
+        msg: Routed<Self::Message>,
+    ) -> Result<bool> {
+        // Grab the customer's IdentityIdentifer from
+        // IdentitySecureChannelLocalInfo or any other place you may
+        // have it lying around.
+        let local_msg = msg.into_local_message();
```
SanjoDeundiak, line 75: Would be great to see example of this working together with a Worker that serves in terms of Request/Response API. Because when resource&action are const for specific workers, it makes sense to move this logic to AccessControl struct

implementations/rust/ockam/ockam_examples/example_projects/access_control/src/authenticated_table_worker.rs
```diff
@@ -0,0 +1,54 @@
+use ockam::abac::{AbacAttributeStorage, Subject};
+use ockam::{Context, Message, Result, Routed, Worker};
+use serde::{Deserialize, Serialize};
+
+/// A simple authenticated table service which serves subject
+/// attributes to a requester
+pub struct AuthenticatedTableWorker<B> {
```
SanjoDeundiak, line 7: Don't we already have such service?

implementations/rust/ockam/ockam/src/abac/access_control.rs
antoinevg: Whoops. Missed one.

implementations/rust/ockam/ockam/src/abac/access_control.rs
antoinevg: We don't have to use it, but I can think of use-cases where it could be useful for customers. If you'd prefer I can move it to the ockam_examples/ section.

implementations/rust/ockam/ockam/src/abac/metadata.rs
antoinevg: `AbacMetadata` for tunneling ABAC triples between nodes. `AbacLocalInfo` for tunneling ABAC triples in-node. Again, shout if you think it belongs in example_projects/ rather!

implementations/rust/ockam/ockam/src/abac/mod.rs
antoinevg: I'm leaning towards a separate crate as well. Unless anyone thinks different by the end of the week I think I'm going to do that.

implementations/rust/ockam/ockam/src/abac/workers.rs
antoinevg: We don't have to use it, but I can think of use-cases where it could be useful for customers. If you'd prefer I can move it to the ockam_examples/ section.

implementations/rust/ockam/ockam_examples/example_projects/access_control/.gitignore
antoinevg: Whoops, that should only be excluding `target/`.

implementations/rust/ockam/ockam_examples/example_projects/access_control/Cargo.toml
antoinevg: Indeed it should.

implementations/rust/ockam/ockam_examples/example_projects/access_control/Cargo.toml
antoinevg: `02-abac-in-place.rs` needs access to:

```
use ockam_core::{
    async_trait,
    compat::{boxed::Box, sync::Arc},
};
```

implementations/rust/ockam/ockam_examples/example_projects/access_control/examples/02-abac-in-place.rs
antoinevg: You've said this before but I think I might be missing some info around how you need ABAC to be integrating with the Request/Response API in Rust. 

Could you maybe come up with a small example to explain what you mean or expand a bit more on how you'd like to use this logic inside an AccessControl struct?

implementations/rust/ockam/ockam_examples/example_projects/access_control/src/authenticated_table_worker.rs
antoinevg: This is an example of how to implement `AbacAttributeStorage` not how to use the authenticated table.  The purpose is to show how you could integrate the authenticated table service or even some other attribute source with ABAC.

Remember, eventually ABAC is going to be present in many parts of the system. The current use-case for `ockam_command` is only one so it doesn't make sense to specialize it just for that case. Things need to be pluggable.

implementations/rust/ockam/ockam_examples/example_projects/access_control/Cargo.toml
antoinevg: Ah nope, it's `ockam/default` because this example project uses the default ockam feature set ("std", "ockam_transport_tcp", "software_vault", "software_vault_storage", "noise_xx")

implementations/rust/ockam/ockam/src/abac/access_control.rs
antoinevg: Moved to `ockam_examples/example_projects/access_control`.

implementations/rust/ockam/ockam/src/abac/mod.rs
antoinevg: Moved to `ockam_examples/example_projects/access_control`.

implementations/rust/ockam/ockam/src/abac/workers.rs
antoinevg: Moved to `ockam_examples/example_projects/access_control`.

implementations/rust/ockam/ockam_examples/example_projects/access_control/examples/02-abac-in-place.rs
antoinevg: Let's merge this PR so long and then re-sync on this when I'm done with the Elixir-side.

implementations/rust/ockam/ockam/src/abac/metadata.rs
antoinevg: Moved to `ockam_examples/example_projects/access_control`.

implementations/rust/ockam/ockam/src/abac/mod.rs
antoinevg: I also took a look at moving `IdentityIdentifier` to `ockam_core` but things got hairy very quickly.

On the balance of things I think we're better off giving it it's own crate.

I've moved the implementation to `ockam_abac`. It's pretty minimal after moving the bits for the `AccessControl` out but it will probably start expanding rapidly once we start implementing the service for Elixir and integrating 3rd party sources etc.

mrinalwadhwa (approved):

PR #2887, opened 2022-06-20
chore(rust): update ockam_transport_ble deps

twittner:
Addresses https://github.com/build-trust/ockam/security/dependabot/8


antoinevg:
This PR is likely to break embedded builds and may require more changes than just bumping the deps.

One option is to leave it open until there's time to migrate the code to the new versions of `atsame54_xpro` and `stm32f4xx-hal`.

Some background on why this vulnerability is not an issue for us here:

https://github.com/build-trust/ockam/pull/2542#issuecomment-1055630787

PR #2894, opened 2022-06-22
refactor(rust): strategy to enable/disable logs in ockam_command

adrianbenavides:
Previously, we had the arguments `verbose`, `quiet`, `spawned` + the OCKAM_LOG env variable to
control how logs were enabled. The OCKAM_LOG was the highest priority parameter, making it
impossible to override with any of mentioned command arguments.
Also, the `quiet` argument was ignored.

Now, the `spawned` argument has been removed and the priority has been reversed. That is, if
`quiet` is set, logs are not enabled. If `verbose` is set, its value will be used to define
the log level; if not, it will try to read the log level from the OCKAM_LOG env variable.
Logs won't be enabled if both `verbose` and OCKAM_LOG are not set.

Some examples:
```bash
$ OCKAM_LOG=debug ockam node create -q # Logs not enabled because `quiet` is set
$ OCKAM_LOG=debug ockam node create -vvv # Logs enabled to INFO because of `verbose` preference
$ OCKAM_LOG=debug ockam node create # Logs enabled to DEBUG because of `OCKAM_LOG` value
$ ockam node create # Logs not enabled because neither `verbose` nor `OCKAM_LOG` are set
```

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -42,12 +42,9 @@ impl CreateCommand {
                 .open(&cfg.log_path.join(format!("{}.log", command.node_name)))
                 .expect("failed to open log path");
 
-            // Override the log level for spawned nodes because debug
-            // logs are significantly more useful for debugging
-            std::env::set_var("OCKAM_LOG", "debug");
             let child = Command::new(ockam)
                 .args([
-                    "--spawned",
+                    "-vv", // Enable logs at DEBUG level
                     "node",
                     "create",
                     "--port",
--- a/implementations/rust/ockam/ockam_command/src/util/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/util/mod.rs
@@ -140,24 +140,24 @@ pub fn setup_logging(verbose: u8) {
         "ockam_vault_sync_core",
     ];
     let builder = EnvFilter::builder();
-    let filter = match env::var("OCKAM_LOG") {
-        Ok(s) if !s.is_empty() => builder.with_env_var("OCKAM_LOG").from_env_lossy(),
-        _ => match verbose {
-            0 => builder
-                .with_default_directive(LevelFilter::WARN.into())
-                .parse_lossy(ockam_crates.map(|c| format!("{c}=info")).join(",")),
-            1 => builder
-                .with_default_directive(LevelFilter::INFO.into())
-                .parse_lossy(""),
-            2 => builder
-                .with_default_directive(LevelFilter::DEBUG.into())
-                .parse_lossy(""),
-            _ => builder
-                .with_default_directive(LevelFilter::TRACE.into())
-                .parse_lossy(""),
+    // If `verbose` is not set, try to read the log level from the OCKAM_LOG env variable.
+    // If both `verbose` and OCKAM_LOG are not set, logging will not be enabled.
+    // Otherwise, use `verbose` to define the log level.
+    let filter = match verbose {
+        0 => match env::var("OCKAM_LOG") {
+            Ok(s) if !s.is_empty() => builder.with_env_var("OCKAM_LOG").from_env_lossy(),
+            _ => return,
         },
+        1 => builder
+            .with_default_directive(LevelFilter::INFO.into())
+            .parse_lossy(ockam_crates.map(|c| format!("{c}=info")).join(",")),
+        2 => builder
+            .with_default_directive(LevelFilter::DEBUG.into())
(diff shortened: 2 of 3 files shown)


implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -128,34 +128,30 @@ pub(crate) fn multiaddr_to_route(ma: &MultiAddr) -> Option<Route> {
...
+            .parse_lossy(""),
+        3 => builder
+            .with_default_directive(LevelFilter::INFO.into())
+            .parse_lossy(""),
+        4 => builder
+            .with_default_directive(LevelFilter::DEBUG.into())
+            .parse_lossy(""),
+        _ => builder
+            .with_default_directive(LevelFilter::TRACE.into())
+            .parse_lossy(""),
```
spacekookie: I think a single `-v` should be enough to enable `INFO` level logging. WARN and ERROR events by themselves are almost useless to see what is going on and INFO is low-frequency enough that it can still be considered "mostly quiet"

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -41,12 +41,9 @@ impl CreateCommand {
                 .open(&cfg.log_path.join(format!("{}.log", command.node_name)))
                 .expect("failed to open log path");
 
-            // Override the log level for spawned nodes because debug
-            // logs are significantly more useful for debugging
-            std::env::set_var("OCKAM_LOG", "debug");
             let child = Command::new(ockam)
                 .args([
-                    "--spawned",
+                    "-vvv", // Enable logs at INFO level
```
spacekookie: The default log level for spawned nodes should be `DEBUG`

implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -128,34 +128,30 @@ pub(crate) fn multiaddr_to_route(ma: &MultiAddr) -> Option<Route> {
...
-        "ockam",
-        "ockam_node",
-        "ockam_core",
-        "ockam_command",
-        "ockam_identity",
-        "ockam_channel",
-        "ockam_transport_tcp",
-        "ockam_vault",
-        "ockam_vault_sync_core",
-    ];
```
spacekookie: I think we need to keep this list because otherwise logging will be also applied to dependency crates which we really don't care about (mostly `tokio` and `mio` are common culprits for spamming debug and trace lines)

spacekookie (changes requested):

implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -128,34 +128,30 @@ pub(crate) fn multiaddr_to_route(ma: &MultiAddr) -> Option<Route> {
...
-                .parse_lossy(""),
-            _ => builder
-                .with_default_directive(LevelFilter::TRACE.into())
-                .parse_lossy(""),
+    // If `verbose` is not set, try to read the log level from the OCKAM_LOG env variable.
+    // If both `verbose` and OCKAM_LOG are not set, logging will not be enabled.
+    // Otherwise, use `verbose` to define the log level.
+    let filter = match verbose {
+        0 => match env::var("OCKAM_LOG") {
+            Ok(s) if !s.is_empty() => builder.with_env_var("OCKAM_LOG").from_env_lossy(),
```
adrianbenavides, line 148: Regarding this https://github.com/build-trust/ockam/pull/2894/files#r903498666

I guess it makes sense to also add the `ockam_dirs` here. Otherwise, when setting the OCKAM_LOG variable, it should include the list rather than only the log level. (@spacekookie)

spacekookie: I'm confused what you mean with `ockam_dirs` here. Anyway, I don't think we really need to expose per-crate overrides of logging via `OCKAM_LOG` but it might be nice to be able to do that. As long as `env OCKAM_LOG=trace ockam ...` (so just one level for all ockam crates) is still possible

adrianbenavides: `ockam_crates` sorry (the variable you mentioned in the linked comment). OK, I'll leave it as it is then.

adrianbenavides:
All comments reviewed @spacekookie, thanks for your feedback!

spacekookie (dismissed):
Great, thanks for implementing this so quickly! :sparkles: :rocket:

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #2916, opened 2022-07-01
fix(rust): correctly setup state directory for foreground nodes

spacekookie:
This PR fixes the issue where foreground nodes failed to start because the state directory wasn't properly set up beforehand.


SanjoDeundiak (approved):

antoinevg (approved):

metaclips:
@mergifyio refresh

PR #2919, opened 2022-07-02
refactor(rust): use same variable names on some ockam_command commands

adrianbenavides:
The `ockam_command` crate is currently under active development by several people and we have some commands that differ in structure and common variable naming.

This PR adds some minor changes to make commands that are more stable look as similar as possible.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/enroll/enrollment_token_generate.rs
+++ b/implementations/rust/ockam/ockam_command/src/enroll/enrollment_token_generate.rs
@@ -1,13 +1,13 @@
-use anyhow::{anyhow, Context};
+use anyhow::anyhow;
 use clap::Args;
 
-use crate::IdentityOpts;
 use ockam::TcpTransport;
 use ockam_api::auth::types::Attributes;
 use ockam_multiaddr::MultiAddr;
 
 use crate::old::identity::load_or_create_identity;
 use crate::util::embedded_node;
+use crate::IdentityOpts;
 
 #[derive(Clone, Debug, Args)]
 pub struct GenerateEnrollmentTokenCommand {
@@ -21,17 +21,17 @@ pub struct GenerateEnrollmentTokenCommand {
     #[clap(display_order = 1002, long, default_value = "default")]
     identity: String,
 
-    /// Comma-separated list of attributes
-    #[clap(last = true, required = true)]
-    attributes: Vec<String>,
+    /// Attributes (use '=' to separate key from value)
+    #[clap(value_delimiter('='), last = true, required = true)]
+    attrs: Vec<String>,
 
     #[clap(flatten)]
     identity_opts: IdentityOpts,
 }
 
 impl GenerateEnrollmentTokenCommand {
-    pub fn run(command: GenerateEnrollmentTokenCommand) {
-        embedded_node(generate, command);
+    pub fn run(cmd: GenerateEnrollmentTokenCommand) {
+        embedded_node(generate, cmd);
     }
 }
 
@@ -47,20 +47,11 @@ async fn generate(
         .ok_or_else(|| anyhow!("failed to parse address: {}", cmd.address))?;
 
     let mut attributes = Attributes::new();
-    for kv in &cmd.attributes {
-        let mut s = kv.split(',');
-        let k = s.next().context(format!(
-            "failed to parse key from pair: {kv:?}. Expected a \"key,value\" pair."
-        ))?;
-        let v = s
-            .next()
-            .context(format!("no value found on pair: {kv:?}"))?;
-        if k.is_empty() {
-            anyhow::bail!("attribute name can't be empty at pair {kv:?}")
-        } else if v.is_empty() {
-            anyhow::bail!("attribute value can't be empty at pair {kv:?}")
-        } else {
+    for entry in cmd.attrs.chunks(2) {
+        if let [k, v] = entry {
             attributes.put(k, v.as_bytes());
+        } else {
+            return Err(anyhow!("{entry:?} is not a key-value pair"));
         }
     }
 
--- a/implementations/rust/ockam/ockam_command/src/message/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/message/mod.rs
@@ -1,8 +1,8 @@
-mod send;
-
 use crate::HELP_TEMPLATE;
 use clap::{Args, Subcommand};
 use send::SendCommand;
+
+mod send;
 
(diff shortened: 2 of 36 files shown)


implementations/rust/ockam/ockam_command/src/enroll/enrollment_token_generate.rs
```diff
@@ -21,17 +21,17 @@ pub struct GenerateEnrollmentTokenCommand {
     #[clap(display_order = 1002, long, default_value = "default")]
     identity: String,
 
-    /// Comma-separated list of attributes
-    #[clap(last = true, required = true)]
-    attributes: Vec<String>,
+    /// Attributes (use '=' to separate key from value)
+    #[clap(value_delimiter('='), last = true, required = true)]
```
spacekookie, line 25: I thought this was supported by default? i.e. that you could either write `--attrs foo` or `--attrs=foo`? I feel like we should support both

implementations/rust/ockam/ockam_command/src/message/mod.rs
```diff
@@ -1,9 +1,11 @@
-mod send;
-
-use crate::HELP_TEMPLATE;
 use clap::{Args, Subcommand};
+
 use send::SendCommand;
 
+use crate::HELP_TEMPLATE;
```
spacekookie: Style wise I'd collapse the spaces between these lines to make the section more concise

implementations/rust/ockam/ockam_command/src/transport/create.rs
```diff
@@ -55,7 +55,7 @@ impl CreateCommand {
         // still bad and should be fixed
         let node = command.api_node.unwrap_or_else(|| cfg.get_api_node());
         match command.create_subcommand {
-            CreateTypeCommand::TcpConnector { address } => {
+            CreateTypeCommand::TcpConnector { addr: address } => {
                 cfg.add_transport(&node, false, true, address)
             }
```
spacekookie: I think re-mapping the argument name here is not really necessary

implementations/rust/ockam/ockam_command/src/util/api.rs
```diff
@@ -27,7 +27,9 @@ pub(crate) fn query_transports() -> Result<Vec<u8>> {
 pub(crate) fn create_transport(cmd: &crate::transport::CreateCommand) -> Result<Vec<u8>> {
     // FIXME: this should not rely on CreateCommand internals!
     let (tt, addr) = match &cmd.create_subcommand {
-        transport::CreateTypeCommand::TcpConnector { address } => (TransportMode::Connect, address),
+        transport::CreateTypeCommand::TcpConnector { addr: address } => {
+            (TransportMode::Connect, address)
+        }
```
spacekookie: Same here

implementations/rust/ockam/ockam_command/src/util/config/mod.rs
```diff
@@ -150,15 +140,6 @@ Otherwise your OS or OS configuration may not be supported!",
...
 
-    /// Check whether another node has been registered with this API
-    /// port.  This doesn't catch all port collision errors, but will
-    /// get us most of the way there in terms of starting a new node.
-    pub fn port_is_used(&self, port: u16) -> bool {
-        let inner = self.inner.read().unwrap();
-
-        inner.nodes.iter().any(|(_, n)| n.port == port)
-    }
-
```
spacekookie: Please don't remove this function even though it's currently not used. It is still needed in the future

implementations/rust/ockam/ockam_command/src/message/mod.rs
adrianbenavides: That's because of the clion's default rules for rustfmt. It separates imports in different sections (local crates, external deps, etc). For some reason clion has this option enabled by default: https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#StdExternalCrate%5C%3A

Maybe we should have a `.rustfmt.toml` to ensure we all apply the same rules.

implementations/rust/ockam/ockam_command/src/transport/create.rs
adrianbenavides: True, I didn't catch this change after renaming the struct field.

implementations/rust/ockam/ockam_command/src/util/config/mod.rs
adrianbenavides: OK, I'll revert the visibility change of `OckamCommand` to avoid the unused warning.

implementations/rust/ockam/ockam_command/src/enroll/enrollment_token_generate.rs
adrianbenavides: The `attrs` argument is defined as a `Vec<String>` where each `String` is a key-value and each entry of the vector an occurrence of the argument (i.e. separated by space). So, an example of usage would be `... -- k1=v1 k2=v2`. 

The `value_delimiter` defaults to `,` as defined [here](https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.value_delimiter) so previously we had to define the attributes like `... -- k1,v1 k2,v2`, so we have to choose one of them.

The good thing about using `=` is that you can parse the key-value pair easily. See [here](https://github.com/build-trust/ockam/pull/2919/files#diff-f9e15f4f9fab6ce0ebcfea91f33f4a60632ba437984d6bbdfb89afbdc49dfb34R52-R56).

implementations/rust/ockam/ockam_command/src/message/mod.rs
spacekookie: Hm, maybe. Although I think net every file needs to have the same rules for imports. I do generally also try to separate different "types" of imports, but when there's just 2 or 3 lines it feels overkill to also add blank lines in between

implementations/rust/ockam/ockam_command/src/enroll/enrollment_token_generate.rs
spacekookie: Aah okay. Hm yea fair enough. Passing a list of arguments in is always a bit tricky so being more explicit is probably better

implementations/rust/ockam/ockam_command/src/message/mod.rs
adrianbenavides: I've reverted those changes for now.

spacekookie (dismissed):

SanjoDeundiak (approved):
Trying to wake up Mergify

PR #2922, opened 2022-07-03
feat(rust): add `message-format` global arg

adrianbenavides:
This argument is only used in one command as an example right now. In future PR's, this argument
will be used to control the format of the command output. This can be helpful to pipe commands,
process the output in json format and select the variables we need as input for the next
command.

As part of this PR, a small refactor has been done to pass a `CommandGlobalOpts` object to
subcommands, which includes the global command arguments as well as the `OckamConfig` instance.
Ideally, all commands should receive this object, so we can handle how the `message-format`
applies to each command individually, for example.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/config/get.rs
+++ b/implementations/rust/ockam/ockam_command/src/config/get.rs
@@ -1,4 +1,4 @@
-use crate::util::OckamConfig;
+use crate::CommandGlobalOpts;
 use clap::Args;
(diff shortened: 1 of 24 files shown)


spacekookie (dismissed):
LGTM

SanjoDeundiak (approved):
Rebase approve

antoinevg (commented):
+1

PR #2929, opened 2022-07-05
Identity command

SanjoDeundiak:


implementations/rust/ockam/ockam_command/src/forwarder/create.rs
```diff
@@ -31,7 +31,7 @@ impl CreateCommand {
         let port = match cfg.select_node(&cmd.node_name) {
             Some(cfg) => cfg.port,
             None => {
-                eprintln!("No such node available. Run `ockam node list` to list available nodes");
+                eprintln!("No such node available.  Run `ockam node list` to list available nodes");
```
adrianbenavides, line 34: I've seen this double-space after a period all over the codebase. Is this some formatting option you use for a reason? Do we want to do this?

implementations/rust/ockam/ockam_command/src/identity/create.rs
```diff
@@ -0,0 +1,54 @@
...
+            base_route.modify().append("_internal.nodeman"),
+            api::create_identity()?,
+        )
+        .await?;
+
+    let response = api::parse_create_identity_response(&resp)?;
+
+    match response.status() {
+        Some(Status::Ok) => {
+            eprintln!("Identity created!",)
```
adrianbenavides, line 46: Shouldn't OK messages be printed to stdout?

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -11,14 +11,20 @@ use ockam::remote::RemoteForwarder;
...
 use ockam_core::compat::{boxed::Box, collections::BTreeMap, string::String};
 use ockam_core::errcode::{Kind, Origin};
-use ockam_identity::authenticated_storage::mem::InMemoryStorage;
 use ockam_identity::{Identity, TrustEveryonePolicy};
 use ockam_multiaddr::MultiAddr;
 use ockam_vault::Vault;
 
+use crate::auth::Server;
+use crate::identity::IdentityService;
+use crate::lmdb::LmdbStorage;
```
twittner, line 20: This is currently only available if feature "lmdb" is enabled.

implementations/rust/ockam/ockam_command/src/identity/create.rs
spacekookie: We've been trying to keep `stdout` clean so that we can support things like piping the output of an ockam command into another application (we probably do currently log some things to stdout but we shouldn't

spacekookie (dismissed):
LGTM!

implementations/rust/ockam/ockam_command/src/forwarder/create.rs
SanjoDeundiak: @adrianbenavides I just wanted it to be consistent. @spacekookie thoughts?

implementations/rust/ockam/ockam_api/src/nodes/service.rs
SanjoDeundiak: @twittner Nice observation. I think it's fine since lmdb feature is enabled by default and we don't want to support other types of storages for now

twittner: It would be unfortunate if the crate does not compile with `default-features = false` so I recommend to either remove the "lmdb" feature and always make LMDB storage available or to conditionally expose it.

metaclips:
@mergifyio refresh

SanjoDeundiak:
@Mergifyio rebase

implementations/rust/ockam/ockam_api/src/nodes/service.rs
SanjoDeundiak: I think conditionally exposing it is the way to go, but doesn't make sense to me until we have some viable alternative storage

SanjoDeundiak:
@Mergifyio refresh

implementations/rust/ockam/ockam_command/src/forwarder/create.rs
spacekookie: I am personally a fan of double spaces after periods.  But I also don't think we're currently consistent about it at all.  Either way works for me :sweat_smile:

spacekookie (approved):
Still LGTM :joy:

PR #2933, opened 2022-07-07
refactor(rust): extract common utils to process api services req/res/err

adrianbenavides:
We use different approaches to handle requests/responses/errors in the ockam_api services. This leads us to misuse cbor encode/decode functionalities and have inconsistent log traces, for example.

This PR extract three functions to process requests/responses/errors (and with logs around these calls) as it's already done in the `auth` and `cloud` services. Previously, these two modules had this logic duplicated. Now they both call the same functions.

What's left to be done (in a separate PR) is refactor the other modules (identity, vault, nodes) to follow the same structure and use these new functions.


spacekookie (dismissed):
Thank you! This looks great. I've been meaning to write a util like this for weeks and never got around to it :green_heart:

implementations/rust/ockam/ockam_api/src/lib.rs
```diff
@@ -595,3 +596,124 @@ pub(crate) fn assert_response_match<'a>(schema: impl Into<Option<&'a str>>, cbor
...
+) -> ockam_core::Result<Vec<u8>>
+where
+    T: Encode<()>,
+    R: Into<Route> + Display,
+{
+    let mut buf = Vec::new();
+    req.encode(&mut buf)?;
+    assert_request_match(schema, &buf);
+    trace! {
+        target = target,
```
twittner: This should be `target: ...`, i.e. use the colon instead of `=`. Cf. https://docs.rs/tracing/0.1.35/tracing/macro.trace.html

implementations/rust/ockam/ockam_api/src/lib.rs
```diff
@@ -595,3 +596,124 @@ pub(crate) fn assert_response_match<'a>(schema: impl Into<Option<&'a str>>, cbor
...
+    let res: Response = dec.decode()?;
+    trace! {
+        target = target,
+        id     = %res.id(),
+        re     = %res.re(),
+        status = ?res.status(),
+        body   = %res.has_body(),
+        "<- {label}"
+    }
+    Ok(res)
```
twittner, line 684: A lot of the functions like this one are probably useful only because of the somewhat involved logging statement which no one wants to repeat everywhere. An alternative approach would be to implement `Value` for `Request` and `Response` which would allow writing `trace!(%res, ...)`. See https://docs.rs/tracing/0.1.35/tracing/trait.Value.html for details.

adrianbenavides: This might be helpful. For now I'll leave it as it is because I have to refactor more stuff in this crate. If this function doesn't need to be extended, I'll implement `Value` and get rid of the function.

implementations/rust/ockam/ockam_api/src/lib.rs
adrianbenavides: good catch! This forces me to use a constant here and remove that variable.

antoinevg (dismissed):

SanjoDeundiak (approved):
Waking up Mergify

PR #2935, opened 2022-07-07
feat(rust): node command usability improvements

spacekookie:
* Check whether a given API port is already in-use and prevent the
  user from spawning a node if it is.  For foreground nodes this check
  is not performed for two reasons:

1. we would have to distinguish between whether a node was
   foregrounded by the user, or is actually a spawned instance which
   has already had this check applied

2. when creating a foreground node it is much easier to see the log
   output which will inform the user that the given port is not
   avalaible

* deleting a node now also deletes the state directory
* deleting a node that was not running no longer prints an error from
  killing it.
* spawning a node is now the default. If you would like to keep a node
  attached to your terminal you must pass `--foreground` instead!
  (@hairyhum @polvorin feels like relevant information for you!)

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/delete.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/delete.rs
@@ -28,16 +28,18 @@ pub fn delete_node(cfg: &OckamConfig, node_name: &String, sigkill: bool) {
     };
 
     if let Some(pid) = pid {
-        if let Err(e) = signal::kill(
+        let _ = signal::kill(
             Pid::from_raw(pid),
             if sigkill {
                 Signal::SIGKILL
             } else {
                 Signal::SIGTERM
             },
-        ) {
-            eprintln!("Error occurred while terminating node process: {}", e);
-        }
+        );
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_command/src/node/delete.rs
```diff
@@ -28,16 +28,18 @@ pub fn delete_node(cfg: &OckamConfig, node_name: &String, sigkill: bool) {
...
-        if let Err(e) = signal::kill(
+        let _ = signal::kill(
             Pid::from_raw(pid),
             if sigkill {
                 Signal::SIGKILL
             } else {
                 Signal::SIGTERM
             },
-        ) {
-            eprintln!("Error occurred while terminating node process: {}", e);
```
SanjoDeundiak, line 39: Might print warning in the future

SanjoDeundiak (dismissed):

SanjoDeundiak:
@Mergifyio rebase

SanjoDeundiak (approved):

Issue #2597, opened 2022-03-25
Implement AWS Kinesis integration fo Stream protocol


hairyhum:
Currently Ockam Stream protocol only supports [in-memory](https://github.com/ockam-network/ockam/blob/develop/implementations/elixir/ockam/ockam/lib/ockam/stream/storage/internal.ex) and [kafka](https://github.com/ockam-network/ockam/blob/develop/implementations/elixir/ockam/ockam_kafka/lib/storage/kafka_storage.ex) storage backends.

AWS Kinesis has very similar stream API and can be the next storage to support.

PR #2943, opened 2022-07-09
feat(elixir): aws kinesis backend proof of concept for the stream service

mszpakowski:
## Current Behaviour

Elixir implementation of Ockam Stream service supports in-memory and Kafka storage backends.

### Summary
Proof of concept of an AWS Kinesis storage backend for the Ockam Stream service.

### Resolves
https://github.com/build-trust/ockam/issues/2597

### What was done
* added an AWS Kinesis storage backend that fully supports `init_stream/3`, `init_partition/4` and `save/4` stream worker callbacks
* partially supports `fetch/5` stream worker callback - due to Kinesis sequence number design the callback implementation is limited to indexes that are lower than initial sequence number for a given shard, indexes that are valid Kinesis sequence numbers in a given shard and indexes that are most recently pulled valid sequence number incremented by one (which allows pulling unconsumed messages) - with these limitations elixir Consumer workers are still able to pull messages from the stream
*  **because of said `fetch/5` callback limitations it does not fully fit within the constraints of the Ockam Stream Protocol and as such shouldn't be considered for production use** (more on that in a `Known issues` section)
* the implementation follows the module structure and configuration used in `ockam_kafka` mix project 

### What was left out
1. To keep the proof of concept simple I kept the optional configuration to minimum (i.e. there is no stream name customization, no AWS client customization through options etc). Of course I can add those if needed.
2. Outside of the scope of this PR (although I'm happy to include it or open a separate PR) is the issue of some elixir stream examples not working. As far as I can tell it's caused by the fact that `Publisher` and `Consumer` workers don't store `all_addresses` in their state, which causes authorization to fail. Adding those (and adding `stream_kafka` and `stream_kafka_index` to the list of services in `ockam_cloud_node` config) fixed the issue for me.

### Known issues
1. **Kinesis backend doesn't work with Rust workers** - Rust workers expect indices to be 64 bit unsigned integers, and Kinesis uses 186 bit unsigned integers as sequence numbers. From my perspective a way to go would be to adjust Rust worker implementation to support larger integers. An adjustment to the protocol design might be needed as well, as it only states that the index is an `uint` (which might indicate that it is supposed to be a 32-bit integer).
3. **Kinesis backend does not currently support all integers as indexes** - Kinesis allows getting messages from the stream using sequence numbers as reference. There are two types of [iterators](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_GetShardIterator.html) that allow that: `AT_SEQUENCE_NUMBER` - which retrieves a message at a provided sequence number, and `AFTER_SEQUENCE_NUMBER` - which retrieves a message after a provided sequence number. Both iterators require valid sequence numbers as references, so requesting an iterator with sequence numbers like 0, 1 or any between one valid sequence number and next valid sequence number would generally be considered invalid and return an error. This implementation works around this in admittedly hacky ways:
* an initial sequence number for a shard (it’s initialized when stream is created) is stored and if requested index is smaller or equal to initial sequence number, messages from the beginning of the stream are retrieved (using `AFTER_SEQUENCE_NUMBER` iterator and initial sequence number).
* a sequence number of the latest returned message is stored in the state - if an index of the next requested message is larger than the stored one by 1, it is assumed that the consumer wants to retrieve the next message, and `AFTER_SEQUENCE_NUMBER` iterator is used with a sequence number stored in the state.
* otherwise `AT_SEQUENCE_NUMBER` iterator is used with the requested index.

An alternative solution would be to first check if an index is valid with `AT_SEQUENCE_NUMBER` iterator, and if not, we would decrement it by one and check it again with `AFTER_SEQUENCE_NUMBER` iterator. That would eliminate the need to rely on a sequence number stored in state, and would work for all valid sequence numbers incremented by one (not only the one most recently fetched), but would come at a cost of an additional call and calls to get shard iterator are rate limited to 5 calls per second.
Neither of these solutions would work if the requested index was larger by more than 1 than a valid sequence number. I have no good idea on how to solve it within the current constraints of the Ockam Stream protocol, so any ideas are welcome.

A more robust solution would require changes to the protocol and I don’t know how feasible it would be at this point. It could look something like that:
```
enum IndexReference {
  AT_INDEX
  AFTER_INDEX
}

type PullRequest {
  request_id: uint
  index: uint
  index_reference: IndexReference
  limit: uint
}

enum Status {
  OK
  ERROR
}

​​type PullResponse {
  request_id: uint
  status: Status
  messages: []StreamMessage
}
```
It would return errors for invalid indexes and would allow clients to specify if they want to get a message at a specific index or after an already processed index (instead of incrementing previously received index by one).

4. Requests to get shard iterator and to get records from AWS Kinesis are rate limited to 5 calls per second per shard. Heavy consumption will lead to hitting these rate limits and therefore some throttling either on consumer side or worker side might be necessary for optimization (currently if that happens stream worker replies with an empty list of messages, consumer waits for 2 seconds, and retries).
5.  Due to design quirks of Kinesis, there might be a possibility that a call to request messages with `limit` bigger than 1 might not return the full number of messages. Shards in a Kinesis stream are split into portions, and not every portion is guaranteed to have messages. In sparsely populated shards, a request to get records might return an empty list, even if there are messages further down the stream. In the implementation shard portions are iterated over until a message is found or the end of the stream is reached. That theoretically (I haven't been able to reproduce it) leaves a possibility that in given shard portion there would be a smaller number of messages than requested, which would get subsequently returned, even if there might be more messages later in the stream.

## Checks


mrinalwadhwa:
@mszpakowski thank you for sending this PR ❤️ 
We'll read/try and come back with questions.

implementations/elixir/ockam/ockam_kinesis/lib/storage/kinesis_storage.ex
```diff
@@ -0,0 +1,272 @@
...
+         limit,
+         %State{
+           initial_sequence_number: initial_sequence_number,
+           previous_sequence_number: previous_sequence_number
+         } = state
+       )
+       when index <= initial_sequence_number do
+    with {:ok, shard_iterator} <-
+           Kinesis.get_shard_iterator(stream_name, partition, :after_sequence_number,
+             starting_sequence_number: to_string(initial_sequence_number)
```
polvorin: shouldn't be `:at_sequence_number` here to not miss the first one?  (or maybe use  `TRIM_HORIZON` to make it explicit)

implementations/elixir/ockam/ockam_kinesis/lib/storage/kinesis_storage.ex
```diff
@@ -0,0 +1,272 @@
...
+        error
+    end
+  end
+
+  # NOTE: When we repeat fetching messages for the same index and we have a shard
+  #       iterator stored in the state, we pick up where we left off.
+  #       This would clause would realistically get invoked only if the index
+  #       is at the end of the stream and we're awaiting new messages.
+  #       If an iterator has expired, we reset it to `nil` so it can be retrieved again.
+  defp fetch_messages(
```
polvorin: there is a bit of repetition between these `fetch_messages()` clauses.  Maybe there is opportunity to simplify?,  something like  
1) first getting the iterator to use (either the cached one is applicable, as you did,  or request a new one)
2) use the iterator and update the state

polvorin (commented):
Thanks for the pull request, looks good overall and you take the effort to try to avoid re-creating the shard iterator if the "previous" one would still be valid for the received query.

Honestly,  as you mention there is room for improvement on the stream API to accommodate for the kinesis case.   I'm not too fond of keeping the shard iterator here,  as this worker is something that would be shared between different clients.   So, perhaps we could return the iterator (as an opaque value) to the clients so they can ask to continue from it.   Or keep the iterator server-side,  but in a session attached to the specific client.    These would require some changes to the API

implementations/elixir/ockam/ockam_kinesis/lib/storage/kinesis_storage.ex
mszpakowski: The first message would have a different sequence number than the initial sequence number and from what I observed (unfortunately official documentation around `StartingSequenceNumber` is lacking) we can use both `AFTER_SEQUENCE_NUMBER` and `AT_SEQUENCE_NUMBER` with the initial sequence number to retrieve the first message. I chose `AFTER_SEQUENCE_NUMBER` as I found it more intuitive, given that, as mentioned, the sequence number of the message would be different than the initial one.

That being said, you might be right that the least confusing option would be to simply use `TRIM_HORIZON` instead.

implementations/elixir/ockam/ockam_kinesis/lib/storage/kinesis_storage.ex
mszpakowski: It did cross my mind when writing this function, but at the moment I thought that the logic is already pretty complicated that abstracting it more might come at the cost of readability.

That's not a strong opinion however, so I'll refactor it.

hairyhum:
Thanks for the PR.
`uint` is  `A variable-length unsigned integer encoded using the Unsigned Little Endian Base 128`
So if Rust implementation decodes it as 64 bit, it's not doing it properly, but still if we want to support 186 bit, we need to improve the encoding format.

As for the iterators, it's a good point that client progressing indexes might not be the best approach. I agree with @polvorin that it would be a better solution to leave index progression to the implementation (stream storage and index callback modules) and make it opaque for the client.
This would also help with 186 bit indexes or more exotic indexes in other backends.

mszpakowski:
closing this, if we want to use some of that in the future, we can revisit it then

PR #2945, opened 2022-07-11
refactor(rust): cloud api endpoints to run through the nodes service

adrianbenavides:
This is the first PR to run all cloud api endpoints through a node secure channel. It only includes the refactor on the ockam_api crate. It also includes some changes to the types, adding `serde` support, that will be used by the commands. 

## Te be done in subsequent PR's
* The second part will include the changes in the ockam_command crate
* There is a major todo in this PR that is updating the tests and removing the cloud's `MessagingClient` will be done in a third PR
* Finally, I still need to figure out how to send the secure channel address from the command to the node as part of the message payload, which will require another PR involving both crates (api + commands). I guess I could create new request types just as the ones used to communicate the node with the cloud, but adding an extra field for the channel address. I want to explore first if I can reuse the same types for both cases to avoid duplicating stuff.

Next: https://github.com/build-trust/ockam/pull/2947


adrianbenavides:
The second part will be ready as soon as this PR gets merged. The rest will be done between today and tomorrow.

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #2965, opened 2022-07-15
signer/authenticator

twittner:
Add an authenticator service that allows enrollers to add members and members to get and verify signed credentials.

### DIFF
--- a/implementations/rust/ockam/ockam_api/schema.cddl
+++ b/implementations/rust/ockam/ockam_api/schema.cddl
@@ -190,3 +190,25 @@ request_enrollment_token = {
     ?0: 8560526,
      1: attributes
 }
+
+;;; Signer ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+credential_request = [0, identity_credential_request]
+
+identity_credential_request = {
+    0: bytes ;; identity id bytes
+}
+
+credential = [0, identity_credential]
+
+identity_credential = {
+    0: bytes, ;; identity id bytes
+    1: signer_signature
+}
+
+signer_signature = [
+    0: identity_id,
+    1: signer_signature_bytes
+]
+
+signer_signature_bytes = bytes
--- a/implementations/rust/ockam/ockam_api/schema.cddl
+++ b/implementations/rust/ockam/ockam_api/schema.cddl
@@ -193,17 +193,12 @@ request_enrollment_token = {
 
 ;;; Signer ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-credential_request = [0, identity_credential_request]
+sign_request = attributes
 
-identity_credential_request = {
-    0: bytes ;; identity id bytes
-}
-
-credential = [0, identity_credential]
-
-identity_credential = {
-    0: bytes, ;; identity id bytes
-    1: signer_signature
+credential = {
+    ?0: 3796735,
+     1: bytes, ;; CBOR-encoded attributes
+     2: signer_signature
 }
 
 signer_signature = [
@@ -212,3 +207,15 @@ signer_signature = [
 ]
 
 signer_signature_bytes = bytes
+
+;;; Authenticator ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+add_enroller = {
+    ?0: 1010815,
+     1: identity_id,
+}
+
+add_member = {
+    ?0: 2820828,
+     1: identity_id,
+}
--- a/implementations/rust/ockam/ockam_api/schema.cddl
+++ b/implementations/rust/ockam/ockam_api/schema.cddl
@@ -208,6 +208,16 @@ signer_signature = [
 
 signer_signature_bytes = bytes
 
+add-identity = {
+    ?0: 1544679,
+     1: identity
+}
+
+get-identity-response = {
+    ?0: 9967409,
+     1: identity
+}
+
 ;;; Authenticator ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 add_enroller = {
--- a/implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
+++ b/implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
@@ -1,11 +1,18 @@
 pub mod types;
 
-use crate::signer;
+use crate::auth::types::Attributes;
+use crate::signer::types::{Credential, IdentityId};
 use crate::util::response;
-use crate::{Error, Method, Request, Response, ResponseBuilder};
+use crate::{assert_request_match, assert_response_match, Cbor};
+use crate::{signer, Timestamp};
+use crate::{Error, Method, Request, RequestBuilder, Response, ResponseBuilder, Status};
 use core::marker::PhantomData;
-use minicbor::Decoder;
-use ockam_core::{self, Result, Routed, Worker};
+use core::time::Duration;
+use core::{fmt, str};
+use minicbor::encode::write::Cursor;
+use minicbor::{Decoder, Encode};
+use ockam_core::errcode::{Kind, Origin};
+use ockam_core::{self, Address, Result, Route, Routed, Worker};
 use ockam_identity::authenticated_storage::AuthenticatedStorage;
 use ockam_identity::{IdentityIdentifier, IdentitySecureChannelLocalInfo};
 use ockam_node::Context;
@@ -15,6 +22,8 @@ use types::{AddEnroller, AddMember, Placeholder};
 // storage scopes:
 const ENROLLER: &str = "enroller";
 const MEMBER: &str = "member";
+
+const MAX_VALIDITY: Duration = Duration::from_secs(2 * 3600);
 
 #[derive(Debug)]
 pub struct Server<M, S> {
@@ -81,7 +90,7 @@ impl<S: AuthenticatedStorage> Server<General, S> {
             Some(Method::Post) => match req.path_segments::<2>().as_slice() {
                 // Enroller wants to add a member.
                 ["member"] => {
-                    if let Some(err) = self.check_enroller_auth(&req, from.key_id()).await? {
+                    if let Some(err) = check_enroller(&self.store, &req, from.key_id()).await? {
                         err.to_vec()?
                     } else {
                         let add: AddMember = dec.decode()?;
@@ -94,10 +103,16 @@ impl<S: AuthenticatedStorage> Server<General, S> {
                 }
                 // Member wants a credential.
                 ["credential"] => {
-                    if let Some(err) = self.check_member_auth(&req, from.key_id()).await? {
+                    if let Some(err) = check_member(&self.store, &req, from.key_id()).await? {
                         err.to_vec()?
                     } else {
-                        let crd = self.signer.sign_id(from).await?;
+                        let mut attrs = Attributes::new();
+                        attrs.put("id", from.key_id().as_bytes());
+                        let ts = Timestamp::now().ok_or_else(invalid_sys_time)?;
+                        let mut timestamp = Cursor::new([0; 9]);
+                        minicbor::encode(ts, &mut timestamp)?;
+                        attrs.put("ts", &timestamp.get_ref()[..timestamp.position()]);
+                        let crd = self.signer.sign(&attrs).await?;
                         Response::ok(req.id()).body(crd).to_vec()?
                     }
                 }
@@ -106,13 +121,28 @@ impl<S: AuthenticatedStorage> Server<General, S> {
             Some(Method::Get) => match req.path_segments::<3>().as_slice() {
                 // Enroller checks member data.
                 ["member", id] => {
-                    if let Some(err) = self.check_enroller_auth(&req, from.key_id()).await? {
+                    if let Some(err) = check_enroller(&self.store, &req, from.key_id()).await? {
                         err.to_vec()?
                     } else if let Some(data) = self.store.get(MEMBER, id).await? {
                         let member = minicbor::decode::<Placeholder>(&data)?;
                         Response::ok(req.id()).body(member).to_vec()?
                     } else {
                         Response::not_found(req.id()).to_vec()?
+                    }
+                }
+                // Validate member credential.
+                ["credential"] => {
+                    let crd: Credential = dec.decode()?;
+                    if let Ok(att) = self.signer.verify(&crd).await {
+                        if let Some(err) = check_credential(&self.store, &req, &att).await? {
+                            err.to_vec()?
+                        } else {
+                            Response::ok(req.id())
+                                .body(Cbor(crd.attributes()))
+                                .to_vec()?
+                        }
+                    } else {
+                        Response::unauthorized(req.id()).to_vec()?
                     }
                 }
                 _ => response::unknown_path(&req).to_vec()?,
@@ -122,48 +152,88 @@ impl<S: AuthenticatedStorage> Server<General, S> {
 
         Ok(res)
     }
+}
 
-    async fn check_enroller_auth<'a>(
-        &self,
-        req: &'a Request<'_>,
-        enroller: &str,
-    ) -> Result<Option<ResponseBuilder<Error<'a>>>> {
-        if let Some(data) = self.store.get(ENROLLER, enroller).await? {
-            minicbor::decode::<Placeholder>(&data)?;
-            return Ok(None);
+async fn check_credential<'a, S: AuthenticatedStorage>(
+    store: &S,
+    req: &'a Request<'_>,
+    attrs: &Attributes<'_>,
+) -> Result<Option<ResponseBuilder<Error<'a>>>> {
+    // Verify that the credential has not expired
+
+    let ts = if let Some(val) = attrs.get("ts") {
+        minicbor::decode(val)?
+    } else {
+        return Ok(Some(response::bad_request(
+            req,
+            "missing credential timestamp",
+        )));
+    };
+    let now = Timestamp::now().ok_or_else(invalid_sys_time)?;
+    if let Some(dur) = now.elapsed(ts) {
+        if dur > MAX_VALIDITY {
+            return Ok(Some(response::forbidden(req, "credential expired")));
         }
-        warn! {
-            target: "ockam_api::authenticator::direct::server",
-            enroller = %enroller,
-            id       = %req.id(),
-            method   = ?req.method(),
-            path     = %req.path(),
-            body     = %req.has_body(),
-            "unauthorised enroller"
-        }
-        Ok(Some(response::forbidden(req, "unauthorized enroller")))
+    } else {
+        return Ok(Some(response::bad_request(
+            req,
+            "invalid credential timestamp",
+        )));
     }
 
-    async fn check_member_auth<'a>(
-        &self,
-        req: &'a Request<'_>,
-        member: &str,
-    ) -> Result<Option<ResponseBuilder<Error<'a>>>> {
-        if let Some(data) = self.store.get(MEMBER, member).await? {
-            minicbor::decode::<Placeholder>(&data)?;
-            return Ok(None);
-        }
-        warn! {
-            target: "ockam_api::authenticator::direct::server",
-            member   = %member,
-            id       = %req.id(),
-            method   = ?req.method(),
-            path     = %req.path(),
-            body     = %req.has_body(),
-            "unauthorised member"
-        }
-        Ok(Some(response::forbidden(req, "unauthorized member")))
+    // Verify that the member is authorised
+
+    let id = if let Some(val) = attrs.get("id") {
+        str::from_utf8(val).map_err(invalid_utf8)?
+    } else {
+        return Ok(Some(response::bad_request(
+            req,
+            "missing credential identity",
+        )));
+    };
+    check_member(store, req, id).await
+}
+
+async fn check_member<'a, S: AuthenticatedStorage>(
+    store: &S,
+    req: &'a Request<'_>,
+    member: &str,
+) -> Result<Option<ResponseBuilder<Error<'a>>>> {
+    if let Some(data) = store.get(MEMBER, member).await? {
+        minicbor::decode::<Placeholder>(&data)?;
+        return Ok(None);
     }
+    warn! {
+        target: "ockam_api::authenticator::direct::server",
+        member   = %member,
+        id       = %req.id(),
+        method   = ?req.method(),
+        path     = %req.path(),
+        body     = %req.has_body(),
+        "unauthorised member"
+    }
+    Ok(Some(response::forbidden(req, "unauthorized member")))
+}
+
+async fn check_enroller<'a, S: AuthenticatedStorage>(
+    store: &S,
+    req: &'a Request<'_>,
+    enroller: &str,
+) -> Result<Option<ResponseBuilder<Error<'a>>>> {
+    if let Some(data) = store.get(ENROLLER, enroller).await? {
+        minicbor::decode::<Placeholder>(&data)?;
+        return Ok(None);
+    }
+    warn! {
+        target: "ockam_api::authenticator::direct::server",
+        enroller = %enroller,
+        id       = %req.id(),
+        method   = ?req.method(),
+        path     = %req.path(),
+        body     = %req.has_body(),
+        "unauthorised enroller"
+    }
+    Ok(Some(response::forbidden(req, "unauthorized enroller")))
 }
 
 impl<S: AuthenticatedStorage> Server<Admin, S> {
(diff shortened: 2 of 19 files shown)


implementations/rust/ockam/ockam_api/schema.cddl
```diff
@@ -190,3 +190,42 @@ request_enrollment_token = {
...
+
+sign_request = attributes
+
+credential = {
+    ?0: 3796735,
+     1: bytes, ;; CBOR-encoded attributes
+     2: signer_signature
+}
+
+signer_signature = [
```
SanjoDeundiak, line 204: Can this be just `signature` or maybe `credential_signature`?

implementations/rust/ockam/ockam_api/schema.cddl
```diff
@@ -190,3 +190,42 @@ request_enrollment_token = {
...
 }
+
+;;; Signer ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+sign_request = attributes
+
+credential = {
+    ?0: 3796735,
+     1: bytes, ;; CBOR-encoded attributes
+     2: signer_signature
```
SanjoDeundiak, line 201: Off-topic: but do we consider possible situations where credential has multiple signatures?

implementations/rust/ockam/ockam_api/schema.cddl
```diff
@@ -190,3 +190,42 @@ request_enrollment_token = {
...
+    ?0: 1544679,
+     1: identity
+}
+
+get-identity-response = {
+    ?0: 9967409,
+     1: identity
+}
+
+;;; Authenticator ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
SanjoDeundiak, line 221: Can this be divided into multiple files? Considering they share some types

implementations/rust/ockam/ockam_api/src/signer.rs
```diff
@@ -0,0 +1,279 @@
...
+use ockam_core::{self, vault, Address, Result, Route, Routed, Worker};
+use ockam_identity::authenticated_storage::AuthenticatedStorage;
+use ockam_identity::change_history::IdentityChangeHistory;
+use ockam_identity::{Identity, IdentityIdentifier, IdentityVault};
+use ockam_node::Context;
+use tracing::{trace, warn};
+use types::{Credential, IdentityId, Signature};
+
+pub struct Server<V: IdentityVault, S> {
+    id: Identity<V>,
```
SanjoDeundiak, line 20: Can it be `identity`? `id` is usually short for `identifier`

implementations/rust/ockam/ockam_api/src/signer.rs
```diff
@@ -0,0 +1,279 @@
...
+impl fmt::Debug for Client {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Client")
+            .field("route", &self.route)
+            .finish()
+    }
+}
+
+impl Client {
+    pub async fn new(r: Route, ctx: &Context) -> Result<Self> {
```
SanjoDeundiak, line 160: `ctx` is usually first argument (after `self`)

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -0,0 +1,419 @@
...
+        }
+
+        let res = match req.method() {
+            Some(Method::Post) => match req.path_segments::<2>().as_slice() {
+                // Admin wants to add an enroller.
+                ["enroller"] if self.is_admin(from) => {
+                    let add: AddEnroller = dec.decode()?;
+                    let enroller = minicbor::to_vec(Placeholder)?;
+                    self.e_store
+                        .set(ENROLLER, add.enroller().as_str().to_string(), enroller)
```
SanjoDeundiak, line 96: Isn't enroller id meant to go as a first arguments? Should `id` argument in `AuthenticatedStorage` be of type `IdentityIdentifier` to help avoid such situations?

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -0,0 +1,419 @@
...
+
+// storage scopes:
+const ENROLLER: &str = "enroller";
+const MEMBER: &str = "member";
+
+const MAX_VALIDITY: Duration = Duration::from_secs(2 * 3600);
+
+#[derive(Debug)]
+pub struct Server<S, T> {
+    m_store: S, // member store
```
SanjoDeundiak, line 29: Could it be one store?

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -0,0 +1,419 @@
...
+    e_store: T, // enroller store
+    signer: signer::Client,
```
SanjoDeundiak, line 31: Is corresponding Server meant to be on the same node? If yes, can we use it directly without this excess Client hop?

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -0,0 +1,419 @@
...
+                    let add: AddEnroller = dec.decode()?;
+                    let enroller = minicbor::to_vec(Placeholder)?;
+                    self.e_store
+                        .set(ENROLLER, add.enroller().as_str().to_string(), enroller)
+                        .await?;
+                    Response::ok(req.id()).to_vec()?
+                }
+                ["enroller"] => Response::unauthorized(req.id()).to_vec()?,
+                // Enroller wants to add a member.
+                ["member"] => {
```
SanjoDeundiak, line 102: Would be great if we could stick to the same conventions for paths and methods. E.g. Rest-like. This would be POST members. This is an example I think we had few people agreed on a while ago https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/identity/identity_service.rs

implementations/rust/ockam/ockam_api/src/authenticator/direct/types.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    pub fn member(&self) -> &IdentityId {
+        &self.member
+    }
+}
+
+/// Used until we know what enroller/member data we want to store.
+#[derive(Debug, Decode, Encode)]
+#[cbor(map)]
+pub struct Placeholder;
```
SanjoDeundiak, line 56: Should it at least have a `TypeTag`?

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -0,0 +1,419 @@
...
+                        let enroller = minicbor::decode::<Placeholder>(&data)?;
+                        Response::ok(req.id()).body(enroller).to_vec()?
+                    } else {
+                        Response::not_found(req.id()).to_vec()?
+                    }
+                }
+                ["enroller", _] => Response::unauthorized(req.id()).to_vec()?,
+                // Enroller checks member data.
+                ["member", id] => {
+                    if let Some(err) = check_enroller(&self.e_store, &req, from.key_id()).await? {
```
SanjoDeundiak, line 144: Is admin allowed to do everything enroller has permissions to do?

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
SanjoDeundiak: Maybe it could look something like this?
```
self.e_store.set(add.enroller_id()), "is_enroller", true)
```

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -0,0 +1,419 @@
...
+                        err.to_vec()?
+                    } else if let Some(data) = self.m_store.get(MEMBER, id).await? {
+                        let member = minicbor::decode::<Placeholder>(&data)?;
+                        Response::ok(req.id()).body(member).to_vec()?
+                    } else {
+                        Response::not_found(req.id()).to_vec()?
+                    }
+                }
+                // Validate member credential.
+                ["credential"] => {
```
SanjoDeundiak, line 154: Just noting: there must exist check not only if credential is valid, but also if it was issued by known authority and if it was issued to the party presenting this credential. The latter check may be performed here as well

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -0,0 +1,419 @@
...
+        minicbor::decode(val)?
+    } else {
+        return Ok(Some(response::bad_request(
+            req,
+            "missing credential timestamp",
+        )));
+    };
+    let now = Timestamp::now().ok_or_else(invalid_sys_time)?;
+    if let Some(dur) = now.elapsed(ts) {
+        if dur > MAX_VALIDITY {
```
SanjoDeundiak, line 208: Off-topic: I think we want to have expiration date in the credential as well

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -0,0 +1,419 @@
...
+
+    let id = if let Some(val) = attrs.get("id") {
+        str::from_utf8(val).map_err(invalid_utf8)?
+    } else {
+        return Ok(Some(response::bad_request(
+            req,
+            "missing credential identity",
+        )));
+    };
+    check_member(store, req, id).await
```
SanjoDeundiak, line 228: I don't get the following fact: if we can verify that presenter is a member according to our storage, why would we need a credential? Or if have a valid credential presented, why would we check membership in our storage?

implementations/rust/ockam/ockam_api/src/signer/types.rs
```diff
@@ -0,0 +1,180 @@
...
+
+#[derive(Debug, Decode, Encode)]
+#[rustfmt::skip]
+pub struct Signature<'a> {
+    #[b(0)] ident: IdentityId<'a>,
+    #[b(1)] signature: CowBytes<'a>
+}
+
+impl<'a> Signature<'a> {
+    pub fn new<S>(id: IdentityId<'a>, sig: S) -> Self
```
SanjoDeundiak, line 77: Can this be `IdentityIdentifier`?

SanjoDeundiak, line 77: Also, I think we usually use `impl Into<Cow<'a, [u8]>>` as more ergonomic if you wish

implementations/rust/ockam/ockam_api/tests/auth.rs
```diff
@@ -0,0 +1,104 @@
...
+        .create_secure_channel("api", TrustEveryonePolicy, &InMemoryStorage::new())
+        .await?;
+
+    // Create a member identity:
+    let v = Vault::create();
+    let m = Identity::create(ctx, &v).await?;
+
+    // Add the member via the enroller's connection:
+    let mut c = direct::Client::new(route![e2a, "auth"], ctx).await?;
+    c.add_member(IdentityId::new(m.identifier()?.key_id()))
```
SanjoDeundiak, line 72: Can we avoid referencing `.key_id()` everywhere? This is kinda implementation detail. `IdentityIdentifier` has its own conversion to string

implementations/rust/ockam/ockam_api/tests/auth.rs
```diff
@@ -0,0 +1,104 @@
...
+    assert!(atts.get("ts").is_some());
+
+    // Here we pretend to be on a different node and want to validate the credential:
+    {
+        let v = Vault::create();
+        let a = Identity::create(ctx, &v).await?;
+        let s = signer::Server::new(a, InMemoryStorage::new());
+        ctx.start_worker("another-signer", s).await?;
+        let mut c = signer::Client::new("another-signer".into(), ctx).await?;
+        c.add_identity(auth_identity.identity()).await?;
```
SanjoDeundiak, line 95: This looks very massive relative to what it does: mostly just verifies the signature. I think we could use `Server` directly (in case this is other member's rust node). Maybe passing identity as an argument could be cleaner, then adding it permanently. Also the name `add_identity` is misleading for me

implementations/rust/ockam/ockam_api/tests/auth.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    // Get a fresh member credential and verify its validity:
+    let cred = c.credential().await?.to_owned();
+    let atts = c.validate(&cred).await?;
+    assert!(atts.get("id").is_some());
+    assert!(atts.get("ts").is_some());
+
+    // Here we pretend to be on a different node and want to validate the credential:
+    {
+        let v = Vault::create();
```
SanjoDeundiak, line 90: I'm not saying it must be added to that test, but noting for now: it's critically important that credential is obtained from the other party through a secure channel AND it's checked that credential subject is the same as secure channel participant. I'm worried that this will be often forgotten (which breaks the whole system), so that maybe we should make verifier take subject id as an extra argument so that it will perform the check as well?

SanjoDeundiak (dismissed):
Have lots of questions, let's keep discussion going, but good to land. Nice job!

mrinalwadhwa:
@Mergifyio refresh

mrinalwadhwa (dismissed):

mrinalwadhwa:
Github seems to not want to run checks on this PR 🧐

mrinalwadhwa:
Seems they are having a downtime https://www.githubstatus.com
Screen Shot 2022-08-09 at 5 41 41 PM

mrinalwadhwa (approved):

implementations/rust/ockam/ockam_api/schema.cddl
twittner: `signature` is already in use in `schema.cddl`, but `credential_signature` should be ok.

implementations/rust/ockam/ockam_api/schema.cddl
twittner: We had some discussions about this in the past. So far the more likely scenario is multiple credentials I think, but if we need more signature we could extend `credential` and allow additional signatures.

implementations/rust/ockam/ockam_api/schema.cddl
twittner: You mean splitting `schema.cddl`?

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
twittner: The authenticated storage layout was not really specified and is currently not restricted to entries of the type `IdentityId × String × Bytes`. Originally the thought was that the `Bytes` values would be signed data to allow validating the authenticity again but this may have been overtaken by events. The first element is merely a scope to hold key-value pairs and a means to avoid collisions when multiple services share a single storage. In the case of globally unique entries such as identity IDs we can change that and store them the way you proposed but I am not sure if `AuthenticatedStorage` should be made specific for identities.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
twittner: It used to be one, but after discussions with @polvorin it was changed so that enrollers live in RAM only and are set when the project node is provisioned. With the current protocol, member information needs to be persistent, which lead to the separation of the stores.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
twittner: It is meant to be on the same node. I have some local changes which I can turn into a PR that allows signature validation with just the reference to `Identity` and the `AuthenticatedStorage`.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
twittner: How would that look here?

implementations/rust/ockam/ockam_api/src/authenticator/direct/types.rs
twittner: Maybe. I hoped that we would clarify this before the PR got merged.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
twittner: Effectively yes. The admin would first have to add itself as an enroller before adding members.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
twittner: When the authenticator runs in the cloud as part of the project's authority and some entity asks for the credential validation it may be that despite a valid credential the enroller has previously removed the member (not implemented yet), hence the validation should fail. I am not sure how useful the authenticator is on client side where we do not issue credentials.

implementations/rust/ockam/ockam_api/src/signer/types.rs
twittner: Long ago, when I started this, I went with this `IdentityId` because `IdentityIdentifier` always allocates. Maybe we can revisit this.

twittner: I think we would only use that if we have just byte slices. Where we have proper types I think we should better use them.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
twittner: https://github.com/build-trust/codebase/issues/375

hairyhum:
@twittner do we have the API spec for this component?

Also from the code it seems like verification is done by the authority, and we need to move it out because we might not have access to the authority when verifying a credential.

twittner:
> @twittner do we have the API spec for this component?

`schema.cddl` contains the relevant parts. As for the credential itself, see https://github.com/build-trust/codebase/issues/375.

> Also from the code it seems like verification is done by the authority, and we need to move it out because we might not have access to the authority when verifying a credential.

Every node can start its own authenticator and signer service locally. Access to a remote authority is not required. As I mentioned [above](https://github.com/build-trust/ockam/pull/2965#discussion_r942206527) I can make the verification part accessible via reference to identity and authenticated storage, but this is an implementation detail.

hairyhum:
> schema.cddl contains the relevant parts. As for the credential itself, see https://github.com/build-trust/codebase/issues/375.

Can we also have an API doc in addition to the schema?

> Every node can start its own authenticator and signer service locally. Access to a remote authority is not required. As I mentioned https://github.com/build-trust/ockam/pull/2965#discussion_r942206527 I can make the verification part accessible via reference to identity and authenticated storage, but this is an implementation detail.

My point is we don't need a signer to verify credentials. My understanding is that signer has some storage, it's using authenticated identities table, it has a private key etc. None of this is needed for a verifier component.

twittner:
> My understanding is that signer has some storage, it's using authenticated identities table, it has a private key etc. None of this is needed for a verifier component.

As I already stated above, I can provide a verification function that is accessible via references to identity and storage, but it is not true that storage is not needed for the verification. Signatures contain the identity ID of the authority and storage is used to load the corresponding public key (which must have been added previously).

The current implementation more or less reflects https://github.com/build-trust/codebase/issues/296 and I am fully aware that this is not the end of the story.

hairyhum:
Can we just provide the identity of authority in the `verify` request instead of adding it to the state and then referring to that?

Also I don't see verify API in the schema definition.

twittner:
> Can we just provide the identity of authority in the verify request instead of adding it to the state and then referring to that?

In case we want to support multiple signatures at some point it would become problematic as we would not know which identity to use for which signature. It would also be highly redundant to include it in every credential exchange between members. And finally we would still need to compare the provided identity with the one we are supposed to use (presumably stored locally), as otherwise the sender could present us a signature from an arbitrary authority.

> Also I don't see verify API in the schema definition.

To verify, the credential is sent as request body.

hairyhum:
> In case we want to support multiple signatures at some point it would become problematic as we would not know which identity to use for which signature.

It's a valid argument, but we can always provide a list of accepted identities.

> It would also be highly redundant to include it in every credential exchange between members. 

Members have control over which authorities they accept the signatures from, so it's not a part of credential exchange in any case.

> And finally we would still need to compare the provided identity with the one we are supposed to use 

We already do that, no? That wouldn't change regardless of where you get the identity from.

My main concern is that `member` and `verifier` are different components and I would prefer if member component had control over which authorities it accepts credentials from. If we have that in the verifier component, then member would not be able to control that, other than adding new ones.
If we go the way when verifier stores authority identities then we need to give more control to the member to chose which ones it wants to verify. Such as not having the "own" identity for the verifier and allowing member to remove identities. In other words not allowing the verifier to verify anything other than what member explicitly tells it to verify.
Still I would prefer if that was per-call basis and without a storage on the verifier side.

hairyhum:
Also verifier defninitely needs to be separate from the signer in that model.

implementations/rust/ockam/ockam_api/schema.cddl
SanjoDeundiak: Yes

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
SanjoDeundiak: Unless we have use-cases where id is not an `IdentityIdentifier`, let's use it this way. Other use-cases could probably have different traits

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
SanjoDeundiak: I guess that means that list of enrollers live somewhere on elixir side? This is very weird.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
SanjoDeundiak: That would be great. Maybe it could be wrapped into a separate struct, which would be used from both here and `Server`

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
SanjoDeundiak: This particular request would be POST members. I believe other requests have good examples in the link I posted. Also REST best practices are available on the Internet if we haven't yet covered some scenarios

implementations/rust/ockam/ockam_api/src/authenticator/direct/types.rs
SanjoDeundiak: Oh, my task was to land this PR :D

implementations/rust/ockam/ockam_api/src/signer/types.rs
SanjoDeundiak: Yes, you can improve `IdentityIdentifier` struct if needed

SanjoDeundiak:
It makes sense that verifier can and should be separate. Also I think that at this point we can define different types of rust components, to further clarify what functionality should be available where. The types are:
1. Rust native code
2. FFI
3. Rust-served services living in a sidecar
4. Rust-served services exposed to nodes living outside of the internal network
5. `NodeManager`

Thoughts?

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
SanjoDeundiak: This is related to the discussion below - about separation of the authenticator and verifiers. Authenticator is mostly dedicated to issuing credentials, while verifier doesn't have access to the members storage. Is that right?

implementations/rust/ockam/ockam_api/schema.cddl
twittner: Yes, it would be possible. The functions `assert_request_match` and `assert_response_match` would need to accept the schema file as another parameter.

implementations/rust/ockam/ockam_api/schema.cddl
twittner: https://github.com/build-trust/ockam/commit/9d3bcb5d6aa7c9716422f011f7755e07a75a56de

implementations/rust/ockam/ockam_api/src/signer.rs
twittner: https://github.com/build-trust/ockam/commit/9d3bcb5d6aa7c9716422f011f7755e07a75a56de

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
twittner: https://github.com/build-trust/ockam/commit/5967700e0150bafef2b26dc14c3b48971ed6246f

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
twittner: There is this: https://github.com/build-trust/ockam/commit/0d8e9b287ddd6e661cfc199f1acb6dfb05b0ab0e. If we make it a struct, would it not be identical to the existing `signer::Server`?

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
twittner: https://github.com/build-trust/ockam/commit/b436e864e43e9e8b839c5fc33ce6bdc725f5fa79

implementations/rust/ockam/ockam_api/src/authenticator/direct/types.rs
twittner: Removed in https://github.com/build-trust/ockam/commit/5967700e0150bafef2b26dc14c3b48971ed6246f.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
twittner: It depends on the kind of verification that is be done. The signer verifies the signature only and leaves it to other parts of the system to inspect the credential further. If this authenticator is used to verify a credential (I imagined this would perhaps be used in the sidecar scenario) it considers its member storage.

twittner: Pure client-side credential verification is not implemented beyond using the signer to verify some credential. I think this would be part of https://github.com/build-trust/codebase/issues/369.

twittner:
> > In case we want to support multiple signatures at some point it would become problematic as we would not know which identity to use for which signature.
> 
> It's a valid argument, but we can always provide a list of accepted identities.

What I meant was that signature verification would require us to try all combinations of signatures and public keys since we do not know which key to use.

> > It would also be highly redundant to include it in every credential exchange between members.
> 
> Members have control over which authorities they accept the signatures from, so it's not a part of credential exchange in any case.

Maybe we are talking past each other, but I understood your comment

> Can we just provide the identity of authority in the `verify` request instead of adding it to the state and then referring to that?

to mean that the verification in `signer::Server` should accept the identity and this part was intended to be used on client-side for member credential verification as well.

> > And finally we would still need to compare the provided identity with the one we are supposed to use
> 
> We already do that, no? That wouldn't change regardless of where you get the identity from.

The current protocol works like this:

The credential contains the identity ID of the signer. The party that receives the credential loads the corresponding public key from storage where it had been added previously when the enroller provided the authority's identity to the member. If the key is not found signature validation fails.

If the public key is provided in the call to verify, would we not still need to consult local storage to see if we have that key?

The protocol was specified here: https://github.com/build-trust/codebase/issues/324

hairyhum:
> If the public key is provided in the call to verify, would we not still need to consult local storage to see if we have that key?

What I mean is that we still make a check that the signature key id is in the "accepted keys" and we have a check for that. Difference would only be that "accepted keys" may be controlled by the member via arguments somehow.  

> Also I think that at this point we can define different types of rust components, to further clarify what functionality should be available where.

Yes. We're currently running `3` and  `4` in sidecars, but ideally would like to run them in different sidecars so maybe we can call them differently.

FFI (`2`) only exposes vault functions and only used in encrryption secure channel from Elixir perspective.

I'd like to run verifier component in the sidecar (`3`) for easier access and also because it's more simple RPC component

I'd like to run the signer (authority) component in more "proper" node, exposed via secure channel listener (`4`), while still running like a sidecar from k8s point of view.
This node secure channel needs to be authorizing enrollers and members.

SanjoDeundiak:
My thoughts are:
1. `2` should be replaced with `3` in the future
2. `3` should avoid having persistent state. The only exception to this rule is Vault storage. Everything else should be stored on the caller side
3. `3` should be managed by rust `NodeManager` and be started by `ockam services [SERVICE_NAME] start`. This way we at least will know what is the exect list of those services and what are their dependencies. After that we can review individual cases
4. `4` should be avoided in general. Every exception should be reviewed carefully in sense of why we can't have such service on elixir side

polvorin:
fwiw,  yesterday' evening tried to integrate with the Authority on rust,   but failed.   Looked at the credential'  issue/retrieval api  @twittner  implemented,   and just implemented that same thing as an Authority on elixir.. to have something to test.

I can show on today's meeting,  but yes the things about having admin creating projects, setting enrollers,  enrollers adding members, member retrieving credentials it's all working.   What to do with that credential once retrieved is a different story :shrug:

PR #2974, opened 2022-07-18
fix(rust): `addr` argument for cloud commands

adrianbenavides:
Removes name flag from `addr` argument so it can be used as in the rest of commands that makes
 use of this attribute: `command /dnsaddr/...` instead of `command --addr /dnsaddr/...`.

Makes the `addr` argument able to accept either a MultiAddr or an Address to make it easier to
test against local secure channels and remote nodes.


mrinalwadhwa:
we should change `addr` to `address`

adrianbenavides:
I'd leave that for another PR because right now, all commands that have an address argument are using `addr` as name. If I do that rename here, we'll have a bunch of commands with `address` and some others with `addr`.

antoinevg (approved):
Awesome, tx for fixing this!

PR #2977, opened 2022-07-18
refactor(rust): rollback `node_name` attribute changes done on node's create command

adrianbenavides:
Changes done to the node's `CreateCommand` attribute to define the node name have been reverted:
* attribute name changed from `api_node` to `node_name`
* removed attribute flag


antoinevg (dismissed):
👍

mrinalwadhwa (approved):

PR #2978, opened 2022-07-18
fix(rust): exit process on command's `embedded_node` error

adrianbenavides:
Commands interacting with background nodes hung when failing because `Context` was never stopped. Output example:

```
Caused by:
    0: failed to load data
    1: failed to load data', implementations/rust/ockam/ockam_command/src/util/mod.rs:58:18
^C
```

This PR fixes that by handling the error's manually instead of bubbling it up, exiting the current process. Now, we show a simple error message to the CLI user, and log the error:

```
encountered an error in command handler code
2022-07-18T17:21:21.585234Z ERROR ockam_command::util: e=Failed to process request
```


antoinevg (dismissed):
👍

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #2993, opened 2022-07-19
feat(rust): derive background node's verbosity from input arguments, defaults to debug

adrianbenavides:
I wanted to record `trace` logs, which are widely used in `ockam_api` but I couldn't set a custom verbosity level to background nodes. With these changes, the default behavior is unchanged, but if you want to modify it, you can just pass the desired level when calling the `create` command:

`ockam create node -vvv` -> will spawn a background node with `trace` verbosity.


antoinevg (approved):
👍

PR #2995, opened 2022-07-19
refactor(rust): auth0 command, extracting auth0 flow out of the node call

adrianbenavides:
This change is needed to avoid timing out while the user is still doing the auth0 flow. Now, once the user is done with the auth0 flow and we receive the token, we send it to the background node and the token is processed right away.

It's also needed to show the `println` calls to the user calling the command showing the information related to the flow (one-time code and status).

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cloud/enroll.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/enroll.rs
@@ -33,23 +33,20 @@ mod node {
     use ockam_node::Context;
 
     use crate::auth::types::Attributes;
-    use crate::cloud::enroll::auth0::{Auth0TokenProvider, AuthenticateAuth0Token};
+    use crate::cloud::enroll::auth0::AuthenticateAuth0Token;
     use crate::cloud::enroll::enrollment_token::{
         AuthenticateEnrollmentToken, EnrollmentToken, RequestEnrollmentToken,
     };
-    use crate::cloud::{BareCloudRequestWrapper, CloudRequestWrapper};
+    use crate::cloud::CloudRequestWrapper;
     use crate::nodes::NodeMan;
-    use crate::request;
+    use crate::{request, CowStr, Id};
     use crate::{Request, Response, Status};
 
     use super::*;
 
     const TARGET: &str = "ockam_api::cloud::enroll";
 
-    impl<A> NodeMan<A>
-    where
-        A: Auth0TokenProvider,
-    {
+    impl NodeMan {
         /// Executes an enrollment process to generate a new set of access tokens using the auth0 flow.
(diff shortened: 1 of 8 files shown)


implementations/rust/ockam/ockam_api/src/cloud/enroll.rs
```diff
@@ -150,11 +150,6 @@ mod node {
 pub mod auth0 {
     use super::*;
 
-    pub const DOMAIN: &str = "dev-w5hdnpc2.us.auth0.com";
-    pub const CLIENT_ID: &str = "sGyXBwQfU6fjfW1gopphdV9vCLec060b";
```
antoinevg, line 154: Was this a secret? If so, did we revoke it?

implementations/rust/ockam/ockam_command/src/enroll/auth0.rs
```diff
@@ -73,6 +76,13 @@ async fn enroll(
 
 pub struct Auth0Service;
 
+impl Auth0Service {
+    const DOMAIN: &'static str = "dev-w5hdnpc2.us.auth0.com";
+    const CLIENT_ID: &'static str = "sGyXBwQfU6fjfW1gopphdV9vCLec060b";
```
antoinevg, line 81: Is this a secret? Should we commit it? If we should, do we have a ticket to revoke it in a timely manner?

implementations/rust/ockam/ockam_command/src/enroll/auth0.rs
```diff
@@ -73,6 +76,13 @@ async fn enroll(
 
 pub struct Auth0Service;
 
+impl Auth0Service {
+    const DOMAIN: &'static str = "dev-w5hdnpc2.us.auth0.com";
```
antoinevg, line 80: It's not great to expose our auth0 domain more than we need to. Can we open a ticket to remove this from the source code in the near future and move it to something we pass in via deployment configuration?

antoinevg (approved):
The auth0.com configuration elements concern me a little as they should probably live in a deployment configuration file or env variable. 

Can you please open an issue for us to do this in the near future?

These have been kicking around in the code for a while though, so no reason not to merge this PR.

adrianbenavides:
> The auth0.com configuration elements concern me a little as they should probably live in a deployment configuration file or env variable.
> 
> Can you please open an issue for us to do this in the near future?
> 
> These have been kicking around in the code for a while though, so no reason not to merge this PR.

Yeah, I talked about this with mrinal a while ago and he told me that we can commit these configuration details for now as they will be removed before the initial release. We'll use a different domain, client, etc, so it should not be a problem.

I have created an issue for this: https://github.com/build-trust/ockam/issues/2997

PR #3031, opened 2022-07-26
Command updates

SanjoDeundiak:


mrinalwadhwa (approved):
@SanjoDeundiak Looks great.
I have a couple of UI suggestions that I think can be handled in separate issues/PRs 

1. `identity print` and `identity export` can be combined into one `identity show` command with slightly different options.
2. I think the output of both of these cases should be `| pbcopy` friendly .. or maybe we need a `--format raw` option
3. `--known-identifier` in secure-channel I think should be called `--authorized-identifier` and it would be nice if we could repeat this option to add multiple identifiers.

PR #3049, opened 2022-07-28
Add a launch configuration parameter for foreground nodes

### LINKED ISSUE #3014: Add commands `ockam node start` and `ockam node stop`
We currently have [node commands](https://github.com/build-trust/ockam/tree/develop/implementations/rust/ockam/ockam_command/src/node)

```
ockam node create green
```

and

```
ockam node delete green
```

We also need

```
ockam node start
```

and

```
ockam node stop
```

so that a node can be started and stopped without losing its identity and other state.

spacekookie:
A small PR to add the ability to configure a foreground node with a launch configuration

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -66,6 +66,7 @@ pub struct NodeManager {
     tcp_transport: TcpTransport,
 
     skip_defaults: bool,
+    foreground: bool,
 
     vault: Option<Vault>,
     identity: Option<Identity<Vault>>,
@@ -81,6 +82,7 @@ impl NodeManager {
         node_name: String,
         node_dir: PathBuf,
         skip_defaults: bool,
+        foreground: bool,
         api_transport: (TransportType, TransportMode, String),
         tcp_transport: TcpTransport,
     ) -> Result<Self> {
@@ -147,6 +149,7 @@ impl NodeManager {
             transports,
             tcp_transport,
             skip_defaults,
+            foreground,
             vault,
             identity,
             authenticated_storage,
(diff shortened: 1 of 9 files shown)


spacekookie:
I see two paths forward for improving the UX of the node configuration/ restart abilities:

**Option A**

1. Separate `NodeCfg` from the main `OckamCfg`, so that each node has a separate file.  This makes writing this file by hand or providing it via different tooling before starting a node much easier.
2. Change the `--launch-state` parameter to a more general `--config-file` parameter, which will be used to setup a node and subsequent commands.

The downside to this approach is that it will be much harder to save other data in the node config (for example identity), if it is being overridden by a launch config.

**Option B**

1. Separate `NodeCfg::composites` into a separate configuration file which is kept per-node
2. Introduce a `--launch-config` parameter which reads a set of instructions to configure a new node with. This could then easily be re-used for both foreground and background nodes.

I think option B is the better approach here but I'm curious what you think. I also don't think this problem is relevant for the `ockam node start` or `ockam node stop` commands.

SanjoDeundiak:
@spacekookie I'll get to reviewing the code tomorrow, but here are my points before that:
1. We need to be able to run start/stop node without extra arguments that would contain state/setup info. I expect it to work docker-like. That means we need to save some state corresponding to that somewhere, probably, in per-node config files. It could be extra field in existing per-node config (I implemented recently, where I store `vault_path` and `identity`). Can be a separate file
2. We need to be able to use `setup` file to create a new node. This is part of `Stage D`. It may be the same as config we store to start already created node, but it may be different (I prefer latter).

spacekookie:
@polvorin @SanjoDeundiak Have a look at this PR again. I changed the launch parameter to be a configuration.
So in the docker image you can just point the CLI at a config like this:

```json
[
   {
      "id" : "<anything>",
      "op" : {
         "address" : "0.0.0.0:6060",
         "mode" : "Listener",
         "protocol" : "Tcp",
         "type" : "Transport"
      },
      "params" : []
   }
]
```

Is this workable for you @polvorin? For reference, these are the commands that can be run this way so far: https://github.com/build-trust/ockam/blob/a847bb9ff5bb326d836bd9a4bf24f2b9abbc7ad6/implementations/rust/ockam/ockam_command/src/util/config/snippets.rs#L39-L66

I would suggest we merge this PR independently of the `ockam node start` command to unblock you.

polvorin:
@spacekookie  for our use case,  it's easier to pass parameters to docker' run , or even env variables, than to have this on a file.  But either case the problem is to make ockam' foreground node pick up vault' existing data after the container is restarted,   we can store it on a persistent volume, but currently we can't pick it up when restarted.

PR #3052, opened 2022-07-29
mrinal/send

mrinalwadhwa:
- feat(rust): add echoer and uppercase workers to command nodes
- feat(rust): make message send command wait for a reply

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -5,7 +5,9 @@ use std::net::SocketAddr;
 use std::{env::current_exe, fs::OpenOptions, process::Command, time::Duration};
 
 use crate::{
+    node::echoer::Echoer,
     node::show::query_status,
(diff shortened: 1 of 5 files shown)


implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -165,5 +167,8 @@ async fn setup(ctx: Context, (c, cfg): (CreateCommand, OckamConfig)) -> anyhow::
     .await?;
     ctx.start_worker(NODEMAN_ADDR, node_man).await?;
 
+    ctx.start_worker("uppercase", Uppercase).await?;
```
SanjoDeundiak, line 170: Not very important, but better place for this is NodeMan, since it's already responsible for starting services (Vault, Identity, Authenticated)

SanjoDeundiak: Also, we'll be improving NodeMan persistence capabilities, so it will remember what it had started, to be able to spawn the same services during node restarts

Issue #3063, opened 2022-07-30
Make `--api-address` option of `ockam node create` default to `127.0.0.1:0`


mrinalwadhwa:
# Current Behavior

```
ockam node create
```

define here https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/node/create.rs

currently has a default `--api-address` value of `127.0.0.1:62526`

This means that the second time `ockam node create` is run a new address must be specified.

# Desired Behavior

We want it to be possible to:

```
ockam node create
ockam node create
ockam node create
ockam node create
```

where a new port is automatically picked for each node.

tmuro17:
Would love to take a crack at this one

mrinalwadhwa:
@tmuro17 awesome, thank you .. a few quick notes:

We can't simply change the value to `127.0.0.1:0`
https://github.com/build-trust/ockam/blob/ee3919eaf07ed0bb6c02171cfa05e9970101a3d8/implementations/rust/ockam/ockam_command/src/util/mod.rs#L16

since the TCP listen is done by the spawned node and the command embedded node needs to know the address to speak with the spawned node

https://github.com/build-trust/ockam/blob/20da0a4112c9ffd6a08ff37d48ec0f544d5018d9/implementations/rust/ockam/ockam_command/src/node/create.rs#L154-L156

The only way I can think of is to create a function that does a TcpListener::bind on `127.0.0.1:0` to find an unused port and then immediately stops listening. Then pass that port along when spawning a node.

tmuro17:
@mrinalwadhwa You answered the question before I had time to look at the code and ask it.  And I like your solution to find an open port.

mrinalwadhwa:
added in https://github.com/build-trust/ockam/pull/3064

PR #3064, opened 2022-07-30
feat(rust): find available port when running `node create`

tmuro17:
find an available port when creating a node when no `api_address` is provided,
so that `ockam node create` can be run multiple times in a row without trying to bind to an existing port

## Current Behavior
`ockam create node` defaults to an address of `127.0.0.1:62526`.  This prevents repeat runs of `ockam create node` from succeeding as it will try to bind to the already in use port of `62526`

## Proposed Changes
Per #3063, find an available port that can be used each time `ockam create node` is run

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -1,14 +1,20 @@
 use clap::Args;
 use rand::prelude::random;
 
-use std::net::SocketAddr;
-use std::{env::current_exe, fs::OpenOptions, process::Command, time::Duration};
+use std::{
+    env::current_exe,
+    fs::OpenOptions,
+    net::{IpAddr, SocketAddr},
+    process::Command,
+    str::FromStr,
+    time::Duration,
+};
 
 use crate::{
     node::echoer::Echoer,
     node::show::query_status,
     node::uppercase::Uppercase,
-    util::{connect_to, embedded_node, OckamConfig, DEFAULT_API_ADDRESS},
+    util::{connect_to, embedded_node, find_available_port, OckamConfig},
     CommandGlobalOpts,
 };
 use ockam::{Context, TcpTransport};
@@ -32,8 +38,8 @@ pub struct CreateCommand {
     skip_defaults: bool,
 
     /// Specify the API address
-    #[clap(default_value = DEFAULT_API_ADDRESS, long, short)]
-    api_address: String,
+    #[clap(long, short)]
+    api_address: Option<String>,
 
     #[clap(long, hide = true)]
     no_watchdog: bool,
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -152,7 +168,9 @@ impl CreateCommand {
 
 async fn setup(ctx: Context, (c, cfg): (CreateCommand, OckamConfig)) -> anyhow::Result<()> {
     let tcp = TcpTransport::create(&ctx).await?;
-    let bind = c.api_address;
+    // We can unwrap `c.api_address` as by this point,
+    // an available port was found, and used to create an api_address in `run`
+    let bind = c.api_address.unwrap();
```
tmuro17: This `unwrap` is a result of changing `api_address` to an Option, which was done because we no longer have a static default value.  While this is sound, I could also leave the default value, then check if `api_address`  is the default value and update the `api_address` accordingly with a new port to be used.  That feels worse to me, but I am open to changing to that strategy if we would prefer to avoid this unwrap.

mrinalwadhwa: What I like about making the default value `127.0.0.1:0` and that showing up in the cli help for `ockam node create` is that it communicates to the user what will happen.

tmuro17: That makes sense to me, will make that change

mrinalwadhwa: Do you want to do that in this PR or in a different followup PR? Since this is about to pass checks, we can merge it if you want.

mrinalwadhwa (dismissed):

implementations/rust/ockam/ockam_command/src/node/create.rs
tmuro17: Whoops, didn't see your comment, included the change in this PR.  (I'm used to my full time job where a change like this would be included in the original PR and not in a follow-up)

mrinalwadhwa: either way works. thank you

mrinalwadhwa (approved):
Just tried it, worked like a charm 🥳 
The changes look great!

mrinalwadhwa:
Thank you for another awesome contribution!!

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -41,10 +47,20 @@ pub struct CreateCommand {
 impl CreateCommand {
     pub fn run(opts: CommandGlobalOpts, command: CreateCommand) {
         let cfg = &opts.config;
-        let address: SocketAddr = command
-            .api_address
-            .parse()
-            .expect("failed to parse api address");
+        let address: SocketAddr = if &command.api_address == "127.0.0.1:0" {
```
SanjoDeundiak, line 50: Should check for the port value here? Ip address can be different (or replaced with domain name)
@mrinalwadhwa @tmuro17

PR #3084, opened 2022-08-01
fix(rust): use runtime handles

### LINKED ISSUE #2996: Diagnose `Cannot drop a runtime ...` error in rust nodes
This error shows up often

```
thread 'tokio-runtime-worker' panicked at 'Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.', /Users/mrinal/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.19.2/src/runtime/blocking/shutdown.rs:51:21
```

---

_If you're looking for other issues to contribute to, checkout issues labeled -  https://github.com/build-trust/ockam/labels/good%20first%20issue or https://github.com/build-trust/ockam/labels/help%20wanted_

twittner:
Hopefully fixes #2996.

### DIFF
--- a/implementations/rust/ockam/ockam_executor/src/runtime.rs
+++ b/implementations/rust/ockam/ockam_executor/src/runtime.rs
@@ -42,14 +42,35 @@ where
 }
 
 /// Runtime
-pub struct Runtime {}
+pub struct Runtime {
+    handle: Handle,
+}
 
 impl Runtime {
     pub fn new() -> io::Result<Runtime> {
-        Ok(Self {})
+        Ok(Self { handle: Handle(()) })
+    }
+
+    pub fn handle(&self) -> &Handle {
+        &self.handle
     }
 
     /// Spawn a future onto the runtime.
+    pub fn spawn<F>(&self, future: F) -> JoinHandle<F::Output>
+    where
+        F: Future + Send + 'static,
+        F::Output: Send + 'static,
+    {
+        executor::current().spawn(future);
+        JoinHandle::new()
+    }
+}
+
+/// Runtime handle
+#[derive(Clone)]
+pub struct Handle(());
+
+impl Handle {
     pub fn spawn<F>(&self, future: F) -> JoinHandle<F::Output>
     where
         F: Future + Send + 'static,
--- a/implementations/rust/ockam/ockam_node/src/async_drop.rs
+++ b/implementations/rust/ockam/ockam_node/src/async_drop.rs
@@ -1,10 +1,9 @@
-use crate::tokio::runtime::Runtime;
 use crate::tokio::sync::{
     mpsc::Sender as DefaultSender,
     oneshot::{self, Receiver, Sender},
 };
 use crate::NodeMessage;
-use ockam_core::{compat::sync::Arc, Address};
+use ockam_core::Address;
 
 /// A helper to implement Drop mechanisms, but async
 ///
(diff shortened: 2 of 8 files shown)


mrinalwadhwa:
I was able to reproduce the error in #2996, on  this branch, with my very crude test:

```
for i in {1..9}; do RUST_BACKTRACE=1 cargo run --bin ockam node create "n$i"; done
for i in {1..100}; do RUST_BACKTRACE=1 cargo run --quiet --bin ockam node list; done
```

pro465:
hey, @mrinalwadhwa! so you somehow still got the panics?

can i get the backtrace?

mrinalwadhwa:
Trying a few runs again, with a `cargo clean`... just to be sure. Let's see.

pro465:
can i get the backtrace if you the error again?

pro465:
@twittner quite elegant (potential) solution btw! runtimes being dropped in async contexts => use static runtimes and leak them. such simplicity! :chefs_kiss:

mrinalwadhwa:
Had to interrupt this run but things are looking quite promising.
Will try another run later in the day.

```
» for j in {1..50}; do for i in {1..50}; do RUST_BACKTRACE=1 cargo run --quiet --bin ockam node list 1>/dev/null; echo -n "."; done; echo ""; done
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..................................................
..............................

pro465:
@mrinalwadhwa so no panics?

mrinalwadhwa:
yes so far none.

pro465:
that's good, hooray!

question: should not the `Handle::spawn` method call's returned JoinHandle block forever causing problems?

edit: and `Runtime::spawn` too?

pro465:
@mrinalwadhwa you seem to be using `--quiet` flag which caused us problems before, perhaps you want to remove it?

mrinalwadhwa:
Just finished running 2500 runs of `cargo run --bin ockam node list` with no failures 🎉

antoinevg (dismissed):
"It is said there is no problem in computer science which cannot be solved by [one more level of indirection](https://wiki.c2.com/?OneMoreLevelOfIndirection)."

This is really good work. 👍

twittner:
> @pro465: [...] static runtimes and leak them

I do not think the runtime is leaked. It is owned by `ockam_node::Executor` and dropped when the executor goes out of scope (which should usually happen in sync code, hence the panic no longer occurs).

> @pro465: question: should not the Handle::spawn method call's returned JoinHandle block forever causing problems? [...] and Runtime::spawn too?

I think so. This part is for no_std environments and I only added the Handle-part to be API compatible. Changes to the `JoinHandle` are out of scope for this PR I would say.

pro465:
> @twittner: I do not think the runtime is leaked.

i meant the runtime in no_std environments.

pro465:
sorry to bother you but i have another question: how is `Handle` different from `Arc<Runtime>`?

the docs say:
> The handle is internally reference-counted and can be freely cloned. A handle can be obtained using the [Runtime::handle](https://docs.rs/tokio/1.20.1/tokio/runtime/struct.Runtime.html#method.handle) method.

so how does just using Handle fix the problem?

twittner:
> @pro465: how is `Handle` different from `Arc<Runtime>`?

The `Runtime` contains a `BlockingPool` which it shuts down when dropped which is itself a blocking operation that panics if used in an async context. Dropping the `Handle` OTOH does not cause this shutdown.

antoinevg (dismissed):

pro465:
@twittner oh so that's why, i wondered why would they behave differently, so thank you for clearing it up!

mrinalwadhwa:
@pro465 Thank you so much for all the research you put into this over the last couple of weeks!!
@twittner Thank you for the fix!! and also thank you for adding attribution to @pro465 in the commit https://github.com/build-trust/ockam/pull/3084/commits/8bd0f21ad4248e8ccab3208b35b1b500bebef63d

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

pro465:
@mrinalwadhwa thank you so much for the encouraging and kind messages!

also @twittner congrats for fixing another issue 🎉 ! Thank you for the effort and also for the attribution! 
(which, well i don't know if i deserve that, because idk if it helped you in fixing this...)

Issue #2836, opened 2022-06-07
Investigate CI Rust Nightly check ran for ~6hrs


mrinalwadhwa:
https://github.com/build-trust/ockam/runs/6769053318?check_suite_focus=true

mrinalwadhwa:
This happened again
https://github.com/build-trust/ockam/runs/7264866857?check_suite_focus=true

mrinalwadhwa:
Another case:
https://github.com/build-trust/ockam/runs/7265288541?check_suite_focus=true

It seems to be the same test everytime:
```
test cloud::project::tests::basic_api_usage has been running for over 60 seconds
```

mrinalwadhwa:
It's one of these tests
https://github.com/build-trust/ockam/search?q=basic_api_usage

mrinalwadhwa:
@adrianbenavides would love your thoughts on why this may be happening.

adrianbenavides:
I'll be working on these tests this week, so I'll make sure to check this out!

adrianbenavides:
This issue can be closed after the following issues are solved:
* https://github.com/build-trust/ockam/issues/2996
* https://github.com/build-trust/ockam/issues/3054

PR #3093, opened 2022-08-01
Repeat issue 3054 tests hang on panic

### LINKED ISSUE #3054: Make async tests not hang if test function panics
Currently async tests just hang whenever panic occurs.
Possible solution may be using [catch_unwind](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html). Below you can find a working prototype

```rs
async fn test_fn(mut ctx: Mutex<Context>) -> Result<()> {
    let _ctx = ctx.get_mut().unwrap();
    panic!("PANIC")
}

#[test]
fn test() {
    use core::time::Duration;
    use ockam_node::{compat::tokio::time::timeout, NodeBuilder};

    let (mut ctx, mut executor) = NodeBuilder::without_access_control().build();
    executor
        .execute(async move {
            let child_ctx = Mutex::new(ctx.new_detached(Address::random_local()).await.unwrap());
            match timeout(Duration::from_millis(1000), test_fn(child_ctx).catch_unwind()).await {
                Ok(r) => match r {
                    Ok(r) => match r {
                        Err(err) => {
                            ctx.stop().await.unwrap();
                            Err(err)
                        }
                        Ok(_) => Ok(()),
                    },
                    Err(_) => {
                        ctx.stop().await.unwrap();
                        panic!("Test panic")
                    }
                },
                Err(_) => {
                    ctx.stop().await.unwrap();
                    panic!("Test timeout")
                }
            }
        })
        .expect("Executor should not fail")
        .expect("Test function should not fail");
}

```

Retamogordo:
## Current Behavior

Fixes https://github.com/build-trust/ockam/issues/3054 and potentially https://github.com/build-trust/ockam/issues/2836

## Proposed Changes

Proc macro in implementations/rust/ockam/ockam_macros/src/node_test_attribute.rs was modified. I also changed tests.rs and delayed.rs files since I needed to for testing the modified macro.
Also a couple of failing pattern demo tests were added to tests.rs, they begin with 'empty_' prefix.
Some observations:
1) There are tests that still hang, all of them repeat the pattern of empty_macro_test_double_ctx_stop(...), namely the calling of ctx.stop() inside their body, which leads to ctx.stop() to be called twice. 
I was unable to discover whether in this case the second invocation leads to hanging inside the runtime, or it is the result of panicking after all. 
I wrapped  ctx.stop() with catch_rewind in the macro, to no avail.
2) The async version of simple_worker__run_node_lifecycle__worker_lifecycle_should_be_full
now fails, there is a suspicious line 
std::thread::sleep(Duration::new(1, 0))
which is a sync call, the async version of this didn't help to fix the failure.
I left the original version commented out so it can be easily reverted.
3) I left 
start_and_shutdown_node__many_iterations__should_not_fail
intact, since  there is an outer loop there, so I did not wanted to change it in order to not change the test's semantics.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_macros/src/node_test_attribute.rs
+++ b/implementations/rust/ockam/ockam_macros/src/node_test_attribute.rs
@@ -70,7 +70,7 @@ fn output(mut cont: Container) -> TokenStream {
             quote! {#ident}
         }
     };
-    let ctx_stop_stmt = quote! { let _ = #ctx_ident.stop().await; };
+    let ctx_stop_stmt = quote! { #ctx_ident.stop().await };
     let test_fn = &cont.test_fn;
     let test_fn_ident = &cont.test_fn.sig.ident;
     let ockam_crate = cont.data.attrs.ockam_crate;
@@ -78,27 +78,58 @@ fn output(mut cont: Container) -> TokenStream {
     cont.original_fn.block = parse2(quote! {
         {
             use core::time::Duration;
-            use #ockam_crate::{NodeBuilder, compat::tokio::time::timeout};
+            use #ockam_crate::{Context, NodeBuilder, compat::tokio::time::timeout};
+            use #ockam_crate::compat::futures::FutureExt;
+
+            async fn stop_map_err(ctx: &mut Context, res: ockam_core::Result<()>) -> ockam_core::Result<()> {
+                match res {
+                    Ok(_) => Ok(()),
+                    Err(err) => {
+                        // ?? ignoring possible error returned by ctx.stop()
+                        let _ = std::panic::AssertUnwindSafe(async { ctx.stop().await.unwrap(); })
+                            .catch_unwind()
+                            .await;
+                        Err(err)
+                    }
+                }
+            }
 
             let (mut #ctx_ident, mut executor) = NodeBuilder::without_access_control().build();
             executor
                 .execute(async move {
-                    match timeout(Duration::from_millis(#timeout_ms), #test_fn_ident(&mut #ctx_ident)).await {
-                        Ok(r) => match r {
-                            Err(err) => {
-                                #ctx_stop_stmt
-                                Err(err)
-                            },
-                            Ok(_) => Ok(())
-                        },
-                        Err(_) => {
-                            #ctx_stop_stmt
-                            panic!("Test timeout")
+                    let catch_res = std::panic::AssertUnwindSafe(async {
+
+                        let test_result
+                                = timeout(Duration::from_millis(#timeout_ms), #test_fn_ident(&mut #ctx_ident))
+                            .await
+                            .map_err(|_| ockam_core::Error::new(ockam_core::errcode::Origin::Node,
+                                                    ockam_core::errcode::Kind::Timeout, "Test timeout") )
+                            // unwrap timeout result to get inner test result
+                            .unwrap();
+                        // stop ctx and pipe result back
+                        stop_map_err(&mut #ctx_ident, test_result).await
+                    })
+                    .catch_unwind()
+                    .await;
+                    // unwrap result from catch_unwind
+                    match catch_res {
+                        // not panicked, get test result
+                        Ok(test_result) => test_result,
+
+                        Err(test_panic_cause) => {
+                            // ?? ignoring possible error returned by ctx.stop() in sake of test_panic_cause
+                            let _ = #ctx_stop_stmt;
+                            // NOTICE: still panicking in an async context after catch_unwinding
+                            // could cause hanging ?
+                            // would be better to return Err ?, like so:
+                            // ockam_core::Result::<()>::Err("Test panicked".into())
+                            // however no conversion from &str to Err is there
+                            panic!("Test panicked, caused by {:?}", test_panic_cause);
                         }
                     }
                 })
                 .expect("Executor should not fail")
-                .expect("Test function should not fail");
+                .expect("Test function should not fail")
         }
     }).expect("Parsing failure");
     let input_fn = &cont.original_fn;
--- a/implementations/rust/ockam/ockam_node/src/delayed.rs
+++ b/implementations/rust/ockam/ockam_node/src/delayed.rs
@@ -80,7 +80,7 @@ impl<M: Message + Clone> DelayedEvent<M> {
 
 #[cfg(test)]
 mod tests {
-    use crate::{Context, DelayedEvent, NodeBuilder};
+    use crate::{Context, DelayedEvent};
     use core::sync::atomic::Ordering;
     use core::time::Duration;
     use ockam_core::compat::{boxed::Box, string::ToString, sync::Arc};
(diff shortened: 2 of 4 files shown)


mrinalwadhwa:
Thank you so much for taking a second run at this 🙌

Retamogordo:
Hope will be useful this time :)

implementations/rust/ockam/ockam_macros/src/node_test_attribute.rs
```diff
@@ -79,26 +79,30 @@ fn output(mut cont: Container) -> TokenStream {
         {
             use core::time::Duration;
             use #ockam_crate::{NodeBuilder, compat::tokio::time::timeout};
+            use futures::FutureExt;
```
SanjoDeundiak: I think the problem with this import is that not every crate has `futures` crate as a dependency. How about replacing it with `use #ockam_crate::compat::futures::FutureExt;`. To make this work such module should be added to `ockam_node/src/compat/mod.rs` like this
```
pub mod futures {
    pub use futures::FutureExt;
}
```

implementations/rust/ockam/ockam_node/src/tests.rs
```diff
@@ -66,7 +76,7 @@ impl Worker for SimpleWorker {
         ctx.send(msg.return_route(), msg.body()).await
     }
 }
-
+/*
```
SanjoDeundiak: Could you please remove that in case test below does the same?

implementations/rust/ockam/ockam_macros/src/node_test_attribute.rs
```diff
@@ -79,26 +79,30 @@ fn output(mut cont: Container) -> TokenStream {
...
+
+                        timeout(Duration::from_millis(#timeout_ms), #test_fn_ident(&mut #ctx_ident))
+                            .await
+                            .map_err(|_| panic!("Test timeout"))
+                            .unwrap()
+                    })
+                    .catch_unwind()
+                    .await;
+                    
+                    std::panic::AssertUnwindSafe(async { #ctx_stop_stmt; })
```
SanjoDeundiak: I think we only need to run this if test function:
1. Panics
2. Returns an error
3. Timeouts
In other scenarios test function stops the context itself

implementations/rust/ockam/ockam_node/src/delayed.rs
```diff
@@ -110,166 +110,122 @@ mod tests {
...
+            .await?;
+        sleep(Duration::from_millis(150)).await;
+        heartbeat
+            .schedule(Duration::from_millis(200))
+            .await?;
+        sleep(Duration::from_millis(100)).await;
+        heartbeat.cancel();
+        sleep(Duration::from_millis(300)).await;
+
+        assert_eq!(1, msgs_count.load(Ordering::Relaxed));
```
SanjoDeundiak, line 180: `ctx.stop().await.unwrap();`

implementations/rust/ockam/ockam_node/src/delayed.rs
```diff
@@ -110,166 +110,122 @@ mod tests {
...
+            .await?;
+        heartbeat
+            .schedule(Duration::from_millis(100))
+            .await?;
+        heartbeat
+            .schedule(Duration::from_millis(100))
+            .await?;
+        sleep(Duration::from_millis(150)).await;
+
+        assert_eq!(1, msgs_count.load(Ordering::Relaxed));
```
SanjoDeundiak, line 156: `ctx.stop().await.unwrap();`

implementations/rust/ockam/ockam_node/src/delayed.rs
```diff
@@ -110,166 +110,122 @@ mod tests {
...
+        heartbeat
+            .schedule(Duration::from_millis(100))
+            .await?;
+        sleep(Duration::from_millis(150)).await;
+        heartbeat
+            .schedule(Duration::from_millis(100))
+            .await?;
+        sleep(Duration::from_millis(150)).await;
+
+        assert_eq!(3, msgs_count.load(Ordering::Relaxed));
```
SanjoDeundiak, line 134: I think , `ctx.stop().await.unwrap();` is missing from rewritten test could you please fix that?

implementations/rust/ockam/ockam_node/src/delayed.rs
```diff
@@ -110,166 +110,122 @@ mod tests {
...
+            .await?;
+        sleep(Duration::from_millis(150)).await;
+        heartbeat
+            .schedule(Duration::from_millis(200))
+            .await?;
+        sleep(Duration::from_millis(100)).await;
+        drop(heartbeat);
+        sleep(Duration::from_millis(300)).await;
+
+        assert_eq!(1, msgs_count.load(Ordering::Relaxed));
```
SanjoDeundiak, line 204: `ctx.stop().await.unwrap();`

implementations/rust/ockam/ockam_node/src/tests.rs
```diff
@@ -147,27 +196,20 @@ fn counting_processor__run_node_lifecycle__processor_lifecycle_should_be_full()
...
+        initialize_was_called: initialize_was_called_clone,
+        shutdown_was_called: shutdown_was_called_clone,
+        run_called_count: run_called_count_clone,
+    };
 
-            assert!(initialize_was_called.load(Ordering::Relaxed));
-            assert!(shutdown_was_called.load(Ordering::Relaxed));
-            assert_eq!(5, run_called_count.load(Ordering::Relaxed));
+    ctx.start_processor("counting_processor", processor)
+        .await?;
```
SanjoDeundiak: `ctx.stop().await.unwrap();`

implementations/rust/ockam/ockam_node/src/tests.rs
```diff
@@ -103,6 +121,37 @@ fn simple_worker__run_node_lifecycle__worker_lifecycle_should_be_full() {
...
+    };
+
+    ctx.start_worker("simple_worker", worker).await?;
+
+    ctx.send(route!["simple_worker"], "Hello".to_string())
+        .await?;
+
+    let msg = ctx.receive::<String>().await.unwrap().take().body();
+    assert_eq!(msg, "Hello");
+
```
SanjoDeundiak, line 100: `ctx.stop().await.unwrap();`

implementations/rust/ockam/ockam_node/src/tests.rs
```diff
@@ -202,38 +244,30 @@ impl Processor for WaitingProcessor {
...
+        .await?;
+    sleep(Duration::new(1, 0)).await;
 
+    ctx.stop_processor("waiting_processor").await?;
     // Wait till tokio Runtime is shut down
     std::thread::sleep(Duration::new(1, 0));
 
     assert!(initialize_was_called.load(Ordering::Relaxed));
     assert!(shutdown_was_called.load(Ordering::Relaxed));
+    Ok(())
```
SanjoDeundiak: `ctx.stop().await.unwrap();`

implementations/rust/ockam/ockam_node/src/tests.rs
```diff
@@ -280,49 +314,41 @@ impl Processor for MessagingProcessor {
...
-            assert_eq!(
-                "I go home",
-                ctx.receive::<String>().await.unwrap().take().body()
-            );
+    ctx.send(route!["messaging_processor"], "Stop working".to_string())
+        .await?;
+    assert_eq!(
+        "I go home",
+        ctx.receive::<String>().await.unwrap().take().body()
+    );
```
SanjoDeundiak, line 302: `ctx.stop().await.unwrap();`

implementations/rust/ockam/ockam_node/src/tests.rs
```diff
@@ -491,6 +522,7 @@ fn worker_calls_stopworker_from_handlemessage() {
         counter_a.load(Ordering::Relaxed),
         counter_b.load(Ordering::Relaxed)
     );
+    Ok(())
```
SanjoDeundiak: `ctx.stop().await.unwrap();`

implementations/rust/ockam/ockam_node/src/tests.rs
```diff
@@ -527,23 +559,46 @@ enum SendReceiveResponse {
...
+    let msg_tx = SendReceiveRequest::Connect();
+    let msg_rx = ctx.send_and_receive("SendReceiveWorker", msg_tx).await?;
 
-            if let SendReceiveResponse::Connect(Err(e)) = msg_rx {
-                panic!("test failure: {}", e)
-            }
+    if let SendReceiveResponse::Connect(Err(e)) = msg_rx {
+        panic!("test failure: {}", e)
+    }
+    Ok(())
```
SanjoDeundiak: `ctx.stop().await.unwrap();`

implementations/rust/ockam/ockam_node/src/tests.rs
```diff
@@ -527,23 +559,46 @@ enum SendReceiveResponse {
...
+}  
 
-            ctx.stop().await
-        })
-        .unwrap()
-        .unwrap();
+#[ockam_macros::test(crate = "crate")]
+async fn empty_macro_test_ok(_ctx: &mut Context) -> Result<()> {
+    Ok(())
```
SanjoDeundiak: `ctx.stop().await.unwrap();`

SanjoDeundiak (changes requested):
Nice job. Left few comments.
P.S. could you please also edit commit message and  run `cargo fmt` + `cargo test`?

SanjoDeundiak:
@Retamogordo commit message guide is [here](https://github.com/build-trust/.github/blob/main/CONTRIBUTING.md#commits)

implementations/rust/ockam/ockam_macros/src/node_test_attribute.rs
Retamogordo: Sure,  I'll see your remarks a bit later.

implementations/rust/ockam/ockam_macros/src/node_test_attribute.rs
Retamogordo: I thought that ctx must be stopped as a fallback no matter how the test ends, kinda 'catch-finally' semantics. So the test business logic would be free of concerning about ctx stopping?

implementations/rust/ockam/ockam_node/src/delayed.rs
Retamogordo: Ok, I'll check all mentioned,  now traveling 👍

implementations/rust/ockam/ockam_macros/src/node_test_attribute.rs
SanjoDeundiak: @Retamogordo I think before your changes we didn't call `ctx.stop()` if test function successfully returns `Ok(())`. Running `ctx.stop` in that case as a part of macro has its own pros&cons, but I think we don't want to change that as a part of this PR

Retamogordo: Right, I'll revert it back to its original logic.

Retamogordo: Another point I'm still not sure about is if I should assume that ctx.stop() may panic inside? Namely, if there is a need to wrap ctx.stop with catch_unwind as I do now ?

adrianbenavides: We've had panics happening when executing ctx.stop() in the past, so I'd wrap that also with catch_unwind, yes.

SanjoDeundiak: I think panic in ctx.stop() is undefined behaviour. Doesn't matter if it's wrapped on not

adrianbenavides:
One quick note @Retamogordo, if you squash all your commits and write something like `fix(rust): tests hang on inner panics`, it should pass the CI check.

implementations/rust/ockam/ockam_macros/src/node_test_attribute.rs
```diff
@@ -70,35 +70,66 @@ fn output(mut cont: Container) -> TokenStream {
...
+                            // however no conversion from &str to Err is there
+                            panic!("Test panicked, caused by {:?}", test_panic_cause);
                         }
                     }
                 })
                 .expect("Executor should not fail")
-                .expect("Test function should not fail");
+                .expect("Test function should not fail")
         }
     }).expect("Parsing failure");
```
adrianbenavides, line 134: I think you can simplify this a lot. Check this out and let me know what you think!
```rust
cont.original_fn.block = parse2(quote! {
    {
        use core::panic::AssertUnwindSafe;
        use core::time::Duration;
        use ockam_core::{Error, errcode::{Origin, Kind}};
        use #ockam_crate::{NodeBuilder, compat::{tokio::time::timeout, futures::FutureExt}};

        let (mut #ctx_ident, mut executor) = NodeBuilder::without_access_control().build();
        executor
            .execute(async move {
                // Wraps the test function call in a `catch_unwind` to catch possible panics.
                match AssertUnwindSafe(async {
                    match timeout(Duration::from_millis(#timeout_ms), #test_fn_ident(&mut #ctx_ident)).await {
                        Ok(r) => match r {
                            // Test went well, return result as is.
                            Ok(_) => Ok(()),
                            Err(e) => {
                                #ctx_stop_stmt
                                Err(e)
                            }
                        },
                        Err(_) => {
                            // Test timed out, so we return a custom error that we can handle.
                            #ctx_stop_stmt
                            Err(Error::new(Origin::Node, Kind::Timeout, "Test timed out"))
                        }
                    }
                })
                .catch_unwind()
                .await
                {
                    Ok(res) => res,
                    Err(_) => {
                        ctx.stop().await?;
                        panic!("Test panicked");
                    }
                }
            })
            .expect("Test panicked")
            .expect("Test function returned error");
    }
}).expect("Parsing failure");
```

implementations/rust/ockam/ockam_node/src/tests.rs
```diff
@@ -527,23 +495,71 @@ enum SendReceiveResponse {
...
+    ctx.start_worker("dummy_worker", DummyWorker).await?;
+    panic!("Will hang on panic ?");
+}
+
+#[ockam_macros::test(crate = "crate", timeout = 0)]
+#[should_panic]
+async fn empty_macro_test_timed_out(ctx: &mut Context) -> Result<()> {
+    ctx.start_worker("dummy_worker", DummyWorker).await?;
+    sleep(Duration::from_millis(100)).await;
+    Ok(())
```
adrianbenavides, line 564: Tests for the different macros are located here: `implementations/rust/ockam/ockam/tests`. It would be great to move the tests you wrote there. Also, you don't need the `DummyWorker` nor start a worker, it's OK if you don't use the `ctx` on a given test.

adrianbenavides (changes requested):
Great job @Retamogordo ! I left a couple of comments with some suggestions.

implementations/rust/ockam/ockam_macros/src/node_test_attribute.rs
Retamogordo: Hi @adrianbenavides, 
as for your code suggestion,
the problem here is that we need to deal with both Results::Errs and inner panics. So whenever the test panics inside, catch_rewind converts it into an Error. So your code line if let Err(err)... actually checks if the test panicked, but if the test returns an error, catch_rewind returns Ok. So we need to unwrap it and check the inner Result.
The second point is that you suggest to re-panic upon catch_rewind error thus getting back to the same situation when a code running in async context panics again.
I tried your code, it hangs whenever the inner test wants to return an Error.
As for commiting messages, it is still terra incognita for me, but I'm trying hard :)

mrinalwadhwa: > it is still terra incognita for me, but I'm trying hard :)

😀 don't worry too much about it, we can correct it for you just before landing the PR.

adrianbenavides: Ohh I see, I didn't understand how `catch_unwind` worked then! I'm going to try something else... now I'm curious :laughing:

Retamogordo: Yes, there are a couple of corner cases:
1) no implicit conversion from timeout::Result to ockam::Result, I needed to map the error
2) it would be nice to get rid of explicit panicking in that snippet while still in async block. My snippet still has that flaw, because I could not find which kind of ockam::Error to return in this case. I left a long comment in that block. Need your suggestion here.
3) yesterday you said that we must assume that ctx.stop may panic too, so the code deals with it
4) not directly related to this issue, rather to #3105, there are scenarios when the code hangs without panicking if ctx.stop called twice. When tha tissue is resolved, probably this code also may be simplified.

adrianbenavides: I've updated the snippet from my previous comment https://github.com/build-trust/ockam/pull/3093#discussion_r937578186 with basically your solution but without that extra function. I think this will fix most of the cases where we have those flaky tests hanging in CI, so we can leave it here and see how it goes.

Retamogordo: Ok, just that regarding explicit panicking, could it be converted into some kind of Error? Otherwise, I'm afraid, we are still in the same mine camp as before. 
So, as for this issue, I just leave it as is ?

Retamogordo: Hi @adrianbenavides !
Please let me know if any action is expected from me in order to close this issue. I am not yet experienced with community contribution process on github, so sometimes I get a bit confused.
I understood, that you took control on it, right ?

mrinalwadhwa: @Retamogordo you did great with the process 👏 👏 
I squashed your commits, rebased and amended the commit message.
When @adrianbenavides @SanjoDeundiak are back online in a some hours, we can merge after their final 👍

SanjoDeundiak:
@mergifyio rebase

SanjoDeundiak (approved):
Huuge improvement! Tnx @Retamogordo

adrianbenavides (approved):
LGTM

PR #3098, opened 2022-08-02
Ockam configuration module refactor

spacekookie:
Currently `ockam_command` and `ockam_api` have two separate configuration modules. This PR moves all config structs to `ockam_api` while keeping command-specific utility functions separate.

This PR also adds a separate `startup.json` config which is updated with every command run, in order to populate a restart script.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2132,6 +2132,7 @@ version = "0.10.0"
 dependencies = [
  "anyhow",
  "cddl-cat",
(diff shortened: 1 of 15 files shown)


antoinevg (dismissed):
I like where this is going! 👍

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3099, opened 2022-08-02
fix(rust): remove `block_future`

twittner:
`block_future` is only used by `Executor::execute` which is itself a blocking method and not invoked from an async context (nor should it), hence it seems unnecessary to use tokio's task::block_in_place.


antoinevg (dismissed):
🎉

mrinalwadhwa (dismissed):

mrinalwadhwa (dismissed):

BeenzSyed (approved):
Looks good to me!

PR #3104, opened 2022-08-03
feat(rust): project node identity

polvorin:
@adrianbenavides  I updated the project' schema to reflect the identity_id field that is already available.   You can take over from here to update the rust code.

Also, all invitation related schema and code should be removed (but could be a separate issue on separate PR),  as that invitation workflow was removed.


adrianbenavides:
I'm on it!

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3106, opened 2022-08-04
fix(rust): space create command when list args are passed

adrianbenavides:
Same as 2963e3d9 (https://github.com/build-trust/ockam/pull/3096) but for space commands.


antoinevg (approved):
👍

PR #3107, opened 2022-08-04
refactor(rust): remove invitations code

adrianbenavides:
As stated here: https://github.com/build-trust/ockam/pull/3104 I'm removing all code related to invitations from commands and api.


polvorin (dismissed):
:+1:

mrinalwadhwa:
@adrianbenavides @polvorin I would like to explore if both inviting people into a project and inviting other members into a project can be the same set of commands for users with slight difference in arguments.

```
ockam invitation create/accept/reject
```

is very similar to:

```
ockam project add-member/new-credential
```

that @twittner proposed [here](https://github.com/build-trust/codebase/issues/324)

so maybe we should keep the command facade for invitations

polvorin:
@mrinalwadhwa  
these are different enough imho:

* When an _admin_  "invite" someone into a project (as: you are now an admin of this project)  its  inviting an _user_  , identified by his _email_.   That user can use whatever (authenticated) identity_id it has or will have,  to admin that project  as long as it authenticate to that email.  

* When and _enroller_  add a _member_ ,  it do that on the project' node (authority).   Its add an identity_id , not an _user_  as member of the project.   These identity_id  could for example be for machines that the enroller is  provisioning.     Nothing prevents an enroller to to add an identity_id that correspond to some _user_   as a member too, of course.  

The adding/listing/removal of admins from projects and spaces is available but I think not exposed right now, via  `put`  request on the specific project/space.

mrinalwadhwa:
@polvorin good points.
I think there is sufficient similarity in the intent of the commands even though the inputs and protocols run are different. Let's merge this for now and keep exploring what the ideal commands are for the two cases.

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3108, opened 2022-08-04
Correctly drop router receiver after shutdown

### LINKED ISSUE #3105: Dangling channel receiver in Router
emerged while resolving #3054 
- Reproduce
```
    info!("*********************** before create detached ****************"); 
    let mut detached = ctx.new_detached(Address::random_local()).await?;
    info!("*********************** after create detached ****************"); 

    info!("----------------------- before stopping detached-----------------"); 
    detached.stop().await?; 
    info!("----------------------- after stopping detached -----------------"); 

    sleep(Duration::from_secs(2)).await;

    info!("*********************** before create start worker ****************"); 
    ctx.start_worker("dummy", DummyWorker).await?;
// never gets here
    info!("*********************** after create start worker ****************"); 

```
or just stop context twice:
```
    ctx.stop().await?; 
    ctx.stop().await?; // hangs
```

- Possible cause
	When stopping context a shutdown request is headed to Router, as a result the Router exits its event loop sending ack back to the context.
	For the second stopping, there is no longer a listening receiver on the routers end, however the receiver has not been dropped.
	Tokio mpsc channel's docs say that the sender returns error if the other end is dropped, however here this is not the case, so tx passes smoothly and the context still waits for the ack, which is not going to arrive, so the thread hangs.
	This gets more complex as there are such entities as detached context and workers which share the same cloned sender.
	So any of them can cause the Router to shutdown the listening loop thus all other entities will wait for acks from the loop which no longer runs.

- Remarks
I am totally unfamiliar with Ockam so my apologies if it is a false alarm or a known thing.

spacekookie:
This commit (potentially) fixes #3105.  Because the router run-loop does not own the Router, it is not dropped after the loop terminates.

Note: the commit isn't signed because I forgot my Yubikey at home.  I will add a signature later.

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/router/mod.rs
+++ b/implementations/rust/ockam/ockam_node/src/router/mod.rs
@@ -46,7 +46,7 @@ pub struct Router {
     /// Externally registered router components
     external: BTreeMap<TransportType, Address>,
     /// Receiver for messages from node
-    receiver: RouterReceiver<NodeMessage>,
+    receiver: Option<RouterReceiver<NodeMessage>>,
 }
 
 enum RouteType {
@@ -69,13 +69,20 @@ impl Router {
             state: RouterState::new(sender),
             map: InternalMap::default(),
             external: BTreeMap::new(),
-            receiver,
+            receiver: Some(receiver),
         }
     }
 
     #[cfg(feature = "metrics")]
     pub(crate) fn get_metrics_readout(&self) -> (Arc<AtomicUsize>, Arc<AtomicUsize>) {
         self.map.get_metrics()
+    }
+
+    /// Get the router receiver
+    fn get_recv(&mut self) -> Result<&mut RouterReceiver<NodeMessage>> {
+        self.receiver
+            .as_mut()
+            .ok_or_else(|| NodeError::NodeState(NodeReason::Corrupt).internal())
     }
 
     pub fn init(&mut self, addr: Address, senders: SenderPair) {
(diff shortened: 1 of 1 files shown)


implementations/rust/ockam/ockam_node/src/router/mod.rs
```diff
@@ -269,11 +276,13 @@ impl Router {
...
     /// Block current task running this router.  Return fatal errors
     async fn run_inner(&mut self) -> Result<()> {
-        while let Some(msg) = self.receiver.recv().await {
+        while let Some(msg) = self.get_recv()?.recv().await {
             let msg_str = format!("{}", msg);
             match self.handle_msg(msg).await {
                 Ok(should_break) => {
                     if should_break {
+                        // We drop the receiver end here
+                        self.receiver.take();
```
twittner, line 304: Rather than dropping the receiver I recommend calling `Receiver::close` instead. This way there is no need to have it as an `Option`al field in `Router`.

spacekookie:
Oh @twittner `Receiver::close` isn't implemented in `no_std`?? @antoinevg 

I can just go back to wrapping it in an Option but this is a bit strange

antoinevg:
> Oh @twittner `Receiver::close` isn't implemented in `no_std`?? @antoinevg
> 
> I can just go back to wrapping it in an Option but this is a bit strange

We can probably switch to `futures::channel` on `no_std` rather than using our own impl. I'll take a look at this and open a PR if it's do-able.

spacekookie:
I just changed the code to be as it was before, let's just merge this option first then so we can see if it makes a difference in how often we have CI failures

twittner:
> [...] `Receiver::close` isn't implemented in `no_std`?? [...]

For the time being it is sufficient to add a stub to https://github.com/build-trust/ockam/blob/2ea26c64036e61669181f17c92b0906a6a434a79/implementations/rust/ockam/ockam_executor/src/channel.rs#L139 as I did here: https://github.com/build-trust/ockam/commit/d6b68e48d25167b9d68f6a81bd9982770536115a

antoinevg:
> We can probably switch to `futures::channel` on `no_std` rather than using our own impl. I'll take a look at this and open a PR if it's do-able.

Ah never mind, `futures::channel::mpsc` is only available on `std`.

Just adding a stub for `close()` is fine as the executor is single-threaded so there's no need to notify anything.

spacekookie:
@twittner either merge this PR or close it and open a PR with your solution. I'm not rebasing it again

SanjoDeundiak (approved):

PR #3111, opened 2022-08-04
Node restart

spacekookie:
WIP node restart mechanism.

* `ockam node start [node_name]` exists and works
* **but** only when re-starting a base node with no additional configuration.

In theory other configuration commands should be supported (for transport & portals), though for some reason the commands hang when run from the startup sequence


mrinalwadhwa:
@spacekookie this introduces an MPL licensed dependency. We can't use that. https://github.com/build-trust/ockam/runs/7727728134?check_suite_focus=true

spacekookie:
Rebased again against current `develop` @build-trust/approvers please review and get this merged today

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3126, opened 2022-08-05
fix(rust): Check address before creating stuff

### LINKED ISSUE #3061: Router doesn't check if address is already taken while starting Worker/Processor

pro465:
## Current Behavior

described in #3061 

## Proposed Changes

fixes #3061.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/router/mod.rs
+++ b/implementations/rust/ockam/ockam_node/src/router/mod.rs
@@ -17,7 +17,7 @@ use crate::channel_types::{router_channel, MessageSender, RouterReceiver, SmallS
 use crate::{
     error::{NodeError, NodeReason},
     relay::{CtrlSignal, RelayMessage},
-    NodeMessage, RouterReply, ShutdownType,
+    NodeMessage, NodeReplyResult, RouterReply, ShutdownType,
 };
 use ockam_core::compat::{collections::BTreeMap, sync::Arc};
 use ockam_core::{Address, Result, TransportType};
@@ -115,6 +115,25 @@ impl Router {
             // If we _are_ still actually running then this is a real
             // failure and needs to be escalated
             e => e,
+        }
+    }
+
+    async fn check_addr_not_exist(
+        &self,
+        addr: &Address,
+        reply: &SmallSender<NodeReplyResult>,
+    ) -> Result<()> {
+        if self.map.internal.contains_key(addr) {
+            let node = NodeError::Address(addr.clone());
+
+            reply
+                .send(Err(node.clone().already_exists()))
+                .await
+                .map_err(|_| NodeError::NodeState(NodeReason::Unknown).internal())?;
+
+            Err(node.already_exists())
+        } else {
+            Ok(())
         }
     }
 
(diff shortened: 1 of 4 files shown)


pro465:
github didn't even let me edit the commit message before pushing 😞

pro465:
okay i accidentally did the opposite 😄

pro465:
oh no, did the same thing because of laziness 😭

pro465:
ok now it should be fixed

pro465:
okay it should pass now

SanjoDeundiak (dismissed):
Hey @pro465 , thank you for the contribution. I think the problem here is that router should reply with an error using `reply` mpsc sender, so that caller could handle the error correctly. Could you please also add a simple test for that scenario? Tnx!

pro465:
@SanjoDeundiak thank you for reviewing this! I don't know what would constitute a good test for this, but i will try nevertheless

pro465:
hey @SanjoDeundiak i have a question: do we need to check the secondary addresses too? or just the primary address?

pro465:
@SanjoDeundiak i dunno a good test, but  i'm guessing this:
```rust
#[test]
#[ockam::node]
async fn no_dup_addresses(ctx: &mut Context) {
    // create router
    // send StartWorker message
    // recieve message
    // assert recieved message is an error
}
```

mrinalwadhwa:
@pro465 before you add the test.
Do you mind if I modify [your branch](https://github.com/pro465/ockam/tree/check-addr) to remove the [merge commit](https://github.com/build-trust/ockam/runs/7702259611?check_suite_focus=true), so its easy to land when ready?

pro465:
@mrinalwadhwa no, I would not mind that. In fact I want that, so thank you for the offer!

mrinalwadhwa:
@pro465 all set. just add new commits from here, don't do any merges and we can squash when landing the pr.

pro465:
ok thank you!

pro465:
@SanjoDeundiak should i create a raw Router for the test,
or is there a high-level fn that i can call, that automatically does what i want, and i can simply just assert the result is an error when calling it twice?

mrinalwadhwa:
@pro465 checkout the tests here https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_node/src/tests.rs

pro465:
@mrinalwadhwa how do i specify that the furst `start_{worker, processor}` call should pass and only the second call should fail?

mrinalwadhwa:
Maybe:

1. start first worker
2. send it a message
3. receive an expected reply based on what the worker does - maybe an echo
4. start a second worker with the same name
5. receive a message
6. assert received message is an error

pro465:
> @mrinalwadhwa: 

> 



> Maybe:
> 
> [..]
> 4. start a second worker with the same name
> 5. receive a message
> 6. assert received message is an error

I'm confused, should not the second `start_worker` call return an error? or should it instead return Ok but send an Err thru the mpsc?

pro465:
(just to clarify, the it does send Err thru the mpsc, but the Context seem to catch it in the `start_worker` call and return if it is an Err, which is why i think the mpsc should not contain an error after returning from `start_worker`)

mrinalwadhwa:
You're right, the tests you've added look great. I'll merge shortly.

mrinalwadhwa (approved):

pro465:
@mrinalwadhwa thank you!

mrinalwadhwa:
@pro465 Thank you for another awesome contribution 👏 👏

PR #3168, opened 2022-08-08
refactor(rust): make it easier to write commands' api req/res handlers

adrianbenavides:
Add an abstraction layer on top of `connect_to` to make it easier to write commands' api req/res handlers. It removes a lot of boilerplate around calling the background node and especially, around handling errors. This can prove especially helpful for new contributors.

## Scope, limitations and future changes

The scope of this PR is to apply the new abstraction layer ONLY on `space` commands. For every command group we want to refactor to use this new layer, we have to implement a new trait (`Output`) that defines how the objects that we receive from the nodes are printed in the console. Note that with that trait implemented for `Space` and `Vec<Space>`, we have all we need to properly handle the output of `show`, `get`, and `list` commands.

The ideal situation would be to have an async runtime at the top level (at lib.rs or even at bin/ockam.rs) and pass down the `Context`. If I were to do this on this PR, I'd have to refactor every command which is way out of the scope, so that's why I have those `node_rpc` + `rpc` functions inside the commands. 

We should gradually refactor all the commands in small PR's, and once we have that ready, we can create the async runtime at the top level and then remove the `node_rpc` + `rpc` functions from all commands and just call the `run_impl` method.

## Commands with custom/partial outputs

We have some commands where we don't want to print the whole response or where we want to show a simple message at the end of the command.

To print a message after we receive a response (without body) from the node we can do the following:

```rust
let mut rpc = Rpc::new(...)?;
rpc.request(...).await?;
rpc.check_response()?;
println!("Success");
Ok(())
```

To manually decide what fields of the response body are printed we can do the following:

```rust
let mut rpc = Rpc::new(...)?;
rpc.request(...).await?;
let b = rpc.parse_response::<Space>()?;
println!("The space id is {}", b.id);
Ok(())
```

## Output for `space` commands

```bash
> space create ...
id: id-9joim12kmd
name: name-ajsndono2
users: u1, u2

> space show $ID
id: id-9joim12kmd
name: name-ajsndono2
users: u1, u2

> space list
+-----+-------+--------------+
| ID  | Name  | Users        |
+-----+-------+--------------+
| id1 | name1 | user1, user2 |
+-----+-------+--------------+
| id2 | name2 | user3, user2 |
+-----+-------+--------------+

> space delete $ID

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/authenticated.rs
+++ b/implementations/rust/ockam/ockam_command/src/authenticated.rs
@@ -43,7 +43,9 @@ pub enum AuthenticatedSubcommand {
 
 impl AuthenticatedCommand {
     pub fn run(c: AuthenticatedCommand) {
-        embedded_node(run_impl, c.subcommand)
+        if let Err(e) = embedded_node(run_impl, c.subcommand) {
+            eprintln!("Ockam node failed: {:?}", e,);
+        }
     }
 }
 
--- a/implementations/rust/ockam/ockam_command/src/util/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/util/mod.rs
@@ -1,21 +1,124 @@
+use std::{env, net::TcpListener, path::Path};
+
+use anyhow::Context;
+use minicbor::{Decode, Decoder, Encode};
+use tracing::error;
+use tracing_subscriber::prelude::*;
+use tracing_subscriber::{filter::LevelFilter, fmt, EnvFilter};
+
+pub use addon::AddonCommand;
+pub use config::*;
+use ockam::{route, Address, NodeBuilder, Route, TcpTransport, TCP};
+use ockam_api::nodes::NODEMANAGER_ADDR;
+use ockam_core::api::{RequestBuilder, Response, Status};
+
+use crate::util::output::Output;
+use crate::{CommandGlobalOpts, OutputFormat};
+
 pub mod api;
 pub mod exitcode;
 pub mod startup;
 
 mod addon;
-pub use addon::AddonCommand;
-
 mod config;
-pub use config::*;
-
-use anyhow::Context;
-use ockam::{route, NodeBuilder, Route, TcpTransport, TCP};
-use std::{env, net::TcpListener, path::Path};
-use tracing::error;
-use tracing_subscriber::prelude::*;
-use tracing_subscriber::{filter::LevelFilter, fmt, EnvFilter};
+mod output;
 
 pub const DEFAULT_CLOUD_ADDRESS: &str = "/dnsaddr/cloud.ockam.io/tcp/62526";
+
+pub struct Rpc<'a> {
+    ctx: &'a ockam::Context,
+    buf: Vec<u8>,
+    opts: &'a CommandGlobalOpts,
+    cfg: NodeConfig,
+}
+
+impl<'a> Rpc<'a> {
+    pub fn new(
+        ctx: &'a ockam::Context,
+        opts: &'a CommandGlobalOpts,
+        api_node: &'a str,
+    ) -> anyhow::Result<Self> {
+        Ok(Rpc {
+            ctx,
+            buf: Vec::new(),
+            opts,
+            cfg: opts.config.get_node(api_node)?,
+        })
+    }
+
+    pub async fn request<T>(&mut self, req: RequestBuilder<'_, T>) -> anyhow::Result<()>
+    where
+        T: Encode<()>,
+    {
+        let buf = req.to_vec()?;
+        let mut rte = connect(self.ctx, &self.cfg).await?;
+        self.buf = self
+            .ctx
+            .send_and_receive(rte.modify().append(NODEMANAGER_ADDR), buf)
+            .await?;
+        Ok(())
+    }
+
+    /// Parse the response body and return it.
+    pub fn parse_response<T>(&'a self) -> crate::Result<T>
+    where
+        T: Decode<'a, ()>,
+    {
+        let mut dec = self.parse_response_impl()?;
+        Ok(dec.decode().context("Failed to decode response body")?)
+    }
+
+    /// Parse response header only to check the status code.
+    pub fn check_response(&self) -> crate::Result<()> {
+        self.parse_response_impl()?;
+        Ok(())
+    }
+
+    /// Parse the response body and return it.
+    fn parse_response_impl(&'a self) -> crate::Result<Decoder> {
+        let mut dec = Decoder::new(&self.buf);
+        let hdr = dec
+            .decode::<Response>()
+            .context("Failed to decode response header")?;
+        match hdr.status() {
+            Some(Status::Ok) if hdr.has_body() => {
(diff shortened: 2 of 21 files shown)

### CODE REVIEW

**Comment by mrinalwadhwa** [comment]
@adrianbenavides I really like where you're going with this PR 👍

File: implementations/rust/ockam/ockam_command/src/util/mod.rs [hunk 1]
```diff
@@ -1,22 +1,140 @@
...
+            .context("Failed to decode response header")?;
+        match hdr.status() {
+            Some(Status::Ok) if hdr.has_body() => {
+                return Ok(dec.decode().context("Failed to decode response body")?);
+            }
+            Some(Status::Ok) => {
+                eprintln!("No body found in response");
+            }
+            Some(status) if hdr.has_body() => {
+                let err = dec.decode::<String>().unwrap_or_else(|_| "".to_string());
```
twittner:
```suggestion
                let err = dec.decode::<String>().unwrap_or_default();
```

implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -1,22 +1,140 @@
...
+                serde_json::to_string_pretty(&b).context("Failed to serialize response body")?
+            }
+        };
+        println!("{}", o);
+        Ok(())
+    }
+
+    /// Parse response header only to check the status code.
+    #[allow(dead_code)]
+    pub fn check_response(&self) -> crate::Result<()> {
```
twittner: `parse_response` and `check_response` could maybe share most of their implementations.

implementations/rust/ockam/ockam_core/src/cbor_utils/cow_str.rs
```diff
@@ -77,3 +78,13 @@ impl<'a, S: ?Sized + AsRef<str>> PartialEq<S> for CowStr<'a> {
         self.0 == other.as_ref()
     }
 }
+
+impl CowStr<'_> {
+    /// Join a slice of CowStr into a String.
+    pub fn slice_to_string(data: &[CowStr<'_>]) -> String {
```
twittner: I would suggest to move this to `ockam_command` as a utility function. It can also be generalised somewhat and it can avoid the vector allocation by using [itertools][2]:

```rust
pub fn comma_separated<T: AsRef<str>>(data: &[T]) -> String {
    use itertools::Itertools;

    #[allow(unstable_name_collisions)]
    data.iter()
        .map(AsRef::as_ref)
        .intersperse(", ")
        .collect()
}
```

If [`core::iter::Iterator::intersperse`][1] is stabilised, the `use` statement can be removed.

[1]: https://doc.rust-lang.org/core/iter/trait.Iterator.html#method.intersperse
[2]: https://crates.io/crates/itertools

implementations/rust/ockam/ockam_command/src/util/output.rs
```diff
@@ -0,0 +1,48 @@
+use cli_table::{Cell, Style, Table};
+use core::fmt::Write;
+
+use ockam_api::cloud::space::Space;
+use ockam_core::CowStr;
+
+pub trait Output {
```
twittner, line 31: This could benefit from some doc comment.

implementations/rust/ockam/ockam_command/src/util/output.rs
```diff
@@ -0,0 +1,48 @@
...
+    fn output(&self) -> anyhow::Result<String> {
+        let table = self
+            .iter()
+            .fold(
+                vec![],
+                |mut acc,
+                 Space {
+                     id, name, users, ..
+                 }| {
+                    let row = vec![
```
twittner:
```suggestion
                    let row = [
```

implementations/rust/ockam/ockam_command/src/util/output.rs
```diff
@@ -0,0 +1,48 @@
...
+                        id.cell(),
+                        name.cell(),
+                        CowStr::slice_to_string(users).cell(),
+                    ];
+                    acc.push(row);
+                    acc
+                },
+            )
+            .table()
+            .title(vec![
```
twittner:
```suggestion
            .title([
```

implementations/rust/ockam/ockam_command/src/util/output.rs
```diff
@@ -0,0 +1,48 @@
...
+        let mut w = String::new();
+        write!(w, "id: {}", self.id)?;
+        write!(w, "\nname: {}", self.name)?;
+        write!(w, "\nusers: {}", CowStr::slice_to_string(&self.users))?;
+        Ok(w)
+    }
+}
+
+impl Output for Vec<Space<'_>> {
+    fn output(&self) -> anyhow::Result<String> {
```
twittner, line 46: Stylistically I find a for loop like `for Space { id, name, users, .. } in self { ... }` easier to read compared to a fold, but YMMV.

twittner (dismissed):
:+1: Looks good to me! I just left some minor suggestions.

implementations/rust/ockam/ockam_core/src/cbor_utils/cow_str.rs
adrianbenavides: I have to implement AsRef for CowStr. Would this be ok?

```rust
impl<'a> AsRef<str> for CowStr<'a> {
    #[inline]
    fn as_ref(&self) -> &str {
        self.0.as_ref()
    }
}
```

I added the `inline` attribute because it looks like std implementations of this trait use it.

twittner: I would not bother with inline unless profiling shows it would benefit the code.

adrianbenavides:
> +1 Looks good to me! I just left some minor suggestions.

Thanks for all the comments! I've fixed all you pointed out in my last commit.

implementations/rust/ockam/ockam_command/src/authenticated.rs
```diff
@@ -43,7 +43,9 @@ pub enum AuthenticatedSubcommand {
 
 impl AuthenticatedCommand {
     pub fn run(c: AuthenticatedCommand) {
-        embedded_node(run_impl, c.subcommand)
+        if let Err(e) = embedded_node(run_impl, c.subcommand) {
+            eprintln!("Ockam node failed: {:?}", e,);
+        }
```
lameferret, line 48: Hey, shouldn't this return `exitcode`? Sorry if I'm missing something.

adrianbenavides: Hey! This is just a workaround. As soon as we refactor that command (and all the others) to use the new methods from this PR, most exit codes will be handled automatically (including the one you mentioned).

mrinalwadhwa (approved):

mrinalwadhwa (dismissed):

mrinalwadhwa:
@Mergifyio refresh

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

Issue #3184, opened 2022-08-09
Fix `ockam node start` on a running node makes it DOWN


mrinalwadhwa:
```
» ockam node create n1

Node Created!

Node:
  Name: n1
  Status: UP
  Services:
    Service:
      Type: TCP Listener
      Address: /ip4/127.0.0.1/tcp/54342
    Service:
      Type: Secure Channel Listener
      Address: /service/api
      Route: /ip4/127.0.0.1/tcp/54342/service/api
      Identity: P7dc502c795201c2d816749ea45ff51a416dce16ef0fe1b65ab4a642899e3b4d8
      Authorized Identities:
        - P7dc502c795201c2d816749ea45ff51a416dce16ef0fe1b65ab4a642899e3b4d8
    Service:
      Type: Uppercase
      Address: /service/uppercase
    Service:
      Type: Echo
      Address: /service/echo
  Secure Channel Listener Address: /service/api

» ockam node start n1
Node 'n1' already appears to be running as PID 8124

» ockam node show n1

Node:
  Name: n1
  Status: DOWN
  Services:
    Service:
      Type: TCP Listener
      Address: /ip4/127.0.0.1/tcp/54342
    Service:
      Type: Secure Channel Listener
      Address: /service/api
      Route: /ip4/127.0.0.1/tcp/54342/service/api
      Identity: N/A
      Authorized Identities:
        - N/A
    Service:
      Type: Uppercase
      Address: /service/uppercase
    Service:
      Type: Echo
      Address: /service/echo
  Secure Channel Listener Address: /service/api
```

The code for this is here https://github.com/build-trust/ockam/tree/develop/implementations/rust/ockam/ockam_command/src/node

---

_We love helping new contributors!_
_If you have questions or need help as you work on your first Ockam contribution, please leave a comment on [this discussion](https://github.com/ockam-network/ockam/discussions/1081)._
_If you're looking for other issues to contribute to, checkout issues labeled -  https://github.com/build-trust/ockam/labels/good%20first%20issue or https://github.com/build-trust/ockam/labels/help%20wanted_

mrinalwadhwa:
Fixed in https://github.com/build-trust/ockam/pull/3187, thank you @leonzchang 🎉

PR #3187, opened 2022-08-10
fix(rust): fix ockam node start and error info

leonzchang:
fix issue (#3184) and some error message output.

## Current Behavior
 If n1 already exists it will print out debug format error message.
```
» ockam create node n1

failed to update node configuration for 'n1': Exists("n1")
```

If starting a non-existing node, error message will be printed out as below.
```
» ockam node start n2

failed to load startup configuration for node 'n2' because: node with name 0 does not exist
```

## Proposed Changes
```
» ockam create node n1

failed to update node configuration for 'n1': node with name n1 already exists

» ockam node start n2

failed to load startup configuration for node 'n2' because: node with name n2 does not exist
```

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -105,7 +105,7 @@ impl CreateCommand {
                     verbose,
                 ) {
                     eprintln!(
-                        "failed to update node configuration for '{}': {:?}",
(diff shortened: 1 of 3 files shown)


mrinalwadhwa:
@leonzchang Thank you for another awesome contribution 🥳

mrinalwadhwa (approved):

PR #3197, opened 2022-08-11
MultiAddr lookup mechanism

spacekookie:
This PR (so far) adds two components:

- A mechanism to replace `/node/foo` entries in a MultiAddr with their backing addresses
- Storage for these node aliases and a new command to add them (`ockam alias set <name> <address>`, or any local node)

What's still missing:

- ~~Changing every command that currently accepts a `String` address to accept `MultiAddr`, use `ockam_api::utils::clean_multiaddr`, and test for the desired behaviour.~~
  - `ockam message send` and `ockam secure-channel create` work with the new schema
- ~~We can use `--to <alias>` in commands as well, but this needs to be implemented~~


spacekookie:
[0c138db](https://github.com/build-trust/ockam/pull/3197/commits/0c138db34c30c31e2d6770650e5610df594dfe05) adds a more generic lookup system where we can store anything else we might want to keep track of. The keys this uses are namespaced similar to multiaddr, so when storing a lookup for a node the key should be `/node/n1` where as storing a lookup for a project should be `/project/p1`, etc

mrinalwadhwa (approved):

Issue #3159, opened 2022-08-07
Improve secure channels to cleanup any spawned workers


adrianbenavides:
When [creating a secure channel](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_identity/src/channel.rs#L35), multiple workers are spawned but only the address of the top-level worker is returned, making it impossible to remove all used resources.

We use "temporary secure channels" in [`ockam_api/src/cloud`](https://github.com/build-trust/ockam/tree/develop/implementations/rust/ockam/ockam_api/src/cloud), and right now, every time we call one of these functions, we increase the worker count on the node. We need to find a different approach to create secure channels to release all the acquired resources.

Retamogordo:
Hi,
a noob question: is it about delete_secure_channel(...) ?
Context that sends a graceful shutdown request to the router ?
And spawned workers are supposed to follow the same shutdown procedure ?

mrinalwadhwa:
@Retamogordo yes, create_secure_channel spawns a few workers .. there is delete_secure_channel [here](https://github.com/build-trust/ockam/blob/ad0f26bbdd9aa9c5899f63145b297bcf5291b3d4/implementations/rust/ockam/ockam_api/src/nodes/service.rs#L231-L237) which naively just stops the worker at the address being used to send messages to the channel.

We need a control protocol that will tell the secure channel to do a graceful shutdown of all workers that it spawned then shut itself down

Retamogordo:
Hi @mrinalwadhwa,
it looks like that the "parent" context needs a mechanism of keeping track its "children" entities ? Does it refer to secure_channel only or to general case of any kind of context ? 
It looks like the way the router functions now assumes that only one graceful shutdown with ack is possible, while all others that share the same  cloned sender will get an error because of dropped receiver ?

mrinalwadhwa:
> it looks like that the "parent" context needs a mechanism of keeping track its "children" entities

@Retamogordo exactly, the long term plan is to model something similar to Erlang's Supervision Trees
https://www.erlang.org/doc/design_principles/des_princ.html#supervision-trees
https://www.erlang.org/doc/man/supervisor.html
https://elixir-lang.org/getting-started/mix-otp/supervisor-and-application.html
https://hexdocs.pm/elixir/Supervisor.html

[...]

Issue #3186, opened 2022-08-10
Change `ockam identity` commands to support multiple identities.


mrinalwadhwa:
Currently:

```
ockam node create n1
```

creates a default identity for the newly created node

we can see it with

```
» ockam identity show --node n1
P7dc502c795201c2d816749ea45ff51a416dce16ef0fe1b65ab4a642899e3b4d8
```

Even thought we have an `identity create` command `SHOW_HIDDEN ockam identity create --help`

We can't create a new identity because a node configuration can currently only have one identity.

```
» ockam identity create --node n1
encountered an error in command handler code
```

We have to modify NodeManager and its configuration here to support multiple identities
https://github.com/build-trust/ockam/blob/83db24993f74af23adcee23228fdc10d3bb1fe68/implementations/rust/ockam/ockam_api/src/nodes/config.rs#L6-L15

The code for the identity commands is here
https://github.com/build-trust/ockam/tree/6093f3a77869738fb4417f5093e5fc7d00acdf8c/implementations/rust/ockam/ockam_command/src/identity

The code for identity apis that identity commands call are here
https://github.com/build-trust/ockam/blob/0c5a39891eddec31056b7838d9efac50d287185c/implementations/rust/ockam/ockam_api/src/nodes/service/identity.rs#L14 

Once we have the ability to have multiple identities, we'll need other commands like `identity show | list`

lameferret:
- What is the reasoning behind multiple identities? 
    - Is it for multiple services (other nodes) to talk to the same node with their own unique ID? 
- Can `Vec<Vec<u8>>` be used for storing multiple identities? Is there a better way to store them?
- How can identity be overridden? I couldn't find the command to do so.

mrinalwadhwa:
@anuvratsingh thank you for looking into this.

> What is the reasoning behind multiple identities?

The underlying `ockam` library supports the idea of a node/application using different identities in different privacy and authorization contexts. But we haven't surfaced that in ockam command yet. So I may communicate with remote service 1 using identity 1 and with remote service 2 using identity 2.

> Can Vec<Vec<u8>> be used for storing multiple identities? Is there a better way to store them?

@SanjoDeundiak may be the best person to add some context here. The private information (like keys) of identities live in vaults. So we'll have to tackle that.

> How can identity be overridden? I couldn't find the command to do so.

We'll need new options in various commands to indicate the `--identity` and the `--vault`
[...]

PR #3200, opened 2022-08-11
feat(rust): check controller's identity id when creating secure channel

adrianbenavides:
https://github.com/build-trust/codebase/issues/330
https://github.com/build-trust/codebase/issues/340

The controller's identity can be loaded in three different ways:

1. ockam binary: the contents from `implementations/rust/ockam/ockam_api/static/controller.id` will be embedded into the binary and will be used by default
2. dev (cargo run ...): the contents from `implementations/rust/ockam/ockam_api/static/controller.id` will be used also in this case, but you can change its contents to try different id's
3. `OCKAM_CONTROLLER_IDENTITY_ID` env var: if it's present, its contents will be used and will have priority over the previous methods

## How to test

If running minikube on `ockam-cloud`, copy the `./minikube/mnt/storage/controller/identity.id` contents. We'll use it to set the `OCKAM_CONTROLLER_IDENTITY_ID` env var.

```bash
OCKAM_CONTROLLER_IDENTITY_ID={identity.id-contents} ockam node create default
ockam enroll --auth0 $CONTROLLER # this method will use a secure channel using the controller's identity id
```

Alternatively, you can copy the contents into the `implementations/rust/ockam/ockam_api/static/controller.id` file.

## Other changes

- Add a new registry entry into the node to cache secure channels and avoid the [resource leak](https://github.com/build-trust/ockam/issues/3159) on deleting them
- Remove cloud tests as they are no longer useful. They were testing against a fake server and were useful at the beginning to test how it all worked, but they are not testing anything really and we'll shortly have new integration tests for this


implementations/rust/ockam/ockam_api/src/nodes/registry.rs
```diff
@@ -1,10 +1,37 @@
...
+        &self.addr
+    }
+}
+
+#[derive(Default, Hash, Ord, PartialOrd, Eq, PartialEq)]
+pub(crate) struct IdentityRouteKey(Vec<u8>);
+
+impl IdentityRouteKey {
+    pub(crate) async fn new(identity: &Identity<Vault>, route: &Route) -> Result<Self> {
+        let mut key = identity.export().await?;
```
SanjoDeundiak, line 29: Could you please elaborate why identity should be present in the key? NodeManager can't have more than one

adrianbenavides: Not needed! This is a remnant of a wrong approach I did when I started to work on this PR. I'll leave just the `route` as a key. The idea is to reuse the secure channels that are used between node and orchestrator (i.e. controller + project nodes).

adrianbenavides: I was now reviewing the open issues we have and it seems we want to support multiple identities on a node: https://github.com/build-trust/ockam/issues/3186

So I guess I should revert and use `IdentityRouteKey` as a key.

spacekookie (dismissed):
LGTM

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3201, opened 2022-08-11
feat(rust): add exitcodes

### LINKED ISSUE #3199: Make cli command use exit code to signal success/failures
In most/all of the commands, there are cases where the command fail,  we print the error,  but we return with a success code anyway., like here:
https://github.com/build-trust/ockam/blob/b80b36bcb407a2b23cdf54618e9ac8f4162fffe3/implementations/rust/ockam/ockam_command/src/secure_channel/create.rs#L98-L103
(in that case,  the exit code will come from the last  `stop_node(ctx).await`  result)-

Instead, commands must be adapted so that they exit with a non-success code if they fail.   Even better if we could use more specific error codes for different scenarios.   Related docs: https://rust-cli.github.io/book/in-depth/exit-code.html

lameferret:
Fix #3199 

- Vast majority of errors were `IOERR`.
- Unused errors in `util/exitcode.rs` were ignored using `_` rather than removing.
- Didn't use `OK` exitcode as it implied by default.
- Didn't add comment describing exitcode usage, as they are self-explanatory and would be redundant.

Clarification required:
1. `message::send.rs`, etcetera have "failed to normalize MultiAddr route" 
    - Is this a `USAGE` error or `SOFTWARE` error?
2. Can error in `create.rs`(`forwarder/create.rs#L99` etcetera) be considered as `CANTCREAT` error?
3. `tcp/connection/delete.rs#L64` "You may have to provide --force to delete the API transport"
    - Can this be `USAGE` error?
4. `util/mod.rs#L68` "encountered an error in command handler code"
    - Is this `USAGE`, `IOERR` or `SOFTWARE` error?
    
Unrelated: Shouldn't `/transport` dir be removed?

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/util/exitcode.rs
+++ b/implementations/rust/ockam/ockam_command/src/util/exitcode.rs
@@ -1,0 +1,78 @@
+#![allow(dead_code)]
+/// Alias for the numeric type that holds system exit codes.
+pub type ExitCode = i32;
+
+/// Successful exit
+pub const OK: ExitCode = 0;
+
+/// The command was used incorrectly, e.g., with the
+/// wrong number of arguments, a bad flag, a bad syntax
+/// in a parameter, etc.
+pub const USAGE: ExitCode = 64;
+
+/// The input data was incorrect in some way.  This
+/// should only be used for user's data and not system
+/// files.
+pub const DATAERR: ExitCode = 65;
+
+/// An input file (not a system file) did not exist or
+/// was not readable.  This could also include errors
+/// like "No message" to a mailer (if it cared to
+/// catch it).
+pub const NOINPUT: ExitCode = 66;
+
+/// The user specified did not exist.  This might be
+/// used for mail addresses or remote logins.
+pub const NOUSER: ExitCode = 67;
+
+/// The host specified did not exist.  This is used in
+/// mail addresses or network requests.
+pub const NOHOST: ExitCode = 68;
+
+/// A service is unavailable.  This can occur if a
+/// support program or file does not exist. This can also
+/// be used as a catchall message when something you
+/// wanted to do doesn't work, but you don't know why.
+pub const UNAVAILABLE: ExitCode = 69;
+
+/// An internal software error has been detected.  This
+/// should be limited to non-operating system related
+/// errors as possible.
+pub const SOFTWARE: ExitCode = 70;
+
+/// An operating system error has been detected.  This
+/// is intended to be used for such things as "cannot
+/// fork", "cannot create pipe", or the like.  It
+/// includes things like getuid returning a user that
+/// does not exist in the passwd file.
+pub const OSERR: ExitCode = 71;
+
+/// Some system file (e.g., /etc/passwd, /var/run/utmp,
+/// etc.) does not exist, cannot be opened, or has some
+/// sort of error (e.g., syntax error).
+pub const OSFILE: ExitCode = 72;
+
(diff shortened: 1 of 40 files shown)


adrianbenavides:
Hey @anuvratsingh thanks for the PR, that's really helpful! 

Answering your questions:

> 1. `message::send.rs`, etcetera have "failed to normalize MultiAddr route"
>    * Is this a `USAGE` error or `SOFTWARE` error?

This is generally because the user has passed an invalid route, so I'd say it's a `USAGE` error.

> 2. Can error in `create.rs`(`forwarder/create.rs#L99` etcetera) be considered as `CANTCREAT` error?

Don't worry about these ones right now, I'm preparing a refactor to handle all these errors in one place. We'll return an error based on the returned error code from the node.

> 3. `tcp/connection/delete.rs#L64` "You may have to provide --force to delete the API transport"
>    * Can this be `USAGE` error?

I'd use either `UNAVAILABLE` or `SOFTWARE` since the cause here is that the background node, for some reason, couldn't delete the TCP connection properly.

> 4. `util/mod.rs#L68` "encountered an error in command handler code"
>    * Is this `USAGE`, `IOERR` or `SOFTWARE` error?

Sames as question `2`.

>  5. Unrelated: Shouldn't `/transport` dir be removed?

We can address that in a separate PR. We are having a lot of movement in this crate so it might get removed later on.

-----

And a final comment regarding

> * Unused errors in util/exitcode.rs were ignored using _ rather than removing.

Since all the errorcode members are public, it won't harm to have them all available, just as in the original library. In other words, we don't have to ignore or remove the unused codes

implementations/rust/ockam/ockam_command/src/util/exitcode.rs
```diff
@@ -0,0 +1,105 @@
...
+pub const _NOPERM: ExitCode = 77;
+
+/// Something was found in an unconfigured or misconfigured state.
+pub const _CONFIG: ExitCode = 78;
+
+/// Check if exit code given by `code` is successful
+///
+/// # Example:
+/// ```
+/// extern crate exitcode;
```
adrianbenavides: I think this should be `use crate::utils::exitcode;`

implementations/rust/ockam/ockam_command/src/util/exitcode.rs
```diff
@@ -0,0 +1,105 @@
...
+/// ```
+pub fn _is_success(code: ExitCode) -> bool {
+    code == _OK
+}
+
+/// Check if exit code given by `code` is an error
+///
+/// # Example:
+/// ```
+/// extern crate exitcode;
```
adrianbenavides: Here too (`use crate::utils::exitcode;`)

implementations/rust/ockam/ockam_command/src/util/exitcode.rs
lameferret: - with `use crate::utils::exitcode;`
    - could not find `utils` in the crate root
- with `use crate::util::exitcode;`
    - unresolved import help: a similar path exists: `tracing_subscriber::util`

lameferret:
Thanks @adrianbenavides, for the review. I will push commit with the suggested changes asap.

lameferret:
I have removed `is_error` and `is_success` with their doc tests.

mrinalwadhwa:
@anuvratsingh the changes look great, thank you for another great contribution 🥳

mrinalwadhwa (approved):

PR #3212, opened 2022-08-12
feat(rust): make `IdentityIdentifier` encodable

twittner:
Derive `minicbor::{Encode, Decode}` for `IdentityIdentifier` to enable direct use of the type in API requests and responses.

Also changes `Identiy::identifier` to return `&IdentityIdentifier` instead of `Result<IdentityIdentifier>` as the method never fails and to better support callers that do not need to get a clone of the value.

### DIFF
--- a/implementations/rust/ockam/ockam_api/tests/auth.rs
+++ b/implementations/rust/ockam/ockam_api/tests/auth.rs
@@ -37,7 +37,7 @@ async fn credential(ctx: &mut Context) -> Result<()> {
         let e_store = InMemoryStorage::new();
         let m_store = InMemoryStorage::new();
         let mut auth = direct::Server::new(m_store, e_store, mk_signer(ctx).await?);
-        auth.set_admin(&admin.identifier()?);
+        auth.set_admin(admin.identifier());
         ctx.start_worker("auth", auth).await?;
     }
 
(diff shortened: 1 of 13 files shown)


implementations/rust/ockam/ockam_api/tests/auth.rs
```diff
@@ -69,7 +69,7 @@ async fn credential(ctx: &mut Context) -> Result<()> {
 
     // Add the member via the enroller's connection:
     let mut c = direct::Client::new(route![e2a, "auth"], ctx).await?;
-    c.add_member(IdentityId::new(m.identifier()?.key_id()))
+    c.add_member(IdentityId::new(m.identifier().key_id()))
```
SanjoDeundiak, line 72: Does it still need to access `key_id`?

implementations/rust/ockam/ockam_api/tests/auth.rs
```diff
@@ -49,7 +49,7 @@ async fn credential(ctx: &mut Context) -> Result<()> {
             .create_secure_channel("api", TrustEveryonePolicy, &InMemoryStorage::new())
             .await?;
         let mut c = direct::Client::new(route![a2a, "auth"], ctx).await?;
-        c.add_enroller(IdentityId::new(e.identifier()?.key_id()))
+        c.add_enroller(IdentityId::new(e.identifier().key_id()))
```
SanjoDeundiak, line 52: Same question here

SanjoDeundiak (changes requested):
Almost perfect

SanjoDeundiak (dismissed):

SanjoDeundiak:
@mergifyio refresh

SanjoDeundiak (approved):

implementations/rust/ockam/ockam_api/tests/auth.rs
twittner: This part will change in future PRs soon. I limited the scope of this one.

PR #3228, opened 2022-08-15
feat(rust): define credential structure in ockam crate

twittner:
Defines the `Credential` structure of https://github.com/build-trust/codebase/issues/375 and sign/verify methods.

### DIFF
--- a/implementations/rust/ockam/ockam/src/credential.rs
+++ b/implementations/rust/ockam/ockam/src/credential.rs
@@ -4,6 +4,7 @@
 
 mod exchange;
 
+use core::marker::PhantomData;
 use core::time::Duration;
 use minicbor::bytes::ByteSlice;
 use minicbor::{Decode, Encode};
@@ -21,6 +22,14 @@ use crate::TypeTag;
 
 pub const MAX_CREDENTIAL_VALIDITY: Duration = Duration::from_secs(6 * 3600);
 
+/// Type to represent data of verified credentials.
+#[derive(Debug, Encode)]
+pub enum Verified {}
+
+/// Type to represent data of unverified credentials.
+#[derive(Debug, Decode)]
+pub enum Unverified {}
+
 #[derive(Debug, Decode, Encode)]
 #[rustfmt::skip]
 #[cbor(map)]
@@ -36,7 +45,7 @@ pub struct Credential<'a> {
 #[derive(Debug, Decode, Encode)]
 #[rustfmt::skip]
 #[cbor(map)]
-pub struct CredentialData<'a> {
+pub struct CredentialData<'a, T> {
     /// A schema identifier to allow distinguishing sets of attributes.
     #[n(1)] schema: Option<SchemaId>,
     /// User-defined key-value pairs.
@@ -50,7 +59,9 @@ pub struct CredentialData<'a> {
     /// The time when this credential was created.
     #[n(6)] created: Timestamp,
     /// The time this credential expires.
-    #[n(7)] expires: Timestamp
+    #[n(7)] expires: Timestamp,
+    /// Term to represent the verification status type.
+    #[n(8)] status: Option<PhantomData<T>>
 }
 
 impl<'a> Credential<'a> {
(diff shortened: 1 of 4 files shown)


implementations/rust/ockam/ockam/src/credential.rs
```diff
@@ -1,44 +1,270 @@
...
+pub const MAX_CREDENTIAL_VALIDITY: Duration = Duration::from_secs(6 * 3600);
+
 #[derive(Debug, Decode, Encode)]
 #[rustfmt::skip]
 #[cbor(map)]
 pub struct Credential<'a> {
     #[cfg(feature = "tag")]
     #[n(0)] tag: TypeTag<3796735>,
+    /// CBOR-encoded [`CredentialData`].
     #[b(1)] attributes: CowBytes<'a>,
```
SanjoDeundiak: Probably should be renamed to credential_data

implementations/rust/ockam/ockam/src/credential.rs
```diff
@@ -1,44 +1,270 @@
...
 }
 
+#[derive(Debug, Decode, Encode)]
+#[rustfmt::skip]
+#[cbor(map)]
+pub struct CredentialData<'a> {
+    /// A schema identifier to allow distinguishing sets of attributes.
+    #[n(1)] schema: Option<SchemaId>,
+    /// User-defined key-value pairs.
+    #[b(2)] attributes: Option<Attributes<'a>>,
```
SanjoDeundiak: Not sure it makes sense to make it Optional. Usually it will have some values (since that the whole point of Credential), otherwise it can also be empty

implementations/rust/ockam/ockam/src/credential.rs
```diff
@@ -1,44 +1,270 @@
...
         &self.signature
     }
+
+    /// Perform a signature check with the given identity.
+    pub async fn is_signature_of<V>(&self, issuer: &Identity<V>) -> Result<bool>
+    where
+        V: IdentityVault,
+    {
+        let dat = CredentialData::try_from(self)?;
+        let sig = Signature::new(self.signature.clone().into_owned());
```
SanjoDeundiak, line 88: Signature can be Cow in the future

implementations/rust/ockam/ockam/src/credential.rs
```diff
@@ -1,44 +1,270 @@
...
         &self.attributes
     }
 
-    /// Signature getter
     pub fn signature(&self) -> &[u8] {
         &self.signature
     }
+
+    /// Perform a signature check with the given identity.
+    pub async fn is_signature_of<V>(&self, issuer: &Identity<V>) -> Result<bool>
```
SanjoDeundiak: I would say if it is doing verification, it should have verify in its name. My first assumption from this name was that it checks if issue in the credential is the same as issuer in the argument

SanjoDeundiak: I would decouple CBOR-encoded Credential struct from any implementation that involves issuing or verification. The could move to Identity, or a separate struct

SanjoDeundiak (dismissed):
Left some comments, but this is good to merge

implementations/rust/ockam/ockam/src/credential.rs
```diff
@@ -1,44 +1,270 @@
...
+            val <= MAX_CREDENTIAL_VALIDITY,
+            "validitiy exceeds allowed maximum"
+        );
+        self.validity = val;
+        self
+    }
+
+    /// Create a signed credential based on the given values.
+    #[cfg(feature = "std")]
+    pub async fn finish<V>(self, issuer: &Identity<V>, key_label: &str) -> Result<Credential<'a>>
```
SanjoDeundiak: Name can be `issue`?

implementations/rust/ockam/ockam/src/credential.rs
SanjoDeundiak: P.S. Probably, moving to Identity would mean moving all of these code to identity crate, so let's not do this. But something like `CredentialIssuer` and `CredentialVerifier` could be great. Also, should simplify further update of corresponding Server and Client structs to use it

implementations/rust/ockam/ockam/src/credential.rs
twittner: I used an `Option` as an easy easy way to not encode the key and the empty map, but this can be postponed (if needed at all).

SanjoDeundiak (approved):

PR #3230, opened 2022-08-15
feat(rust): use `IdentityChangeHistory` for verification

twittner:
No need for the full `Identity`.


implementations/rust/ockam/ockam/src/credential.rs
```diff
@@ -13,7 +13,9 @@ use ockam_core::compat::collections::BTreeMap;
 use ockam_core::errcode::{Kind, Origin};
 use ockam_core::vault::{Signature, SignatureVec};
```
SanjoDeundiak: I think Verifier can be imported here as well

SanjoDeundiak (changes requested):

implementations/rust/ockam/ockam/src/credential.rs
twittner: Too much automation.

SanjoDeundiak (approved):

PR #3236, opened 2022-08-15
Credentials Exchange

SanjoDeundiak:


implementations/rust/ockam/ockam/src/credential.rs
```diff
@@ -81,12 +83,19 @@ impl<'a> Credential<'a> {
     pub async fn verify_signature<'b: 'a, V>(
         &'b self,
         issuer: &IdentityChangeHistory,
-        verifier: V,
+        verifier: &V,
```
twittner, line 86: I was thinking that we could perhaps add a blanket impl, like in here: https://github.com/build-trust/ockam/blob/9596b93498239f0496d09d409608a87c21c7743a/implementations/rust/ockam/ockam_core/src/vault/verifier.rs

SanjoDeundiak: What are the benefits?
P.S. Let's merge this PR and keep the discussion

twittner (approved):

mrinalwadhwa (approved):

PR #3240, opened 2022-08-16
feat(rust): use auth0 + controller production settings

adrianbenavides:
I'll leave it as a draft while we generate the new elixir image that points to the auth0 production environment.


implementations/rust/ockam/ockam_command/src/enroll/auth0.rs
```diff
@@ -80,9 +80,9 @@ async fn enroll(
 pub struct Auth0Service;
 
 impl Auth0Service {
-    const DOMAIN: &'static str = "dev-w5hdnpc2.us.auth0.com";
-    const CLIENT_ID: &'static str = "sGyXBwQfU6fjfW1gopphdV9vCLec060b";
-    const API_AUDIENCE: &'static str = "https://dev-w5hdnpc2.us.auth0.com/api/v2/";
+    const DOMAIN: &'static str = "ockam.us.auth0.com";
```
mrinalwadhwa, line 83: @polvorin this should be `account.ockam.io` but the api audience should be `https://ockam.us.auth0.com/api/v2/`

mrinalwadhwa (approved):

PR #3253, opened 2022-08-16
verifier and authenticator

twittner:
Continuation of #2965. Renames signer to verifier and changes the direct authenticator to work on enroller files. Also adds ways to start the services from command line.


hairyhum:
> Also adds ways to start the services from command line.

We currently run the sidecar node as 
`ockam node create -v -f --api-address=127.0.0.1:4100 sidecar`
and the identity service is started with the node.

Are we supposed to do that differently now?

Also can we have an API doc for those services with response codes data formats for request/response?

implementations/rust/ockam/ockam_core/src/vault/verifier.rs
```diff
@@ -13,3 +13,10 @@ pub trait Verifier {
         data: &[u8],
     ) -> Result<bool>;
 }
+
```
SanjoDeundiak, line 16: Let's add it to all traits (at list in this module)

implementations/rust/ockam/ockam_api/src/verifier.rs
```diff
@@ -0,0 +1,130 @@
...
+            target: "ockam_api::verifier",
+            id     = %req.id(),
+            method = ?req.method(),
+            path   = %req.path(),
+            body   = %req.has_body(),
+            "request"
+        }
+
+        let res = match req.method() {
+            Some(Method::Get) => match req.path_segments::<2>().as_slice() {
```
SanjoDeundiak: I guess it should be POST actions/verify

mrinalwadhwa:
@hairyhum the updated command should be:

```
ockam node create sidecar --foreground --tcp-listener-address 127.0.0.1:4100 -vvv
```

implementations/rust/ockam/ockam_api/src/verifier.rs
```diff
@@ -0,0 +1,130 @@
...
+                    let vr: VerifyRequest = dec.decode()?;
+                    match self.verify(req.id(), &vr).await {
+                        Ok(Either::Left(err)) => err.to_vec()?,
+                        Ok(Either::Right(dat)) => {
+                            // copy expiration date into attributes
+                            let expiration = dat.expires_at();
+                            let mut attr = dat.into_attributes();
+                            let mut buff = Cursor::new([0; 9]);
+                            minicbor::encode(expiration, &mut buff)?;
+                            attr.put(EXPIRES_AT, &buff.get_ref()[..buff.position()]);
```
SanjoDeundiak: How about having VerifyResponse, that would have attributes and expire fields?

implementations/rust/ockam/ockam_api/src/verifier.rs
```diff
@@ -0,0 +1,130 @@
...
+        };
+
+        Ok(res)
+    }
+
+    async fn verify<'a>(
+        &self,
+        id: Id,
+        req: &'a VerifyRequest<'a>,
+    ) -> Result<Either<ResponseBuilder<Error<'_>>, CredentialData<'a, Verified>>> {
```
SanjoDeundiak, line 85: Well, that looks ugly, but we haven't had pretty solution to that problem I guess

implementations/rust/ockam/ockam_api/src/verifier.rs
```diff
@@ -0,0 +1,130 @@
...
+        if req.subject() != data.subject() {
+            let err = Error::new("/verify").with_message("invalid subject");
+            return Ok(Either::Left(Response::forbidden(id).body(err)));
+        }
+
+        let now = Timestamp::now().ok_or_else(|| {
+            ockam_core::Error::new(Origin::Core, Kind::Internal, "invalid system time")
+        })?;
+
+        if now < data.expires_at() {
```
SanjoDeundiak: It's `>=`, isn't it?

implementations/rust/ockam/ockam_api/src/verifier.rs
```diff
@@ -0,0 +1,130 @@
...
+            let err = Error::new("/verify").with_message("expired");
+            return Ok(Either::Left(Response::forbidden(id).body(err)));
+        }
+
+        if let Some(PROJECT_MEMBER_SCHEMA) = data.schema() {
```
SanjoDeundiak: I don't think we should check any attributes here, this verification can be generic to verify any credential. Caller should figure out what are the schema and needed keys&values for that credential to authorize certain action

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -1,41 +1,49 @@
...
-#[derive(Debug)]
-pub struct Server<S, T> {
-    m_store: S, // member store
-    e_store: T, // enroller store
-    signer: signer::Client,
-    admin: Option<IdentityId<'static>>,
+pub struct Server<S, V: IdentityVault> {
+    project: Vec<u8>,
+    store: S,
+    ident: Arc<Identity<V>>,
```
SanjoDeundiak: What is the purpose of Arc here?

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -87,197 +95,93 @@ where
...
-                        let ts = Timestamp::now().ok_or_else(invalid_sys_time)?;
-                        let mut timestamp = Cursor::new([0; 9]);
-                        minicbor::encode(ts, &mut timestamp)?;
-                        attrs.put("ts", &timestamp.get_ref()[..timestamp.position()]);
-                        let crd = self.signer.sign(&attrs).await?;
+                ["credential"] => match self.check_member(&req, from).await {
+                    Ok(None) => {
+                        let crd = Credential::builder(from.clone())
+                            .with_schema(PROJECT_MEMBER_SCHEMA)
+                            .with_attribute(PROJECT_ID, &self.project)
```
SanjoDeundiak, line 115: Should be {"project_id": "123ef"}, {"role": "member"}

implementations/rust/ockam/ockam_api/src/lib.rs
```diff
@@ -20,33 +20,21 @@ pub mod lmdb;
...
-    }
-}
-
-#[cfg(feature = "std")]
-impl From<Timestamp> for u64 {
-    fn from(t: Timestamp) -> Self {
-        t.0
+/// Newtype around a byte-slice that is assumed to be CBOR-encoded.
+#[derive(Debug, Copy, Clone)]
+pub struct Cbor<'a>(pub &'a [u8]);
```
SanjoDeundiak: Isn't that present in ockam_core::api?

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -70,7 +70,7 @@ pub struct NodeManager {
     skip_defaults: bool,
 
     vault: Option<Vault>,
-    identity: Option<Identity<Vault>>,
+    identity: Option<Arc<Identity<Vault>>>,
```
SanjoDeundiak: Why it should be Arc?

SanjoDeundiak (changes requested):
My main concern is Arc that wraps Identity. Other things can be tackled later

implementations/rust/ockam/ockam_api/src/verifier.rs
twittner: :facepalm:

implementations/rust/ockam/ockam_api/src/verifier.rs
twittner: Why the "actions" prefix?

twittner:
> > Also adds ways to start the services from command line.
> 
> We currently run the sidecar node as `ockam node create -v -f --api-address=127.0.0.1:4100 sidecar` and the identity service is started with the node.
> 
> Are we supposed to do that differently now?

TBD. Currently the verifier and authenticator services are not started automatically so they require additional commands, e.g.

```
> ockam service start verifier --node foo
Verifier service started at address: verifier
> ockam service start authenticator --node foo --enrollers /tmp/enrollers.json --project project42
Authenticator service started at address: authenticator
```

> Also can we have an API doc for those services with response codes data formats for request/response?

schema.cddl has been updated with verify request/response structures.

implementations/rust/ockam/ockam_api/src/verifier.rs
SanjoDeundiak: Because POST verify means something like "add an object of type verify". Therefore, me and @mrinalwadhwa decided to add actions prefix

SanjoDeundiak:
@mergifyio rebase

SanjoDeundiak (approved):

PR #3260, opened 2022-08-17
Fix schema validation

SanjoDeundiak:


SanjoDeundiak:
@mergifyio rebase

implementations/rust/ockam/ockam_core/src/api.rs
```diff
@@ -516,7 +518,7 @@ impl<T: Encode<()>> ResponseBuilder<T> {
 }
 
 #[allow(unused_variables)]
-pub fn assert_request_match<'a>(schema: impl Into<Option<&'a str>>, cbor: &[u8]) {
+pub fn assert_request_match<'a>(struct_name: impl Into<Option<&'a str>>, cbor: &[u8]) {
```
twittner, line 521: While we currently send structs we could also send enum bodies in requests. The previous naming reads better to me, as in "assert the request matches a schema".

twittner (approved):
The `cddl-cat` dependency in `ockam_api` is no longer used and can be removed.

implementations/rust/ockam/ockam_core/src/api.rs
SanjoDeundiak: Previous naming was misleading, as there are 2 schemas in this function: what you're going to send, and schema from schema.cddl. Not saying current naming is perfect

PR #3263, opened 2022-08-17
Two-way cred exchange

SanjoDeundiak:


implementations/rust/ockam/ockam/src/credential.rs
```diff
@@ -230,6 +234,15 @@ impl<'a> Attributes<'a> {
     pub fn iter(&self) -> impl Iterator<Item = (&str, &[u8])> {
         self.attrs.iter().map(|(k, v)| (*k, &***v))
     }
+
+    pub fn to_owned(&self) -> BTreeMap<String, Vec<u8>> {
+        let mut map = BTreeMap::<String, Vec<u8>>::default();
```
twittner:
```suggestion
        let mut map = BTreeMap::default();
```

implementations/rust/ockam/ockam/src/credential/exchange.rs
```diff
@@ -26,14 +50,117 @@ impl CredentialExchange {
...
+                )
+            }
+        };
+
+        // TODO: Implement expiration mechanism in Storage
+        let entry = AttributesEntry::new(credential_data.attributes, credential_data.expires);
+
+        let mut entry_buf = vec![];
+        let mut encoder = Encoder::new(&mut entry_buf);
+        encoder.encode(entry)?;
```
twittner:
```suggestion
        let entry_buf = minicbor::to_vec(&entry)?;
```

implementations/rust/ockam/ockam/src/credential/exchange.rs
```diff
@@ -64,68 +191,80 @@ impl CredentialExchange {
...
+                    Origin::Application,
+                    Kind::Invalid,
+                    "credential presentation failed",
+                ))
+            }
+        }
+
+        let credential: Credential = dec.decode()?;
+
+        Self::receive_presented_credential(
```
twittner: Should this not act on error?

implementations/rust/ockam/ockam/src/credential/exchange.rs
```diff
@@ -150,12 +289,43 @@ impl<S: AuthenticatedStorage, V: IdentityVault> CredentialExchangeWorker<S, V> {
...
+            (Post, ["actions", "present_mutual"]) => {
+                let credential: Credential = dec.decode()?;
+
+                CredentialExchange::receive_presented_credential(
+                    req.id(),
+                    req.path(),
+                    sender,
+                    credential,
+                    &self.authorities,
+                    self.present_back.as_ref(),
```
twittner: If this is `None` it will just send an OK response without the credential body. Should we not error in this case?

twittner (commented):
Some doc comments would be helpful I think.

implementations/rust/ockam/ockam/src/credential.rs
SanjoDeundiak: Sure, I wanted Default::default, but rust refused to defer the type, so I decided to be clear enough for it 😄

implementations/rust/ockam/ockam/src/credential/exchange.rs
SanjoDeundiak: Oh, right, I had feeling that there was a way to do it simpler, but forgot what it is

SanjoDeundiak: Maybe Encoder::to_vec() could also exist, since this is what I was trying to find

implementations/rust/ockam/ockam/src/credential/exchange.rs
SanjoDeundiak: What do you mean?

implementations/rust/ockam/ockam/src/credential/exchange.rs
SanjoDeundiak: Interesting question. So, if initiator presents credential for us and expects credential in return, but we don't have it, should we accept his credential or not? I think we should accept it. Initiator will error either way, though error may be not super-friendly atm, there is space for improvement

implementations/rust/ockam/ockam/src/credential/exchange.rs
twittner: `receive_presented_credential` may fail to validate the credential and returns an error response in this case which is ignored here. I had some difficulties following the code so I may be missing something.

implementations/rust/ockam/ockam/src/credential/exchange.rs
```diff
@@ -64,68 +187,80 @@ impl CredentialExchange {
             )),
         }
     }
+
+    async fn present_credential_mutual(
```
twittner: This method is only used in tests. Should it be public?

SanjoDeundiak: Yes

implementations/rust/ockam/ockam/src/credential/exchange.rs
SanjoDeundiak: Nice catch. I'll fix, but this requires refactor in the future :D

SanjoDeundiak:
@mergifyio rebase

implementations/rust/ockam/ockam/src/credential/exchange.rs
```diff
@@ -26,18 +55,126 @@ impl CredentialExchange {
         })
     }
 
+    /// Create a generic bad request response.
+    pub fn bad_request<'a>(id: Id, path: &'a str, msg: &'a str) -> ResponseBuilder<Error<'a>> {
+        let e = Error::new(path).with_message(msg);
+        Response::bad_request(id).body(e)
+    }
+
+    #[allow(clippy::too_many_arguments)]
```
twittner, line 64: This should not be needed anymore.

implementations/rust/ockam/ockam_node/src/api.rs
```diff
@@ -28,6 +28,7 @@ where
         body   = %req.header().has_body(),
         "-> {label}"
     };
+    // TODO: Check IdentityId here?
```
twittner, line 31: Could you elaborate why this should be done here?

twittner (approved):

PR #3268, opened 2022-08-18
fix(rust): align serialisations of `IdentityIdentifier`

twittner:
Currently `IdentityIdentifier`s are serialized differently, depending on the method:

- `IdentityIdentifier::to_string` yields a string with a leading 'P'.
- Serializing via serde or minicbor will not prepend the 'P' prefix.

When deserializing from a string via `TryFrom`, the 'P' must be present and is removed. Deserializing via serde and minicbor accepts any string. In practice this can cause problems if the `to_string` representation is read via serde, as the key ID itself would contain the 'P' character.

This commit suggests the following approach:

The 'P' is present in all serialisations and all deserialisations expect it to be there or error otherwise.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2293,6 +2293,7 @@ dependencies = [
  "ockam_node",
  "ockam_transport_tcp",
  "ockam_vault",
+ "quickcheck",
  "rand 0.8.5",
  "rand_xorshift",
  "serde",
--- a/implementations/rust/ockam/ockam_command/src/old/identity.rs
+++ b/implementations/rust/ockam/ockam_command/src/old/identity.rs
@@ -106,7 +106,7 @@ pub fn parse_identities(idents: &str) -> anyhow::Result<Vec<IdentityIdentifier>>
                     "Failed to parse identifier (should be 64 hexadecimal ASCII characters): {s:?}"
                 ))
             } else {
-                Ok(IdentityIdentifier::from_key_id(s.into()))
+                Ok(IdentityIdentifier::from_key_id(s))
             }
         })
         .collect()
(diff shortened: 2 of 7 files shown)


SanjoDeundiak:
@polvorin @hairyhum is that change should also be applied to elixir to avoid breaking things?
@mrinalwadhwa do you think it's time to change "P" from Profile to "I" from "Identity"?

polvorin:
> @polvorin @hairyhum is that change should also be applied to elixir to avoid breaking things? @mrinalwadhwa do you think it's time to change "P" from Profile to "I" from "Identity"?

I think it's ok either way as long as it's consistent :+1: ,  elixir code treat them as opaque strings

mrinalwadhwa:
My opinion is the prefix shouldn't exist at all ... but we can come back to that conversation in a some months once functionality stabilizes a bit more. In the meantime a consistent serialization makes sense. Use of `I` instead of `P` also makes sense.

twittner:
Would it be a problem to remove the prefix now if that is anyway the goal?

mrinalwadhwa:
Not a decided goal. We should collect everyone's opinion and have a discussion about pros & cons... when we have some time to focus on that question.

twittner:
I think this decision needs to happen very soon. A change of the `IdentityIdentifier` format is a breaking change and values of this type are persisted and exchanged in various places which would be hard to update simultaneously.

mrinalwadhwa:
That's a good point. Let's change to the I-prefix and come back to this conversation in a couple of weeks.

twittner:
> A change of the `IdentityIdentifier` format is a breaking change and values of this type are persisted and exchanged in various places which would be hard to update simultaneously.

This could already be witnessed by the failing tests when changing from a "P" prefix to an "I" prefix: https://github.com/build-trust/ockam/runs/7918942832?check_suite_focus=true

I reverted to "P" as this PR was anyway meant to be a bugfix only.

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3280, opened 2022-08-19
feat(rust): derive `Default` for `Id` and `Error`

twittner:
To allow construction of responses and errors without having access to the request (e.g. because decoding it failed), this commit derives `Default` for `Id` and ensures that random Ids are never 0. It also makes the `path` in `Error` optional, so that an error response can be constructed without knowing the request path. This functionality should be used judiciously, e.g. only when request decoding failed, otherwise when the data is available it should be used.

### DIFF
--- a/implementations/rust/ockam/ockam_core/src/api.rs
+++ b/implementations/rust/ockam/ockam_core/src/api.rs
@@ -116,7 +116,7 @@ pub fn internal_error<'a>(r: &'a Request, msg: &'a str) -> ResponseBuilder<Error
 }
 
 /// A request/response identifier.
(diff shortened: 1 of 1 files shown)


SanjoDeundiak:
@twittner If a client sent us something that is not a valid `Request`, unlikely that he expects `Response` in response. Thoughts?

SanjoDeundiak:
PR itself is fine though

twittner:
> If a client sent us something that is not a valid `Request`, unlikely that he expects `Response` in response. Thoughts?

Maybe. We do not know and I think it is still better to return a well-formed `Response` and keep our end of the API contract.

SanjoDeundiak:
@mergifyio rebase

SanjoDeundiak (approved):

PR #3294, opened 2022-08-22
feat(rust): add launch configuration to start services

twittner:
Example:

```json
{
  "startup_services": {
    "vault": {},
    "identity": {
      "address": "identity_service"
    },
    "verifier": {
      "address": "verify"
      "disabled": true
    },
    "authenticator": {
      "enrollers": "/tmp/enrollers.json",
      "project": "p42"
    }
  }
}
```
Opening as a draft PR as it currently abuses the existing `launch_config` parameter and needs better integration with the rest of the system.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -10,6 +10,9 @@ use std::{
     str::FromStr,
 };
 
+use crate::secure_channel_listener::create as secure_channel_listener;
+use crate::service::config::Config;
+use crate::service::start::{self, StartCommand, StartSubCommand};
 use crate::util::{bind_to_port_check, exitcode};
 use crate::{
     node::show::query_status,
@@ -28,6 +31,7 @@ use ockam_api::{
     nodes::models::transport::{TransportMode, TransportType},
     nodes::{NodeManager, NODEMANAGER_ADDR},
 };
+use ockam_core::LOCAL;
 use ockam_vault::storage::FileStorage;
 use ockam_vault::Vault;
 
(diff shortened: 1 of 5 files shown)


twittner:
CC: @polvorin, @hairyhum, @spacekookie

hairyhum:
> "disabled": true

What does that mean? Why is it in the config if it's disabled?

Also can you please do two configs, one for the sidecar and one for the CA node?

twittner:
> > "disabled": true
> 
> What does that mean? Why is it in the config if it's disabled?

It means that the service is not started and allows for some flexibility. Since JSON does not support comments and it may sometimes be useful to temporarily disable the configuration for a service without having to delete the part from the file, "disabled" can be set to true. The setting is entirely optional.

> Also can you please do two configs, one for the sidecar and one for the CA node?

What do you mean? As implemented one provides a config file with the services to start. The configuration can be different per node.

hairyhum:
> What do you mean? As implemented one provides a config file with the services to start. The configuration can be different per node.

Can you give examples of a node, which requires secure channel to connect to and a node which doesn't? Or it's not controlled by this configuration?

mrinalwadhwa:
@twittner this looks great! I didn't get a chance to try it today but I will tomorrow. Nice!!

twittner:
> Can you give examples of a node, which requires secure channel to connect to and a node which doesn't? Or it's not controlled by this configuration?

It was not. I now added a config section with the same parameters as the command-line interface to start a secure-channel listener, except that it is always started at the local node.

Example 1 (using defaults): 

```
"secure_channel_listener": {}
```

Example 2:

```
"secure_channel_listener": {
  "address": "api",
  "authenticated_identifiers": ["P18face426da251fbee0c0d3a564547c2c5b38c9b1359574eb4aa7e4dd526c00a"]
}
```

hairyhum (dismissed):

mrinalwadhwa (approved):

PR #3295, opened 2022-08-22
feat(rust): create `PublicIdentity`, cleanup identity

SanjoDeundiak:


implementations/rust/ockam/ockam_api/src/verifier.rs
```diff
@@ -95,27 +94,22 @@ where
...
-        if now >= data.expires_at() {
-            let err = Error::new("/verify").with_message("expired");
-            return Ok(Either::Left(Response::forbidden(id).body(err)));
-        }
+        let data = match ident
+            .verify_credential(cre, req.subject(), &self.vault)
+            .await
+        {
+            Ok(data) => data,
+            Err(_) => {
```
twittner: Would be helpful to preserve the error information (at least in logs).

The whole method can be simplified somewhat:


```rust
async fn verify<'a>(
    &self,
    id: Id,
    req: &'a VerifyRequest<'a>,
    cre: &'a Credential<'a>,
) -> Result<Either<ResponseBuilder<Error<'_>>, CredentialData<'a, Verified>>> {
    let data = CredentialData::try_from(cre)?;

    if let Some(ident) = req.authority(data.unverfied_issuer()) {
        let ident = PublicIdentity::import(ident, &self.vault).await?;
        let data = ident.verify_credential(cre, req.subject(), &self.vault).await?;
        Ok(Either::Right(data))
    } else {
        let err = Error::new("/verify").with_message("unauthorised issuer");
        Ok(Either::Left(Response::unauthorized(id).body(err)))
    }
}
```

implementations/rust/ockam/ockam_identity/src/credential.rs
```diff
@@ -1,90 +1,316 @@
...
+    /// The time when this credential was created.
+    #[n(6)] created: Timestamp,
+    /// The time this credential expires.
+    #[n(7)] expires: Timestamp,
+    /// Term to represent the verification status type.
+    #[n(8)] status: Option<PhantomData<T>>
+}
+
+impl<'a> CredentialData<'a, Unverified> {
+    pub(crate) fn make_verified(self) -> CredentialData<'a, Verified> {
```
twittner:
```suggestion
    pub(crate) fn into_verified(self) -> CredentialData<'a, Verified> {
```

Cf. https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

implementations/rust/ockam/ockam_api/src/verifier.rs
SanjoDeundiak: @twittner the difference is that I have 2 different error messages returned to the client depending if authority wasn't found in the list or verification failed. No big deal if it's one error, but I wouldn't call that "simplification". Therefore, I would leave as is

twittner: At least log the error message.

twittner (changes requested):

implementations/rust/ockam/ockam_api/src/verifier.rs
SanjoDeundiak: @twittner I think we want to develop an approach for better error handling in our Services overall. My suggestion is that there should be a clear boundary between following:
1. Serving a service (deserializing requests, responding with responses) 
1. Performing an action caused by the request
 
In this case `2` should operate using ockam rust errors, while `1` should return meaningful error in case of:
1. Invalid requests
1. Transforming underlying error `1` returned wrapped in a proper Response

This should happen across whole codebase, maybe some changes to rust error should also happen to make this possible. Until then, errors we return in services are a mess and will be a mess. But, if you think propagating error in this specific function is an improvement worth time we've spent - I'll push the change

twittner: I agree with the distinction and I am not against the generic error response message, but once an error response is received it may be necessary to analyse logs to understand what went wrong and seeing a potentially more detailed error in the logs can save a lot of time. Regardless of where the code is right now I think errors should always be logged, unless there is a good reason not to (e.g. because of sensitive information).

twittner (approved):

PR #3304, opened 2022-08-22
feat(rust): route to authority as part of project responses

polvorin:
WIP
@adrianbenavides  please take over from here.     Just need to display this as part of the project' info,  and update the tests.  (server is already including this field on responses so you can test it)

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cloud/project.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/project.rs
@@ -22,6 +22,8 @@ pub struct Project<'a> {
     #[b(6)] pub users: Vec<CowStr<'a>>,
     #[b(7)] pub space_id: CowStr<'a>,
     #[b(8)] pub identity: Option<CowStr<'a>>,
+    #[b(9)] pub authority_access_route: Option<CowStr<'a>>,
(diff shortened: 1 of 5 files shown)


implementations/rust/ockam/ockam_api/src/cloud/project.rs
```diff
@@ -19,7 +20,9 @@ pub struct Project<'a> {
     #[b(5)] pub access_route: CowStr<'a>, //TODO: should be optional, waiting for changes on the elixir side
     #[b(6)] pub users: Vec<CowStr<'a>>,
     #[b(7)] pub space_id: CowStr<'a>,
-    #[b(8)] pub identity: Option<CowStr<'a>>,
+    #[b(8)] pub identity: Option<IdentityIdentifier>,
+    #[b(9)] pub authority_access_route: Option<CowStr<'a>>,
+    #[b(10)] pub authority_identity: Option<CowStr<'a>>,
```
twittner, line 27: This should use bytes instead of a hex encoded string.

twittner:
I included the additional info in the output and updated the tests. @adrianbenavides, please let me know if anything is missing. In a subsequent PR we should send the authority identity as bytes instead of a hex encoded string.

adrianbenavides (approved):
Thanks @twittner, LGTM!

PR #3306, opened 2022-08-23
feat(rust): add method to create projects' secure channels given an input multiaddr

### LINKED ISSUE #3210: Enroll command should create project and space if they don't exist
When enrolling an identity with the controller of the orchestrator the `enroll` command should create a project and space if they don't already exist.

The enroll command should also add the project ID and access route into the configuration lookup table (via a `/project/default` key) for future use

### LINKED ISSUE #3239: Block when in space create, project create or ockam enroll
Currently, when `ockam space create`, `ockam project create` and `ockam enroll` are run, the command immediately returns after sending a request to the Orchestrator even though the space/project is not created or the enrollment has not completed. Any further commands that are run during this time won't succeed.

To not run into this issue, we should block until `ockam space create`, `ockam project create` and `ockam enroll` are complete. While it is blocked, it should keep polling the Orchestrator to know when to stop.

### LINKED ISSUE #3250: Replace /project alias values via configuration lookup
When using a `/project/default` (or similar) key in the `ockam forwarder` or `ockam secure-channel` commands to construct a message to the project access route with the correct identifier.

### LINKED ISSUE #3279: Refactor `ockam message send` command to use the new simpler utility functions.
In https://github.com/build-trust/ockam/pull/3168 @adrianbenavides introduced some simplifications to how various ockam commands interact with other ockam nodes.

We would like to change `ockam message send` to use this newer style.
The code for `ockam message send` is here https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/message/send.rs

Here's a PR that did a similar refactor of `ockam secure-channel-listener list`
https://github.com/build-trust/ockam/pull/3267/files

Related #https://github.com/build-trust/ockam/issues/3241

adrianbenavides:
Built on top of https://github.com/build-trust/ockam/pull/3312. We have to merge that first.

Project lookup added to:
- `forwarder create`
- `message send`
- `secure-channel create`
- `project add-member`
- `project get-credential`

Other changes:
- `enroll` now does not create the secure channel, as it will be created when doing the project lookup
- `project create` command now updates the config file with the project info, so it can be used by the lookup method
- Refactor `Error` struct to simplify its use. It can now be built passing an `anyhow` error so that we can output the error message, instead of just the exit code

## :warning: Important note :warning:

Currently, these changes will only work if the projects' info is stored in the config file, which is done when creating projects with the `enroll` command and the `project create` command.

When a project name from the input MultiAddr doesn't exist in the config file it will try to fetch it from the cloud to create the secure channel. This logic won't work until we have an API method that enables us to retrieve projects by name. We should add this endpoint on the elixir side.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cloud/project.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/project.rs
@@ -1,9 +1,11 @@
 use minicbor::{Decode, Encode};
 use serde::Serialize;
+use std::str::FromStr;
 
 use ockam_core::CowStr;
 #[cfg(feature = "tag")]
 use ockam_core::TypeTag;
+use ockam_multiaddr::MultiAddr;
 
 #[derive(Encode, Decode, Serialize, Debug)]
 #[rustfmt::skip]
@@ -46,6 +48,10 @@ impl Project<'_> {
 
     pub fn is_ready(&self) -> bool {
         !self.access_route.is_empty() & self.identity.is_some()
+    }
+
+    pub fn access_route(&self) -> MultiAddr {
+        MultiAddr::from_str(&self.access_route).expect("Invalid access route")
     }
 }
 
--- a/implementations/rust/ockam/ockam_api/src/config/atomic.rs
+++ b/implementations/rust/ockam/ockam_api/src/config/atomic.rs
@@ -9,6 +9,7 @@
 //! generates an AtomicUpdater.
 
 use crate::config::ConfigValues;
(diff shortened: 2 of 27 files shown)


mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

mrinalwadhwa:
The `addr` argument from `CloudOpts` added as a positional on every command I think will create problems/confusion in places we need other positional. I think we should make it a hidden option `--route-to-controller`

PR #3312, opened 2022-08-24
fix(rust): mutliaddr support for projects

adrianbenavides:
Needed for https://github.com/build-trust/ockam/pull/3306

* Fixes support for `project` prefix on MultiAddr.
* Add missing config updates

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/config/lookup.rs
+++ b/implementations/rust/ockam/ockam_api/src/config/lookup.rs
@@ -11,8 +11,10 @@ use std::{
 #[derive(Debug, Default)]
 pub struct LookupMeta {
     /// Append any project name that is encountered during look-up
-    pub project: VecDeque<String>,
+    pub project: VecDeque<Name>,
 }
+
+pub type Name = String;
 
 /// A generic lookup mechanism for configuration values
 #[derive(Clone, Debug, Serialize, Deserialize)]
@@ -52,13 +54,13 @@ impl ConfigLookup {
     /// Store a project route and identifier as lookup
     pub fn set_project(
         &mut self,
-        prj_name: String,
+        name: String,
         node_route: String,
         id: String,
         identity_id: String,
     ) {
         self.map.insert(
-            format!("/project/{}", prj_name),
+            format!("/project/{}", name),
             LookupValue::Project(ProjectLookup {
                 node_route,
                 id,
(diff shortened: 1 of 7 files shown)


implementations/rust/ockam/ockam_api/src/config/lookup.rs
```diff
@@ -11,9 +11,11 @@ use std::{
 #[derive(Debug, Default)]
 pub struct LookupMeta {
     /// Append any project name that is encountered during look-up
-    pub project: VecDeque<String>,
+    pub project: VecDeque<Name>,
```
twittner, line 14: `LookupMeta::project` is never read. What is the purpose of this field?

implementations/rust/ockam/ockam_command/src/util/config.rs
```diff
@@ -265,25 +262,38 @@ impl OckamConfig {
...
-
         Ok(())
     }
 }
 
+impl Drop for OckamConfig {
+    fn drop(&mut self) {
+        // To reduce temporary coupling between config setters/updaters and actually
+        // updating the config file, we try to update the config file when it is dropped.
+        let _ = self.atomic_update().run();
```
twittner: Could you explain a bit more why this is needed?

implementations/rust/ockam/ockam_api/src/config/lookup.rs
adrianbenavides: This is needed in https://github.com/build-trust/ockam/pull/3306. I've separated the `multiaddr` related changes to keep the other PR simpler.

implementations/rust/ockam/ockam_command/src/util/config.rs
adrianbenavides: Every time you call one of the setters/updaters you have to remember to call the `atomic_update` method. We already forgot about calling the `atomic_update` method in a couple of places, so I thought it would be nice to force calling it somehow.

For example, [this command](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/configuration/set.rs#L26) was not updating the config file, and therefore, it was doing nothing.

twittner: I see. I am not sure that the `Drop` impl is the right place. For now, can you at least log the error?

adrianbenavides: Sure, done!

twittner (approved):

implementations/rust/ockam/ockam_command/src/util/config.rs
adrianbenavides: I think I've found a case where this won't work. Let me check something before merging this.

adrianbenavides:
@twittner I've removed the `Drop` implementation for the `OckamConfig` because it was producing a dead lock when running the `node delete` command. Instead of doing that, I've added the `config update` calls where they were missing.

adrianbenavides:
Somehow I pushed those changes to the other PR. I've created a new one just to delete de `Drop` impl so we don't have the deadlock on `develop`: https://github.com/build-trust/ockam/pull/3315

PR #3328, opened 2022-08-24
Credential commands

SanjoDeundiak:


implementations/rust/ockam/schema.cddl
```diff
@@ -59,7 +59,7 @@ attributes = {
 }
 
 attribute = {
-    ?0: 6844116,
+    ?0: 4592146,
```
SanjoDeundiak, line 62: @hairyhum @polvorin FYI

mrinalwadhwa:
I cherry picked the multi-process config fix and merged it here https://github.com/build-trust/ockam/pull/3345#issue-1351533145 .. this PR branch is now rebased to a develop that already has that fix.

mrinalwadhwa:
Had to revert the cherry-picked commit. This error started happening very often after applying that change.

```
» ockam enroll && bats implementations/rust/ockam/ockam_command/tests/commands.bats
thread 'tokio-runtime-worker' panicked at 'Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.', /Users/mrinal/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.19.2/src/runtime/enter.rs:39:9
Error channel closed

Caused by:
    channel closed
```

removing the commit made that error go away.

The commit is [here](https://github.com/build-trust/ockam/pull/3345) if you want to apply it back.

implementations/rust/ockam/ockam_api/src/nodes/models/credentials.rs
```diff
@@ -0,0 +1,70 @@
...
+use ockam_core::TypeTag;
+use ockam_multiaddr::MultiAddr;
+
+#[derive(Clone, Debug, Decode, Encode)]
+#[rustfmt::skip]
+#[cbor(map)]
+pub struct SetAuthorityRequest {
+    #[cfg(feature = "tag")]
+    #[n(0)] tag: TypeTag<1864428>,
+    #[b(1)] pub authorities: Vec<Vec<u8>>,
```
twittner, line 17: Would be nice to have a bit more type information here.

implementations/rust/ockam/ockam_api/src/nodes/models/credentials.rs
```diff
@@ -0,0 +1,70 @@
...
+    #[b(1)] pub route: CowStr<'a>,
+    #[n(2)] pub overwrite: bool,
+}
+
+impl<'a> GetCredentialRequest<'a> {
+    pub fn new(route: &MultiAddr, overwrite: bool) -> Self {
+        Self {
+            #[cfg(feature = "tag")]
+            tag: TypeTag,
+            route: route.to_string().into(),
```
twittner, line 45: `route : MultiAddr` should work now.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -67,8 +69,9 @@ pub struct NodeManager {
 
     skip_defaults: bool,
 
-    pub(crate) vault: Option<Vault>,
+    vault: Option<Vault>,
     identity: Option<Identity<Vault>>,
+    authorities: Option<Vec<PublicIdentity>>,
```
twittner, line 74:
```suggestion
    authorities: Vec<PublicIdentity>,
```

implementations/rust/ockam/ockam_command/src/util/api.rs
```diff
@@ -210,6 +210,39 @@ pub(crate) fn start_authenticated_service(addr: &str) -> Result<Vec<u8>> {
...
+    use ockam_api::nodes::models::credentials::{
+        GetCredentialRequest, PresentCredentialRequest, SetAuthorityRequest,
+    };
+
+    pub(crate) fn present_credential(
+        to: &MultiAddr,
+        oneway: bool,
+    ) -> RequestBuilder<PresentCredentialRequest> {
+        let b = PresentCredentialRequest::new(to, oneway);
+        Request::builder(Method::Post, "/node/credentials/actions/present").body(b)
```
twittner:
```suggestion
        Request::post("/node/credentials/actions/present").body(b)
```

implementations/rust/ockam/ockam_command/src/util/api.rs
```diff
@@ -210,6 +210,39 @@ pub(crate) fn start_authenticated_service(addr: &str) -> Result<Vec<u8>> {
...
+    }
+
+    pub(crate) fn set_authority(authorities: &[String]) -> RequestBuilder<SetAuthorityRequest> {
+        let authorities: std::result::Result<Vec<Vec<u8>>, FromHexError> =
+            authorities.iter().map(hex::decode).collect();
+        let authorities = authorities.unwrap();
+
+        let b = SetAuthorityRequest::new(authorities);
+        Request::builder(Method::Post, "/node/credentials/authority").body(b)
```
twittner:
```suggestion
        Request::post("/node/credentials/authority").body(b)
```

implementations/rust/ockam/ockam_command/src/util/api.rs
```diff
@@ -210,6 +210,39 @@ pub(crate) fn start_authenticated_service(addr: &str) -> Result<Vec<u8>> {
...
+    }
+
+    pub(crate) fn get_credential(
+        from: &MultiAddr,
+        overwrite: bool,
+    ) -> RequestBuilder<GetCredentialRequest> {
+        let b = GetCredentialRequest::new(from, overwrite);
+        Request::builder(Method::Post, "/node/credentials/actions/get").body(b)
```
twittner:
```suggestion
        Request::post("/node/credentials/actions/get").body(b)
```

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -79,6 +82,26 @@ pub struct IdentityOverride {
...
+
+    pub(crate) fn vault(&self) -> Result<&Vault> {
+        self.vault
+            .as_ref()
+            .ok_or_else(|| ApiError::generic("Vault doesn't exist"))
+    }
+
+    pub(crate) fn authorities(&self) -> Result<Vec<PublicIdentity>> {
+        self.authorities
+            .clone()
```
twittner: Whether the list of authorities needs to be cloned or not should be for the caller to decide.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
SanjoDeundiak: Could you please elaborate why you think this should change?

implementations/rust/ockam/ockam_api/src/nodes/models/credentials.rs
SanjoDeundiak: What do you mean by "type information"?

implementations/rust/ockam/ockam_api/src/nodes/models/credentials.rs
SanjoDeundiak: You mean having MultiAddr field in the structure?
```
29 | #[derive(Clone, Debug, Decode, Encode)]
   |                        ^^^^^^ the trait `minicbor::Decode<'_, Ctx>` is not implemented for `MultiAddr`
```

implementations/rust/ockam/ockam_command/src/util/api.rs
SanjoDeundiak: Too bad this better approach wasn't propagated everywhere when added, now the old approach is copy&pasted from different places. I'll fix that

implementations/rust/ockam/ockam_api/src/nodes/service.rs
twittner: There is no need for an `Option` on top of a `Vec` when the `Vec` already has a neutral element. If the vector is empty there are no authorities configured. Adding a `Option` on top of that only complicates things as can be seen by the `authorities` getter method for instance.

implementations/rust/ockam/ockam_api/src/nodes/models/credentials.rs
twittner: I mean a more specific type, e.g. a `Vec<PublicIdentity>`.

implementations/rust/ockam/ockam_api/src/nodes/models/credentials.rs
twittner: Yes, it requires the `cbor` feature of `ockam_multiaddr`.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
SanjoDeundiak: I think it's important to differentiate "haven't been configured" value from "configured to be empty" value. Otherwise it's very easy to cause bugs. E.g. one of such bugs I found recently with `pid` value set to be `Some(0)` instead of `None`, which then lead to panics

SanjoDeundiak: If it doesn't have `Option` wrapper, we'll eventually forget to check for emptiness in some place

implementations/rust/ockam/ockam_api/src/nodes/models/credentials.rs
SanjoDeundiak: @twittner any reason this feature is not enabled by default?

implementations/rust/ockam/ockam_api/src/nodes/models/credentials.rs
SanjoDeundiak: Enabling cbor for `PublicIdentity` would take a lot of time considering how big and composite this struct is + how time consuming is adding cbor attributes

implementations/rust/ockam/ockam_api/src/nodes/service.rs
twittner: I do not think there is a meaningful difference between not having an authority because none has been configured and not having an authority because it has been configured like that. Validating a credential can only succeed if there is a matching authority. In that sense it is not possible to "forget to check".

implementations/rust/ockam/ockam_api/src/nodes/models/credentials.rs
SanjoDeundiak: And too bad when cbor support was added it wasn't propagated to the places where it could be useful...

SanjoDeundiak: I'll fix that...

implementations/rust/ockam/ockam_api/src/nodes/models/credentials.rs
twittner: Not sure. The `Encode` and `Decode` impls could use the existing `PublicIdentity::export` and `PublicIdentity::import` functionality I guess. It can be done in another PR.

SanjoDeundiak: 1. Then it will use serde bare under the hood
2. `Identity` and `PublicIdentity` types also provide some guarantees about the data they contain, you shouldn't be able to construct those types with invalid inner `ChangeHistory`, this is why their import functions are doing checks, which are not possible to do during simple Decode

implementations/rust/ockam/ockam_api/src/nodes/models/credentials.rs
twittner: My opinion is that features should be opt-in, especially when they activate additional dependencies. You can change that of course.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
SanjoDeundiak: Yes, it won't end up with accidentally verified credential either way. But the error should be different. "Not configured authorities" vs "Trusted authority not found". The first one obviously point out what needs to be fixed, the second one makes you think that authorities were configured, but they don't match for some reason

implementations/rust/ockam/ockam_api/src/nodes/models/credentials.rs
twittner: > this is why their import functions are doing checks, which are not possible to do during simple Decode

Could you not create a `Vault` in a `Decode` impl and use it to do the validation?

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -79,6 +82,26 @@ pub struct IdentityOverride {
...
+        self.vault
+            .as_ref()
+            .ok_or_else(|| ApiError::generic("Vault doesn't exist"))
+    }
+
+    pub(crate) fn authorities(&self) -> Result<&Vec<PublicIdentity>> {
+        self.authorities
+            .as_ref()
+            .ok_or_else(|| ApiError::generic("Authorities don't exist"))
+    }
```
twittner, line 102:
```suggestion
    pub(crate) fn authorities(&self) -> Result<&[PublicIdentity]> {
        self.authorities
            .as_deref()
            .ok_or_else(|| ApiError::generic("Authorities don't exist"))
    }
```

twittner (approved):

PR #3334, opened 2022-08-25
fix(rust): creation of static forwarder without heartbeats

adrianbenavides:
Adds a more elaborate logic to derive from the `--at` argument if it references a local node or a remote node. Still a hacky approach until rust forwarder service supports the heartbeat protocol.

Also, the `address` argument has been removed because it was not being used and the `for_node` argument has been renamed to `to`.

Finally, the forwarder service has been moved to its own module (no logic changes done, just moved the `create_forwarder` function).

```bash
$ ockam forwarder create --from forwarder_to_n2 --to /node/n2 --at /node/n1
/service/forwarder_to_n2

$ ockam forwarder create --to /node/n2 --at /node/n1
/service/register # !!! this is not what we want

$ ockam forwarder create --from /service/forwarder_to_d --to /node/default --at /project/default
/service/forward_to_forwarder_to_d

$ ockam forwarder create --at /project/default --to /node/default
/service/FWD_9a311b2755e4c6c0
```

## Note

When executing the command the way we agreed to, it doesn't return a consistent address for the forwarder:

```bash
$ ockam forwarder create --at /project/default --from /service/forwarder_to_d --for /node/default
/service/forward_to_forwarder_to_d
```

Should we stop prepending `forwarder_to` to the forwarder name? If not, it doesn't make too much sense to pass the alias argument as `/service/<alias>`.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/error.rs
+++ b/implementations/rust/ockam/ockam_command/src/error.rs
@@ -24,7 +24,7 @@ impl Error {
 
 impl Display for Error {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        std::fmt::Display::fmt(&self.inner, f)
+        std::fmt::Debug::fmt(&self.inner, f)
     }
 }
 
--- a/implementations/rust/ockam/ockam/src/remote.rs
+++ b/implementations/rust/ockam/ockam/src/remote.rs
@@ -187,7 +187,7 @@ impl RemoteForwarder {
         let registration_route = hub_route
             .into()
             .modify()
-            .append("forwarding_service")
+            .append("static_forwarding_service")
             .into();
 
         // let remote_address = Address::random_local().without_type().to_string();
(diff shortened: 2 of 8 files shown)


adrianbenavides:
@Mergifyio rebase

mrinalwadhwa:
> Should we stop prepending forwarder_to to the forwarder name?

The service should not force the prefix. The address should be what the user asked for @polvorin @hairyhum

implementations/rust/ockam/ockam_command/src/error.rs
```diff
@@ -24,7 +24,7 @@ impl Error {
 
 impl Display for Error {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        std::fmt::Display::fmt(&self.inner, f)
+        std::fmt::Debug::fmt(&self.inner, f)
```
adrianbenavides, line 27: This is to output the anyhow's "caused by" message, instead of just the last error message, and at the same time keep the `ockam_command`'s Error exit code hidden.

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3336, opened 2022-08-25
feat(rust): add authorities config

twittner:


implementations/rust/ockam/ockam_command/src/util/config.rs
```diff
@@ -391,3 +396,25 @@ impl StartupConfig {
         self.inner.readlock_inner().commands.clone()
     }
 }
+
+pub struct AuthoritiesConfig {
+    inner: Config<cli::AuthoritiesConfig>,
```
SanjoDeundiak, line 401: Nice, ideally all configs should use that approach

implementations/rust/ockam/ockam_api/src/config/cli.rs
```diff
@@ -140,3 +143,35 @@ impl ConfigValues for StartupConfig {
         Self::default()
     }
 }
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct AuthoritiesConfig {
+    authorities: BTreeMap<Vec<u8>, MultiAddr>,
```
SanjoDeundiak, line 149: That is fine, but for small number of elements it's actually makes sense to use `Vec` (faster, simpler, smaller memory footprint)

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -196,6 +193,12 @@ impl NodeManager {
         Ok(s)
     }
 
+    pub async fn configure_authorities(&mut self, ac: AuthoritiesConfig) -> Result<()> {
+        let v = Vault::default();
```
SanjoDeundiak: Ideally we should use existing `Vault` instance. The idea is that Vault is pluggable, it's hard to achieve that if it's created many times in different places

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -329,12 +335,16 @@ async fn run_background_node_impl(
         Some(get_identity_override(ctx, &cfg).await?)
     };
 
+    if let Some(path) = &c.project {
```
SanjoDeundiak, line 322: Have you tested it? I may be wrong, but that looks like it shouldn't work, because `project` value is not forwarded in `startup::spawn_node`

SanjoDeundiak (changes requested):

implementations/rust/ockam/ockam_command/src/node/create.rs
twittner: Changed in https://github.com/build-trust/ockam/pull/3336/commits/8a1b8744d23f46fc8d3bd454471a902b3933a099

implementations/rust/ockam/ockam_api/src/nodes/service.rs
twittner: Changed in https://github.com/build-trust/ockam/pull/3336/commits/eb01611e5f7a513429a2f3df1665fc639aa25b62. Let me know if you want to have `configure_authorities` return an error instead of falling back to a default vault if none exists.

SanjoDeundiak: Yeah, ideally it should return an error. Not critical for now though

SanjoDeundiak (approved):

PR #3344, opened 2022-08-25
refactor(rust): move `auth0::node` to `util`

lameferret:
Fixes: [TODO: move to util/node module](https://github.com/build-trust/ockam/blob/8f4d0391eae726a19d6e204189111474d447830b/implementations/rust/ockam/ockam_command/src/enroll/auth0.rs#L58)


mrinalwadhwa (approved):
That looks great. Thank you!
Thank you for another awesome contribution! 🎉

PR #3353, opened 2022-08-26
feat(rust): use `MultiAddr` cbor support

SanjoDeundiak:
<

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cloud/project.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/project.rs
@@ -2,7 +2,6 @@ use minicbor::{Decode, Encode};
 use ockam_core::CowStr;
 use ockam_identity::IdentityIdentifier;
 use serde::Serialize;
-use std::str::FromStr;
 
 #[cfg(feature = "tag")]
 use ockam_core::TypeTag;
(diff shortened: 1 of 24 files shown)


implementations/rust/ockam/ockam_api/src/cloud/project.rs
```diff
@@ -19,6 +19,7 @@ pub struct Project<'a> {
     #[b(2)] pub name: CowStr<'a>,
     #[b(3)] pub space_name: CowStr<'a>,
     #[b(4)] pub services: Vec<CowStr<'a>>,
+    // TODO: Implement Serialize for MultiAddr and change type here
```
SanjoDeundiak, line 22: @twittner

twittner (approved):

implementations/rust/ockam/ockam_api/src/cloud/project.rs
```diff
@@ -355,7 +355,8 @@ mod tests {
                     name: String::arbitrary(g).into(),
                     space_name: String::arbitrary(g).into(),
                     services: vec![String::arbitrary(g).into(), String::arbitrary(g).into()],
-                    access_route: String::arbitrary(g).into(),
+                    access_route: MultiAddr::from_str(&format!("node/{}", String::arbitrary(g)))
+                        .unwrap(),
```
twittner, line 359: The string here can not be arbitrary, as the '/' has special meaning when parsing textual multi-addresses.

```suggestion
                    access_route: {
                        let mut ma = MultiAddr::default();
                        ma.push_back(ockam_multiaddr::proto::Node::new(String::arbitrary(g)).unwrap();
                        ma
                    },
```

twittner (changes requested):
The schema.cddl needs to be updated as the project's access route is no longer a text string and this change probably requires coordination with the elixir team. Short-term it may be better to stick with text for interoperability with Elixir.

PR #3383, opened 2022-08-30
fix(rust): check if project node can establish a secure channel before marking it as ready

adrianbenavides:
* `is_ready` stops checking authority fields, since it seems that we now always receive `None` on the `authority access route`
* when creating a project on the `enroll` command, it now waits until a secure channel can be established. In my tests, I've found that after a project is TCP-ready, sometimes it establishes the secure channel right away, and sometimes it takes 4-8 seconds until it can do it (I could only reproduce this case one time out of 10-15 runs)

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cloud/project.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/project.rs
@@ -87,10 +87,7 @@ impl Project<'_> {
     }
 
     pub fn is_ready(&self) -> bool {
-        !(self.access_route.is_empty()
-            || self.authority_access_route.is_none()
-            || self.identity.is_none()
-            || self.authority_identity.is_none())
+        !(self.access_route.is_empty() || self.identity.is_none())
     }
 
     pub async fn is_reachable(&self) -> Result<bool> {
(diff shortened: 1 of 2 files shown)


twittner:
> it seems that we now always receive `None` on the `authority access route`

Is this intentional?

adrianbenavides:
> > it seems that we now always receive `None` on the `authority access route`
> 
> Is this intentional?

I don't know, but yesterday we were receiving that field. We don't need it to check if the project is ready, so that's why I just removed it. (cc @polvorin)

polvorin:
> > > it seems that we now always receive `None` on the `authority access route`
> > 
> > 
> > Is this intentional?
> 
> I don't know, but yesterday we were receiving that field. We don't need it to check if the project is ready, so that's why I just removed it. (cc @polvorin)

it's not.   In which project this is happening?.    Or on newly created ones? .   How long after the creation it still didn't have an authority returned?

polvorin:
investigating this,  I see it as None on a project of mine as well :thinking:

mrinalwadhwa (dismissed):

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3384, opened 2022-08-30
fix(rust): authority config keys must be strings

twittner:
Change the authorities JSON format from `{ Vec<u8> -> MultiAddr }` to `{ IdentityIdentifier -> { identity : String, MultiAddr } }` with `identity` as the hex-encoded bytes.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/config/cli.rs
+++ b/implementations/rust/ockam/ockam_api/src/config/cli.rs
@@ -5,9 +5,10 @@ use crate::config::{
     snippet::ComposableSnippet,
     ConfigValues,
 };
+use crate::HexByteVec;
 use directories::ProjectDirs;
 use ockam_core::Result;
-use ockam_identity::{IdentityVault, PublicIdentity};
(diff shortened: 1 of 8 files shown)


implementations/rust/ockam/ockam_api/src/config/cli.rs
```diff
@@ -175,3 +176,26 @@ impl ConfigValues for AuthoritiesConfig {
         Self::default()
     }
 }
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct Authority {
+    identity: HexByteVec,
```
twittner, line 182: Ideally we would use `PublicIdentity` here, but the type does not implement `Serialize` or `Deserialize`. It is a bit sad, because in `ockam_command` we already have to import the bytes as a `PublicIdentity` in order to get the `IdentityIdentifier` only to serialize it again and converting it back yet again in `AuthoritiesConfig::to_public_identities`.

SanjoDeundiak (approved):

PR #3388, opened 2022-08-31
fix(rust): `node create` command waits until node is up before returning response

adrianbenavides:
I've tested this by artificially delaying the foreground node creation by 5 seconds.


antoinevg (dismissed):
Good workaround for now but we should probably track down a Rusty IPC lock implementation for the longer term.

mrinalwadhwa (approved):

mrinalwadhwa (approved):

PR #3389, opened 2022-08-31
feat(rust): add project info command

twittner:
Previously the output of project show was given to new members that enroll, however it contained private information (e.g. the users list) which are of no concern to an individual new member. This PR adds a project info structure that contains only the relevant information for the enrolling member.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -10,6 +10,7 @@ use std::{
     str::FromStr,
 };
 
+use crate::project::ProjectInfo;
 use crate::secure_channel::listener::create as secure_channel_listener;
 use crate::service::config::Config;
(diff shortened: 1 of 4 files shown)


adrianbenavides:
How is the `project info` command different from the `project show`? Couldn't we just modify the `impl Output for Project<'_>` to hide the fields you mentioned?

twittner:
> How is the `project info` command different from the `project show`?

The command is identical but instead of printing all project data it collects the relevant subset to print. One could probably refactor `project show` to not immediately print its output, but instead return the `Project` and `project info` could reuse that part then.

> Couldn't we just modify the `impl Output for Project<'_>` to hide the fields you mentioned?

That depends on the requirements. Even if it would work right now though it may be preferrable to have a separate command tailored for the use-case of producing data that is meant to be shared with enrolling members.

adrianbenavides (approved):
OK, I understand now, thanks for the clarification!

PR #3390, opened 2022-08-31
feat(rust): refactor rpc struct so it allows working with embedded nodes

adrianbenavides:
The first stage of a major refactor where we'll gradually stop using background nodes in favor of embedded nodes. This means that we won't need to `ockam node create` to run most of the current commands.

In this PR:
* I've extended the `Rpc` struct to handle embedded nodes while keeping its previous functionality intact
* I've modified some commands (all space commands and project list) to work with an embedded node

In the following PR's, I'll be modifying more commands to use an embedded node and probably I'll have to refactor a little bit more the `Rpc` struct to make working with embedded nodes the default mode.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/enroll.rs
+++ b/implementations/rust/ockam/ockam_command/src/enroll.rs
@@ -61,7 +61,7 @@ async fn run_impl(ctx: &Context, opts: CommandGlobalOpts, cmd: EnrollCommand) ->
     };
     let cloud_opts = cmd.cloud_opts.clone();
 
-    let space = default_space(ctx, &opts, &tcp, &nc, &node_opts, &cloud_opts).await?;
+    let space = default_space(ctx, &opts, &tcp, &nc, &cloud_opts).await?;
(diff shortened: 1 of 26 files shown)


antoinevg (dismissed):
I suspect we're going to run into race conditions here once folk start using `ockam` in anger but we also have this problem in other parts of the code so this isn't a reason to not merge this PR.

Anyone (@twittner ?) know if Rust has something equivalent to `boost.interprocess`? I did come across https://crates.io/crates/named_lock which looks promising…

mrinalwadhwa:
@adrianbenavides could you resolve the conflict here with latest develop, please.

adrianbenavides:
> I suspect we're going to run into race conditions here once folk start using `ockam` in anger but we also have this problem in other parts of the code so this isn't a reason to not merge this PR.

Do you mean race conditions when accessing the shared config files?

antoinevg:
> Do you mean race conditions when accessing the shared config files?

Race conditions when creating processes. If I read the code correctly, e.g. if the parent process exits after starting child process but before child process has finished starting. I could have the wrong end of the stick here :-)

mrinalwadhwa (approved):

PR #3394, opened 2022-08-31
chore: revert commit 90ce008a108f5a8d9032288c536899c92d97b974

mrinalwadhwa:
@adrianbenavides I merged #3390 but then found out that `ockam node delete --all` hangs with that commit. Reverting for now so we can release.


mrinalwadhwa:
@adrianbenavides I merged #3390 but then found out that `ockam node delete --all` hangs with that commit. Reverting for now so we can release.

PR #3396, opened 2022-09-01
feat(rust): refactor rpc struct so it allows working with embedded nodes

adrianbenavides:
Same PR as https://github.com/build-trust/ockam/pull/3390 adding the fix to the `node delete` command, which was reported to hang when running `commands.bats`.

I could reproduce the hang when running `commands.bats` on macos, and it now works on the same setup with the changes of this branch.


antoinevg (dismissed):
Looks sane 👍

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3397, opened 2022-09-01
Add credential exchange support to secure channel involved commands

SanjoDeundiak:


implementations/rust/ockam/ockam_command/src/util/output.rs
```diff
@@ -182,16 +182,14 @@ impl Output for ShowSecureChannelResponse<'_> {
...
-                    "\n  Secure Channel:\n{} {}\n{} {}\n{} {}\n{} {}",
+                    "\n  Secure Channel:\n{} {}\n{} {}\n{} {}",
                     "  •         At: ".light_magenta(),
                     route_to_multiaddr(&route![addr.to_string()])
                         .context("Invalid Secure Channel Address")?
                         .to_string()
                         .light_yellow(),
                     "  •         To: ".light_magenta(),
                     self.route.as_ref().unwrap().light_yellow(),
-                    "  • Identifier: ".light_magenta(),
```
SanjoDeundiak, line 193: @mrinalwadhwa it was initiator id, not responder (so, always the same and equals to node's id). I removed it during resolving conflicts, tell me if you want it back

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -322,7 +328,7 @@ impl NodeManager {
                 self.set_authorities(req, dec).await?.to_vec()?
             }
             (Post, ["node", "credentials", "actions", "get"]) => {
-                self.get_credential(ctx, req, dec).await?.to_vec()?
+                self.get_credential(req, dec).await?.to_vec()?
```
adrianbenavides, line 373: Why is this method called `get` if it's setting the credential and it's not returning anything?

implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
```diff
@@ -2,32 +2,43 @@ use super::map_multiaddr_err;
...
-    pub(super) async fn create_secure_channel_impl<'a>(
+    async fn get_credential_if_needed(&self) -> Result<()> {
+        let identity = self.identity()?;
+
+        if identity.credential().await.is_some() {
+            return Ok(());
+        }
+
+        self.get_credential_impl(identity, false, "" /* FIXME */)
+            .await?;
```
adrianbenavides: IIUC, calling this method without calling the `get_credential` method first will cause an error, right? So, somehow, when we have this `FIXME` figured out, we'll be able to skip that call from the command side, is that correct?

implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
```diff
@@ -127,41 +157,37 @@ impl NodeManager {
...
+        let res = match identity.stop_secure_channel(&sc_address).await {
+            Ok(()) => {
+                trace!(%sc_address, "Removed secure channel");
+                self.registry.secure_channels.remove_by_addr(&sc_address);
+                Some(sc_address)
+            }
+            Err(err) => {
+                trace!(%sc_address, "Error removing secure channel: {err}");
+                None
+            }
```
adrianbenavides, line 182: To improve the logged info, maybe we could check if the `sc_address` is in the registry before trying to stop it, and log a "Not found" kind of message instead of the current "Error" kind of message.

```rust
let res = if !self.registry.secure_channels.contains(&sc_address) {
    trace!(%sc_address, "No secure channel found for the passed address");
    None
} else {
    match identity.stop_secure_channel(&sc_address).await {
        ...
    }
};
```

implementations/rust/ockam/ockam_api/src/nodes/service.rs
SanjoDeundiak: @adrianbenavides in this case it means that background node should go get a credential for itself

SanjoDeundiak: Path also indicates that this is an action, not a getter

implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
SanjoDeundiak:
```
So, somehow, when we have this FIXME figured out, we'll be able to skip that call from the command side, is that correct?
```
Correct

```
calling this method without calling the get_credential method first will cause an error, right?
```
Didn't get the question

implementations/rust/ockam/ockam_api/src/nodes/service.rs
adrianbenavides: ok, got it, thanks!

implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
adrianbenavides: Both questions were more or less pointing in the same direction, so we can consider them both as answered :laughing:

implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
SanjoDeundiak: Well, channels in registry and actual running channels should hopefully match 1:1, in practice that's not guaranteed. I think if we successfully stopped the channel, and registry didn't contain such channel, we're in perfectly valid state. Channel is stopped and not present in the registry. Yay

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
```diff
@@ -1,31 +1,59 @@
...
+    }
+
+    pub fn insert(
+        &mut self,
+        addr: Address,
+        route: Route,
+        authorized_identifiers: Option<Vec<IdentityIdentifier>>,
+    ) {
+        self.channels
+            .push(SecureChannelInfo::new(route, addr, authorized_identifiers))
```
twittner, line 27: I would not allow duplicate address or route registrations.

implementations/rust/ockam/ockam_api/src/lib.rs
```diff
@@ -19,6 +19,20 @@ pub mod lmdb;
 #[macro_use]
 extern crate tracing;
 
+pub struct DefaultAddress;
+
+impl DefaultAddress {
```
twittner, line 24: Why not a module?

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
```diff
@@ -1,31 +1,59 @@
...
+    }
+
+    pub fn remove_by_addr(&mut self, addr: &Address) {
+        self.channels.retain(|x| x.addr() != addr)
+    }
+
+    pub fn list(&self) -> &[SecureChannelInfo] {
+        &self.channels
+    }
```
twittner, line 36:
```suggestion
    pub fn iter(&self) -> impl Iterator<Item = &SecureChannelInfo> {
        self.channels.iter()
    }
```

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -57,6 +59,29 @@ fn map_anyhow_err(err: anyhow::Error) -> ockam_core::Error {
...
+
+impl AsRef<[AuthorityInfo]> for Authorities {
+    fn as_ref(&self) -> &[AuthorityInfo] {
+        self.0.as_ref()
+    }
+}
+
+pub(crate) struct AuthorityInfo {
+    identity: PublicIdentity,
+    addr: Option<MultiAddr>, // TODO: Should be not optional in the future
```
twittner, line 82: Why in the future?

implementations/rust/ockam/ockam_command/src/node/util.rs
```diff
@@ -40,21 +40,18 @@ pub async fn start_embedded_node(ctx: &Context, cfg: &OckamConfig) -> Result<Str
...
     tcp.listen(&bind).await?;
     let node_dir = cfg.get_node_dir_raw(&cmd.node_name)?;
-    let mut node_man = NodeManager::create(
+    let node_man = NodeManager::create(
         ctx,
         cmd.node_name.clone(),
         node_dir,
         identity_override,
         cmd.skip_defaults || cmd.launch_config.is_some(),
+        Some(&cfg.authorities(&cmd.node_name)?.snapshot()),
```
twittner, line 49: Why the `Option` if we always use a `Some`?

SanjoDeundiak: Not always

implementations/rust/ockam/ockam_api/src/nodes/service.rs
SanjoDeundiak: Because that is not compatible with existing credentials commands I added a while ago

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
SanjoDeundiak: Why do you think that's better?

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
SanjoDeundiak: Collision check wouldn't hurt, however those addresses are randomly generated, so I'm not sure what should happen so that we end up with duplicates

implementations/rust/ockam/ockam_api/src/lib.rs
SanjoDeundiak: What is the difference?

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
twittner: 1. It hides an implementation detail.
2. The only use of this when I was reviewing it was `.list().iter()` which could be simplified to just `.iter()`.

implementations/rust/ockam/ockam_command/src/node/util.rs
twittner: AFAICS the only case where it was `None` is in a test.

SanjoDeundiak: Probably yes, so I think we should leave it Optional

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
SanjoDeundiak: Well, there is actually a trade-off between:
  1. Hiding implementation detail (and therefore leaving space to change that detail in the future)
  2. Giving more detailed information about the internal structure (which may allow you to optimize certain things)

In this particular place, don't you think slice will work just fine?

SanjoDeundiak:
@mergifyio rebase

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
twittner: I do not think we miss any optimisation potential here, but feel free to ignore the suggestion.

implementations/rust/ockam/ockam_command/src/node/util.rs
twittner: I disagree. The whole configuration mechanism is designed around defaults (see `ConfigValues`) and loading a configuration will usually not return a `None` but in practice some configuration, even if empty. Catering for some test that could just as well use `ConfigValues::default_values()` would be wrong especially because it is not something we are going to see in practice.

twittner (dismissed):

implementations/rust/ockam/ockam_command/src/node/util.rs
SanjoDeundiak: @twittner  So what would you do if you don't have Authorities? Which is often the case. Empty list? We already had conversation about empty list vs Optional value

BeenzSyed (approved):

implementations/rust/ockam/ockam_command/src/node/util.rs
SanjoDeundiak: I'm resolving that to unblock the merge. You can add more comments after it

PR #3398, opened 2022-09-01
fix(rust): project enroll

twittner:
Currently `project enroll` establishes a secure channel to the project node instead of the project's authority. This commit extends the `ProjectLookup` data to include the authority access route. `project enroll` will attempt to map a `/project` protocol as given in the input multi-address to the corresponding authority address and establish a secure channel to the authority.

**Example**:

```
ockam project enroll \
    --member ... \
    --to /project/myproject/service/authenticator
```

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/project/enroll.rs
+++ b/implementations/rust/ockam/ockam_command/src/project/enroll.rs
@@ -4,7 +4,7 @@ use anyhow::anyhow;
 use ockam::identity::IdentityIdentifier;
 use ockam::{Context, TcpTransport};
 use ockam_api::authenticator::direct::types::AddMember;
-use ockam_api::config::lookup::ConfigLookup;
+use ockam_api::config::lookup::{ConfigLookup, ProjectAuthority};
 use ockam_api::nodes::models::secure_channel::CreateSecureChannelResponse;
 use ockam_core::api::Request;
 use ockam_multiaddr::{proto, MultiAddr, Protocol};
@@ -56,9 +56,8 @@ impl Runner {
     async fn run(self) -> Result<()> {
         let tcp = TcpTransport::create(&self.ctx).await?;
         let map = self.opts.config.get_lookup();
-        let to = if let Some(addr) = authority_addr(&self.cmd.to, &map)? {
-            debug!(%addr, "establishing secure channel to project authority");
-            let mut addr = self.secure_channel(&tcp, &addr).await?;
+        let to = if let Some(a) = project_authority(&self.cmd.to, &map)? {
+            let mut addr = self.secure_channel(&tcp, a).await?;
             for proto in self.cmd.to.iter().skip(1) {
                 addr.push_back_value(&proto).map_err(anyhow::Error::from)?
             }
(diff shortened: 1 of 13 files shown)


implementations/rust/ockam/ockam_command/src/project/enroll.rs
```diff
@@ -32,33 +35,81 @@ pub struct EnrollCommand {
...
+
+    async fn secure_channel(
+        &self,
+        tcp: &TcpTransport,
+        addr: &MultiAddr,
+    ) -> anyhow::Result<MultiAddr> {
+        let mut rpc = RpcBuilder::new(&self.ctx, &self.opts, &self.cmd.node_opts.api_node)
+            .tcp(tcp)
+            .build()?;
+        rpc.request(api::create_secure_channel(addr, None)).await?;
```
SanjoDeundiak: Shouldn't we know authority IdentityIdentifier here?

implementations/rust/ockam/ockam_command/src/project/enroll.rs
```diff
@@ -32,33 +35,81 @@ pub struct EnrollCommand {
...
+        let addr = res.addr()?;
+        Ok(addr)
+    }
+}
+
+/// Get the authority address (if any) of the given address.
+///
+/// If the input address begins with a `/project` protocol we look for the
+/// corresponding authority access route and return it.
+fn authority_addr(input: &MultiAddr, map: &ConfigLookup) -> anyhow::Result<Option<MultiAddr>> {
```
SanjoDeundiak: I'll move that later, since it'll be used to get credential

implementations/rust/ockam/ockam_command/src/project/enroll.rs
```diff
@@ -32,33 +35,81 @@ pub struct EnrollCommand {
...
+    fn new(ctx: Context, opts: CommandGlobalOpts, cmd: EnrollCommand) -> Self {
+        Self { ctx, opts, cmd }
+    }
 
-        let req = Request::post("/members").body(AddMember::new(cmd.member));
-        let mut rpc = RpcBuilder::new(ctx, opts, &cmd.node_opts.api_node)
+    async fn run(self) -> Result<()> {
+        let tcp = TcpTransport::create(&self.ctx).await?;
+        let map = self.opts.config.get_lookup();
+        let to = if let Some(addr) = authority_addr(&self.cmd.to, &map)? {
```
SanjoDeundiak: That is fine for now, but I found it potentially misleading that multiaddr starting with `/project/name` may be resolved to either project node or authority node depending on the situation

SanjoDeundiak (changes requested):

implementations/rust/ockam/ockam_command/src/project/enroll.rs
SanjoDeundiak: cc @mrinalwadhwa

implementations/rust/ockam/ockam_command/src/project/util.rs
```diff
@@ -217,48 +217,62 @@ pub async fn check_project_readiness<'a>(
...
+            .as_ref()
+            .context("Project should have identity set")?;
+        let authority = if let Some(r) = &project.authority_access_route {
+            let rte = MultiAddr::try_from(&**r).context("Invalid project authority address")?;
+            let a = project
+                .authority_identity
+                .as_ref()
+                .context("Missing project authority")?;
+            let a = hex::decode(&**a).context("Invalid project authority")?;
+            let v = Vault::default();
```
SanjoDeundiak, line 243: I don't like that we need to create a Vault here, but let's not bother with that problem for now

SanjoDeundiak (dismissed):

SanjoDeundiak (approved):

PR #3402, opened 2022-09-01
chore(rust): adapt bats test to easy run on different setups

polvorin:
* The first two tests (create node with and without a name)  need to run the command wrapped in bats'  `RUN` to make the assert_success pass 
* Let easily point to  which ockam binary to use
   * Further TODO:  let that be changed through env variable


mrinalwadhwa (dismissed):
Thank you. The changes look great.
Could you address the trailing space space lint and then let's merge.

mrinalwadhwa (approved):

PR #3415, opened 2022-09-03
Remove homebrew from CI

metaclips:
This PR removes brew installation from ockam-builder docker image and  installs bats from source. Now users have to indicate bats-assert and bats-support path to run `bats commands.bat` instead of the script assuming brew to be the source of bats installation.

### DIFF
--- a/implementations/rust/ockam/ockam_command/tests/commands.bats
+++ b/implementations/rust/ockam/ockam_command/tests/commands.bats
@@ -25,13 +25,12 @@
 # ORCHESTRATOR_TESTS=1 LONG_TESTS=1 bats implementations/rust/ockam/ockam_command/tests/commands.bats
 
 # bats_lib=$NVM_DIR/versions/node/v18.8.0/lib/node_modules # linux
-# bats_lib=$(brew --prefix)/lib # macos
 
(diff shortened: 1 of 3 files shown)


implementations/rust/ockam/ockam_command/tests/commands.bats
```diff
@@ -25,10 +25,14 @@
...
 
 # bats_lib=$NVM_DIR/versions/node/v18.8.0/lib/node_modules # linux
-bats_lib=$(brew --prefix)/lib # macos
+# bats_lib=$(brew --prefix)/lib # macos
 
 # Ockam binary to use
 OCKAM=ockam
+if [[ -z $BATS_LIB ]]; then
+  echo "Please set path to bats-support and bat-assert lib"
+  exit 1
```
mrinalwadhwa: instead of exiting here, let's default to `bats_lib=$(brew --prefix)/lib`

metaclips: Done thanks

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3420, opened 2022-09-05
fix(rust): add default value for `nodes` and remove `api_node` config

adrianbenavides:
This will help support config files without `nodes` and/or `api_node` fields when upgrading from older versions.


adrianbenavides:
cc @polvorin

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3423, opened 2022-09-05
feat(rust): switch to arch agnostic integers for secret length

SanjoDeundiak:


implementations/rust/ockam/ockam_core/src/vault/types.rs
```diff
@@ -208,14 +219,14 @@ impl SecretAttributes {
...
-    pub fn length(&self) -> usize {
+    pub fn length(&self) -> u32 {
         self.length
     }
 }
 
 impl SecretAttributes {
     /// Create a new secret attribute.
-    pub fn new(stype: SecretType, persistence: SecretPersistence, length: usize) -> Self {
+    pub fn new(stype: SecretType, persistence: SecretPersistence, length: u32) -> Self {
```
twittner, line 229:
```suggestion
    pub fn new<N: Into<u32>>(stype: SecretType, persistence: SecretPersistence, length: N) -> Self {
```

It may simplify the handling of integers to be more generic here. Most of the length constants are fairly small and could be declared as a `u8` which makes them compatible with all other unsigned integers, e.g. instead of 

```rust
pub const CURVE25519_SECRET_LENGTH_U32: u32 = 32;
pub const CURVE25519_SECRET_LENGTH_USIZE: usize = 32;
```

it would suffice to have only

```rust
pub const CURVE25519_SECRET_LENGTH: u8 = 32;
```

which can easily be converted to larger integers with an `.into()` and it could be passed directly to generic functions such as the suggested `SecretAttributes::new` from above.

twittner (dismissed):

implementations/rust/ockam/ockam_core/src/vault/types.rs
mrinalwadhwa: really like that suggestion, let's merge this PR after that change.

SanjoDeundiak: That makes sense, and I tried something similar, but that lead to many conversions that I wanted to avoid. So I decided to go with separate consts

SanjoDeundiak: I can make change to this function though, but that doesn't mean we can go with only 1 type for consts after that

SanjoDeundiak: Ok, I tried to use that approach, that lead to 3 different types of conversions that appeared all over the codebase. Also, it wasn't possible to convert `match` statements to use conversions, so they needed to be rewritten. At this point I regret spending my and @twittner (who was helping me) time on that. Not worth it atm

mrinalwadhwa (approved):

Issue #3436, opened 2022-09-07
Detect new release is available and display upgrade instructions in command


mrinalwadhwa:
We will continue to release often.

When any ockam command/subcommand is run, we want to check if a new version has been released and display upgrade instruction.

We'll do this by fetching this file over HTTP and checking if it is not empty.

```
https://github.com/build-trust/ockam/releases/download/ockam_v0.72.0/upgrade.txt
```

Where `ockam_v0.72.0` would be replaced with the current version of the command binary.

The code for the check would likely go here before invoking the command
https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/lib.rs#L251

The output message to suggest an upgrade should go to stderr.

We don't want the invoked command to fail incase github is down on unreachable for some reason .. so this functionality should ignore any errors, timeout quickly and move on to running the command as quickly as possible.

---

_We love helping new contributors!_
_If you have questions or need help as you work on your first Ockam contribution, please leave a comment on [this discussion](https://github.com/ockam-network/ockam/discussions/1081)._
_If you're looking for other issues to contribute to, checkout [this discussion](https://github.com/build-trust/ockam/discussions/3206) and labels -  https://github.com/build-trust/ockam/labels/good%20first%20issue or https://github.com/build-trust/ockam/labels/help%20wanted_

mrinalwadhwa:
Added in https://github.com/build-trust/ockam/pull/3439
Thank you @leonzchang 🥳

PR #3439, opened 2022-09-08
feat(rust): warning if user current version is not update to date

leonzchang:
output message if user current version is not update to date when command runs. (#3436)

example:
```
> ockam node list
Current version is not update to date.
TO keep up on the latest version please check at: https://crates.io/crates/ockam

No nodes registered on this system!
```

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/lib.rs
+++ b/implementations/rust/ockam/ockam_command/src/lib.rs
@@ -18,6 +18,7 @@ mod service;
 mod space;
 mod tcp;
 mod terminal;
+mod upgrade;
 mod util;
 mod vault;
 mod version;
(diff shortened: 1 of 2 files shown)


mrinalwadhwa:
Awesome!! Thank you for sending this!

A couple of small suggestions:

1. Change the message to look like this:

```
A new release of ockam is available: 0.72.0 → 0.73.0
To upgrade, run: `brew update && brew upgrade ockam`
https://github.com/build-trust/ockam/releases/tag/ockam_v0.73.0
```

2. Instead of 404 on upgrade.txt as signal of no upgrade ... use 404 or if file empty, both as signal of no upgrade.

3. 
```
To upgrade, run: `brew update && brew upgrade ockam`
```

this line should only show if its a macos binary.

leonzchang:
2. Instead of 404 on upgrade.txt as signal of no upgrade ... use 404 or if file empty, both as signal of no upgrade.
-- if the `upgrade.txt` does not exist(empty), the fetch context will show `Not Found`.

mrinalwadhwa (approved):
Thank you for another awesome contribution!

PR #3464, opened 2022-09-12
Permission

metaclips:


.github/workflows/scorecards.yml
```diff
@@ -1,6 +1,7 @@
 name: OSSF Scorecards
 
-permissions: read-all
+permissions:
+  contents: read
```
mrinalwadhwa, line 4: 🤔 wonder if the scorecards action needs more than `contents: read`

metaclips: Yes it does, the permissions are defined per jobs here https://github.com/build-trust/ockam/blob/ef708342fb9c098f2235c542d3b922f063bbfe8a/.github/workflows/scorecards.yml#L17-L21

mrinalwadhwa: Ah the job already has what it needs 👍 

https://github.com/build-trust/ockam/blob/d09cd0f85658db26e7710521060a52f55df7f3b5/.github/workflows/scorecards.yml#L18-L22

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3475, opened 2022-09-14
test(rust): add inlets&outlets with credentials bats test

SanjoDeundiak:


implementations/rust/ockam/ockam_command/tests/commands.bats
```diff
@@ -32,12 +32,18 @@
...
+  OCKAM=ockam
+fi
 
 if [[ -z $BATS_LIB ]]; then
   BATS_LIB=$(brew --prefix)/lib # macos
 fi
 
+# Where node-specific data would be stored, when nodes don't share identities.
+# /tmp/blue , /tmp/green , etc.
+NODE_PATH=/tmp
```
mrinalwadhwa, line 45: 🤔 bats creates a test run specific tmp dir, maybe we should use that.

polvorin: yes, I wasn't sure about this and initially used a tempdir,  the problem is (when running locally at least)  how to terminate nodes if the test is interrupted / exit without the proper `teardown()` called,  in this case the `green` and `blue` nodes.    The `ockam node delete --all`  at `setup()` time  doesn't "see" nodes created with  non-default `OCKAM_PROJECT_PATH` , that we use to get a different identites.    We must pass the same folder when calling `node delete`,  so having it in a fixed, known place solved that.

mrinalwadhwa: that makes sense.

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3481, opened 2022-09-16
feat(rust): move credentials check decision to `NodeManager` field

SanjoDeundiak:


implementations/rust/ockam/ockam_command/src/util/startup.rs
```diff
@@ -94,6 +96,10 @@ pub fn spawn_node(
         args.push("--skip-defaults".to_string());
     }
 
+    if enable_credential_checks {
+        args.push("--enable-credential-checks".to_string());
+    }
```
adrianbenavides, line 101: We don't have this argument in any command now, do we?

adrianbenavides (approved):

PR #3483, opened 2022-09-16
feat(rust): add `--config` argument to `node create` command

adrianbenavides:
It loads a json file containing a list of commands that will be executed after creating a node. To generate this json file, the user can use the global `--export`, `--export-as`, and `--depends-on` arguments on any command, that will copy the command input to the desired file.

## How to test

```bash
ockam node list --export commands.json
ockam node create blue --config commands.json
```

## Options to export commands to a json file

```bash
# Will generate random name for the command
$ ockam node list --export commands.json

# Set name for command 
$ ockam node list --export commands.json --export-as node-list

# Depend on other command
$ ockam node list --export commands.json --depends-on node-create
```

## Structure of the config file 

```json
{
  "node-list": {
    "args": [
      "node",
      "list"
    ],
    "depends_on": null
  }
}
```

### DIFF
--- a/implementations/rust/ockam/ockam_command/Cargo.toml
+++ b/implementations/rust/ockam/ockam_command/Cargo.toml
@@ -87,3 +87,4 @@ ockam_core = { path = "../ockam_core", version = "^0.70.0" }
 
 [dev-dependencies]
 assert_cmd = "2"
+tempfile = "3"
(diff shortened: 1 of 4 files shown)


adrianbenavides:
@mergifyio rebase

mrinalwadhwa (dismissed):

adrianbenavides:
@mergifyio rebase

mrinalwadhwa (approved):

Issue #3487, opened 2022-09-17
Write a `curl | sh` script to install precompiled ockam binaries


mrinalwadhwa:
We release precompiled binaries in every release. We want to have a curl | sh script to make it easy for people to install these binaries.

https://github.com/build-trust/ockam/releases/tag/ockam_v0.75.0

danieled-it:
Hey @mrinalwadhwa , I'd like to work on this, can I get this issue assigned?

metaclips:
> Hey @mrinalwadhwa , I'd like to work on this, can I get this issue assigned?

Done! Thanks 🚀

danieled-it:
Forgot to ask before: got any preferences on where the new script should be located and what its name should be?

metaclips:
> Forgot to ask before: got any preferences on where the new script should be located and what its name should be?

We can have the script [here](https://github.com/build-trust/ockam/tree/develop/tools/scripts). I think `release-installer.sh` sounds cool 😎but it’s left to you if you have a better name 😎.

danieled-it:
Probably a very dumb question, but what implementation scope would you guys like to see in the commit message for this? 😃 
I got a first "working" copy on my machine but I don't want to get it wrong at commit time 😄

metaclips:
😄 we can scope it under ci. `ci: add ockam release installer script` will be pretty neat 🥶

mrinalwadhwa:
Added in https://github.com/build-trust/ockam/pull/3488
Thank you @danieled-it 🥳 🎈

PR #3488, opened 2022-09-17
ci: add ockam release installer script

danieled-it:
## Current Behavior

No installer is available to `curl | sh` precompiled ockam binaries

## Proposed Changes

This PR is meant to solve #3487 .

The new script in this PR performs the following when executed on Linux environments:

- Checks the architecture of the machine executing it, throws an error and exits if no related binary is released.
- Checks if `cosign` is available for signature verification, fetches the latest version if it's missing.
- Downloads the latest version of the precompiled `ockam` binary based on the architecture.
- Downloads `sha256sums.txt`, the related `ssha256sums.txt.sig` and the ockam public key for validation.
- Performs checksum validation on the downloaded binary
- Informs the user on how to start using ockam and suggests two ways of storing it in `$PATH` for future usages.

**Usage:**

```bash
curl https://raw.githubusercontent.com/danieled-it/ockam/ci/ockam_installer/tools/scripts/release-installer.sh | bash
```
**Missing features:**

- Handle Mac OS environments (I don't have a Mac to validate the results it would return with `uname -m` and `lscpu`, happy to keep working on this if someone can help)

**Execution results:** (excuse the screenshots, I just wanted to make sure the colors would show up 😄 ):

Case 1: `cosign` already available:
image

Case 2: `cosign` is missing:
image

Further testing needed;

- Unfortunately I can only test on a "Windows running Debian on WSL2" environment, so a validation with other architectures to ensure the initial `case` is correct would be nice.

## Checks

### DIFF
--- a/tools/scripts/release-installer.sh
+++ b/tools/scripts/release-installer.sh
@@ -1,0 +1,175 @@
+#!/usr/bin/env bash
+set -e
+
+# This script automatically fetches the precompiled binaries
+# for the latest version released of ockam and stores them.
+
+# Constants definition
+# GitHub repos API URLs
+API_OCKAM_RELEASES="https://api.github.com/repos/build-trust/ockam/releases"
+# Colors
+RED="\033[0;31m"
+GREEN="\033[0;32m"
+ORANGE="\033[0;33m"
+NC="\033[0m"  # No Color
+
+# main
+
+OS_TYPE=$(uname)
+OS_ARCH=$(uname -m)
+
+case $OS_TYPE in
+  "Darwin")
+    # Both Intel and M1/M2 Macs will report "x86_64" with "uname -m"
+    # https://apple.stackexchange.com/questions/420452/running-uname-m-gives-x86-64-on-m1-mac-mini
+    OS_ARCH=$(arch)
+    case $OS_ARCH in
+      "i386")
+        OCKAM_BINARY_FILE="ockam.x86_64-apple-darwin"
+        ;;
+      "arm64")
+        OCKAM_BINARY_FILE="ockam.aarch64-apple-darwin"
+        ;;
+      *)
+        printf "${RED}ERROR${NC} - Unsupported architecture $OS_ARCH discovered, sorry about that!\n"
+        ;;
+    esac
+    ;;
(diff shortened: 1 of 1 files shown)


mrinalwadhwa:
@danieled-it thank you for an awesome first contribution!! 🥳 
I'll run some tests for it in various environments and report back soon.

mrinalwadhwa:
@danieled-it The script looks great. A couple of suggestions:

1. I think we should use cosign, if it's available in the host machine's path, but if it's not available it doesn't make sense to download it via curl and then verify using it.

2. If possible, it would be nice to not limit the script to bash and try to be a bit more portable. For instance, I wasn't able to test the script in an alpine container where the default shell is ash.

On an M1 mac with the latest macos, `uname -m` returns `arm64` and `lscpu` is not found.

danieled-it:
> I think we should use cosign, if it's available in the host machine's path, but if it's not available it doesn't make sense to download it via curl and then verify using it.

This one was easy to change, done in [35731a3](https://github.com/build-trust/ockam/pull/3488/commits/35731a3d5a4c461fea8e24e6a6a067741768131c).
The new execution in case `cosign` is missing looks like this:

image


> If possible, it would be nice to not limit the script to bash and try to be a bit more portable. For instance, I wasn't able to test the script in an alpine container where the default shell is ash.

I'll try to tweak it for some more shell-intercompatibility, I see the problem for now is the `$OSTYPE` variable on top, that's available only in bash, I'll find some replacement for it.

danieled-it:
I think both shell inter-compatibility and MacOS management are ok now, at least from information I found online, of course some real-life testing will be needed for Mac (at least I was able to test the different shells on Linux on my local machine).

As I'm not familiar with ockam itself, I'd ask for an extra-careful check on the whole `case $OS_TYPE in` block 😄

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

mrinalwadhwa:
@danieled-it the script looks great! Thank you for an awesome first contribution!!
I can think of some small improvements but we can tackle those in follow on issues/prs.

PR #3497, opened 2022-09-20
feat(rust): recover remote forwarder

twittner:
Currently any interruption to the underlying TCP connection of a remote forwarder will permanently make it unusable. To remedy the sitution a session "medic" is added to the node manager that maintains sessions which in turn provide actions to replace themselves. Sessions are each pinged at regular frequency and upon three subsequent failures the replacement action is triggered.

Currently only remote forwarders to API nodes, i.e. with addresses like

  /dnsaddr/hostname/tcp/12345/service/api
  /ip4/1.2.3.4/tcp/5678/service/api
  /ip6/::1/tcp/12345/service/api

are considered and only if an authorised identity is present. Other addresses or adresses like these but without authorised identities will not enable recovery and not even attenpt to open a secure channel.

To allow the node manager to re-establish a secure channel the command layer no longer performs the establishment of the secure channel, but instead only project and node references are resolved and the resulting address is passed as is to the node manager. NB that this assumes that the project route does not change which is not always guaranteed. This requires making project resolution available in the manager (tbd).

### DIFF
--- a/implementations/rust/ockam/ockam/src/forwarder.rs
+++ b/implementations/rust/ockam/ockam/src/forwarder.rs
@@ -70,7 +70,6 @@ impl Forwarder {
             forward_route,
             payload: Some(registration_payload.clone()),
         };
-        let _ = ctx.stop_worker(address.clone()).await;
         ctx.start_worker(address, forwarder).await?;
 
         Ok(())
--- a/implementations/rust/ockam/ockam/src/forwarder.rs
+++ b/implementations/rust/ockam/ockam/src/forwarder.rs
@@ -70,6 +70,7 @@ impl Forwarder {
             forward_route,
             payload: Some(registration_payload.clone()),
         };
+        let _ = ctx.stop_worker(address.clone()).await;
         ctx.start_worker(address, forwarder).await?;
 
         Ok(())
--- a/implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
@@ -1,11 +1,16 @@
+use std::collections::HashMap;
+
 use minicbor::{Decode, Encode};
 
 use ockam::remote::RemoteForwarderInfo;
 use ockam_core::CowStr;
+use ockam_identity::IdentityIdentifier;
 use ockam_multiaddr::MultiAddr;
 
 #[cfg(feature = "tag")]
 use ockam_core::TypeTag;
+
+use super::secure_channel::CredentialExchangeMode;
 
 /// Request body when instructing a node to create a forwarder
 #[derive(Debug, Clone, Decode, Encode)]
@@ -14,23 +19,38 @@ use ockam_core::TypeTag;
 pub struct CreateForwarder<'a> {
     #[cfg(feature = "tag")]
     #[n(0)] tag: TypeTag<3386455>,
-    /// Ockam's cloud forwarder node address
-    #[b(1)] pub(crate) address: CowStr<'a>,
-    /// Forwarder alias
-    #[n(2)] pub(crate) alias: Option<CowStr<'a>>,
-    /// Forwarding service is at rust node
+    /// Address to create forwarder at.
+    #[b(1)] pub(crate) address: MultiAddr,
+    /// Forwarder alias.
+    #[b(2)] pub(crate) alias: Option<CowStr<'a>>,
+    /// Forwarding service is at rust node.
     #[n(3)] pub(crate) at_rust_node: bool,
+    /// Authorised identities per project.
+    #[n(4)] pub(crate) identities: HashMap<MultiAddr, IdentityIdentifier>,
+    #[n(5)] pub(crate) mode: CredentialExchangeMode
 }
 
 impl<'a> CreateForwarder<'a> {
-    pub fn new(address: &MultiAddr, alias: Option<String>, at_rust_node: bool) -> Self {
+    pub fn new(address: MultiAddr, alias: Option<String>, at_rust_node: bool) -> Self {
         Self {
             #[cfg(feature = "tag")]
             tag: Default::default(),
-            address: address.to_string().into(),
+            address,
             alias: alias.map(|s| s.into()),
             at_rust_node,
+            identities: HashMap::new(),
+            mode: CredentialExchangeMode::None,
         }
+    }
+
+    pub fn set_identities(&mut self, ids: HashMap<MultiAddr, IdentityIdentifier>) -> &mut Self {
+        self.identities = ids;
+        self
+    }
+
+    pub fn set_credentials_mode(&mut self, m: CredentialExchangeMode) -> &mut Self {
+        self.mode = m;
+        self
     }
 }
 
@@ -105,7 +125,7 @@ mod tests {
             let mut buf = vec![];
             Request::post("/node/forwarder")
                 .body(CreateForwarder::new(
-                    &route_to_multiaddr(&route).unwrap(),
+                    route_to_multiaddr(&route).unwrap(),
                     None,
                     false,
                 ))
--- a/implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
@@ -26,31 +26,30 @@ pub struct CreateForwarder<'a> {
     /// Forwarding service is at rust node.
     #[n(3)] pub(crate) at_rust_node: bool,
     /// Authorised identities per project.
+    ///
+    /// The current handling of this request will only create a secure channel
+    /// for addresses which have a corresponding authorised identity configured.
     #[n(4)] pub(crate) identities: HashMap<MultiAddr, IdentityIdentifier>,
     #[n(5)] pub(crate) mode: CredentialExchangeMode
 }
 
 impl<'a> CreateForwarder<'a> {
-    pub fn new(address: MultiAddr, alias: Option<String>, at_rust_node: bool) -> Self {
+    pub fn new(
+        address: MultiAddr,
+        alias: Option<String>,
+        at_rust_node: bool,
+        identities: HashMap<MultiAddr, IdentityIdentifier>,
+        mode: CredentialExchangeMode,
+    ) -> Self {
         Self {
             #[cfg(feature = "tag")]
             tag: Default::default(),
             address,
             alias: alias.map(|s| s.into()),
             at_rust_node,
-            identities: HashMap::new(),
-            mode: CredentialExchangeMode::None,
+            identities,
+            mode,
         }
-    }
-
-    pub fn set_identities(&mut self, ids: HashMap<MultiAddr, IdentityIdentifier>) -> &mut Self {
-        self.identities = ids;
-        self
-    }
-
-    pub fn set_credentials_mode(&mut self, m: CredentialExchangeMode) -> &mut Self {
-        self.mode = m;
-        self
     }
 }
 
@@ -97,6 +96,7 @@ mod tests {
     use ockam_core::api::{Request, Response, Status};
     use ockam_core::{compat::rand, compat::rand::Rng};
     use ockam_core::{route, Address, Result, Routed, Worker};
+    use std::collections::HashMap;
 
     use crate::nodes::NodeManager;
     use crate::*;
@@ -128,6 +128,8 @@ mod tests {
                     route_to_multiaddr(&route).unwrap(),
                     None,
                     false,
+                    HashMap::new(),
+                    CredentialExchangeMode::None,
                 ))
                 .encode(&mut buf)?;
             buf
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -1,17 +1,24 @@
 //! Node Manager (Node Man, the superhero that we deserve)
 
 use std::collections::BTreeMap;
+use std::error::Error as _;
 use std::path::PathBuf;
-use std::sync::Arc;
 
 use minicbor::Decoder;
 
 use ockam::{Address, Context, ForwardingService, Result, Routed, TcpTransport, Worker};
-use ockam_core::api::{Method, Request, Response, Status};
-use ockam_core::compat::{boxed::Box, string::String};
+use ockam_core::api::{Error, Method, Request, Response, Status};
+use ockam_core::compat::{
+    boxed::Box,
+    string::String,
+    sync::{Arc, Mutex},
+};
 use ockam_core::errcode::{Kind, Origin};
+use ockam_core::AsyncTryClone;
 use ockam_identity::{Identity, IdentityIdentifier, PublicIdentity};
 use ockam_multiaddr::MultiAddr;
+use ockam_node::tokio;
+use ockam_node::tokio::task::JoinHandle;
 use ockam_vault::storage::FileStorage;
 use ockam_vault::Vault;
 
@@ -22,6 +29,7 @@ use crate::lmdb::LmdbStorage;
 use crate::nodes::config::NodeManConfig;
 use crate::nodes::models::base::NodeStatus;
 use crate::nodes::models::transport::{TransportMode, TransportType};
+use crate::session::{Medic, Sessions};
 use crate::DefaultAddress;
 
 pub mod message;
@@ -84,6 +92,7 @@ pub(crate) struct AuthorityInfo {
 
 /// Node manager provides a messaging API to interact with the current node
 pub struct NodeManager {
+    address: Address,
     node_name: String,
     node_dir: PathBuf,
     config: Config<NodeManConfig>,
@@ -99,6 +108,14 @@ pub struct NodeManager {
     authorities: Option<Authorities>,
     pub(crate) authenticated_storage: LmdbStorage,
     pub(crate) registry: Registry,
+    sessions: Arc<Mutex<Sessions>>,
+    medic: JoinHandle<Result<(), ockam_core::Error>>,
+}
+
+impl Drop for NodeManager {
+    fn drop(&mut self) {
+        self.medic.abort()
+    }
 }
 
 pub struct IdentityOverride {
@@ -136,7 +153,8 @@ impl NodeManager {
 impl NodeManager {
     /// Create a new NodeManager with the node name from the ockam CLI
     #[allow(clippy::too_many_arguments)]
-    pub async fn create(
+    pub async fn create<A: Into<Address>>(
+        addr: A,
         ctx: &Context,
         node_name: String,
         node_dir: PathBuf,
@@ -221,7 +239,11 @@ impl NodeManager {
             ));
         }
 
+        let medic = Medic::new();
+        let sessions = medic.sessions();
+
         let mut s = Self {
+            address: addr.into(),
             node_name,
             node_dir,
             config,
@@ -237,6 +259,11 @@ impl NodeManager {
             authorities: None,
             authenticated_storage,
             registry: Default::default(),
+            medic: {
+                let ctx = ctx.async_try_clone().await?;
+                tokio::spawn(medic.start(ctx))
+            },
+            sessions,
         };
 
         if !skip_defaults {
@@ -246,6 +273,9 @@ impl NodeManager {
                 s.configure_authorities(ac).await?;
             }
         }
+
+        s.start_echoer_service_impl(ctx, DefaultAddress::ECHO_SERVICE.into())
+            .await?;
 
         Ok(s)
     }
@@ -285,8 +315,6 @@ impl NodeManager {
             .await?;
         self.start_uppercase_service_impl(ctx, DefaultAddress::UPPERCASE_SERVICE.into())
             .await?;
-        self.start_echoer_service_impl(ctx, DefaultAddress::ECHO_SERVICE.into())
-            .await?;
 
         ForwardingService::create(ctx).await?;
 
@@ -315,7 +343,7 @@ impl NodeManager {
         req: &Request<'_>,
         dec: &mut Decoder<'_>,
     ) -> Result<Vec<u8>> {
-        trace! {
+        debug! {
             target: TARGET,
             id     = %req.id(),
             method = ?req.method(),
@@ -437,7 +465,7 @@ impl NodeManager {
                 .to_vec()?,
 
             // ==*== Forwarder commands ==*==
-            (Post, ["node", "forwarder"]) => self.create_forwarder(ctx, req, dec).await?,
+            (Post, ["node", "forwarder"]) => self.create_forwarder(ctx, req.id(), dec).await?,
 
             // ==*== Inlets & Outlets ==*==
             (Get, ["node", "inlet"]) => self.get_inlets(req).to_vec()?,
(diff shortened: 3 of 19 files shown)


implementations/rust/ockam/ockam/src/forwarder.rs
```diff
@@ -70,6 +70,7 @@ impl Forwarder {
             forward_route,
             payload: Some(registration_payload.clone()),
         };
+        let _ = ctx.stop_worker(address.clone()).await;
```
SanjoDeundiak: Shouldn't we at least add prefix to the provided address, as in elixir? Not doing that looks bad even for demo purposes...
@mrinalwadhwa 

P.S. Oh, is this line the reason why forwarders to rust nodes are created without heartbeats?

implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -244,9 +244,18 @@ impl<'a> Rpc<'a> {
         };
         match hdr.status() {
             Some(status) if hdr.has_body() => {
-                let err = match dec.decode::<String>() {
-                    Ok(msg) => format!("Message: {msg}"),
-                    Err(_) => String::default(),
+                let err = if matches!(dec.datatype(), Ok(Type::String)) {
```
SanjoDeundiak, line 247: That would be great in form of a separate commit

implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
```diff
@@ -1,37 +1,57 @@
...
-    #[n(2)] pub(crate) alias: Option<CowStr<'a>>,
-    /// Forwarding service is at rust node
+    /// Address to create forwarder at.
+    #[b(1)] pub(crate) address: MultiAddr,
+    /// Forwarder alias.
+    #[b(2)] pub(crate) alias: Option<CowStr<'a>>,
+    /// Forwarding service is at rust node.
     #[n(3)] pub(crate) at_rust_node: bool,
+    /// Authorised identities per project.
+    #[n(4)] pub(crate) identities: HashMap<MultiAddr, IdentityIdentifier>,
```
SanjoDeundiak, line 32: Is empty HashMap considered no authorized identities, or any identity is authorized?

implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
```diff
@@ -1,37 +1,57 @@
...
+    #[n(5)] pub(crate) mode: CredentialExchangeMode
 }
 
 impl<'a> CreateForwarder<'a> {
-    pub fn new(address: &MultiAddr, alias: Option<String>, at_rust_node: bool) -> Self {
+    pub fn new(address: MultiAddr, alias: Option<String>, at_rust_node: bool) -> Self {
```
SanjoDeundiak: I would extend constructor to take all fields as arguments, that will force users to make decisions about authorized identities and credentials mode instead of default values (which are likely to cause security vulnerabilities)

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -84,6 +92,7 @@ pub(crate) struct AuthorityInfo {
 
 /// Node manager provides a messaging API to interact with the current node
 pub struct NodeManager {
+    address: Address,
```
SanjoDeundiak: AFAIK, you can take needed address from `Context` instance, no need to store it in the field

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -99,6 +108,14 @@ pub struct NodeManager {
     authorities: Option<Authorities>,
     pub(crate) authenticated_storage: LmdbStorage,
     pub(crate) registry: Registry,
+    sessions: Arc<Mutex<Sessions>>,
+    medic: JoinHandle<Result<(), ockam_core::Error>>,
+}
+
+impl Drop for NodeManager {
+    fn drop(&mut self) {
+        self.medic.abort()
```
SanjoDeundiak: Should it be in shutdown function instead?

implementations/rust/ockam/ockam_command/src/error.rs
```diff
@@ -62,3 +62,9 @@ impl From<std::io::Error> for Error {
         Error::new(exitcode::IOERR, e.into())
     }
 }
+
+impl From<ockam_multiaddr::Error> for Error {
```
SanjoDeundiak, line 66: Could be a separate commit

implementations/rust/ockam/ockam_command/src/forwarder/create.rs
```diff
@@ -46,48 +51,74 @@ impl CreateCommand {
...
-            CredentialExchangeMode::Oneway,
-        )
-        .await?;
-        let at = crate::project::util::clean_projects_multiaddr(at, projects_sc)?;
-        let mut rpc = RpcBuilder::new(ctx, opts, api_node).tcp(&tcp)?.build();
-        let cmd = CreateCommand { at, ..cmd };
-        rpc.request(req(&cmd, at_rust_node)?).await?;
-        rpc.parse_and_print_response::<ForwarderInfo>()?;
-        Ok(())
+async fn rpc(ctx: Context, (opts, cmd): (CommandGlobalOpts, CreateCommand)) -> Result<()> {
```
SanjoDeundiak, line 54: I suspect this code can be used for other commands as well, would be great to move it to some shared place in the future

implementations/rust/ockam/ockam_command/src/forwarder/create.rs
```diff
@@ -46,48 +51,74 @@ impl CreateCommand {
...
-/// Construct a request to create a forwarder
-fn req(cmd: &CreateCommand, at_rust_node: bool) -> anyhow::Result<RequestBuilder<CreateForwarder>> {
-    let alias = if at_rust_node {
-        let mut name = "forward_to_".to_owned();
-        name.push_str(&cmd.forwarder_name);
-        name
-    } else {
-        cmd.forwarder_name.clone()
+    let req = {
+        let alias = if at_rust_node {
```
SanjoDeundiak, line 106: Would be great to unify that, this is out of context of this PR though

implementations/rust/ockam/ockam_command/src/forwarder/create.rs
```diff
@@ -46,48 +51,74 @@ impl CreateCommand {
...
+            format!("forward_to_{}", cmd.forwarder_name)
+        } else {
+            cmd.forwarder_name.clone()
+        };
+        let mut body = CreateForwarder::new(ma, Some(alias), at_rust_node);
+        body.set_identities(pa)
```
SanjoDeundiak: Should we allow empty HashMap here?

implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
SanjoDeundiak: P.S. Looking at the single usage of that constructor I don't see anything against having all fields in the constructor

implementations/rust/ockam/ockam/src/forwarder.rs
twittner: Sorry, this was a leftover from a previous revision. While it does not affect naming it is not relevant for this PR.

implementations/rust/ockam/ockam_api/src/session/sessions.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    pub fn clear_pings(&mut self) {
+        self.pings.clear()
+    }
+}
+
+#[derive(Default, Debug, Copy, Clone, PartialEq, Eq, Hash, Encode, Decode)]
+#[rustfmt::skip]
+pub struct Key {
+    #[n(0)] ctr: u64,
```
SanjoDeundiak: What is the point of the counter?

implementations/rust/ockam/ockam_api/src/session.rs
```diff
@@ -0,0 +1,194 @@
...
+
+impl Decodable for Message {
+    fn decode(m: &[u8]) -> Result<Self, Error> {
+        minicbor::decode(m).map_err(Error::from)
+    }
+}
+
+impl ockam_core::Message for Message {}
+
+/// A collector receives messages from a [`Responder`] and forwards them.
```
SanjoDeundiak: What is Responder here?

implementations/rust/ockam/ockam_api/src/session.rs
```diff
@@ -0,0 +1,194 @@
...
+                        let l = {
+                            let v = Encodable::encode(&m).expect("message can be encoded");
+                            let r = route![session.address().clone(), DefaultAddress::ECHO_SERVICE];
+                            let t = TransportMessage::v1(r, Collector::address(), v);
+                            LocalMessage::new(t, Vec::new())
+                        };
+                        let sender = ctx.clone();
+                        self.pings
+                            .spawn(async move { (key, sender.forward(l).await) });
+                        continue;
```
SanjoDeundiak: `else if` is cleaner IMHO

implementations/rust/ockam/ockam_api/src/nodes/service/forwarder.rs
```diff
@@ -1,44 +1,89 @@
...
+                                enable_recovery(
+                                    &mut s,
+                                    this,
+                                    c,
+                                    req.address,
+                                    alias.to_string(),
+                                    auth,
+                                    req.mode,
+                                );
+                                self.sessions.lock().unwrap().add(s);
```
adrianbenavides, line 71: I've seen that you always unwrap the `lock()` result. Is this done on purpose to let the mutex in a "poisoned" state if something panics?

twittner: It is sort of the default thing to do unless one is sure that no invariants are broken if a panic occurs while holding the lock. While panics are not expected when dealing with sessions they are propagated across thread boundaries this way should they arise.

implementations/rust/ockam/ockam_api/src/session.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    pub async fn start(self, ctx: Context) -> Result<(), Error> {
+        let ctx = ctx.new_detached(Address::random_local()).await?;
+        let (tx, rx) = mpsc::channel(32);
+        ctx.start_worker(Collector::address(), Collector(tx))
+            .await?;
+        self.go(ctx, rx).await
+    }
+
+    async fn go(mut self, ctx: Context, mut rx: mpsc::Receiver<Message>) -> ! {
```
SanjoDeundiak, line 62: This function would benefit from some comments

implementations/rust/ockam/ockam_api/src/session.rs
```diff
@@ -0,0 +1,194 @@
...
+            }
+
+            let _ = timeout(self.delay, self.get_results(&mut rx)).await;
+        }
+    }
+
+    async fn get_results(&mut self, rx: &mut mpsc::Receiver<Message>) {
+        loop {
+            tokio::select! {
+                p = self.pings.join_next(), if !self.pings.is_empty() => match p {
```
SanjoDeundiak, line 106: Why `if !self.pings.is_empty()` check is needed?

implementations/rust/ockam/ockam_api/src/session.rs
```diff
@@ -0,0 +1,194 @@
...
+                }
+            }
+
+            let _ = timeout(self.delay, self.get_results(&mut rx)).await;
+        }
+    }
+
+    async fn get_results(&mut self, rx: &mut mpsc::Receiver<Message>) {
+        loop {
+            tokio::select! {
```
SanjoDeundiak, line 105: Could you please elaborate on how this works? Is that intended that we try to run pings and session reestablishments and receive pongs concurrently? Shouldn't we do that in parallel? Also, `tokio::select` is meant to cancel all futures except the one that finished. So, in a loop, we start polling all of the futures I mentioned above one-by-one (allowing them to make some progress), wait till one of them finishes, abort everything else (and remove all the progress that they may achieved, also not cleaning up any state they may have changed so far), then start all of them again on next loop iteration. Am I wrong about how this works?

implementations/rust/ockam/ockam_api/src/session.rs
```diff
@@ -0,0 +1,194 @@
...
+                        }
+                    }
+                }
+            }
+
+            let _ = timeout(self.delay, self.get_results(&mut rx)).await;
+        }
+    }
+
+    async fn get_results(&mut self, rx: &mut mpsc::Receiver<Message>) {
```
SanjoDeundiak, line 103: Require some refactoring in the future, it's a little bit hard to read

SanjoDeundiak (changes requested):
Still reviewing, submitting what I have so far...

implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
twittner: The way it is currently used by the node manager is that without an authorised identity, no secure channel will be established.

implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
twittner: These are still optional as creating a forwarder at another rust node for example does not require a secure channel.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
twittner: I will add a shutdown method and stop the task there.

implementations/rust/ockam/ockam_api/src/session.rs
twittner: Operationally there are ping tasks and replacement tasks that execute in the context of tokio `JoinSet`s. Their progress is driven by the tokio executor. The `tokio::select!` call only evaluates cancellation-safe futures which eventually yield a result from those tasks (or a message in case of `mpsc::Receiver::recv`). No progress is removed and no state needs to be cleaned up when the loosing futures care cancelled.

For details see:
- https://docs.rs/tokio/latest/tokio/macro.select.html
- https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html

implementations/rust/ockam/ockam_api/src/session.rs
twittner: `JoinSet::join_next` immediately returns `None` if empty. By checking first if the set is empty, the future returned from `JoinSet::join_next` is not evaluated, leaving the other branches. This makes the whole loop less busy.

implementations/rust/ockam/ockam_api/src/session.rs
twittner: Another leftover from before https://github.com/build-trust/ockam/pull/3485.

implementations/rust/ockam/ockam_api/src/session/sessions.rs
twittner: The key is used to index sessions. The counter gives a strongly monotonic number and avoids collisions which a purely random key could cause.

implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
SanjoDeundiak: Could you please add that comment to the code?

implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
SanjoDeundiak: Not using Secure Channel or authorized identifiers explicitly may be fine for the time being, I'm worried about the fact that this happens implicitly. User of this code can easily create the structure, use insecure flow and not even know that those fields exist

implementations/rust/ockam/ockam_api/src/session.rs
SanjoDeundiak: @twittner oh, I guess `JoinSet` is the owner of the futures in this case, so futures are not cancelled in `tokio::select`. Tnx, for the clarification. Do you know if futures are able to progress in the background, or only while `join_next` is called? That wasn't clear for me after reading the doc

SanjoDeundiak: Also, what are you thoughts about having centralized Medic that tracks all sessions (as in this PR) vs having dedicated Medic per session? Is there a chance latter approach would look simpler?

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -237,6 +250,11 @@ impl NodeManager {
             authorities: None,
             authenticated_storage,
             registry: Default::default(),
+            medic: {
+                let ctx = ctx.async_try_clone().await?;
+                tokio::spawn(medic.start(ctx))
```
SanjoDeundiak, line 255: We ideally want to limit direct usage of tokio to `ockam_node` crate. It's totally fine for now though, just FYI

implementations/rust/ockam/ockam_api/src/nodes/service/forwarder.rs
```diff
@@ -49,14 +94,135 @@ impl NodeManager {
...
+
+async fn replace_sec_chan(
+    ctx: &Context,
+    manager: &Address,
+    prev: Address,
+    addr: &MultiAddr,
+    authorised: &HashMap<MultiAddr, IdentityIdentifier>,
+    mode: CredentialExchangeMode,
+) -> Result<Address> {
+    debug!(%addr, %prev, "recreating secure channel");
```
SanjoDeundiak, line 197: Is sending a message to itself is better than simple function call?

implementations/rust/ockam/ockam_api/src/nodes/service/forwarder.rs
```diff
@@ -49,14 +94,135 @@ impl NodeManager {
...
+    if res.status() != Some(Status::Ok) {
+        if res.has_body() {
+            let e: Error = d.decode()?;
+            warn!(%addr, %prev, err = ?e.message(), "failed to create secure channel");
+        }
+        return Err(ApiError::generic("error creating secure channel"));
+    }
+    let res: CreateSecureChannelResponse = d.decode()?;
+    let mad = res.addr()?;
+    if let Some(p) = mad.first() {
```
SanjoDeundiak: Isn't in cleaner to convert this multiaddr to a route (we have a function for that) and check that there is only one hop?

implementations/rust/ockam/ockam_api/src/nodes/service/forwarder.rs
```diff
@@ -49,14 +94,135 @@ impl NodeManager {
...
+            if &*p == "api" {
+                return MultiAddr::default().try_with(input.iter().take(3)).ok();
+            }
+        }
+    }
+    None
+}
+
+/// Create a secure channel to an API node or return the address route as is.
+async fn connect(
```
SanjoDeundiak: Should it be a NodeManager's function?

implementations/rust/ockam/ockam_api/src/session/sessions.rs
SanjoDeundiak: Isn't that potentially leaking data about number of sessions of this node to the other node?

implementations/rust/ockam/ockam_api/src/session.rs
twittner: > Do you know if futures are able to progress in the background, or only while `join_next` is called?

A `JoinSet` spawns the futures as regular tokio tasks and they make progress independently of the `join_next` call.

> Also, what are you thoughts about having centralised Medic that tracks all sessions (as in this PR) vs having dedicated Medic per session?

I think tracking all sessions here is more flexible, for instance when expressing dependencies. Isolated session tracking in actors makes such reasoning comparatively difficult. For example, when a secure channel is shared by multiple sessions, it would be enough to ping it once and restart all sessions depending on it. This is easier to implement when all sessions are treated as data in a single `Medic`.

implementations/rust/ockam/ockam_api/src/nodes/service/forwarder.rs
twittner: Most methods of `NodeManager` require unique access (`&mut self`). The recovery futures run asynchronously and we would have to wrap the manager into an `Arc<Mutex<_>>` to apply those methods. Such a change would be much larger so I chose the message sending instead.

implementations/rust/ockam/ockam_api/src/nodes/service/forwarder.rs
twittner: I deliberately made it a private function of this module for the time being so that other parts of the `NodeManager` would not be tempted to use it.

implementations/rust/ockam/ockam_api/src/session/sessions.rs
twittner: Not potentially, but it definitely exposes the number of sessions. If that is a concern we would have to resort to just random keys and perform a collision check when generating it. If preferred I can implement that change.

implementations/rust/ockam/ockam_api/src/nodes/service/forwarder.rs
SanjoDeundiak: But it can be private even if it's a NodeManager's function

implementations/rust/ockam/ockam_api/src/nodes/service/forwarder.rs
SanjoDeundiak: Yeah, probably there should be `NodeManager` that is doing all the heavy lifting, and `NodeManagerWorker` that is a Worker that delegates calls to `NodeManager`. In this case either `NodeManager` can be wrapped inside `Arc<Mutex>` of its fields could be turned into thread-safe structures. Definitely out of the context of this PR though. Thoughts?

implementations/rust/ockam/ockam_api/src/nodes/service/forwarder.rs
twittner: Not sure I get what you mean. If it is a private *method* of the `NodeManager` it would be available in the whole scope of it vs. just being a private function of this module. Did you have something else in mind?

implementations/rust/ockam/ockam_api/src/nodes/service/forwarder.rs
twittner: That would be a good change I think.

implementations/rust/ockam/ockam_api/src/nodes/service/forwarder.rs
SanjoDeundiak: AFAIK private function of the struct, that is defined in submodule, is only available from that submodule. In other words this `connect` function would be available from this file only doesn't matter if it's a static function of a struct's function

implementations/rust/ockam/ockam_api/src/session/sessions.rs
SanjoDeundiak: Yeah, I think leaking such a counter is a bad idea. How about random keys that are long enough so you don't worry about collisions?

SanjoDeundiak: BTW we can have `ctr` and `rnd` if you like that, it's just usage should be vice versa: `rnd` is unique identifier of the session, while `ctr` is an intra-session ping counter

implementations/rust/ockam/ockam_api/src/nodes/service/forwarder.rs
twittner: Thanks, I did not think it would behave like this but it makes sense.

implementations/rust/ockam/ockam_api/src/session/sessions.rs
twittner: I switched to a 24-bytes random key.

twittner:
@SanjoDeundiak: Anything left that needs to be addressed?

SanjoDeundiak:
@mergifyio rebase

SanjoDeundiak (approved):

PR #3503, opened 2022-09-21
feat(rust): add `secure` protocol to multi-addr

twittner:
As an empty protocol `secure` can denote a secure channel in a multi-addr, e.g. `/ip4/127.0.0.1/tcp/8000/secure`

### DIFF
--- a/implementations/rust/ockam/ockam_multiaddr/tests/id.rs
+++ b/implementations/rust/ockam/ockam_multiaddr/tests/id.rs
@@ -108,7 +108,7 @@ quickcheck! {
                         prot.push_back(Ip6::CODE)
                     }
                     Secure::CODE => {
-                        addr.push_back(Secure::new()).unwrap();
+                        addr.push_back(Secure::new("secure")).unwrap();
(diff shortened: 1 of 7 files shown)


mrinalwadhwa:
Nice, I think this should have an address after because we still want to indicate which secure channel listener to establish the channel with. 

```
/ip4/127.0.0.1/tcp/8000/secure/api
```

implementations/rust/ockam/ockam_multiaddr/tests/id.rs
```diff
@@ -110,7 +138,17 @@ quickcheck! {
     }
 }
 
-const PROTOS: &[Code] = &[Tcp::CODE, DnsAddr::CODE, Ip4::CODE, Ip6::CODE];
+const PROTOS: &[Code] = &[
```
SanjoDeundiak, line 141: Could be a static function like `Code::all()`, if you like it

SanjoDeundiak (dismissed):

SanjoDeundiak (approved):

PR #3519, opened 2022-09-24
feat(rust): extend the declarative config support

adrianbenavides:
## Scope

Being able to execute a set of commands after a node is created.

This PR contains the work described in `Implement commands parser` of the [design issue](https://github.com/build-trust/codebase/issues/400), specifically:
- Support piping
- After a node is created, wait for user input before running next command
- Section to define arguments to be used when creating the node
- Sections to define commands to be run after node creation/restart

## How to test

Create a json file with the following contents:

```json
{
  "run": {
    "name": "green"
  },
  "on_node_init": [
    {
      "command": "secure-channel create --from /node/green --to /node/blue/service/api",
      "pipe": true
    },
    "tcp-inlet create --at /node/green --from 127.0.0.1:7000 --to -/service/echo"
  ]
}
```

Execute the command `ockam node run <path-to-file>.json`.

This command will create the `green` node without the default arguments and then it will execute the commands defined in the `on_node_init` section. Note that the `secure-channel` command needs a node called `blue`. If it doesn't exist, it will be created automatically.

## Next up

- Use this feature on the `ockam node start` command

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
@@ -92,6 +92,7 @@ impl NodeManagerWorker {
         let alias = alias.map(|a| a.0.into()).unwrap_or_else(random_alias);
 
         info!("Handling request to create inlet portal");
+        debug!(%bind_addr, %alias, ?outlet_route, "Creating inlet portal");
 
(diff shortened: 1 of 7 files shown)


SanjoDeundiak (dismissed):

SanjoDeundiak:
@mergifyio rebase

SanjoDeundiak (approved):

PR #3528, opened 2022-09-26
feat(rust): resolve forwarder project name in manager

twittner:
Since project routes are not guaranteed to be stable this commit moves the resolution of project references to routes when creating a remote forwarder to the node manager. The command-line layer merely maps a project name to a project ID but leaves everything else to the node manager.

On top of https://github.com/build-trust/ockam/pull/3497.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
@@ -1,5 +1,3 @@
-use std::collections::HashMap;
-
 use minicbor::{Decode, Encode};
 
 use ockam::remote::RemoteForwarderInfo;
@@ -10,8 +8,6 @@ use ockam_multiaddr::MultiAddr;
 #[cfg(feature = "tag")]
 use ockam_core::TypeTag;
 
-use super::secure_channel::CredentialExchangeMode;
-
 /// Request body when instructing a node to create a forwarder
 #[derive(Debug, Clone, Decode, Encode)]
 #[rustfmt::skip]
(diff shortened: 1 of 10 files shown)


implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
```diff
@@ -20,35 +16,42 @@ pub struct CreateForwarder<'a> {
...
-    pub fn new(
+    pub fn at_project(address: MultiAddr, alias: Option<String>, cloud: MultiAddr) -> Self {
+        Self {
+            #[cfg(feature = "tag")]
+            tag: Default::default(),
+            address,
+            alias: alias.map(|s| s.into()),
+            at_rust_node: false,
+            cloud_addr: Some(cloud),
+            authorized: None,
```
SanjoDeundiak, line 42: Why `at_project` forwarder doesn't have authorized value?

implementations/rust/ockam/ockam_command/src/forwarder/create.rs
```diff
@@ -108,13 +90,11 @@ async fn rpc(ctx: Context, (opts, cmd): (CommandGlobalOpts, CreateCommand)) -> R
...
         };
-        let body = CreateForwarder::new(
-            ma,
-            Some(alias),
-            at_rust_node,
-            pa,
-            CredentialExchangeMode::Oneway,
-        );
+        let body = if Some(Project::CODE) == cmd.at.first().map(|p| p.code()) {
+            CreateForwarder::at_project(ma, Some(alias), cmd.cloud_opts.route())
```
SanjoDeundiak, line 97: Should we error if `cmd.authorized` was provided as well?

implementations/rust/ockam/ockam_command/src/forwarder/create.rs
```diff
@@ -108,13 +90,11 @@ async fn rpc(ctx: Context, (opts, cmd): (CommandGlobalOpts, CreateCommand)) -> R
...
+        } else {
+            CreateForwarder::at_node(ma, Some(alias), at_rust_node, cmd.authorized)
```
SanjoDeundiak, line 99: Should we error if `cmd.cloud_opts` was provided as well?

implementations/rust/ockam/ockam_api/src/util.rs
```diff
@@ -127,29 +132,47 @@ pub fn multiaddr_to_addr(ma: &MultiAddr) -> Option<Address> {
...
+            return Err(ApiError::message(format!(
+                "unknown transport type: {other}"
+            )));
+        }
     }
-    Some(ma)
+    Ok(ma)
+}
+
+pub fn is_secure_channel_addr(a: &MultiAddr) -> bool {
```
SanjoDeundiak: Does it check if secure channel will be created while using that MultiAddr, and doesn't detect if there is already existent secure channel in that MultiAddr? Would be nice to explain that in the comments (and maybe even in the name)

SanjoDeundiak: Also wanted to clarify, is it possible that we want to check if next hop uses secure channel, e.g.
`/ipv4/127.0.0.1/tcp/1234/secure/api/ipv4/127.0.0.1/tcp4321/secure/api`, but for example this multiaddr also passes the check I guess? `/ipv4/127.0.0.1/tcp/1234/ipv4/127.0.0.1/tcp4321/secure/api`
I'm afraid of misusing that function

SanjoDeundiak (changes requested):

implementations/rust/ockam/ockam_api/src/nodes/models/forwarder.rs
twittner: It uses the project authority.

implementations/rust/ockam/ockam_command/src/forwarder/create.rs
twittner: Done in https://github.com/build-trust/ockam/pull/3528/commits/08d78aed923f83858d6996fcc98fa4a09f71a9a8

implementations/rust/ockam/ockam_command/src/forwarder/create.rs
twittner: AFAIK they read from an environment variable and default to a standard address if not present. `CloudOpts` has no arguments to provide.

implementations/rust/ockam/ockam_api/src/util.rs
twittner: To avoid confusion I have removed it in https://github.com/build-trust/ockam/pull/3528/commits/08d78aed923f83858d6996fcc98fa4a09f71a9a8. Reading the match inline seems to be more obvious to readers.

twittner:
@SanjoDeundiak: Could you review again please?

SanjoDeundiak (approved):

PR #3545, opened 2022-09-30
feat(rust): recover tcp inlet 

twittner:
Continuation of https://github.com/build-trust/ockam/pull/3497.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
@@ -1,11 +1,15 @@
 //! Inlets and outlet request/response types
 
+use std::net::SocketAddr;
+
 use minicbor::{Decode, Encode};
 use ockam_core::compat::borrow::Cow;
 
 use ockam_core::CowStr;
 #[cfg(feature = "tag")]
 use ockam_core::TypeTag;
+use ockam_identity::IdentityIdentifier;
+use ockam_multiaddr::MultiAddr;
 
 /// Request body to create an inlet or outlet
 #[derive(Clone, Debug, Decode, Encode)]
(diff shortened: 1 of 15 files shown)


implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
```diff
@@ -85,31 +93,74 @@ impl NodeManagerWorker {
...
+            if !sec1.is_empty() && rest.matches(0, &[Service::CODE.into(), Secure::CODE.into()]) {
+                let addr = sec1.clone().try_with(rest.iter().take(2))?;
+                let (sec2, _) = node_manager
+                    .connect(&addr, CredentialExchangeMode::Mutual, None)
+                    .await?;
+                (sec1, sec2.try_with(rest.iter().skip(2))?)
+            } else {
+                (MultiAddr::default(), sec1.try_with(&rest)?)
+            }
+        };
```
adrianbenavides: I'm having a hard time understanding the context of various parts of the code. For example, in this block, it would be very helpful to have a comment explaining what's going to happen, why it's needed, etc., so anyone can understand a chunk of code without having to read it.

twittner: Thanks @adrianbenavides! I added a bit of context and simplified this block. The complication results mostly from the various ways how to connect, e.g. via project and forwarder or directly to another node. The inlet must understand those in order to recreate the route from scratch if the connection is interrupted at some point. Let me know if there is anything else unclear.

adrianbenavides: Great, thanks toralf! I'll finish the review as soon as I can :+1:

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
```diff
@@ -205,3 +280,78 @@ impl NodeManagerWorker {
...
+                    }
+                };
+                let r = multiaddr_to_route(&rest)
+                    .ok_or_else(|| ApiError::message(format!("invalid multiaddr: {rest}")))?;
+                if let Some(wa) = data.get::<Address>(INLET_WORKER) {
+                    let _ = this.tcp_transport.stop_inlet(wa).await;
+                }
+                let opts = InletOptions::new(bind, r, access);
+                let wa = this.tcp_transport.create_inlet_extended(opts).await?.0;
+                data.put(INLET_WORKER, wa);
```
adrianbenavides: IIUC, in this block you stop the `INLET_WORKER` if it's set in `data` before adding the new worker address (`wa`). Is that right? My main concern is that I wouldn't feel confident debugging/refactoring/modifying this code without your help.

I know it's not a complex piece of code, but it's hard to contextualize it within all that's going on in the whole function. My problem in this case is that I have to read the entire function to understand what's going on in this specific block and figure out how it depends on the previous code. In my experience, it helps a lot to add a simple visual separation with a small comment so people that don't have the context of this code can quickly navigate through it. 

For example:

```rust
let rest = ...;

// Replace INLET_WORKER with a new worker address
let r = ...
...
data.put(...);
```

This way I could easily identify what the next lines of code are going to do without having to read the code first. Let me know what you think, just trying to make things easier for our future selves!

twittner: Thanks @adrianbenavides! I added more comments to provide context. Let me know if it helps.

adrianbenavides: It looks great now @twittner, thank you so much for adding those comments :pray:

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
```diff
@@ -14,34 +18,75 @@ use ockam_core::TypeTag;
...
+
+    pub fn authorized(&self) -> Option<IdentityIdentifier> {
+        self.authorized.clone()
+    }
+
+    pub fn alias(&self) -> Option<&str> {
+        self.alias.as_deref()
+    }
+
+    pub fn is_check_credential(&self) -> bool {
```
SanjoDeundiak, line 87: Is that correct grammar?

implementations/rust/ockam/ockam_command/tests/commands.bats
```diff
@@ -316,7 +316,21 @@ teardown() {
   assert_success
 }
 
-@test "inlet/outlet example with credentials, not provided" {
+@test "create an inlet/outlet pair with relay through a forwarder in an orchestrator project and move tcp traffic through it 2/2" {
```
SanjoDeundiak: Could you please replace these `1/2` `2/2` with something more meaningful? It required me to go through the whole test to figure out what's the difference

implementations/rust/ockam/ockam_command/tests/commands.bats
```diff
@@ -382,6 +429,37 @@ teardown() {
...
+  run $OCKAM project enroll --member $green_identifer --to /project/default/service/authenticator
+  assert_success
+
+  run $OCKAM tcp-outlet create --at /node/blue --from /service/outlet --to 127.0.0.1:5000 --check-credential
+  assert_success
+  run  $OCKAM forwarder create blue --at /project/default --to /node/blue
+  assert_output --partial "forward_to_blue"
+  assert_success
+
+  run $OCKAM tcp-inlet create --at /node/green --from 127.0.0.1:7000 --to /project/default/service/forward_to_blue/secure/api/service/outlet --check-credential
```
SanjoDeundiak, line 456: Would be great to test that whole session recovery is working. Unit/integration/bats tests would be great. Not sure how hard it would be to do. Not in this PR for sure

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
twittner: Years ago I was reading a rust style guide somewhere that recommended to use an `is_` prefix for predicates. It does not seem to be the case anymore, at least it is not mentioned in the API naming conventions. That being said, it still has its merits as `check_credential` sounds a bit imperative and the `is_` prefix reads like an inquiry, which it sort of is. Do you have a better name in mind?

implementations/rust/ockam/ockam_command/tests/commands.bats
twittner: :heavy_check_mark:

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
adrianbenavides: I agree with @twittner . That function without the `is_` prefix could lead to confusion, as one might think the function will actually do something when it's just reading a bool value.

adrianbenavides (dismissed):
LGTM!

adrianbenavides (dismissed):

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3580, opened 2022-10-04
feat(elixir): implement elixir ABAC policy checks

hairyhum:
Policies contain attribute rules
Attribute rules match against ABAC.Request
ABAC.Request is associated to policy by ActionId (action name + resource name) ABAC.Request contains action resource and subject attributes

ABAC.Authorization implements message flow authorization using ABAC:

Subject attributes are taken from the credential attributes storage Resource attributes are taken from the worker state (e.g. "address") Action attributes are taken from the Message (e.g. "to", "from")

Workers can have :attributes configured on start to go to resource attributes Message local metadata goes to action attributes (except :identity and :identity_id)

## Checks


hairyhum:
API docs for policies:
```
List policies:
Request: GET path: "" body: ""
Response: map(action_id: string => attribute_rule: string)

Get policy:
Request: GET path: action_id body: ""
Response: attribute_rule : string
Example: GET "echo/handle_message"
Response: "(= subject.foo \"bar\")"

Set policy:
Request: PUT path: action_id body: attribute_rule (binary utf8)
Example:
PUT "echo/handle_message" "(= subject.foo \"bar\")"

Delete policy
Request: DELETE path: action_id
```

implementations/elixir/ockam/ockam_abac/lib/attribute_rules.ex
```diff
@@ -0,0 +1,239 @@
...
+  defp do_match_rule?(false, _request) do
+    false
+  end
+
+  defp do_match_rule?(_rule, _request) do
+    false
+  end
+
+  defp compare(:eq, val1, val2), do: val1 == val2
+  defp compare(:neq, val1, val2), do: val1 == val2
```
polvorin: typo on `==`

implementations/elixir/ockam/ockam_abac/lib/attribute_rules.ex
```diff
@@ -0,0 +1,239 @@
...
+  alias Ockam.ABAC.AttributeRule.Formatter
+
+  alias Ockam.ABAC.Request
+
+  ## Only binaries for now
+  ## TODO: support more types for lt/gt
+  @type attribute_source() :: :resource | :action | :subject
+  @type key() :: {source :: attribute_source(), name :: binary()}
+  @type value() :: binary()
+
```
polvorin, line 13: wonder if a an expression language with terminals, literals, etc won't be easier to reason about long term. And introduce typed values.    Made an attempt [here](https://gist.github.com/polvorin/9f49fd92c1462f433e5ed2e6d047882c)

hairyhum: We could use that, but I was trying to actually make it more strict and specific to the rules that we have rather than introducing generic language.
Because if we make it more generic, then we'll have to enforce specific rules again when we validate them. For example make sure that the top rule is boolean and that something like `:eq` only has 2 arguments.

implementations/elixir/ockam/ockam_abac/lib/attribute_rules.ex
hairyhum: Updated

implementations/elixir/ockam/ockam_abac/lib/attribute_rules.ex
hairyhum: Also I've removed all usages of `AttributeRule.new` from the tests. Let's keep the internal implementation opaque for the user and only suggest to use the s-expression format.

metaclips-botter:
☑️ Orchestrator test on PR #3580 at commit `8a9930445420e1e00e868db8fed757934cbe09c2` passed

polvorin (dismissed):

documentation/architecture/authorization/ABAC.md
```diff
@@ -0,0 +1,194 @@
...
+Mostly useful in logical rules, but can be used as allow-all or deny-all policies
+
+##### Comparison rules:
+
+- equality `(= subject.foo "bar")` `(!= subject.foo "bar")`
+- numeric comparison `(> subject.foo 1)`, `(< subject.foo 3)`
+
+Supports operators `=` `!=` `>` and `<`
+First argument MUST be an Attribute
+Second argument can be an Attribute or a Value
```
twittner, line 72: Why constrain the order and limit the number of arguments to two? Equality for instance is a symmetric relation so the order of arguments does not matter.

documentation/architecture/authorization/ABAC.md
```diff
@@ -0,0 +1,194 @@
...
+
+##### Lists
+
+- `[1 2 3]`
+- `["foo" "bar"]`
+
+List consists of multiple values.
+
+**Attributes, values and lists can be arguments in rules, but cannot be rules,
+because they don't resolve to boolean**
```
twittner, line 128: Except for the boolean values.

documentation/architecture/authorization/ABAC.md
```diff
@@ -0,0 +1,194 @@
...
+
+Matches that an attribute belongs to a list.
+
+- `(member? subject.key [1 2 3])`
+- `(member? subject.key resource.keys)`
+- `(member? "foo" action.keys)`
+
+First argument can be an Attribute or a Value
+Second argument can be an Attribute or a List
+If second argument is an Attribute, the corresponding Attribute should resolve to a List
```
twittner:
```suggestion
If second argument is an Attribute it MUST resolve to a List
```

documentation/architecture/authorization/ABAC.md
```diff
@@ -0,0 +1,194 @@
...
+
+##### Logical rules:
+
+Combine other rules
+
+- `(not (in "foo" action.keys))`
```
mszpakowski: Is there a difference between `in` and `member?` used above?

hairyhum: We decided to not use `in`, will update

documentation/architecture/authorization/ABAC.md
hairyhum: To make rules more strict.

documentation/architecture/authorization/ABAC.md
hairyhum: `true` and `false` are rules, attributes with boolean values should not be rules because we can't guarantee attribute types.

implementations/elixir/ockam/ockam_abac/lib/policy_storage/dets.ex
```diff
@@ -0,0 +1,114 @@
...
+
+  Runs a gen_server opening the DETS table.
+
+  Options:
+  - filename: atom | string - file name to use for DETS table
+  """
+
+  @behaviour Ockam.ABAC.PolicyStorage
+
+  use GenServer
```
mszpakowski, line 13: Just curious why are we spinning up `GenServer`s here and with ETS storage? Doesn't look like they're used for anything else than opening a file/creating ETS table. Is there a plan to add more functionality in callbacks later to those?

hairyhum: All it does here is opens the table, yes. I don't think we need much to add much more.

hairyhum: If you have any ideas on better ways to manage ets and dets tables without bringing in something like mnesia - please share.

mszpakowski: Ah, nevermind, it makes sense to me now. I thought we could simply initialize those tables when we start the application instead of spinning up GenServers, but it would be less configurable and more difficult to manage.

implementations/elixir/ockam/ockam_services/test/services/abac_policies_api_test.exs
```diff
@@ -0,0 +1,64 @@
...
+  alias Ockam.ABAC.ActionId
+  alias Ockam.ABAC.PolicyCheck
+  alias Ockam.ABAC.PolicyStorage
+  alias Ockam.ABAC.Request
+
+  alias Ockam.API.Client, as: ApiClient
+  alias Ockam.Services.API.ABAC.PoliciesApi
+
+  setup do
+    Application.put_env(:ockam_abac, :policy_storage, Ockam.ABAC.PolicyStorage.ETS)
```
mszpakowski, line 14: If we're changing the application config in the test, then I think we should restore it to the previous state in the `on_exit/1` callback, otherwise it might affect other tests.

polvorin (dismissed):

hairyhum:
@Mergifyio refresh

hairyhum:
@Mergifyio queue

hairyhum:
@Mergifyio queue default

polvorin (approved):

Issue #3572, opened 2022-10-04
Lint crago.lock file when updated


metaclips:
When a PR updates `Cargo.lock` file, we should ensure that the lock file was generated by `cargo command` and not tampered with. This could entail a CI that
- Gets the current SHA of `Cargo.lock` generated by PR/user.
- Deleting the current `Cargo.lock`
- Generating a new `Cargo.lock`, possibly calling `cargo check` command.
- Getting the SHA of the new lock file
- Ensure the SHA generated by `user PR` is same as that of `CI`

elliotwutingfeng:
Can I try this?

mrinalwadhwa:
@elliotwutingfeng that would be great! let us know if you have questions along the way.

elliotwutingfeng:
Thanks! I believe this can be closed now.

PR #3583, opened 2022-10-05
ci: lint cargo.lock in CI

elliotwutingfeng:
## Current Behavior

When a PR updates `Cargo.lock` file, we should ensure that the lock file was generated by `cargo command` and not tampered with. 

## Proposed Changes

Add the following jobs to CI `rust.yml` workflow:

   - Gets the current SHA of Cargo.lock generated by PR/user.
   - Deleting the current Cargo.lock
   - Generating a new Cargo.lock, possibly calling cargo check command.
   - Getting the SHA of the new lock file
   - Ensure the SHA generated by user PR is same as that of CI

Addresses #3572

## Checks

### DIFF
--- a/.github/workflows/lint-cargo.yml
+++ b/.github/workflows/lint-cargo.yml
@@ -1,0 +1,47 @@
+name: Lint Cargo.lock
+
+permissions:
+  contents: read
+
+on:
+  workflow_dispatch:
+    inputs:
+      commit_sha:
+        description: SHA to test workflow
+  pull_request:
+    paths:
+      - "**.toml"
+      - "**/Cargo.lock"
+  push:
+    paths:
+      - "**.toml"
+      - "**/Cargo.lock"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  lint_cargo_lock:
+    name: Rust - Lint Cargo.lock
+    runs-on: ubuntu-20.04
+    container:
+      image: ghcr.io/build-trust/ockam-builder@sha256:e43dd94652096b03cc472a3c709c7335e8b166cab77b7a7b56f88fa38f3d24cc
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+        with:
+          ref: ${{ github.event.inputs.commit_sha }}
+      - uses: ./.github/actions/cargo_home_cache
+      - uses: ./.github/actions/cargo_target_dir_cache
+      - name: Check whether Cargo.lock is same as CI-generated Cargo.lock
+        run: |
+          old_hash=`sha384sum Cargo.lock`
+          rm Cargo.lock
+          RUSTFLAGS='--cfg tokio_unstable -Dwarnings' cargo check
+          expected_hash=`sha384sum Cargo.lock`
+          if [ "$old_hash" = "$expected_hash" ]; then
+              echo "Cargo.lock hash matches CI-generated Cargo.lock hash"
+          else
+              echo "Cargo.lock hash does not match CI-generated Cargo.lock hash. Expected: $expected_hash Got: $old_hash" && exit 1
+          fi
+      - uses: ./.github/actions/cargo_target_dir_pre_cache


.github/workflows/rust.yml
```diff
@@ -77,6 +77,27 @@ jobs:
...
+  lint_cargo_lock:
+    name: Rust - Lint Cargo.lock
+    runs-on: ubuntu-20.04
+    container:
+      image: ghcr.io/build-trust/ockam-builder@sha256:e43dd94652096b03cc472a3c709c7335e8b166cab77b7a7b56f88fa38f3d24cc
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+        with:
+          ref: ${{ github.event.inputs.commit_sha }}
+      - uses: ./.github/actions/gradle_cache
```
metaclips: Let's also cache cargo build so that `cargo check` command will be fast, example [here](https://github.com/build-trust/ockam/blob/d792ddcec1194c0d666ed69c5bb7bf5409470562/.github/workflows/rust.yml#L89-L93)

metaclips (changes requested):
Thanks for an amazing PR!

.github/workflows/rust.yml
metaclips: @elliotwutingfeng seems our CI isn't happy with the current commit message as it doesn't allow uppercase, can you instead rename the commit to
ci: lint cargo.lock in \`CI\`
or
ci: lint cargo.lock in ci
or whichever amazing and descriptive commit message you can think of 🙂

mrinalwadhwa:
@elliotwutingfeng Thank you for spending time on this, before we can merge please

- [x] Accept the Ockam [Contributor License Agreement](https://github.com/build-trust/ockam-contributors/blob/main/CLA.md) by adding your Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/blob/main/CONTRIBUTORS.csv) file in a separate pull request to the [build-trust/ockam-contributors](https://github.com/build-trust/ockam-contributors) repository.

The easiest way to do this is to [edit the CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/edit/main/CONTRIBUTORS.csv) file in the github web UI and create a PR, this will make the commit as verified.

elliotwutingfeng:
Replaced `sha512sum` with `sha384sum` as `sha512sum` is vulnerable to [length-extension attacks](https://pycryptodome.readthedocs.io/en/latest/src/hash/sha512.html#sha-512-sha-512-224-sha-512-256) while `sha384sum` is [not](https://pycryptodome.readthedocs.io/en/latest/src/hash/sha384.html#sha-384).

mrinalwadhwa:
@elliotwutingfeng @metaclips I've been thinking about this PR. I think we should run this check only if a PR changes Cargo.lock so we're not forcing contributors to update cargo locks before sending unrelated PRs. What do you think?

elliotwutingfeng:
@mrinalwadhwa Yes, I think that makes sense.

.github/workflows/rust.yml
```diff
@@ -86,6 +86,10 @@ jobs:
       - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
         with:
           ref: ${{ github.event.inputs.commit_sha }}
+      - uses: technote-space/get-diff-action@v4
```
mrinalwadhwa: @elliotwutingfeng Instead of using an extra dependency. Would be better to just add another workflow, since github can trigger only on specific path changes:

https://github.com/build-trust/ockam/blob/develop/.github/workflows/rust.yml#L12-L32

mrinalwadhwa (changes requested):
left a suggestion

.github/workflows/rust.yml
elliotwutingfeng: I set it to check for `**.toml` as well since changing it would usually change `Cargo.lock`.

metaclips (dismissed):
Approved 🚀. Thanks @elliotwutingfeng for taking this through. Can you squash your commits so that we can merge right away.

elliotwutingfeng:
Done!

mrinalwadhwa (approved):
@elliotwutingfeng Thank you for an awesome first contribution 🥳 🎈

PR #3590, opened 2022-10-05
refactor(rust): change `vault create` command to use `rpc` abstraction

### LINKED ISSUE #3564: Refactor `vault create` command to use `rpc` abstraction and to stop mapping errors manually
Command: [`vault create`](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/vault/create.rs)

## Current

The current implementation is calling `std::process::exit`, which is something we want to stop using to handle errors properly and give the user a better-formatted output when a command fails.

It's also using the `connect_to` function, which can be simplified by using the `Rpc` utils. Check out how this is done in the [`subscription`](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/subscription.rs#L52) commands, for example.

## Desired

- `run` method (main entry point of the subcommand) should contain a single line of code calling a function (run_impl/rpc/...) that returns a `crate::Result<()>`
- `std::process::exit` should not be used manually to map the errors of the command's logic
- `run` method will return `()`. The error will be internally handled within the `node_rpc` function 
- If you need to replace an `std::process::exit` by an actual error, you can use something like `Err(crate::error::Error::new(exitcode::..., anyhow!(...))`
- Use `rpc` abstraction instead of `connect_to`

You can take the [`subscription`](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/subscription.rs#L52) commands as a reference.

---

kriogenia:
## Current Behavior

The current implementation is calling `std::process::exit`, which is something we want to stop using to handle errors properly and give the user a better-formatted output when a command fails.

It's also using the `connect_to` function, which can be simplified by using the `Rpc` utils.

## Proposed Changes

Following #3564 this PR:
* Uses `node_rpc` in the `run` function, invoking the new `run_impl`. Propagated errors will be handled now in that function.
* Changes the invokation of `connect_to`, `api::create_vault` and `api::parse_create_vault_response` with the use of the `Rpc` abstraction (using `Rpc::request` and `Rpc::parse_response` instead.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/util/api.rs
+++ b/implementations/rust/ockam/ockam_command/src/util/api.rs
@@ -103,15 +103,6 @@ pub(crate) fn delete_tcp_listener(cmd: &crate::tcp::listener::DeleteCommand) -> 
     Ok(buf)
 }
 
-/// Construct a request to create a Vault
-pub(crate) fn create_vault(path: Option<String>) -> Result<Vec<u8>> {
-    let mut buf = vec![];
-    Request::post("/node/vault")
-        .body(models::vault::CreateVaultRequest::new(path))
-        .encode(&mut buf)?;
-    Ok(buf)
-}
-
 /// Construct a request to export Identity
 pub(crate) fn long_identity() -> Result<Vec<u8>> {
     let mut buf = vec![];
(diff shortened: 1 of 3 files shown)


adrianbenavides (approved):
Thank you @kriogenia for an awesome first contribution! :partying_face:

PR #3607, opened 2022-10-07
refactor(rust): `node create` to return result

adrianbenavides:
Clean up `node create` command so it returns a `Result` instead of calling `std::process::exit()`.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -1,7 +1,7 @@
 use clap::Args;
 use rand::prelude::random;
 
-use anyhow::{Context as _, Result};
+use anyhow::{anyhow, Context as _, Result};
 use std::{
     net::{IpAddr, SocketAddr},
     path::{Path, PathBuf},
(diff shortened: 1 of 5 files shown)


mrinalwadhwa (approved):

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -398,6 +398,31 @@ where
     Ok(r)
 }
 
+pub fn embedded_node_daemon<A, F, Fut, T>(f: F, a: A) -> crate::Result<T>
```
mrinalwadhwa: @adrianbenavides `daemon` leads me to to think of an OS level process here. Since we're not doing that here let's call this function something else.

mrinalwadhwa (changes requested):
rename `embedded_node_deamon`

implementations/rust/ockam/ockam_command/src/util/mod.rs
adrianbenavides: I don't really know what to call it. The idea is that this function creates an embedded node and the context that is used is not stopped, so it will run indefinitely. I don't know the word for that :laughing:

mrinalwadhwa: I'm a fan of very explicit names `embedded_node_that_is_not_stopped`

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -244,76 +134,94 @@ impl CreateCommand {
...
+        // HACK: try to get the current node dir.  If it doesn't
+        // exist the user PROBABLY started a non-detached node.
+        // Thus we need to create the node dir so that subsequent
+        // calls to it don't fail
+        if cfg.get_node_dir(&cmd.node_name).is_err() {
+            println!("Creating node directory...");
+            cfg.create_node(&cmd.node_name, addr, verbose)?;
+            cfg.persist_config_updates()?;
+        }
+        embedded_node_daemon(run_background_node, (opts.clone(), cmd, addr))?;
```
mrinalwadhwa: The `run_background_node` mention in this line is strange since we're in the `if cmd.foreground` branch. We should rename that function

implementations/rust/ockam/ockam_command/src/util/mod.rs
adrianbenavides: I like that!

implementations/rust/ockam/ockam_command/src/node/create.rs
adrianbenavides: That's certainly a bad name. We are calling that function only at that line, so we could rename it to `run_foreground_node`

mrinalwadhwa (dismissed):

adrianbenavides:
@mergifyio rebase

adrianbenavides:
@Mergifyio refresh

twittner (approved):

PR #3608, opened 2022-10-07
refactor(rust): move `get_final_element` to `extract_node_name`

michealkeines:
Changes:

1. `get_final_element` with the suggested impl of `extract_node_name`
2. replaced `extract_node_name` in all the required places
3. added a test case for `extract_node_name` 

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/util/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/util/mod.rs
@@ -3,6 +3,7 @@ use std::{
     env,
     net::{SocketAddr, TcpListener},
     path::Path,
+    str::FromStr,
 };
 
 use anyhow::{anyhow, Context as _, Result};
@@ -17,7 +18,7 @@ pub use config::*;
 use ockam::{route, Address, Context, NodeBuilder, Route, TcpTransport, TCP};
 use ockam_api::nodes::NODEMANAGER_ADDR;
 use ockam_core::api::{RequestBuilder, Response, Status};
-use ockam_multiaddr::MultiAddr;
+use ockam_multiaddr::{proto::Node, MultiAddr, Protocol};
 
 use crate::node::util::start_embedded_node;
 use crate::util::output::Output;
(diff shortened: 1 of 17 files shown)


implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -534,3 +539,24 @@ pub async fn query_pid(
...
     Ok(())
 }
+
+#[test]
+fn test_extract_node_name() {
+    let test_cases = vec![
+        ("/node/test", "test", false),
+        ("test", "test", false),
+        ("/test", "/test", false),
+        ("/node/", "", false),
```
adrianbenavides: I think this case should return an error instead of an empty string. Maybe you can add a "non_empty" check before returning the string.

A couple of test cases you could add:
- `"/node"`: I think this should fail when casting the `Node` value (at line 465)
- `""`: the other empty string case

implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -534,3 +539,24 @@ pub async fn query_pid(
...
+        ("/node/test/tcp/22", "test", false),
+    ];
+
+    for (input, check, fails) in test_cases {
+        if !fails {
+            assert!(extract_node_name(input).unwrap() == check);
+        } else {
+            assert!(extract_node_name(input).is_err());
+        }
+    }
```
adrianbenavides: You can simplify this if you define the test cases with `Result`'s directly:

```rust
let test_cases = vec![
    ("/node/test", Ok("test")),
    ("/node/test/test", Err(())),
    ...
];
```

Then here you can just 

```rust
for (input, res) in test_cases {
    if let Ok(n) = res {
        assert!(extract_node_name(input).unwrap() == n);
    } else {
        assert!(extract_node_name(input).is_err());
    }
}
```

adrianbenavides (changes requested):
Almost there @michealkeines, just a couple of improvements!

adrianbenavides:
Also, make sure to run `cargo fmt --all` to fix any lint errors :pray:

michealkeines:
Hi @adrianbenavides I have added suggested changes, and also fixed to clippy warnings, please review it again, thankyou.

adrianbenavides (approved):
Awesome, thanks @michealkeines, LGTM! :rocket:

Issue #3578, opened 2022-10-04
Rust. ABAC Policy and Request format revision.


hairyhum:
Current implementation of ABAC policies does not support rules for Resource and Action attributes.
https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_abac/src/policy.rs
Also it's using entire Resource or Action data structure to retrieve policies.
https://github.com/build-trust/ockam/blob/5af0e43e7156bcd97050baae6b60c3d48df42fdf/implementations/rust/ockam/ockam_abac/src/mem.rs#L82

This makes it not possible to defined rules which would use Resource attributes as variable, for example something like:
`A subject with "domain" attribute should have access resources with the same "domain" attribute`

As an alternative, we can use resource and action names as an identifier for policy. Then ABAC request would contain:
`action_id, action_attributes, resource_attributes, subject_attributes`. This way we can find policies for an ABAC request and then use policy rules to match against all three attribute sets.

hairyhum:
@twittner feel free to rephrase the issue and add more details.

PR #3639, opened 2022-10-12
feat(rust): add small language for abac

twittner:
The language is represented using [s-expressions](https://en.wikipedia.org/wiki/S-expression) and uses the following grammar:

```
Expr    = Comment | Atom | List | Seq
Atom    = String | Int | Float | Bool | Ident
String  = UTF-8 encoded string [1]
Int     = signed 64-bit integer
Float   = 64-bit floating point value [2]
Bool    = "true" | "false"
List    = "(" Expr* ")"
Seq     = "[" Expr* "]"
Ident   = ASCII sequence without whitespace, comma, semicolon, parentheses or brackets
Comment = A line starting with ";;"

---
[1]: https://webassembly.github.io/spec/core/text/values.html#strings
[2]: https://webassembly.github.io/spec/core/text/values.html#floating-point
```

The fist element of a list (if any) is interpreted as an operator, whereas sequences are always treated as data. A number of operators are built-in:

- `and`           (logical and)
- `or`            (logical or)
- `not`           (logical not)
- `if`            (logical choice)
- ~~`+`             (addition)~~
- ~~`-`             (subtraction)~~
- ~~`*`             (multiplication)~~
- ~~`/`             (division)~~
- `<`             (less than)
- ~~`<=`            (less than or equal)~~
- `>`             (greater than)
- ~~`>=`            (greater than or equal)~~
- ~~`in` |~~ `member?` (set membership)
- `=`  ~~| `eq?`~~    (equivalence)
- `!=` ~~| `ne?`~~    (negated equivalence)

The included REPL can be used to play around with this, e.g. by executing `cargo run -p ockam_abac --bin repl --features=repl`


hairyhum:
I think this can make a nice tool for the CLI users to test their policies.
If not REPL (which can be a useful tool) then at least something like:
`ockam policy test --subject={"foo": "bar} --action={} --resource={"foos": ["bar", "baz"]} --rules=(member subject.foo resource.foos)`

implementations/rust/ockam/ockam_abac/src/eval.rs
```diff
@@ -0,0 +1,195 @@
...
+                    "*"   => eval_arith(&es[1..], env, Op::Mul),
+                    "-"   => eval_arith(&es[1..], env, Op::Sub),
+                    "/"   => eval_arith(&es[1..], env, Op::Div),
+                    "<"   => eval_pred(&es[1..], env, "<",  |a, b| a < b),
+                    "<="  => eval_pred(&es[1..], env, "<=", |a, b| a <= b),
+                    ">"   => eval_pred(&es[1..], env, ">",  |a, b| a > b),
+                    ">="  => eval_pred(&es[1..], env, ">=", |a, b| a >= b),
+                    "in" | "member" => eval_in(&es[1..], env),
+                    "="  | "eq?"    => eval_eq(&es[1..], env),
+                    "!=" | "ne?"    => eval_ne(&es[1..], env),
```
hairyhum: Let's have just one option for each.

twittner: I removed several operators. The current set is `and`, `or`, `not`, `if`, `<`, `>`, `=`, `!=`, and `member?`[^1].

[^1]: I would like to suggest to use `member?` instead of `member`. Having predicates end with a question mark is one of the nice [conventions](https://scheme.com/tspl4/intro.html#./intro:h2) in Scheme.

hairyhum (dismissed):

tools/cargo-deny/deny.toml
```diff
@@ -14,11 +14,14 @@ allow = [
...
     "ISC",
-    "BlueOak-1.0.0"
+    "BlueOak-1.0.0",
+    "BSL-1.0"
 ]
 exceptions = [
     { allow = ["LicenseRef-Ring-ISC-OpenSSL"], name = "ring", version = "*" },
     { allow = ["Unicode-DFS-2016"], name = "unicode-ident", version = "*" },
+    { allow = ["Apache-2.0 WITH LLVM-exception"], name = "wast", version = "*" },
+    { allow = ["Apache-2.0 WITH LLVM-exception"], name = "wasm-encoder", version = "*" },
```
twittner, line 24: @mrinalwadhwa: Added `BSL-1.0` and `Apache-2.0 WITH LLVM-exception` for `wast` and its dependency `wasm-encoder` only.

adrianbenavides:
hey @twittner! I'm in the process of reviewing the PR and I have some questions, mainly about its context:

- what was the state of this before your PR in terms of functionality? For example, was there some kind of parser before?

- do you have any examples like the one you showed in the demo last week? I think it would be awesome to have a readme file with those examples and a high-level explanation of the design (the references/inspirations and differences with respect to them, main features)

- not surprised coming from you, but it's really cool! Have you done something similar in the past? I'm sure you did, but just curious

twittner:
> * what was the state of this before your PR in terms of functionality? For example, was there some kind of parser before?

The previous implementation was based on an old sketch [[1]] which just evaluated logical predicates against subject attributes. Later, attributes where meant to be extended to resources and potentially actions too. Parsing was not implemented yet.

> * do you have any examples like the one you showed in the demo last week? I think it would be awesome to have a readme file with those examples and a high-level explanation of the design (the references/inspirations and differences with respect to them, main features)

At this point I don't. The PR is a stepping stone I wanted to get in. Eventually I think the documentation will need to be expanded. Currently the grammar is all that is available, plus the REPL has a :help command.

It is really just a tiny AST with some primitive types (integers, floats, booleans and strings) and two ways to combine them: lists and sequences and it uses s-expression as syntax (and CBOR for binary encoding). In a proper Lisp there would be only lists. The reason for the distinction is that the first element of a list is interpreted as a function. Scheme has a quote operator to prevent a list from being evaluated, but for the current scope it seemed simpler to me to have a dedicated expression for that. See also [4] for details regarding quote.

One thing that must change eventually is to either limit recursion depth or use a heap-allocated stack for evaluation. Otherwise there is the risk of stack overflow with deeply nested expressions.

> * not surprised coming from you, but it's really cool! Have you done something similar in the past? I'm sure you did, but just curious

Thank you for the compliment! Somehow, whenever there is a need to express syntax trees for human consumption s-expressions are the first thing that come to mind :wink: 

[1]: https://github.com/build-trust/ockam/discussions/2621#discussioncomment-2553657
[2]: https://github.com/build-trust/ockam/issues/3578
[3]: https://scheme.com/tspl4/start.html#./start:h2

adrianbenavides:
@mergifyio rebase

adrianbenavides (approved):
LGTM!

PR #3656, opened 2022-10-14
fix(rust): fix service start from launch config file

polvorin:
## Current Behavior

Starting a (foreground, but not sure that makes any difference) node with a `--launch-config`  with > 0 services to start, like:

```
{
        "startup_services": {
                "authenticator": [..edited authenticator config],
                "secure_channel_listener": {}
        }
}
```

fails,  as the tcp transport is tried to be register multiple times, once in each service' start call,  resulting in an error:

```
2022-10-13T15:15:14.388197Z TRACE ockam_transport_tcp::router::tcp_router: Registering TCP router for type = 1
2022-10-13T15:15:14.388273Z TRACE ockam_node::router: Marking address 0#e0b5687c5fefcce472175087037031be as ready!
2022-10-13T15:15:14.388450Z TRACE ockam_node::context: De-allocated detached context 0#86ad523d2e0f45410168f0891490c929
2022-10-13T15:15:14.388543Z TRACE ockam_node::context: De-allocated detached context 0#a927bc71b2ba39a670ab4c1559eb5d94
Background node error failed because router state: a router for this type already exists

Caused by:
    failed because router state: a router for this type already exists
```
 

## Proposed Changes
reuse the TcpTransport created [here](https://github.com/build-trust/ockam/blob/polvorin/fix-services-from-startup-config/implementations/rust/ockam/ockam_command/src/node/create.rs#L300) , passing it to  services start calls like [here](https://github.com/build-trust/ockam/blob/polvorin/fix-services-from-startup-config/implementations/rust/ockam/ockam_command/src/node/create.rs#L364)  .

Note: didn't pass it to the[ secure_channel_listener creator ](https://github.com/build-trust/ockam/blob/polvorin/fix-services-from-startup-config/implementations/rust/ockam/ockam_command/src/node/create.rs#L381),  as its implementation is different than the rest :man_shrugging: .   But this one didn't cause the problem that the others did.

 

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/service/start.rs
+++ b/implementations/rust/ockam/ockam_command/src/service/start.rs
@@ -1,10 +1,10 @@
 use crate::node::NodeOpts;
-use crate::util::{api, node_rpc, Rpc};
+use crate::util::{api, node_rpc, RpcBuilder};
 use crate::CommandGlobalOpts;
 use anyhow::{anyhow, Result};
 use clap::{Args, Subcommand};
 use minicbor::Encode;
-use ockam::Context;
+use ockam::{Context, TcpTransport};
 use ockam_api::DefaultAddress;
 use ockam_core::api::{RequestBuilder, Status};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_command/src/service/start.rs
```diff
@@ -137,11 +138,12 @@ async fn start_service_impl<T>(
     serv_addr: &str,
     serv_name: &str,
     req: RequestBuilder<'_, T>,
+    tcp: Option<&'_ TcpTransport>,
 ) -> Result<()>
 where
     T: Encode<()>,
 {
-    let mut rpc = Rpc::background(ctx, opts, node_name)?;
+    let mut rpc = Rpc::background(ctx, opts, node_name, tcp)?;
```
adrianbenavides: Instead of modifying the `Rpc::background` function, you have to use the `RpcBuilder`:

```rust
let mut rpc = RpcBuilder::new(ctx, opts, &cmd.node_opts.api_node).tcp(tcp)?.build();
```

adrianbenavides (dismissed):
Just a small comment about how to reuse `TcpTransport`s between different rpc calls. But overall, the solution makes total sense, thanks for taking the time to look into this!

implementations/rust/ockam/ockam_command/src/service/start.rs
polvorin: > Instead of modifying the `Rpc::background` function, you have to use the `RpcBuilder`:
> 
> ```rust
> let mut rpc = RpcBuilder::new(ctx, opts, &cmd.node_opts.api_node).tcp(tcp)?.build();
> ```

oh right!,  I didn't notice that  `tcp/1`  change the builder from `RpcMode::Embedded`  to `RpcMode::Background`  ,  I misread and assumed things created with the builder where always on `RpcMode::Embedded` mode.

Will modify :+1:

adrianbenavides:
@mergifyio rebase

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #3679, opened 2022-10-18
chore(rust): allow to store custom attributes for issuing credentials

polvorin:
## Current Behavior
Up to now, we store just a boolean indicating that given identity identifier is a _member_ of the project.  When issuing credentials, this was checked and a `role=member` credential was issued

## Proposed Changes
This change allows to store attributes instead of a boolean, paving the way to attach different attributes and values to credentials issued by a given authority (ej,  role other than 'member').

Code and test is added to ensure new code can interpret the legacy format.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
+++ b/implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
@@ -98,7 +98,9 @@ where
                 ["members"] => match self.check_enroller(&req, from).await {
                     Ok(None) => {
                         let add: AddMember = dec.decode()?;
-                        let tru = minicbor::to_vec(true)?;
+                        let member_attributes =
+                            HashMap::from([(ROLE.to_string(), MEMBER.to_string())]);
+                        let tru = minicbor::to_vec(member_attributes)?;
                         self.store
                             .set(add.member().key_id(), MEMBER.to_string(), tru)
                             .await?;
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -98,7 +98,9 @@ where
                 ["members"] => match self.check_enroller(&req, from).await {
                     Ok(None) => {
                         let add: AddMember = dec.decode()?;
-                        let tru = minicbor::to_vec(true)?;
+                        let member_attributes =
+                            HashMap::from([(ROLE.to_string(), MEMBER.to_string())]);
```
twittner, line 102: Will all member attribute values be just strings?

twittner (dismissed):

adrianbenavides:
@mergifyio rebase

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
polvorin: good question! :) 
I think in the near future it would be.   But we need to have a more broad conversation about this,   as the underling data struct is likely what would end up on credentials,  and what the  ABAC implementation run on top of.

Another thing to come back latter is the whole "schema update" of things stored on the storage.  The approach here is a workaround,  but would be better if we have a way to update _all_  legacy values on the storage to the new schema.   What complicates that right now is that there are different modules possible putting things on this storage, and we would like to update the schema for one of them, but not touch the values of the other ones.

mrinalwadhwa (approved):

Issue #3594, opened 2022-10-06
Refactor `tcp connection delete` command to use `rpc` abstraction and to stop mapping errors manually


adrianbenavides:
Command: [`tcp connection delete`](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/tcp/connection/delete.rs)

## Current

The current implementation is calling `std::process::exit`, which is something we want to stop using to handle errors properly and give the user a better-formatted output when a command fails.

It's also using the `connect_to` function, which can be simplified by using the `Rpc` utils. Check out how this is done in the [`subscription`](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/subscription.rs#L52) commands, for example.

## Desired

- `run` method (main entry point of the subcommand) should contain a single line of code calling a function (run_impl/rpc/...) that returns a `crate::Result<()>`
- `std::process::exit` should not be used to manually map the errors of the command's logic
- `run` method will return `()`. The error will be internally handled within the `node_rpc` function 
- If you need to replace an `std::process::exit` by an actual error, you can use something like `Err(crate::error::Error::new(exitcode::..., anyhow!(...))`
- Use `rpc` abstraction instead of `connect_to`

You can take the [`subscription`](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/subscription.rs#L52) commands as a reference.

---

_We love helping new contributors!_
_If you have questions or need help as you work on your first Ockam contribution, please leave a comment on [this discussion](https://github.com/ockam-network/ockam/discussions/1081)._
_If you're looking for other issues to contribute to, checkout [this discussion](https://github.com/build-trust/ockam/discussions/3206) and labels -  https://github.com/build-trust/ockam/labels/good%20first%20issue or https://github.com/build-trust/ockam/labels/help%20wanted_

Abhicodes-crypto:
I would like to take this up

divyank-aggarwal:
Hi @adrianbenavides 

I've refactored the code but is there a test which I can run or write that'd let me check a successful refactor. Would also appreciate it if you can help me with [this](https://github.com/build-trust/ockam/issues/3601#issuecomment-1282365076) issue.

Thanks for the help!

adrianbenavides:
> Hi @adrianbenavides
> 
[...]

PR #3680, opened 2022-10-19
refactor(rust): refactored node show command

### LINKED ISSUE #3601: Refactor `node show` command to use `rpc` abstraction and to stop mapping errors manually
Command: [`node show`](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/node/show.rs)

## Current

The current implementation is calling `std::process::exit`, which is something we want to stop using to handle errors properly and give the user a better-formatted output when a command fails.

It's also using the `connect_to` function, which can be simplified by using the `Rpc` utils. Check out how this is done in the [`subscription`](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/subscription.rs#L52) commands, for example.

## Desired

- `run` method (main entry point of the subcommand) should contain a single line of code calling a function (run_impl/rpc/...) that returns a `crate::Result<()>`
- `std::process::exit` should not be used to manually map the errors of the command's logic
- `run` method will return `()`. The error will be internally handled within the `node_rpc` function 
- If you need to replace an `std::process::exit` by an actual error, you can use something like `Err(crate::error::Error::new(exitcode::..., anyhow!(...))`
- Use `rpc` abstraction instead of `connect_to`

You can take the [`subscription`](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/subscription.rs#L52) commands as a reference.

---

divyank-aggarwal:
## Current Behavior

Currently the node show command is using the `connect_to` abstraction and mapping errors manually using `std::process::exit`

## Proposed Changes

As proposed by #3601 , I've created a `run_impl` function which is used along with the `node_rpc` abstraction.
The code is passing the test `create a node with a name and do show on it` in the `command.bats` test file

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/show.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/show.rs
@@ -1,16 +1,18 @@
-use crate::util::{api, connect_to, exitcode, OckamConfig};
+use crate::util::{api, node_rpc, OckamConfig, RpcBuilder};
 use crate::{help, node::HELP_DETAIL, CommandGlobalOpts};
 use anyhow::Context;
 use clap::Args;
 use colorful::Colorful;
 use minicbor::Decoder;
+use ockam::{route, TcpTransport, TCP};
 use ockam_api::config::cli::NodeConfigOld;
+use ockam_api::nodes::models::identity::ShortIdentityResponse;
 use ockam_api::nodes::models::portal::{InletList, OutletList};
 use ockam_api::nodes::models::services::ServiceList;
 use ockam_api::nodes::models::transport::TransportList;
 use ockam_api::nodes::NODEMANAGER_ADDR;
 use ockam_api::{addr_to_multiaddr, route_to_multiaddr};
-use ockam_core::api::{Response, Status};
+use ockam_core::api::{Request, Response, Status};
 use ockam_core::{Result, Route};
 use ockam_multiaddr::proto::{DnsAddr, Node, Tcp};
 use ockam_multiaddr::MultiAddr;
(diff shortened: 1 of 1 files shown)


implementations/rust/ockam/ockam_command/src/node/show.rs
```diff
@@ -31,20 +33,89 @@ pub struct ShowCommand {
...
+            eprintln!("No such node available.  Run `ockam node list` to list available nodes");
+            return Err(crate::error::Error::new(
+                exitcode::IOERR,
+                anyhow!("No nodes available for node show command"),
+            ));
+        }
+        Some(x) => x.port(),
+    };
+    let mut base_route = route![(TCP, format!("localhost:{}", port))];
+    let route = base_route.modify().append(NODEMANAGER_ADDR).into();
```
adrianbenavides: You can simplify this a little to improve readability:
```rust
let route = {
    let port = cfg.get_node_port(&node_name)?;
    let mut base_route = route![(TCP, format!("localhost:{}", port))];
    base_route.modify().append(NODEMANAGER_ADDR).into()
};
```

implementations/rust/ockam/ockam_command/src/node/show.rs
```diff
@@ -31,20 +33,89 @@ pub struct ShowCommand {
...
+    let mut child_ctx = ctx.new_detached(Address::random_local()).await?;
```
adrianbenavides: You don't really need this. You can make the `run_impl` context mutable and use it on the `is_node_up` function.

adrianbenavides (changes requested):
Looking great! I left a couple of comments on things to improve. Also, can you also commit the test you added to the `commands.bats` file (the one you mention in the description)?

divyank-aggarwal:
Thanks for the review @adrianbenavides 

I've made the 2 changes as you suggested above. The test I was referencing already exists [here](https://github.com/build-trust/ockam/blob/65a3e23b0785cff93f5ed5897d653d459da908af/implementations/rust/ockam/ockam_command/tests/commands.bats#L87) 

Would also appreciate it if you can suggest a test for another issue I'm working on #3594

metaclips-botter:
☑️ Orchestrator test on PR #3680 at commit `03f0b505f27164063e05945d90bb05b1d5d6f3a5` passed

adrianbenavides:
> The test I was referencing already exists [here](https://github.com/build-trust/ockam/blob/65a3e23b0785cff93f5ed5897d653d459da908af/implementations/rust/ockam/ockam_command/tests/commands.bats#L87)

Oh I see, perfect!

adrianbenavides (approved):
LGTM! Thanks @divyank-rs for another great contribution 🙏

PR #3695, opened 2022-10-20
feat(rust): add policy command

twittner:
Also use configured policies when checking inlet/outlet messages.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -4,7 +4,7 @@ use minicbor::Decoder;
 
 use ockam::compat::asynchronous::RwLock;
 use ockam::{Address, Context, ForwardingService, Result, Routed, TcpTransport, Worker};
-use ockam_core::api::{Error, Method, Request, Response, Status};
+use ockam_core::api::{Error, Method, Request, Response, ResponseBuilder, Status};
 use ockam_core::compat::{
     boxed::Box,
     string::String,
@@ -42,6 +42,7 @@ pub mod message;
 mod credentials;
 mod forwarder;
 mod identity;
+mod policy;
 mod portals;
 mod secure_channel;
 mod services;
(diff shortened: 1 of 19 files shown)


metaclips-botter:
☑️ Orchestrator test on PR #3695 at commit `f83612445976a5b7d60cdaaba8234d6084fd511b` passed

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -642,6 +660,25 @@ impl NodeManagerWorker {
...
+            (Get, ["policy", resource, action]) => {
+                let this = self.node_manager.read().await;
+                match this.get_policy(req.id(), resource, action).await? {
+                    Either::Left(res) => {
+                        let err = Error::new(req.path())
+                            .with_method(Method::Get)
+                            .with_message("policy not found");
+                        res.body(err).to_vec()?
+                    }
+                    Either::Right(res) => res.to_vec()?,
```
adrianbenavides: What's the benefit of using `Either` here? I see a couple of drawbacks:

1. If you are converting the `Left` variant into an error, wouldn't it be easier to create the error within `get_policy`? 
2. Also, if the `get_policy` method evolves so it returns multiple error kinds, we would have to handle them here to adapt the error message instead of just returning the error as is, right?

implementations/rust/ockam/ockam_command/src/policy.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl PolicyCommand {
+    pub fn run(self, opts: CommandGlobalOpts) {
+        node_rpc(rpc, (opts, self))
+    }
+}
+
+async fn rpc(ctx: Context, (opts, cmd): (CommandGlobalOpts, PolicyCommand)) -> Result<()> {
+    let tcp = TcpTransport::create(&ctx).await?;
```
adrianbenavides: Nit: if you are using a single RPC per command, you can remove this and create the rpc as: `Rpc::background(ctx, &opts, &node)?;`

implementations/rust/ockam/ockam_api/src/nodes/service.rs
twittner: Thanks, I moved the error creation into `NodeManager::get_policy`.

adrianbenavides:
@mergifyio rebase

adrianbenavides (approved):
LGTM!

PR #3723, opened 2022-10-24
Avoid recursion when processing policy expressions.

twittner:
Malicious input may cause stack overflows when parsing or evaluating policy expressions by recursion. Instead use heap-allocated stacks to avoid this problem.

Also make identifier matching more strict.

On top of https://github.com/build-trust/ockam/pull/3695. WIP until that one is merged.


adrianbenavides:
@mergifyio rebase

adrianbenavides (approved):

PR #3727, opened 2022-10-24
feat(rust): add command to list policies of a resource

twittner:
E.g.
```
> ockam policy set --at alice -r foo -a get -e true
> ockam policy set --at alice -r foo -a set -e false
> ockam policy set --at alice -r bar -a get -e false
> ockam policy list --at alice -r foo
foo/get: true
foo/set: false
> ockam policy list --at alice -r bar
bar/get: false
```

### DIFF
--- a/implementations/rust/ockam/ockam_abac/src/mem.rs
+++ b/implementations/rust/ockam/ockam_abac/src/mem.rs
@@ -6,6 +6,7 @@ use ockam_core::async_trait;
 use ockam_core::compat::boxed::Box;
 use ockam_core::compat::collections::BTreeMap;
 use ockam_core::compat::sync::{Arc, RwLock};
+use ockam_core::compat::vec::Vec;
 use ockam_core::Result;
 
 #[derive(Default)]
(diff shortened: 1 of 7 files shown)


implementations/rust/ockam/ockam_abac/src/mem.rs
```diff
@@ -73,6 +84,10 @@ impl PolicyStorage for Memory {
         self.inner.write().unwrap().set_policy(r, a, p);
         Ok(())
     }
+
+    async fn policies(&self, r: &Resource) -> Result<Vec<(Action, Expr)>> {
+        Ok(self.inner.write().unwrap().policies(r))
```
adrianbenavides, line 89: Given that these methods return a `Result`, why don't handle the errors with `?`? AFAIK, we do these unwraps because we are using the std locks and they are usually used like this, but is there any specific reason for not bubbling up the errors in this case?

twittner: The `Inner` methods should not panic, hence the lock should never be poisoned. The only reason I can think of is a panic due to allocation failure and this aborts the process. The module can otherwise ensure that it is panic free.

Even when mapping the `PoisonError` to `ockam_core::Error` there is also no good way to handle it because the various `Kind` and `Origin` variants do not uniquely determine the error and the calling code could not be sure it really is a `PoisonError` it is seeing and not some other I/O error (at least when abstractly calling through the trait interface).

adrianbenavides:
@mergifyio rebase

adrianbenavides (approved):

PR #3735, opened 2022-10-25
feat(rust): complete policy delete functionality

twittner:
Also disallow `':'` in identifiers to ensure resource names can not use it, as `':'` is used as the separator character in the composite LMDB key.

Finally change LMDB's delete functionality to ignore not-found errors. This also affects authenticated attributes.


adrianbenavides:
@mergifyio rebase

adrianbenavides (approved):

PR #3738, opened 2022-10-25
feat(rust): copy custom attributes from okta profiles into credentials

polvorin:
When configuring the okta' addon,  allow to pass `--attributes` argument to specify which attributes from the okta userprofile to include in the ockam credential.

Example:
```
ockam project addon configure okta --project default  --tenant $OKTA_TENANT --client-id $OKTA_CLIENT_ID  --cert-path $PATH_TO_OKTA_CERT --attributes email --attributes name
```
This will copy the `email`  and `name`  attributes from the okta profile to the ockam credential.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/project/addon.rs
+++ b/implementations/rust/ockam/ockam_command/src/project/addon.rs
@@ -107,6 +107,10 @@ pub enum ConfigureAddonCommand {
             value_parser(NonEmptyStringValueParser::new())
         )]
         client_id: String,
+
+        /// Attributes to copy from Okta userprofile into Ockam credential.
+        #[arg(long, id = "attributes", value_name = "ATTRIBUTES")]
+        attributes: Vec<String>,
     },
 }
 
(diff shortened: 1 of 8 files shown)


implementations/rust/ockam/ockam_command/src/project/addon.rs
```diff
@@ -107,6 +107,10 @@ pub enum ConfigureAddonCommand {
             value_parser(NonEmptyStringValueParser::new())
         )]
         client_id: String,
+
+        /// Attributes to copy from Okta userprofile into Ockam credential.
+        #[arg(long, id = "attributes", value_name = "ATTRIBUTES")]
+        attributes: Vec<String>,
```
adrianbenavides, line 113: Will the user always want to pass this argument? If not, it should be an `Option` that we can convert to an empty `Vec` if `None`

polvorin: If it doesn't get any  `--attributes` argument,  it already get an empty Vec.   If the user  pass  `--attributes a  --attributes b`   ,  command gets  a  `["a", "b"]` vector.

adrianbenavides: ooh ok got it! I thought clap would mark it as required

adrianbenavides:
@mergifyio rebase

adrianbenavides (approved):

PR #3759, opened 2022-10-27
refactor(rust): reduce repeated logic, make use of `rpc`

neil2468:
## Current Behavior

In `ockam_command/node`...
* `show.rs` has duplicate code.
*  `list.rs`, `create.rs` and `start.rs` are using `connect_to()` and not the newer `Rpc`. 

## Proposed Changes

* Simplify code and remove duplicates. 
* Use `Rpc` where possible.
* Avoid `std::process::exit()` where possible.

Note: `InletList` and `OutletList` changed to use non-borrowing fields so that `Rpc` instances can be re-used.

Apologies for the large PR, but `list`, `create` and `start` depend on `show`, so it kind of makes sense to do the changes as one. :pray:

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
@@ -131,12 +131,12 @@ impl<'a> CreateOutlet<'a> {
 pub struct InletStatus<'a> {
     #[cfg(feature = "tag")]
     #[n(0)] tag: TypeTag<9302588>,
-    #[b(1)] pub bind_addr: Cow<'a, str>,
-    #[b(2)] pub worker_addr: Cow<'a, str>,
-    #[b(3)] pub alias: Cow<'a, str>,
+    #[b(1)] pub bind_addr: CowStr<'a>,
+    #[b(2)] pub worker_addr: CowStr<'a>,
+    #[b(3)] pub alias: CowStr<'a>,
     /// An optional status payload
-    #[b(4)] pub payload: Option<Cow<'a, str>>,
-    #[b(5)] pub outlet_route: Cow<'a, str>,
+    #[b(4)] pub payload: Option<CowStr<'a>>,
+    #[b(5)] pub outlet_route: CowStr<'a>,
 }
 
 impl<'a> InletStatus<'a> {
@@ -153,11 +153,11 @@ impl<'a> InletStatus<'a> {
     }
 
     pub fn new(
-        bind_addr: impl Into<Cow<'a, str>>,
-        worker_addr: impl Into<Cow<'a, str>>,
-        alias: impl Into<Cow<'a, str>>,
-        payload: impl Into<Option<Cow<'a, str>>>,
-        outlet_route: impl Into<Cow<'a, str>>,
+        bind_addr: impl Into<CowStr<'a>>,
+        worker_addr: impl Into<CowStr<'a>>,
+        alias: impl Into<CowStr<'a>>,
+        payload: impl Into<Option<CowStr<'a>>>,
+        outlet_route: impl Into<CowStr<'a>>,
     ) -> Self {
         Self {
             #[cfg(feature = "tag")]
@@ -178,11 +178,11 @@ impl<'a> InletStatus<'a> {
 pub struct OutletStatus<'a> {
     #[cfg(feature = "tag")]
     #[n(0)] tag: TypeTag<4012569>,
-    #[b(1)] pub tcp_addr: Cow<'a, str>,
-    #[b(2)] pub worker_addr: Cow<'a, str>,
-    #[b(3)] pub alias: Cow<'a, str>,
+    #[b(1)] pub tcp_addr: CowStr<'a>,
+    #[b(2)] pub worker_addr: CowStr<'a>,
+    #[b(3)] pub alias: CowStr<'a>,
     /// An optional status payload
-    #[b(4)] pub payload: Option<Cow<'a, str>>,
+    #[b(4)] pub payload: Option<CowStr<'a>>,
 }
 
 impl<'a> OutletStatus<'a> {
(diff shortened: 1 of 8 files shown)


implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
```diff
@@ -131,12 +131,12 @@ impl<'a> CreateOutlet<'a> {
 pub struct InletStatus<'a> {
     #[cfg(feature = "tag")]
     #[n(0)] tag: TypeTag<9302588>,
-    #[b(1)] pub bind_addr: Cow<'a, str>,
-    #[b(2)] pub worker_addr: Cow<'a, str>,
-    #[b(3)] pub alias: Cow<'a, str>,
+    #[n(1)] pub bind_addr: Cow<'a, str>,
```
adrianbenavides: Types with lifetimes should have [the `b` attribute](https://twittner.gitlab.io/minicbor/minicbor_derive/index.html#n-and-b-or-cborn-and-cborb) so they can borrow values instead of copying them when decoding.

Also, we could replace `Cow<'a, str>` with `CowStr<'a>`, which is just a wrapper we introduced to enforce borrowing.

adrianbenavides (changes requested):
This is great, we've been wanting to do this for a long time now. I've left a small comment regarding CBOR attributes. The changes on the commands look great :+1:

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
neil2468: Hi @adrianbenavides.

Thanks for the feedback. 

I've made changes (and pushed them to this PR), but now I get a compile error which has got me stuck (see below). Do you have any pointers that might help?

PS. The previous `InletList` and `OutletList` were based on what I saw for `ServiceList` and `TransportList`.

Thanks in advance 😄 

```
warning: /Users/neil/dev/ockam/Cargo.toml: unused manifest key: ockam_command
   Compiling ockam_api v0.19.0 (/Users/neil/dev/ockam/implementations/rust/ockam/ockam_api)
error: lifetime may not live long enough
   --> implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs:218:24
    |
218 | #[derive(Debug, Clone, Decode, Encode)]
    |                        ^^^^^^
    |                        |
    |                        lifetime `'bytes` defined here
    |                        type annotation requires that `'bytes` must outlive `'a`
...
221 | pub struct InletList<'a> {
    |                      -- lifetime `'a` defined here
    |
    = help: consider adding the following bound: `'bytes: 'a`
    = note: this error originates in the derive macro `Decode` (in Nightly builds, run with -Z macro-backtrace for more info)

error: lifetime may not live long enough
   --> implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs:238:24
    |
238 | #[derive(Debug, Clone, Decode, Encode)]
    |                        ^^^^^^
    |                        |
    |                        lifetime `'bytes` defined here
    |                        type annotation requires that `'bytes` must outlive `'a`
...
241 | pub struct OutletList<'a> {
    |                       -- lifetime `'a` defined here
    |
    = help: consider adding the following bound: `'bytes: 'a`
    = note: this error originates in the derive macro `Decode` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `ockam_api` due to 2 previous errors
```

neil2468: I changed the structs to the `b` attribute (and force pushed them to this PR). Now, `Rpc` can't be reused for the Inlet and Outlet requests.

```
error[E0502]: cannot borrow `*rpc` as mutable because it is also borrowed as immutable
   --> implementations/rust/ockam/ockam_command/src/node/show.rs:174:9
    |
171 |         let inlets = rpc.parse_response::<InletList>()?;
    |                      --------------------------------- immutable borrow occurs here
...
174 |         rpc.request(api::list_outlets()).await?;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here
...
185 |             Some((&inlets, &outlets)),
    |                   ------- immutable borrow later used here

For more information about this error, try `rustc --explain E0502`.
```

https://github.com/build-trust/ockam/blob/8f364c0f33c658114f5c2ddc40543ce452121a18/implementations/rust/ockam/ockam_command/src/node/show.rs#L169-L186

The previous `InletList` and `OutletList` were based on what I see for `ServiceList` and `TransportList`.

Would you like me to __(1)__ change `print_query_status()` so it stops re-using `Rpc` between the Inlet and Outlet requests? or __(2)__ roll back so `InletList` and `OutletList` are like `ServiceList` and `TransportList` and do not borrow values?

Thanks in advance! 😄

adrianbenavides: We can roll back to clone values and tackle this in a separate PR. Otherwise, this PR can get too big. Also, we have other structures that don't have the correct CBOR attributes, so it might be a good idea to step back and plan this carefully.

Sorry for the confusion!

mrinalwadhwa:
Since today is the last day of `hacktoberfest`, I've add the `hacktoberfest-accepted` label in case you're participating. We can merge the PR after @adrianbenavides has a chance to review.

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
neil2468: > Sorry for the confusion!

That's totally fine, this is all good experience for me.

In this last force push, I've kept the borrowing CBOR fields and implemented `Clone` for `Rpc`. Let me know what you think. If you'd prefer, I can roll back to the non-borrowing CBOR fields.

Thanks!

adrianbenavides: That's perfect! It's what I had in mind since we don't care if we clone the responses on the CLI side. To be honest, I didn't think it would be that easy :laughing:

adrianbenavides (approved):
Awesome PR @neil2468 as always, thank you so much! :pray:

PR #3770, opened 2022-10-28
feat(elixir): allow configuring idle timeout for workers

hairyhum:
Add new option to worker creation: `:idle_timeout`, which can be a number in milliseconds or `:infinity`.

If `idle_timeout` is a number, the worker will terminate if no messages were received during the timeout.

## Checks

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/worker.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/worker.ex
@@ -120,6 +120,10 @@ defmodule Ockam.Worker do
         Ockam.Worker.handle_message(__MODULE__, message, state)
       end
 
+      def handle_info(:idle_timeout, state) do
+        Ockam.Worker.handle_idle_timeout(state)
+      end
+
       @doc false
       @impl true
       def handle_continue(:post_init, options) do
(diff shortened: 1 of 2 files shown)


implementations/elixir/ockam/ockam/lib/ockam/worker.ex
```diff
@@ -246,6 +255,63 @@ defmodule Ockam.Worker do
     end
   end
 
+  def schedule_idle_timeout(state) do
```
polvorin, line 258: seems simpler to return the timeout as part of the  `handle_info| handle_continue|  handle_call`  callbacks?,  and rely on genserver to track it for us.     Unless there is some functionality that we want to enforce other than plain time between received msgs

hairyhum: I don't think we can do that from generic worker without intercepting all worker `handle_*` functions, which we currently don't do.
Also gen_server events are not all ockam messages and gen_server timeouts will not happen if there are gen_server events. For example our metrics poller may check secure channels using gen_server events.

hairyhum:
https://github.com/Mergifyio rebase

hairyhum:
@Mergifyio rebase

polvorin (approved):

PR #3777, opened 2022-10-31
refactor(elixir): `TypedCBOR.encode()` returns a tuple on success and…

### LINKED ISSUE #3189: TypedCBOR: `encode` wrapping exception does not return `{:ok ...}` tuple
There are two functions defined in `Ockam.TypedCBOR` for encoding: `encode` and `encode!`.

Former is calling `wrap_exception(&encode!...` which will return `{:error, reason}` if there is an exception in `encode!` instead of throwing the exception.

But if there is no exception, the return value is `binary()` and not `{:ok, binary()}` and the function return value is used directly in the code.

This means that if we encounter exception, there will be `{:error, reason}` tuples in the data instead of encoded values because it's not checked anywhere.

As a solution we can switch all current usage of `encode` to `encode!` and fix the function to return `{:ok, binary()}` on success instead of `binary()` so we can use `encode` in the future code.

davidag:
… failure

Fixes #3189

## Current Behavior

`TypedCBOR.encode()` returns a value on success and a tuple `{:error, reason}` on error.

## Proposed Changes

Make `TypedCBOR.encode()` return a tuple `{:ok, data}` on success in symmetry with `TypedCBOR.decode()`. Solution described in #3189

## Checks

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/api/discovery/service_info.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/api/discovery/service_info.ex
@@ -29,8 +29,8 @@ defmodule Ockam.API.Discovery.ServiceInfo do
     end
   end
 
-  def encode_list(service_infos) do
-    service_infos |> Enum.map(&normalize/1) |> NormalizedServiceInfo.encode_list()
+  def encode_list!(service_infos) do
+    service_infos |> Enum.map(&normalize/1) |> NormalizedServiceInfo.encode_list!()
   end
 
   def decode_list(data) do
@@ -47,8 +47,8 @@ defmodule Ockam.API.Discovery.ServiceInfo do
     end
   end
 
-  def encode(%__MODULE__{} = service_info) do
-    normalize(service_info) |> NormalizedServiceInfo.encode()
+  def encode!(%__MODULE__{} = service_info) do
+    normalize(service_info) |> NormalizedServiceInfo.encode!()
   end
 
   def decode(data) do
--- a/implementations/elixir/ockam/ockam/lib/ockam/api/request.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/api/request.ex
@@ -27,7 +27,7 @@ defmodule Ockam.API.Request do
     end
   end
 
-  def encode(request) when is_map(request) do
+  def encode!(request) when is_map(request) do
     body = request.body
 
     has_body =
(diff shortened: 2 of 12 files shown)


hairyhum:
@davidag thank you for the contribution. Can you also please check usages of `encode` for `Ockam.API.Discovery.ServiceInfo`, `Ockam.API.Request` and `Ockam.API.Response`?

davidag:
@hairyhum Updated so that all three use `encode!()` instead of `encode()`, which I think is better because now errors will raise.

implementations/elixir/ockam/ockam/lib/ockam/api/discovery/service_info.ex
```diff
@@ -48,7 +48,7 @@ defmodule Ockam.API.Discovery.ServiceInfo do
   end
 
   def encode(%__MODULE__{} = service_info) do
```
hairyhum: For consistency this should also be called `encode!`

implementations/elixir/ockam/ockam/lib/ockam/api/request.ex
```diff
@@ -38,7 +38,7 @@ defmodule Ockam.API.Request do
       end
 
     base =
-      Header.encode(%Header{
```
hairyhum, line 41: Let's rename `Request.encode/1` -> `Request.encode!/1` for consistency. We can address mixed use of `:bare` and `CBOR` later.

implementations/elixir/ockam/ockam/lib/ockam/api/response.ex
```diff
@@ -36,7 +36,7 @@ defmodule Ockam.API.Response do
       end
 
     base =
-      Header.encode(%Header{
```
hairyhum, line 39: Let's rename `Response.encode/1` -> `Response.encode!/1` for consistency. We can address mixed use of `:bare` and `CBOR` later.

hairyhum (changes requested):

implementations/elixir/ockam/ockam/lib/ockam/api/response.ex
davidag: Good point, I've changed all three 👌🏼

hairyhum:
I've just found that update in return type of `encode/1` changes the return type of `encode_list/1` for the modules using typed cbor plugin.
@davidag  Can you please check and update usages of `encode_list/1` to `encode_list!/1`?

davidag:
@hairyhum fixed one usage of `encode_list()` and rebased. Let me know if there's anything else missing 🙏🏼

hairyhum (dismissed):

hairyhum (approved):

PR #3783, opened 2022-11-01
feat(rust): always load policy from storage

twittner:
The current `PolicyAccessControl` keeps a policy expression in memory and only reads identity attributes when determining if a message is authorised or not. The policy storage is not used, except when initially creating a `PolicyAccessControl` value.

As a potential alternative, this PR implements a more flexible but slower approach where the policy expression is always read from the underlying policy storage and therefore reflects any changes to policies immediately. If no policy is found, access is denied.

To allow updating policies of inlets and outlets, they may be given alias names which can then be used as resource names to link them to policies.

### DIFF
--- a/implementations/rust/ockam/ockam_abac/src/policy.rs
+++ b/implementations/rust/ockam/ockam_abac/src/policy.rs
@@ -10,6 +10,8 @@ use tracing as log;
 
 use crate::eval::eval;
 use crate::expr::str;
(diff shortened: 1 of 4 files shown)


adrianbenavides (approved):
Thanks for adding all those comments, it made the PR a lot easier to navigate and review. LGTM!

PR #3786, opened 2022-11-02
tests(rust): add utility to run commands tests that depend on local nodes only

adrianbenavides:
Adding small tests to the `commands.bats` file to test the commands is slowing down the CI pipelines a lot. The core idea behind the proposed changes is to be able to run the tests that depend on local state to rust-based tests and reuse a small set of nodes, so we can avoid creating and destroying nodes before every test (like it's done in the commands.bats file), which is the operation that takes the most time to complete.

## Discarded ideas - Using an object pool

The initial solution I explored was to have an object pool that could be accessed in parallel. When running multiple tests, I found that sometimes the node creation fail, probably due to a filesystem error (more specifically, the error was `VaultError::StorageError`).

After trying different pool sizes and different object initialization strategies, I've dropped this solution.

## Proposed solution

Due to the limitations described above, and to keep the PR simple and with the minimum side effects possible, I've opted for the simplest solution: a vector of nodes behind a mutex. The direct implication of this is that all tests using this vector will run serially, but we still get the speed up from not having to recreate nodes for every test.

## CI Improvement

- The `Ockam Command Test` stage has dropped from 17m to 14m (this can be improved after porting more tests)
- The `Rust Test` stage remains a ~7m


implementations/rust/ockam/ockam_command/src/test_utils/node.rs
```diff
@@ -0,0 +1,61 @@
...
+    Arc::new(Mutex::new(node_names))
+}
+
+pub(crate) fn test_nodes() -> MutexGuard<'static, Vec<TestNode>> {
+    NODES_POOL.lock().unwrap()
+}
+
+pub(crate) struct TestNode {
+    name: String,
+    init: RefCell<bool>,
```
twittner: If `TestNode` is not required to be `Sync` I would go with `Cell` or else with `AtomicBool`.

implementations/rust/ockam/ockam_command/src/test_utils/node.rs
```diff
@@ -0,0 +1,61 @@
+use assert_cmd::prelude::*;
+use duct::cmd;
+use lazy_static::lazy_static;
+use rand::random;
+use std::cell::RefCell;
+use std::sync::{Arc, Mutex, MutexGuard};
+
+lazy_static! {
```
twittner: Have you considered `once_cell`?

implementations/rust/ockam/ockam_command/src/test_utils/cmd.rs
```diff
@@ -0,0 +1,184 @@
...
+use std::time::SystemTime;
+use std::{fs, io};
+use std::{fs::File, io::Read, thread::sleep, time::Duration};
+use tempfile::NamedTempFile;
+use thiserror::Error;
+use tracing::info;
+use tracing_subscriber::fmt::TestWriter;
+use tracing_subscriber::{filter::LevelFilter, fmt, EnvFilter};
+
+const POLL_MS: u32 = 250;
```
twittner: Why a unit suffix instead of `Duration`?

implementations/rust/ockam/ockam_command/src/test_utils/node.rs
adrianbenavides: Old habits... once_cell definitely looks better. Thanks!

implementations/rust/ockam/ockam_command/src/test_utils/cmd.rs
adrianbenavides: Changed to durations.

implementations/rust/ockam/ockam_command/src/test_utils/node.rs
adrianbenavides: Since `TestNode`'s are behind a mutex, is better to use `Cell`over `AtomicBool`, right?

twittner: If `TestNode` is used behind a mutex it is better to use a regular `bool` and use `&mut self` instead of `&self`.

etorreborre:
@adrianbenavides is it something that you plan to resume working on? Or should it be changed with your recent work on tests parallelization?

adrianbenavides:
> @adrianbenavides is it something that you plan to resume working on?

Yes, I was waiting until we removed the shared data between nodes entirely, but I don't think that's going to happen until after we are done with adding new commands.

> Or should it be changed with your recent work on tests parallelization?

In any case, after having worked on that, I believe we could resume this PR as the only shared file (the config.json) is not used in unit tests, which are the kind of tests I wanted to port to rust.

I'll try to come back to this as soon as I can.

PR #3793, opened 2022-11-03
feat(rust): `check-credential` argument defaults to node's `enable-credential-checks` value

adrianbenavides:
The `tcp-inlet create` and `tcp-outlet create` commands now don't have to explicit the `check-credential` value. The node will default to its `enable-credential-checks` value passed upon creation.

These commands can override the node's value using one of the following attributes: `check-credential` and `disable-check-credential`


polvorin (dismissed):

polvorin (dismissed):

mrinalwadhwa:
@Mergifyio rebase

mrinalwadhwa (approved):

PR #3791, opened 2022-11-03
feat(rust): add default value for `to` argument and change the `attributes` argument name

adrianbenavides:
Apply the following changes to the `project enroll` command:
- change `–attr` to `–attribute`
- make `--to` attribute optional

PR #3796, opened 2022-11-03
chore(rust): rename attr argument to attribute on okta addon config

polvorin:
Similar to https://github.com/build-trust/ockam/pull/3791,  but when configuring okta addon


polvorin:
@Mergifyio rebase

mrinalwadhwa (approved):

Issue #3701, opened 2022-10-20
Configure Rust nodes to support Elixir ports


hairyhum:
Elixir Ockam nodes sometimes use Rust nodes to fill the gaps in the implementation. For example Identity and Credential Verifier services are delegated to Rust nodes.

Currently this means that there should be TCP-accessible Rust node deployed next to the Elixir node and there needs to be some orchestration to manage them. This if fairly simple if Elixir nodes are deployed with orchestration (like k8s), but makes it more challenging to do more portable deployments.

Elixir has [Ports](https://hexdocs.pm/elixir/1.13/Port.html) as a tool to integrate with external programs. This tool allows to start external programs from Elixir code and also lets Elixir processes to link with them, so they can be a part of the supervision trees.

We can start a Rust node from the Port, but currently when Port terminates the Rust node does not. This is because termination of the Port is communicated to external program by sending it the `EOF`.
The idea behind Ports is that they communicate with Elixir on `STDIN` and `STDOUT`. 

If we could have an argument we can add to the Rust node creation so it will monitor `STDIN` and terminate on `EOF` it should be easy enough to integrate Rust "sidecar" nodes as a part of Elixir nodes, simplifying deployment.

Another thing we can do is to add a `STDIN/STDOUT` transport protocol to leverage the Ports even more.

---

_We love helping new contributors!_
_If you have questions or need help as you work on your first Ockam contribution, please leave a comment on [this discussion](https://github.com/ockam-network/ockam/discussions/1081)._
_If you're looking for other issues to contribute to, checkout [this discussion](https://github.com/build-trust/ockam/discussions/3206) and labels -  https://github.com/build-trust/ockam/labels/good%20first%20issue or https://github.com/build-trust/ockam/labels/help%20wanted_

hairyhum:
I think the first step would be to add an argument to the `node create` command to make it listen on `STDIN` and terminate after receiving `EOF`, this argument should be probably somehow related to the `foreground` argument since it doesn't make much sense for background nodes.

phillyphil91:
hey @hairyhum is this a rust only issue? if so, I would like to work on this

mrinalwadhwa:
@phillyphil91 yes the following part is Rust only:

> I think the first step would be to add an argument to the node create command to make it listen on STDIN and terminate after receiving EOF, this argument should be probably somehow related to the foreground argument since it doesn't make much sense for background nodes.

[...]

PR #3804, opened 2022-11-06
add function to check STDIN for EOF

phillyphil91:
## Current Behavior 

Elixir nodes use Rust nodes sometimes through the use of Ports. When an Elixir node terminates the Rust node which it created does not.  

## Proposed Changes
 
This change adds functionality to monitor STDIN and terminate a RUST node on EOF input (e.g. from the Elixir node)
This PR address the issue #3701

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -2,6 +2,7 @@ use clap::Args;
 use minicbor::Decoder;
 use ockam_identity::credential::Credential;
 use rand::prelude::random;
+use tokio::io::AsyncBufReadExt;
 
 use anyhow::{anyhow, Context as _};
 use std::{
@@ -58,6 +59,10 @@ pub struct CreateCommand {
     #[arg(display_order = 900, long, short)]
     pub foreground: bool,
 
+    /// Watch stdin for EOF
+    #[arg(display_order = 900, long = "exit-on-eof", short)]
+    pub exit_on_eof: bool,
+
     /// TCP listener address
     #[arg(
         display_order = 900,
@@ -98,6 +103,7 @@ impl Default for CreateCommand {
     fn default() -> Self {
         Self {
             node_name: hex::encode(random::<[u8; 4]>()),
+            exit_on_eof: false,
             tcp_listener_address: "127.0.0.1:0".to_string(),
             foreground: false,
             child_process: false,
@@ -193,7 +199,7 @@ fn create_foreground_node(opts: &CommandGlobalOpts, cmd: &CreateCommand) -> crat
 }
 
 async fn run_foreground_node(
-    ctx: Context,
+    mut ctx: Context,
     (opts, cmd, addr): (CommandGlobalOpts, CreateCommand, SocketAddr),
 ) -> crate::Result<()> {
     let cfg = &opts.config;
@@ -285,6 +291,32 @@ async fn run_foreground_node(
         }
     }
 
+    if cmd.exit_on_eof {
+        stop_node_on_eof(&mut ctx, &opts, &cmd.node_name).await?;
+    }
+
+    Ok(())
+}
+
+// Read STDIN until EOF is encountered and then stop the node
+async fn stop_node_on_eof(
+    ctx: &mut Context,
+    opts: &CommandGlobalOpts,
+    node_name: &str,
+) -> crate::Result<()> {
+    let reader = tokio::io::BufReader::new(tokio::io::stdin());
+    let mut lines = reader.lines();
+
+    loop {
+        match lines.next_line().await {
+            Ok(Some(_)) => (),
+            Ok(None) => break,
+            Err(_) => unreachable!(),
+        }
+    }
+
+    ctx.stop().await?;
+    opts.state.nodes.get(node_name)?.kill_process(false)?;
     Ok(())
 }
 


phillyphil91:
Hi @hairyhum 
I wasn't entirely sure how to implement this, given if I have even understood the problem correctly. I found multiple ways of monitoring the STDIN for EOF alongside the code that `ockam node create --foreground` runs anyways. 
One which is currently commented out just monitors STDIN on a new thread and then communicates those events to a channel. I was not really sure when and how to query the messages on that channel (e.g. if/when to use `loop`).

The other approach which is implemented atm is using an async fn that calls `StopCommand.run()` when EOF is seen on STDIN and then terminates the process and node. It works manually, although it prints that `Node n1 is not running!`, it looks like the node is stopped. 

I am also not sure how to test this properly. Maybe this needs to be tested on the Elixir side from which the Rust node is called?

phillyphil91:
Another similar idea would be to use `lines()` on the `BufReader` to create a stream over the lines of the reader and thus no buffer would be needed. 
```rust
async fn stop_node_on_eof(node_name: String, opts: CommandGlobalOpts) {
    let stop = StopCommand {
        node_name: node_name,
        force: true,
    };

    let reader = tokio::io::BufReader::new(tokio::io::stdin());
    let mut lines = reader.lines();

    loop {
        match lines.next_line().await {
            Ok(Some(_)) => (),
            Ok(None) => break, //EOF
            Err(_) => panic!("Something went wrong"),
        }
    }

    stop.run(opts)
}
```

hairyhum:
To test Elixir integration you can use:
Assuming the flag to monitor stdin is called `--watch-stdin`

```
# run elixir shell
bash$ iex 

# open port with the node
iex> port = Port.open({:spawn, "ockam node create --foreground --watch-stdin"}, [:binary])
# close the port and terminate the node
iex> Port.close(port)

# make sure there is no ockam
bash$ pgrep ockam
```

You can also add the step to check that node is running between `Port.open` and `Port.close`.

Ideally if we can hook up to other input from STDIN it would help in the future if we want to implement such an interface. 
But for now EOF is the only thing we need to handle.
But let's have some separate flag to enable that anyway.

phillyphil91:
Thanks @hairyhum the test via the elixir shell worked. 
A couple of points from my side:

1. I added the flag for `--watch-stdin`
2. I decided to check for `EOF` in `STDIN` by looping through the stream of lines that come in and break when `EOF` is encountered. I think this is better than reading into a buffer that is not used anyways, but I'm not really sure, feel free to correct me. (The two different approaches are in the two commits)
3. In the first commit I just called `StopCommand.run()` but that ends up returning `Ok(_)` when trying to call `cfg.get_node_pid()` and thus executing:
```rust
eprintln!("Node {} is not running!", &self.node_name);                                                                                                         
std::process::exit(exitcode::IOERR);
```
I am not sure why this is happening but maybe the node is not able to find its own PID? But since this is the case I just called `std::process::exit(exitcode::IOERR);` directly without calling `StopCommand.run()`

4. The node is now stopped when receiving `EOF` from `STDIN` but the node remains registered with status: `DOWN`. Wouldn't it be better if the node is completely deleted once this happens. Because otherwise there is this orphan node that was created by elixir, right?

hairyhum:
> I added the flag for --watch-stdin

👍 

> I decided to check for EOF in STDIN by looping through the stream of lines that come in and break when EOF is encountered. I think this is better than reading into a buffer that is not used anyways, but I'm not really sure, feel free to correct me. (The two different approaches are in the two commits)

That should be fine for now, I think.

> I am not sure why this is happening but maybe the node is not able to find its own PID? But since this is the case I just called std::process::exit(exitcode::IOERR); directly without calling StopCommand.run()

I'm not sure what's the consequence of this, TBH. @SanjoDeundiak @twittner can you please take a look?

> The node is now stopped when receiving EOF from STDIN but the node remains registered with status: DOWN. Wouldn't it be better if the node is completely deleted once this happens. Because otherwise there is this orphan node that was created by elixir, right?

Just stopping the node is better because we don't want to delete the node data in case it needs to restart.

phillyphil91:
Hi @hairyhum and @SanjoDeundiak any update on this?

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -270,14 +277,34 @@ async fn run_foreground_node(
...
+
+    loop {
+        match lines.next_line().await {
+            Ok(Some(_)) => (),
+            Ok(None) => break,
+            Err(_) => panic!("Something went wrong"),
+        }
+    }
+
+    std::process::exit(exitcode::IOERR);
```
adrianbenavides: You can pass the `ockam::Context` as a variable to this function and then call `ctx.stop().await();`. That way the node will be gracefully stopped and exit the process without throwing an error.

Also, since this stops the process, you should do the equivalent to what's happening in the Stop command which is just calling the next function: `opts.state.nodes.get(&cmd.node_name)?.kill_process(false);`. This call will kill the process and delete all the node's state data we store on disk.

adrianbenavides (commented):
Hey @phillyphil91 sorry for taking so long to review your PR, last week we were all working on some stuff and didn't have the time to review community PR's.

First off, thank you for taking the time to look into this, it looks pretty good! We've done some simplifications on this command and you have some conflicts now, but they're pretty straightforward to solve. Other than that, I've left a comment on how to exit the process after receiving the EOF signal.

Let me know if I can help you with anything else!

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -53,6 +55,10 @@ pub struct CreateCommand {
     #[arg(display_order = 900, long, short)]
     pub foreground: bool,
 
+    /// Watch stdin for EOF
+    #[arg(display_order = 900, long = "watch-stdin", short)]
+    pub watch_stdin: bool,
```
adrianbenavides: @hairyhum should this be the default behavior? Also, would it make more sense to rename it to `exit-on-eof`? (cc @phillyphil91)

hairyhum: I don't think it should be default, but I like the `exit-on-eof` arg name.

phillyphil91:
Hi @hairyhum and @adrianbenavides I tried to include all suggestions. 
I had to make ctx mutable in  `run_foreground_node()` to be able to call `ctx.stop().await();` if I understand correctly. 

The elixir shell examples work via manual tests.

Please let me know what you think and I will change to the correct commit massage..

adrianbenavides (approved):
Awesome, I think we can merge it now! I've squashed and fixed the commit messages for you to pass the CI lint. Thanks @phillyphil91 for a great contribution :partying_face:

phillyphil91:
Hi @adrianbenavides thanks for the approval. it was a great issue. looking forward to more.

PR #3812, opened 2022-11-07
chore(rust): add comment about always starting echoer service

neil2468:
## Current Behavior

When an `ockam_api::NodeManagerWorker` is initialised, the code that starts the `echoer` service is apart from the code that starts the other default services.

This means that the `echoer` service does not mirror what the other services do. For example, if `ockam node create n1 --skip-defaults` is run then the `echoer` service is started and the other default services are not.

In my opinion, the `echoer` serivce should behave like the other default services. 

## Proposed Changes

Move the code for starting the `echoer` service to be with the code that starts the other default services. 

## Checks

Thanks! :smile:

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -356,6 +356,8 @@ impl NodeManager {
             }
         }
 
+        // Always start the echoer service as ockam_api::Medic assumes it will be
+        // started unconditionally on every node. It's used for liveness checks.
         s.start_echoer_service_impl(ctx, DefaultAddress::ECHO_SERVICE.into())
             .await?;
 


twittner:
The reason why echo is started separately and unconditionally is because it serves as checkpoint for liveness checks, see: https://github.com/build-trust/ockam/blob/0b828228d06ac367a8ebdb2a5c7de1da8c482dbf/implementations/rust/ockam/ockam_api/src/session.rs#L80

neil2468:
Thanks @twittner. That makes sense. :smile:

I've updated the PR to only add a comment detailing why the echoer is started there.

adrianbenavides (approved):
LGTM! Thanks @neil2468 for bringing this up, I think it's a detail that would have confused anyone reading this code. 🚀

PR #3817, opened 2022-11-08
fix(rust): change if/and/or evaluation of abac expressions

twittner:
Currently if expressions evaluate both, their true branch and their false branch. This does not work when using `exists?` to check if identifiers are bound as the use of unbound identifiers in either branch would cause an error, defeating the purpose of `exists?`. To fix this, the evaluation of `if` expressions now only evaluates the test predicate and only then, depending on the outcome, will either the true or the false branch be evaluated.

**Edit**: Additionally, `and` and `or` expressions now evaluate their arguments lazily, stopping as soon as the logical result can be determined, i.e. for `and` as soon as a `false` and for `or` as soon as a `true` value is encountered.

### DIFF
--- a/implementations/rust/ockam/ockam_abac/src/eval.rs
+++ b/implementations/rust/ockam/ockam_abac/src/eval.rs
@@ -40,8 +40,46 @@ pub fn eval(expr: &Expr, env: &Env) -> Result<Expr, EvalError> {
                 [Expr::Ident(id), ..] => {
                     let nargs = xs.len() - 1; // number of arguments
                     match id.as_str() {
-                        "and" => ctrl.push(Op::And(nargs)),
-                        "or"  => ctrl.push(Op::Or(nargs)),
+                        "and" => {
+                            // 'and' evaluates its arguments lazily. As soon as a
+                            // false value is encountered evaluation stops and
+                            // the remaining arguments are just popped off the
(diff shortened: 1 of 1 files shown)


implementations/rust/ockam/ockam_abac/src/eval.rs
```diff
@@ -52,7 +52,13 @@ pub fn eval(expr: &Expr, env: &Env) -> Result<Expr, EvalError> {
...
                                 return Err(EvalError::malformed("'if' requires three arguments"))
                             }
-                            ctrl.push(Op::If)
+                            // We first evaluate the test and only then, depending on the result
+                            // do we either evaluate the true branch or the false branch.
+                            ctrl.push(Op::Eval(&xs[3])); // false branch
+                            ctrl.push(Op::Eval(&xs[2])); // true branch
+                            ctrl.push(Op::If);
+                            ctrl.push(Op::Eval(&xs[1])); // test
+                            continue
```
adrianbenavides, line 99: Is there a test evaluating an expression like this as an example? Or at least, I think it would be great to have an example in the comment (line 56) to help visualize the `xs`'s expected elements.

polvorin (dismissed):

adrianbenavides:
@mergifyio rebase

adrianbenavides (approved):

PR #3826, opened 2022-11-08
feat(rust): adapt and re enable complex bats tests

polvorin:
Tests involving credentials and project enrollment where disabled,  here we adapt them to take into account the changes introduced,  and re-enabled them. 
Added the case of  enrollment-token member enrollment.


adrianbenavides:
@mergifyio rebase

adrianbenavides (approved):

PR #3831, opened 2022-11-10
Stricter ABAC lang

twittner:
Some proposals to tweak ABAC processing:

- Reverts the addition of `null` and brings back `EvalError::unbound`.
- Returns an error when comparing values of different types instead of evaluating to `false`.
- Requires sequences to be homogeneous.

### DIFF
--- a/implementations/rust/ockam/ockam_abac/src/env.rs
+++ b/implementations/rust/ockam/ockam_abac/src/env.rs
@@ -1,74 +1,64 @@
-use crate::error::MergeError;
+use crate::error::{EvalError, MergeError};
 use crate::expr::Expr;
 use ockam_core::compat::collections::BTreeMap;
-use ockam_core::compat::string::String;
+use ockam_core::compat::string::{String, ToString};
 
-#[derive(Debug, Clone)]
-pub struct Env {
-    map: BTreeMap<String, Expr>,
-    null: Expr,
-}
-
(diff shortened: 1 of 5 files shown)


hairyhum:
What's the impact of this?

twittner:
> What's the impact of this?

The first change (removing `null`), restores the behaviour as of c7f2195103d48aa6d6e8d336220fb51514cbca60, i.e. unbound identifiers cause an evaluation error. The theme of this PR is to avoid mixing types in evaluations. Something like `(!= subject.role "guest")` should error instead of returning `true` if `subject.role` is unbound.

The other changes are along similar lines. Comparing values of different types (e.g. numbers and strings) is an error whereas currently it would return `false`. Homogeneous sequences means that elements of a sequence have to have the same type.

Policies after this PR which want to work with potentially unbound identifiers need to guard them with `exists?`. Logical operators are short-circuting, i.e. their evaluation stops as soon as the result can be determined, so instead of `(op x y ...)` where `x` is potentially unbound, one could write `(and (exists? x) (op x y ...))` or `(if (exists? x) (op x y ...) ...)` is evaluation to a bool is desired.

hairyhum (dismissed):

adrianbenavides:
@mergifyio rebase

adrianbenavides (approved):

PR #3839, opened 2022-11-17
Access controls

SanjoDeundiak:
## What is done in that PR:
1. Every `Context` (Worker, Processor or detached) creation now must provide mandatory incoming and outgoing access controls
1. Provide/fix following access controls implementations: `LocalOnwardOnly`, `LocalSourceOnly`, `AllowOnwardAddress`, `AllowSourceAddress`, etc.
1. Mark every questionable place in the code related to access controls with `@ac` tag
1. Setup access controls for most of the workers started internally: tcp, secure channels, identity secure channels, forwarders, etc.
1. Related updates to `ockam_node` crate: `RelayMessage` struct, etc.
1. Update test, examples and documentation
1. Other things: see changes and commit messages


implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -302,21 +287,28 @@ impl Context {
...
+    ///     ctx.start_worker("my-worker-address", MyWorker, Arc::new(AllowAll), Arc::new(AllowAll)).await
     /// }
     /// ```
-    pub async fn start_worker<NM, NW>(&self, address: impl Into<Address>, worker: NW) -> Result<()>
+    pub async fn start_worker<NM, NW>(
+        &self,
+        address: impl Into<Address>,
+        worker: NW,
+        incoming: Arc<dyn AccessControl>,
+        outgoing: Arc<dyn AccessControl>,
```
twittner: I noticed that in almost all calls to `start_worker`, a fresh `Arc` is created which makes the API a little bit involved. How about accepting `impl AccessControl` here for `incoming` and `outgoing` and creating the `Arc`s in the body of `start_worker`? This would simplify many calls. If the calling code wants to retain control over the allocation it can always use the `WorkerBuilder` directly. Similar reasoning applies to `start_processor` and `new_detached` methods.

implementations/rust/ockam/ockam_core/src/access_control.rs
```diff
@@ -32,66 +32,18 @@ pub trait AccessControl: Debug + Send + Sync + 'static {
     async fn is_authorized(&self, relay_msg: &RelayMessage) -> Result<bool>;
 }
```
twittner, line 34: Some forward impls for `Arc` and `Box` may be useful, e.g.

```rust
#[async_trait]
impl<T: AccessControl + ?Sized> AccessControl for Arc<T> {
    async fn is_authorized(&self, relay_msg: &RelayMessage) -> Result<bool> {
        (**self).is_authorized(relay_msg).await
    }
}
```

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
```diff
@@ -153,6 +153,7 @@ pub(crate) struct Registry {
     pub(crate) okta_identity_provider_services: BTreeMap<Address, OktaIdentityProviderServiceInfo>,
     pub(crate) uppercase_services: BTreeMap<Address, UppercaseServiceInfo>,
     pub(crate) echoer_services: BTreeMap<Address, EchoerServiceInfo>,
+    pub(crate) hop_services: BTreeMap<Address, EchoerServiceInfo>,
```
twittner: Should there be a `HopServiceInfo`?

implementations/rust/ockam/ockam_transport_core/src/error.rs
```diff
@@ -38,6 +38,8 @@ pub enum TransportError {
     PortalInvalidState,
     /// InvalidRouterResponseType
     InvalidRouterResponseType,
+
+    TempError,
```
twittner: What kind of error does this denote?

SanjoDeundiak: This is me trying to catch WebSocket bug on CI, apparently it doesn't reproduce now 🥲
I'll remove it before the merge

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
SanjoDeundiak: Nice catch!

implementations/rust/ockam/ockam_node/src/context.rs
SanjoDeundiak: Yes, I was thinking about it, but was a little bit uncertain if there are cases where we want to be able to clone existing Arc. Using `WorkerBuilder` in those cases sounds reasonable. I'll try to do the change, should simplify many things

implementations/rust/ockam/ockam_core/src/access_control.rs
SanjoDeundiak: What is the benefit of that?

twittner: It gives client code flexibility. Assume you have a value of a type that implements `AccessControl` in an `Arc` and you want to call some API that accepts any implementation of `AccessControl`. Without a forwarding impl you can not call the function with the value you have as `Arc<T>` for `T: AccessControl` is not itself an implementation of `AccessControl`. Similarly for `Box` or shared references.

SanjoDeundiak: Tnx for the explanation. I decided to use dynamic dispatch for `AccessControl`, to support changing type of the implementation in the runtime, e.g. a `Context` could want to change its access control from `DenyAll` to `AllowAll`. Because of that I think we should pass Access Control mostly as a reference or wrapped in an `Arc`. And I think it's advisable to be able to distinguish between `T` and `Arc<T>`, for example in cases like your suggestion above: to let `ctx.start_worker` accept `impl AccessControl`. If `Arc<T>` is passed their, we'll end up creating `Arc<Arc<T>>`, don't we?

twittner: `Arc<Arc<T>>` is not wrong. Another question is of course why the `Arc` is needed at all? Maybe a `Clone` constraint would be better to express the need to create duplicates?

SanjoDeundiak: How would you use `Clone` in dynamic dispatch usecase?

twittner: I only glanced over the bit about storing trait objects, but yes, for storing a value as a field without involving generics some indirection is needed.

SanjoDeundiak: That's why I decided to go with `Arc` for now

twittner: I still do not understand why you want to prevent someone with an `Arc<T>` or `Box<T>` where `T` implements `AccessControl` to pass that to functions that are generic over `AccessControl` types, but :shrug:

documentation/guides/rust/get-started/05-secure-channel/README.md
```diff
@@ -38,19 +38,22 @@ Add the following code to this file:
...
+    ctx.start_worker("echoer", Echoer, AllowAll, AllowAll).await?;
 
     // Initialize the TCP Transport.
     let tcp = TcpTransport::create(&ctx).await?;
 
     // Create a TCP listener and wait for incoming connections.
-    tcp.listen("127.0.0.1:4000").await?;
+    // Use port 4000, unless otherwise specified by command line argument.
+    let port = std::env::args().nth(1).unwrap_or_else(|| "4000".to_string());
+    tcp.listen(format!("127.0.0.1:{port}")).await?;
```
mrinalwadhwa: Let's revert the port input changes back to the way they were. We want examples to stay focused on teaching the core idea we're introducing.

SanjoDeundiak: The problem is that these examples were not checked on CI, I added that check. Fixed ports can cause problems during parallel runs. Thoughts?

mrinalwadhwa: 🤔 Maybe we can tell the test runner to run the tests that involve ports in sequence.

SanjoDeundiak: Maybe, but as far as I remember - using posts 3000 4000 on CI is road to random failures

mrinalwadhwa: We can make some other, less likely to collide, ports the hardcoded values - maybe 4441, 4442, 4443.

That would be a good change, we've come across similar collisions with macos as well.

But we should make it, consistently, across all examples and diagrams.
There is this PR that didn't change diagrams https://github.com/build-trust/ockam/pull/2700

SanjoDeundiak: @mrinalwadhwa I just noticed that our get started guide for transports (that precedes this one) already uses port numbers from cli arguments. Therefore I don't get why doing the same in this guide is bad

mrinalwadhwa: It's bad because this adds noise to the example that takes attention away from the main ideas we're trying to teach here. Think of a non rust developer or not a developer at all reading this.

If the transport example uses it, it must've missed my check .. let's change that to hardcoded values as well

SanjoDeundiak: @mrinalwadhwa I reverted get started examples to the simpler state and created an issue for further improvement instructions. I think we should not prevent this PR from landing because of that issue
https://github.com/build-trust/ockam/issues/4015

mrinalwadhwa: sounds good 👍

examples/rust/get_started/examples/04-udp-transport-initiator.rs
```diff
@@ -9,7 +9,10 @@ async fn main(mut ctx: Context) -> Result<()> {
     let _udp = UdpTransport::create(&ctx).await?;
 
     // Send a message to the "echoer" worker, on a different node, over an udp transport.
-    let r = route![(UDP, "localhost:4000"), "echoer"];
+    // Wait to receive a reply and print it.
+    // Use port 4000, unless otherwise specified by command line argument.
+    let port = std::env::args().nth(1).unwrap_or_else(|| "4000".to_string());
+    let r = route![(UDP, format!("localhost:{port}")), "echoer"];
```
mrinalwadhwa: let's change this to fixed ports as well

examples/rust/get_started/examples/04-udp-transport-responder.rs
```diff
@@ -11,10 +12,12 @@ async fn main(ctx: Context) -> Result<()> {
     let udp = UdpTransport::create(&ctx).await?;
 
     // Create a UDP listener and wait for incoming datagrams.
-    udp.listen("127.0.0.1:4000").await?;
+    // Use port 4000, unless otherwise specified by command line argument.
+    let port = std::env::args().nth(1).unwrap_or_else(|| "4000".to_string());
+    udp.listen(format!("127.0.0.1:{port}")).await?;
```
mrinalwadhwa: let's change this to fixed ports as well

mrinalwadhwa (changes requested):
let's also change the UDP examples to fixed ports.

examples/rust/get_started/examples/04-udp-transport-responder.rs
SanjoDeundiak: @mrinalwadhwa done

examples/rust/get_started/examples/04-udp-transport-initiator.rs
SanjoDeundiak: @mrinalwadhwa done

mrinalwadhwa (approved):

PR #3842, opened 2022-11-18
feat(rust): define a new structure to store CLI state and config

adrianbenavides:
Re: https://github.com/build-trust/codebase/issues/460

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/state.rs
+++ b/implementations/rust/ockam/ockam_command/src/state.rs
@@ -1,0 +1,550 @@
+use anyhow::Context;
+use ockam_identity::change_history::{IdentityChangeHistory, IdentityHistoryComparison};
+use ockam_identity::{Identity, IdentityIdentifier};
+use ockam_vault::{storage::FileStorage, Vault};
+use serde::{Deserialize, Serialize};
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+
+#[derive(Clone)]
+pub struct CliState {
+    pub vaults: VaultsState,
+    pub identities: IdentitiesState,
+    pub nodes: NodesState,
+    dir: PathBuf,
+}
+
+impl CliState {
+    pub fn new() -> anyhow::Result<Self> {
+        let dir = Self::dir()?;
+        Ok(Self {
+            vaults: VaultsState::new(&dir)?,
+            identities: IdentitiesState::new(&dir)?,
+            nodes: NodesState::new(&dir)?,
+            dir,
+        })
+    }
+
+    fn dir() -> anyhow::Result<PathBuf> {
+        Ok(match std::env::var("OCKAM_HOME") {
+            Ok(dir) => PathBuf::from(&dir),
+            Err(_) => dirs::home_dir()
+                .context("no $HOME directory")?
+                .join(".ockam"),
+        })
+    }
+}
+
+#[derive(Clone)]
+pub struct VaultsState {
+    dir: PathBuf,
+}
+
+impl VaultsState {
+    fn new(cli_path: &Path) -> anyhow::Result<Self> {
+        let dir = cli_path.join("vaults");
+        std::fs::create_dir_all(&dir)?;
+        Ok(Self { dir })
+    }
+
+    pub fn create(&self, name: &str, config: VaultConfig) -> anyhow::Result<VaultState> {
+        let path = {
+            let mut path = self.dir.clone();
+            path.push(format!("{}.json", name));
+            if path.exists() {
+                return Err(anyhow::anyhow!("vault `{name}` already exists"));
+            }
+            path
+        };
+        let contents = serde_json::to_string(&config)?;
+        std::fs::write(&path, contents)?;
+        if !self.default_path().exists() {
+            self.set_default(name)?;
+        }
+        Ok(VaultState { path, config })
+    }
+
+    pub fn get(&self, name: &str) -> anyhow::Result<VaultState> {
+        let path = {
+            let mut path = self.dir.clone();
+            path.push(format!("{}.json", name));
+            if !path.exists() {
+                return Err(anyhow::anyhow!("vault `{name}` does not exist"));
+            }
+            path
+        };
+        let contents = std::fs::read_to_string(&path)?;
+        let config = serde_json::from_str(&contents)?;
+        Ok(VaultState { path, config })
+    }
+
+    pub fn default_path(&self) -> PathBuf {
+        let mut path = self.dir.clone();
+        path.push("default");
+        path
+    }
+
+    pub fn default(&self) -> anyhow::Result<VaultState> {
+        let path = std::fs::canonicalize(&self.default_path())?;
+        let contents = std::fs::read_to_string(&path)?;
+        let config = serde_json::from_str(&contents)?;
+        Ok(VaultState { path, config })
+    }
+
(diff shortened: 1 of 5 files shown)


implementations/rust/ockam/ockam_command/src/state.rs
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+    pub fn stderr_log(&self) -> PathBuf {
+        self.path.join("stderr.log")
+    }
+}
+
+#[derive(Default, Debug, Clone, Eq, PartialEq)]
+pub struct NodeConfig {
+    vault: Option<String>,
```
polvorin: would call these `default_vault`  and `default_identity`

implementations/rust/ockam/ockam_command/src/state.rs
```diff
@@ -0,0 +1,535 @@
...
+        file_stem(&path)
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct NodeState {
+    version: String,
+    name: String,
+    path: PathBuf,
+    vault: VaultState,
```
polvorin: `default_vault`   and `default_identity`

implementations/rust/ockam/ockam_command/src/state.rs
```diff
@@ -0,0 +1,535 @@
...
+        std::fs::File::create(state.stdout_log())?;
+        std::fs::File::create(state.stderr_log())?;
+        std::os::unix::fs::symlink(&state.vault.path, state.path.join("vault"))?;
+        std::os::unix::fs::symlink(&state.identity.path, state.path.join("identity"))?;
+        Ok(state)
+    }
+
+    pub fn get(
+        &self,
+        vault: VaultState,
```
polvorin: the default vault and default identity for the existing node are  on ` state.path.join("vault")`  and ` state.path.join("identity")`  .   We shouldn't pass them here

adrianbenavides:
@mergifyio rebase

implementations/rust/ockam/ockam_command/src/state.rs
```diff
@@ -0,0 +1,535 @@
...
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct VaultState {
+    path: PathBuf,
+    pub config: VaultConfig,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
+#[serde(untagged)]
+pub enum VaultConfig {
```
polvorin, line 119: I'm not entirely sure how things are structured, but have the _feeling_ this should be somewhere else.
The reasoning is

1) Create a node `$ockam node create n1`
2) Start a secure channel with a _specific_  vault and identity:  `$ockam [...]  --identity id1  --vault v1`  

The question is how the cli,  running on step 2,   pass the vault and identity to use to `n1` (the running background node) .   I _think_  serializing the entire change history is not the best approach as that's something that can change  (like rotating the keys on the identity).   So perhaps passing the path to the file is better?. 

What do you think?   I don't have this part particularly clear. 

(also,  I'm assuming the background node doesn't have knowledge of this top-level OCKAM_HOME thing,  and instead only knows _his_ particular node' state directory,   and any other path that the cli communicate to it)

implementations/rust/ockam/ockam_command/src/state.rs
```diff
@@ -0,0 +1,535 @@
...
+        let mut path = self.dir.clone();
+        path.push(name);
+        path.push("identity");
+        let path = std::fs::canonicalize(&path)?;
+        file_stem(&path)
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct NodeState {
```
polvorin, line 323: this looks like something the background running nodes would use.   Is this definition available to them?  or they code is defined on a different crate like ockam_api or something?

polvorin (changes requested):
Requested name changes from  `identity` and  `vault`  to `default_identity`  and `default_vault` 
(as those are the default ones that will be used by the node,  for any operation that requires vault and identities  *if*  specific ones aren't provided with the call).

The other comment is more open-ended,  perhaps it's ok just need some clarification for me to understand :)

implementations/rust/ockam/ockam_command/src/state.rs
adrianbenavides: In my opinion, the `default` concept is only valid from the `NodeManager` perspective, where you can override the usage of the vault/identity found in the node's state directory with the ones from the command payload. 

At this layer though, given that there is only one possible vault and one possible identity, introducing the `default` notion might be misleading as there is a non-default.

Let me know if that makes sense to you!

implementations/rust/ockam/ockam_command/src/state.rs
adrianbenavides: I've done some changes to update the `Node` sections.

implementations/rust/ockam/ockam_command/src/state.rs
adrianbenavides: Ideally, the ockam_api crate would **not** need to know anything about this state.rs file, but we might need to move something to the ockam_api crate. I'd wait until that happens.

For this PR, I wanted to keep this first version on a single file and on the ockam_command crate to keep it as simple as possible.

implementations/rust/ockam/ockam_command/src/state.rs
adrianbenavides: When running `ockam node create --identity id1 --vault v1 [...]` the command internals will look like this:

```rust
let vault_path = options.vaults.get(&vault_name).path;
let identity_path = options.identities.get(&identity_name).path;
let node_config = NodeConfig::new(&vault_path, &identity_path);
let node_state = options.nodes.create(&node_name, node_config);
``` 

The `node_state` contains the _paths_ to the vault and identity. That's what we will send to the node upon creation. Then, every time we need to use the vault or identity, we will recreate them from scratch, use them, and drop them. That way, if we need to mutate an identity (rotate the key), we can do that from a command for example, and the nodes using that identity will use the updated version.

Does this make sense to you?

implementations/rust/ockam/ockam_command/src/state.rs
polvorin: disagree :shrug: .     Neither the vault or identity is **the** node' vault and identity.    They are the defaults that will be used if other is not specified

implementations/rust/ockam/ockam_command/src/state.rs
polvorin: ok.  There are two path when talking  file-based vaults:    One to the file that said that the vault is a file-based one (instead of a kms based one, etc)  and that has the path to the actual file-backed vault.  And the other is that path to the file-backed vault itself.

From the command to the running node we pass the first path,  the path to vault' definition.     The background running node needs to be able to parse that file struct (what is called `VaultConfig` here).   I mean either parse it from disk,  or the client need to serialize that `VaultConfig` struct and give it to the background node.   In either case the server code needs to have access to `VaultConfig`'  definition,   that's what I'm unsure how is done.
But ok to move definitions around latter.

polvorin (dismissed):
lets address the changes latter so following PRs don't get blocked.

polvorin:
@Mergifyio rebase

polvorin (approved):

PR #3848, opened 2022-11-21
chore(rust): identity show do not run on embedded node

polvorin:
do not call delete_embedded_node/2, as it is not an embedded node, and end up unlinking the node' state directory..


polvorin:
@Mergifyio rebase

adrianbenavides (approved):
A couple of weeks ago I did a review of every command trying to find misuses of the `delete_embedded_node` and overlooked this one. Thanks @polvorin!

PR #3849, opened 2022-11-22
fix(rust): replace atty dependency with is-terminal

adrianbenavides:
Fixes https://github.com/build-trust/ockam-artifacts/issues/98

Also, `atty` is pretty much unmaintained, while `is-terminal` has recent activity.


adrianbenavides:
@mrinalwadhwa should we move this forward? I would do it even if it's just to remove the (unmaintained) `atty` dependency.

mrinalwadhwa:
Yes, let's do it

mrinalwadhwa (approved):

Issue #3808, opened 2022-11-07
Ockam command man page generation


hairyhum:
`ockam_command` is using `clap` for command line parsing. There is a tool to generate man files from `clap` config https://crates.io/crates/clap_mangen

With some improvements around https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/lib.rs to explain environment variables it can be used to produce decent enough documentation to existing commands.

adrianbenavides:
I can take a look into this.

My main concern is about how to integrate it with our CI pipeline: maybe adding the man page generation in the `build.rs` will slow it down too much. And I never heard of the proposed alternative ([xtask](https://github.com/matklad/cargo-xtask)) Any preference/idea here @metaclips ?

adrianbenavides:
I've done a bit of research on this. Here are the main findings:
- Trying to generate the man page from ockam_command's `build.rs` is not feasible as it needs to depend on itself to generate the docs. I've tried it and failed miserably
- The proper approach would be to have a separate package with the sole purpose of building the man page. The building process can then be automated with [cargo xtask](https://github.com/matklad/cargo-xtask). Here is [an example](https://github.com/ctrlcctrlv/humnumsort) of how to set this up

Other notes:
- `clap_mangen` is relatively new (February 2022)

### Generating manpages

I was only able to generate the manpage for the main command (the output of `ockam --help`). To generate manpages for subcommands, we should derive `Parser` instead of `Args` in every subcommand so we can run the following:

```rust
let cmd = <ockam_command::ProjectCommand as clap::CommandFactory>::command();
```

Some questions:
- Not sure this is the right way to do this. We need to figure out if `clap_manpage` can discover subcommands automatically
- How should the different manpages be organized?

hargut:
Hi @hairyhum,
Hi @adrianbenavides,

I've looked into the topic and started to implement a subcommand that generates the man pages.

The PR adds a `manpages` subcommand, which when called, writes the man page files to one of the following directories:
- value from user provider `--dir|-d` option
- `~/.local/share/man/man1`
- current working directory `$PWD/ockam_man_pages` in case `$HOME` cannot be determined


Additionally I've added an option to disable gzip compression for the generated man pages. The current implementation generates one man page for the main command, and one per subcommand. Generation only done on users request calling the subcommand.

I'd be happy to receive your inputs and feedback.

Many thanks in advance.

Best regards,
Harald

hairyhum:
[...]

PR #3852, opened 2022-11-22
chore(rust): ockam command add subcommand to generate man pages

hargut:
## Current Behavior

Currently there are no man pages available for ockam.

## Proposed Changes

The PR adds a subcommand called `manpages` which generates man pages and writes the files to a user selected directory, the users default man page location within the home directory or to a fallback location based on the current working directory.

One man page is generated for the main command, and one for each subcommand. For further details please refer to the according issue:
https://github.com/build-trust/ockam/issues/3808

## Checks


implementations/rust/ockam/ockam_command/src/manpages.rs
```diff
@@ -0,0 +1,134 @@
...
+    match generate_man_page(man_dir, cmd.get_name(), cmd, no_compression) {
+        Ok(()) => (),
+        Err(error) => error!(
+            "Error generating man page for command \"ockam\": {:?}",
+            error
+        ),
+    }
+
+    // generate man page for sub commands
+    for s_cmd in cmd.get_subcommands() {
```
adrianbenavides, line 93: I think this loop should be run recursively, as most (if not all) of the subcommands have 1 or 2 layers of nested subcommands.

adrianbenavides (commented):
This looks pretty awesome, thanks @hargut for taking a look into this!

I'm not sure we need a subcommand for this, but rather a simple function to generate the man files from CI. Also, we don't know yet how we want to organize the man files, but we can leave that for a future PR.

In any case, this is a great start. I left a small comment :+1:

implementations/rust/ockam/ockam_command/src/manpages.rs
hargut: Good input, thank you.
I've have switched that part to a recursive call.

hargut:
@adrianbenavides 
with the input from your comment I've done the following adaptions:
- extracted the manpage generation to a function
- switched to a recursive call for generating the man pages to cover all levels of subcommands
- added the hidden flag to the manpages subcommand so it does not appear for users in the help, but could be used e.g. for CI integration
- fixed some formatting issues

Initial my thought was to add the subcommand so the distribution process can further be based on a single binary.

implementations/rust/ockam/ockam_command/src/manpages.rs
```diff
@@ -0,0 +1,140 @@
...
+            }
+            None => {
+                let mut man_dir = env::current_dir()?;
+                man_dir.push("ockam_man_pages/");
+                man_dir
+            }
+        },
+    };
+
+    println!("man pages directory: {}", man_dir.display());
```
adrianbenavides: If the user provides a directory, it's already known. If the output is stored in `~/.local/share/man/man1/`, the user doesn't really care because it's the default directory and will be able to access them directly when running `man ockam`.

So, I'd show the directory only in the case we store the output in a non expected directory (after line 70). And I would change the message to `println!("Man pages stored at: {}", man_dir.display());`

implementations/rust/ockam/ockam_command/src/manpages.rs
```diff
@@ -0,0 +1,140 @@
...
+        default_value = "false",
+        action = ArgAction::SetTrue,
+        help = "disable gzip compression for man page output",
+    )]
+    no_compression: bool,
+}
+
+impl ManpagesCommand {
+    pub fn run(self) {
+        write_man_pages(&self.dir, &self.no_compression);
```
adrianbenavides: Given that the `run` function only calls one function, I'd include the `write_man_pages` contents directly here to avoid a level of indirection.

implementations/rust/ockam/ockam_command/src/manpages.rs
```diff
@@ -0,0 +1,140 @@
...
+    }
+}
+
+fn write_man_pages(dir: &Option<String>, no_compression: &bool) {
+    let man_dir = match get_man_page_directory(dir) {
+        Ok(path) => path,
+        Err(error) => panic!("Error getting man page directory: {:?}", error),
+    };
+    let clap_command = <OckamCommand as CommandFactory>::command();
+    generate_man_pages(man_dir.as_path(), &clap_command, None, no_compression);
```
adrianbenavides: You should move the `no_compression` boolean instead of passing the reference to this function. It's cheap to clone (although you don't need to in this case) and it's more comfortable to use.

implementations/rust/ockam/ockam_command/src/manpages.rs
```diff
@@ -0,0 +1,140 @@
...
+                man_dir.push("ockam_man_pages/");
+                man_dir
+            }
+        },
+    };
+
+    println!("man pages directory: {}", man_dir.display());
+    if !man_dir.exists() {
+        create_dir(man_dir.clone())?;
+    }
```
adrianbenavides: Nit: you can replace this with `create_dir_all`, as you don't have to check if it already exists.

adrianbenavides (commented):
I left some more suggestions. And here are a couple of extra things:
- When you are done with all the changes, could you squash all the commits and force push?
- When I tried the command on my machine, the output contained colors (like `[1;4mFeedback: [0m    `). I think this is due to the `COLORFGBG` environment variable. Can you take a look at this? Maybe there is a way to unset that variable. If you don't see an easy way to do this, we can tackle it on a separate PR.

implementations/rust/ockam/ockam_command/src/manpages.rs
hargut: Inlined that part.

implementations/rust/ockam/ockam_command/src/manpages.rs
hargut: Done.

implementations/rust/ockam/ockam_command/src/manpages.rs
hargut: Changed.

implementations/rust/ockam/ockam_command/src/manpages.rs
hargut: Switched.

hargut:
Hi @adrianbenavides,l

I've addressed the inputs from the review and squashed all changes into a single commit.

Some background and information on the colour topic:
- Currently the colours are added in [help.rs](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/help.rs#L21) which is called during adding the help messages for clap (e.g [node/mod.rs](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/node/mod.rs#L124)).
- Within clap there is only the colourized string version available, which ended up in the man page.
- Without removing the syntax highlighting for colour enabled terminals within the help messages a way to solve that problem is to filter the ascii control characters before writing the man page to disk. This is what I've added to the PR.

I'm not very happy with that solution, and think it might be cleaner to think about removing the colourization within help messages.


Best regards,
Harald

adrianbenavides (approved):
Awesome! I agree that we have to revisit how we handle colorization, but for now, this solution is perfect. We'll create more issues on this topic to keep improving the CLI docs.

Again, thanks @hargut for this contribution! :pray:

PR #3854, opened 2022-11-23
refactor(rust): use new state structure on `identity create` command

adrianbenavides:


implementations/rust/ockam/ockam_command/src/vault/create.rs
```diff
@@ -38,14 +37,11 @@ async fn run_impl(ctx: Context, (options, cmd): (CommandGlobalOpts, CreateComman
         rpc.is_ok()?;
         println!("Vault created for the Node {}!", node_name);
     } else {
-        // Only file-backed vaults for now
         let path = state::VaultConfig::fs_path(&cmd.name, cmd.path)?;
-        if !path.exists() {
-            FileStorage::create(path.clone()).await?;
-        }
```
adrianbenavides, line 45: This part here is hidden behind the `config.get().await?;` call. I want to keep all the details about vault creation scoped within the `state` boundaries.

polvorin (approved):

PR #3860, opened 2022-11-25
refactor(rust): `identity create` creates default vault if doesn't exist

adrianbenavides:


polvorin (approved):
we might need to refactor a bit in the future to make the default vault creation accessible from other commands (thinking node create would also need the functionality to create an initial vault if none exists).   But we will see that once there

PR #3861, opened 2022-11-25
feat(rust): on `node create` create symlinks to default vault/identity

adrianbenavides:
It includes some TODO's that will be addressed in future PR's, like the `pid`, `version` and `NodeSetupConfig`.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/util.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/util.rs
@@ -1,6 +1,7 @@
 use std::sync::Arc;
 
 use anyhow::{anyhow, Context as _, Result};
+use rand::random;
 use sysinfo::{get_current_pid, ProcessExt, System, SystemExt};
 use tracing::trace;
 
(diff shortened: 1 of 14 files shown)


implementations/rust/ockam/ockam_command/src/node/util.rs
```diff
@@ -105,6 +109,22 @@ pub(super) async fn create_default_identity_if_needed(
 
     cfg.persist_config_updates()?;
 
+    // Store vault and identity in the new directory
+    let vault_name = hex::encode(random::<[u8; 4]>());
```
polvorin, line 113: I'm not sure I follow this part,   but it's already different on https://github.com/build-trust/ockam/pull/3866 where the default' symlinks are copied around

polvorin:
@adrianbenavides   I'll be in favor of just closing this one and https://github.com/build-trust/ockam/pull/3865 ,   and merge https://github.com/build-trust/ockam/pull/3866 that contains both already

implementations/rust/ockam/ockam_command/src/node/util.rs
adrianbenavides: Yeah exactly, this was a workaround before what I did on https://github.com/build-trust/ockam/pull/3866

adrianbenavides:
Merged through https://github.com/build-trust/ockam/pull/3866

PR #3865, opened 2022-11-28
feat(rust): use cli_state when creating nodes

adrianbenavides:
## Changes

* Move `CliState` to `ockam_api`  
* Refactor the to-be-deprecated current NodeConfig so it points to the new directories, so nodes are created using the new structure and the new vault/identity symlinks.
* Removes the `default_vault.json` file and the `default_identity` from the global `config.json`
* Removes the methods on the NodeManager to create vaults/identities.

## Current state

Now, when creating a node with `ockam node create`, it will create the default vault and identity, and a random node using those vault and identity.


adrianbenavides:
Merged through https://github.com/build-trust/ockam/pull/3866

PR #3866, opened 2022-11-28
feat(rust): `node create` can be used with custom vault and identity

adrianbenavides:
Add new arguments to `node create` to pass a custom `vault` and `identity` that will be set as the default for the new node.

## Open questions

If the user runs the command `ockam node create --vault v1 --identity idt1` on a **fresh install**:
a. should the command return an error telling the user that the passed vault/identity don't exist?
b. or the command should create the vault/identity?

I would say `a` because `b` could lead to errors if the user mistypes the vault/identity name.


polvorin:
yes,  agree that if the user specify some identity/vault and it doesn't already exist, it must be an error :+1:

polvorin (approved):
This is looking good! :clap: .  Did some tests and all seems to be working as expected.
As said in the others PR,  I'd close the others and merge this one,  but up to you. 

Now that different nodes can actually share the underling vault (as well as the cli itself),   we would need https://github.com/build-trust/codebase/issues/479  (or some similar thing,  479 needs a round of feedback)

PR #3869, opened 2022-11-30
refactor(rust): remove old configs and move missing fields to new node directory

adrianbenavides:
Big PR, mainly removing old stuff about CLI/nodes config files:
- Remove `NodeConfigOld` struct, which contained the node's data (name, address, port, verbose)
- Remove `NodeConfig` struct, which contained the path to the `commands.json` and `state.json` files
- Remove the `Commands` struct, as we are not using it, and probably we'll change our approach on this (declarative config file). We can recover it if we need it in the future
- Remove `nodes` section from the `LookupConfig`. Everything now relies on the new `CliState` to retrieve information about the nodes

It also removes some unused commands:
- `ockam configuration set`: was used to change a node name. It doesn't make much sense right now to keep it
- `ockam node run`: was used to run the `commands.json`, which is now deleted

Finally, refactors a bunch of nodes to use the CliState instead of the NodeConfig.

Note that we still have the `OckamConfig`, which holds the reference to the Lookup. Eliminating this will require some effort since we'll have to refactor a bunch of nodes to stop relying on the lookup data to retrieve projects/spaces data.


adrianbenavides:
@mergifyio rebase

polvorin (approved):

PR #3870, opened 2022-11-30
ecdsa p-256 and aws kms integration

twittner:
Implements https://github.com/build-trust/codebase/issues/461.
Replaces #3859.
~~NB: MPL-2.0 license needs to be allowed (CC: @mrinalwadhwa)~~
**Edit**: Removed `evercrypt` as it pulls in outdated crates cargo-deny is rejecting.


implementations/rust/ockam/ockam_vault/src/aws.rs
```diff
@@ -0,0 +1,270 @@
...
+        Ok(Self { client, config: c })
+    }
+
+    /// Create an AWS KMS client using the default configutation.
+    pub async fn default() -> Result<Self> {
+        Self::new(Config::default()).await
+    }
+
+    /// Create a new NIST P-256 key-pair in AWS KMS and return its ID.
+    pub async fn create_key(&self) -> Result<KeyId> {
```
polvorin, line 48: can we change the `KeyId`  type here to be something else (maybe `KMSKeyId`  or something like that),  to make it clear that these are  *not*  the same thing as the `KeyId`  we index on?  (the ones that are a function of the public key).   IIUC  when signing for example we go from  
    `PubKey -->[hash_function()]-->  KeyId -->[lookup()]-->KMSKeyId -->[aws_kms_call(KMSKeyId)]--> signature`

implementations/rust/ockam/ockam_command/src/identity/create.rs
```diff
@@ -44,7 +55,16 @@ async fn run_impl(
         options.state.vaults.default()?.config
     };
     let vault = vault_config.get().await?;
-    let identity = Identity::create(&ctx, &vault).await?;
+    let identity = if let Some(kid) = cmd.key_id {
```
polvorin, line 58: I'm not entirely sure that passing a key_id is only applicable for kms vaults. 

Suppose we have keys already on a non-kms vault  (like the existing file-based one).   Could we want to generate an identity using an existing key there instead of generating a new one?.    That looks like something that could be general and not specific to kms.   *But*  that brings the question of what the key_id parameter *is*,  to be general it shouldn't be something special about kms.

We could do something like:
create the vault
`$ ockam vault create v1 --kms`

import the kms key.  This would compute the `KeyId`  and add the mapping in our  storage from `KeyId -> $kms-key-id`,  and return the `KeyId`   
` $ ockam vault import-key  v1  $kms-key-id` 
      
Then the --key-id  we optionally pass when creating an identity is always a (ockam)  KeyId,   and is not specific to a specific kind of vault.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -105,6 +105,14 @@ pub struct CreateCommand {
...
     #[arg(long = "identity", value_name = "IDENTITY")]
     identity: Option<String>,
+
+    /// Enable AWS KMS integration.
+    #[arg(long, default_value = "false")]
+    aws_kms: bool,
+
+    /// Use an existing AWS KMS key.
+    #[arg(long)]
+    key_id: Option<String>,
```
polvorin, line 115: Just my opinion here,
these `aws_kms` and `key_id` are  used if there is no vault already,   I _feel_ it is clearer if we don't allow to automatically create kms vaults here,  and instead go only with the simplest one (the file based one,  on the default location).     To use a kms vault as the default one for a node,   first explicitly create the vault.    For example we allow to use a different file system path than the default for storage,  but don't have that option here on node create  (so to use a different path,  explicitly create the vault first).

adrianbenavides: I agree on this. We want to provide "defaults" everywhere we can, but only to cover the simplest/default use case, and the KMS feature is not part of that.

As [noted here](https://github.com/build-trust/ockam/pull/3870/files#r1036895504), we can keep this PR as it is and I can do the needed changes after this PR is merged.

implementations/rust/ockam/ockam_command/src/identity/create.rs
adrianbenavides: I like that approach BUT, since that is more a `command` related stuff, we can create an issue and I can do that on a separate PR. Let me know what you guys think.

implementations/rust/ockam/ockam_vault/src/secret_impl.rs
```diff
@@ -39,6 +48,25 @@ impl Vault {
                 ))
                 .await?
             }
+            SecretType::NistP256 => '_block: {
+                #[cfg(feature = "aws")]
+                if attributes.persistence() == SecretPersistence::Persistent {
+                    if let Some(kms) = &self.aws_kms {
+                        if let Secret::Aws(kid) = secret {
+                            let pk = kms.public_key(kid).await?;
+                            break '_block self.compute_key_id_for_public_key(&pk).await?;
```
adrianbenavides, line 57: What does this syntax mean? It will step out of the `SecretType::NistP256` match, but when? After `self.compute_key_id_for_public_key(&pk).await?;` is done? only if it fails?

adrianbenavides: Ok I think got it. It breaks to avoid continuing with the below `cfg_if!` code block right?

implementations/rust/ockam/ockam_vault/src/signer_impl.rs
```diff
@@ -47,6 +55,26 @@ impl Signer for Vault {
...
+                    if #[cfg(feature = "rustcrypto")] {
+                        use p256::ecdsa::{self, signature::Signer as _};
+                        use p256::pkcs8::DecodePrivateKey;
+                        let sec = ecdsa::SigningKey::from_pkcs8_der(key).map_err(from_pkcs8)?;
+                        let sig = sec.sign(data);
+                        Ok(Signature::new(sig.to_der().as_bytes().to_vec()))
+                    } else {
+                        compile_error!("NIST P-256 requires feature `rustcrypto`")
+                    }
+                }
```
adrianbenavides, line 76: A general question about this pattern. Couldn't this be written as follows?

```rust
cfg_if! {
    if #[cfg(feature = "aws")] { ... }
    else if #[cfg(feature = "rustcrypto")] { ... }
    else { ... }
}
```

adrianbenavides (commented):
Looks good to me. I initially thought the internals would be more obscure but it looks pretty straightforward. I left a couple of questions about syntax and a suggestion regarding the potential changes on `ockam_command`.

implementations/rust/ockam/ockam_vault/src/aws.rs
twittner: NB that `KeyId` is not a type but just an alias for `String`. This is why I felt I could reuse it here because it still is some sort of identifier of a key.

implementations/rust/ockam/ockam_command/src/identity/create.rs
twittner: I was not at all concerned with the command-line interface. In fact I closed the previous PR because of the number of changes on the CLI layer that happened in the meantime. What you find in here is the bare minimum to have some workable example and I absolutely agree that it can and should be reworked to fit better into the intended CLI.

implementations/rust/ockam/ockam_vault/src/signer_impl.rs
twittner: No, because this would not make it possible to fall back to the `rustcrypto` block from the `aws` block in case the `Secret` is not a `Secret::Aws` or because the vault is not configured for AWS. Generally, the `cfg_if!` tests are static (happen at compile time) and here we have to make some dynamic (at runtime) decisions too.

adrianbenavides: Ok I understand, I knew I was missing something but couldn't see what. Thanks!

implementations/rust/ockam/ockam_command/src/identity/create.rs
polvorin: :+1:  lets ticket it up @adrianbenavides

implementations/rust/ockam/ockam_vault/src/aws.rs
polvorin: Hum I see,   somehow I thought using a plain string where a KeyId was specified was not going to work,   but you are right is just an alias. 
I still feel it a bit confusing,   but can live with it.    Maybe in some future iteration we make the KeyId a newtype and so KMSKeyId ,   or  at least make another alias  with a different name    :man_shrugging:

polvorin (dismissed):
I'd be in favor of merging it,   and we can tweak it after that

implementations/rust/ockam/ockam_command/src/identity/create.rs
adrianbenavides: For reference: https://github.com/build-trust/codebase/issues/493

adrianbenavides:
@mergifyio rebase

adrianbenavides (approved):

PR #3872, opened 2022-11-30
fix(rust): cli state dir used on tests on macos

adrianbenavides:
Using the default temp directory on macos doesn't work. It looks like the OS eagerly moves the files around before the test finishes.

This is solved by creating the temp dir under `~/.ockam/tests`.


polvorin (dismissed):

adrianbenavides:
@mergifyio rebase

SanjoDeundiak (approved):
Didn't review the changes, but test is now passing on my MacOS machine

PR #3877, opened 2022-12-02
refactor(rust): cleanup `node create` arguments

adrianbenavides:
- Removes `kms` related arguments
- Removes other unused arguments


implementations/rust/ockam/ockam_command/tests/commands.bats
```diff
@@ -83,6 +83,19 @@ teardown() {
...
 }
 
+@test "node is restarted with default services" {
+  # Create node, check that it has one of the default services running
+  run $OCKAM node create n1
+  assert_success
+  assert_output --partial "/service/vault_service"
+
+  # Stop node, restart it, and check that the service is up again
+  $OCKAM node stop n1
```
polvorin, line 93: would `$OCKAM node create n1`   work? .  

I'm not sure about the difference between  create and start.     At the end what we need is something that can either start from scratch if there is no data for that node (no `node/n1`)  _or_ pick up from existing data if there is already (there already exists a `node/n1`).

polvorin (dismissed):

implementations/rust/ockam/ockam_command/tests/commands.bats
adrianbenavides: > would $OCKAM node create n1 work? .

no, because it already exists, i.e. we have a directory with the node's state.

> I'm not sure about the difference between create and start. At the end what we need is something that can either start from scratch if there is no data for that node (no node/n1) or pick up from existing data if there is already (there already exists a node/n1).

I'm not sure I understand this part. We have `create` to start from scratch and `start` to pick up from existing data. Are you proposing to merge both commands?

adrianbenavides:
@mergifyio rebase

adrianbenavides:
@mergifyio rebase

implementations/rust/ockam/ockam_command/tests/commands.bats
polvorin: Not here and now.    But we need a way to ,  using *exactly the same command line command*  to start a node creating all folders,  data-files etc if it doesn't exists  (current  `create`)   or pick it up from existing data (current `start`,   haven't tested that part to confirm).  
This *could*  be a bash script that do one or the other depending if it's the first time it's run.   But this script needs to be included on the ockam'  docker image  (that's basically the issue:   we restart ockam' containers and they need to pick up from where they where)

polvorin (approved):

PR #3927, opened 2022-12-15
feat(rust): add unix domain sockets transport and getting-started example

caoakleyii:
## Current Behavior

Currently there is no implementation of Unix Domain Socket Transports.

## Proposed Changes

This PR will add functionality to listen, connect, send and receive messages over UDS within Ockam Routing.

https://github.com/build-trust/codebase/issues/474

## Checks

### DIFF
--- a/examples/rust/get_started/examples/04-unix-domain-socket-transport-responder.rs
+++ b/examples/rust/get_started/examples/04-unix-domain-socket-transport-responder.rs
@@ -1,0 +1,21 @@
+// This node starts a uds listener and an echoer worker.
+// It then runs forever waiting for messages.
+
+use hello_ockam::Echoer;
+use ockam::{Context, Result};
+use ockam_transport_uds::UdsTransport;
+
+#[ockam::node]
+async fn main(ctx: Context) -> Result<()> {
+    // Initialize the Uds Transport.
+    let uds = UdsTransport::create(&ctx).await?;
+
+    // Create a Uds listener and wait for incoming connections.
+    uds.listen("/tmp/ockam-example-echoer").await?;
+
+    // Create an echoer worker
+    ctx.start_worker("echoer", Echoer).await?;
+
+    // Don't call ctx.stop() here so this node runs forever.
+    Ok(())
+}
--- a/examples/rust/get_started/examples/04-unix-domain-socket-transport-responder.rs
+++ b/examples/rust/get_started/examples/04-unix-domain-socket-transport-responder.rs
@@ -2,7 +2,7 @@
 // It then runs forever waiting for messages.
 
 use hello_ockam::Echoer;
-use ockam::{Context, Result};
+use ockam::{access_control::AllowAll, Context, Result};
 use ockam_transport_uds::UdsTransport;
 
 #[ockam::node]
@@ -14,7 +14,7 @@ async fn main(ctx: Context) -> Result<()> {
     uds.listen("/tmp/ockam-example-echoer").await?;
 
     // Create an echoer worker
-    ctx.start_worker("echoer", Echoer).await?;
+    ctx.start_worker("echoer", Echoer, AllowAll, AllowAll).await?;
 
     // Don't call ctx.stop() here so this node runs forever.
     Ok(())
--- a/implementations/rust/ockam/ockam_transport_uds/src/router/handle.rs
+++ b/implementations/rust/ockam/ockam_transport_uds/src/router/handle.rs
@@ -1,7 +1,7 @@
 use std::os::unix::net::SocketAddr;
 
 use ockam_core::{
-    async_trait, compat::sync::Arc, Address, AsyncTryClone, Mailbox, Mailboxes, Result,
+    async_trait, compat::sync::Arc, Address, AsyncTryClone, DenyAll, Mailbox, Mailboxes, Result,
 };
 use ockam_node::Context;
 use ockam_transport_core::TransportError;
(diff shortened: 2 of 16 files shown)


polvorin (dismissed):
This looks good to me, will like to have some :eyes: from somebody with more background on this part (rust transports) as well. 

I was hoping we could factor out more common code between this and the tcp transport,  but it's ok.

examples/rust/get_started/examples/04-unix-domain-socket-transport-responder.rs
```diff
@@ -0,0 +1,21 @@
...
+use ockam::{Context, Result};
+use ockam_transport_uds::UdsTransport;
+
+#[ockam::node]
+async fn main(ctx: Context) -> Result<()> {
+    // Initialize the Uds Transport.
+    let uds = UdsTransport::create(&ctx).await?;
+
+    // Create a Uds listener and wait for incoming connections.
+    uds.listen("/tmp/ockam-example-echoer").await?;
```
SanjoDeundiak, line 14: FYI: we should prefer first setting up everything on our node (in this case - starting an `Echoer`) and only then start listening for incoming connections. But since other examples have it wrong, this one should stay wrong as well :D

implementations/rust/ockam/ockam_transport_uds/src/router/handle.rs
```diff
@@ -0,0 +1,176 @@
...
+    api_addr: Address,
+}
+
+#[async_trait]
+impl AsyncTryClone for UdsRouterHandle {
+    async fn async_try_clone(&self) -> Result<Self> {
+        // TODO @ac 0#UdsRouterHandle.async_try_clone.detached
+        // in:  n/a - DenyAll?
+        // out: n/a - DenyAll?
+        let mailboxes = Mailboxes::new(
```
SanjoDeundiak, line 29: Should be DenyAll for incoming and outgoing

implementations/rust/ockam/ockam_transport_uds/src/router/uds_router.rs
```diff
@@ -0,0 +1,319 @@
...
+    api_addr: Address,
+    map: BTreeMap<Address, Address>,
+    allow_auto_connection: bool,
+}
+
+/// Public Implementaions to instantiate a UDS Router and UDS Router Handler
+impl UdsRouter {
+    /// Create and register a new UDS router with the given node context
+    pub async fn register(ctx: &Context) -> Result<UdsRouterHandle> {
+        // TODO: Perhaps should not use random_ragged, it uses `LOCAL` as the address
```
SanjoDeundiak: `random_tagged` generates random `LOCAL` addresses, if that was your concern

implementations/rust/ockam/ockam_transport_uds/src/router/handle.rs
SanjoDeundiak: Could you please refer to `tcp` crate and see what are the access controls in similar cases?

implementations/rust/ockam/ockam_transport_uds/src/workers/receiver.rs
```diff
@@ -0,0 +1,106 @@
...
+
+        // Insert the peer address into the return route so that
+        // reply routing can be properly resolved
+        msg.return_route.modify().prepend(self.peer_addr.clone());
+
+        trace!("Message onward route: {}", msg.onward_route);
+        trace!("Message return route: {}", msg.return_route);
+
+        // Mark that message originates from some other node
+        let local_info = ExternalLocalInfo::new(UDS).to_local_info()?;
```
SanjoDeundiak: This was changed recently, we no longer carry `ExternalLocalInfo`

implementations/rust/ockam/ockam_transport_uds/src/workers/sender.rs
```diff
@@ -0,0 +1,322 @@
...
+        ctx: &Context,
+        router_handle: UdsRouterHandle,
+        stream: Option<UnixStream>,
+        peer: SocketAddr,
+        hostnames: Vec<String>,
+    ) -> Result<WorkerPair> {
+        trace!("Creating new UDS worker pair");
+        let (mut worker, pair) = Self::new_pair(router_handle, stream, peer, hostnames).await?;
+
+        // TODO: @uds (Oakley) determine why his is considered bad in equivalent usage found here:
```
SanjoDeundiak: Tcp crate now generates `rx_addr` in `new_pair`, this way there is only one place where this address is generated, which leads to a little bit cleaner code. Could you please also make that change for uds?

implementations/rust/ockam/ockam_transport_uds/src/workers/sender.rs
```diff
@@ -0,0 +1,322 @@
...
+                    debug!(addr = %path_display, err = %e, "Failed to connect");
+                    self.stop_and_unregister(ctx).await?;
+
+                    return Err(TransportError::from(e).into());
+                }
+            };
+
+            let sock = SockRef::from(&connection);
+
+            // TODO: @uds enable an automatic keep alive system
```
SanjoDeundiak: What is the conclusion from this? Should we enable call `set_keepalive`?

SanjoDeundiak (changes requested):
Few questions, but overall looks great

caoakleyii:
@SanjoDeundiak  Thanks for the feedback, this pretty much exactly what I was working on addressing.

I'll have this PR updated shortly with those fixes.

implementations/rust/ockam/ockam_transport_uds/src/workers/sender.rs
caoakleyii: So my understanding on this is that socket2, only has the setting  to enable the Unix Domain Socket to allow receive keepalive packets.

However, unlike it's TCP counterpart, it does not have an internal system to automatically send those packets and keep it alive.

I wasn't able to find an existing library that enabled this. So we would need to write a small implementation of this if we wanted it.

However, because the sockets are local to a machine, and less likely to have any issues (e.g. network inconsistencies) I think this low priority.

caoakleyii: Longstory short, I don't see a reason why we should disable keepalive packets.

However, there won't be any we're sending through this existing implementation.

examples/rust/get_started/examples/04-unix-domain-socket-transport-responder.rs
caoakleyii: Noted! 
Is there a specific reason behind this? (Outside of just a more logical sequence and potential race conditions.)

SanjoDeundiak: I think race conditions is the main reason

SanjoDeundiak (approved):

caoakleyii:
@mrinalwadhwa When you're available to merge this, I'd appreciate it! Thank you.

Issue #3906, opened 2022-12-13
add optional `--identity` argument to `secure-channel create` and modify implementation on `ockam_api`


adrianbenavides:
## Desired

Add a new optional `--identity` argument to the `secure-channel create` command. This argument must be of type `Option<String>` (the identity name) and it will be sent "as is" to the node as part of the request, which must be modified to include this new field.

Additionally, the request handler on `ockam_api` must be refactored so it prioritizes the passed optional `Identity` if available, or default to the `self.identity()` if not.

---

_We love helping new contributors!_
_If you have questions or need help as you work on your first Ockam contribution, please leave a comment on [this discussion](https://github.com/ockam-network/ockam/discussions/1081)._
_If you're looking for other issues to contribute to, checkout [this discussion](https://github.com/build-trust/ockam/discussions/3206) and labels -  https://github.com/build-trust/ockam/labels/good%20first%20issue or https://github.com/build-trust/ockam/labels/help%20wanted_

Mo-Fatah:
I can work on this

mrinalwadhwa:
@Mo-Fatah awesome, all yours.

Mo-Fatah:
@adrianbenavides here 
> and it will be sent "as is" to the node as part of the request, which must be modified to include this new field.

I assume by the "request" you mean the [`api::create_secure_channel`](https://github.com/build-trust/ockam/blob/b9861eaaf1de7a73117c9cae18ede8546767d0bb/implementations/rust/ockam/ockam_command/src/secure_channel/create.rs#L166), right ?

Also, the [request handler](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/nodes/models/secure_channel.rs#L39) on `ockam_api`  (correct me if I lost my way) , I couldn't find where exactly should I use the `identity` option from the request

adrianbenavides:
> I assume by the "request" you mean the [`api::create_secure_channel`](https://github.com/build-trust/ockam/blob/b9861eaaf1de7a73117c9cae18ede8546767d0bb/implementations/rust/ockam/ockam_command/src/secure_channel/create.rs#L166), right ?

[...]

PR #3930, opened 2022-12-16
feat(rust): store vault name on idt config (cli state)

adrianbenavides:
This will be needed to complete the issues related to adding the optional `--identity` to some commands (ex: https://github.com/build-trust/ockam/issues/3906). With this information, we will be able to retrieve an identity without forcing the user to provide a vault.

I've also added a method to read the authorities config from the node's state, which will be needed for the credentials commands issues.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state.rs
@@ -1,3 +1,4 @@
+use crate::config::cli::AuthoritiesConfig;
 use crate::nodes::models::transport::{CreateTransportJson, TransportMode, TransportType};
 use ockam_identity::change_history::{IdentityChangeHistory, IdentityHistoryComparison};
 use ockam_identity::{Identity, IdentityIdentifier};
@@ -171,6 +172,12 @@ pub struct VaultState {
     pub config: VaultConfig,
 }
 
+impl VaultState {
+    pub fn name(&self) -> Result<String> {
+        file_stem(&std::fs::canonicalize(&self.path)?)
+    }
+}
+
 #[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
 #[serde(untagged)]
 pub enum VaultConfig {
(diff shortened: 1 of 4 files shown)


mrinalwadhwa (dismissed):

implementations/rust/ockam/ockam_api/src/cli_state.rs
```diff
@@ -523,6 +513,13 @@ impl NodeState {
         self.path.join("policies_storage.lmdb")
     }
 
+    pub fn authorities(&self) -> Result<AuthoritiesConfig> {
```
polvorin, line 516: this is ok now.
In a next iteration we can re visit how this authority file is used and for what,  given than the same info is present on the project' information  (project.json right now).

polvorin:
I'll prefer to not store the vault location as part of the identity file unless we reach a point of being stuck if not,  but I think we haven't found such situation so far.  
It introduces ties between the identity file and vault locations, _but_ the identity file is something that can be used _without any vault or secret key_  and maybe even from different machines than the ones that created them (for example, just for verifying a signature).    

For example for the `secure-channel create`  case,   I'd prefer to:
- Use default vault if one is not given
- Use the one provided otherwise
- Fail if the needed key isn't present on the vault
 
rather than storing the vault path on the identity file. 

Yes,  this put on the user the hurdle of remembering in which vault is the secret key for some identity.  But I think this is ok:

- If you have > than one vault,  and explicitly create an identity on some non-default vault,  is because you have a particular need to solve and have good understanding of the "vault" concept.   So requiring you to give that info do not present a difficulty/learning curve.
- For most users, on most cases, there will be just one vault, the default one,  so you don't need to specify it.  All our scenarios so far fails in this case.
- For _reading_   we _might_  at some point use a "union"   vault that would be the union of a set of underling vaults.   In that case,  you never specify the vault when _using_  an identity  (as the keys will be looked up on all underling vaults on the union).   You only need to specify the vault when _creating_ the identity, to know where to store the secret key.   Note  this is not yet decided, we need to thing a bit more and see the pros and cons of the approach.

adrianbenavides:
Thanks for the detailed response @polvorin 🙏 . If we agree on failing on those cases, we can close this PR. I created it preemptively as one of the potential requirements for some of the community issues I've created today.

I can update the issues' description with this information, explaining how to retrieve the identity and what should happen if it can't be retrieved from the default vault. For now, I'd only add the `identity` argument and leave the `vault` for a future change so we can better plan how to handle that logic. Otherwise, we'll probably end up duplicating this logic in a lot of commands.

mrinalwadhwa:
marking this as draft for now. please close it in case we no longer need it.

adrianbenavides:
I'll close it for now 👍

PR #3977, opened 2022-12-25
feat(rust): added optional `identity_name` field to `createproject`

### LINKED ISSUE #3905: `command/api`: add field to `CreateProject` request to pass an optional identity name
Depends on: https://github.com/build-trust/ockam/issues/3904

## Desired

Add a field called `identity_name` of type `<Option<String>` to the `CreateProject` struct. Then, on the `ockam_api/src/cloud/project::create_project` function replace the hardcoded `None` passed to the `request_controller` function with an `identity` build from the  `identity_name` field of the request using the `CliState`.

---

nidnogg:
## Current Behavior

`CreateProject` has no default name for identities.

## Proposed Changes

This PR solves #3905 by adding the requested `identity_name` field to `CreateProject` requests, as well as by adding the `ident` argument to its corresponding `request_controller`. 

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cloud/project.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/project.rs
@@ -364,25 +364,25 @@ mod node {
 
             let req_builder = Request::post(format!("/v0/{space_id}")).body(&req_body);
             let cli_state = cli_state::CliState::new()?;
-            let inner = self.get().read().await;
-            let ident = match &req_body.identity_name {
-                Some(existing_identity_name) => {
-                    let identity_cfg = cli_state
-                        .identities
-                        .get(existing_identity_name.as_ref())?
-                        .config;
-                    match identity_cfg.get(ctx, inner.vault()?).await {
-                        Ok(idt) => idt,
-                        Err(_) => {
-                            let vault_cfg = cli_state.vaults.default()?.config;
-                            identity_cfg.get(ctx, &vault_cfg.get().await?).await?
+            let ident = {
+                let inner = self.get().read().await;
+                match &req_body.identity_name {
+                    Some(existing_identity_name) => {
+                        let identity_cfg = cli_state
+                            .identities
+                            .get(existing_identity_name.as_ref())?
+                            .config;
+                        match identity_cfg.get(ctx, inner.vault()?).await {
+                            Ok(idt) => idt,
+                            Err(_) => {
+                                let vault_cfg = cli_state.vaults.default()?.config;
+                                identity_cfg.get(ctx, &vault_cfg.get().await?).await?
+                            }
                         }
                     }
+                    None => inner.identity()?.async_try_clone().await?,
                 }
-                None => inner.identity()?.async_try_clone().await?,
             };
-
-            drop(inner);
 
             self.request_controller(
                 ctx,
@@ -409,9 +409,10 @@ mod node {
 
             let req_builder = Request::get("/v0");
 
-            let inner = self.get().read().await;
-            let ident = inner.identity()?.async_try_clone().await?;
-            drop(inner);
+            let ident = {
+                let inner = self.get().read().await;
+                inner.identity()?.async_try_clone().await?
+            };
 
             self.request_controller(
                 ctx,
@@ -439,9 +440,10 @@ mod node {
 
             let req_builder = Request::get(format!("/v0/{project_id}"));
 
-            let inner = self.get().read().await;
-            let ident = inner.identity()?.async_try_clone().await?;
-            drop(inner);
+            let ident = {
+                let inner = self.get().read().await;
+                inner.identity()?.async_try_clone().await?
+            };
 
             self.request_controller(
                 ctx,
@@ -470,9 +472,10 @@ mod node {
 
             let req_builder = Request::delete(format!("/v0/{space_id}/{project_id}"));
 
-            let inner = self.get().read().await;
-            let ident = inner.identity()?.async_try_clone().await?;
-            drop(inner);
+            let ident = {
+                let inner = self.get().read().await;
+                inner.identity()?.async_try_clone().await?
+            };
 
             self.request_controller(
                 ctx,
@@ -501,9 +504,10 @@ mod node {
 
             let req_builder = Request::post(format!("/v0/{project_id}/enrollers")).body(req_body);
 
-            let inner = self.get().read().await;
-            let ident = inner.identity()?.async_try_clone().await?;
-            drop(inner);
+            let ident = {
+                let inner = self.get().read().await;
+                inner.identity()?.async_try_clone().await?
+            };
 
             self.request_controller(
                 ctx,
@@ -531,9 +535,10 @@ mod node {
 
             let req_builder = Request::get(format!("/v0/{project_id}/enrollers"));
 
-            let inner = self.get().read().await;
-            let ident = inner.identity()?.async_try_clone().await?;
-            drop(inner);
+            let ident = {
+                let inner = self.get().read().await;
+                inner.identity()?.async_try_clone().await?
+            };
 
             self.request_controller(
                 ctx,
@@ -563,9 +568,10 @@ mod node {
             let req_builder =
                 Request::delete(format!("/v0/{project_id}/enrollers/{enroller_identity_id}"));
 
-            let inner = self.get().read().await;
-            let ident = inner.identity()?.async_try_clone().await?;
-            drop(inner);
+            let ident = {
+                let inner = self.get().read().await;
+                inner.identity()?.async_try_clone().await?
+            };
 
             self.request_controller(
                 ctx,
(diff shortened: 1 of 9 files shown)


implementations/rust/ockam/ockam_api/src/cloud/project.rs
```diff
@@ -211,7 +211,8 @@ pub struct CreateProject<'a> {
     #[b(1)] pub name: CowStr<'a>,
     #[b(2)] pub services: Vec<CowStr<'a>>,
     #[b(3)] pub users: Vec<CowStr<'a>>,
-    #[b(4)] pub enforce_credentials: Option<bool>
+    #[b(4)] pub enforce_credentials: Option<bool>,
+    #[b(5)] pub identity_name: Option<String>
```
adrianbenavides: You can use a `CowStr<'a>` instead of a `String` here. Sorry for not specifying this correctly in the issue description

implementations/rust/ockam/ockam_api/src/cloud/project.rs
```diff
@@ -220,12 +221,14 @@ impl<'a> CreateProject<'a> {
         enforce_credentials: Option<bool>,
         users: &'a [T],
         services: &'a [T],
+        identity_name: Option<String>,
```
adrianbenavides: You will have to change this to something like `Option<&'a T>`.

implementations/rust/ockam/ockam_api/src/cloud/project.rs
```diff
@@ -301,14 +305,26 @@ mod node {
...
+
+            let default_vault = cli_state.vaults.default()?.config.get().await?;
+            let ident = Some(
+                cli_state
+                    .identities
+                    .default()?
+                    .config
+                    .get(ctx, &default_vault)
+                    .await?,
+            );
```
adrianbenavides: This part is a little bit tricky to understand and it's not doing what we need to. At this point, the `req_body` will have an optional identity name:
- If `None`, we want to use the node's default identity
- If `Some`, we want to load that identity using the node's vault and if it fails, fallback to the default vault and rise an error if the identity can't be loaded then

Now, to access the node's vault/identity you can do the following:
```rust
let inner = self.get().read().await;
let vault = inner.vault()?;
let identity = inner.identity()?;
```

Let me know if you have any questions on this! 👍

adrianbenavides (changes requested):
Thanks for submitting this PR @nidnogg ! I've left a couple of comments, let me know if I can help you with anything.

Happy new year! ☃️

implementations/rust/ockam/ockam_api/src/cloud/project.rs
nidnogg: No worries, fixing it in the upcoming changes!

implementations/rust/ockam/ockam_api/src/cloud/project.rs
nidnogg: Gotcha, I might have some time to do these tomorrow night. I'll beep you back if I run into any trouble!

implementations/rust/ockam/ockam_api/src/cloud/project.rs
nidnogg: Shouldn't it be `Option<CowStr<'a>>` like in the previous comment as well? The compiler keeps giving an error that these two should match (i.e. lines 215 and 224).

implementations/rust/ockam/ockam_api/src/cloud/project.rs
nidnogg: Thanks for chipping in! I've tried my hand at it but I'm not quite understanding a few steps:
1. Just to confirm - if `None` is a match for `req_body`'s identity name then I can use the `default_vault` approach like it's been currently done in lines 310~318, right?

2. From this tidbit: 
```rust
let inner = let inner = self.get().read().await;
let vault = inner.vault()?;
let identity = inner.identity()?;
```
When you say load that identity from `red_body` into the node's vault, do you mean just using `inner.identity()?` I'm confused as to how you would associate that identity with a vault, having just an identity name. 

3. If it is a matter of running `inner.identity()`, that function seems to return a borrowed value, rendering me unable to pass it onto the  `request_controller` afterwards. I couldn't find a clone method implemented for getting that value, should that be implemented?

implementations/rust/ockam/ockam_api/src/cloud/project.rs
adrianbenavides: Oh, sorry, it should be `Option<S>`, I got confused with the traits of the method definition.

Then you can transform that into a `Option<Cow<str>>` with `identity_name.map(|x| x.into())`.

implementations/rust/ockam/ockam_api/src/cloud/project.rs
adrianbenavides: 1. No, that's the key point here. We want to use the node's default vault, not the global default vault.
2. Check the code below
3. There is a method for this: `async_try_clone`

Since this is a little tricky, here is a possible solution for this:

```rust
let cli_state = cli_state::CliState::new()?;
let inner = self.get().read().await;
let identity = match &req_body.identity_name {
    Some(idt_name) => {
        let idt_cfg = cli_state.identities.get(idt_name)?.config;
        match idt_cfg.get(ctx, inner.vault()?).await {
            Ok(idt) => idt,
            Err(_) => {
                let vault_cfg = cli_state.vaults.default()?.config;
                idt_cfg.get(ctx, &vault_cfg.get().await?).await?
            }
        }
    }
    None => inner.identity()?.async_try_clone().await?,
};
```

adrianbenavides: One final note: once you have this in place, you should modify the `request_controller` and the `controller_secure_channel` functions to actually use this identity.

You can remove [this block](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/cloud/mod.rs#L152-L155) and pass the identity to replace [this one](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/cloud/mod.rs#L121).

implementations/rust/ockam/ockam_api/src/cloud/project.rs
nidnogg: Should be solved in the latest commit

implementations/rust/ockam/ockam_api/src/cloud/project.rs
nidnogg: Thanks! That worked out fine.

implementations/rust/ockam/ockam_api/src/cloud/project.rs
nidnogg: Thanks a lot for the previous snippet, that was much simpler solution than a series of `if let`s that I was employing at the time and it did help out a lot.

Regarding this last note:
> One final note: once you have this in place, you should modify the `request_controller` and the `controller_secure_channel` functions to actually use this identity.
> 
> You can remove [this block](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/cloud/mod.rs#L152-L155) and pass the identity to replace [this one](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/cloud/mod.rs#L121).

After performing those steps, around 24 calls to the `request_controller()` function that were using `None` as a default `ident` value had to be modified to take in an `Identity<Vault>`.

The only working way I found to get an `Identity<Vault>` in those calls was using the initial tip you provided, like such (as seen in [this snippet](https://github.com/build-trust/ockam/pull/3977/files#diff-ac92631eb44c2b0ea45d6c7ef8bb4dec67639f735f6a46d708a61b483125d2feR90-R92)):

```rust 
let inner = let inner = self.get().read().await;
let ident = inner.identity()?.async_try_clone().await;
drop(inner); // Without this drop the compiler would implode complaining of multiple borrows from inner
```

I'm not sure if this is the intended behavior or if we need that same approach in `CreateProject` for every one of these calls. Let me know if this works!

adrianbenavides: > After performing those steps, around 24 calls to the request_controller() function that were using None as a default ident value had to be modified to take in an Identity<Vault>.

Yeah, that's right, I have to create more issues to modify the other endpoints so we have the same behavior as the `create_project` method.

Regarding the use of `drop(inner)`, just in case this is useful, it happened because you were trying to hold a reference to the lock (`inner`) after the function returned its result. The "clean" way to do that is to create an inner scope to work with the lock, so it can be automatically dropped when we get out of that scope:

```rust
let ident = {
    let inner = self.get().read().await;
    inner.identity()?.async_try_clone().await?
};
```

Let me know if you want to do that change, it can be a helpful exercise to better understand the borrow checker. If not, I can do it myself!

The PR looks great now, I'll merge it as soon as we apply the `drop` changes.

nidnogg: Great to know! I'll apply those ASAP. Thanks again @adrianbenavides!

nidnogg: Solved with latest changes

adrianbenavides (dismissed):
Ready to merge, thank you so much for your contribution @nidnogg!! 🥳

adrianbenavides (approved):

Issue #3575, opened 2022-10-04
Split schema.cddl


twittner:
[`schema.cddl`](https://github.com/build-trust/ockam/blob/127df6de454784813cb69c785faf2e627afa41c5/implementations/rust/ockam/ockam_core/src/schema.cddl) contains schema definitions for various API types. It was originally located in `ockam_api` but as use of API types spread it was moved to `ockam_core` along with supporting infrastructure (e.g. [to assert some CBOR value matches a certain rule](https://github.com/build-trust/ockam/blob/127df6de454784813cb69c785faf2e627afa41c5/implementations/rust/ockam/ockam_core/src/api.rs#L521-L561)). However this single file still contains all rules, even for types outside of `ockam_core`. Ideally each crate has its own schema file, containing only those rules that apply to its own types and not those of other crates.

Assuming those files exist their contents must nevertheless be merged as every API interaction contains a mixture of types from different crates (e.g. `Request` from `ockam_core` and `Project` from `ockam_api`). The existing functions `assert_request_match` and `assert_response_match` could accept an additional parameter, implementing cddl-cat's [`LookupContext`](https://docs.rs/cddl-cat/0.6.1/cddl_cat/context/trait.LookupContext.html) and instead of using `cddl_cat::validate_cbor_bytes` they use [`cddl_cat::validate_cbor`](https://docs.rs/cddl-cat/0.6.1/cddl_cat/fn.validate_cbor.html). Before using those assertion functions a crate needs to merge the contexts of its dependent crates. For example:

## `ockam_core::api`

```rust
pub const SCHEMA: &str = core::include_str!("schema.cddl");

pub fn schema() -> Result<BasicContext, MyValidationError> {
    let cddl = cddl_cat::parse_cddl(SCHEMA)?;
    let cddl = flatten(&cddl)?;
    Ok(BasicContext::new(cddl))
}
```

## `ockam_node::api`

```rust
pub const SCHEMA: &str = core::include_str!("schema.cddl");

pub fn schema() -> Result<BasicContext, MyValidationError> {
    let mut ctx = ockam_core::api::schema()?;
    let cddl = cddl_cat::parse_cddl(SCHEMA)?;
    let cddl = flatten(&cddl)?;
    for (k, v) in cddl {
        match ctx.rules.entry(k) {
            Entry::Vacant(e) => {
                e.insert(v);
            }
            Entry::Occupied(e) => {
                return Err(MyValidationError::Duplicate(e.key().to_string()))
            }
[...]

PR #3980, opened 2022-12-26
add schema fn

phillyphil91:
## Current Behavior

Currently the CDDL schema for various API types is in `ockam_core` independent of whether the API type is actually part of `ockam_core` or any other crate.

## Proposed Changes

Allow schema.cddl files to be placed in other crates and merge them when CBOR API validation is needed.

Addresses issue: #3575 

 
## Checks


phillyphil91:
TODO: 

- [ ] Implement `validate_cbor`
- [ ] Include `serde_cbor::value::Value` to be used in `validate_cbor` instead of &[u8] for `validate_cbor_bytes`

phillyphil91:
Hi @twittner 
I tried to implement everything more or less as suggested in #3575 

A couple of questions:
- There are a lot of `unwrap()` calls at the moment. I am not really familiar with ockam error handling. Where should I propagate these possible errors to?
```rust
pub fn schema() -> BasicContext {
    let cddl = parse_cddl(SCHEMA).unwrap();
    let cddl = flatten(&cddl).unwrap();
    BasicContext::new(cddl)
}
``` 
Same with:
```rust
 for (k, v) in cddl {
        match ctx.rules.entry(k) {
            Entry::Vacant(e) => {
                e.insert(v);
            }
            Entry::Occupied(_e) => {
                todo!()
                // return Err(MyValidationError::Duplicate(e.key().to_string()))
            }
        }
    }
    Ok(ctx)
}
```
Where should `MyValidationError` be placed at?
- Regarding the cfg tag flag. That is just used for cddl related stuff right? I removed if for now, because it was causing some issues, but I can add it again later. For example the `assert_request_match` doesn't do anything when the flag is disabled right? 
- Is `assert_response_match` actually called anywhere?
- All calls to `assert_request_match` pass `None` as the `struct_name` and hence only the `request` (i.e. the header) section of the cddl is checked, correct? No "body" checks are performed anywhere? 
- Regarding the merging of the cddl files. Would it make sense to check if a crate has the file and if not just use the one from `ockam_core` ?

phillyphil91:
Hi @twittner or anyone else.

Any comment or request for this? Would love to continue with it.

adrianbenavides:
> Hi @twittner or anyone else.
> 
> Any comment or request for this? Would love to continue with it.

Hey @phillyphil91, I'll be reviewing it shortly, thanks for keeping your interest in this issue!

implementations/rust/ockam/ockam_node/src/api.rs
```diff
@@ -1,10 +1,34 @@
+#![allow(missing_docs)]
+
 use crate::Context;
+use alloc::collections::btree_map::Entry;
+use cddl_cat::{context::BasicContext, flatten::flatten, parse_cddl};
 use core::fmt::Display;
 use minicbor::Encode;
 use ockam_core::api::{assert_request_match, RequestBuilder};
 use ockam_core::compat::vec::Vec;
 use ockam_core::{Address, LocalInfo, Result, Route};
```
adrianbenavides: > There are a lot of unwrap() calls at the moment. I am not really familiar with ockam error handling. Where should I propagate these possible errors to?

There is a central `Error` type that you can use at `ockam_core`. In this case, you can just import that type and use it from both `ockam_core::api` and `ockam_node::api`, as you are already using the `ockam_core::Result`.

implementations/rust/ockam/ockam_node/src/api.rs
```diff
@@ -1,10 +1,34 @@
...
+    let cddl = parse_cddl(SCHEMA).unwrap();
+    let cddl = flatten(&cddl).unwrap();
+    for (k, v) in cddl {
+        match ctx.rules.entry(k) {
+            Entry::Vacant(e) => {
+                e.insert(v);
+            }
+            Entry::Occupied(_e) => {
+                todo!()
+                // return Err(MyValidationError::Duplicate(e.key().to_string()))
```
adrianbenavides: You can create an ockam error with `Origin::Core` and `Code::AlreadyExists`.

implementations/rust/ockam/ockam_core/src/api.rs
```diff
@@ -523,43 +531,50 @@ impl<T: Encode<()>> ResponseBuilder<T> {
 }
 
 #[allow(unused_variables)]
-pub fn assert_request_match<'a>(struct_name: impl Into<Option<&'a str>>, cbor: &[u8]) {
-    #[cfg(feature = "tag")]
```
adrianbenavides, line 527: > Regarding the cfg tag flag. That is just used for cddl related stuff right? I removed if for now, because it was causing some issues, but I can add it again later. For example the assert_request_match doesn't do anything when the flag is disabled right?

We have a [`tag` field](https://github.com/build-trust/ockam/blob/495c3a04bc20081fc21220f054a1f38d5c199e19/implementations/rust/ockam/ockam_api/src/cloud/addon.rs#L15) on all cbor structs that can be used to identify the struct and it's enabled with that flag.

I assume we have to somehow avoid validating that field when the `tag` feature is disabled.

The best thing would be to have some unit tests to check that the validation works as expected with and without the `flag` feature.

adrianbenavides (commented):
> All calls to assert_request_match pass None as the struct_name and hence only the request (i.e. the header) section of the cddl is checked, correct? No "body" checks are performed anywhere?

Since this feature was half-baked when we started writing the request/response bodies, it makes sense that we didn't enforce these checks, but we would like to add them sooner than later.

> Regarding the merging of the cddl files. Would it make sense to check if a crate has the file and if not just use the one from ockam_core ?

I guess that if a crate doesn't have its own cddl file, then the merge is not needed and can use the ockam_core::api::schema() without modification. Does that make sense to you?

phillyphil91:
thanks for the review @adrianbenavides . I will take a look at your comments and try to implement them

phillyphil91:
Hi @adrianbenavides 
I did some rather big changes. Hopefully it's not over-engineered or too complicated for the use case.
I didn't like the idea of defining a `fn schema()` in multiple files so I only added one in `ockam_core::api`. 
The `fn schema()` takes a valid cddl schema `&str` and converts that into a `BasicContext`.
The `fn merged_cddl()` takes a reference to a slice of valid cddl schema `&str`s and basically calls `fn schema()` on every `&str` and afterwards merges the X number of cddl schemas and returns the final context or an appropriate error (hopefully).

This means any other crate just needs to import the `fn merged_cddl()` and provide it with all valid cddl schema `&str`s it wants to merge. That means for example in `ockam_node::api` we need to remember to bring in the `ockam_core::api::SCHEMA` as well. I could of course have that already baked into `fn merged_cddl()` if we know that it will always be used no matter which other crate adds a cddl schema, but maybe that's not always the case that the core cddl schema is needed.

I provided two tests in core::api to check if `fn merged_cddl()` works as well as if it returned the appropriate error. 

One thing I noticed is that the docs for the Error `Kind::AlreadyExists` was not correct and just the description from `NotFound`. I tried to fix that.

TODO imo:

- [ ] reactivate `#[cfg(feature = "tag")]`
- [x] handle remaining `unwrap()` calls, e.g. in `assert_request_match`

implementations/rust/ockam/ockam_core/src/api.rs
```diff
@@ -13,8 +14,48 @@ use tinyvec::ArrayVec;
 #[cfg(feature = "tag")]
 use crate::TypeTag;
 
+use cddl_cat::{context::BasicContext, flatten::flatten, parse_cddl, validate_cbor};
+
 pub const SCHEMA: &str = core::include_str!("schema.cddl");
 
+pub fn schema(cddl_schema: &str) -> Result<BasicContext> {
```
phillyphil91: i guess this doesn't have to to be `pub`

adrianbenavides:
> I did some rather big changes. Hopefully it's not over-engineered or too complicated for the use case.
I didn't like the idea of defining a fn schema() in multiple files so I only added one in ockam_core::api.
The fn schema() takes a valid cddl schema &str and converts that into a BasicContext.
The fn merged_cddl() takes a reference to a slice of valid cddl schema &strs and basically calls fn schema() on every &str and afterwards merges the X number of cddl schemas and returns the final context or an appropriate error (hopefully).

It's looking great, I think the implementation you propose is easy to use and flexible enough for our current use cases.

> This means any other crate just needs to import the fn merged_cddl() and provide it with all valid cddl schema &strs it wants to merge. That means for example in ockam_node::api we need to remember to bring in the ockam_core::api::SCHEMA as well. I could of course have that already baked into fn merged_cddl() if we know that it will always be used no matter which other crate adds a cddl schema, but maybe that's not always the case that the core cddl schema is needed.

About this, the `ockam_core` schema should just contain the first 52 lines (request/response/error/message). The rest is only used at `ockam_api`. If you do this change, then it would make sense to always load the `ockam_core` schema.

phillyphil91:
hey @adrianbenavides 
i think almost everything should be covered now. Still not really sure what to do about the `#[cfg(feature = "tag")]` flag. should this be added to basically everything I created now?

Also I wasn't sure if the `tracing::error!` usage that I have now included, makes sense. 

Maybe you can give some advice on what to do with that. If there is anything else that needs to be changed still, let me know.

adrianbenavides:
> hey @adrianbenavides i think almost everything should be covered now. Still not really sure what to do about the `#[cfg(feature = "tag")]` flag. should this be added to basically everything I created now?

[What I wrote here](https://github.com/build-trust/ockam/pull/3980#discussion_r1066651411) is still pretty much valid. Note that in the schema.cddl files the `tag` field is always present, so:
- if the `tag` feature is enabled, then the `tag` field must be checked
- but if it's disabled, the `tag` field must be skipped

I think the easiest way to approach this is to write some unit tests checking that the validation methods work as expected with and without tag fields (i.e. the `assert_request_match` and `assert_response_match` methods).

> Also I wasn't sure if the `tracing::error!` usage that I have now included, makes sense.

They make sense to me. Not sure though if any of those branches should bubble up the error instead of just logging it. Again, I think that writing some tests will clarify this.

phillyphil91:
I replaced the calls to `fn assert_response_match` in ockam_api/src/authenticator/direct.rs. 
I was wondering it it makes sense to call `fn merged_cddl` in `fn assert_request_match` and `fn assert_response_match` instead of passing it from the outside, because it's really just used in those functions and it would remove the calls to it in the functions that call `fn assert_request_match`.

Regarding #`[cfg(feature = "tag")]` I reinstated the flag in both `fn assert_request_match` and `fn assert_response_match`. 
It should do everything that it did before now. 
The question that I still have is that the "type tag" field which is enable by "tag" was not checked before was it? Basically if "tag" was not enabled then `fn assert_request_match` did nothing (some for the reponse fn).
Maybe I am misunderstanding that it was there before or is something new that needs to be implemented. 
As the "type tag" field is marked as optional in the cddl file wouldn't it work without the "tag" flag at all? If the field is there, it will be checked and if not then the rest of the request/response will be checked regardlessly. Maybe I am not correctly understanding cddl notation yet. 
Or rather I am wondering if there should be two versions of `fn assert_request_match`. One that checks the request against the cddl if the tag field is there and one if it's not. but then again... it's optional right now anyways.

Sorry for the confusion around this.

adrianbenavides:
> I was wondering it it makes sense to call fn merged_cddl in fn assert_request_match and fn assert_response_match instead of passing it from the outside, because it's really just used in those functions and it would remove the calls to it in the functions that call fn assert_request_match.

Definitely, this could be improved. The core idea behind this PR is to have crate-specific cddl files that can be merged with a base cddl. What would make sense to me then, is to have a statically initialized `BasicContext` per crate, that is, call the `merged_cddl` function only once per crate. This can be done with once_cell ([here's an example](https://github.com/build-trust/ockam/blob/7e62aed2222aaef50c5c3bddb589187d626be0de/implementations/rust/ockam/ockam_abac/src/parser.rs#L13)). If you try this, you will have to change the signature of the function so it receives a `&BasicContext`, and it should be ok since we only need to read it.

> The question that I still have is that the "type tag" field which is enable by "tag" was not checked before was it?

It wasn't checking the "tag" field because we didn't implement it. But we want to have that functionality. We can keep that out of the scope and tackle it on a separate PR if that makes more sense, and focus on polishing what you have right now.

Thank you so much for your efforts on this PR @phillyphil91, it's not been easy to figure out the right approach but it's almost there! We could merge it after fixing the conflicts + the `merge_cddl` improvement.

phillyphil91:
hi @adrianbenavides 
the once_cell approach wasn't immediately clear too me. Basically it creates a static heap allocated value that only gets initialized once and every subsequent call to the function returns a `Result` now that wraps the static reference to the `BasicContext` and doesn't create a new one?

Regardless of my possibly flawed understanding I think the code is now doing what you suggested. Feel free to review.

Regarding the tag field, i would agree to leave this out of the scope for the PR. Maybe in a subsequent one I can take another look at it.

Thanks

adrianbenavides:
> Basically it creates a static heap allocated value that only gets initialized once and every subsequent call to the function returns a Result now that wraps the static reference to the BasicContext and doesn't create a new one?

Exactly that 👍

implementations/rust/ockam/ockam_core/src/api.rs
```diff
@@ -5,15 +5,62 @@ use crate::compat::rand;
...
+const SCHEMA: &str = core::include_str!("schema.cddl");
+
+fn schema(cddl_schema: &str) -> Result<BasicContext> {
+    let cddl = parse_cddl(cddl_schema).map_err(|e| crate::Error::new(Origin::Core, Kind::Io, e))?;
+    let cddl = flatten(&cddl).map_err(|e| crate::Error::new(Origin::Core, Kind::Io, e))?;
+    Ok(BasicContext::new(cddl))
+}
+
+pub fn merged_cddl(cddl_schemas: &[&str]) -> Result<&'static BasicContext> {
+    static INSTANCE: OnceBox<BasicContext> = OnceBox::new();
```
adrianbenavides: This is not exactly how you would use this.

Note from the [example](https://github.com/build-trust/ockam/blob/7e62aed2222aaef50c5c3bddb589187d626be0de/implementations/rust/ockam/ockam_abac/src/parser.rs#L13) I posted on the previous command that the once_cell is used in isolation (behind a function to make it easy to access it).

Since each crate has their schemas and are known at compile time, the idea is that we can create a static reference to the compiled cddl *of each crate*.

If each crate has their own static cddl, you could then rewrite this:

```rust
let cddl_context = merged_cddl(&[])?;
assert_request_match(schema, &buf, cddl_context);
```

to:

```rust
assert_request_match(schema, &buf, cddl());
```

The static cdd, on the `ockam_api` crate for example, can be built as follows:

```rust
fn cddl() -> &'static BasicContext {
    static INSTANCE: OnceBox<BasicContext> = OnceBox::new();
    INSTANCE.get_or_init(|| {
        Box::new(merged_cddl(&[])).unwrap())
    })
}
```

phillyphil91: Thanks for the suggestions. But how would you handle the errors that `merged_cddl` might produce? Because `assert_request_match` only returns (). Should they be outputted via tracing::error! ?

Also can you maybe explain how your suggestion would differ from mine? Because in the end in my approach we would also just get the `&'static BasicContext` assigned to `cddl_context` and then passed to `assert_request_match` in 
```rust
let cddl_context = merged_cddl(&[])?;
assert_request_match(schema, &buf, cddl_context);
```

How does this differ from calling a function that wraps the `Result` output from `merged_cddl` in the `OnceBox` and calling and passing that to `assert_request_match`:
```rust
fn cddl() -> &'static BasicContext {
    static INSTANCE: OnceBox<BasicContext> = OnceBox::new();
    INSTANCE.get_or_init(|| {
        Box::new(merged_cddl(&[])).unwrap())
    })
}
``` 
Also wouldn't the above function return a `&'static Result<BasicContext>` due to merged_cddl"s return value?

If i call my version of merged_cddl (with OnceBox) in ockam_api for example this will create it's own cddl at compile time, just as it would for ockam_core or any other crate that imports `merged_cddl`, no? 
Or is the output of both approaches kind of the same, but you would suggest that we use the approach that you suggested?

Sorry for further delay on this. I think this is the last question that I had and hopefully we can merge this soon. Thanks for all the help thus far @adrianbenavides

adrianbenavides: > But how would you handle the errors that merged_cddl might produce?

We can't handle that error, since the cddl's are compiled as part of the binary. If there is an error with the cddl's we better fail fast, so it's okay to unwrap that error.

> Also can you maybe explain how your suggestion would differ from mine?

There are two differences:

1. Instead of having to merge cddls **everytime** before doing an assert_match with `merged_cddl(&[])?` you can just load the **pre-merged** cddl with `cddl()`. 
2. The second benefit is that you don't have to explicitly write the cddl's you want to merge (the `merged_cddl` argument), since that's constant within a crate.

> Or is the output of both approaches kind of the same, but you would suggest that we use the approach that you suggested?

I think the output should be the same, but my suggestion has the previous benefits.

> Sorry for further delay on this. I think this is the last question that I had and hopefully we can merge this soon.

This conversation has been absolutely necessary to discover a good solution for this issue, so don't be sorry! And thank you for your patience with this 🙏 !

phillyphil91: hi @adrianbenavides thanks. 
That of course makes sense. Not sure why I didn't catch that myself. 
In your solution the check for an existing value in the `OnceBox` happens before the call to `merged_cddl` (and calls `merged_cddl` on the first time), while in my version `merged_cddl` get's called everytime and afterwards `OnceBox` `get_or_init` get's called, leading to unnecessary merging of the cddl everytime. 

I changed it to the solution that you suggested. I am not a big fan of having to define the `cddl` function in every crate now, but i guess that's not too bad considering the advantages of `OnceBox`. 

One thing that I just caught is that the `schema.cddl` file that I split, I accidentally added it to `ockam_node` instead of `ockam_api`. `ockam_node` only needs the types defined in the schema from `ockam_core` I assume. But `ockam_api` needs other ones too so I moved the file to `ockam_api`. I am not sure where I should import the `SCHEMA` and create the `cddl()` function in `ockam_api` so I did in `ockam_api::authenticator::direct` because that's where it's used. 

Tests in `ockam_api::cloud::project::tests::schema::create_project` and other schema tests in `ockam_api` need the `SCHEMA` from the crate to work. Please have a look if it makes sense the way I am bringing the schema into scope here from `ockam_api::authenticator::direct`.

In those test `validate_cbor_bytes` is still used instead of `validate_cbor`. Should this be changed?

And last but not least, there are still some tests failing in `ockam_api` now where I am not entirely sure why this is the case.

Thanks for your help.

P.s. I had to edit the Cargo.toml files a little bit. Hopefully that's okay for now.

adrianbenavides: > I am not sure where I should import the SCHEMA and create the cddl() function in ockam_api so I did in ockam_api::authenticator::direct because that's where it's used.

That should live at `schema.rs`. 

> In those test validate_cbor_bytes is still used instead of validate_cbor. Should this be changed?

Let's keep that as it is for now, just to make sure everything works as before.

> And last but not least, there are still some tests failing in ockam_api now where I am not entirely sure why this is the case.

- You have to uncomment [this line](https://github.com/build-trust/ockam/pull/3980/files#diff-1956a8258c1f4f6fe6f14fe58aace63355c293b7a09192ea689f90fdf888c087R614)
- The [schema tests form ockam_api](https://github.com/build-trust/ockam/blob/f0451afbc6b6c09e92fa9bae36da5ac50cd53d0e/implementations/rust/ockam/ockam_api/tests/schema.rs#L65) should be moved to `ockam_core` because that's were we now have the types that are used in those tests

phillyphil91: [schema test ](https://github.com/build-trust/ockam/blob/f0451afbc6b6c09e92fa9bae36da5ac50cd53d0e/implementations/rust/ockam/ockam_api/tests/schema.rs#L65) from `ockam_api` has been moved to `ockam_core::api` at the bottom as a test module. Not sure if that makes sense, as it's a bit cluttered now. But there is no dedicated test directory in `ockam_core`.

> That should live at schema.rs.

Not sure if you meant `ockam_api/tests/schema.rs `but I figured that does not make sense so I created a new file in `ockam_api`.

Should the `merged_cddl/ cddl` code in `ockam_core::api` also be moved to a `schema.rs` file within the crate to be consistent with `ockam_api`. Or just keep it as is? Same with `ockam_node`

mrinalwadhwa:
@phillyphil91 I just wanted to jump in here and say, thank you for continuing to invest time on this pull request.

implementations/rust/ockam/ockam_core/src/api.rs
adrianbenavides: > [schema test ](https://github.com/build-trust/ockam/blob/f0451afbc6b6c09e92fa9bae36da5ac50cd53d0e/implementations/rust/ockam/ockam_api/tests/schema.rs#L65)from ockam_api has been moved to ockam_core::api at the bottom as a test module. Not sure if that makes sense, as it's a bit cluttered now. But there is no dedicated test directory in ockam_core.

Agree, the current code organization of the `ockam_core/api.rs` looks weird. I wouldn't worry about that now, let's land this PR first and see how it all feels. We can do that refactor later. 

> Not sure if you meant ockam_api/tests/schema.rs but I figured that does not make sense so I created a new file in ockam_api.

Sorry, I didn't have that ambiguity in mind when I wrote that. And yes, I meant creating that new file at `ockam_api/schema.rs`.

> Should the merged_cddl/ cddl code in ockam_core::api also be moved to a schema.rs file within the crate to be consistent with ockam_api. Or just keep it as is? Same with ockam_node

Not sure how to approach this right now, but I can think of a couple of different ways to organize this. We can open a different issue for this refactor and discuss the details there if that makes sense to you.

adrianbenavides:
@phillyphil91 I've rebased `develop` onto this branch and fixed the merge conflicts. 

Before we can merge it, I'd ask you to squash all commits into a single one with a comment that makes the CI happy (you can check the [expected format here](https://github.com/build-trust/ockam/actions/runs/4190609999/jobs/7264194555#step:9:111)). Something like `refactor(rust): ...` would do.

Also, before merging the PR we'll have to fix some problems with how the dependencies are defined in the different Cargo.toml files, but I can help you with that. The problem is that the `cddl` dependency is `std` only, so all the code using that dependency must be behind the `tag` feature flag, so we don't break the ockam's no-std builds.

phillyphil91:
@adrianbenavides I try to embed everything that I did i.e. everything that uses `cddl-cat` into the tag feature. 

I want to avoid having to specicfy the `cfg` flag many times so I grouped code into `mod` where possible and applied `#[cfg(feature = "tag")]` to the entire `mod`.
I'm not sure if there is an easier way to do this, afaik outside of a function there is not easy way to apply a `cfg` flag to multiple lines besides the `mod` approach.

All crates that use `cddl-cat` now have that crate behind the `tag` feature in the respective `Cargo.toml` files. Also when a crate other than `core` enables the feature, it should also enable it for `core`.

Since beforehand `assert_response_match` and `assert_request_match` didn't really do much or have a body either with the `tag` feature disabled, I moved the entire function behind the feature, because otherwise I would have problems if only `cddl()` is behind it. Now the functions are just used when the `tag` feature is enabled otherwise the code skips them.

I'm not super familiar with the INs and OUTs of Cargo.tomal and conditional compilation yet. So hopefully this approach makes some sense. Let me know if there is something that needs to be changed.

I would suggest that I squash the commits as soon as everything is done.

adrianbenavides:
That was fast! Thanks @phillyphil91 :pray:

> I want to avoid having to specicfy the cfg flag many times so I grouped code into mod where possible and applied #[cfg(feature = "tag")] to the entire mod.
>
> I'm not sure if there is an easier way to do this, afaik outside of a function there is not easy way to apply a cfg flag to multiple lines besides the mod approach.

That's perfect. FYI there is a crate to work around this limitation called [`cfg-if`](https://github.com/rust-lang/cfg-if). Feel free to use it if you you think it can simplify some parts of the code, but it's perfectly fine as it is right now.

> All crates that use cddl-cat now have that crate behind the tag feature in the respective Cargo.toml files. Also when a crate other than core enables the feature, it should also enable it for core.
> 
> Since beforehand assert_response_match and assert_request_match didn't really do much or have a body either with the tag feature disabled, I moved the entire function behind the feature, because otherwise I would have problems if only cddl() is behind it. Now the functions are just used when the tag feature is enabled otherwise the code skips them.
> 
> I'm not super familiar with the INs and OUTs of Cargo.tomal and conditional compilation yet. So hopefully this approach makes some sense. Let me know if there is something that needs to be changed.

The things you mentioned are the important things and I could successfully build the project with and without the `tag` feature on `std`, so well done! I'm trying to figure out why it's failing on `no-std` though. To test this part, you can run the following:

```bash
cargo build

# Compile all crates enabling the `tag` feature if available
cargo build --features tag

# Compile a specific crate that uses this flag
cargo build -p ockam_api --features "tag, direct-authenticator" 

# Compile the no-std env to see if some `std` dependency was leaked
cargo build -p ockam --no-default-features --features="alloc, no_std, software_vault" 
```

adrianbenavides:
> I'm trying to figure out why it's failing on no-std though.

There are some new dependencies that should also be optional because they are only available for `std`.

- ockam_api: once_cell should be optional
- ockam_core: serde_cbor should be optional; quickcheck should be declared in the `dev-dependencies` section
- ockam_node: once_cell should be optional

phillyphil91:
added all your suggestions @adrianbenavides .
Compilation works now for me on `std` and `no-std`.

Let me know if there is anything missing

etorreborre:
@phillyphil91 we are almost there :-). Can you please rebase your PR and do some renaming on your commits so that they follow [the convention](https://github.com/build-trust/.github/blob/main/CONTRIBUTING.md#commit-messages)? You need to [format a file](https://github.com/build-trust/ockam/actions/runs/4235238527/jobs/7359332298) as well. Thanks!

etorreborre:
@phillyphil91 unfortunately the last commit is not going to pass because it is a merge commit. You really need to do a rebase.

etorreborre:
And also we have a specific commit message convention that you need to follow: https://github.com/build-trust/.github/blob/main/CONTRIBUTING.md#commit-messages

phillyphil91:
hi @etorreborre 
I'm doing the rebase right now :+1:

etorreborre:
Thanks @phillyphil91 I think you just need to modify the commit message now so that there are no capital cases: 
> split CDDL schema files & merge when cbor api

If you want to keep capitals you need to enclose the words with backticks, like "\`CDDL\`"

etorreborre:
@adrianbenavides just one last check with you since you followed this PR from the start. It is ok to merge now, isn't it?

adrianbenavides:
@etorreborre yes, it's ready now! 👍 

thanks a lot @phillyphil91 for the titanic work you put into this PR, it was great! 💪

adrianbenavides (approved):
LGTM!

davide-baldo (approved):

phillyphil91:
Hi @adrianbenavides 

thanks a lot for all your help on this one. It was a very rewarding issue. 

Looking forward to the next one. 

Also thanks @etorreborre

PR #4018, opened 2023-01-03
feat(rust): add optional --identity argument to secure-channel create

### LINKED ISSUE #3906: add optional `--identity` argument to `secure-channel create` and modify implementation on `ockam_api`

Mo-Fatah:
Closes #3906 
## Proposed Changes

in `ockam_command` : add optional --identity argument to secure-channel create. The `Option<String>` argument is sent to `ockam_api` in the request. 

in `ockam_api`: the `create_secure_channel_impl` function checks if the `identity` argument is provided, otherwise use `self.identity()`

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
@@ -1,6 +1,7 @@
 use std::time::Duration;
 
 use super::{map_multiaddr_err, NodeManagerWorker};
+use crate::cli_state::CliState;
 use crate::error::ApiError;
 use crate::nodes::models::secure_channel::{
     CreateSecureChannelListenerRequest, CreateSecureChannelRequest, CreateSecureChannelResponse,
@@ -14,9 +15,10 @@ use minicbor::Decoder;
 use ockam::identity::TrustEveryonePolicy;
 use ockam::{Address, Result, Route};
 use ockam_core::api::{Request, Response, ResponseBuilder};
-use ockam_core::{route, AsyncTryClone};
+use ockam_core::{route, AsyncTryClone, CowStr};
 use ockam_identity::{Identity, IdentityIdentifier, TrustMultiIdentifiersPolicy};
 use ockam_multiaddr::MultiAddr;
+use ockam_node::Context;
 use ockam_vault::Vault;
 
 impl NodeManager {
@@ -90,8 +92,22 @@ impl NodeManager {
         authorized_identifiers: Option<Vec<IdentityIdentifier>>,
         credential_exchange_mode: CredentialExchangeMode,
         timeout: Option<Duration>,
+        identity_name: Option<CowStr<'_>>,
+        ctx: &Context,
     ) -> Result<Address> {
-        let identity = self.identity()?.async_try_clone().await?;
+        let identity = if let Some(identity) = identity_name {
+            let state = CliState::new()?;
+            let idt_config = state.identities.get(&identity)?.config;
+            match idt_config.get(ctx, self.vault()?).await {
+                Ok(idt) => idt,
+                Err(_) => {
+                    let default_vault = &state.vaults.default()?.config.get().await?;
+                    idt_config.get(ctx, default_vault).await?
+                }
+            }
+        } else {
+            self.identity()?.async_try_clone().await?
+        };
 
         let sc_addr = self
             .create_secure_channel_internal(&identity, sc_route, authorized_identifiers, timeout)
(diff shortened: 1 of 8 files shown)


implementations/rust/ockam/ockam_command/src/util/api.rs
```diff
@@ -87,6 +87,7 @@ pub(crate) fn create_secure_channel(
         addr,
         authorized_identifiers,
         credential_exchange_mode,
+        None,
```
adrianbenavides: Can you check if this function is used elsewhere? If not, you could delete it and build the request in the command directly.

Also, IIUC, I think the new argument will have no effect if you are passing `None` here.

implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
```diff
@@ -246,12 +260,14 @@ impl NodeManagerWorker {
         let route = crate::multiaddr_to_route(&addr)
             .ok_or_else(|| ApiError::generic("Invalid Multiaddr"))?;
 
+        let identity_ctx = identity.map(|id| Some((id, ctx))).unwrap_or(None);
```
adrianbenavides: We shouldn't pair the Context with any variable as it's a pretty broad variable that could be useful in different scenarios. Just pass the two variables separately.

implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
```diff
@@ -90,8 +92,18 @@ impl NodeManager {
         authorized_identifiers: Option<Vec<IdentityIdentifier>>,
         credential_exchange_mode: CredentialExchangeMode,
         timeout: Option<Duration>,
+        identity_ctx: Option<(CowStr<'_>, &Context)>,
```
adrianbenavides: In the context of this function, you might want to rename this to `identity_name` as it could get easily confused with the actual identity that you build in the next line.

implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
```diff
@@ -90,8 +92,18 @@ impl NodeManager {
...
     ) -> Result<Address> {
-        let identity = self.identity()?.async_try_clone().await?;
+        let identity = if let Some((identity, ctx)) = identity_ctx {
+            CliState::new()?
+                .identities
+                .get(&identity)?
+                .config
+                .get(ctx, self.vault()?)
```
adrianbenavides: Just to make sure we can handle more situations, let's do the following.
- Try building the identity with the node's vault, as you have right now, but don't bubble up the error if it fails.
- If it fails, try again with the default vault as a fallback, and if that fails again, then we bubble up the error.

adrianbenavides (changes requested):
Hey @Mo-Fatah , thanks for submitting another PR 🥳

It looks almost ready to me, but I left some comments pointing out some details that could be improved. Let me know if you have any questions.

implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
Mo-Fatah: At first I put both of them separately, but since `create_secure_channel_impl` is called in multiple scopes (e.g. [here](https://github.com/build-trust/ockam/blob/2de6faaac4bc1ee7c7dc06dc914dd7408be11de8/implementations/rust/ockam/ockam_api/src/nodes/service.rs#L349)) this led to various changes to bring `Context` into scope each time `create_secure_channel_impl` is called.
 Anyway, I have separated both of them.

implementations/rust/ockam/ockam_command/src/util/api.rs
Mo-Fatah: Sorry didn't get that, did you mean `create_secure_channel` ? if yes, then yes it is used in two different locations in `ockam_command/src/project/util.rs` [here](https://github.com/build-trust/ockam/blob/2de6faaac4bc1ee7c7dc06dc914dd7408be11de8/implementations/rust/ockam/ockam_command/src/project/util.rs#L111) and [here](https://github.com/build-trust/ockam/blob/2de6faaac4bc1ee7c7dc06dc914dd7408be11de8/implementations/rust/ockam/ockam_command/src/project/util.rs#L111)

adrianbenavides: Yeah, I meant `create_secure_channel` sorry. Now that I see the changes again, let us remove that method and build the request just as you do in the `secure_channel create` command. 

Without the context of this PR, anyone reading the `create_secure_channel` method contents will find it confusing trying to figure out why we are passing None as the identity.

adrianbenavides (dismissed):
Ready to merge, thanks @Mo-Fatah for another great contribution! 🥳

adrianbenavides (approved):

PR #4055, opened 2023-01-11
feat(elixir): identity service with custom vaults

polvorin:
This allow loading specific `identities` from the identity service,  as well as using specific `vaults` when creating signatures and establishing secure channels. 

Examples:

```
    {:ok, identity, identity_id} = Ockam.Identity.get(some_identity_name)
```
load a named identity from the sidecar (if that's the Identity module used)

```
    Ockam.Services.start_service(
      :identity_secure_channel,
      [identity: own_identity, vault_name: vault_name] ++ options
    )
```
Starts an identity secure channel listener,  using `own_identity`  *and*  using the vault named `vault_name` to perform signatures with that identity.

This paves the way to use `kms` vaults from elixir code through sidecars.

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/identity/secure_channel/secure_channel.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/identity/secure_channel/secure_channel.ex
@@ -17,6 +17,7 @@ defmodule Ockam.Identity.SecureChannel do
 
   Options:
   - identity :: binary() | :dynamic - identity to use in spawned channels, :dynamic will generate a new identity
+  - vault_name :: String.t() | nil - vault to use in the channel, nil to use the default one
   - identity_module (optional) :: module() - module to generate dynamic identity
   - encryption_options (optional) :: Keyword.t() - options for Ockam.SecureChannel.Channel
   - address (optional) :: Ockam.Address.t() - listener address
@@ -82,6 +83,7 @@ defmodule Ockam.Identity.SecureChannel do
 
     with {:ok, identity} <- get_identity(handshake_options) do
       new_handshake_options = Keyword.put(handshake_options, :identity, identity)
+
       new_worker_options = Keyword.put(worker_options, :handshake_options, new_handshake_options)
       {:ok, Keyword.put(options, :worker_options, new_worker_options), state}
     end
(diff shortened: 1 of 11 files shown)


implementations/elixir/ockam/ockam/lib/ockam/identity/secure_channel/secure_channel.ex
```diff
@@ -17,6 +17,7 @@ defmodule Ockam.Identity.SecureChannel do
 
   Options:
   - identity :: binary() | :dynamic - identity to use in spawned channels, :dynamic will generate a new identity
+  - vault_name :: String.t() | nil - vault to use in the channel, nil to use the default one. Only available for listeners
```
hairyhum: Why is it only available for listeners?
It is used from initiator code.

implementations/elixir/ockam/ockam/lib/ockam/identity/sidecar.ex
```diff
@@ -39,14 +46,14 @@ defmodule Ockam.Identity.Sidecar do
     end
   end
 
-  @spec create_signature(identity :: t(), auth_hash :: binary()) ::
+  @spec create_signature(vault_name :: String.t() | nil, identity :: t(), auth_hash :: binary()) ::
           {:ok, proof :: proof()} | {:error, reason :: any()}
-  def create_signature(identity, auth_hash) do
+  def create_signature(vault_name, identity, auth_hash) do
```
hairyhum, line 51: Should we use default argument here, if we support `nil` as an argument?

hairyhum:
Looks good. Minor comments regarding `vault_name` argument usage

implementations/elixir/ockam/ockam/lib/ockam/identity/sidecar.ex
polvorin: yes will do it

implementations/elixir/ockam/ockam/lib/ockam/identity/secure_channel/secure_channel.ex
polvorin: hum,  will remove the comment .  I thought I added it only for listener handshake but got a bit lost on the code,  the comment probably got stale

implementations/elixir/ockam/ockam/lib/ockam/identity/sidecar.ex
polvorin: Did it on the Identity module   that act as the api in front of sidecar/stub implementations

mszpakowski (approved):
Looks good as far as I can tell

adrianbenavides (approved):
Approving to trigger the merge 🚂

PR #4059, opened 2023-01-12
fix(rust): fix vault locking on nfs volumes

polvorin:
the lock file must be opened with read permissions in order to obtain a shared lock on it.  Write permission is not enough.  This was working on local filesystem but failed on attached volumes on the cloud.


polvorin:
@Mergifyio rebase

adrianbenavides (approved):

PR #4070, opened 2023-01-13
feat(elixir): abstraction to implement rest-like endpoints

polvorin:
allow to define routes and handlers, handle path bindings and provide user-implemented authorization rules for endpoints.

Route definitions have the pattern
` {auth_type :: any(), method :: atom(), path :: String.t(), handler :: atom()}` 
  
`auth_type`  is any term.     It will be passed to the `authorize` callback to perform custom authorization on that particular endpoint. 

Example from test cases:
```
defmodule Ockam.Services.API.Tests.EndpointAPI do
  use Ockam.Services.API.Endpoint
  alias Ockam.API.Request

  @impl true
  def routes() do
    [
      {:all, :get, "/", &list/3},
      {:member, :get, "/:id/:role", &show/3},
      {:admin, :put, "/item/:id/:role", &edit/3}
    ]
  end

  def list(_req, %{}, %{}), do: {:ok, "LIST"}
  def show(_req, %{id: id}, %{auth_data: auth_data}), do: {:ok, id <> auth_data}
  def edit(%Request{body: body}, %{id: _id}, %{}), do: {:ok, body}

  # Note: an actual implementation will look at the identity information attached to the request.
  # In this example we just pass a "role" in the url as it's easier to setup the test.
  def authorize(:all, _req, _bindings), do: true
  def authorize(:member, _req, %{id: "a", role: "member"}), do: {true, %{auth_data: "EXTRA"}}
  def authorize(:admin, _req, %{id: "a", role: "admin"}), do: true
  def authorize(_auth_type, _req, _bindings), do: false
end
```

### DIFF
--- a/implementations/elixir/ockam/ockam_services/lib/services/api/endpoint.ex
+++ b/implementations/elixir/ockam/ockam_services/lib/services/api/endpoint.ex
@@ -1,0 +1,132 @@
+defmodule Ockam.Services.API.Endpoint do
+  defmodule DispatchTable do
+    def compile_dispatch_table(rules) do
+      catch_all_rule = {:_, :not_found, %{}}
+
+      case Enum.all?(rules, fn
+             {_auth_type, method, <<"/", _rest::binary>> = _path, handler}
+             when method in [:get, :post, :put, :delete] and
+                    is_function(handler, 3) ->
+               true
+
+             _bad_rule ->
+               false
+           end) do
+        true ->
+          table =
+            Enum.reduce(rules, %{}, fn {auth_type, method, <<"/", _rest::binary>> = path, handler},
+                                       acc ->
+              rule = {path, handler, %{auth_type: auth_type}}
+              Map.update(acc, method, [rule], &[rule | &1])
+            end)
+
+          {:ok,
+           table
+           |> Enum.map(fn {method, rules} ->
+             {method, :cowboy_router.compile([{:_, Enum.reverse([catch_all_rule | rules])}])}
+           end)
+           |> Enum.into(%{})}
+
+        false ->
+          {:error, :bad_dispatch_rules}
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


polvorin:
@Mergifyio rebase

polvorin:
@Mergifyio rebase

implementations/elixir/ockam/ockam_services/lib/services/api/endpoint.ex
```diff
@@ -0,0 +1,109 @@
...
+        {method, :cowboy_router.compile([{:_, Enum.reverse([catch_all_rule | rules])}])}
+      end)
+      |> Enum.into(%{})
+    end
+
+    # Paths *must* start with a "/" (due to how cowboy' routing works)
+    def match(dispatch_table, method, <<"/", _rest::binary>> = path) do
+      case Map.get(dispatch_table, method) do
+        nil ->
+          nil
```
hairyhum: This function returns `{:ok, handler, bindings, auth_type}` or `nil`. I think it would be cleaner to return `:error` instead of `nil`

implementations/elixir/ockam/ockam_services/lib/services/api/endpoint.ex
```diff
@@ -0,0 +1,109 @@
...
+        case handler.(req, bindings, extra_data) do
+          {:ok, body} ->
+            {:reply, :ok, body, state}
+
+          {:error, reason} ->
+            {:error, reason}
+        end
+      end
+
+      defp authorize(_auth_type, _req, _bindings), do: true
```
hairyhum, line 127: I think `authorize` should also be a `@callback`.

hairyhum (commented):
Few code quality comments.

polvorin:
@Mergifyio rebase

PR #4109, opened 2023-01-20
feat(rust): cli project enrollment

polvorin:
augment the existing project authenticate command to support enrollment tokens
```
$ ockam project authenticate -h
Authenticate with a project node

Usage: ockam project authenticate [OPTIONS] --project <PROJECT_JSON_PATH>

Options:
      --project <PROJECT_JSON_PATH>  Project config file
      --okta                         
      --token <ENROLLMENT TOKEN>     
      --identity <IDENTITY>          

```
if `--okta` is given,  it's authenticate using the okta addon.
If `--token`is given,  it authenticate using that enrollment token  (generated by an enroller)

If neither `--okta`  or `--token`  is given,  it just try to retrieve the credential without any further step (for the enrollment workflow where the enroller directly adds the member' identity).

`--identity` let us choose over which identity we are working  (use the default one unless specified)


polvorin:
@Mergifyio rebase

polvorin:
@Mergifyio rebase

PR #4106, opened 2023-01-19
feat(rust): implement vaults delete command

0x00A5:
## Current Behavior

Missing `vault delete` command

## Proposed Changes

Implement `vault delete` command so that vaults can be deleted via `vault delete ${vault_name}` when they are not in use by any existing node.

Fix issue #3936

## Checks

- [x] All commits in this Pull Request are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and Verified by Github.
- [x] All commits in this Pull Request follow the Ockam [commit message convention](https://github.com/build-trust/.github/blob/main/CONTRIBUTING.md#commit-messages).
- [x] I accept the Ockam Community [Code of Conduct](https://github.com/build-trust/.github/blob/main/CODE_OF_CONDUCT.md).
- [x] I have accepted the Ockam [Contributor License Agreement](https://github.com/build-trust/ockam-contributors/blob/main/CLA.md) by adding my Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/blob/main/CONTRIBUTORS.csv) file in a separate pull request to the [build-trust/ockam-contributors](https://github.com/build-trust/ockam-contributors) repository.

PR #4115, opened 2023-01-23
fix(rust): vault deletion logic from `CliState`

adrianbenavides:
When deleting a vault it doesn't have to affect identities, as they can be loaded using any available vault.

Bug introduced at: https://github.com/build-trust/ockam/pull/4106 which caused orchestrator CI tests to fail.

Orchestrator CI ✅ : https://github.com/build-trust/ockam-artifacts/actions/runs/3984269164

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state.rs
@@ -1,21 +1,15 @@
 use crate::nodes::models::transport::{CreateTransportJson, TransportMode, TransportType};
 use ockam_identity::change_history::{IdentityChangeHistory, IdentityHistoryComparison};
 use ockam_identity::{Identity, IdentityIdentifier, SecureChannelRegistry};
-use ockam_node::tokio::fs::File;
 use ockam_vault::{storage::FileStorage, Vault};
 use rand::random;
 use serde::{Deserialize, Serialize};
-use std::alloc::System;
-use std::ffi::OsStr;
 use std::fmt;
 use std::fmt::{Display, Formatter};
-use std::os::unix::ffi::OsStrExt;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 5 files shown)


implementations/rust/ockam/ockam_api/src/cli_state.rs
```diff
@@ -159,22 +155,30 @@ impl VaultsState {
...
-    pub fn delete_by_state(&self, vault: &VaultState) -> Result<()> {
+    pub async fn delete(&self, name: &str) -> Result<()> {
+        // Retrieve vault. If doesn't exist do nothing.
+        let vault_state = match self.get(name) {
+            Ok(v) => v,
+            Err(CliStateError::NotFound(_)) => return Ok(()),
+            Err(e) => return Err(e),
+        };
+
+        // Set default to another vault if it's the default
```
polvorin, line 166: I think we should revisit this  (and even more the identity case https://github.com/build-trust/ockam/blob/d5bfa7ff85363378cdfc0791f1528c16e53f90db/implementations/rust/ockam/ockam_api/src/cli_state.rs#L348).

If users remove the identity that is pointed by as default,   then node will fail the next time it's tried to be run unless a new default is set.    That make sense and is expected,    while changing the default behind the scenes is not.
But we can do that in a separate PR

polvorin (dismissed):
Left a comment about changing default behind the scene,   we can address that in a different PR

polvorin:
@Mergifyio rebase

polvorin (approved):

implementations/rust/ockam/ockam_api/src/cli_state.rs
adrianbenavides: I'll leave it as "resolved" so we can merge this PR. I'll take a look at this shortly, thanks for the review!

PR #4118, opened 2023-01-23
Kafka outlet implementation for Elixir nodes

hairyhum:
Implements kafka interceptor worker to be included in the inlet-outlet route which can parse kafka protocol messages, modify and/or react to them.

### Implementation:

- Add `Ockam.Transport.Portal.Interceptor` with pluggable interceptors for portals protocol
- Add `Ockam.Kafka.Interceptor` implementing portal interceptor to handle kafka protocol messages
- Implement protocol parsing **only for metadata api**. Support API versions 0-12 for metadata.
- Add `Ockam.Kafka.Interceptor.MetadataHandler` handler functions to react to metadata messages
  - `outlet_response/3` - creates dynamic outlets for each broker in metadata response **does not modify the response**
  - `inlet_response/3` - creates dynamic inlets for each broker in metadata response **modifies the response**
- Add support for `ssl` connections in `Ockam.Transport.Portal.OutletWorker`, configured by passing the `ssl: true` option
- Add `:kafka_interceptor` service to `Ockam.Services.Kafka.Provider` (see moduledoc)

### Caveats:

- Protocol parsing/formatting implemented directly without schema or code generation
- Only metadata API is supported at the moment
- Interceptor implemented as a part of ockam_kafka application, which is not necessary as it's not using kafka dependencies
- If interceptor does not support an API version, the message will be forwarded as is without errors.

### Future improvements
- Use schema or codegen to parse/format the protocol
- Move kafka interceptor to a separate application with less dependencies
- Figure out a way to run unit tests for SSL outlets
- ~~Support connection interruption on some critical errors~~
- Support non-localhost endpoints for dynamic inlets

## Checks


implementations/elixir/ockam/ockam/lib/ockam/transport/portal/interceptor.ex
```diff
@@ -0,0 +1,274 @@
...
+  :init_message - message from `Ockam.Session.Spawner`
+  """
+  use Ockam.AsymmetricWorker
+
+  alias Ockam.Message
+  alias Ockam.Router
+  alias Ockam.Transport.Portal.TunnelProtocol
+
+  ## TODO: enforce that on inlet/outlet level
+  @max_payload_size 48 * 1024
```
mszpakowski, line 41: Could you explain why this is the max payload size?

implementations/elixir/ockam/ockam/lib/ockam/transport/portal/interceptor.ex
```diff
@@ -0,0 +1,274 @@
...
+        small when small < @max_payload_size ->
+          [TunnelProtocol.encode({:payload, payload})]
+
+        large ->
+          Enum.map(chunk_payload(large, @max_payload_size - 1), fn payload ->
+            TunnelProtocol.encode({:payload, payload})
+          end)
+      end
+    end)
+  end
```
mszpakowski: Is the actual max payload size that we can handle `48 * 1024` or `48 * 1024 - 1`? Because something like this would be a bit more intuitive:
```elixir
        small when small <= @max_payload_size ->
          [TunnelProtocol.encode({:payload, payload})]

        large ->
          Enum.map(chunk_payload(large, @max_payload_size), fn payload ->
            TunnelProtocol.encode({:payload, payload})
          end)
```

implementations/elixir/ockam/ockam/lib/ockam/transport/portal/outlet_worker.ex
```diff
@@ -40,30 +76,69 @@ defmodule Ockam.Transport.Portal.OutletWorker do
...
+    {:noreply, state}
+  end
+
+  def handle_info({:EXIT, _from, reason}, state) do
+    {:stop, reason, state}
+  end
+
+  @impl true
+  def terminate(reason, %{peer: peer, connected: true} = _state) do
+    Logger.info("Outlet terminate with reason: #{inspect(reason)}, disconnecting")
```
mszpakowski: 💅 
```suggestion
    Logger.info("Outlet terminated with reason: #{inspect(reason)}, disconnecting")
```

implementations/elixir/ockam/ockam/lib/ockam/transport/portal/outlet_worker.ex
```diff
@@ -40,30 +76,69 @@ defmodule Ockam.Transport.Portal.OutletWorker do
...
+    :ok = Router.route(%Message{payload: TunnelProtocol.encode(:disconnect), onward_route: peer})
+  end
+
+  def terminate(reason, _state) do
+    Logger.info("Outlet terminate with reason: #{inspect(reason)}, already disconnected")
```
mszpakowski:
```suggestion
    Logger.info("Outlet terminated with reason: #{inspect(reason)}, already disconnected")
```

implementations/elixir/ockam/ockam_kafka/lib/interceptor/inlet_manager.ex
```diff
@@ -0,0 +1,124 @@
...
+       outlet_prefix: outlet_prefix,
+       inlets: %{}
+     }}
+  end
+
+  def list_inlets(server \\ __MODULE__, timeout \\ 5000) do
+    GenServer.call(server, :list_inlets, timeout)
+  end
+
+  def set_inlets(server \\ __MODULE__, inlets, timeout \\ 5000) when is_list(inlets) do
```
mszpakowski: It's not very clear from the function definition that the `inlets` are actually port offsets here

implementations/elixir/ockam/ockam_kafka/lib/interceptor/kafka_interceptor.ex
```diff
@@ -0,0 +1,334 @@
...
+            case reconstruct_request(old_header, updated_request_content) do
+              {:ok, new_request} -> {:ok, new_request, state}
+              {:error, reason} -> {:error, reason, state}
+            end
+
+          {:error, reason} ->
+            {:error, reason, state}
+        end
+
+      {:error, {:unsupported_api, api, %RequestHeader{} = header0}} ->
```
mszpakowski: Curious, why `header0` and not simply `header`?

implementations/elixir/ockam/ockam_kafka/lib/interceptor/inlet_manager.ex
```diff
@@ -0,0 +1,124 @@
...
+  def handle_call(:list_inlets, _from, %{inlets: inlets} = state) do
+    {:reply, inlets, state}
+  end
+
+  def handle_call(
+        {:set_inlets, requested_inlets},
+        _from,
+        %{allowed_ports: allowed_ports, inlets: inlets} = state
+      ) do
+    case Enum.any?(requested_inlets, fn port_offset -> port_offset > allowed_ports end) do
```
mszpakowski: From the provider inline docs:
```
base_port: optional<integer>, default 9001 - port from which to start dynamic inlets
allowed_ports: optional<integer>, default 20 - number of ports allowed for dynamic inlets (adds to base_port)
```
This does not exactly fit the implementation. The implementation assumes (but doesn't enforce) that the `requested_inlets` will all be positive and therefore in such case no inlet will be started on the base port. If base port is `9001` then we would start on `9002`.

But I think the bigger issue here that we don't check the input, so nothing stops a client to specify requested inlets using negative numbers and 0, so the allowed ports number can be easily exceeded (and any port <= base_port + allowed_ports can be used).

implementations/elixir/ockam/ockam_kafka/test/interceptor/interceptor_test.exs
```diff
@@ -0,0 +1,233 @@
...
+    assert <<size::signed-big-integer-size(32), message::binary-size(size)>> = packet
+    request_header = request_header(1)
+
+    assert {:ok, _header, _size, %MetadataResponse{} = response} =
+             Parser.parse_kafka_response_for_request(request_header, message)
+
+    ## All hosts should be changed to "localhost"
+    Enum.each(response.brokers, fn broker ->
+      assert broker.host == "localhost"
+    end)
```
mszpakowski: This seems to be duplicated later on line 188

implementations/elixir/ockam/ockam/lib/ockam/transport/portal/interceptor.ex
```diff
@@ -0,0 +1,274 @@
...
+  def handle_outer_message(message, state) do
+    handle_message(:outer, message, state)
+  end
+
+  @impl true
+  def handle_inner_message(message, state) do
+    handle_message(:inner, message, state)
+  end
+
+  def handle_message(direction, %Message{} = message, state) do
```
mszpakowski: Just a general comment - I'm curious how do we decide here what are public vs private functions. For example in this module most of the public functions (or even all outside of callback implementations) are internal to the module. Could we make them private, so it's clear that these are not APIs?

implementations/elixir/ockam/ockam_kafka/lib/interceptor/kafka_interceptor.ex
```diff
@@ -0,0 +1,334 @@
...
+                {:error, reason}
+            end
+
+          {:error, reason} ->
+            {:error, reason}
+        end
+
+      {:error, reason} ->
+        {:error, reason}
+    end
```
mszpakowski: No need to change this in this PR, but having `case` nested inside of `case` nested inside of `case` makes the code a bit difficult to follow (similar for `process_kafka_request/2` and `process_kafka_response/2`). There's a refactoring opportunity here for the future

mszpakowski (commented):
Overall looks great. I have some doubts about how we handle requested inlets and allowed ports in the inlet manager, but other than that my comments are mostly cosmetic

implementations/elixir/ockam/ockam/lib/ockam/transport/portal/interceptor.ex
hairyhum: Because of https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_transport_tcp/src/portal/portal_receiver.rs#L9
It's an arbitrary number we chose for our portals.

implementations/elixir/ockam/ockam/lib/ockam/transport/portal/interceptor.ex
hairyhum: I think we can do that. Just need to test it with Rust.

implementations/elixir/ockam/ockam_kafka/lib/interceptor/kafka_interceptor.ex
hairyhum: Because it's header version 0, but it doesn't have any meaning in this context.

implementations/elixir/ockam/ockam_kafka/lib/interceptor/inlet_manager.ex
hairyhum: > The implementation assumes (but doesn't enforce) that the requested_inlets will all be positive and therefore in such case no inlet will be started on the base port

I'm not sure why would you think so. 0 is a totally valid port offset.

> But I think the bigger issue here that we don't check the input, so nothing stops a client to specify requested inlets using negative numbers

WIll add the check.

implementations/elixir/ockam/ockam/lib/ockam/transport/portal/interceptor.ex
hairyhum: Generally it's better to make functions private unless they need to be public (for API or testing reasons), but I tend to forget.

implementations/elixir/ockam/ockam_kafka/lib/interceptor/kafka_interceptor.ex
hairyhum: Yes. I've tried several times, this is the best I could do for now. The main problem was controlling context for things like state.

implementations/elixir/ockam/ockam_kafka/lib/interceptor/inlet_manager.ex
mszpakowski: > I'm not sure why would you think so. 0 is a totally valid port offset.

I agree that 0 can be a valid offset, but if we allow it, then the math with the `port_offset > allowed_ports` condition doesn't check out, because 0..20 is 21. That's why I assumed from the implementation that we wanted offset to be positive, but maybe it's just the condition that is incorrect.

implementations/elixir/ockam/ockam_kafka/lib/interceptor/inlet_manager.ex
```diff
@@ -0,0 +1,127 @@
...
+    {:reply, inlets, state}
+  end
+
+  def handle_call(
+        {:set_inlets, requested_port_offsets},
+        _from,
+        %{allowed_ports: allowed_ports, inlets: inlets} = state
+      ) do
+    case Enum.any?(requested_port_offsets, fn port_offset ->
+           port_offset > allowed_ports or port_offset < 0
```
mszpakowski: Mentioned in this [comment](https://github.com/build-trust/ockam/pull/4118#discussion_r1093495527)

I think this condition should be `port_offset >= allowed_ports or port_offset < 0`, otherwise we will allow `allowed_ports + 1` ports

EDIT: alternatively we can rename `allowed_ports` to something like `max_offset`. I have no preference here

mszpakowski (dismissed):

polvorin (approved):

PR #4105, opened 2023-01-19
feat(rust): add kafka commands to request starting the producer/consumer services

adrianbenavides:
Related issue https://github.com/build-trust/ockam/issues/4012

This PR only adds the commands to request the initialization of the services. The services are not developed yet so I've added a couple of TODO's with the `@confluent` tag.

PR #4120, opened 2023-01-24
refactor(rust): set default_node_name using clap

### LINKED ISSUE #4080: `ockam node start` command: set default value to `node` argument using `Clap` attributes
Currently, the optional `node` field is an `Option<String>`. When is `None` (i.e. the argument is not passed), it defaults to the default node name. This is done [here](https://github.com/build-trust/ockam/blob/7511258aa3785056a7bfeadf22debd6033f43b40/implementations/rust/ockam/ockam_command/src/node/start.rs#L31-L34).

We want to change how the argument is set up so this transformation is done under the hood by `Clap`. This can be done by defining the argument as follows:

```rust
#[arg(default_value_t = default_node_name())]
node_name: String,
```

And modifying the `default_node_name` so it doesn't receive any argument and, instead, build a new `CliState::new()` instance to retrieve the default node name.

More context [in this comment](https://github.com/build-trust/ockam/pull/3948/files#r1056592882).

---

_If you're looking for other issues to contribute to, checkout issues labeled -  https://github.com/build-trust/ockam/labels/good%20first%20issue or https://github.com/build-trust/ockam/labels/help%20wanted_

raysonkoh:
## Current Behavior

Currently the node name is set using `Option<String>` and checking whether it is `None` in which case the code will apply the default node name. 

## Proposed Changes

Since it is possible to extract the default node name from `CliState::new()` instance, I have modified the `default_node_name` method to take in 0 arguments and use it to set the default node name using Clap attributes.

Fixes #4080 

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/util.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/util.rs
@@ -249,13 +249,3 @@ pub fn spawn_node(
 
     Ok(())
 }
-
-/// Retreive the default node name from the global options
-pub fn default_node_name(opts: &CommandGlobalOpts) -> String {
-    opts.state
-        .nodes
-        .default()
-        .unwrap_or_else(|_| panic!("No default Node found"))
-        .config
-        .name
-}
--- a/implementations/rust/ockam/ockam_command/src/node/delete.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/delete.rs
@@ -1,4 +1,5 @@
-use crate::node::util::{default_node_name, delete_all_nodes, delete_node};
+use crate::node::default_node_name;
+use crate::node::util::{delete_all_nodes, delete_node};
 use crate::{help, node::HELP_DETAIL, CommandGlobalOpts};
 use clap::Args;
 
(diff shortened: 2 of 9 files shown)


implementations/rust/ockam/ockam_command/src/node/util.rs
```diff
@@ -250,12 +251,12 @@ pub fn spawn_node(
...
+pub fn default_node_name() -> String {
+    let state = CliState::new().expect("Failed to load CLI state. Try running 'ockam reset'");
+    state
         .nodes
         .default()
-        .unwrap_or_else(|_| panic!("No default Node found"))
-        .config
-        .name
+        .map(|ns| ns.config.name)
+        .unwrap_or_else(|_| "".to_string())
```
adrianbenavides: This doesn't have to change: if there is no default it must panic.

adrianbenavides (changes requested):
Almost ready to merge! I left a small change request.

Thank you so much @raysonkoh for submitting this PR 🙏

implementations/rust/ockam/ockam_command/src/node/util.rs
raysonkoh: Thanks for the review! I tried to panic if there is no default but it seems that the tests are failing because of that. I think this requires more investigation but I suspect that the default config may not be present.

I found that you have implemented a `default_node_name` fn in #4105, hence I updated my code to use your implementation instead. Let me know if it is okay.

adrianbenavides: Oh right, I remember now. It seems to me like a bug from Clap: for some reason, when running any command, the `default_node_name` function is executed (even if the actual command that you are running doesn't use that function).

adrianbenavides (approved):
Merging now, thanks again @raysonkoh 💪

PR #4123, opened 2023-01-24
Polvorin/token lease manager api

polvorin:
Functional wise,  this PR:
* Switch the token manager service to use the[ request-response  protocol  ](https://github.com/build-trust/ockam/blob/develop/implementations/elixir/ockam/ockam/lib/ockam/api/request.ex),  using the [endpoint](https://github.com/build-trust/ockam/blob/develop/implementations/elixir/ockam/ockam_services/lib/services/api/endpoint.ex)  helper
* Leases generated are split by identity:   identity `A` can't see leases generated for identity `B`
* Some changes on service config and startup,  to ease the path of having multiple lease managers on a same node

Simplifications/refactors
* storage and cloud service are simple callbacks,  no separate process nor workers.   Simplifies the testing and debugging,  and was enough at least for the influxdb use case.
*  to simplify deployment,  there is an `in-memory`  storage that keeps leases on memory only (no external DB),  and uses the influxdb api as source-of-truth to reload data in case of restarts.
* test cases added


polvorin:
@Mergifyio rebase

polvorin:
@Mergifyio rebase

PR #4133, opened 2023-01-26
chore(rust): fix no_std example

SanjoDeundiak:
This no_std example is currently not checked on CI, so it got broken recently. Fixing it. Waiting for @metaclips to reenable that check on CI


adrianbenavides (dismissed):
Checked with both:
- `cargo run --example hello --no-default-features --features="alloc, no_std"`
- `cargo +nightly run --example hello --target thumbv7em-none-eabihf --no-default-features --features="qemu"`

adrianbenavides (approved):

PR #4144, opened 2023-01-27
fix(rust): allow deleting a project by id so it doesn't have to wait for it to be ready

adrianbenavides:
When a project fails to be created, it can't be deleted from the CLI because it was forced to wait until the project was successfully created.

To get around this limitation, the `ockam project delete` command should allow the user to delete a command passing the project ID to prevent the CLI from fetching the project details to learn its route given its name.

The new variation looks like this: `ockam project delete {s-name} --project-id {p-id}`


adrianbenavides:
Not needed for now as discussed with @mszpakowski

PR #4145, opened 2023-01-27
properly unset variables in scripts

metaclips:
Use unset command to delete arrays

### DIFF
--- a/tools/scripts/release/crate-bump.sh
+++ b/tools/scripts/release/crate-bump.sh
@@ -61,7 +61,7 @@ bumping_transitive_deps=false
 # on the second iteration, (crateC and crateB) version is then bumped, on the third iteration
 # we do not bump crateC version even though its dep has been modified as it's version has already
 # been bumped for a release.
-while [[ "$updated_crates" != "$recently_updated_crates" ]]; do
+while [[ "${updated_crates[*]}" != "${recently_updated_crates[*]}" ]]; do
   for crate in "${updated_crates[@]}"; do
     if [[ -n "${bumped_crates[$crate]}" ]]; then
       echo "===> $crate has been bumped recently ignoring"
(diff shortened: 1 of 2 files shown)


polvorin (dismissed):

tools/scripts/release/crate-bump.sh
```diff
@@ -88,11 +88,15 @@ while [[ "$updated_crates" != "$recently_updated_crates" ]]; do
     echo y | cargo release "$version" --config tools/scripts/release/release.toml --no-push --no-publish --no-tag --no-dev-version --package "$name" --execute
   done
 
-  recently_updated_crates=$updated_crates
+  unset recently_updated_crates
```
polvorin: feels strange to unset then set it again immediately after?

metaclips: We are trying to handle scenarios like below
- `ockam_core` code was updated by developer, so it version gets bumped by the script
- `ockam_core` is a transitive dependency of `ockam` crate so we have to bump the `version` of `ockam_core` in `ockam` Cargo.toml file and also `ockam` version itself (since there was an update to the crate which is publishable).

To ensure we bump all updated crates, we need to track crates that were bumped due to code changes and their transitive deps

metaclips:
@mergify-io rebase

metaclips:
@mergify rebase

tools/scripts/release/crate-bump.sh
metaclips: @polvorin can you please re-review? Apologies.

polvorin: I mean,   what I don't get is the difference between
```
    unset recently_updated_crates
    recently_updated_crates="${updated_crates[@]}"
```
that the script does,  vs just

```
    recently_updated_crates="${updated_crates[@]}"
```

unless there is something I'm not seeing :man_shrugging:   .  If it is needed for something non-obvious lets put a comment on it noting why we need to unset before immediately setting it again.

polvorin (dismissed):

tools/scripts/release/crate-bump.sh
metaclips: Works better. Thanks.

metaclips: Please re-review

polvorin (approved):

PR #4153, opened 2023-01-28
feat(rust): fix errors in ockam status command

### LINKED ISSUE #4096: Diagnose and fix errors in `ockam status` command
This command has three known issues that must be addressed.

### How to reproduce

```bash
ockam reset -y
ockam status
```

### Fix proposal

[This section of the code](https://github.com/build-trust/ockam/blob/7cbe20ebbeeb7515ae63aae6804581f0fbb20044/implementations/rust/ockam/ockam_command/src/status.rs#L39-L47) is incorrect because `list()` will return an empty list instead of an error where there are no nodes in the system. That part should be rewritten as follows:

```rust
let node_states = opts.state.nodes.list()?;  
if node_states.is_empty() {  
    return Err(anyhow!("No nodes registered on this system!").into());  
}
```

### How to reproduce

Note: the previous error must be solved to reproduce this one.

```bash
ockam reset -y
ockam status
```

### Fix proposal

Currently, when running `ockam status` on a fresh install the command exits successfully without showing any output. The command should return an error when there are no enrolled identities. Before calling the `print_status`, there should be a check like this:

```rust
if status_identities.is_empty() {  
    return Err(anyhow!("No enrolled identities found! Try passing the `--all` argument to see all identities.").into());  
}
```

### How to reproduce

```shell
ockam reset -y
ockam node create
ockam node create
ockam status
```

### Fix proposal

For every registered node, this command runs an `Rpc` request. Currently, it tries to create a new `TcpTransport` on each request, resulting in the following error:

```
» ockam status
Version 0.77.0, hash: 99c3431ec67affcdf80d393f75e40bd69575c33b
failed because router state: a router for this type already exists

Caused by:
    failed because router state: a router for this type already exists
```

To fix this, the `get_node_status` must be refactored so it accepts a `TcpTransport` reference as a parameter. The `TcpTransport` must be created before [the `node_states` loop](https://github.com/build-trust/ockam/blob/7cbe20ebbeeb7515ae63aae6804581f0fbb20044/implementations/rust/ockam/ockam_command/src/status.rs#L50).

---

_If you're looking for other issues to contribute to, checkout [this discussion](https://github.com/build-trust/ockam/discussions/3206) and labels - https://github.com/build-trust/ockam/labels/good%20first%20issue or https://github.com/build-trust/ockam/labels/help%20wanted_

hweawer:
Closes https://github.com/build-trust/ockam/issues/4096

## Current Behavior

`ockam status`

1. Is failing with IO error when no nodes are registered.
2. When running on a fresh install the command exits successfully without showing any output.
3. Fails when there is more than 1 node registered

## Proposed Changes

1. Checking if the list of node states is empty, if so fail with appropriate message.
2. Checking if the list of identities is empty, if so fail with appropriate message.
3. Share `TcpTransport` between `Rpc`s for different nodes. 

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/status.rs
+++ b/implementations/rust/ockam/ockam_command/src/status.rs
@@ -1,6 +1,6 @@
 use crate::util::{api, node_rpc, RpcBuilder};
+use crate::CommandGlobalOpts;
 use crate::Result;
-use crate::{exitcode, CommandGlobalOpts};
 use anyhow::anyhow;
 use clap::Args;
 use ockam::{Context, TcpTransport};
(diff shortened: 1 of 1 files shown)


adrianbenavides (approved):
Awesome, thanks @hweawer for this PR. LGTM 🚀

PR #4155, opened 2023-01-30
rust kafka sidecar

davide-baldo:
Kafka sidecar for secure tunneling.

From the last feedback I made some simplifications: there are now two instances of `KafKaWorker` rather than a single one with multiple addresses.

In order to keep track of which answers originated from which requests I had to add a shared state between the 2 workers: `ProtocolState`.
To create a dynamic inlet only once for each broker I also created another shared state across every worker `KafkaInletMap`.

these are the areas are ones I most uncertain of:
- In order to create new inlets dynamically I had to call node worker API (the ones used from CLI) from within the KafkaInletMap, which seems like I'm calling an API above the level of abstraction and feels a bit unnatural
- To test the worker I've created an integration test that also starts the node manager, is it ok? should have used bats instead?


implementations/rust/ockam/ockam_transport_tcp/src/portal/mod.rs
```diff
@@ -6,6 +6,8 @@ mod portal_worker;
 
 pub(crate) use inlet_listener::*;
 pub(crate) use outlet_listener::*;
+pub use portal_message::PortalMessage;
```
SanjoDeundiak: We would probably want to move Portals protocol-related things out of Tcp, probably to `ockam_core`, or `ockam_transport_core`. This is for the future

implementations/rust/ockam/ockam_api/src/kafka/decoder.rs
```diff
@@ -0,0 +1,62 @@
+use crate::kafka::kafka_portal_worker::MAX_KAFKA_MESSAGE_SIZE;
+use bytes::BytesMut;
+use futures::Stream;
+use ockam::compat::tokio;
+use ockam_transport_tcp::MAX_PAYLOAD_SIZE;
+use std::io::{Error, ErrorKind};
```
SanjoDeundiak: Can we safely replace this with no_std counterparts?

implementations/rust/ockam/ockam_api/src/kafka/kafka_inlet_map.rs
```diff
@@ -0,0 +1,115 @@
+use crate::nodes::models::portal::{CreateInlet, CreateOutlet};
+use crate::nodes::NODEMANAGER_ADDR;
+use crate::route_to_multiaddr;
+use ockam::compat::tokio::sync::Mutex;
+use ockam_core::api::Request;
+use ockam_core::errcode::{Kind, Origin};
+use ockam_core::{route, Error};
+
+use ockam_node::Context;
+use std::collections::HashMap;
```
SanjoDeundiak: We use `ockam_core::compat` to import `std`-things to ease no_std support now/in the future

implementations/rust/ockam/ockam_api/src/kafka/kafka_inlet_map.rs
```diff
@@ -0,0 +1,115 @@
...
+use std::net::SocketAddr;
+use std::str::FromStr;
```
SanjoDeundiak: Avoiding `std` dependency is preferable when possible, do you mind using `core::str::FromStr` instead?

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_listener.rs
```diff
@@ -0,0 +1,85 @@
...
+        message: Routed<Self::Message>,
+    ) -> ockam::Result<()> {
+        trace!("received message: {:?}", &message);
+
+        let result = KafkaPortalWorker::start(context, self.inlet_map.clone()).await;
+        if let Err(err) = result {
+            warn!("cannot spawn worker; err = {:?}", err);
+            return Err(err);
+        }
+        let worker_address = result.unwrap();
```
SanjoDeundiak: Could you please consider rewriting this part to avoid unwrapping? Since you already handle error condition

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_listener.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    pub(crate) async fn start(
+        context: &Context,
+        portal_address: Address,
+        inlet_map: Arc<KafkaInletMap>,
+    ) -> ockam_core::Result<()> {
+        //TODO: only allow the inlet
+        context
+            .start_worker(portal_address, Self { inlet_map }, AllowAll, AllowAll)
```
SanjoDeundiak: Ping me if you need some help with figuring out AccessControl here

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_worker.rs
```diff
@@ -0,0 +1,423 @@
...
+
+///Acts like a relay for messages between tcp inlet and outlet for both directions.
+/// It's meant to be created by the portal listener.
+///
+/// This instance manage both streams inlet and outlet in two different workers, one dedicated
+/// to the requests (inlet=>outlet) the other for the responses (outlet=>inlet).
+/// since every kafka message is length-delimited every message is read and written
+/// through a framed encoder/decoder.
+///
+/// ```text
```
SanjoDeundiak: Like that 👍

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_worker.rs
```diff
@@ -0,0 +1,423 @@
...
+use ockam_core::{
+    errcode::{Kind, Origin},
+    Address, AllowAll, Decodable, Encodable, Error, LocalInfo, LocalMessage, Route, Routed,
+    TransportMessage, Worker,
+};
+use ockam_transport_tcp::{PortalMessage, MAX_PAYLOAD_SIZE};
+use std::sync::Arc;
+
+///by default kafka supports up to 1MB messages
+pub(crate) const MAX_KAFKA_MESSAGE_SIZE: usize = 16 * 1024 * 1024;
```
SanjoDeundiak: Should it be 1MB or 16MB?

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_worker.rs
```diff
@@ -0,0 +1,423 @@
...
+    async fn handle_message(
+        &mut self,
+        context: &mut Self::Context,
+        routed_message: Routed<Self::Message>,
+    ) -> ockam::Result<()> {
+        trace!("received message: {:?}", &routed_message);
+
+        let onward_route = routed_message.onward_route();
+        let return_route = routed_message.return_route();
+        let portal_message = PortalMessage::decode(routed_message.payload())?;
```
SanjoDeundiak: I think that should do the same but without extra decoding
```
let portal_message = routed_message.body();
```

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_worker.rs
```diff
@@ -0,0 +1,423 @@
...
+                    .await?;
+                }
+            }
+            PortalMessage::Disconnect => {
+                //the shutdown will take care of sending disconnect
+                context
+                    .stop_worker(self.other_worker_address.clone())
+                    .await?;
+                context.stop_worker(context.address()).await?;
+                self.forward(context, routed_message).await?;
```
SanjoDeundiak: I'm not sure it's safe to use `Context` instance after it was shutdown

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_worker.rs
davide-baldo: 16mb it's the maximum "sane" message size suggested, it'll add a note

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_worker.rs
```diff
@@ -0,0 +1,423 @@
...
+/// │        ├────────────►│ Kafka   ├────────────►│        │
+/// │        │             │ Request │             │        │
+/// │  TCP   │             └─────────┘             │  TCP   │
+/// │ Inlet  │             ┌─────────┐             │ Outlet │
+/// │        │  encoder    │ Kafka   │   decoder   │        │
+/// │        │◄────────────┤ Response│◄────────────┤        │
+/// └────────┘             └─────────┘             └────────┘
+///```
+pub(crate) struct KafkaPortalWorker {
+    other_worker_address: Address,
```
SanjoDeundiak: Would be nice to have some comment about what is the "other" worker here. Also, what is the relation between them, is one of them responsible for shutting down the other?

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_worker.rs
```diff
@@ -0,0 +1,423 @@
...
+
+impl KafkaPortalWorker {
+    ///returns address used for inlet communications
+    pub(crate) async fn start(
+        context: &mut Context,
+        inlet_map: Arc<KafkaInletMap>,
+    ) -> ockam_core::Result<Address> {
+        let shared_protocol_state = Arc::new(ProtocolState::new());
+
+        let inlet_address = Address::random_local();
```
SanjoDeundiak: Could you please use Address::random_tagged so that's it's easier to debug?

implementations/rust/ockam/ockam_api/src/kafka/kafka_inlet_map.rs
davide-baldo: I'm not sure I can deliver the functionality without std, I'll try to migrate dependencies

implementations/rust/ockam/ockam_api/src/lib.rs
```diff
@@ -22,6 +23,7 @@ pub mod lmdb;
 
 #[macro_use]
 extern crate tracing;
+extern crate core;
```
SanjoDeundiak: Could you please elaborate what is the purpose of that?

davide-baldo: shouldn't be there, and it's not needed

davide-baldo: removing

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_worker.rs
davide-baldo: make sense, reversing the order in doubt!

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_worker.rs
davide-baldo: yep, adding comments!

implementations/rust/ockam/ockam_api/src/kafka/decoder.rs
davide-baldo: fixed everything

implementations/rust/ockam/ockam_api/src/kafka/kafka_inlet_map.rs
davide-baldo: done

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_listener.rs
davide-baldo: Yep, I need help, I tried to naively allow only inlets listeners, but what I actually need are the TCP inlet workers instead. 

https://github.com/build-trust/ockam/blob/davide-baldo/kafka-sidecar/implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_listener.rs#L64

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_worker.rs
davide-baldo: thanks!

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_worker.rs
davide-baldo: done!

implementations/rust/ockam/ockam_api/src/kafka/kafka_portal_listener.rs
davide-baldo: removed the code, I will fix permissions on the next iteration

implementations/rust/ockam/ockam_api/src/kafka/portal_listener.rs
```diff
@@ -0,0 +1,72 @@
...
+            .replace(worker_address.clone());
+
+        trace!(
+            "forwarding message: onward={:?}; return={:?}; worker={:?}",
+            &message.transport().onward_route,
+            &message.transport().return_route,
+            worker_address
+        );
+
+        context.forward(message).await?;
```
hairyhum, line 45: What does this do with the message local metadata?

davide-baldo: The message is converted into a LocalMessage which retains all local metadata, so they are just forwarded "as-is"

davide-baldo: while in `portal_worker.rs:67` instead, where the ockam message probably won't translate to 1 to 1 to kafka message and is merged and split again, the local information of the last message is copied for every message sent

hairyhum: > the local information of the last message is copied for every message sent

"Last message" case would happen if kafka message is in multiple ockam messages, right?
And if it's other way around - we copy the message metadata over multiple outgoing messages, or we repack batched kafka messages into one ockam message?

In Elixir we usually don't forward local metadata. @SanjoDeundiak do we preserve local metadata when forwarding? What's the security implication if we do that?

implementations/rust/ockam/ockam_api/src/kafka/inlet_map.rs
```diff
@@ -0,0 +1,144 @@
+use core::str::FromStr;
+use minicbor::Decoder;
+
+use ockam::compat::tokio::sync::Mutex;
```
SanjoDeundiak, line 4: We want to use sync Mutexes/Locks instead of async where possible. Could we use sync here?

implementations/rust/ockam/ockam_api/src/kafka/inlet_map.rs
```diff
@@ -0,0 +1,144 @@
...
+                    return Err(Error::new(
+                        Origin::Transport,
+                        Kind::Invalid,
+                        "cannot convert route to multiaddr",
+                    ));
+                }
+                Some(to) => to,
+            };
+
+            let buffer: Vec<u8> = context
```
SanjoDeundiak, line 104: If the `NodeManager` is run on the same node (is it?), could we replace that with a function call?

SanjoDeundiak: Also, converting an ockam route to a multiaddr is technically correct. Still, we should be careful with it, as it may lose some information we expect in that multiaddr later (like `node`, `project`, etc.)

implementations/rust/ockam/ockam_api/src/kafka/inlet_map.rs
```diff
@@ -0,0 +1,144 @@
...
+
+            let to = route_to_multiaddr(
+                &self_guard
+                    .interceptor_route
+                    .clone()
+                    .modify()
+                    .append(kafka_outlet_address(broker_id))
+                    .into(),
+            );
+            let to = match to {
```
SanjoDeundiak: Do you think using `.ok_or()` would be cleaner here?

implementations/rust/ockam/ockam_api/src/kafka/portal_listener.rs
```diff
@@ -0,0 +1,72 @@
...
+    type Message = Any;
+    type Context = Context;
+
+    async fn handle_message(
+        &mut self,
+        context: &mut Self::Context,
+        message: Routed<Self::Message>,
+    ) -> ockam::Result<()> {
+        trace!("received first message!");
+        let worker_address = KafkaPortalWorker::start(context, self.inlet_map.clone()).await?;
```
SanjoDeundiak, line 27: Could it be easier to pass the message as an argument here instead of forwarding it later?

implementations/rust/ockam/ockam_api/src/port_range.rs
```diff
@@ -0,0 +1,90 @@
...
+impl TryFrom<&str> for PortRange {
+    type Error = Error;
+
+    fn try_from(text: &str) -> Result<Self, Self::Error> {
+        Self::from_str(text)
+    }
+}
+
+#[test]
+fn port_range_parse() -> () {
```
SanjoDeundiak, line 80: Nice!

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
```diff
@@ -0,0 +1,472 @@
...
+                //stop both workers
+                let disconnect_received = self.disconnect_received.swap(true, Ordering::SeqCst);
+                if !disconnect_received {
+                    context
+                        .stop_worker(self.other_worker_address.clone())
+                        .await?;
+                    context.stop_worker(context.address()).await?;
+                }
+            }
+            _ => self.forward(context, routed_message).await?,
```
SanjoDeundiak: Could we please make other cases explicit, so that it requires update if we change the `PortalMessage`

SanjoDeundiak (changes requested):
I haven't done deep dive into details of Kafka protocol (and won't for now), everything else looks almost good. Leaving few comments below

implementations/rust/ockam/ockam_api/src/kafka/portal_listener.rs
SanjoDeundiak: It's an interesting case; as far as I understand, messages that get in the Worker may be split or united (and generally, there is an arbitrary dependency between the number of messages in and out). I'm not sure what is the correct behavior of `LocalInfo` in that case. For now, the only thing we put into `LocalInfo` is SecureChannel metadata (`IdentitiIdentifier`); I think we should be fine for now if we check that the `LocalInfo` of every message coming in is the same. That way, it's OK to clone it to every message that goes out. Not sure it's a future-proof solution, though.

davide-baldo: > "Last message" case would happen if kafka message is in multiple ockam messages, right?

Yes, we may receive pieces of kafka message 1 byte at a time until it's complete, in this case, we take the local metadata from the ockam message containing the last byte

> And if it's other way around - we copy the message metadata over multiple outgoing messages, or we repack batched kafka messages into one ockam message?

In case multiple kafka messages are received with a single ockam message usually they would be processed and re-forwarded within a single message, but there is no guarantee between ockam messages in and ockam messages out. 
For example, the incoming buffer could be almost full and a single ockam message could bring the last piece of current message and other 3 kafka messages, in this case the current kafka message will be send in several pieces and the most recent 3 kafka message may not end up in the same ockam message.

implementations/rust/ockam/ockam_api/src/kafka/inlet_map.rs
davide-baldo: >  could we replace that with a function call?

That would be ideal and would simplify quite a few things as well as testing, but i don't see any easy way to get a reference to the NodeWorker

implementations/rust/ockam/ockam_api/src/kafka/portal_listener.rs
davide-baldo: I think it would make the worker code more complex because then I would have a special case for the first message

implementations/rust/ockam/ockam_api/src/kafka/inlet_map.rs
davide-baldo: it is! :)

implementations/rust/ockam/ockam_api/src/kafka/inlet_map.rs
davide-baldo: I don't  think I can without risking a deadlock, I'm keeping the lock while exchanging messages to keep the locking mechanism simpler, and if another task tries to lock while waiting for a message it could cause a deadlock (if tokio assigned the same thread to both tasks or if we are running single-threaded)

implementations/rust/ockam/ockam_api/src/kafka/inlet_map.rs
davide-baldo: > Also, converting an ockam route to a multiaddr is technically correct. Still, we should be careful with it, as it may lose some information we expect in that multiaddr later (like `node`, `project`, etc.)

If we could use directly a NodeWorker reference I could avoid the MultiAddress altogether

implementations/rust/ockam/ockam_api/src/kafka/portal_listener.rs
hairyhum: > In case multiple kafka messages are received with a single ockam message usually they would be processed and re-forwarded within a single message, but there is no guarantee between ockam messages in and ockam messages out.

Do I understand correctly that there is some mechanism which re-batches messages? Can you please try to document it somehow, at least in the code comments?

hairyhum: > For now, the only thing we put into LocalInfo is SecureChannel metadata (IdentitiIdentifier)

Yes. And this means that all messages coming from the interceptor are considered coming from a secure channel source. Is that something that we do in other forwarding workers in Rust?
In Elixir I was changing the "source" of the message to "local" when forwarding, because if message is forwarded it's no longer originating from a secure channel.

This behaviour may need some more consistent approach maybe, especially if we add source attributes to the message metadata for ABAC.

davide-baldo: > Do I understand correctly that there is some mechanism which re-batches messages?

Kafka messages content is currently unchanged, it just gets split and merged back.

> Can you please try to document it somehow, at least in the code comments?

Sure!

implementations/rust/ockam/ockam_api/src/kafka/inlet_map.rs
SanjoDeundiak: I don't think there can be a deadlock. Sync mutex lock won't allow you to carry it across async calls

implementations/rust/ockam/ockam_api/src/kafka/inlet_map.rs
SanjoDeundiak: Maybe it can be an abstract dependency? Expressed as a trait maybe? `KafkaInletMap` need a way to spawn inlets, it's fine to provide that functionality as an external dependency

implementations/rust/ockam/ockam_api/src/kafka/inlet_map.rs
davide-baldo: > I don't think there can be a deadlock. Sync mutex lock won't allow you to carry it across async calls

You're right, the compiler stop me from doing it! Should I change the locking schema to use sync mutex?

implementations/rust/ockam/ockam_api/src/kafka/inlet_map.rs
davide-baldo: I've drafted a prototype:  https://github.com/build-trust/ockam/commit/beac50eb5955d489797744837d13511d5ade165f

I expect to need similar access also for secure channels, I think it's beneficial to be able to access node manager pieces directly, but It works only for services (where we can inject the dependency) and then we are duplicating APIs entry points, which is fine as long as it's clear how they are meant to be used.

I wonder if it's worth having a direct API or if RPCs utilities can do the same. RPCs utilities are scheduled to be moved from the command to the API crate.

implementations/rust/ockam/ockam_api/src/kafka/portal_listener.rs
davide-baldo: just pushed the comment commit, marking as resolved

implementations/rust/ockam/ockam_api/src/kafka/inlet_map.rs
davide-baldo: >  RPCs utilities are scheduled to be moved from the command to the API crate.

Sorry, I double-checked, and there is no issue yet!

implementations/rust/ockam/ockam_api/src/kafka/inlet_map.rs
SanjoDeundiak: That would be great!

SanjoDeundiak (dismissed):

SanjoDeundiak (approved):

PR #4162, opened 2023-01-30
chore(rust): continue on process not found when killing a node

### LINKED ISSUE #4136: Diagnose and fix `ockam node delete --all` error
## Expected Behavior

`ockam node delete --all` should not throw an error when a foreground node is stopped.

## Current Behavior

```
» ockam node create n1 --foreground
^C

» ockam node list

» ockam node delete --all
Version 0.78.0, hash: 7b1d00e8b11881d6b7c422f9e7acc4c77bbc7ba5
{code: 70, err: IO error}
```

## Steps to Reproduce

1. Start a foreground node `ockam node create n1 --foreground`
2. Then press `Ctrl-C` to exit it
3. `ockam node list` Will show an empty list of nodes
4. `ockam node delete --all` Will throw the error

---

hargut:
## Current Behavior
During some actions e.g. `node delete --all` the ockam command fails in certain scenarios with an `IO Error`.

## Proposed Changes
Only warn on `No such process` (Posix ESRCH) during `NodeState::kill_process()` calls.
Print the root cause in the `Display` implementation of `ockam_command::error:Error`.

Fixes https://github.com/build-trust/ockam/issues/4136, but was also reproducible when e.g. killing a node using `kill` which leaves a stale `pid` file in the `nodes` config directory.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state.rs
@@ -1,5 +1,6 @@
 use crate::cloud::project::Project;
 use crate::nodes::models::transport::{CreateTransportJson, TransportMode, TransportType};
+use nix::errno::Errno;
 use ockam_identity::change_history::{IdentityChangeHistory, IdentityHistoryComparison};
 use ockam_identity::{Identity, IdentityIdentifier, SecureChannelRegistry};
 use ockam_vault::{storage::FileStorage, Vault};
@@ -738,6 +739,14 @@ impl NodeState {
                     nix::sys::signal::Signal::SIGTERM
                 },
             )
+            .or_else(|e| {
+                if e == Errno::ESRCH {
+                    tracing::warn!(node = %self.config.name, %pid, "No such process");
+                    Ok(())
+                } else {
+                    Err(e)
+                }
+            })
             .map_err(|e| {
                 CliStateError::Io(std::io::Error::new(
                     std::io::ErrorKind::Other,
--- a/implementations/rust/ockam/ockam_command/src/error.rs
+++ b/implementations/rust/ockam/ockam_command/src/error.rs
@@ -32,7 +32,13 @@ impl Debug for Error {
 impl Display for Error {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         writeln!(f, "{}", Version::short())?;
-        writeln!(f, "{{code: {}, err: {}}}", self.code, self.inner)
+        writeln!(
+            f,
+            "{{code: {}, err: {}, cause: {}}}",
+            self.code,
+            self.inner,
+            self.inner.root_cause()
+        )
     }
 }
 
--- a/implementations/rust/ockam/ockam_command/src/node/delete.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/delete.rs
@@ -11,11 +11,11 @@ pub struct DeleteCommand {
     #[arg(default_value_t = default_node_name(), group = "nodes")]
     node_name: String,
 
-    /// Terminate all nodes
+    /// Terminate all node processes and delete all node configurations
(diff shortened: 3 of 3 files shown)


implementations/rust/ockam/ockam_api/src/cli_state.rs
```diff
@@ -738,6 +739,17 @@ impl NodeState {
                     nix::sys::signal::Signal::SIGTERM
                 },
             )
+            .or_else(|e| {
+                if e == Errno::ESRCH {
+                    println!(
+                        "Warning: no such process - PID `{pid}` for node `{}`",
+                        self.config.name
+                    );
```
adrianbenavides: If the user doesn't have logging enabled, probably doesn't care if a PID wasn't found, so this message should be logged instead of printed: `tracing::warn(%node, %pid, "No such process")`.

adrianbenavides (commented):
I couldn't reproduce any of the errors: when creating a foreground node, the `pid` file is not created, so it won't try to stop any process when deleting it. Ignoring the "no such process" is still a good improvement.

One thing you could do to reduce the side effects of running foreground nodes is integrating [`ctrlc`](https://docs.rs/ctrlc/latest/ctrlc/) to properly handle the `Ctrl+C`. This handler should only be created when passing the `--foreground` flag AND is not a child process. On receiving that signal, we could call the `NodeState::delete()` function. Finally, the node state should always be removed [before exiting gracefully](https://github.com/build-trust/ockam/blob/d96b2cdd94a0ed5a1a5636e224ab1ce64f6b6aff/implementations/rust/ockam/ockam_command/src/node/create.rs#L304) and under the same conditions (`--foreground` flag passed and `--child-process` is false).

Also, while we are here, we could improve the docs for the [`--force` argument](https://github.com/build-trust/ockam/blob/d96b2cdd94a0ed5a1a5636e224ab1ce64f6b6aff/implementations/rust/ockam/ockam_command/src/node/delete.rs#L18). It should state that when passing that flag, the nodes' processes will be terminated immediately (and not gracefully).

implementations/rust/ockam/ockam_api/src/cli_state.rs
hargut: Updated.

hargut:
Hi @adrianbenavides,

at the start of analysing the issue I could reproduce the problem, but had some previous nodes on the system. Likely one of these had a stale pid file, with no corresponding process due to reboots.

After a successful `node delete --all` command I was as well not able to reproduce the error described in the issue.  As during the `node create --foreground` call there is no pid file created this should as well not be an error scenario for the listed commands.

The idea with `ctrlc` and cleaning up the directory after interruption is good, but I'm a bit hesitant if this will keep command expectations aligned. The command is named `node create` and therefore it might be expected that the node is created as well when using `--foreground`.

In case the desired behaviour is a deleted node after foreground creation and interruption I'll (try to) implement that change.

adrianbenavides:
> Hi @adrianbenavides,
> 
> at the start of analysing the issue I could reproduce the problem, but had some previous nodes on the system. Likely one of these had a stale pid file, with no corresponding process due to reboots.

That makes sense yes, thanks!

> The idea with `ctrlc` and cleaning up the directory after interruption is good, but I'm a bit hesitant if this will keep command expectations aligned. The command is named `node create` and therefore it might be expected that the node is created as well when using `--foreground`.

Let's keep that out of the scope of this PR. I'll open a separate issue so we can study that proposal in detail before implementing a solution.

adrianbenavides:
I could reproduce the error with:

```rust
ockam node create
# Manually stop/kill the process
ockam node delete --all
```

adrianbenavides (dismissed):
It seems to work following the steps that reproduced the error on `develop` 🥳 

Thanks @hargut this great PR!

adrianbenavides (approved):

PR #4163, opened 2023-01-31
chore(rust): fix for `node create /node/n1` bug

### LINKED ISSUE #4137: Diagnose and fix `ockam node create /node/n1` error
## Currently

Fails when passing a `MultiAddr` as a node name.

## Desired

This command only needs to extract the node name from the `node_name` argument (i.e. it doesn't need to transform it in any way). Currently, the command will take the command argument value as is, whereas we also need to support valid MultiAddr strings with the format `/node/{name}`.

To support both inputs, the argument type must be `String` so it can be processed with the [`extract_address_value` function](https://github.com/build-trust/ockam/blob/f1b551e0e69cefeaa52133c4dd84546d38cab4cb/implementations/rust/ockam/ockam_command/src/util/mod.rs#L450) function. You can check how this function works in the [unit tests](https://github.com/build-trust/ockam/blob/f1b551e0e69cefeaa52133c4dd84546d38cab4cb/implementations/rust/ockam/ockam_command/src/util/mod.rs#L542).

A potential solution would be to create a wrapper function around `extract_address_value` with the following logic:
- If the input string is a multiaddr, then check that it's of type `Node`. If it's not, then return an error saying that the input MultiAddr must follow the format `/node/{name}`. If it is, call the `extract_address_value` function
- else, return the input unprocessed

## How to reproduce the error

```bash
» ockam node create /node/n2
Version 0.78.0, hash: 7b1d00e8b11881d6b7c422f9e7acc4c77bbc7ba5
IO error

Caused by:
Read-only file system (os error 30)
```

nidnogg:
## Current Behavior

Whenever a node is created via `node create <node_name>` and `node_name` is a `MultiAddr` String, an IO error is produced. 

## Proposed Changes

This PR fixes #4137 by adding a `parse_node_name()` function before the node's creation that uses the `extract_address_value` util (when applicable) to circumvent this.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -148,7 +148,7 @@ impl CreateCommand {
     fn parse_node_name(&self) -> crate::Result<Self> {
         let cmd = self.clone();
 
-        if self.node_name.starts_with('/') {
+        if self.node_name.contains('/') {
             if self.node_name.starts_with("/node/") {
                 let node_name = extract_address_value(&self.node_name)?;
                 return Ok(Self { node_name, ..cmd });
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -12,7 +12,6 @@ use std::{
 };
 use tracing::error;
 
-use crate::node::util::spawn_node;
 use crate::secure_channel::listener::create as secure_channel_listener;
 use crate::service::config::Config;
 use crate::service::start;
@@ -22,6 +21,7 @@ use crate::{
     help, node::show::print_query_status, node::HELP_DETAIL, project, util::find_available_port,
     CommandGlobalOpts,
 };
+use crate::{node::util::spawn_node, util::extract_address_value};
 use crate::{
     node::util::{add_project_authority_from_project_info, init_node_state},
     util::RpcBuilder,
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -144,6 +144,23 @@ impl CreateCommand {
             ..cmd
         })
     }
+
+    fn parse_node_name(&self) -> crate::Result<Self> {
+        let cmd = self.clone();
+
+        if self.node_name.contains('/') {
+            if self.node_name.starts_with("/node/") {
```
adrianbenavides: This is a little hack we should avoid as it doesn't express what we are trying to parse here. You should parse the string into a MultiAddr instance and check the protocol like it's done inside `extract_address_value`. It's more verbose, but it doesn't depend on manual/hard-coded parsing rules.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -144,6 +144,23 @@ impl CreateCommand {
...
+                let node_name = extract_address_value(&self.node_name)?;
+                return Ok(Self { node_name, ..cmd });
+            } else {
+                return Err(crate::Error::new(
+                    exitcode::IOERR,
+                    anyhow!("Error: A MultiAddr node must follow the format /node/{{name}}"),
+                ));
+            }
+        }
+        Ok(Self { ..cmd })
```
adrianbenavides: This function should be static and should live in the utils.rs close to the `extract_address_value` and should return a `String` so we can reuse it elsewhere if needed. The input should be a `&str` too.

implementations/rust/ockam/ockam_command/src/node/create.rs
nidnogg: Modified it in the latest commit to work similarly to `extract_address_value`

implementations/rust/ockam/ockam_command/src/node/create.rs
nidnogg: Return types changed and function moved closer to `extract_address_value` for reusability in the latest commit

adrianbenavides (dismissed):
Looks perfect, thank you so much @nidnogg for this great contribution!

adrianbenavides (approved):

PR #4172, opened 2023-02-01
Rework transport

SanjoDeundiak:
1. Get rid of TCP auto-connection
2. Get rid of `TcpRouter`
3. Get rid of `TCP` Address type
4. Switch to explicit TCP connections
5. Switch to addressing individual connection Workers by their random local Address
6. Allow multiple connections to the same TCP address
7. +Supporting changes to avoid breaking things after the architecture redesign


implementations/rust/ockam/ockam_command/src/authenticated.rs
```diff
@@ -54,25 +54,26 @@ pub enum AuthenticatedSubcommand {
 
 impl AuthenticatedCommand {
     pub fn run(self) {
+        // FIXME: Requires update to the latest format
```
SanjoDeundiak, line 57: @adrianbenavides FYI

documentation/guides/rust/get-started/04-transport/README.md
```diff
@@ -166,11 +169,20 @@ async fn main(ctx: Context) -> Result<()> {
...
 
+    // Create a TCP connection to the responder node.
+    let connection_to_responder = tcp.connect("127.0.0.1:4000").await?;
+
     // Create a TCP listener and wait for incoming connections.
     tcp.listen("127.0.0.1:3000").await?;
 
-    // Create a Hop worker
-    ctx.start_worker("hop", Hop, AllowAll, AllowAll).await?;
+    // Create a Forwarder worker
```
etorreborre, line 175: Maybe we could move the forwarder creation near the `connection_to_responder` and create the TCP listener last?

documentation/guides/rust/get-started/05-secure-channel/README.md
```diff
@@ -85,17 +85,22 @@ Add the following code to this file:
...
 async fn main(ctx: Context) -> Result<()> {
-    ctx.start_worker("hop", Hop, AllowAll, AllowAll).await?;
-
     // Initialize the TCP Transport.
     let tcp = TcpTransport::create(&ctx).await?;
 
+    // Create a TCP connection to Bob.
+    let connection_to_bob = tcp.connect("127.0.0.1:4000").await?;
+
+    // Start a Forwarder to forward messages to Bob using the TCP connection.
```
etorreborre, line 100: BTW I wonder if this example really makes sense. I have the impression that in all cases Bob would create his own forwarder on the middle node and not have to open a TCP listener. Maybe we can add a comment saying that real-life scenarios are more likely to use a forwarding service on the Orchestrator.

examples/rust/get_started/examples/04-routing-over-transport-two-hops-middle.rs
```diff
@@ -2,7 +2,7 @@
 // Starts a tcp listener at 127.0.0.1:3000
 // It then runs forever waiting to route messages.
 
-use hello_ockam::Hop;
```
etorreborre, line 5: It would be nice to find a way to reduce the duplication between those examples and the same examples in the READMEs because this adds up to the maintenance every time we do a change.

implementations/rust/ockam/ockam_api/src/cloud/mod.rs
```diff
@@ -51,10 +52,11 @@ impl<'a, T> CloudRequestWrapper<'a, T> {
         }
     }
 
-    pub fn route(&self) -> Result<Route> {
+    pub async fn route(&self, tcp: &TcpTransport) -> Result<Route> {
```
etorreborre, line 55: I would propose a different approach here, in order to make the code more resilient to further changes if we start using a UDP transport for example:

 - leave `CloudRequestWrapper` as is it but with a `multiaddr()` function just  returning a parsed `MultiAddr`
 
 - have the function `multiaddr_to_route` be a method of `NodeManagerWorker/NodeManager` in order to be able to access all the created transports internally to do the necessary connections

This means that many calls in ockam_api files will look like:
```rust
let cloud_route = self.multiaddr_to_route(req_wrapper.multiaddr()?).await?;
```

implementations/rust/ockam/ockam_api/src/nodes/models/transport.rs
```diff
@@ -148,27 +148,31 @@ pub struct TransportStatus<'a> {
     /// The mode the transport should operate in
     #[n(3)] pub tm: TransportMode,
     /// The status payload
-    #[b(4)] pub payload: CowStr<'a>,
+    #[b(4)] pub socket_addr: CowStr<'a>,
+    /// The status payload
```
etorreborre: This comment needs to be changed

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -94,12 +97,14 @@ pub(crate) struct AuthorityInfo {
     addr: MultiAddr,
 }
 
+type TransportsType = BTreeMap<Alias, (TransportType, TransportMode, Address, String)>;
```
etorreborre: I would prefer simply `Transports` as a name here, `Type` is a bit redundant since this is a type alias.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -209,13 +214,13 @@ impl<'a> NodeManagerProjectsOptions<'a> {
 }
 
 pub struct NodeManagerTransportOptions {
-    api_transport: (TransportType, TransportMode, String),
+    api_transport: (TransportType, TransportMode, Address, String),
```
etorreborre, line 217: Maybe it's time to create a proper struct for `ApiTransport` and reuse it in the definition of `Transports`? 
This way we can also document what it means and have `TransportStatus::new` just take the transport id + api transport so that we duplicate a bit less code.

implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -188,19 +190,18 @@ impl<'a> Rpc<'a> {
...
                     None => {
                         let tcp = TcpTransport::create(ctx).await?;
-                        tcp.connect(addr_str).await?;
+                        tcp.connect(addr_str).await?
                     }
                     Some(tcp) => {
-                        // Ignore "already connected" error.
-                        let _ = tcp.connect(addr_str).await;
+                        // Create a new connection anyway
+                        tcp.connect(addr_str).await?
```
adrianbenavides, line 201: So this won't fail now when trying to connect more than once to the same address?

implementations/rust/ockam/ockam_api/src/session/util.rs
```diff
@@ -5,10 +5,18 @@ use std::time::Duration;
 pub(crate) const MAX_RECOVERY_TIME: Duration = Duration::from_secs(10);
 pub(crate) const MAX_CONNECT_TIME: Duration = Duration::from_secs(5);
 
-pub(crate) fn starts_with_host_tcp_secure(addr: &MultiAddr) -> Option<usize> {
+pub(crate) fn starts_with_host_tcp(addr: &MultiAddr) -> Option<usize> {
```
etorreborre, line 8: It would be nicer to return the prefix and the rest of the `MultiAddr` rather than an index to later use for manual splitting. So instead of
```rust
let Some(pos1) = starts_with_host_tcp(addr) {
            debug!(%addr, "creating a tcp connection");
            let (a1, b1) = addr.split(pos1);
```
We could have
```rust
impl MultiAddr {

   // split the MultiAddr into 2 MultiAddr where
   //   - the first is a multi addr is a Dns, Ipv4 or Ipv6 address
   //   - the second part is the remaining of the MultiAddr
   //  For example etc....  
   fn split_network_host(&self) -> Result<(MultiAddr, MultiAddr)> {
     ...
  }

  // split the MultiAddr into 2 MultiAddr where
  // ... 
  fn split_secure(&self) -> Result<(MultiAddr, MultiAddr)> {
}

...
let Some(host_address, rest) = addr.split_network_host() {
            debug!(%tcp_address, "creating a tcp connection");
            
```

implementations/rust/ockam/ockam_api/src/util.rs
```diff
@@ -1,81 +1,114 @@
...
+            Service::CODE => {
+                let local = p.cast::<Service>()?;
+                rb = rb.append(Address::new(LOCAL, &*local))
+            }
+            Secure::CODE => {
+                let local = p.cast::<Secure>()?;
+                rb = rb.append(Address::new(LOCAL, &*local))
+            }
+
+            // If your code crashes here then the front-end CLI isn't
```
etorreborre, line 28: I think it could be a good idea to lift that comment to the documentation of the function itself. It is kind of important. Especially because I don't know what kind of error message that would give if the `unreachable` code was triggered and how long it would take to debug it. Maybe we could investigate the idea of using a newtype like `LocalMultiAddr` to enforce the fact that a `MultiAddr` has been "cleaned" before being passed to that function.

Also `clean_multiaddr` does not exist anymore so the comment so refer to `clean_nodes_multiaddr`

implementations/rust/ockam/ockam_api/src/util.rs
```diff
@@ -1,81 +1,114 @@
...
+        }
+    }
+    Some(rb.into())
+}
 
-    let it = input.iter().peekable();
-    for p in it {
+/// If the input MultiAddr is "/dnsaddr/localhost/tcp/4000/service/api",
+/// then this will return string format of the SocketAddr: "127.0.0.1:4000".
+pub fn multiaddr_to_socket_addr(ma: &MultiAddr) -> Option<String> {
```
etorreborre, line 45: This function could be put directly onto the `MultiAddr` struct.

implementations/rust/ockam/ockam_command/src/authenticated.rs
```diff
@@ -54,25 +54,26 @@ pub enum AuthenticatedSubcommand {
...
         if let Err(e) = embedded_node(run_impl, self.subcommand) {
             eprintln!("Ockam node failed: {e:?}",);
         }
     }
 }
 
 async fn run_impl(ctx: Context, cmd: AuthenticatedSubcommand) -> crate::Result<()> {
-    TcpTransport::create(&ctx).await?;
+    let tcp = TcpTransport::create(&ctx).await?;
```
etorreborre, line 65: I think we can put the transport creation directly in the client creation.

implementations/rust/ockam/ockam_command/src/node/show.rs
```diff
@@ -101,7 +101,8 @@ fn print_node_info(
             println!("    Transport:");
             println!("      Type: {}", e.tt);
             println!("      Mode: {}", e.tm);
-            println!("      Address: {}", e.payload);
+            println!("      Socket: {}", e.socket_addr);
```
etorreborre, line 104: Nice!

implementations/rust/ockam/ockam_command/src/util/mod.rs
SanjoDeundiak: It won't. The other problem is that we never close these connections... Which I think is fine for now, but let's keep it in mind

implementations/rust/ockam/ockam_command/src/tcp/connection/create.rs
```diff
@@ -45,7 +45,7 @@ impl CreateCommand {
             OutputFormat::Plain => {
                 let from = &self.node_opts.from;
 
-                let to = response.payload.parse::<SocketAddrV4>()?;
+                let to = response.socket_addr.parse::<SocketAddrV4>()?;
```
etorreborre, line 48: Could we put the parsing as `SocketAddrV4` as a `socket_addr()` method on `TransportStatus`?

implementations/rust/ockam/ockam_command/src/tcp/connection/create.rs
```diff
@@ -71,12 +71,10 @@ impl CreateCommand {
                     .default_tcp_listener()?
                     .addr
                     .port();
-                let route: Route = route![(TCP, format!("localhost:{port}"))]
-                    .modify()
-                    .append_t(TCP, response.payload.to_string())
-                    .into();
-                let multiaddr = route_to_multiaddr(&route)
-                    .context("Couldn't convert given address into `MultiAddr`")?;
+                let mut multiaddr = MultiAddr::default();
```
etorreborre, line 74: A `route`-like macro could be nice to build `MultiAddr`. 
Maybe that code could go to the `TransportStatus` struct since it is duplicated in `create.rs`?

implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -188,19 +190,18 @@ impl<'a> Rpc<'a> {
                 ref tcp,
             } => {
                 let port = node_state.setup()?.default_tcp_listener()?.addr.port();
-                let addr = Address::from((TCP, format!("localhost:{port}")));
-                let addr_str = addr.address();
-                match tcp {
+                let addr_str = format!("localhost:{port}");
+                let addr = match tcp {
```
etorreborre, line 194: An idea. Why not make `TcpTransport.create` idempotent? If it is created we don't recreate it. Then clients don't really have to worry about that

implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -548,6 +552,46 @@ pub fn process_multi_addr(addr: &MultiAddr, cli_state: &CliState) -> Result<Mult
     Ok(processed_addr)
 }
 
+/// Go through a multiaddr and remove all instances of
+/// `/node/<whatever>` out of it and replaces it with a fully
+/// qualified address to the target
+pub fn clean_nodes_multiaddr(
```
etorreborre, line 558: That code looks convoluted to me. I would leave it to just do a pass on node address replacement, which involves the `cli_state` nodes and not do anything with the projects (and call the function `resolve_node_address` or something similar). 

Then in `clean_project_multiaddr` (which I'd rather call `resolve_project_secure_channel_address` for example) I would work on the project side of things.

implementations/rust/ockam/ockam_core/src/access_control/any.rs
```diff
@@ -1,35 +1,53 @@
-use crate::access_control::IncomingAccessControl;
+use crate::access_control::{IncomingAccessControl, OutgoingAccessControl};
+use crate::compat::sync::Arc;
+use crate::compat::vec::Vec;
 use crate::{async_trait, compat::boxed::Box, RelayMessage, Result};
 
-use core::fmt::{self, Debug};
+/// Allows message that are allowed by any of [`IncomingAccessControl`]s
+#[derive(Debug)]
+pub struct AnyIncomingAccessControl(Vec<Arc<dyn IncomingAccessControl>>);
```
etorreborre, line 8: Good generalization

implementations/rust/ockam/ockam_core/src/routing/address.rs
```diff
@@ -1,222 +1,14 @@
...
 };
-use crate::{debugger, DenyAll, OutgoingAccessControl, RelayMessage, Result};
-use core::cmp::Ordering;
+use crate::{AddressParseError, AddressParseErrorKind, Result, TransportType, LOCAL};
 use core::fmt::{self, Debug, Display};
 use core::ops::Deref;
 use core::str::from_utf8;
 use serde::{Deserialize, Serialize};
 
-/// A `Mailbox` controls the dispatch of incoming messages for a
```
etorreborre, line 15: Thanks I also wanted to do this :-)

implementations/rust/ockam/ockam_core/src/routing/address.rs
```diff
@@ -239,54 +31,10 @@ impl Mailboxes {
 #[derive(Serialize, Deserialize, Clone, Hash, Ord, PartialOrd, Eq, PartialEq)]
 pub struct Address {
     tt: TransportType,
+    // TODO: It's binary but in most cases we assume it to be an UTF-8 string
```
etorreborre: Not sure about the `TODO` part. What is there to do?

implementations/rust/ockam/ockam_multiaddr/src/proto.rs
```diff
@@ -203,6 +203,7 @@ macro_rules! gen_str_proto {
     };
 }
 
+gen_str_proto!(Worker, 102526, "worker");
```
etorreborre, line 206: Out of curiosity where do these numerical values come from?

implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -596,10 +596,10 @@ impl Context {
         let (reply_tx, mut reply_rx) = small_channel();
         let next = match route.next() {
             Ok(next) => next,
-            Err(_) => {
+            Err(err) => {
                 // TODO: communicate bad routes to calling function
                 tracing::error!("Invalid route for message sent from {}", sending_address);
-                panic!("invalid destination route");
```
etorreborre, line 602: Nice

implementations/rust/ockam/ockam_node/src/context.rs
```diff
@@ -182,8 +182,8 @@ impl Context {
     }
 
     /// Return all addresses of the current worker
-    pub fn aliases(&self) -> Vec<Address> {
-        self.mailboxes.aliases()
+    pub fn addresses(&self) -> Vec<Address> {
```
etorreborre, line 185: I am not sure about this comment. It says "all" the addresses but this actually does not include the address of the main mailbox. Maybe it would be better to stay with the `main/additional` vocabulary that we have on mailboxes.

implementations/rust/ockam/ockam_node/src/router/mod.rs
```diff
@@ -166,10 +167,13 @@ impl Router {
                     .map_err(|_| NodeError::NodeState(NodeReason::Unknown).internal())?
             }
             // Rejected router registration command
-            Router(_, _, sender) => sender
-                .send(RouterReply::router_exists())
-                .await
-                .map_err(|_| NodeError::NodeState(NodeReason::Unknown).internal())?,
+            Router(_, _, sender) => {
```
etorreborre, line 170: Nice!

implementations/rust/ockam/ockam_transport_tcp/src/lib.rs
```diff
@@ -26,30 +26,25 @@ extern crate core;
...
-/// TCP address type constant
-pub const TCP: TransportType = TransportType::new(1);
-
 pub(crate) const CLUSTER_NAME: &str = "_internals.transport.tcp";
 
-fn parse_socket_addr<S: AsRef<str>>(s: S) -> Result<SocketAddr> {
-    Ok(s.as_ref()
-        .parse()
-        .map_err(|_| TransportError::InvalidAddress)?)
+fn parse_socket_addr(s: &str) -> Result<SocketAddr> {
```
etorreborre: That's another topic but in general I think it would be nice to have no floating functions / structs / types in top-level libs. This way when we open the `lib` file we just see the top-level modules, re-exports and features.

implementations/rust/ockam/ockam_transport_tcp/src/portal/portal_receiver.rs
```diff
@@ -13,18 +13,25 @@ use tracing::{error, warn};
 /// `TcpPortalWorker` after a call is made to
 /// [`TcpPortalWorker::start_receiver`](crate::TcpPortalWorker::start_receiver)
 pub(crate) struct TcpPortalRecvProcessor {
+    registry: TcpRegistry,
     buf: Vec<u8>,
-    rx: OwnedReadHalf,
+    read_half: OwnedReadHalf,
```
etorreborre, line 18: Thanks for the renaming :-)

implementations/rust/ockam/ockam_transport_tcp/src/portal/portal_worker.rs
```diff
@@ -167,10 +174,14 @@ impl TcpPortalWorker {
 
     /// Start a `TcpPortalRecvProcessor`
     async fn start_receiver(&mut self, ctx: &Context, onward_route: Route) -> Result<()> {
-        if let Some(rx) = self.rx.take() {
+        if let Some(rx) = self.read_half.take() {
             let next_hop = onward_route.next()?.clone();
-            let receiver =
-                TcpPortalRecvProcessor::new(rx, self.internal_address.clone(), onward_route);
+            let receiver = TcpPortalRecvProcessor::new(
```
etorreborre, line 179: Talking about names I think I would prefer some expanded names, `TcpSender` and `TcpReceiver` (without insisting on whether they are have a worker or a processor instance)

implementations/rust/ockam/ockam_transport_tcp/src/registry.rs
```diff
@@ -0,0 +1,159 @@
+use ockam_core::compat::sync::{Arc, RwLock};
+use ockam_core::Address;
+
+#[derive(Default)]
```
etorreborre: I would nice to have a comment to explain why we even need such a registry. Ah, I see there's a comment on `TcpRegistry`. So then I would add a comment here saying to refer to `TcpRegistry` and on `TcpRegistry` I would expand to comment to say why we need to do registrations/deregistrations

implementations/rust/ockam/ockam_transport_tcp/src/registry.rs
```diff
@@ -0,0 +1,159 @@
...
+    sender_workers: Vec<Address>,
+    receiver_processors: Vec<Address>,
+}
+
+impl InternalRegistry {
+    fn add_portal_worker(&mut self, addr: &Address) {
+        self.portal_workers.push(addr.clone())
+    }
+    fn remove_portal_worker(&mut self, addr: &Address) {
+        if let Some(index) = self.portal_workers.iter().position(|x| x == addr) {
```
etorreborre: I don't know if that is super idiomatic rust but I would expect more something like
```
self.portal_workers.retain(|x| x != addr)
```
as a one-liner

implementations/rust/ockam/ockam_transport_tcp/src/transport.rs
```diff
@@ -46,11 +50,47 @@ use crate::{parse_socket_addr, TcpOutletListenWorker, TcpRouter, TcpRouterHandle
...
+    pub fn registry(&self) -> &TcpRegistry {
+        &self.registry
+    }
+}
+
+impl TcpTransport {
+    /// Resolve the given peer to a [`SocketAddr`](std::net::SocketAddr)
+    fn resolve_peer(peer: String) -> Result<SocketAddr> {
+        // Try to parse as SocketAddr
+        if let Ok(p) = parse_socket_addr(&peer) {
```
etorreborre, line 74: Maybe we can move the `parse_socket_addr` function to this file instead of `lib.rs`

implementations/rust/ockam/ockam_transport_tcp/src/transport.rs
```diff
@@ -106,9 +161,29 @@ impl TcpTransport {
     /// let tcp = TcpTransport::create(&ctx).await?;
     /// tcp.listen("127.0.0.1:8000").await?;
     /// # Ok(()) }
-    pub async fn listen<S: AsRef<str>>(&self, bind_addr: S) -> Result<SocketAddr> {
+    pub async fn listen(&self, bind_addr: impl AsRef<str>) -> Result<(SocketAddr, Address)> {
         let bind_addr = parse_socket_addr(bind_addr.as_ref())?;
-        self.router_handle.bind(bind_addr).await
+        // Could be different from the bind_addr
```
etorreborre: Could we explain why/how it could be different?

implementations/rust/ockam/ockam_transport_tcp/src/transport.rs
```diff
@@ -106,9 +161,29 @@ impl TcpTransport {
...
+            bind_addr,
+            Arc::new(LocalSourceOnly),
+            Arc::new(LocalOnwardOnly),
+        )
+        .await?;
+
+        Ok((socket_addr, address))
+    }
+
+    /// Interrupt an active TCP connection given it's `Address`
```
etorreborre: s/it's/its

implementations/rust/ockam/ockam_transport_tcp/src/transport.rs
```diff
@@ -106,9 +161,29 @@ impl TcpTransport {
...
+    pub async fn disconnect(&self, address: &Address) -> Result<()> {
+        self.ctx.stop_worker(address.clone()).await
+    }
+
+    /// Interrupt an active TCP listener given it's `Address`
```
etorreborre: s/it's/its

implementations/rust/ockam/ockam_transport_tcp/src/workers/sender.rs
```diff
@@ -58,120 +47,164 @@ pub(crate) enum TcpSendWorkerMsg {
...
+        receiver_outgoing_access_control: Arc<dyn OutgoingAccessControl>,
+    ) -> Result<Address> {
         trace!("Creating new TCP worker pair");
-        let (sender, pair) = Self::new_pair(router_handle, stream, peer, hostnames).await?;
-
-        // Allow messages routed from Tcp Router
-        let tx_mailbox = Mailbox::new(
-            pair.tx_addr(),
-            Arc::new(AllowSourceAddress(tcprouter_main_addr)),
+        let sender_worker = Self::create(
```
etorreborre, line 140: This is a bit confusing. We create a `sender_worker` and we pass it a `receiver_outgoing_access_control`. Is it a typo? No, because the `TcpSendWorker` actually creates and starts a `TcpRecvProcessor` inside its `initialize` method. 

Why don't we do both the creation of the sender and the receiver inside of the `start` function, and use the required access control there? There will be more symmetry in the code and less indirection IMO.

implementations/rust/ockam/ockam_transport_tcp/tests/keepalive.rs
```diff
@@ -0,0 +1,48 @@
+use ockam_core::compat::rand::{self, Rng};
+use ockam_core::{route, Result};
+use ockam_node::Context;
+use ockam_transport_tcp::TcpTransport;
+use std::time::Duration;
+use tracing::info;
+
+#[ignore]
+#[ockam_macros::test(timeout = 400000)]
+async fn tcp_keepalive_test(ctx: &mut Context) -> Result<()> {
```
etorreborre, line 10: Nice test!

etorreborre (dismissed):
Nice work, I really like the simplification of the routing!

I left many comments, some of them which I think can really help with future maintenance. You can choose to tackle some of them in that PR or in follow-up PRs if you want. We can also discuss about them if some of them are not clear.

implementations/rust/ockam/ockam_transport_tcp/src/workers/sender.rs
SanjoDeundiak: Nice catch. That was impossible before I moved TcpStream creation out of TcpSendWorker. Now should be possible

documentation/guides/rust/get-started/04-transport/README.md
SanjoDeundiak: I will review examples and put everything in the correct order

documentation/guides/rust/get-started/05-secure-channel/README.md
SanjoDeundiak: We should revisit these guides and figure out if they make sense, but I don't want to do that in that PR

examples/rust/get_started/examples/04-routing-over-transport-two-hops-middle.rs
SanjoDeundiak: You can create an issue for that, otherwise this feedback will get lost

implementations/rust/ockam/ockam_api/src/cloud/mod.rs
SanjoDeundiak: @adrianbenavides @CAOakleyII

implementations/rust/ockam/ockam_api/src/nodes/service.rs
SanjoDeundiak: @adrianbenavides @CAOakleyII

implementations/rust/ockam/ockam_api/src/session/util.rs
SanjoDeundiak: @adrianbenavides @CAOakleyII

implementations/rust/ockam/ockam_api/src/util.rs
SanjoDeundiak: @adrianbenavides @CAOakleyII

implementations/rust/ockam/ockam_api/src/util.rs
SanjoDeundiak: @adrianbenavides @CAOakleyII

implementations/rust/ockam/ockam_command/src/authenticated.rs
SanjoDeundiak: I think command implementation should be rewritten to new format

SanjoDeundiak: @adrianbenavides @CAOakleyII

implementations/rust/ockam/ockam_command/src/tcp/connection/create.rs
SanjoDeundiak: @adrianbenavides @CAOakleyII

implementations/rust/ockam/ockam_command/src/tcp/connection/create.rs
SanjoDeundiak: @adrianbenavides @CAOakleyII

implementations/rust/ockam/ockam_command/src/util/mod.rs
SanjoDeundiak: That way we'll end up with some global shared state. If a client code has problems with having access to the shared `TcpTransport` instance, that probably means the client code needs a refactor

implementations/rust/ockam/ockam_command/src/util/mod.rs
SanjoDeundiak: @adrianbenavides @CAOakleyII

implementations/rust/ockam/ockam_core/src/routing/address.rs
SanjoDeundiak: I'll remove the `TODO` part :)

implementations/rust/ockam/ockam_multiaddr/src/proto.rs
SanjoDeundiak: @adrianbenavides I was always curious as well :D

implementations/rust/ockam/ockam_node/src/context.rs
SanjoDeundiak: It should include the main address in the list

implementations/rust/ockam/ockam_transport_tcp/src/lib.rs
SanjoDeundiak: 100% Agree

implementations/rust/ockam/ockam_transport_tcp/src/registry.rs
SanjoDeundiak: Love that

implementations/rust/ockam/ockam_transport_tcp/src/workers/sender.rs
SanjoDeundiak: I'll do it in another PR

etorreborre:
@adrianbenavides @CAOakleyII I'm resolving my comments so that we can proceed with that PR. I will create separate issues to tackle the proposed modifications / refactorings if that's ok with you

etorreborre (dismissed):

polvorin (dismissed):

polvorin (approved):

implementations/rust/ockam/ockam_multiaddr/src/proto.rs
adrianbenavides: Me too I'm afraid. I'm pretty sure they are random numbers and we started with the magic ockam number (62526 was it?). I'd keep it monotonic.

adrianbenavides:
> @adrianbenavides @CAOakleyII I'm resolving my comments so that we can proceed with that PR. I will create separate issues to tackle the proposed modifications / refactorings if that's ok with you

Just to keep the actionable comments in a single comment:

- https://github.com/build-trust/ockam/pull/4172#discussion_r1119863015 -> https://github.com/build-trust/ockam/issues/4388
- https://github.com/build-trust/ockam/pull/4172#discussion_r1119875952 -> https://github.com/build-trust/ockam/issues/4442
- https://github.com/build-trust/ockam/pull/4172#discussion_r1119906096 -> https://github.com/build-trust/ockam/issues/4389
- https://github.com/build-trust/ockam/pull/4172#discussion_r1119909628
- https://github.com/build-trust/ockam/pull/4172#discussion_r1119912671 -> https://github.com/build-trust/ockam/issues/4390
- https://github.com/build-trust/ockam/pull/4172#discussion_r1119915225
- https://github.com/build-trust/ockam/pull/4172#discussion_r1120084471 -> https://github.com/build-trust/ockam/issues/4391
- https://github.com/build-trust/ockam/pull/4172#discussion_r1120087117
- https://github.com/build-trust/ockam/pull/4172#discussion_r1120116376
- https://github.com/build-trust/ockam/pull/4172#discussion_r1120125291 -> https://github.com/build-trust/ockam/issues/4392

@etorreborre since next week I'll be managing the open-source side, I can review all the comments and create the necessary issues. I think some of them could be https://github.com/build-trust/ockam/labels/good%20first%20issue 's.

etorreborre:
Thanks @adrianbenavides for leaving a summary and taking care of the follow-up!

implementations/rust/ockam/ockam_multiaddr/src/proto.rs
mrinalwadhwa: I think these have to do with the fact that multiaddr address types need a number that is unlikely to conflict with other address types ... why DNS is 56.
I suspect there might be a doc in the multiaddr spec specifying these.

mrinalwadhwa: yup https://github.com/multiformats/multiaddr/blob/master/protocols.csv

adrianbenavides: We could still add this link to the code as a reference, pick a high enough number and start from there.

Issue #1819, opened 2021-09-02
Module structure of the Rust libraries needs cleaning up


thomcc:
*Caveat: this is subjective and quite broad.*

When updating the example projects as part of #1816, I noticed a number of cases where the organization<sup>1</sup> of different parts of the ockam rust libraries felt... suboptimal.

That is, this is just about how we group/organize things, and not about the actual APIs themselves. Specifically it feels very much like it grew organically rather than was deliberately designed.

## Problems

That said, here are a couple of (mostly-) concrete problems I saw after spending a couple hours updating the examples:

1. There are two `ockam::stream` modules, one is from `ockam/src/stream.rs` and the other is from `ockam/src/protocols/stream/mod.rs`, which is re-exposed by a `pub use protocols::*;` in the crate root (e.g. `ockam/src/lib.rs`). Unfortunately, the second of these rendered inaccessable<sup>2</sup> by the first (note that `protocols` is not a public module).

    I suspect we have similar issues elsewhere, or could easily get them accidentally in the future. (I wish we could lint<sup>3</sup> for this one, honestly)

2. The subset of things that different crates re-export from eachother seems fairly arbitrary, and isn't done consistently
    - `ockam_core` vs `ockam` -- `ockam` re-exports a subset of `ockam_core`, and does not expose `ockam_core`.
    - `ockam_vault_core` vs `ockam_vault` -- `ockam_vault` re-exports a subset of `ockam_value_core`, but also has `pub extern trait ockam_vault_core` (exposing ockam_vault_core via the somewhat repetitive `ockam_vault::ockam_vault_core` path).

3. We have a lot of types with colliding names, — a regex search says that we have at least 6 `struct Signature`s, for example.

   This isn't inherently bad — I think some of these make sense, but some of them are redundant, and others can and do cause collisions (in the examples, anyway).

   However, we should generally assume users are going to glob import the types from the traits that they need to use, and see if we can avoid that.

   Note: This one probably requires more thought...

I need to use the library a bit more, since I'm sure there's more I just didn't notice. There's also a few cases where I'm pretty sure some of the types are in a weird/seemingly-wrong crate, but I need to look more to say for certain. 

## Rough / Opinionated notes

Here are some rough thoughts that reflect some of my opinions on how we should maybe do things. Most of this is not specific to ockam at all.

- Stuff like `ockam_foo`/`ockam_foo_core` should probably either:
[...]

PR #4179, opened 2023-02-01
refactor: #1819 signature and secret renamings

etorreborre:
The purpose of this PR is to get a more consistent terminology:

 - A `Signature` is a just a list of bytes transferred over the wire
 - An `IdentityChangeSignature` is a signature authenticating an `IdentityChange`. This gives it more context and avoid the naming clashes identified in #1819
 
 - the `SecretVault` interface has been renamed to `KeyVault` since it is dealing with keys and returning `KeyIds`. So instead of "generating a secret and getting back a key id" we "generate a key and get back a key id"
 - this change has been extended to `KeyAttributes` (instead of `SecretAttributes` for the attributes of a key)
 - the above renaming clashed with the existing `KeyAttribute` struct (containing both a key label and some key attributes). That structure is actually not very useful since a key label is used independently from the key attributes:
     - the key label is used in conjunction with identity changes
     - while the key attributes are used with the Vault API which doesn't know about key labels


implementations/rust/ockam/ockam_core/src/vault/types.rs
```diff
@@ -70,33 +70,33 @@ cfg_if! {
     }
 }
 
-/// Binary representation of a Secret.
+/// Binary representation of a Private key.
 #[derive(Serialize, Deserialize, Clone, Zeroize, Encode, Decode)]
 #[zeroize(drop)]
 #[cbor(transparent)]
-pub struct SecretKey(#[n(0)] SecretKeyVec);
+pub struct PrivateKey(#[n(0)] PrivateKeyVec);
```
SanjoDeundiak, line 77: Could you please elaborate what is the reasoning on that renaming?

implementations/rust/ockam/ockam_core/src/vault/types.rs
```diff
@@ -183,11 +183,11 @@ impl From<Signature> for SignatureVec {
     }
 }
 
-/// All possible [`SecretType`]s
+/// All possible [`KeyType`]s
 #[derive(Serialize, Deserialize, Copy, Clone, Debug, Encode, Decode, Eq, PartialEq, Zeroize)]
 #[rustfmt::skip]
 #[cbor(index_only)]
-pub enum SecretType {
+pub enum KeyType {
```
SanjoDeundiak, line 190: I'm not sure renaming Secret -> Key is technically correct. Key is more narrow term, for example I don't think a Buffer with sensitive data can be called a Key

implementations/rust/ockam/ockam_identity/src/change/create_key.rs
```diff
@@ -2,23 +2,29 @@ use crate::authenticated_storage::AuthenticatedStorage;
...
+use ockam_core::vault::{KeyAttributes, KeyId, PublicKey};
 use ockam_core::{Encodable, Result};
 use serde::{Deserialize, Serialize};
 
 /// Key change data creation
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct CreateKeyChangeData {
     prev_change_id: ChangeIdentifier,
-    key_attributes: PrivateKeyAttributes,
+    key_label: String,
```
SanjoDeundiak, line 16: That would probably break backwards compatibility and therefore break Ockam Orchestrator

implementations/rust/ockam/ockam_core/src/vault/types.rs
```diff
@@ -244,6 +244,15 @@ impl KeyAttributes {
             length,
         }
     }
+
+    /// Create default key attributes
+    pub fn default_attributes() -> Self {
```
SanjoDeundiak, line 249: I think being explicit when you generate a secret could be very useful

SanjoDeundiak (changes requested):

implementations/rust/ockam/ockam_core/src/vault/types.rs
etorreborre: That's the one presented in the description, the fact that the `SecretVault` API talks about secrets but return key ids. It would be better to make the naming consistent. What would you propose to solve this discrepancy?

implementations/rust/ockam/ockam_identity/src/change/create_key.rs
etorreborre: Interesting, that raises 2 questions in my mind:

 1. do we have tests to check the compatibility with Orchestrator? Because they should have been broken with that change
 2. we should probably have a data structure that is purely used to communicate with the Orchestrator. Are we sending the full structure of changes as-is to the Orchestrator? Where is that done?

implementations/rust/ockam/ockam_core/src/vault/types.rs
etorreborre: Do you mean that we should actually *not* have such a function?

implementations/rust/ockam/ockam_core/src/vault/types.rs
etorreborre: Yes and that might be a hint that we should separate this data from the rest. For example in `hasher_impl.rs` we try to retrieve the salt and [fail at runtime if it doesn't have the right type](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_vault/src/hasher_impl.rs#L52). Maybe this is an indication that this kind of data (that is always `Ephemeral` as far as I understand) should be stored separately inside a `Vault`.

implementations/rust/ockam/ockam_core/src/vault/types.rs
SanjoDeundiak: Yes

implementations/rust/ockam/ockam_core/src/vault/types.rs
SanjoDeundiak: Oh, I thought renaming SecretVault -> KeyVault and SecretKey -> PrivateKey are two separate things, are they related?

implementations/rust/ockam/ockam_identity/src/change/create_key.rs
mrinalwadhwa: Change history is overly complex, we can break backward compatibility here once but let's do it soon and in one sweep.

implementations/rust/ockam/ockam_core/src/vault/types.rs
SanjoDeundiak: Agree, having different types for these things may be a great idea, but before we have that I'm not sure this renaming  makes sense to me

implementations/rust/ockam/ockam_identity/src/change/create_key.rs
SanjoDeundiak: We use rust sidecar for Identity implementation, so there are no second implementation and no second structure for `Identity`. The problem is that we have existing persistent identities in our system (e.g. Orchestrator has its Identity stored somewhere in its db), which probably won't deserialize because of the changed structure. Unfortunately, there is no backwards compatibility test

davide-baldo: Can't we create one right now by hardcoding a fully populated encoded Identity in the test?

etorreborre:
I'm closing this PR for now. In particular since it highlighted the need to do some preparation work on:

 - the serialization of identity data with other nodes in a versioned way
 - possibly the management of non-key, ephemeral, data outside of the vault

PR #4183, opened 2023-02-01
fix(rust): named identities allowed on lease command, add test cases

polvorin:
named identites are now supported on `message send` and `lease *` commands. A proper credential is requested and used in these cases.  New test cases added.

Left some FIXME and TODO for things that are workarounds,   or that need to be fixed.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/message/send.rs
+++ b/implementations/rust/ockam/ockam_command/src/message/send.rs
@@ -8,7 +8,7 @@ use ockam_api::nodes::service::message::SendMessage;
 use ockam_core::api::{Request, RequestBuilder};
 use ockam_multiaddr::MultiAddr;
 
-use crate::node::util::{delete_embedded_node, start_embedded_node};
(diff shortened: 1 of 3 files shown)


caoakleyii:
Thanks for adding the FIXME comments to the `with_embedded_node` method.

I don't see any urgent issues with this PR; If you're okay with it, I'd like to just fix that method up, so that order doesn't matter.

But other than that LGTM.

polvorin:
@CAOakleyII  :+1:  lets say we want this merged tomorrow to think about doing another release with the fix.     So not merging for now,  if you can fix that ordering issue lets merge once that's incorporated.   If we didn't finish that for tomorrow,  lets merge this as-is  and fix the ordering on a separate PR  (that is less urgent,  but still needed as otherwise we will  shot ourselves on foot sooner or latter)

polvorin:
@Mergifyio rebase

polvorin:
@Mergifyio rebase

PR #4156, opened 2023-01-30
refactor(rust): recipient returns an error instead of a panic

MavenRain:
This PR takes care of a TODO calling for Route::recipient to return a Result<Address> instead of lying by returning Address but being a partial function that panics under certain situations.

Clippy also complained about a bunch of redundant cloning, so those are addressed here as well, mainly in local.rs and onward.rs.

This is the second attempt at this PR, and the discussion from the first attempt can be found here: https://github.com/build-trust/ockam/pull/4142

PR #4188, opened 2023-02-02
refactor(rust): recipient returns an error instead of panicking

MavenRain:
This PR takes care of a TODO calling for Route::recipient to return a Result

Instead of lying by returning Address but being a partial function that panics under certain situations.

Clippy also complained about a bunch of redundant cloning, so those are addressed here as well, mainly in local.rs and onward.rs.

This is the third attempt at correctly creating the PR.  The second can be found here: https://github.com/build-trust/ockam/pull/4156

### DIFF
--- a/implementations/rust/ockam/ockam/src/channel/listener.rs
+++ b/implementations/rust/ockam/ockam/src/channel/listener.rs
@@ -59,7 +59,7 @@ impl Worker for ChannelListener {
             rx_int_addr,
             tx_int_addr,
         } = msg.as_body();
-        let peer_channel_addr = msg.return_route().recipient();
+        let peer_channel_addr = msg.return_route().recipient()?;
 
         let peer_rx_base: Route = msg.return_route().modify().pop_back().into();
         let peer_rx_pub = peer_rx_base.clone().modify().append(rx_addr.clone()).into();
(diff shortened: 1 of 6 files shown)


MavenRain:
@etorreborre I think I don't understand this message primarily because I am a Git newbie.

```
Run unverified=$(cat commits.json | jq --raw-output '.[] | [.sha, .commit.verification.verified] | @csv' | grep false || echo '')

    We require that all commits in a pull request are signed and Verified by Github

    Please read about signing commits at:
    https://docs.github.com/en/authentication/managing-commit-signature-verification

    ERROR: The following commits are not Verified by Github.
  
"[33](https://github.com/build-trust/ockam/actions/runs/4075282813/jobs/7024712088#step:15:34)76963e214558aabebe7d438d392baec7c8a85c",false
```

etorreborre:
@MavenRain you indeed need to go to this page which explains the process of signing commits: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits. The idea is that you generate a private/public key pair, you upload the public key to github and locally you sign all your commits with the private key. Then github can verify that those commits were indeed created by you and will show them with a "verified" checkmark. The good thing is that, once it is in place, all your github commits will always be signed!

I am afraid that this means that you need to effectively recreate your commits :-( (there is a way to do it with `git rebase -i` but maybe the easiest thing is just to recreate them). Thanks for your perseverance!

mrinalwadhwa:
@MavenRain @etorreborre I pushed a signed commit on @MavenRain behalf.
@etorreborre if the changes look good to you we can merge.

etorreborre:
@mrinalwadhwa thanks for signing the commit. I will merge as soon as you make me part of the approvers list :-)

mrinalwadhwa (approved):

PR #4216, opened 2023-02-07
feat(rust): NodeState implement check whether a node is running

### LINKED ISSUE #4174: Implement a function on the `NodeState` to check whether a node is running
## Current

There is no explicit way to easily check if a node is running or is stopped without sending a message to it.

## Desired

There are two potential solutions: one involves messaging the node, and another checking the node's state. In either case, we need a function on the `NodeState` that checks a node status given its name.

### Messaging

This solution consists of sending a `short_identity` request to the node with a [short timeout](https://github.com/build-trust/ockam/blob/6d21ff5e3a6180b6de486b75cd09d2ddfd645f54/implementations/rust/ockam/ockam_command/src/util/mod.rs#L160). If the request receives an Ok response, then the node is up, otherwise it's not.

### Node state

When a node is executing as a background process, [we store the `pid` in a file in its state directory](https://github.com/build-trust/ockam/blob/6d21ff5e3a6180b6de486b75cd09d2ddfd645f54/implementations/rust/ockam/ockam_api/src/cli_state.rs#L702). This can be done by checking the existence of this file and, if possible, if the PID exists in the system. 

---

andrescaroc:
## Current Behavior

There is no explicit way to easily check if a node is running or is stopped without sending a message to it.
Resolves https://github.com/build-trust/ockam/issues/4174

## Proposed Changes

NodeState implement a function called `is_running` with fn signature:
 ```
pub fn is_running(&self) -> bool {}
 ```
That returns `true` if pid file is in its state directory and the process is running in the system `/proc/{pid}/status`, otherwise returns `false`

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state.rs
@@ -12,6 +12,7 @@ use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::SystemTime;
+use sysinfo::{Pid, System, SystemExt};
 
 use crate::lmdb::LmdbStorage;
 use ockam::compat::tokio;
@@ -733,6 +734,16 @@ impl NodeState {
     pub fn set_pid(&self, pid: i32) -> Result<()> {
         std::fs::write(self.path.join("pid"), pid.to_string())?;
         Ok(())
+    }
+
+    pub fn is_running(&self) -> bool {
+        if let Ok(Some(pid)) = self.pid() {
+            let mut sys = System::new();
+            sys.refresh_processes();
+            sys.process(Pid::from(pid as usize)).is_some()
+        } else {
+            false
+        }
     }
 
     pub fn stdout_log(&self) -> PathBuf {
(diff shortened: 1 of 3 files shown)


implementations/rust/ockam/ockam_api/src/cli_state.rs
```diff
@@ -735,6 +735,15 @@ impl NodeState {
         Ok(())
     }
 
+    pub fn is_running(&self) -> bool {
+        if let Ok(Some(pid)) = self.pid() {
+            let system_path = PathBuf::from(format!("/proc/{}/status", pid));
```
adrianbenavides: This would only work for linux... you can try with `sysinfo`. It has a method to iterate through the running processes that you can filter.

andrescaroc: Yes, that make sense. Thank you for the guidance. I just pushed the changes with `sysinfo`. Let me know what do you think.

andrescaroc: @adrianbenavides when you find some time feel free to verify and provide feedback

adrianbenavides (approved):
Looks perfect now, thank you so much @andrescaroc for an awesome first contribution! 🚀

PR #4220, opened 2023-02-07
feat(elixir): add unix domain sockets transport and getting-started example

tzumby:
## Current Behavior

Currently there is no implementation of Unix Domain Socket Transports for Elixir implementation.

## Proposed Changes

This PR is still Work in Progress

- [x] Prototype the unix domain sockets transport
- [x] Add examples
- [ ] Add tests
- [ ] Clean up sockets after crashes / shutdown
- [ ] Allow multiple UDS gen_tcp servers to run (currently using named processes for the Connection supervision)

```
       +-----------------+                                                                 
       | UDS(Supervisor) |                                                                 
       +--------/----\---+                                                                 
               /      \                                                                    
              /        \                                                                   
             /       +-----------+                                                         
            /        | Listener  --------|                                                 
           /         +-----------+       |                                                 
          /                              | Start Connection children                       
   +-----------------------+             |                                                 
   |Connection Supervisor  --------------|                                                 
   |(dynamic)              |                                                               
   +------/----------\-----+                                                               
         /            \                                                                    
        /              -\                                                                  
       /                 \                                                                 
      /                   \                                                                
+--------------------++--------------------+                                               
| Connection Process || Connection Process |                                               
|                    ||                    |                                               
| +--------+         || +--------+         |                                               
| | socket |         || | socket |         |                                               
| +--------+         || +--------+         |                                               
+--------------------++--------------------+                                               
                                                      
```

## Checks

### DIFF
--- a/examples/elixir/get_started/06-secure-channel-over-two-uds-transport-hops-initiator.exs
+++ b/examples/elixir/get_started/06-secure-channel-over-two-uds-transport-hops-initiator.exs
@@ -4,11 +4,15 @@
 Ockam.Node.register_address("app")
 
 # Start the TCP Transport Add-on for Ockam Routing.
-Ockam.Transport.UDS.start()
+Ockam.Transport.UDS.start(path: "/tmp/hop1.sock")
+Ockam.Transport.UDS.start(path: "/tmp/hop2.sock")
 
 # Create a vault and an identity keypair.
 {:ok, vault} = Ockam.Vault.Software.init()
 {:ok, identity} = Ockam.Vault.secret_generate(vault, type: :curve25519)
+
+{:ok, hop1} = Ockam.Transport.UDS.Client.create(path: "/tmp/hop1.sock")
+{:ok, hop2} = Ockam.Transport.UDS.Client.create(path: "/tmp/hop2.sock")
 
 # Connect to a secure channel listener and perform a handshake.
 alias Ockam.Transport.UDSAddress
--- a/examples/elixir/get_started/06-secure-channel-over-two-uds-transport-hops-initiator.exs
+++ b/examples/elixir/get_started/06-secure-channel-over-two-uds-transport-hops-initiator.exs
@@ -1,0 +1,27 @@
+["setup.exs"] |> Enum.map(&Code.require_file/1)
+
+# Register this process as worker address "app".
+Ockam.Node.register_address("app")
+
+# Start the TCP Transport Add-on for Ockam Routing.
+Ockam.Transport.UDS.start()
+
+# Create a vault and an identity keypair.
+{:ok, vault} = Ockam.Vault.Software.init()
+{:ok, identity} = Ockam.Vault.secret_generate(vault, type: :curve25519)
+
+# Connect to a secure channel listener and perform a handshake.
+alias Ockam.Transport.UDSAddress
+r = [UDSAddress.new("/tmp/sock1"), "h1", UDSAddress.new("/tmp/sock2"), "secure_channel_listener"]
+{:ok, c} = Ockam.SecureChannel.create(route: r, vault: vault, identity_keypair: identity)
+
+# Prepare the message.
+message = %{onward_route: [c, "echoer"], return_route: ["app"], payload: "Hello Ockam!"}
+
+# Route the message.
+Ockam.Router.route(message)
+
+# Wait to receive a reply.
+receive do
+  message -> IO.puts("Address: app\t Received: #{inspect(message)}")
+end
--- a/examples/elixir/get_started/06-secure-channel-over-two-uds-transport-hops-responder.exs
+++ b/examples/elixir/get_started/06-secure-channel-over-two-uds-transport-hops-responder.exs
@@ -1,0 +1,20 @@
+["setup.exs", "echoer.exs"] |> Enum.map(&Code.require_file/1)
+
+alias Ockam.Transport.UDSAddress
+
+# Create a Echoer type worker at address "echoer".
+{:ok, _echoer} = Echoer.create(address: UDSAddress.new("/tmp/sock2"))
+
+# Create a vault and an identity keypair.
+{:ok, vault} = Ockam.Vault.Software.init()
+{:ok, identity} = Ockam.Vault.secret_generate(vault, type: :curve25519)
+
+# Create a secure channel listener that will wait for requests to initiate an Authenticated Key Exchange.
+Ockam.SecureChannel.create_listener(
+  vault: vault,
+  identity_keypair: identity,
+  address: "secure_channel_listener"
+)
+
+# Start the TCP Transport Add-on for Ockam Routing and a TCP listener on port 4000.
+Ockam.Transport.UDS.start("/tmp/sock2")
--- a/implementations/elixir/ockam/ockam/lib/ockam/examples/routing/uds.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/examples/routing/uds.ex
@@ -1,0 +1,43 @@
+defmodule Ockam.Examples.Routing.UDS do
+  @moduledoc """
+  Example message routing through the UDS transport
+
+  server() - start UDS server with echoer
+  client() - establish a connection and send a message to the echoer and wait for response
+
+  """
+  alias Ockam.Examples.Echoer
+  alias Ockam.Transport.UDS
+
+  require Logger
+
+  def server do
+    ## Start a transport with listener on port
+    UDS.start(path: "/tmp/server.sock")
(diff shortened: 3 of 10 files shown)


examples/elixir/get_started/06-secure-channel-over-two-uds-transport-hops-initiator.exs
```diff
@@ -0,0 +1,27 @@
...
+# Start the TCP Transport Add-on for Ockam Routing.
+Ockam.Transport.UDS.start()
+
+# Create a vault and an identity keypair.
+{:ok, vault} = Ockam.Vault.Software.init()
+{:ok, identity} = Ockam.Vault.secret_generate(vault, type: :curve25519)
+
+# Connect to a secure channel listener and perform a handshake.
+alias Ockam.Transport.UDSAddress
+r = [UDSAddress.new("/tmp/sock1"), "h1", UDSAddress.new("/tmp/sock2"), "secure_channel_listener"]
```
hairyhum, line 19: We're trying to steer away from using implicit connections expressed in specific address types and towards explicit connection workers.
I think we can simplify example by first creating a connection worker like `{:ok, client} = UDS.Client.create("/tmp/sock1")` and then sending messages through it.
That removes support for multi-hop thing and simplifies the example as it would only have 2 nodes.

implementations/elixir/ockam/ockam/lib/ockam/transport/uds/address.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Ockam.Transport.UDSAddress do
```
hairyhum, line 1: As mentioned in other comment, we're trying to steer away from specific address types, so let's not add new ones.

implementations/elixir/ockam/ockam/lib/ockam/transport/uds.ex
```diff
@@ -0,0 +1,59 @@
...
+  def init(opts) do
+    children = [
+      ConnectionSupervisor,
+      {Listener, opts}
+    ]
+
+    Supervisor.init(children, strategy: :rest_for_one)
+  end
+
+  def handle_transport_message(message) do
```
hairyhum, line 29: If we don't use a special transport type, this becomes unnecessary.

implementations/elixir/ockam/ockam/lib/ockam/transport/uds.ex
```diff
@@ -0,0 +1,59 @@
...
+    with {:ok, destination, message} <-
+           pick_destination_and_set_onward_route(message),
+         {:ok, encoded_message} <- Wire.encode(message),
+         {:ok, socket} <-
+           :gen_tcp.connect({:local, destination}, 0, [:binary, active: true, reuseaddr: true]),
```
hairyhum, line 34: For connecting to a remote node, if we do that in the message handler, what happens with the connection?
Should we have some "client" worker to manage the connection in that case?

implementations/elixir/ockam/ockam/lib/ockam/transport/uds/connection.ex
```diff
@@ -0,0 +1,49 @@
...
+  defstruct [:socket]
+
+  @impl true
+  def init(socket) do
+    Logger.debug("Started connection handler")
+    {:ok, %__MODULE__{socket: socket}}
+  end
+
+  @impl true
+  def handle_info({:tcp, socket, data}, %__MODULE__{socket: socket} = state) do
```
hairyhum, line 26: This module currently only supports TCP->Ockam messages.
How do we reply to those messages?

In other transport handler workers we were tracing the address of the worker in the message return route (see `Ockam.Transport.TCP.Handler` around line 70).

I see this is not yet a worker, this comment is to keep in mind when you make it a worker.

implementations/elixir/ockam/ockam/lib/ockam/transport/uds/connection_supervisor.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Ockam.Transport.UDS.ConnectionSupervisor do
```
hairyhum, line 1: I like this approach.

implementations/elixir/ockam/ockam/lib/ockam/transport/uds/listener.ex
```diff
@@ -0,0 +1,40 @@
...
+  @default_socket "/tmp/ockam_uds_server.sock"
+
+  def start_link(opts \\ []) do
+    Task.start_link(__MODULE__, :run, [Keyword.get(opts, :path, @default_socket)])
+  end
+
+  def run(path) do
+    case :gen_tcp.listen(0, [
+           :binary,
+           {:active, true},
```
hairyhum, line 17: If we listen in the active mode, does it mean that it's possible to get connection messages on the wrong process if there is some race condition between accept and change of controlling process?

examples/elixir/get_started/06-secure-channel-over-two-uds-transport-hops-responder.exs
```diff
@@ -0,0 +1,20 @@
+["setup.exs", "echoer.exs"] |> Enum.map(&Code.require_file/1)
+
+alias Ockam.Transport.UDSAddress
+
+# Create a Echoer type worker at address "echoer".
+{:ok, _echoer} = Echoer.create(address: UDSAddress.new("/tmp/sock2"))
```
hairyhum, line 6: I understand it's WIP, but the `address` option we pass to the worker would be this worker address.
Passing transport addresses in this option is not going to work because transport addresses are handled differently. And I believe that wasn't intended behaviour in the first place.

hairyhum (commented):
Hi. Thanks for contributing!

The architecture makes sense.

Left some comments. Mostly related to the address type implementation.
We're currently working to phase out and eventually deprecate this feature in favour of connection workers. This would simplify transport implementation and examples.

The way it should work is that you would create a connection worker like:
`{:ok, client} = UDS.Client.create(uds_address)`

And then send messages to this client address instead of the UDAddress:
`Ockam.Router.route(%Ockam.Message{onward_route: [client, "echo"], payload: "HI"})`

See `Ockam.Transport.TCP.Client`

Also added some notes about connection management.

Let us know if you have any questions about this approach or how to make ockam workers out of gen servers.

examples/elixir/get_started/06-secure-channel-over-two-uds-transport-hops-initiator.exs
tzumby: Thanks @hairyhum, this is great feedback! I went back and did some more reading and implemented an `Ockam.Transport.UDS.Client` and used it in a [Routing example](https://github.com/build-trust/ockam/blob/71c1de648282e7e081ff2f5c875f10203fa93b72/implementations/elixir/ockam/ockam/lib/ockam/examples/routing/uds.ex), similar to how you specified it in your main comment. I still have to clean up the get_started examples and the UDSAddress.

implementations/elixir/ockam/ockam/lib/ockam/transport/uds/listener.ex
tzumby: Good point, perhaps one way to deal with this would be to start with passive or active once and switch it once the Connection child is up.

implementations/elixir/ockam/ockam/lib/ockam/examples/routing/uds.ex
```diff
@@ -0,0 +1,43 @@
...
+
+    ## Register this process to receive messages
+    my_address = "example_run"
+    Ockam.Node.register_address(my_address)
+
+    {:ok, server_host_address} = Ockam.Transport.UDS.Client.create(path: "/tmp/server.sock")
+    {:ok, hop} = Ockam.Transport.UDS.Client.create(path: "/tmp/client.sock")
+
+    Ockam.Router.route(%{
+      onward_route: [server_host_address, hop, "echoer"],
```
hairyhum, line 31: This route sends a message through server uds transport and then a "client.sock" uds transport and then to "echo".
Although that should probably work, it can be a bit confusing.
I think `hop` and "client.sock" can be removed to make a simpler example.

etorreborre:
Hi @tzumby I can see it has been a while since this PR has been opened. Is there anything we can do to help you finish it?

tzumby:
Hey @etorreborre, thanks for messaging, I would love to push this over the finish line, but I'm absolutely swamped with work. If this is blocking anything maybe it makes sense that I either close it or have someone else co-author this with me ?

etorreborre:
@tzumby no worries this is not on the critical path at the moment but we will see if we can maybe find someone to help you out with this.

PR #4223, opened 2023-02-08
Showcase the use of of credentials with the Rust library

etorreborre:
## Current Behavior

We currently a command-line example showing how to use authenticated attributes to [create a TCP portal between a device and its server, using a relay](https://docs.ockam.io/use-cases/apply-fine-grained-permissions-with-attribute-based-access-control-abac).

## Proposed Changes

This PR provides an implementation of this scenario only using the Rust library.
In order to support it some code had to be added, with dependencies to additional crates, to:

  - retrieve project information
  - create a one time token (because the CBOR serialization needed to retrieve credentials is very specific)

The next PR will propose a refactoring so that:

  1. the `OneTimeCode` struct is moved to the `ockam_identity` crate so that it becomes available by using the `ockam` crate
  2. 
  3. the example can use the `ockam_command` crate to read the project information (see the `ProjectInfo` struct in `ockam_command`) from a JSON file. This is mostly for convenience. It would also be possible to manually parse the data contained into the project JSON file into the necessary routes and identities

### DIFF
--- a/examples/rust/get_started/examples/11-attribute-based-authentication-edge-plane.rs
+++ b/examples/rust/get_started/examples/11-attribute-based-authentication-edge-plane.rs
@@ -20,8 +20,6 @@ use std::time::Duration;
 ///
 #[ockam::node]
 async fn main(ctx: Context) -> Result<()> {
-    env::set_var("OCKAM_LOG", "trace");
-
     // Use the TCP transport
     let tcp = TcpTransport::create(&ctx).await?;
 
--- a/examples/rust/get_started/examples/11-attribute-based-authentication-edge-plane.rs
+++ b/examples/rust/get_started/examples/11-attribute-based-authentication-edge-plane.rs
@@ -3,7 +3,6 @@ use ockam::identity::credential::Credential;
 use ockam::identity::{Identity, TrustEveryonePolicy, TrustMultiIdentifiersPolicy};
 use ockam::{route, vault::Vault, Context, Result, TcpTransport};
 use ockam_core::IncomingAccessControl;
-use std::env;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -18,8 +17,28 @@ use std::time::Duration;
 ///   - create a TCP inlet with some access control checking the authenticated attributes of the caller
 ///   - connect the TCP inlet to a server outlet
 ///
+/// The node needs to be started with:
+///
+///  - a project.json file created with `ockam project information --output json  > project.json`
+///  - a token created by an enroller node with `ockam project enroll --attribute component=edge`
+///
 #[ockam::node]
 async fn main(ctx: Context) -> Result<()> {
+    // create a token (with `ockam project enroll --attribute component=edge`)
+    // In principle this token is provided by another node which has enrolling privileges for the
+    // current project
+    let token: OneTimeCode = create_token("component", "edge").await?;
+    println!("token: {token:?}");
+
+    // set the path of the project information file
+    // In principle this file is provided by the enrolling node by running the command
+    // `ockam project information --output json  > project.json`
+    let project_information_path = "project.json";
+    start_node(ctx, project_information_path, token).await
+}
+
+/// start the edge node
+async fn start_node(ctx: Context, project_information_path: &str, token: OneTimeCode) -> Result<()> {
     // Use the TCP transport
     let tcp = TcpTransport::create(&ctx).await?;
 
@@ -27,11 +46,11 @@ async fn main(ctx: Context) -> Result<()> {
     let vault = Vault::create();
     let edge_plane = Identity::create(&ctx, &vault).await?;
 
-    // 1. create a secure channel to the authority
+    // 2. create a secure channel to the authority
     //    to retrieve the node credentials
 
-    // Import the authority identity and route from the project.json file
-    let project = import_project("project.json", &vault).await?;
+    // Import the authority identity and route from the information file
+    let project = import_project(project_information_path, &vault).await?;
 
     // create a secure channel to the authority
     // when creating the channel we check that the opposite side is indeed presenting the authority identity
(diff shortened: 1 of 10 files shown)


examples/rust/get_started/examples/11-attribute-based-authentication-edge-plane.rs
```diff
@@ -0,0 +1,116 @@
...
+///
+/// This example shows how to:
+///
+///   - retrieve credentials from an authority
+///   - create a TCP inlet with some access control checking the authenticated attributes of the caller
+///   - connect the TCP inlet to a server outlet
+///
+#[ockam::node]
+async fn main(ctx: Context) -> Result<()> {
+    env::set_var("OCKAM_LOG", "trace");
```
SanjoDeundiak: Should it stay here?

examples/rust/get_started/examples/11-attribute-based-authentication.rs
```diff
@@ -0,0 +1,91 @@
...
+///                     ^          to control
+///                     |         ^       |
+///  secure channel to  |  create |       |
+///  project            |         |       v
+///                     |         |    "untrusted"
+///                   +--------------secure channel listener--+
+///                   |                                       |
+///                   | Control node                          | "outlet" --> 127.0.0.1:5000
+///                   |                                       |
+///                   +---------------------------------------+
```
polvorin: This looks good!.
One comment,   there is one role missing here, the `enroller`  .    get_credentials  can be used _after_   that identity is a member,  and that can be done only by the `enroller`  (either  by using a one-time-token or directly adding the identity identifier of the new member).    

In the example,  looks like  *both*  the edge and control node are able to act as  `enroller`  due to the underling identity on the _machine_ they run.

I find that a bit confusing and hiding how and _why_  things works. 
Would it be better if instead it's done as:

```
$ ockam enroll
$ ockam project information --output json  > project.json
$ ockam project enroll --attribute  component=edge
     -> gives EDGE_ENROLLMENT_TOKEN
$ ockam project enroll --attribute  component=control_plane
     > gives CONTROL_PLANE_ENROLLMENT_TOKEN
```

Now,  from any machine,  you can start:    
```
 $ attribute-based-authentication-control-plane   project.json   $CONTROL_PLANE_ENROLLMENT_TOKEN
 $ attribute-based-authentication-edge   project.json   $EDGE_ENROLLMENT_TOKEN
 ```

if we want to make it simple to run locally,    a bash script can be used   (that do the enroll,  generate the tokens,   and start the components passing the right argument).  That bash script will server as doc of how things are done in general

File: examples/rust/get_started/src/attribute_access_control.rs [hunk 3]
```diff
@@ -0,0 +1,93 @@
...
+    fn new(attributes_storage: S, expression: Expr) -> Self {
+        Self {
+            attributes_storage,
+            expression,
+        }
+    }
+}
+
+#[async_trait]
+impl<S> IncomingAccessControl for AbacAccessControl<S>
```
SanjoDeundiak, line 56: I wonder what are the difference between this and the original implementation

examples/rust/get_started/examples/11-attribute-based-authentication.rs
```diff
@@ -0,0 +1,91 @@
...
+///                   | Edge node               | Inlet <-- 127.0.0.1:7000
+///                   |                         | connected to "outlet" via secure channel
+///                   +-------------------------+
+///                     |                 |
+///  secure channel to  |                 | create secure channel to control
+///  project            |                 | via the forwarder
+///                     v                 |
+///                   +-------------------+-------+                 +--------------+
+///                   | Orchestrator      |       | get credentials | Authority    |
+///                   |                   |       |---------------->|              |
```
SanjoDeundiak: Aren't nodes talking to the Authority directly?

SanjoDeundiak (dismissed):

polvorin (changes requested):
This looks amazing!.   And very well documented :clap: 

I requested a change to eliminate some _magic_   and be more explicit about  how nodes gets enrolled into the project,  as that's a key point. 
The same end result  (not needing manual intervention to enroll things when doing everything from the admin' machine) can still be provided with the bash script suggested on the comment

examples/rust/get_started/examples/11-attribute-based-authentication.rs
polvorin: good point, yes

examples/rust/get_started/examples/11-attribute-based-authentication-edge-plane.rs
etorreborre: Right, I've removed it.

examples/rust/get_started/src/attribute_access_control.rs
etorreborre: This one is a simpler implementation which does not store the policy

examples/rust/get_started/examples/11-attribute-based-authentication.rs
etorreborre: Fixed, thanks.

examples/rust/get_started/src/attribute_access_control.rs
SanjoDeundiak: What about whitespaces?

examples/rust/get_started/examples/11-attribute-based-authentication.rs
etorreborre: Thanks @polvorin for your comment. It is indeed important to show that the enrollment is a separate phase of the process. I tried a different approach to show this though in order to stay inside the development environment as much as possible. I separated the code between a setup phase where we find out what the token and the project information should be and a `start_node` phase where we start the node with this information.

I did this because I found very useful to be able to stay in the IDE while developing this scenario, and even be able to access the debugger to understand what was wrong. So I think that this format will also help people iterate faster when they try to expand on this use case.

polvorin (approved):

PR #4232, opened 2023-02-09
refactor: move the `OneTimeCode` struct from the ockam_api crate to the ockam_identity crate

etorreborre:
This PR moves the `OneTimeCode` struct to the same crate as the `Credential` struct.

When retrieving credentials we can send a `OneTimeCode` and get back a `Credential` so it makes sense to put them together in the same crate.

Additional improvements:

  - I better encapsulated the encoding / decoding of the `OneTimeCode` as a `String`
  - I removed some duplicated code to parse a `OneTimeCode`
  - I added a command line arguments parsing test for authenticated with a token

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3125,6 +3125,7 @@ dependencies = [
  "ockam_transport_tcp",
  "ockam_vault",
  "quickcheck",
+ "quickcheck_macros",
  "rand 0.8.5",
  "rand_xorshift",
  "serde",
@@ -3696,6 +3697,17 @@ dependencies = [
  "env_logger",
  "log",
  "rand 0.8.5",
+]
+
+[[package]]
+name = "quickcheck_macros"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b22a693222d716a9587786f37ac3f6b4faedb5b80c23914e7303ff5a1d8016e9"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
 ]
 
 [[package]]
--- a/implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
+++ b/implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
@@ -3,13 +3,15 @@ pub mod types;
 use core::{fmt, str};
 use lru::LruCache;
 use minicbor::{Decoder, Encode};
+use ockam::identity::authenticated_storage::AuthenticatedStorage;
+use ockam::identity::credential::{Credential, OneTimeCode, SchemaId};
+use ockam::identity::{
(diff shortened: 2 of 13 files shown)


implementations/rust/ockam/ockam_identity/src/lib.rs
```diff
@@ -18,6 +18,12 @@ extern crate core;
 #[macro_use]
 extern crate alloc;
 
+#[cfg(test)]
+extern crate quickcheck;
+#[cfg(test)]
+#[macro_use(quickcheck)]
```
SanjoDeundiak: Is `#[macro_use]` even a thing now? I thought you don't need it anymore

implementations/rust/ockam/ockam_identity/src/lib.rs
```diff
@@ -18,6 +18,12 @@ extern crate core;
 #[macro_use]
 extern crate alloc;
 
+#[cfg(test)]
+extern crate quickcheck;
```
SanjoDeundiak: Could you please elaborate on why we need this?

SanjoDeundiak (changes requested):
Looks almost perfect

implementations/rust/ockam/ockam_identity/src/lib.rs
etorreborre: I got it from there: https://github.com/BurntSushi/quickcheck#the-quickcheck-attribute. On the other hand you're right, `#[macro_use]` is not necessary anymore, I'll remove it

etorreborre: And maybe to answer the original question I'm using `quickcheck` [here](https://github.com/build-trust/ockam/pull/4232/files#diff-7ee4043c5dc1f294955fee55a11b5e3abea18e271aed8b94e1fbe7c835880a01R82) for a round-trip test.

implementations/rust/ockam/ockam_identity/src/lib.rs
etorreborre: Actually this seems to be necessary: https://github.com/build-trust/ockam/actions/runs/4142577738/jobs/7163412333

implementations/rust/ockam/ockam_identity/src/lib.rs
SanjoDeundiak: I'm fine with introducing this dependency, why should it be declared as an extern crate?

etorreborre: I was just following the tutorial on the quickcheck repository. I guess they recommend `extern crate` to make the macro attribute available everywhere. I removed those 2 statements and did an import locally instead. Thanks for correcting me on what is idiomatic Rust, I have much to learn in that domain :-).

SanjoDeundiak (approved):

PR #4243, opened 2023-02-10
feat(rust): pre-trusted identity identifiers attributes

polvorin:
This implement pre-trusted identity attributes. Define specific traits for attribute storage.  Add more information (when it was added, attested by who),  latter this should be made into proofs.  Adapt the "authenticated" service and command.
This is a first step towards the goal of refactoring the `Authority` to use ABAC rules to identify enrollers  (authority will then need a way to have some trust anchor,   what we call pre-trusted identities here). 

Specifically:
* Added a set of traits (`IdentityAttributeStorageReader`,   `IdentityAttributeStorageWriter` ,  `IdentityAttributeStorage`)   instead of using the raw `AuthenticatedStorage` that is untyped
* Change the places where the code was reading/writting identity' attributes to use the above traits
* Remove it from the `Identity` struct   (altough the underling `AuthenticatedStorage`  is still a template parameter because it's used on identity history operations,  unrelated to this) 
* Added implementation for these traits:
    * one that uses an underling AuthenticatedStorage as storage  (so, basically what we had so far)
    * one that implements the `IdentityAttributeStorageReader`  trait only,  and is instantiated either from a string or a file on disk  
    * one that implements a cascade mechanism, instantiated with two underling stores:  look it up on the first, and if not found look it on the second. 
* Added way to list all identities/attributes
* Add option to start up a node with a set of pre-trusted identities using the cascading behavior described above
* Adapted the existing "authenticated" command

What's not included:
* Type simplifications.    When referring to identity' attributes,  there are differences on how they are treated on different components.  (ej:  they are u8[]  when on  a Credential,    but  must be a valid  String  (utf8)  when used for ABAC or other cases.)
* simplify the UX for the "authenticated" command  (we likely need to change the name of that command,   also likely make it understand the "/node"  multiaddr,  and think how we want to access control it.    I feel this will end up on nodemanager rather than separated)

Sample trusted-identities file  `trusted-identities.json`:
```{
        "P13bde7244c3883863458f50c7dc3c3ec8ce115afce7054f7c1be4ae4bb8d2431": {"role" : "sensor", "location" : "xx"},
        "P13bde7244c3883863458f50c7dc3c3ec8ce115afce7054f7c1be4ae4bb8d2432": {"role" : "other", "location" : "xx"}
}
```

```
$ ockam node create  --trusted-identities-file  trusted-identities.json   [..other create options..]
[...]
$ ockam authenticated get --id P13bde7244c3883863458f50c7dc3c3ec8ce115afce7054f7c1be4ae4bb8d2431 [addr]
Authenticated Identities

▐ Identifier: P13bde7244c3883863458f50c7dc3c3ec8ce115afce7054f7c1be4ae4bb8d2431
▐ Attributes: {"location":"xx","role":"sensor"}
▐ Added At: 1676052962
▐ Expires At: -
▐ Attested By: -

```
This comes from the pre-trusted identities,  so there is no expiration date, nor is attested by anyone.

If we made this node connect receive `credentials` from some other node (`P757af2f95e437ba5b552802915dc43c7a7f500f6e2f46a95b6fe3e3d1f645325` in this case),  we can see it:

```
.$ ockam authenticated list  [addr]
Authenticated Identities

▐ Identifier: P757af2f95e437ba5b552802915dc43c7a7f500f6e2f46a95b6fe3e3d1f645325
▐ Attributes: {"location":"ny","project_id":"f3557590-86ae-4dfc-b6fc-973c036ac23d","service":"sensor"}
▐ Added At: 1676049100
▐ Expires At: 1678640728
▐ Attested By: P13248b4a771875600e00f6c6edde59717c0e5820e2c413d380ed5c1136cb6022

▐ Identifier: P13bde7244c3883863458f50c7dc3c3ec8ce115afce7054f7c1be4ae4bb8d2432
▐ Attributes: {"location":"xx","role":"other"}
▐ Added At: 1676052962
▐ Expires At: -
▐ Attested By: -

▐ Identifier: P13bde7244c3883863458f50c7dc3c3ec8ce115afce7054f7c1be4ae4bb8d2431
▐ Attributes: {"location":"xx","role":"sensor"}
▐ Added At: 1676052962
▐ Expires At: -
▐ Attested By: -
```

Note there is an expiration date for `P757af2f95e437ba5b552802915dc43c7a7f500f6e2f46a95b6fe3e3d1f645325` attributes (as where obtained from a credential),  and that the `Attested By` field is the ID of the `Authority`  that signed the credential.   (in some future stage,  we can include _proof_ of that as well).


polvorin:
@Mergifyio rebase

caoakleyii (approved):

PR #4248, opened 2023-02-13
feat(rust): command shell abstraction

adrianbenavides:
Add a basic implementation of a shell abstraction for the CLI. 

The main goals are centralizing and hiding the details of:
- Detecting interactive/non-interactive
- Detecting color usage (NO_COLOR env, no-color arg)
- Using custom streams for stdout/stderr. This will allow us to use buffered streams that can be checked in tests
- Abstracting how messages are printed to the logs/terminal
- Abstracting (to some degree) the format of the messages 

The shell abstraction has two uses:
- Writing log messages to the stderr stream
- Writing the output of the commands to the stdout stream

The UX is pretty straightforward and automatically applies the restrictions defined by the arguments like `no-color` or `quiet`.

To write logs:

```rust
terminal.write("1")?;
terminal.rewrite("2")?;
terminal.write_line(&"2".red().to_string())?;
```

To ask for user input:

```rust
if terminal.can_ask_for_user_input() {
    let mut b = String::new();
    println!("Enter your name :");
    std::io::stdin().read_line(&mut b)?;
    ...
}
```

To write the command output:

```rust
terminal.stdout()
    .human("This is a human message")
    .machine("This is a machine message")
    .write_line()?;
```

Calling the `stdout` method changes the methods available to prevent the user from misusing the API: see [here](https://github.com/build-trust/ockam/pull/4248/files#diff-d7024cce8bdc7581cbb4f7e138d707c97a299c4deb7172195bf29081e2ea05bdR137) and [here](https://github.com/build-trust/ockam/pull/4248/files#diff-d7024cce8bdc7581cbb4f7e138d707c97a299c4deb7172195bf29081e2ea05bdR174)).


implementations/rust/ockam/ockam_command/src/node/delete.rs
```diff
@@ -34,7 +34,12 @@ fn run_impl(opts: CommandGlobalOpts, cmd: DeleteCommand) -> crate::Result<()> {
         delete_all_nodes(opts, cmd.force)?;
     } else {
         delete_node(&opts, &cmd.node_name, cmd.force)?;
-        println!("Deleted node '{}'", &cmd.node_name);
+        opts.shell
+            .stdout()
+            .human(format!("Deleted node_name '{}'", &cmd.node_name))
```
caoakleyii: Unless i'm missing something here, the output should be more human friendly, `Deleted node '{}'` should be fine

implementations/rust/ockam/ockam_command/src/node/delete.rs
```diff
@@ -34,7 +34,12 @@ fn run_impl(opts: CommandGlobalOpts, cmd: DeleteCommand) -> crate::Result<()> {
...
+            .machine(&cmd.node_name)
+            .json(&serde_json::json!({ "node_name": &cmd.node_name }))
```
caoakleyii: This is a very interesting interface.

One of the things that concerns me is how much this will balloon if we add more supported formats, and then having to go back and update existing outputs as well.

I'll need to take some time to think about this more.

adrianbenavides: Right now, all formats are optional and it defaults to the `human` format. This might not be the best approach and that's something I'm starting to see now that I'm working on the commands' output (e.g. we could tweak the internals to always have a fallback output message, whatever that is). In any case, this design should be forward-compatible right?

implementations/rust/ockam/ockam_command/src/node/delete.rs
adrianbenavides: Correct, that was a copy-paste error.

implementations/rust/ockam/ockam_command/src/node/delete.rs
adrianbenavides: I've added a simple mechanism to select a fallback output message, hidden from the public interface.

implementations/rust/ockam/ockam_command/src/shell/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+        // will be stripped out from output messages. Otherwise, let the terminal decide.
+        no_color || std::env::var("NO_COLOR").is_ok()
+    }
+
+    fn should_disable_user_input(no_input: bool) -> bool {
+        // If global argument `--no-input` is passed or the `NO_INPUT` env var is set we won't be able
+        // to ask the user for input.  Otherwise, let the terminal decide based on the `is_tty` value
+        no_input || std::env::var("NO_INPUT").is_ok()
+    }
+}
```
mrinalwadhwa: We should look at what the conventional name for this argument is across popular command line. `--no-input` is likely not it.

mrinalwadhwa:
Cli, Term, Terminal, Shell all mean the same things. Let's pick one word and use it in all cases. Also, merge with this module https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/terminal.rs

implementations/rust/ockam/ockam_command/src/shell/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+impl<W: ShellWriter> Shell<W, Finished> {
+    pub fn human<T: Display>(mut self, msg: T) -> Self {
+        self.mode.output.human = Some(msg.to_string());
+        self
+    }
+
+    pub fn machine<T: Display>(mut self, msg: T) -> Self {
+        self.mode.output.machine = Some(msg.to_string());
+        self
+    }
```
mrinalwadhwa: I think we should rename
output.human to output.normal
output.machine to output.tty

the human / machine idea is very leaky. For example json is also machine readable.

implementations/rust/ockam/ockam_command/src/shell/mod.rs
mrinalwadhwa: `--no-input-prompts` and `NO_INPUT_PROMPTS`

adrianbenavides: I got that from: https://clig.dev/#interactivity

Maybe it's not widely used in general?

implementations/rust/ockam/ockam_command/src/shell/mod.rs
adrianbenavides: Actually, `human` is what is shown when the terminal is interactive and `machine` when is not.
- stdout is tty -> show human-readable format
- stdout is not tty -> show machine-readable format

Context on the terms used: https://clig.dev/#output

I agree though that `machine` specifically doesn't make 100% sense, but I think it's the most accurate. By default, the format can be either `human` or `machine` and is derived from the stdout tty value. The user can override the format with the `--format` argument.

Maybe we could use `plain` instead of `human`? For the `machine` format I don't have a clear answer as any other format falls into the "non-plain" category.

> For example json is also machine readable.

Yes, but json is a known format in itself. We can clarify the kind of output the user can expect from our `machine` format.

adrianbenavides:
> Cli, Term, Terminal, Shell all mean the same things. Let's pick one word and use it in all cases. Also, merge with this module https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/terminal.rs

After re-reading [this](https://unix.stackexchange.com/questions/4126/what-is-the-exact-difference-between-a-terminal-a-shell-a-tty-and-a-con/4132#4132) I'll go for `Terminal`.

implementations/rust/ockam/ockam_command/src/shell/mod.rs
caoakleyii: I'm unable to find many concrete examples of this, outside of the link @adrianbenavides has already provided.

Most `no-[resource]` flags i've seen though have been brief, and i'd imagine given the nature of most cli flags don't lean into verbosity. 

`--no-input` would be what I'd go with.

implementations/rust/ockam/ockam_command/src/shell/mod.rs
caoakleyii: Both very great points here.

I personally would see us go with plain, as opposed to normal or human.

Because of exactly what JSON/YAML, etc are.

Especially in this web based world where plain-text is often the go to vernacular for anything that isn't a standardized format that would work here.

adrianbenavides: Renamed `human` to `plain`.

adrianbenavides:
@CAOakleyII this should be ready to merge now.

caoakleyii (dismissed):

caoakleyii (approved):

PR #4254, opened 2023-02-14
feat(rust): parse `/node/n1` to `/worker/addr` after connecting to the node via tcp

adrianbenavides:


implementations/rust/ockam/ockam_api/src/cloud/invitation.rs
```diff
@@ -92,7 +92,9 @@ mod node {
             dec: &mut Decoder<'_>,
         ) -> Result<Vec<u8>> {
             let req_wrapper: CloudRequestWrapper<CreateInvitation> = dec.decode()?;
-            let cloud_route = req_wrapper.route()?;
+            let cloud_route = req_wrapper
```
etorreborre, line 95: I would propose to invert the calls if possible. Have a function on `NodeManager` taking a "route" from the request and accessing its own `tcp_transport` to do the conversion to a `cloud_route`. This avoids a bunch of repetition and you can write directly:
```
let cloud_route = self.resolve_route(req_wrapper.route()).await?;
```

adrianbenavides: That would be great, yes. We have a bunch of `multiaddr_to_X` functions that we should reconsider... it's very confusing and intricate how we modify the route as it travels from the command's argument down to the final request.

I'd delay this after sanjo finishes his work on `sanjo/disable_auto_connect` to avoid more merge conflicts. I'll create an issue for this.

adrianbenavides:
This PR's commit was cherry-picked into `sanjo/disable_auto_connect`

PR #4258, opened 2023-02-14
chore: upgrade cortex-m-rt

etorreborre:
in order to fix this advisory issue: https://github.com/build-trust/ockam-artifacts/issues/110.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1076,9 +1076,9 @@ dependencies = [
 
 [[package]]
 name = "cortex-m-rt"
-version = "0.7.2"
+version = "0.7.3"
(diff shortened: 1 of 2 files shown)


davide-baldo:
I think it's OK to update the library, but we are *not* affected since the version is <= 0.7.0

etorreborre:
Actually it is in our dependency tree as `0.7.2` via other dependencies

davide-baldo:
~~Please update the commit message to reflect that~~

davide-baldo (approved):

etorreborre:
Thanks @davide-baldo!

PR #4274, opened 2023-02-15
feat(rust): standarize where authority stores membership information

polvorin:
makes the authority store membership (authenticated identity attributes) information using the standard method for that.  This opens the way to use ABAC to identity enrollers latter on.
Also:
- tweak  enroller'  `project enroll` command to be able to actually work with a project.json file
- add an "hex"  export format to  `identity show`   (to export the full change history as hex string.   This is used to populate a project.json to be able to contact an standalone authority node)
- allow authority node to be run on background  (so far it only worked on foreground)
- add bats test [with a standalone authority](https://github.com/build-trust/ockam/pull/4274/files#diff-76069ee844324858f89f101eb1c6788f22573c9d945ccb48fede3136c3c427d7)

After this change,   since the storage is the same, the pre-trusted identities support is available to authority nodes,  as well as the commands to inspect the authority' authenticated identities table  (both where added for  "normal" nodes  [here](https://github.com/build-trust/ockam/pull/4243)):
```
$ockam authenticated list [ addr ] 
▐ Identifier: P2b7344b142ad96231df68f24e9b45b1ceadfc11674a140a21ce09873d23dd9db
▐ Attributes: {"location":"ny","type":"sensor"}
▐ Added At: 1676553756
▐ Expires At: -
▐ Attested By: Pd311fb0cf1aaf2b0cd83d9e3748f83d263e4ee7c46e85cdc2767d3b9bcfc573b

▐ Identifier: P98bcfa2969ba8da0654e9dc80a33341eaf456dab43deb6dc82bfaaa544bb34a8
▐ Attributes: {"location":"sf","type":"sensor"}
▐ Added At: 1676553753
▐ Expires At: -
▐ Attested By: Pd311fb0cf1aaf2b0cd83d9e3748f83d263e4ee7c46e85cdc2767d3b9bcfc573b

▐ Identifier: Pb0dbffd908eea053ee184e0cece3c69ee709e7fa1dbc46161cc665987ff6d9c0
▐ Attributes: {"sample_attr":"sample_val"}
▐ Added At: 1676553741
▐ Expires At: -
▐ Attested By: -
``` 

In the above example there where 3 members registered on the authority:   - 
- `Pb0dbffd908eea053ee184e0cece3c69ee709e7fa1dbc46161cc665987ff6d9c0`   was passed as a pre-trusted identifier at authority' creation,  so it isn't attested by anyone. 
- `P98bcfa2969ba8da0654e9dc80a33341eaf456dab43deb6dc82bfaaa544bb34a8` and `P2b7344b142ad96231df68f24e9b45b1ceadfc11674a140a21ce09873d23dd9db`  were added by enroller `Pd311fb0cf1aaf2b0cd83d9e3748f83d263e4ee7c46e85cdc2767d3b9bcfc573b`.

Both have no expiration date,  we don't expire _members_  from authority right now.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
+++ b/implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
@@ -3,8 +3,10 @@ pub mod types;
 use core::{fmt, str};
 use lru::LruCache;
 use minicbor::{Decoder, Encode};
-use ockam::identity::authenticated_storage::AuthenticatedStorage;
-use ockam::identity::credential::{Credential, OneTimeCode, SchemaId};
+use ockam::identity::authenticated_storage::{
+    AttributesEntry, AuthenticatedStorage, IdentityAttributeStorage,
+};
+use ockam::identity::credential::{Credential, OneTimeCode, SchemaId, Timestamp};
 use ockam::identity::{
     Identity, IdentityIdentifier, IdentitySecureChannelLocalInfo, IdentityVault,
 };
@@ -25,7 +27,7 @@ use crate::authenticator::direct::types::CreateToken;
 
 use self::types::Enroller;
 
-const MEMBER: &str = "member";
+const LEGACY_MEMBER: &str = "member";
 const MAX_TOKEN_DURATION: Duration = Duration::from_secs(600);
 
 /// Schema identifier for a project membership credential.
@@ -36,12 +38,11 @@ const MAX_TOKEN_DURATION: Duration = Duration::from_secs(600);
 /// - `role`: b"member"
 pub const PROJECT_MEMBER_SCHEMA: SchemaId = SchemaId(1);
 pub const PROJECT_ID: &str = "project_id";
-pub const ROLE: &str = "role";
 
-pub struct Server<S, IS: AuthenticatedStorage, V: IdentityVault> {
+pub struct Server<S: AuthenticatedStorage, IS: IdentityAttributeStorage, V: IdentityVault> {
     project: Vec<u8>,
-    store: S,
-    ident: Identity<V, IS>,
+    store: IS,
+    ident: Identity<V, S>,
     filename: Option<String>,
     enrollers: HashMap<IdentityIdentifier, Enroller>,
     reload_enrollers: bool,
@@ -50,6 +51,7 @@ pub struct Server<S, IS: AuthenticatedStorage, V: IdentityVault> {
 
 struct Token {
     attrs: HashMap<String, String>,
+    generated_by: IdentityIdentifier,
     time: Instant,
 }
 
@@ -57,7 +59,7 @@ struct Token {
 impl<S, IS, V> Worker for Server<S, IS, V>
 where
     S: AuthenticatedStorage,
-    IS: AuthenticatedStorage,
+    IS: IdentityAttributeStorage,
     V: IdentityVault,
 {
     type Context = Context;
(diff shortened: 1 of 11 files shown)


polvorin:
@Mergifyio rebase

implementations/rust/ockam/ockam_command/src/identity/show.rs
```diff
@@ -32,7 +35,10 @@ fn run_impl(opts: CommandGlobalOpts, cmd: ShowCommand) -> crate::Result<()> {
         );
     }
     let state = opts.state.identities.get(&cmd.name)?;
-    if cmd.full {
+    if cmd.hex {
+        let identity = state.config.change_history.export()?;
+        print_output(identity, &opts.global_args.output_format)?;
+    } else if cmd.full {
```
caoakleyii: This is something I've been thinking about trying to support globally as well.

Perhaps we have a flag similar to `--output json` but instead `--encoding hex`  (which would allow for base64 use cases in the future etc)

caoakleyii: Does not needed to be fully added in this PR, maybe just change this command and examples to accept encoding.

Write a `// TODO `

And I can update it to be a global option in my coming PR.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -171,10 +201,20 @@ where
...
-                            self.store
-                                .set(from.key_id(), MEMBER.to_string(), attributes)
-                                .await?;
+                            //TODO: fixme:  unify use of hashmap vs btreemap
+                            let attrs = tkn
+                                .attrs
+                                .iter()
+                                .map(|(k, v)| (k.to_string(), v.as_bytes().to_vec()))
+                                .collect();
+                            let entry = AttributesEntry::new(
```
etorreborre, line 210: I think this deserves a constructor for `AttributesEntry` taking a map and an `IdentityIdentifier` to remove the duplication.

implementations/rust/ockam/ockam_command/src/identity/show.rs
etorreborre: @CAOakleyII why not `--output hex`?

implementations/rust/ockam/ockam_command/tests/commands.bats
```diff
@@ -409,6 +409,53 @@ teardown() {
...
+  assert_success
+
+  echo "{\"id\": \"1\",
+  \"name\" : \"default\",
+  \"identity\" : \"P6c20e814b56579306f55c64e8747e6c1b4a53d9a3f4ca83c252cc2fbfc72fa94\",
+  \"access_route\" : \"/dnsaddr/127.0.0.1/tcp/4000/service/api\",
+  \"authority_access_route\" : \"/dnsaddr/127.0.0.1/tcp/4200/service/api\",
+  \"authority_identity\" : \"$authority_identity_full\"}" > /tmp/project.json
+
+  # m1 is a member (it's on the set of pre-trusted identifier) so it can get it's own credential
```
etorreborre: s/pre-trusted identifier/pre-trusted identifiers and s/it's/its

implementations/rust/ockam/ockam_command/src/identity/show.rs
polvorin: replaced `--hex`  with `--encoding hex`   (only applicable if `--full` was specified) and added a TODO to review if that would be something general on other commands   as well.  The only other that I think right now could  make sense is for credential show.   But even there perhaps settling on some json format is more user friendly).

etorreborre (dismissed):
Better types, that's nice!

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
polvorin: agree,  will let it as TODO to address latter.    _Ideally_  we should settle on use either  btreemap or hashmap  for this and avoid the conversions.   And it's mostly about ergonomics,   as the cardinality of this map is not going to be large.    (I _think_  why btree map was introduced on some parts was because it serialized out-of-the-box through cbor,  but I might be wrong on that)

implementations/rust/ockam/ockam_command/src/identity/show.rs
caoakleyii: > @CAOakleyII why not --output hex?

@etorreborre 
I thought about that at first.

Few reasons, some very opinionated and could be wrong - 

* I view output primarily as formatting, human readable but machine serializable outputs. (JSON, YAML, TOML). 

* They are technically mutually exclusive, you could technically have a base64/hex encoded json output.

* _Slightly_ more refactoring required, our current `--output` defines that the struct has traits of `Output` and `serde::Serialize` where as a hex encoding does not need that.

caoakleyii: >The only other that I think right now could make sense is for credential show. But even there perhaps settling on some json format is more user friendly).
@polvorin 

Yes, allowing credentials to go into a JSON format is very use friendly from a readable stand point.

However, in the case of commands where a user wants to copy/paste a credential, this is more user friendly.

etorreborre (approved):

PR #4277, opened 2023-02-16
kafka message encryption

davide-baldo:
## Proposed Changes

This PR adds support for message encryption for kafka, for each topic/partition a secure channel is created and is used to exchange keys and encrypt messages.

Every kafka `produce` and `fetch` operations are intercepted and the content is transparently encrypted and wrapped around a structure declaring the secure channel identifier used to encrypt it.

The integration with the orchestrator was also added, now the consumer adds a forwarder inside the orchestrator redirecting `consumer__{topic_name}_{partition}` to its own secure channel listener.

The current implementation supports multiple producers and multiple consumers(only 1 per partition).

## How to use/test

1. First enroll in the orchestrator and the project
2. Create the credential for the project
```sh
$ ockam credential get
``` 
3. Start the confluent addon
```sh
$ ockam project addon configure confluent --bootstrap-server my-bootstrap-server:9092
```
4. Create a dedicated node for consumer and producer
```sh
$ ockam node create consumer
$ ockam node create producer
```
5. Start service consumer and producer
```sh
$ ockam service start kafka-consumer --node consumer --ip 127.0.0.1 --forwarding-addr /project/default --bootstrap-port 5555 --port-range 5556-5599
$ ockam service start kafka-producer --node producer --ip 127.0.0.1 --forwarding-addr /project/default --bootstrap-port 4444 --port-range 4445-4499
```
6. Create the configuration file the kafka client in `~/.kafka_sasl_config`, adding with your credentials
```text
request.timeout.ms=3000
security.protocol=SASL_PLAINTEXT
sasl.mechanism=PLAIN
sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required \
        username="USERNAME" \
        password="PASSWORD";
```
7. start the consumer client
```sh
./kafka-console-consumer.sh --topic topic-name --bootstrap-server localhost:5555 --consumer.config ~/.kafka_sasl_config
```
8. start the producer client
```sh
./kafka-console-producer.sh --topic topic-name --bootstrap-server localhost:4444 --producer.config ~/.kafka_sasl_config
```

## Out of scope for now

- multiple consumers per partition
- credential exchange between producer and consumer

## Checks


implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
```diff
@@ -199,38 +207,49 @@ impl KafkaPortalWorker {
             .await
             .map_err(InterceptError::Io)?;
 
+        let mut written_kafka_message = false;
         loop {
```
etorreborre: A `while` expression would be slightly bette here IMO

implementations/rust/ockam/ockam_api/src/util.rs
```diff
@@ -304,7 +316,10 @@ pub mod test {
         )
         .await?;
 
-        let node_manager_worker = NodeManagerWorker::new(node_manager);
+        // node_manager.identity()?.vault().
```
etorreborre: let's remove this comment

implementations/rust/ockam/ockam_api/src/util.rs
```diff
@@ -239,26 +239,38 @@ pub mod test {
...
     use crate::nodes::{NodeManager, NodeManagerWorker, NODEMANAGER_ADDR};
+    use ockam::compat::tokio::sync::RwLock;
     use ockam::Result;
     use ockam_core::AsyncTryClone;
     use ockam_identity::Identity;
     use ockam_node::Context;
+    use ockam_transport_tcp::TcpTransport;
+    use std::sync::Arc;
 
     ///guard to delete the cli state at the end of the test
```
etorreborre: We could modify this comment as well to explain that we need to be able to access the `NodeManager` in the tests and also to eventually delete the `CliState` (and maybe explain what that means, a clean-up of some local files I guess?)

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
```diff
@@ -288,38 +293,96 @@ impl NodeManager {
     pub(super) async fn start_kafka_service_impl<'a>(
         &mut self,
         context: &Context,
-        listener_address: Address,
+        local_interceptor_address: Address,
```
etorreborre, line 297: It would be good to have comments for all these parameters and describe what they represent exactly. Maybe by referring to a diagram somewhere showing the overall topology.

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
```diff
@@ -288,38 +293,96 @@ impl NodeManager {
...
+                    Some(identity.identifier().clone()),
+                    Some(Duration::from_secs(60)),
+                    context,
+                )
+                .await?;
+
+            let project_multiaddr = maybe_tunnel_multiaddr.try_with(&suffix_address)?;
+            let project_route = try_multiaddr_to_route(&project_multiaddr)?;
+
+            let bootstrap_address_route: Route = RouteBuilder::default()
```
etorreborre: Maybe we can use the `route` macro to be a bit more concise here:
```
let bootstrap_address_route = route![
  local_interceptor_address.clone(),
  project_route.clone(),
  ORCHESTRATOR_KAFKA_INTERCEPTOR_ADDRESS,
  ORCHESTRATOR_KAFKA_BOOTSTRAP_ADDRESS];
```
I hope I got the order right :-). But that's also the point of using `route` rather than having to think about prepend vs append.

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
```diff
@@ -288,38 +293,96 @@ impl NodeManager {
...
+                .prepend(local_interceptor_address.clone())
+                .append(ORCHESTRATOR_KAFKA_INTERCEPTOR_ADDRESS)
+                .into();
+
+            debug!("bootstrap_address_route: {bootstrap_address_route:?}");
+            debug!("interceptor_route: {interceptor_route:?}");
+            debug!("project_route: {project_route:?}");
+            (bootstrap_address_route, interceptor_route, project_route)
+        } else {
+            //when executing test the provided address change semantic a little bit
```
etorreborre: It would be good to say which test we are talking about here.

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+use ockam_identity::api::{
+    DecryptionRequest, DecryptionResponse, EncryptionRequest, EncryptionResponse,
+};
+use ockam_identity::authenticated_storage::AuthenticatedStorage;
+use ockam_identity::{Identity, IdentityVault, SecureChannelRegistryEntry, TrustEveryonePolicy};
+use ockam_node::compat::tokio::sync::Mutex;
+use ockam_node::Context;
+use serde::{Deserialize, Serialize};
+
+///Offer simple APIs to encrypt and decrypt kafka messages.
```
etorreborre: This is a small thing but I notice that some comments have no space after `///` while others have a space. It would be good to leave a space for consistency (and that's the overall rust practice AFAIU)

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+/// Underneath it create secure channels for each topic/partition
```
etorreborre: s/create/creates

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+/// and uses them to encrypt the content.
+/// Multiple secure channels may be created for the same topic/partition
+/// but each will be explicitly label.
```
etorreborre: s/label/labelled

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+/// It's the same for both producer and consumer although it could be split
+/// into two distinct implementation.
```
etorreborre: s/implementation/implementations

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+///This is proxy trait to avoid propagating the vault implementation.
+#[async_trait]
+pub(crate) trait KafkaSecureChannelController: Send + Sync {
+    ///encrypt the content specifically for the consumer waiting for that topic name and partition
+    /// the the secure channel is missing it'll be created, this may take up to few seconds
```
etorreborre: s/the the/if the

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+    ) -> Result<(UniqueSecureChannelId, Vec<u8>), Error>;
+
+    ///decrypt the content based on the unique secure channel identifier
+    /// the secure channel is expected to be already initialized
+    async fn decrypt_content_for(
+        &self,
+        context: &mut Context,
+        secure_channel_id: UniqueSecureChannelId,
+        encrypted_content: Vec<u8>,
+    ) -> Result<Vec<u8>, Error>;
```
etorreborre: I think you should be able to reuse the `ockam_core::Result` here and just write `Result<Vec<u8>>`

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+/// Underneath it create secure channels for each topic/partition
+/// and uses them to encrypt the content.
+/// Multiple secure channels may be created for the same topic/partition
+/// but each will be explicitly label.
+/// It's the same for both producer and consumer although it could be split
+/// into two distinct implementation.
+///This is proxy trait to avoid propagating the vault implementation.
+#[async_trait]
+pub(crate) trait KafkaSecureChannelController: Send + Sync {
+    ///encrypt the content specifically for the consumer waiting for that topic name and partition
```
etorreborre: It could be a good idea to explain that we setup a secure channel just to agree on a key to encrypt/decrypt data but not to route it as we would do with a regular secure channel. I wonder if that should deserve a different name, like "EncryptedPortal" or something (but naming is hard as well all know :-)). 

Then one thing we can do later on is to refactor the secure channel code so that the initialization phase is performed by some workers to go until the identity exchange and then some simple decryption/encryption workers are started to process messages. Those 2 workers would then not be necessary in a Kafka scenario.

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+///Unique identifier for a specific secure_channel.
+/// Used in order to distinguish between secure channels created between
+/// the same identities.
+#[derive(Debug, Clone, Serialize, Deserialize, Message)]
+struct SecureChannelIdentifierMessage {
+    secure_channel_identifier: UniqueSecureChannelId,
+}
+
+pub(crate) type UniqueSecureChannelId = u64;
+type TopicPartition = (String, i32);
```
etorreborre, line 94: I would create a struct here with `partition_name: String, partition_id: i32`

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+    ///encrypt the content specifically for the consumer waiting for that topic name and partition
+    /// the the secure channel is missing it'll be created, this may take up to few seconds
+    /// the first time
+    async fn encrypt_content_for(
+        &self,
+        context: &mut Context,
+        topic_name: &str,
+        partition_id: i32,
+        content: Vec<u8>,
+    ) -> Result<(UniqueSecureChannelId, Vec<u8>), Error>;
```
etorreborre: Maybe it's worth introducing a struct
```
struct EncryptedContent {
  channel_id: UniqueSecureChannelId,
  content: Vec<u8>
}
```
as a return value of `encrypt_content_for` and an input value of `decrypt_content_for` to indicate that both are correlated and having less chances to mix them up.

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+/// Used in order to distinguish between secure channels created between
+/// the same identities.
+#[derive(Debug, Clone, Serialize, Deserialize, Message)]
+struct SecureChannelIdentifierMessage {
+    secure_channel_identifier: UniqueSecureChannelId,
+}
+
+pub(crate) type UniqueSecureChannelId = u64;
+type TopicPartition = (String, i32);
+struct InnerSecureChannelControllerImpl<V: IdentityVault, S: AuthenticatedStorage> {
```
etorreborre, line 95: I would be good to put this struct closer to `KafkaSecureChannelControllerImpl` so that we see that it's almost the same thing minus the `Arc<Mutex<...>>`

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+    //we are using encryptor api address as unique _local_ identifier
+    //of the secure channel
+    id_encryptor_map: HashMap<UniqueSecureChannelId, Address>,
```
etorreborre, line 98: Maybe a [`bimap`](https://docs.rs/bimap/latest/bimap) is even more appropriate there then?

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+
+        for partition in partitions {
+            let topic_key: TopicPartition = (topic_name.to_string(), partition);
+            if inner.topic_forwarder_set.contains(&topic_key) {
+                continue;
+            }
+
+            let alias = format!("{topic_name}_{partition}");
+
+            if inner.is_using_project {
```
etorreborre, line 372: I am not a big fan of having test code mixed with production code. At some stage it would be nice to refactor this so that we can plug some different logic for creating forwarders for tests

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+    async fn add_mapping(&self, id: UniqueSecureChannelId, encryptor_address: Address) {
+        self.inner
+            .lock()
+            .await
+            .id_encryptor_map
+            .insert(id, encryptor_address);
+    }
+}
+
+struct SecureChannelControllerListener<V: IdentityVault, S: AuthenticatedStorage> {
```
etorreborre, line 153: Why do we need this additional struct and cannot reuse `KafkaSecureChannelControllerImpl` directly for the `Worker` implementation?

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+    async fn get_secure_channel_for(
+        &self,
+        secure_channel_id: UniqueSecureChannelId,
+    ) -> Result<SecureChannelRegistryEntry, Error> {
+        let inner = self.inner.lock().await;
+        if let Some(encryptor_address) = inner.id_encryptor_map.get(&secure_channel_id) {
+            inner
+                .identity
+                .secure_channel_registry()
+                .get_channel_list()
```
etorreborre, line 262: Maybe we could add a function to the `SecureChannelRegistry` to do that kind of search. There is already `get_channel_by_encryptor_address`, we can add `get_responder_channel_by_encryptor_address`

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+        topic_name: &str,
+        partition_id: i32,
+        content: Vec<u8>,
+    ) -> Result<(UniqueSecureChannelId, Vec<u8>), Error> {
+        let (unique_id, secure_channel_entry) = self
+            .get_or_create_secure_channel_for(context, topic_name, partition_id)
+            .await?;
+
+        trace!("encrypting content with {unique_id}");
+        let encryption_response: EncryptionResponse = context
```
etorreborre, line 302: At some stage I would like to be able to encrypt / decrypt messages locally without having to send messages to a worker because this is not ultimately necessary.

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -0,0 +1,387 @@
...
+///Offer simple APIs to encrypt and decrypt kafka messages.
+/// Underneath it create secure channels for each topic/partition
+/// and uses them to encrypt the content.
+/// Multiple secure channels may be created for the same topic/partition
+/// but each will be explicitly label.
+/// It's the same for both producer and consumer although it could be split
+/// into two distinct implementation.
+///This is proxy trait to avoid propagating the vault implementation.
+#[async_trait]
+pub(crate) trait KafkaSecureChannelController: Send + Sync {
```
etorreborre, line 37: Food for thought: how much of this is really Kafka-specific? It seems to me that the main idea is to allow the creation and tracking of secure channels and forwarders for a specific set of keys (in the HashMap sense). It turns out that those keys are made of topic names and partition ids for the Kafka use case but this could possibly be applicable to other use cases.

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
```diff
@@ -239,17 +258,19 @@ impl KafkaPortalWorker {
     /// used for requests.
     pub(crate) async fn start(
         context: &mut Context,
+        secure_channel_controller: Arc<dyn KafkaSecureChannelController>,
+        uuid_to_name: TopicUuidMap,
         inlet_map: KafkaInletMap,
     ) -> ockam_core::Result<Address> {
-        let shared_protocol_state = ProtocolState::new();
+        let shared_protocol_state = ProtocolState::new(secure_channel_controller, uuid_to_name);
```
etorreborre, line 251: Would it make sense to put the `inlet_map` inside the `ProtocolState` in order to avoid having to pass it to `intercept_response`?

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
```diff
@@ -239,17 +258,19 @@ impl KafkaPortalWorker {
...
 
         let inlet_address = Address::random_tagged("KafkaPortalWorker.inlet");
         let outlet_address = Address::random_tagged("KafkaPortalWorker.outlet");
         let disconnect_received = Arc::new(AtomicBool::new(false));
 
         let inlet_worker = Self {
             inlet_map: inlet_map.clone(),
-            shared_protocol_state: shared_protocol_state.clone(),
+            shared_protocol_state: shared_protocol_state.async_try_clone().await?,
```
etorreborre, line 259: Would it be possible to clone the `inlet_worker` and just set the `receiving` field to `Receiving`? to limit the duplication?

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
```diff
@@ -239,17 +258,19 @@ impl KafkaPortalWorker {
     /// used for requests.
     pub(crate) async fn start(
```
etorreborre: Having a function like `KafkaPortalWorker::start` is a bit misleading since it starts 2 workers. I'd rather have a simple top level function called `start_kafka_portal` instead.

implementations/rust/ockam/ockam_api/src/kafka/protocol_aware.rs
```diff
@@ -59,16 +95,120 @@ impl ProtocolState {
...
                 buffer.len(),
+                header.correlation_id,
                 header.request_api_version,
                 api_key
             );
 
-            #[allow(clippy::single_match)]
             match api_key {
-                ApiKey::MetadataKey => {
+                ApiKey::ProduceKey => {
```
etorreborre, line 106: This is a large function. It could be a good idea to just see the top-level `match api_key` with the different cases and then delegate to specific functions (same thing for the response side).

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
```diff
@@ -3,16 +3,17 @@ use core::sync::atomic::Ordering;
...
 };
 use ockam_node::Context;
 use ockam_transport_tcp::{PortalMessage, MAX_PAYLOAD_SIZE};
 
 use crate::kafka::decoder::KafkaDecoder;
 use crate::kafka::encoder::KafkaEncoder;
 use crate::kafka::inlet_map::KafkaInletMap;
-use crate::kafka::protocol_aware::ProtocolState;
+use crate::kafka::protocol_aware::{ProtocolState, TopicUuidMap};
+use crate::kafka::secure_channel_map::KafkaSecureChannelController;
```
etorreborre, line 16: Just a note on the diagram below. I think it would be even better it we could trace 2 boxes showing where each `KafkaWorker` in the portal is actually placed, with their 'receiving' role labelled. 
Right now I find the comments a bit confusing since they mention an "instance" of the "KafkaPortalWorker" which is actually "two different workers".

etorreborre (dismissed):
@davide-baldo it was good for me to be able to spend some time reviewing your code. I now understand a lot better what it is doing. 

I left mostly comments that are cosmetic which you can fix at your discretion.

It is nice that you wrote integration tests, and in particular, exercising some boundary conditions. I wonder if it would be possible to also add a few unit tests to specify some of the handling we do with routes and addresses. I am thinking about the case where we could change that code, make a small mistake, and spend a long time to understand why a message never arrives.

davide-baldo:
> It is nice that you wrote integration tests, and in particular, exercising some boundary conditions. I wonder if it would be possible to also add a few unit tests to specify some of the handling we do with routes and addresses. I am thinking about the case where we could change that code, make a small mistake, and spend a long time to understand why a message never arrives.

I also wish to work more with unit tests, so far it wasn't easy because you can't mock side effects outside of the worker without turning everything into a dynamic trait. In the current situation, it's by far easier to do an integration test rather than a unit test. Definitely, I'll make incremental improvements in this direction, but I think I'll also try to play around with some mock frameworks and make some compromises in run-time code as well.

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
davide-baldo: For consistency probably i should beef up the command line help rather than comment on the message

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
davide-baldo: unfortunately, it doesn't work with routes but only with addresses or strings, it certainly would be a nicer syntax improvement.

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
davide-baldo: 100% agree

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
davide-baldo: diagrams and overall architecture will be explained elsewhere

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
davide-baldo: I actually like the simplicity of the tuple here, and being different types there is no risk of getting them wrong.
Since the type is internal I don't think is worth making it a struct.

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
davide-baldo: yeah, make sense

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
davide-baldo: every map has different key,value

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
davide-baldo: I'm not a fan of having conditional code to support tests either, I'll refactor it in the next iteration

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
davide-baldo: Just separation, I think it's a bit cleaner, and we might end up with multiple workers in the SecureChannelController so it's best to keep them separated

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
davide-baldo: It's a bit early to optimize it, let's see how the API gets used

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
davide-baldo: I think that's the idea in the long run

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
davide-baldo: The ProtocolState is shared across workers, while the inlet_map is global across every kafka service (granted normally you should have 1).
Then the TopicUuidMap came in within the ProtocolState breaking a bit that separation, yep maybe a review of states is due.

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
davide-baldo: They have different parameters, mainly `receiving: Receiving::Requests,` and `receiving: Receiving::Responses` so no

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
davide-baldo: make sense

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
etorreborre: I thought I tried it and that it worked. I will do a PR to improve this.

implementations/rust/ockam/ockam_api/src/kafka/mod.rs
```diff
@@ -7,13 +7,593 @@ mod inlet_map;
...
+pub const ORCHESTRATOR_KAFKA_BOOTSTRAP_ADDRESS: &str = "kafka_bootstrap";
+
+pub const KAFKA_SECURE_CHANNEL_CONTROLLER_ADDRESS: &str = "kafka_secure_channel_controller";
+pub const KAFKA_SECURE_CHANNEL_LISTENER_ADDRESS: &str = "kafka_consumer_secure_channel";
 
 pub fn kafka_outlet_address(broker_id: i32) -> Address {
     Address::from_string(format!("kafka_outlet_{}", broker_id))
 }
+
+#[cfg(test)]
```
SanjoDeundiak: Could you please move it to the `tests` dir or at least to a different file? (Could be done in a follow-up PR)

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
SanjoDeundiak: We have an (unwritten) convention of calling a constructor `new` and something that creates and does some additional setup/starts workers is `create`

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
SanjoDeundiak: @etorreborre that would be great, unfortunately, that's non-trivial to implement, because we can't send unserializable structs over the messaging system

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
SanjoDeundiak: @etorreborre I like that idea

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
SanjoDeundiak: My two cents: not a huge fan of maps - in terms of memory&time it's often worse (and causes memory segmentation) when there are not many elements. But sometimes it makes code simpler, then it's fine

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
SanjoDeundiak: That would be a good change in the future

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
SanjoDeundiak: Some find&replace would be great, in a separate PR

SanjoDeundiak (dismissed):
Nice job!

davide-baldo:
I've simplified a little, by rewriting the message length encoder/decoder and removing the tokio utils dependency.
Fixed a bug with messages bigger than 1024.

davide-baldo:
few renames:

- `--ip` => `--bootstrap-server-ip`
- `--bootstrap-port` to `--bootstrap-server-port`
- `--port-range` => `--brokers-port-range`
- `--forwarding-addr` => `--project-route`

implementations/rust/ockam/ockam_api/src/kafka/mod.rs
davide-baldo: done! For now, I've just moved it to a dedicated file, I'll try to figure out how to move it under tests.

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
etorreborre: > every map has different key,value

@davide-baldo just to make sure I understood your answer. My comment was saying "if there is only one UniqueSecureChannelId for each Address in the id_encryptor_map, then we could enforce this invariant with a bimap instead of a hashmap". Are you saying that this is not the case?

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
etorreborre: @SanjoDeundiak I would like to understand this constraint better because I don't see why the code for encrypting messages can not be in principle called directly from the `encrypt_content_for` function. Let's say that you kept a pointer to the encoder of a secure channel in the secure channel registry and called the encoder directly without messaging it. Would that work?

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
etorreborre: @davide-baldo now you can use the new `route!` macro to simplify the code :-)

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
davide-baldo: We could enforce it by aligning keys and values, but we would need to send extra information just for the sake of it.
The 2 maps are actually completely independent (because I used only 1 implementation for both consumer and producer to keep it simple, maybe in the future I'll split them)

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
davide-baldo: :+1:  done!

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
etorreborre: I still don't understand your answer 😬, so I will try one last time, this is not a major thing after all: is there a bijection between `UniqueSecureChannelId` and `Address` in the `id_encryptor_map`?

etorreborre (approved):

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
davide-baldo: Yes, there is a bijection between `UniqueSecureChannelId` and `Address`.

For each secure channel id that is created, there is a relative encryptor.

davide-baldo: ps: I'll try to split the file into consumer and producer structures, after the refactor it'll be clearer.

etorreborre: > For each secure channel id that is created, there is a relative encryptor.

Ok, thanks. This is why I was suggesting to use a Bimap in that case instead of a HashMap :-). Not a big deal but I found it useful in the past to make this kind of relationship more precise.

davide-baldo: Won't it index both key and value? I only need to retrieve data one way

PR #4280, opened 2023-02-16
doc: simple example of credentials exchange

etorreborre:
This PR introduces:

  - a refactoring of the `AbacAccessControl` struct to the `ockam` crate in order to allow users to create simple access control policies based on authenticated attributes and an expression over those attributes. This simplifies example 11. and example 6. (introduced by this PR)

  - an simple version of an `Authority` to be used in prototypes and tests. Any `Identity` can act as a `Credential` issuer so this `Authority` gives the possibility to make credentials certified by its identity. Additionally there is a `Worker` implementation and an `AuthorityClient` allowing users to start the authority as a worker and still access it with regular rust functions in other nodes

  - a simple example starting an Authority worker, and 2 other nodes, Alice and Bob. Both Alice and Bob retrieve some credentials from the Authority then proceed to a credential exchange over a secure channel. We then show that it is possible to restrict the access to one of Bob's services only to Alice, thanks to the credential exchange


implementations/rust/ockam/ockam_identity/src/authority.rs
```diff
@@ -0,0 +1,234 @@
...
+    /// Create an authenticated credential for a list of attribute name/value pairs
+    /// and a given subject
+    async fn get_credential(
+        &self,
+        subject: &IdentityIdentifier,
+        attributes: HashMap<String, String>,
+    ) -> Result<Credential>;
+}
+
+impl Authority {
```
polvorin: if this is going to be used on tests only,   would it be possible to move out of here and into something only available on tests?.
If not,    I think we should rename it to something else,   like   `ToyCredentialIssuer` or something.

etorreborre: I think that it's on the same level as an `InMemoryVault`. My inclination is that, at this stage it is more valuable to leave it in the toolkit in order to let people build prototypes more easily rather than enforcing a strong separation between "production" code and "test/prototype" code. 

That being said we could maybe get a better control on the situation by putting this behind a feature flag. What do you think?

Edit: actually, since _any identity_ can act as a credential issuer maybe it makes sense to formally provide this capability?

etorreborre: I think that `CredentialIssuer` might also be a better name.

mrinalwadhwa: Every identity should be able to issue credentials.

CredentialIssuer is a better name, Authority implies trust in some context.
Any identity should be able to issue credentials but it may not be a trusted authority in any context.

examples/rust/get_started/examples/06-credentials-exchange-alice.rs
```diff
@@ -0,0 +1,49 @@
...
+    let alice = Identity::create(&ctx, &Vault::create()).await?;
+
+    // Create a client to the Authority
+    let authority_route = route![(TCP, "127.0.0.1:5000"), "authority"];
+    let authority = AuthorityClient::new(&ctx, authority_route).await?;
+
+    // Get a credential for Alice
+    let credential = authority
+        .get_attribute_credential(alice.identifier(), "name", "alice")
+        .await?;
```
mrinalwadhwa: There must be some criteria for why the authority will give the credential for alice when someone asks for it.

examples/rust/get_started/examples/06-credentials-exchange-bob.rs
```diff
@@ -0,0 +1,55 @@
...
+    bob.start_credential_exchange_worker(
+        vec![authority_identity],
+        "credential_exchange",
+        true,
+        AuthenticatedAttributeStorage::new(bob.authenticated_storage().clone()),
+    )
+    .await?;
+
+    // Create a TCP listener and wait for incoming connections
+    tcp.listen("127.0.0.1:4000").await?;
```
SanjoDeundiak: Ideally, that should be the last step

implementations/rust/ockam/ockam/src/access_control/attribute_access_control.rs
```diff
@@ -0,0 +1,105 @@
...
+            a
+        } else {
+            return Ok(false);
+        };
+
+        let mut e = Env::new();
+
+        for (k, v) in entry.attrs() {
+            if let Ok(s) = str::from_utf8(v) {
+                e.put(format!("subject.{k}"), Str(s.to_string()));
```
SanjoDeundiak: I think the original implementation had some additional checks (e.g. whitespace checks). Do you think they are important?

examples/rust/get_started/examples/06-credentials-exchange-alice.rs
SanjoDeundiak: I'm slightly concerned about how this example differs from the real-world use case we would like to demonstrate to people eventually. In real-world:
1. Attributes are part of the Authority's preknowledge about the subjects obtained through some side-channel
2. Authority info (at least public key) is all subjects' preknowledge obtained through some side-channel
3. Subject's IdentityId in case of using this API through the routing should be obtained from the SecureChannel.

SanjoDeundiak:
@etorreborre could you please add `(rust)` part to the commit message where applicable?

implementations/rust/ockam/ockam_identity/src/authority.rs
etorreborre: I changed the name

examples/rust/get_started/examples/06-credentials-exchange-bob.rs
etorreborre: done.

implementations/rust/ockam/ockam/src/access_control/attribute_access_control.rs
etorreborre: I did not put them here because it seems to me that the attributes should be checked before being put in storage in the first place. Putting them in storage then always ignoring them with a warning is a bit late in the cycle.

etorreborre:
> @etorreborre could you please add (rust) part to the commit message where applicable?

Ok, I thought it was mostly important for `feat` but I will add it to other commit topics.

examples/rust/get_started/examples/06-credentials-exchange-alice.rs
etorreborre: Thanks all for the comments. I pushed a new commit to adjust things:

 - the credential issuer is now pre-loaded with attributes for both bob and alice
 - they both retrieve credentials over a secure channel
 - they are the only 2 identities allowed to access their credential

I am trying to strike a good balance between:

 - representing correctly the interactions between the 3 entities
 - keeping it simple and not showing the full enrollment process
 - having some code which can potentially be extended / modified for prototyping

What do you all think? @mrinalwadhwa @SanjoDeundiak @polvorin

hairyhum: Would it make more sense instead of pre-loading the credential issuer with attributes to provide a mechanism to retrieve them? To be more in-line with `Attributes are part of the Authority's preknowledge about the subjects obtained through some side-channel`
This is basically how the current implementation handles `credential` requests (using authenticated storage `store` and `get_member`), but current implementation mixes in enrollers and tokens into the same module, complicating it unnecessary.
This way we can separate mechanical creation of credential structure (with signature) and attesting to a set of attributes.

etorreborre: I am not sure I understand what you mean exactly. This example tries to avoid showing an enrollment process and provides a function to declare which identity possesses which attribute during the setup of the `CredentialIssuer` worker. It is a "preloading" in that sense but this function could also be made available through an external API.

I mostly tried to show how credentials could be retrieved and used rather than how the attributes were setup. This corresponds to the first part of what Pablo describes in https://github.com/build-trust/codebase/issues/511.

hairyhum: Yes. My point is that we keep mixing up the attribute and authorized identities management with credential issuance. 
If we had an API like the command proposed earlier:
```
ockam credential issue \
	--as i1
	--for Pce361cd9c4e8c1764ea201e6b1f7fab6426b8481cab0323f658e0da7341ffc1d \
	--attribute application="Smart Factory" \
	--attribute city="New York" \
  > credential.txt
 ```
Which accepts subject identity and attributes as an input, this to me sounds like a credential issuer component. I'm going to call that credential issuer.
Credential issuer has limited use in real-life scenarios because we want authentication to happen, but it is useful for development scenarios.
This API does not authenticate the subject and does not attest to a set of attributes. All it does is mechanical credential creation. This is represented in your case as `self.identity.issue_credential`.

Then we have the authenticator component, which takes subject identity and some authentication metadata and either confirms a set of attributes for it or denies authentication. This is what we have for okta or token or enrolled authentication.
We want to have pluggable authenticator components to support different mechanisms like okta, token, ldap etc.

Then we use those attributes from authenticator, and create a credential based on them, this is an integration component which combines authenticator and credential issuer.
In the current implementation we mix together enrollment based authenticator with credential issuer and then plug in other authenticators into the enrollment authenticator, which is confusing.

In this implementation it seems like you're creating a manual authenticator (using function API) and then combine it with credential issuance code and then call it credential issuer.

I think the better approach would be if we had a credential issuer which can either accept attributes as an input, or as a "source" component like a storage we use right now. Maybe we could support both or just one of those.
The credential issuer in this case has only one api: issue_credential(id, optional<attributes>)
Then for the "manual attributes mapping" to set up which attributes are allowed to be attested, we can have a "manual authenticator" component, which will have an api like: register_identity(id, attributes) (`put_attribute_value` in your implementation)
And if we have a clear integration point between those, we can develop credentials component and authentication components independently.

etorreborre: If I understand you correctly I am not sure if we can easily do that without more refactoring of the `Identity` code in the rust library. An `Identity` is used to back-up the `CredentialIssuer` in PR and that `Identity` bundles its own `Vault` to be able to sign credentials but also an `AuthenticatedStorage` to store attributes. 
What I would like is to decouple all of this and have an `Identity` being just an identifier and a history of key changes, then have a separate `Vault` and a separate `AuthenticatedAttributeStorage`.

That being said the purpose of this example is not really to disentangle the attribute authentication and the credential issuance. It is more to show how the credential exchange is supposed to happen once a credential has been issued. All the rest is part of the setup.

So maybe I can:

 - leave it as it is for now and merge
 - refactor that code with what @polvorin has started on his side when it is ready
 - start documenting the use case in the gitbook
 - start working on how to integrate the credential exchange to the secure channel creation as the 3rd step after key exchange and identity exchange

polvorin (dismissed):

polvorin (dismissed):

SanjoDeundiak (approved):

PR #4285, opened 2023-02-17
feat(rust): add more information to `EntryNotFound` errors

etorreborre:
This can help the logs investigation when an entry is missing from the `Vault`. The messages could be possibly improved with some additional structure since we have a notion of `Origin` in our errors but this PR at least allows us to point faster to the code which created the error.


davide-baldo:
like this! super useful to debug

polvorin (approved):

PR #4289, opened 2023-02-17
feat(rust): allow the route macro to use both routes and addresses

etorreborre:
to create larger routes.

Here is a small example
```
let route1 = route!["s1", "s2"];
let route2 = route!["s4"];
let route3 = route![route1, "s3", route2];
assert_eq!(route3, route!["s1", "s2", "s3", "s4"])
```


davide-baldo (approved):
nice!

PR #4292, opened 2023-02-20
Use ABAC rules in authority services

polvorin:
## Current Behavior

There is a  "Direct"  authenticator  worker,  that implements all functionality related to authorities  (enrolling new members directly,   issuing enrollment tokens,  accepting enrollment tokens,  issuing credentials, etc).  There is a separate set of entities called "enrollers", and those are handled through an explicit file or argument.

## Proposed Changes

Split functionality to separate services.  Give each service appropiate ABAC access rule.    Make "enroller" a _role_,  any member can act as enroller if it has the appropriate attributes according to the ABAC rule for enrollers.

Services:

* `credential_issuer` :    Any `member`  can retrieve its own credential.   ABAC rule:  it must be a member (must be already known by the authority)

* `direct_authenticator`:    Add new members.   ABAC rule:   caller must be an enroller (this is requiring a `ockam-role=enroller`  attribute right now,   but latter we will allow arbitrary rules to be specified at the authority level)    
 
* `enrollment_token_issuer`:  Generates a new enrollment token.   ABAC rule:  caller must be an enroller
 
* `enrollment_token_acceptor`:   Uses an enrollment token to authenticate themselves.

Each service can be started individually, for example:   it is possible to start an  authority node that just serve credentials from trust-anchors,  and not allow any enrollment at all.    For this,   only the `CredentialIssuer` worker (at  `credential_issuer`)  is started,  and only using a PreTrusted set of identities).    `enrollment_token_issuer` and `enrollment_token_acceptor`  together are the `EnrollmentTokenAuthenticator`,  the are split to be able to apply different rules on each. 

Note that existing ABAC access control implementation are attached to [worker' addresses](https://github.com/build-trust/ockam/blob/18b7a0a743cf5c176b88a0a1f27da1852fa83d30/implementations/rust/ockam/ockam_api/src/nodes/service/services.rs#L315-L318) and don't have access to the message' content  (they work at [ockam msg level](https://github.com/build-trust/ockam/blob/18b7a0a743cf5c176b88a0a1f27da1852fa83d30/implementations/rust/ockam/ockam_abac/src/policy.rs#L60),  they don't know about Request-Response protocol for example).    So,  if we want two  operations   `a()`  and `b()` to have different ABAC rules,    _with the current support for abac that we have_ ,  these a()  and b()  must be on different addresses. 

## Compatibility
old clients [are compatible with the new authority](https://github.com/build-trust/ockam/blob/polvorin/split-authority-workers/implementations/rust/ockam/ockam_api/src/authenticator/direct.rs#L45).  This compatibility support can be removed once we don't need to support these clients anymore.

## Testing
Tested to ensure  previously existing data is preserved.  There is two data move/conversions that where needed:
* [ converts from old format to new one](https://github.com/build-trust/ockam/blob/f44507f02b06d282abef7c4727e7bbfa42a242a4/implementations/rust/ockam/ockam_api/src/authenticator/direct.rs#L147-L165)
* [ move file around](https://github.com/build-trust/ockam/blob/f44507f02b06d282abef7c4727e7bbfa42a242a4/implementations/rust/ockam/ockam_api/src/cli_state.rs#L430-L438) (file was relocated on an unrelated PR,  we need this to not lost this data)


implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -45,38 +40,204 @@ const MAX_TOKEN_DURATION: Duration = Duration::from_secs(600);
 /// - `role`: b"member"
 pub const PROJECT_MEMBER_SCHEMA: SchemaId = SchemaId(1);
 pub const PROJECT_ID: &str = "project_id";
+pub const LEGACY_MEMBER: &str = "member";
+
+// This act as a facade, modifying and forwarding incoming messages from legacy clients
+// to the new endpoints.   It's going to be removed once we don't need to maintain compatibility
+// with old clients anymore.
+pub struct LegacyAPIConverter {}
```
etorreborre: That's a small thing but let's write `LegacyApiConverter` instead (for example we write `TcpSendWorker` even if TCP is an acronym)

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -45,38 +40,204 @@ const MAX_TOKEN_DURATION: Duration = Duration::from_secs(600);
 /// - `role`: b"member"
 pub const PROJECT_MEMBER_SCHEMA: SchemaId = SchemaId(1);
 pub const PROJECT_ID: &str = "project_id";
+pub const LEGACY_MEMBER: &str = "member";
+
+// This act as a facade, modifying and forwarding incoming messages from legacy clients
```
etorreborre: s/act/acts

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -45,38 +40,204 @@ const MAX_TOKEN_DURATION: Duration = Duration::from_secs(600);
...
+    fn default() -> Self {
+        LegacyAPIConverter::new()
+    }
+}
 
-pub struct Server<S: AuthenticatedStorage, IS: IdentityAttributeStorage, V: IdentityVault> {
+#[ockam_core::worker]
+impl Worker for LegacyAPIConverter {
+    type Context = Context;
+    type Message = Vec<u8>;
```
etorreborre, line 66: Can we set a proper type for the `Message` type? We are not supposed to decode messages in the body of `handle_message` in the design of workers. Since there seems to be 2 alternatives, `Request` and `Response` we can introduce an enum 
```
enum RequestResponseMessage { 
  RequestMessage(Request), 
  ResponseMessage(Response),
}
```
and pattern-match on that.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -45,38 +40,204 @@ const MAX_TOKEN_DURATION: Duration = Duration::from_secs(600);
...
-    attrs: HashMap<String, String>,
-    generated_by: IdentityIdentifier,
-    time: Instant,
+impl<S, IS, V> CredentialIssuer<S, IS, V>
+where
+    S: AuthenticatedStorage,
+    IS: IdentityAttributeStorageReader,
+    V: IdentityVault,
+{
+    pub async fn new(project: Vec<u8>, store: IS, identity: Identity<V, S>) -> Result<Self> {
```
etorreborre, line 177: Note for later, it would be nice to have a real type for `project` and not a bare list of bytes.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -45,38 +40,204 @@ const MAX_TOKEN_DURATION: Duration = Duration::from_secs(600);
...
+                from   = %from,
+                id     = %req.id(),
+                method = ?req.method(),
+                path   = %req.path(),
+                body   = %req.has_body(),
+                "request"
+            }
+            let res = match (req.method(), req.path()) {
+                (Some(Method::Post), "/") | (Some(Method::Post), "/credential") => {
+                    match self.store.get_attributes(from).await {
```
etorreborre: I would suggest to put all this code inside `impl CredentialIssuer` and just leave inside the `impl Worker` the Request/Response API part for accessing the `CredentialIssuer`. This way we can reuse more code outside of workers and potentially write unit tests more easily for the core logic without having to spawn workers.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -45,38 +40,204 @@ const MAX_TOKEN_DURATION: Duration = Duration::from_secs(600);
...
-impl<S, IS, V> Worker for Server<S, IS, V>
+impl<S, IS, V> Worker for CredentialIssuer<S, IS, V>
 where
     S: AuthenticatedStorage,
-    IS: IdentityAttributeStorage,
+    IS: IdentityAttributeStorageReader,
     V: IdentityVault,
 {
     type Context = Context;
     type Message = Vec<u8>;
```
etorreborre, line 211: Same thing here about the `Message` type which could be a `Request` directly so that we don't have to decode it in `handle_message`

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -45,38 +40,204 @@ const MAX_TOKEN_DURATION: Duration = Duration::from_secs(600);
...
+                            api::forbidden(&req, "unauthorized member").to_vec()?
+                        }
+                        Err(error) => api::internal_error(&req, &error.to_string()).to_vec()?,
+                    }
+                }
+                _ => api::unknown_path(&req).to_vec()?,
+            };
+            c.send(m.return_route(), res).await
         } else {
+            // This was, actually, already checked by the access control. So if we reach this point
```
etorreborre: I think that we should find a better design so that we don't have to second-guess ourselves in the code. But I don't have such design yet :-).

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -85,367 +246,409 @@ where
...
+                id     = %req.id(),
+                method = ?req.method(),
+                path   = %req.path(),
+                body   = %req.has_body(),
+                "request"
+            }
+            let res = match (req.method(), req.path()) {
+                (Some(Method::Post), "/") | (Some(Method::Post), "/members") => {
+                    let add: AddMember = dec.decode()?;
+                    let attrs = add
```
etorreborre: This code seems to be duplicated with [this one](https://github.com/build-trust/ockam/pull/4292/files#diff-60f3798cefa23ecbdd4cfae9d89304aa8d15c4b1a0fb0dec2f29c57c42f73381R468) and [this one](https://github.com/build-trust/ockam/pull/4292/files#diff-a63321d694cf3f5ab5e07b211ec48b17071569f1f5a49b1e7a77c25f1d10f865R94). Removing the duplication would help when making sure that we add the project id all the time.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -85,367 +246,409 @@ where
...
+                            api::internal_error(&req, "Failed to get read lock on tokens table")
+                                .to_vec()?
+                        }
+                    }
+                }
+                _ => api::unknown_path(&req).to_vec()?,
+            };
+            c.send(m.return_route(), res).await
+        } else {
+            // This was, actually, already checked by the access control. So if we reach this point
```
etorreborre: Since this block is repeated many times it could be a good idea to encapsulate it into just one function.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -85,367 +246,409 @@ where
...
+            // identity, not an Option to the peer' identity.
+            let mut dec = Decoder::new(m.as_body());
+            let req: Request = dec.decode()?;
+            let res = api::forbidden(&req, "secure channel required").to_vec()?;
+            c.send(m.return_route(), res).await
+        }
+    }
+}
+
+struct Token {
```
etorreborre, line 511: It could make sense to move this data type to `ockam_identity/credential`, closer to something like `OneTimeCode`

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -85,367 +246,409 @@ where
...
-        Ok(Some(api::forbidden(req, "unauthorized enroller")))
+    pub async fn present_token(&mut self, c: &OneTimeCode) -> Result<()> {
+        self.0
+            .request_no_resp_body(&Request::post("/").body(c))
+            .await
     }
 }
 
-pub struct Client {
+pub struct RPCClient {
```
etorreborre: Naming: `RpcClient` would be better

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -85,367 +246,409 @@ where
...
+        }
+        let msg = err.message().unwrap_or(label);
+        ockam_core::Error::new(Origin::Application, Kind::Protocol, msg)
+    } else {
+        ockam_core::Error::new(Origin::Application, Kind::Protocol, label)
+    }
+}
 
-        Ok(res)
+pub struct CredentialIssuerClient(RPCClient);
```
etorreborre: Nice separation into several clients. Maybe we could:

 - put them into a separate file
 - have a `new` method that takes directly a `Context`, a `Route` and has a default `with_timeout` function so that we don't even expose a `RpcClient` which can potentially do everything

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
```diff
@@ -221,39 +224,170 @@ impl NodeManager {
...
+    async fn build_access_control(
+        &self,
+        r: &Resource,
+        a: &Action,
+        project_id: &str,
+        default: &Expr,
+    ) -> Result<Arc<dyn IncomingAccessControl>> {
+        // Populate environment with known attributes:
+        let mut env = Env::new();
+        env.put("resource.id", str(r.as_str()));
```
etorreborre, line 236: `resource.id` and `action.id` still seem redundant to me but let's tackle this in another PR :-).

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
```diff
@@ -221,39 +224,170 @@ impl NodeManager {
...
+            return Err(ApiError::generic(
+                "Direct Authenticator  service already started",
+            ));
+        }
+        let action = actions::HANDLE_MESSAGE;
+        let resource = Resource::new(&addr.to_string());
+        let project = std::str::from_utf8(proj).unwrap();
+        let rule = and([
+            eq([ident("resource.project_id"), ident("subject.project_id")]),
+            eq([ident("subject.ockam-role"), str("enroller")]),
```
etorreborre, line 304: We should make those strings, `project_id` and `ockam-role` constants. And either use underscore or hyphen as a separator but not both :-).

Another extension to the `ockam_abac` crate would be to provide standard function to create identifiers for `subject`, `action`, `resource`:
```
/// example: resource(PROJECT_ID)
pub fn resource(resource_attribute: &str) -> Expr {
  ident(format!("resource.{}", resource_attribute))
}
```

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
```diff
@@ -221,39 +224,170 @@ impl NodeManager {
...
+
+        // TODO: remove this once compatibility with old clients is not required anymore
+        let legacy_api = crate::authenticator::direct::LegacyAPIConverter::new();
+        ctx.start_worker("authenticator", legacy_api, AllowAll, AllowAll)
+            .await?;
+
+        Ok(())
+    }
+
+    pub(super) async fn start_enrollment_token_authenticator(
```
etorreborre: The function name gives the impression that we are only starting one thing. It would be good to explain that we are starting 2 workers, one to issue token, and another one to accept them.

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
```diff
@@ -221,39 +224,170 @@ impl NodeManager {
...
+        ctx: &Context,
+        issuer_addr: Address,
+        acceptor_addr: Address,
+        proj: &[u8],
+    ) -> Result<()> {
+        if self
+            .registry
+            .authenticator_service
+            .contains_key(&issuer_addr)
+            || self
```
etorreborre, line 347: Given that we always add both keys, is it necessary to check both?

implementations/rust/ockam/ockam_api/src/okta/mod.rs
```diff
@@ -24,7 +23,7 @@ pub struct Server<S> {
 #[ockam_core::worker]
 impl<S> Worker for Server<S>
 where
-    S: AuthenticatedStorage,
+    S: IdentityAttributeStorageWriter,
```
etorreborre, line 26: Nice restriction

implementations/rust/ockam/ockam_api/src/okta/mod.rs
```diff
@@ -88,10 +87,27 @@ where
                         dec.decode()?;
                     debug!("device code received: {token:#?}");
                     if let Some(attrs) = self.check_token(&token.access_token.0).await? {
-                        let encoded_attrs = minicbor::to_vec(attrs)?;
-                        self.store
-                            .set(from.key_id(), MEMBER.to_string(), encoded_attrs)
-                            .await?;
+                        //TODO in some future, we will want to track that this entry
```
etorreborre, line 90: Good point

implementations/rust/ockam/ockam_command/src/service/start.rs
```diff
@@ -354,7 +330,11 @@ pub async fn start_okta_identity_provider(
         cfg.attributes.iter().map(|s| s as &str).collect(),
         cfg.project.as_bytes(),
     );
-    let req = Request::post("/node/services/okta_identity_provider").body(payload);
+    let req = Request::post(format!(
+        "/node/services/{}",
+        DefaultAddress::OKTA_IDENTITY_PROVIDER
```
etorreborre, line 335: 💯

etorreborre (dismissed):
That's a massive piece of work! 

I would mostly reinforce 2 things in workers:

 - have typed messages and not do the decoding inside workers. There are few places in code where we started doing it because we wanted to segregate some interfaces (like an "admin" interface and a "message" interface) but there are many places where we have untyped messages for no particular reason (and even the case of "admin" interfaces could have been done differently)

 - for workers representing request/response services just leave the code doing path matching / request decoding / response encoding in the worker and put all the business logic in the struct supporting the worker

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
polvorin: Not here,  existing clients that this is intended to serve send  `Request`  and that's what we need to handle here,  we can't change to `RequestResponseMessage`.    The fact that it receive `Response`  as well is a bit weird,  but I won't worry much about this particular code as it's going to be removed soon once this compatibility layer is not needed anymore.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
polvorin: :100:  .    And also, lets make that a String not a general byte array.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
polvorin: :+1:   I'll let this for a future issue.      What is on `direct.rs`  now would be better to be split across different files.   And the `RpcClient`   doesn't belong here..  it should likely be on core as it is something not tied to authorities in particular but to the Request-Response  protocol.   And there are couple of such clients  across the code already :/ ,  we need to settle down on just one that is general enough to handle all the cases we need.

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
polvorin: those addrs could potentially be user-supplied like other services addresses.    So  user could start with with  

`"a",  "b"`


and then latter do it again with 

`"aa",  "b"` 

that shouldn't be allowed as "b" is already registered.

That said,  there is higher level discussion to have:
* Do we really want to let user specify addresses?  (so far we do for all services)
* What's the purpose and usage of the `registry`?

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
polvorin: OH, nice,  I didn't even know that functionality was there.     But I'll let this for a future issue..   `Request`  have a lifetime parameter,   and couldn't figure out how to make that play nicely with the existing expectation around `Message`

etorreborre: Oh, I did not realize that the type parameter would get in the way. It seems to me that the `path: Cow<'a, str>` is a premature optimization. I'd rather remove it to make the code simpler at the expense of a string copy per request.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
polvorin: I'll let it out for now,  as the types are not really the same (some are btree, some hashmap,  on some the keys are Strings,  on other CowStr, etc).    This need to be unified.  
A separate thing is why we add project id,   I suspect that is not going to be needed,  or need to be changed to something else   (what's the purpose of "project_id"  if there is no _project_  ,  just nodes and one authority?.    Maybe that's a "project"  but then we need to clarify that vocabulary on other parts)

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
polvorin: yes,  I think a more user-friendly  "Worker"  implementation tailored to handle the Req-Resp protocol  (to do for example dispatch based on the method and path)  and consider abac rules as well.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
polvorin: this should be internal to the EnrollmentTokenIssuer/Acceptor implementation

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
polvorin: agree that this shouldn't belong here.    It must be 2x simpler to create an abac policy and attach it to some worker.  For now this was the only place where we have abac rules created,  so added it here but this is prone to improvement

etorreborre (approved):

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
polvorin: :100:  agree on that

PR #4302, opened 2023-02-21
fix(rust): commands shows concise errors with a more human-readable format

adrianbenavides:
On `ockam_command`, when a `ctx.receive` or `ctx.send_and_receive` call times out it will now show the following error:

```bash
The request timed out, please try again
```

instead of:

```bash
Version 0.80.0, hash: 3e8f61c11df56d116321c92926ef45aafbe16c2d
Version 0.80.0, hash: 3e8f61c11df56d116321c92926ef45aafbe16c2d
{code: 70, err: Failed to receive response from node, cause: deadline has elapsed}


Caused by:
    0: Failed to receive response from node
    1: deadline has elapsed
    2: deadline has elapsed
```

## Another example

Validation errors happening within ockam_command will look like:

```bash
Argument '--to' is invalid. Caused by: `node `n1`` not found
```

instead of:

```bash
Version 0.80.0, hash: 3e8f61c11df56d116321c92926ef45aafbe16c2d
Argument '--to' is invalid

Caused by:
    0: `node `n1`` not found
    1: `node `n1`` not found
```

## Underlying problem: nested errors

If you observe the first example, the `Version 0.80...` string and the description and cause messages are duplicated. This happens because there is a nested conversion happening between `anyhow` and `ockam_command` errors. This can be fixed by changing the signature of the `ockam_command` methods to return an `ockam_command::Error` instead of an `anyhow::Error`. This is out of the scope of this PR, but will be addressed shortly.

To avoid confusion, I've removed printing the binary version. I've found cases where it's printed three times. This can be readded after fixing the problem described above.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/util/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/util/mod.rs
@@ -154,7 +154,7 @@ impl<'a> Rpc<'a> {
             .await
             .map_err(|_err| {
                 // Overwrite error to swallow inner cause and hide it from end-user
-                anyhow!("The request timed out, please try again")
+                anyhow!("The request timed out, please make sure the command's arguments are correct or try again")
             })?;
         Ok(())
     }
@@ -174,7 +174,7 @@ impl<'a> Rpc<'a> {
             .await
             .map_err(|_err| {
                 // Overwrite error to swallow inner cause and hide it from end-user
-                anyhow!("The request timed out, please try again")
+                anyhow!("The request timed out, please make sure the command's arguments are correct or try again")
             })?;
         Ok(())
     }
--- a/implementations/rust/ockam/ockam_command/src/util/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/util/mod.rs
@@ -152,11 +152,13 @@ impl<'a> Rpc<'a> {
             .ctx
             .send_and_receive(route.clone(), req.to_vec()?)
             .await
-            .context("Failed to receive response from node")?;
+            .map_err(|_err| {
+                // Overwrite error to swallow inner cause and hide it from end-user
+                anyhow!("The request timed out, please try again")
+            })?;
         Ok(())
     }
 
-    #[allow(unused)]
     pub async fn request_with_timeout<T>(
         &mut self,
         req: RequestBuilder<'_, T>,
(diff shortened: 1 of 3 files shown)


implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -152,11 +152,13 @@ impl<'a> Rpc<'a> {
             .ctx
             .send_and_receive(route.clone(), req.to_vec()?)
             .await
-            .context("Failed to receive response from node")?;
+            .map_err(|_err| {
+                // Overwrite error to swallow inner cause and hide it from end-user
+                anyhow!("The request timed out, please try again")
```
etorreborre: Is a timeout the only possible error here? `_err` does not bring any useful information? Should it still be logged and not just dropped?

davide-baldo: Since it's a generic code path shouldn't we avoid the `please try again` part?

adrianbenavides: > Is a timeout the only possible error here?

Yes, since the NodeManager [will return a response with the error data](https://github.com/build-trust/ockam/blob/f0a16e18a77b60e0f6c1b993db699e44c478c40c/implementations/rust/ockam/ockam_api/src/nodes/service.rs#L757).

> _err does not bring any useful information?

I think not in this context. It contains low-level information of what went wrong. I'm not sure how to translate that information into meaningful data for the command user. I guess that, in case the user wants to know the technical detail of what went wrong, they should look at the logs?

> Should it still be logged and not just dropped?

It is [logged in the NodeManager](https://github.com/build-trust/ockam/blob/f0a16e18a77b60e0f6c1b993db699e44c478c40c/implementations/rust/ockam/ockam_api/src/nodes/service.rs#L746).

Logging the error on the ockam_command would still be useful if the user runs the command with logs enabled, so they don't have to go to the node's logs. Do you think is worth it?

etorreborre: >  I guess that, in case the user wants to know the technical detail of what went wrong, they should look at the logs?
> Do you think is worth it?

No I think it's ok as long as the log message exists somewhere to do the diagnostic.

etorreborre (dismissed):

implementations/rust/ockam/ockam_command/src/util/mod.rs
adrianbenavides: > Since it's a generic code path shouldn't we avoid the please try again part?

@davide-baldo do you think "The request timed out" would be enough? Or would you add a generic help text as to what to do next?

davide-baldo: Maybe something along the lines "check your parameters or try again" to highlight the possibility that the timeout might not the root issue

adrianbenavides: I've changed the message a little.

etorreborre (approved):

PR #4314, opened 2023-02-21
feat(rust): override inlet policy when starting kafka services

davide-baldo:
The current node manager policy expects the project identity to have `project_id` attribute, but currently, this is not true. When starting the kafka service it overrides the policy, checking if the secure channel identity is the same as the project.

## Current Behavior

A manual policy override is needed to make the kafka demo work:
`ockam policy set --at consumer -r tcp-inlet -a handle_message -e true`

## Proposed Changes
The policy override is automatic and specific for the project identity, no manual policy override is needed

### DIFF
--- a/implementations/rust/ockam/ockam_abac/src/policy.rs
+++ b/implementations/rust/ockam/ockam_abac/src/policy.rs
@@ -93,60 +93,53 @@ where
             return Ok(false);
         };
 
+        let mut environment = self.environment.clone();
+
         // Get identity attributes and populate the environment:
-        let attrs = if let Some(a) = self.attributes.get_attributes(&id).await? {
-            a
-        } else {
-            log::debug! {
-                resource = %self.resource,
-                action   = %self.action,
-                id       = %id,
-                "attributes not found; access denied"
-            }
-            return Ok(false);
-        };
-
-        let mut e = self.environment.clone();
-
-        for (k, v) in attrs.attrs() {
-            if k.find(|c: char| c.is_whitespace()).is_some() {
-                log::warn! {
-                    resource = %self.resource,
-                    action   = %self.action,
-                    id       = %id,
-                    key      = %k,
-                    "attribute key with whitespace ignored"
-                }
-            }
-            match str::from_utf8(v) {
-                Ok(s) => {
-                    if !self.overwrite && e.contains(k) {
-                        log::debug! {
-                            resource = %self.resource,
-                            action   = %self.action,
-                            id       = %id,
-                            key      = %k,
-                            "attribute already present"
-                        }
-                        continue;
-                    }
-                    e.put(format!("subject.{k}"), str(s.to_string()));
-                }
-                Err(e) => {
+        if let Some(attrs) = self.attributes.get_attributes(&id).await? {
+            for (key, value) in attrs.attrs() {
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
```diff
@@ -354,6 +355,21 @@ impl NodeManager {
...
 
+        // override default policy to allow incoming packets from the project
+        if let Some(project_name) = &project_name {
+            let (_addr, identity_identifier) = self.resolve_project(project_name)?;
+            self.policies
+                .set_policy(
+                    &resources::INLET,
+                    &actions::HANDLE_MESSAGE,
+                    &eq([
+                        ident("subject.identity"),
```
hairyhum: What happens if this attribute is missing?
We had some issue with missing attributes evaluation and I'm not sure what implementation ended up being like.

implementations/rust/ockam/ockam_abac/src/policy.rs
```diff
@@ -93,60 +93,53 @@ where
...
-                        err      = %e,
-                        "failed to interpret attribute as string"
                     }
                 }
             }
-        }
+        };
+
+        //add the identity itself as a subject parameter
+        environment.put("subject.identity", str(id.to_string()));
```
hairyhum: Should we call this `identity_id` or `identifier` for clarity?

hairyhum (commented):
Two comments regarding identity.

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
davide-baldo: Being the identity identifier itself can't be missing from a secure channel context, the check would fail before being evaluated.

https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_abac/src/policy.rs#L93

implementations/rust/ockam/ockam_abac/src/policy.rs
davide-baldo: maybe identifier is a bit better to be more consistent

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
hairyhum: What if we're not in a secure channel context? When we start a service are we always guaranteed to have a secure channel?

hairyhum: I must be missing the place where we enforce that the message comes from a secure channel.
There was an issue with attributes missing vs attribute values mismatch in the `eq` rule in the rust implementation, which might cause some errors.

hairyhum:
It seems to be working fine with orchestrator and with local project node.

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
etorreborre: As @davide-baldo pointed out, on line 93 we check if some identity info is associated to the message. If the message does not come from a secure channel, then this is not the case and we don't evaluate the policy

etorreborre (approved):

PR #4320, opened 2023-02-22
refactor(rust): all functions from ockam_command now return a `crate::Result`

adrianbenavides:
To avoid problems with conversions between `crate::Result` and `anyhow::Result`, now all functions from `ockam_command` return the local error.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/error.rs
+++ b/implementations/rust/ockam/ockam_command/src/error.rs
@@ -1,6 +1,6 @@
-use std::convert::Infallible;
 use std::fmt::{Debug, Display, Formatter};
 
+use crate::version::Version;
 use crate::{exitcode, ExitCode};
 
 pub type Result<T> = std::result::Result<T, Error>;
@@ -29,6 +29,7 @@ impl Error {
 
 impl Display for Error {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "{}", Version::short())?;
         if let Some(cause) = &self.cause {
             writeln!(f, "{}. Caused by: {}", self.description, cause)?;
         } else {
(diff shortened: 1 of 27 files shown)


implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
```diff
@@ -105,10 +104,9 @@ async fn rpc(ctx: Context, (opts, mut cmd): (CommandGlobalOpts, CreateCommand))
     Ok(())
 }
 
-fn alias_parser(arg: &str) -> anyhow::Result<String> {
-    ensure! {
-        !arg.contains(':'),
-        "an inlet alias must not contain ':' characters"
+fn alias_parser(arg: &str) -> Result<String> {
+    if arg.contains(':') {
```
etorreborre, line 108: Why `panic` here and not return an `Err`?

implementations/rust/ockam/ockam_command/src/tcp/outlet/create.rs
```diff
@@ -94,10 +93,9 @@ fn make_api_request<'a>(cmd: CreateCommand) -> crate::Result<RequestBuilder<'a,
     Ok(request)
 }
 
-fn alias_parser(arg: &str) -> anyhow::Result<String> {
-    ensure! {
-        !arg.contains(':'),
-        "an outlet alias must not contain ':' characters"
+fn alias_parser(arg: &str) -> Result<String> {
+    if arg.contains(':') {
+        panic!("an inlet alias must not contain ':' characters");
```
etorreborre: Same thing here

implementations/rust/ockam/ockam_command/src/util/config.rs
```diff
@@ -104,7 +104,8 @@ impl AuthoritiesConfig {
         let mut cfg = self.inner.write();
         cfg.add_authority(i, a);
         drop(cfg);
-        self.inner.persist_config_updates()
+        self.inner.persist_config_updates()?;
```
etorreborre: I think that if you drop the `?;` then the last `Ok(())` is not necessary

etorreborre (dismissed):
Nice one!

implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
adrianbenavides: I thought about that, but I didn't want to change the behaviour. It doesn't make much sense, so I'm going to change it.

implementations/rust/ockam/ockam_command/src/util/config.rs
adrianbenavides: Not 100% sure, but I think I added that to make the compiler happy (due to error conversions). I'll review it.

adrianbenavides: Yeah exactly, it would look like: `Ok(self.inner.persist_config_updates()?)`

etorreborre: Just `self.inner.persist_config_updates()` doesn't compile? Isn't it returning the same `Result` type?

implementations/rust/ockam/ockam_command/src/error.rs
```diff
@@ -46,50 +47,24 @@ impl From<anyhow::Error> for Error {
...
+gen_from_impl!(std::fmt::Error, SOFTWARE);
+gen_from_impl!(std::net::AddrParseError, DATAERR);
+gen_from_impl!(hex::FromHexError, DATAERR);
+gen_from_impl!(serde_bare::error::Error, DATAERR);
+gen_from_impl!(serde_json::Error, DATAERR);
+gen_from_impl!(minicbor::encode::Error<std::convert::Infallible>, DATAERR);
+gen_from_impl!(minicbor::decode::Error, DATAERR);
+gen_from_impl!(ockam::Error, SOFTWARE);
+gen_from_impl!(ockam_api::cli_state::CliStateError, SOFTWARE);
+gen_from_impl!(ockam_multiaddr::Error, SOFTWARE);
```
davide-baldo, line 70: nice trick! :+1:

implementations/rust/ockam/ockam_command/src/util/config.rs
adrianbenavides: > Isn't it returning the same Result type?

No, that part of the code is old and it works with anyhow (it's also part of another crate: `ockam_api`). I have plans to remove it altogether, that's why I didn't modify it.

etorreborre: I get it, thanks.

etorreborre (approved):

etorreborre (approved):

etorreborre (approved):

mrinalwadhwa (approved):

PR #4322, opened 2023-02-22
fix(rust): increase waiting time after addon config change

polvorin:
this is a temporary workaround, project is not fully ready after addon re-configuration. A real solution is being worked on.


etorreborre (approved):
Time for a constant at the beginning of the file :-)?

PR #4323, opened 2023-02-22
feat(rust): add `vault` argument to `secure-channel listener` command

adrianbenavides:
The `secure-channel listener` was only accepting an `identity` argument to override the node's default identity, but that wasn't enough for scenarios where the overridden identity is using a vault that is not the node's default.

This PR adds a new `vault` argument to that command to support the following scenarios:

```bash
# Setup
$OCKAM node delete --all &>/dev/null
$OCKAM vault delete v1 &>/dev/null
$OCKAM vault delete v2 &>/dev/null
$OCKAM identity delete i1 &>/dev/null
$OCKAM identity delete i2 &>/dev/null
echo "+ removed state"

$OCKAM vault create v1 &>/dev/null
$OCKAM identity create i1 --vault v1 &>/dev/null
idt1=$($OCKAM identity show i1)

$OCKAM vault create v2 &>/dev/null
$OCKAM identity create i2 --vault v2 &>/dev/null
idt2=$($OCKAM identity show i2)
```

Scenario 1: setting vault/identity at node creation
```bash
$OCKAM node create n1 &>/dev/null
$OCKAM node create n2 &>/dev/null --vault v2 --identity i2
echo "+ setup done"

$OCKAM secure-channel-listener create l --at n2 --authorized-identifiers $idt1 &>/dev/null
echo "+ secure-channel-listener created"
$OCKAM secure-channel create --from n1 --to /node/n2/service/l --authorized $idt2
```

Scenario 2: setting vault/identity at sc listener creation
```bash
$OCKAM node create n1 &>/dev/null
$OCKAM node create n2 &>/dev/null
echo "+ setup done"

$OCKAM secure-channel-listener create l --at n2 --vault v2 --identity i2 --authorized-identifiers $idt1 &>/dev/null
echo "+ secure-channel-listener created"
$OCKAM secure-channel create --from n1 --to /node/n2/service/l --authorized $idt2
```

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
@@ -155,6 +155,7 @@ impl NodeManager {
         &mut self,
         addr: Address,
         authorized_identifiers: Option<Vec<IdentityIdentifier>>,
+        vault_name: Option<CowStr<'_>>,
         identity_name: Option<CowStr<'_>>,
         ctx: &Context,
     ) -> Result<()> {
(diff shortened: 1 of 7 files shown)


implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
```diff
@@ -166,12 +167,11 @@ impl NodeManager {
         let identity = if let Some(identity) = identity_name {
```
etorreborre, line 167: Should we warn users if `vault` is set but not `identity`? Because then the `vault` argument is silently not used.

etorreborre (dismissed):
👍 with maybe a possible slight improvement

implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
adrianbenavides: Yes, that makes total sense. We can enforce that at the command layer as well, so the user can't run the command passing only the `vault` argument.

etorreborre (approved):

PR #4326, opened 2023-02-23
fix(rust): update project readiness check to include authority

caoakleyii:
## Current Behavior

Currently the check project readiness function only attempts to establish a secure channel with the project and NOT the project authority.

When sending API calls to orchestrator the api calls will also time out within 30seconds

## Proposed Changes

Checking the project readiness should also include testing a secure channel to the project authority.

API Calls made to orchestrator that can possibly require longer responses due to pod/node restarts, have a timeout increased to 180seconds.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cloud/enroll.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/enroll.rs
@@ -19,6 +19,8 @@ impl Token {
 }
 
 mod node {
+    use std::time::Duration;
+
     use minicbor::Decoder;
     use tracing::trace;
 
@@ -29,7 +31,7 @@ mod node {
 
     use crate::cloud::enroll::auth0::AuthenticateAuth0Token;
     use crate::cloud::enroll::enrollment_token::{EnrollmentToken, RequestEnrollmentToken};
-    use crate::cloud::CloudRequestWrapper;
+    use crate::cloud::{CloudRequestWrapper, ORCHESTRATOR_RESTART_TIMEOUT};
     use crate::nodes::NodeManagerWorker;
     use ockam_identity::credential::Attributes;
 
@@ -55,7 +57,7 @@ mod node {
                 inner.identity()?.async_try_clone().await?
             };
 
-            self.request_controller(
+            self.request_controller_with_timeout(
                 ctx,
                 api_service,
                 None,
@@ -63,6 +65,7 @@ mod node {
                 api_service,
                 req_builder,
                 ident,
+                Duration::from_secs(ORCHESTRATOR_RESTART_TIMEOUT),
             )
             .await
         }
@@ -118,7 +121,7 @@ mod node {
             };
 
             trace!(target: TARGET, "authenticating token");
-            self.request_controller(
+            self.request_controller_with_timeout(
                 ctx,
                 api_service,
                 None,
@@ -126,6 +129,7 @@ mod node {
                 api_service,
                 req_builder,
                 ident,
+                Duration::from_secs(ORCHESTRATOR_RESTART_TIMEOUT),
             )
             .await
         }
--- a/implementations/rust/ockam/ockam_command/src/project/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/project/create.rs
@@ -1,4 +1,7 @@
+use std::time::Duration;
+
 use clap::Args;
+use ockam_api::cloud::ORCHESTRATOR_RESTART_TIMEOUT;
 use rand::prelude::random;
 
 use ockam::Context;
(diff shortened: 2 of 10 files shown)


implementations/rust/ockam/ockam_api/src/cloud/enroll.rs
```diff
@@ -118,14 +121,15 @@ mod node {
...
                 ctx,
                 api_service,
                 None,
                 cloud_route,
                 api_service,
                 req_builder,
                 ident,
+                Duration::from_secs(ORCHESTRATOR_RESTART_TIMEOUT),
             )
             .await
```
caoakleyii, line 134: My current understanding is that this request and the above request require the timeout.

However, I'm not entirely certain that Orchestrator projects are restarted in this process

implementations/rust/ockam/ockam_command/src/project/create.rs
```diff
@@ -51,11 +54,10 @@ async fn run_impl(
     let space_id = space::config::try_get_space(&opts.config, &cmd.space_name)?;
     let node_name = start_embedded_node(ctx, &opts, None).await?;
     let mut rpc = RpcBuilder::new(ctx, &opts, &node_name).build();
-    rpc.request(api::project::create(
-        &cmd.project_name,
-        &space_id,
-        &cmd.cloud_opts.route(),
-    ))
+    rpc.request_with_timeout(
```
adrianbenavides, line 57: I'm not sure if using longer timeouts on the command's RPC will do the effect we want. This will send a request to a NodeManager instance, which will in turn do a `ctx.send_and_receive` call with the default 30s timeout, receiving the timeout error on the command after 30s.

Maybe [retrying a few times like its done here](https://github.com/build-trust/ockam/blob/a1367541da9826f248a682baeed0b18936e63e87/implementations/rust/ockam/ockam_command/src/project/addon.rs#L452) would be more reliable.

davide-baldo:
I've manually tested this PR 5 times, the issue seems to be resolved :+1:

implementations/rust/ockam/ockam_command/src/project/create.rs
hairyhum: The idea is not to fix the problem with it. It is to give a possibility for the orchestrator to fix the problem by making the orchestrator wait with responding to this request until all resources are updated.
This in turn will cause a timeout error on the command side if it's not configured with a longer timeout.

caoakleyii: Just as an fyi, this PR updates the request in **both** places:

Command -> NodeManager, as well as NodeManager -> Orchestrator.

If this is a workable solution, both need to be updated

adrianbenavides: > Command -> NodeManager, as well as NodeManager -> Orchestrator.

Yeah, that's what I didn't see, sorry. I thought the modified `Command -> NodeManager` endpoints were different from the `NodeManager -> Orchestrator` ones.

caoakleyii:
image

implementations/rust/ockam/ockam_command/src/project/util.rs
```diff
@@ -174,91 +176,137 @@ pub async fn check_project_readiness<'a>(
...
+                &project_route,
+                &project_identity,
+                CredentialExchangeMode::None,
+                None,
+            )
+            .await
+            {
                 // Try to delete secure channel, ignore result.
                 let _ = delete_secure_channel(ctx, opts, api_node, tcp, &sc_addr).await;
+                println!("✅");
```
mrinalwadhwa: @CAOakleyII could you change the emojis to `✔︎`. We don't use emojis anywhere else in the interface so far, so let's make that change more consistently if decide its a good option.

caoakleyii: Absolutely!

caoakleyii: I wanted some sort of feedback that it worked, took a guess on the emoji, but agreed it feels out of place until a style is decided upon.

mrinalwadhwa: great call on - we need some feedback there 👍

caoakleyii:
No more emojis

image

implementations/rust/ockam/ockam_command/src/project/util.rs
```diff
@@ -174,91 +177,138 @@ pub async fn check_project_readiness<'a>(
     tcp: Option<&TcpTransport>,
     mut project: Project<'a>,
 ) -> Result<Project<'a>> {
+    // 3 Mins (180 Seconds) per strategy with 5 second intervals
```
hairyhum: Let's make it 10 minutes.

hairyhum (approved):

BeenzSyed (approved):

PR #4337, opened 2023-02-24
kafka refactors and test fix

davide-baldo:
## Current Behavior
Pieces of test code are present in the production code.
MacOSX tests are broken since it supports `127.0.0.1` and not the whole RFC block `127/8`

## Proposed Changes

The test code was removed from production.
Using port `0` to let the os choose an available port, fixing the MacOSX issue and removing any possibility of conflict.
Some minor changes were also applied: few cleanups and some files were split into multiple files for readability.


implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -117,6 +117,7 @@ pub struct NodeManager {
         BootstrapedIdentityStore<PreTrustedIdentities, AuthenticatedAttributeStorage<LmdbStorage>>,
 }
 
+#[derive(Clone)]
```
SanjoDeundiak: Interesting, where is that used?

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_api/src/nodes/service.rs
davide-baldo: I forgot to remove it, I was going to call APIs directly from within the tests, but they are not accessible from outside and are practically equivalent to sending the messages, so I ended up not using it.

davide-baldo: Modified the first commit to remove it :+1:

SanjoDeundiak (approved):

PR #4338, opened 2023-02-24
fix(rust): fix broken test by preventing a premature drop

davide-baldo:
## Current Behavior

test `kafka_portal_worker__metadata_exchange__response_changed` fails randomly.
The issue is `OCKAM_HOME` being read multiple times during tests, this creates a race condition between tests, causing random failure.

## Proposed Changes

Having a single read of `OCKAM_HOME` during the command bootstrap and having everyone else rely on a provided path instead.

## NOTES

This makes the patch quite time-consuming, some items that were built using `Default::default()` cannot be built this way anymore,  and it quickly propagates to a sizable piece of the codebase.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state.rs
@@ -62,26 +62,28 @@ pub struct CliState {
 }
 
 impl CliState {
-    pub fn new() -> Result<Self> {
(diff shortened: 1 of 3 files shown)


SanjoDeundiak (approved):

adrianbenavides:
Should this be closed now that https://github.com/build-trust/ockam/pull/4347 is merged?

PR #4341, opened 2023-02-27
feat(rust): added cmd to delete a tcp-outlet by alias and node

### LINKED ISSUE #4268: Create clap command `ockam tcp-outlet delete` to delete a TCP Outlet on a node
## Summary

A node has set of TCP Outlets. This command should delete an outlet of a specific alias
```
ockam tcp-outlet delete [alias] --node [node-name]
```
The command should have a positional argument `[alias]` which is the alias of the tcp outlet to be deleted.  
The command should accept a `--node` flag, however default to the default node if one is not provided.

Code of where outlets are created and stored can be found [here](https://github.com/build-trust/ockam/blob/65c7259a2b96e85e2745397eaaf77e7dcf427960/implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs#L257-L323).

`tcp-outlet` commands are defined [here](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/tcp/outlet/mod.rs).
## Expected Behavior

```sh
ockam tcp-outlet create --at /node/n1 --from /service/outlet --to 127.0.0.1:5000 --alias "my-outlet"
...
ockam tcp-outlet delete my-outlet --node n1
Deleted TCP Outlet 'my-outlet' on node 'n1'
```

nidnogg:
## Current Behavior

There is no way to delete a tcp-outlet from a node directly.

## Proposed Changes

This PR resolves https://github.com/build-trust/ockam/issues/4268 by adding a `tcp-outlet delete` to `ockam::command`. 

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
@@ -112,6 +112,8 @@ impl<'a> CreateOutlet<'a> {
 #[rustfmt::skip]
 #[cbor(map)]
 pub struct DeleteOutlet<'a> {
+    #[cfg(feature = "tag")]
+    #[n(0)] tag: TypeTag<1193889>,
     /// The alias of the TCP outlet to be deleted
     #[b(1)] pub alias: CowStr<'a>,
     #[b(2)] pub addr: CowStr<'a>
--- a/implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
@@ -116,18 +116,14 @@ pub struct DeleteOutlet<'a> {
     #[n(0)] tag: TypeTag<1193889>,
     /// The alias of the TCP outlet to be deleted
     #[b(1)] pub alias: CowStr<'a>,
-    #[b(2)] pub addr: CowStr<'a>
 }
 
 impl<'a> DeleteOutlet<'a> {
-    pub fn new(
-        alias: impl Into<CowStr<'a>>,
-        addr: impl Into<CowStr<'a>>,
-
-    ) -> Self {
+    pub fn new(alias: impl Into<CowStr<'a>>) -> Self {
         Self {
+             #[cfg(feature = "tag")]
+            tag: TypeTag,
             alias: alias.into(),
-            addr: addr.into(),
         }
     }
 }
--- a/implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
@@ -11,7 +11,7 @@ use ockam_core::TypeTag;
 use ockam_identity::IdentityIdentifier;
 use ockam_multiaddr::MultiAddr;
 
-/// Request body to create an inlet or outlet
+/// Request body to create an inlet
 #[derive(Clone, Debug, Decode, Encode)]
 #[rustfmt::skip]
 #[cbor(map)]
@@ -76,7 +76,7 @@ impl<'a> CreateInlet<'a> {
     }
 }
 
-/// Request body to create an inlet or outlet
+/// Request body to create an outlet
 #[derive(Clone, Debug, Decode, Encode)]
 #[rustfmt::skip]
 #[cbor(map)]
@@ -103,6 +103,29 @@ impl<'a> CreateOutlet<'a> {
             tcp_addr: tcp_addr.into(),
             worker_addr: worker_addr.into(),
             alias: alias.into(),
+        }
+    }
+}
+
+/// Request body to delete an outlet
+#[derive(Clone, Debug, Decode, Encode)]
+#[rustfmt::skip]
+#[cbor(map)]
+pub struct DeleteOutlet<'a> {
+    /// The alias of the TCP outlet to be deleted
+    #[b(1)] pub alias: CowStr<'a>,
+    #[b(2)] pub addr: CowStr<'a>
+}
+
+impl<'a> DeleteOutlet<'a> {
+    pub fn new(
+        alias: impl Into<CowStr<'a>>,
+        addr: impl Into<CowStr<'a>>,
+
+    ) -> Self {
+        Self {
+            alias: alias.into(),
+            addr: addr.into(),
         }
     }
 }
--- a/implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
@@ -121,7 +121,7 @@ pub struct DeleteOutlet<'a> {
 impl<'a> DeleteOutlet<'a> {
     pub fn new(alias: impl Into<CowStr<'a>>) -> Self {
         Self {
-             #[cfg(feature = "tag")]
+            #[cfg(feature = "tag")]
             tag: TypeTag,
             alias: alias.into(),
         }
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
@@ -1,7 +1,7 @@
 use crate::error::ApiError;
 use crate::nodes::connection::Connection;
 use crate::nodes::models::portal::{
-    CreateInlet, CreateOutlet, InletList, InletStatus, OutletList, OutletStatus, DeleteOutlet,
+    CreateInlet, CreateOutlet, DeleteOutlet, InletList, InletStatus, OutletList, OutletStatus,
 };
 use crate::nodes::registry::{InletInfo, OutletInfo, Registry};
 use crate::nodes::service::random_alias;
@@ -11,7 +11,7 @@ use crate::{local_multiaddr_to_route, try_multiaddr_to_addr};
 use minicbor::Decoder;
 use ockam::compat::asynchronous::RwLock;
 use ockam::compat::tokio::time::timeout;
-use ockam::{Address, AsyncTryClone, Result, TCP};
+use ockam::{Address, AsyncTryClone, Result};
 use ockam_abac::expr::{eq, ident, str};
 use ockam_abac::{Action, Env, PolicyAccessControl, PolicyStorage, Resource};
 use ockam_core::api::{Request, Response, ResponseBuilder};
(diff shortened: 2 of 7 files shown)


implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
```diff
@@ -126,6 +126,29 @@ impl<'a> CreateOutlet<'a> {
     }
 }
 
+/// Request body to delete an outlet
+#[derive(Clone, Debug, Decode, Encode)]
+#[rustfmt::skip]
+#[cbor(map)]
+pub struct DeleteOutlet<'a> {
+    /// The alias of the TCP outlet to be deleted
```
davide-baldo, line 117: You can add a TypeTag here, it's just a random number to validate we are decoding the right message
https://www.random.org/integers/?num=100&min=1000000&max=9999999&col=5&base=10&format=html&rnd=new

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
```diff
@@ -319,6 +319,62 @@ impl NodeManagerWorker {
...
+            addr,
+        } = dec.decode()?;
+
+        let alias = alias.into_owned();
+        info!("Handling request to delete outlet portal");
+
+        let tcp_addr = Address::new(TCP, addr.clone().into_owned());
+        let res = node_manager
+            .tcp_transport
+            .stop_outlet(tcp_addr)
```
davide-baldo: the `Address` argument actually refers to a worker address, and not the `TCP` `SocketAddress`

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
```diff
@@ -126,6 +126,29 @@ impl<'a> CreateOutlet<'a> {
...
+    #[b(1)] pub alias: CowStr<'a>,
+    #[b(2)] pub addr: CowStr<'a>
```
davide-baldo: alias (~~aka worker address~~) is the only parameter you need to identify a TCP outlet within a node

implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs
```diff
@@ -0,0 +1,92 @@
...
+    // This request will use the `.remove()` method from the registry's BTree. (See portals.rs AND portal.rs under ockam API)
+    let node = extract_address_value(&cmd.node_opts.api_node)?;
+    let mut rpc = Rpc::background(&ctx, &options, &node)?;
+    rpc.request(make_api_request(cmd)?).await?;
+
+    let OutletStatus { worker_addr, .. } = rpc.parse_response()?;
+
+    let addr = route_to_multiaddr(&route![worker_addr.to_string()])
+        .ok_or_else(|| ApiError::generic("Invalid Outlet Address"))?;
+    println!("{addr}");
```
davide-baldo: here you can print

`Deleted TCP Outlet 'my-outlet' on node 'n1'`

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
nidnogg: From what I understood, the `worker_addr` refers to the `/service/outlet` argument in the TCP-outlet command. But how can this be derived from the node name + the alias in the command or api side?

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
nidnogg: Does this mean the node's TCP outlet does not have to be stopped before removing the alias from the registry? 

On a related note, isn't the alias distinct from the worker address? I figured that was the case from the [create command here](https://github.com/nidnogg/ockam/blob/nidnogg/delete-tcp-outlet/implementations/rust/ockam/ockam_command/src/tcp/outlet/create.rs#L90-L91).

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
nidnogg: Added in latest commit

implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs
nidnogg: Corrected in latest commit

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
davide-baldo: > Does this mean the node's TCP outlet does not have to be stopped before removing the alias from the registry?

You still have to stop the outlet using  `node_manager.tcp_transport.stop_outlet()` by passing the outlet worker address, you can retrieve the outlet worker address from the registry using the alias.
The alias is **always** present, if it wasn't specified it would be randomly generated, you check the alias of unnamed outlets using `ockam tcp-outlet list`

> On a related note, isn't the alias distinct from the worker address?

yes, you're right, the alias is distinct from the worker's address,

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
davide-baldo:
```
$ ockam tcp-outlet create --at /node/my-node --from my-worker-address --to 127.0.0.1:123 --alias my-alias
$ ockam tcp-outlet list
Outlet:
    Alias: my-alias
    From Outlet: /service/my-worker-address
    To TCP: 127.0.0.1:123
```
You can already specify the node using `--node`, and everything you change in the `node_manager` is relative to the node executing your API on, so there is no need to specify the node again, so `/service/my-worker-address` or `Address::from_string("my-worker-address")` is enough so univocally identify the outlet.

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
```diff
@@ -312,6 +312,58 @@ impl NodeManagerWorker {
...
+            let payload_log = format!("There is no corresponding outlet for the {}", alias);
+            info!(payload_log);
+            Ok(Response::bad_request(req.id()).body(OutletStatus::new(
+                "",
+                "",
+                alias,
+                Some(payload_log.into()),
+            )))
+        }
+    }
```
adrianbenavides, line 360: There are some problems in this function:
- You can get and remove in a single operation with `remove_entry`, which helps simplify the code a little
- Removing an item from the registry should always succeed. This is also fixed by applying the previous change
- It's OK to provide some context in the `payload` field, but we shouldn't rely on it to report errors, as this is probably the only response struct that contains that field (see my comment below about the error handling in the command).

Here is a commit I did on a separate branch with the mentioned changes: https://github.com/build-trust/ockam/commit/2e00a734b015165fdb0d96c4ad9d09a1e8f8c7c4

implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs
```diff
@@ -0,0 +1,64 @@
...
+) -> crate::Result<()> {
+    let alias = cmd.alias.clone();
+
+    let node = extract_address_value(&cmd.node_opts.api_node)?;
+    let mut rpc = Rpc::background(&ctx, &options, &node)?;
+    rpc.request(make_api_request(cmd)?).await?;
+
+    let OutletStatus { payload, .. } = rpc.parse_response()?;
+
+    if let Some(payload_response) = payload {
```
adrianbenavides: You should check for the response status instead of relying on the payload field, as you will only have access to it when a handled error occurs within that specific command handler.

There are some things we want to improve to parse the rpc results better, but for the time being,  calling `rpc.is_ok()?;` to replace this `if let Some` block would be just fine.

implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs
```diff
@@ -0,0 +1,64 @@
...
+/// Construct a request to delete a tcp outlet
+fn make_api_request<'a>(cmd: DeleteCommand) -> crate::Result<RequestBuilder<'a, DeleteOutlet<'a>>> {
+    // let alias = cmd.alias.map(|a| a.into());
+    let payload = DeleteOutlet::new(cmd.alias);
+    let request = Request::delete("/node/outlet").body(payload);
+    Ok(request)
+}
+
+fn alias_parser(arg: &str) -> anyhow::Result<String> {
+    ensure! {
```
adrianbenavides: We recently replaced the `ensure` calls to make functions return an ockam_command::Result instead of an anyhow::Result.

[Check this out](https://github.com/build-trust/ockam/blob/5730fcd4953383f64cc464a3b1caff5caed5691a/implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs#L87) to see how we handle this now.

implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs
```diff
@@ -0,0 +1,64 @@
...
+    rpc.request(make_api_request(cmd)?).await?;
+
+    let OutletStatus { payload, .. } = rpc.parse_response()?;
+
+    if let Some(payload_response) = payload {
+        println!("Error deleting Outlet: {}", payload_response);
+        return Ok(());
+    }
+
+    println!("Deleted TCP Outlet '{}' on node '{}'", alias, node);
```
adrianbenavides: I think clippy will complain about this. It recently started recommending to inline the variables if possible, like: `println!("Deleted TCP Outlet '{alias}' on node '{node}'");`

adrianbenavides (changes requested):
Hey @nidnogg! I left some comments with some suggestions to make the command more consistent. Let me know if you have any further questions 👍

implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs
nidnogg: Solved on latest commit - although clippy has been complaining about several unrelated occurrences in `ockam_vault`. I'm assuming these are not breaking checks?

implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs
nidnogg: Solved in latest commit

implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs
nidnogg: Thanks! I had a lot of trouble deciding on that `payload` field and it didn't really sit well with me in the end. Solved in the latest commit.

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
nidnogg: Thanks! These look much simpler in the end.

Just a small side note - since we ignore the removed `key` in  `remove_entry` with `_` anyway, shouldn't `remove` also fit the bill here? (since it also returns the removed value). 

I edited the function with your changes, but I kept the `remove` for curiosity's sake and it seems to work out OK as well.

implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs
adrianbenavides: Don't worry about those. We'll fix them shortly on a separate PR.

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
adrianbenavides: Yeah, that's right. The remove returns the entry's value, while remove_entry returns both the key and the value. So, using `remove` is the right choice in this case!

adrianbenavides (changes requested):
The code looks perfect now! One last thing I'd ask you is to add a simple CRUD test to verify that is working properly.

You can add it [here](https://github.com/build-trust/ockam/blob/a57a616bea115931ec903f73e22b47f7fafe2f18/implementations/rust/ockam/ockam_command/tests/bats/unit.bats#L331).

Since we don't have the `tcp-outlet show` command yet, a good enough test would be something similar to the [tcp-connection test](https://github.com/build-trust/ockam/blob/a57a616bea115931ec903f73e22b47f7fafe2f18/implementations/rust/ockam/ockam_command/tests/bats/unit.bats#L196). Additionally, you can test that trying to delete the same outlet twice will result in an error.

nidnogg:
> The code looks perfect now! One last thing I'd ask you is to add a simple CRUD test to verify that is working properly.
> 
> You can add it [here](https://github.com/build-trust/ockam/blob/a57a616bea115931ec903f73e22b47f7fafe2f18/implementations/rust/ockam/ockam_command/tests/bats/unit.bats#L331).
> 
> Since we don't have the `tcp-outlet show` command yet, a good enough test would be something similar to the [tcp-connection test](https://github.com/build-trust/ockam/blob/a57a616bea115931ec903f73e22b47f7fafe2f18/implementations/rust/ockam/ockam_command/tests/bats/unit.bats#L196). Additionally, you can test that trying to delete the same outlet twice will result in an error.

Added this test in the latest commit.

adrianbenavides:
Looks great! Just one last step before we can merge it: could you rebase it on top of develop and fix the conflict?

nidnogg:
> Looks great! Just one last step before we can merge it: could you rebase it on top of develop and fix the conflict?

Just pushed the rebased conflict fix

adrianbenavides (approved):
LGTM! Thanks @nidnogg for this contribution, it was great! 💪

PR #4347, opened 2023-02-27
refactor(rust): refactor `CliState` so it can be built using an explicit directory

### LINKED ISSUE #4345: Concurrent tests and CLI state
**Describe the bug**
In cargo, tests are executed in parallel in the same process.
Every test expects to have isolation from other tests, but every test using `okam_api` actually **shares `OCKAM_HOME` across test**, which causes random failures.
Even if tests set `OCKAM_HOME` this causes even more issues since tests will be competing for the same global state.

**To Reproduce**
Randomly reproduce it when running tests with a high amount of threads.
So far has been reproduced by @SanjoDeundiak.

**Solution**
The obvious solution is to avoid the usage of `OCKAM_HOME`  as a global state and make it explicit in every test context, however,  `OCKAM_HOME` is **always implicit** through `CliState::dir()` or in a more indirect way.
Almost every usage of `ockam_api` needs to add the `ockam_home` in the context to pass it along.
To further complicate things, some types are created using `Default::default()` but now they actually require `ockam_home`.

adrianbenavides:
It also refactors the `NodeManager` so it includes a `CliState` instance instead of building a new default instance every time it needs it.

Tries to fix https://github.com/build-trust/ockam/issues/4345

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state.rs
@@ -58,19 +58,23 @@ pub struct CliState {
     pub identities: IdentitiesState,
     pub nodes: NodesState,
     pub projects: ProjectsState,
-    dir: PathBuf,
+    pub dir: PathBuf,
 }
 
 impl CliState {
-    pub fn new() -> Result<Self> {
-        let dir = Self::dir()?;
+    pub fn try_default() -> Result<Self> {
+        let dir = Self::default_dir()?;
+        Self::new(&dir)
+    }
+
+    pub fn new(dir: &Path) -> Result<Self> {
         std::fs::create_dir_all(dir.join("defaults"))?;
         Ok(Self {
-            vaults: VaultsState::new(&dir)?,
-            identities: IdentitiesState::new(&dir)?,
-            nodes: NodesState::new(&dir)?,
-            projects: ProjectsState::new(&dir)?,
-            dir,
+            vaults: VaultsState::new(dir)?,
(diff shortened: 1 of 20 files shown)


davide-baldo:
Seems like `IdentityState` is the only remaining piece using the `CliState::try_default()` (outside a few marginal usages within `ockam_command`), can you remove it to eliminate completely the possibility of future issues?

adrianbenavides:
Ready now, all the CliState::try_default usages come from ockam_command.

implementations/rust/ockam/ockam_api/src/cli_state.rs
```diff
@@ -750,8 +783,8 @@ impl NodeState {
         self.path.join("stderr.log")
     }
 
-    pub fn policies_storage_path(&self) -> PathBuf {
-        self.path.join("policies_storage.lmdb")
+    pub async fn policies_storage(&self) -> Result<LmdbStorage> {
+        Ok(LmdbStorage::new(self.path.join("authenticated_storage.lmdb")).await?)
```
davide-baldo: why `policies_storage.lmdb` becomes `authenticated_storage.lmdb`?

davide-baldo (changes requested):

implementations/rust/ockam/ockam_api/src/cli_state.rs
adrianbenavides: Wow, that's a copy-paste error 🤦  Fixing it!

davide-baldo (approved):

PR #4349, opened 2023-02-27
refactor: removed type parameters exposing implementation details

etorreborre:
The main objective of this PR is to simplify the code using `Identity` by removing its type parameters. 
Those type parameters specify which `IdentityVault` and which `AuthenticatedStorage` are used to implement an `Identity` and don't have to be known by the users of `Identity`. The benefits are:

 - a simpler rust entity to manipulate
 - an isolation of the rest of the code from changes in the `Identity` implementation
 - a removal of the same same type parameters which had to be set on other structs using `Identity`

Removing these type parameters consists in using `Arc<dyn Trait>` (for example `Arc<dyn IdentityVault>`) in the implementation of `Identity`. This removal has a number of knock-on effects on other structs and functions which also had to be parameterized by the type parameters used by `Identity` hence the unfortunately large PR.

I left some comments on the PR itself to help the reviewers, please tell me if I have introduced any possible regression or broken a rust idiom.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/bootstrapped_identities_store.rs
+++ b/implementations/rust/ockam/ockam_api/src/bootstrapped_identities_store.rs
@@ -1,7 +1,8 @@
 use ockam_core::async_trait;
+use ockam_core::compat::sync::Arc;
 use ockam_core::compat::{collections::HashMap, string::String, vec::Vec};
 use ockam_core::errcode::{Kind, Origin};
-use ockam_core::{AsyncTryClone, Result};
+use ockam_core::Result;
 use ockam_identity::authenticated_storage::{
     AttributesEntry, IdentityAttributeStorage, IdentityAttributeStorageReader,
     IdentityAttributeStorageWriter,
@@ -12,19 +13,17 @@ use serde_json as json;
 use std::path::PathBuf;
 use tracing::trace;
 
-#[derive(AsyncTryClone)]
-#[async_try_clone(crate = "ockam_core")]
-#[derive(Debug)]
-pub struct BootstrapedIdentityStore<B: IdentityAttributeStorageReader, S: IdentityAttributeStorage>
-{
-    bootstrapped: B,
-    storage: S,
+#[derive(Clone)]
+pub struct BootstrapedIdentityStore {
+    bootstrapped: Arc<dyn IdentityAttributeStorageReader>,
+    storage: Arc<dyn IdentityAttributeStorage>,
 }
 
-impl<B: IdentityAttributeStorageReader, S: IdentityAttributeStorage>
-    BootstrapedIdentityStore<B, S>
-{
-    pub fn new(bootstrapped: B, storage: S) -> Self {
+impl BootstrapedIdentityStore {
+    pub fn new(
+        bootstrapped: Arc<dyn IdentityAttributeStorageReader>,
+        storage: Arc<dyn IdentityAttributeStorage>,
+    ) -> Self {
         Self {
             bootstrapped,
             storage,
@@ -33,9 +32,7 @@ impl<B: IdentityAttributeStorageReader, S: IdentityAttributeStorage>
 }
 
 #[async_trait]
-impl<B: IdentityAttributeStorageReader, S: IdentityAttributeStorage> IdentityAttributeStorageReader
-    for BootstrapedIdentityStore<B, S>
-{
+impl IdentityAttributeStorageReader for BootstrapedIdentityStore {
     async fn get_attributes(
         &self,
         identity_id: &IdentityIdentifier,
@@ -60,9 +57,7 @@ impl<B: IdentityAttributeStorageReader, S: IdentityAttributeStorage> IdentityAtt
 }
 
 #[async_trait]
-impl<B: IdentityAttributeStorageReader, S: IdentityAttributeStorage> IdentityAttributeStorageWriter
-    for BootstrapedIdentityStore<B, S>
-{
+impl IdentityAttributeStorageWriter for BootstrapedIdentityStore {
     async fn put_attributes(
         &self,
         sender: &IdentityIdentifier,
@@ -92,9 +87,14 @@ impl<B: IdentityAttributeStorageReader, S: IdentityAttributeStorage> IdentityAtt
     }
 }
 
-impl<B: IdentityAttributeStorageReader, S: IdentityAttributeStorage> IdentityAttributeStorage
-    for BootstrapedIdentityStore<B, S>
-{
+impl IdentityAttributeStorage for BootstrapedIdentityStore {
+    fn as_identity_attribute_storage_reader(&self) -> Arc<dyn IdentityAttributeStorageReader> {
+        Arc::new(self.clone())
+    }
+
+    fn as_identity_attribute_storage_writer(&self) -> Arc<dyn IdentityAttributeStorageWriter> {
+        Arc::new(self.clone())
+    }
 }
 
 #[derive(Clone, Debug)]
--- a/implementations/rust/ockam/ockam_api/src/config/cli.rs
+++ b/implementations/rust/ockam/ockam_api/src/config/cli.rs
@@ -79,7 +79,7 @@ impl AuthoritiesConfig {
     ) -> Result<Vec<PublicIdentity>> {
         let mut v = Vec::new();
         for a in self.authorities.values() {
-            v.push(PublicIdentity::import_arc(a.identity.as_slice(), vault.clone()).await?)
+            v.push(PublicIdentity::import(a.identity.as_slice(), vault.clone()).await?)
         }
         Ok(v)
     }
--- a/implementations/rust/ockam/ockam_api/src/config/cli.rs
+++ b/implementations/rust/ockam/ockam_api/src/config/cli.rs
@@ -2,6 +2,7 @@
 
 use crate::config::{lookup::ConfigLookup, ConfigValues};
 use crate::{cli_state, HexByteVec};
+use ockam_core::compat::sync::Arc;
 use ockam_core::Result;
 use ockam_identity::{IdentityIdentifier, IdentityVault, PublicIdentity};
 use ockam_multiaddr::MultiAddr;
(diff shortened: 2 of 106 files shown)


implementations/rust/ockam/ockam_api/src/config/cli.rs
```diff
@@ -72,13 +73,13 @@ impl AuthoritiesConfig {
...
-        V: IdentityVault,
-    {
+    pub async fn to_public_identities(
+        &self,
+        vault: Arc<dyn IdentityVault>,
+    ) -> Result<Vec<PublicIdentity>> {
         let mut v = Vec::new();
         for a in self.authorities.values() {
-            v.push(PublicIdentity::import(a.identity.as_slice(), vault).await?)
+            v.push(PublicIdentity::import_arc(a.identity.as_slice(), vault.clone()).await?)
```
etorreborre: There are 2 variants of `PublicIdentity::import`. One function `import` can be used with a `Vault` directly, typically in examples (to make them nicer to write). The other function `import_arc` takes an `Arc<dyn IdentityVault>` when only the reference to the trait is available.

implementations/rust/ockam/ockam_api/src/identity/identity_service.rs
```diff
@@ -123,7 +127,7 @@ impl IdentityService {
             },
             Post => match req.path_segments::<2>().as_slice() {
                 [""] => {
-                    let identity = Identity::create(&self.ctx, &self.vault).await?;
+                    let identity = Identity::create_arc(&self.ctx, self.vault.clone()).await?;
```
etorreborre: We do something similar to `PublicIdentity::import` here.

implementations/rust/ockam/ockam_api/src/bootstrapped_identities_store.rs
```diff
@@ -92,9 +87,14 @@ impl<B: IdentityAttributeStorageReader, S: IdentityAttributeStorage> IdentityAtt
     }
 }
 
-impl<B: IdentityAttributeStorageReader, S: IdentityAttributeStorage> IdentityAttributeStorage
-    for BootstrapedIdentityStore<B, S>
-{
+impl IdentityAttributeStorage for BootstrapedIdentityStore {
+    fn as_identity_attribute_storage_reader(&self) -> Arc<dyn IdentityAttributeStorageReader> {
```
etorreborre, line 91: This is necessary due to the lack of [stable trait upcasting in rust](https://doc.rust-lang.org/beta/unstable-book/language-features/trait-upcasting.html) (if `S : T` there is no implicit coercion of `Arc<dyn S>` to `Arc<dyn T>`).

implementations/rust/ockam/ockam_identity/src/channel/common.rs
```diff
@@ -79,3 +79,126 @@ impl CreateResponderChannelMessage {
         }
     }
 }
+
+/// This struct is used to compensate for the lack of non-experimental trait upcasting in Rust
+/// We encapsulate an IdentityVault and delegate the implementation of all the functions of
+/// the various traits inherited by IdentityVault: SymmetricVault, SecretVault, etc...
+struct CoercedIdentityVault {
```
etorreborre, line 86: This is again a work-around for the lack of trait upcasting.

implementations/rust/ockam/ockam_identity/src/channel/common.rs
```diff
@@ -79,3 +79,126 @@ impl CreateResponderChannelMessage {
...
+        self.vault.ec_diffie_hellman(secret, peer_public_key).await
+    }
+
+    async fn compute_key_id_for_public_key(&self, public_key: &PublicKey) -> Result<KeyId> {
+        self.vault.compute_key_id_for_public_key(public_key).await
+    }
+}
+
+/// Return this vault as a symmetric vault
+pub fn to_symmetric_vault(vault: &Arc<dyn IdentityVault>) -> Arc<dyn SymmetricVault> {
```
etorreborre: Unfortunately it is not possible to make those functions act on `self: Arc<dyn IdentityVault>` and write `self.to_symmetric_vault()` directly.

implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
```diff
@@ -244,16 +235,22 @@ impl<V: IdentityVault, K: SecureChannelKeyExchanger, S: AuthenticatedStorage>
...
+            let mut exchanger = state.key_exchanger.write().await;
+            let _ = exchanger.handle_response(payload).await?;
         }
 
         // If we'll need to generate another request
         let mut request_was_sent = false;
 
         // Key exchange hasn't been completed -> generate and send next request
-        if !state.key_exchanger.is_complete().await? {
+        if !state.key_exchanger.read().await.is_complete().await? {
```
etorreborre: The `KeyExchanger` is potentially mutated, so it is implemented as `Arc<RwLock<dyn KeyExchanger>>`

implementations/rust/ockam/ockam_identity/src/channel/decryptor_state.rs
```diff
@@ -1,22 +1,24 @@
...
-use crate::{IdentityIdentifier, IdentityVault};
+use crate::IdentityIdentifier;
+use ockam_core::compat::sync::Arc;
+use ockam_core::KeyExchanger;
+use ockam_node::compat::asynchronous::RwLock;
 
-pub(crate) struct KeyExchange<K: SecureChannelKeyExchanger> {
-    pub key_exchanger: K,
+pub(crate) struct KeyExchange {
+    pub key_exchanger: Arc<RwLock<dyn KeyExchanger + Send + Sync>>,
```
etorreborre: Important note: this requires the `RwLock` implemented in `ockam_node`. This lock is already used in a few places but marked as unsafe. A follow-up PR should either:

 - implement a safe version
 - tweak the `ockam_core::compat::sync::RwLock` no_std implementation to accept `?Sized` values
 - find a different way to encode the state changes of a `Decryptor`

implementations/rust/ockam/ockam_core/src/key_exchanger.rs
```diff
@@ -21,7 +21,7 @@ pub trait KeyExchanger {
     /// Returns true if the key exchange process is complete.
     async fn is_complete(&self) -> Result<bool>;
     /// Return the data and keys needed for channels. Key exchange must be completed prior to calling this function.
-    async fn finalize(self) -> Result<CompletedKeyExchange>;
```
davide-baldo, line 24: Object destruction in `finalize()` was a pretty good semantic, i'm trying to find a way around the reference

davide-baldo: Nope, that's a fundamental rust limitation :(

implementations/rust/ockam/ockam_identity/src/channel/decryptor_state.rs
davide-baldo: You can actually avoid the usage of the lock by switching to a `Box` instead of an `Arc`.
[I've drafted it here](https://github.com/build-trust/ockam/commit/ae719e28d6ce5f20557b0702a5b488b08b356390)

davide-baldo (changes requested):

etorreborre:
@davide-baldo thanks for the refactoring idea, that makes sense to me, I'm going to incorporate that change.

implementations/rust/ockam/ockam_identity/src/channel/decryptor_state.rs
etorreborre: Thanks @davide-baldo I added your commit to the PR.

implementations/rust/ockam/ockam_core/src/key_exchanger.rs
```diff
@@ -11,7 +11,7 @@ use zeroize::Zeroize;
 
 /// A trait implemented by both Initiator and Responder peers.
 #[async_trait]
-pub trait KeyExchanger {
+pub trait KeyExchanger: Send + Sync + 'static {
```
SanjoDeundiak, line 14: Nice

implementations/rust/ockam/ockam_identity/src/identity.rs
```diff
@@ -5,31 +5,32 @@ use crate::change::IdentitySignedChange;
...
 };
 use ockam_core::{Address, Result};
 use ockam_core::{AsyncTryClone, DenyAll};
 use ockam_node::compat::asynchronous::RwLock;
 use ockam_node::Context;
+use ockam_vault::{Hasher, Vault};
 use ockam_vault::{KeyId, SecretAttributes};
 
 /// Identity implementation
 #[derive(AsyncTryClone)]
```
SanjoDeundiak, line 24: I think Identity now can derive Clone instead of AsyncTryClone

SanjoDeundiak: Oh, actually it can't because of the Context...

SanjoDeundiak: It's a little disappointing that we need to wrap `Identity` in an `Arc`. I think we should consider alternatives:
1. Wrap `Context` inside `Identity` in an `Arc`
2. Remove `Context` from `Identity` and derive `Clone` for `Identity` instead of `AsyncTryClone`

implementations/rust/ockam/ockam_identity/src/identity.rs
```diff
@@ -199,9 +207,9 @@ impl<V: IdentityVault, S: AuthenticatedStorage> Identity<V, S> {
     }
 }
 
-impl<V: IdentityVault> Identity<V, InMemoryStorage> {
+impl Identity {
     /// Create an `Identity` with a new secret key and `InMemoryStorage`
-    pub async fn create(ctx: &Context, vault: &V) -> Result<Self> {
+    pub async fn create_arc(ctx: &Context, vault: Arc<dyn IdentityVault>) -> Result<Self> {
```
SanjoDeundiak: Maybe if `Vault::create()` returns `Arc<Vault>` we could avoid having functions with `_arc` suffix? What do you think?

implementations/rust/ockam/ockam_identity/src/channel/common.rs
SanjoDeundiak: Yeah, rust support for dyn casts is very disappointing. One of the reasons why generics are often used instead... I would like to take a look at alternatives, but that should not block this PR from getting merged

SanjoDeundiak (dismissed):
Nice job, left few comments, but PR is in a good shape. Let's merge it as it is and create follow up PRs if needed

implementations/rust/ockam/ockam_identity/src/identity.rs
etorreborre: I will try to tackle this with my next PR

davide-baldo (dismissed):

implementations/rust/ockam/ockam_identity/src/identity.rs
etorreborre: That's a fantastic idea, I was able to remove the 3 `_arc` functions in my latest commit!

davide-baldo (approved):

PR #4376, opened 2023-03-01
`TCP` <-> `SecureChannel` Message Flow Auth

SanjoDeundiak:
Implement Sessions that allow creating boundaries around individual TCP connections and SecureChannels which will make TcpReceiver send messages to this Secure Channel's Decryptor and no other Address

### DIFF
--- a/examples/rust/get_started/examples/11-attribute-based-authentication-edge-plane.rs
+++ b/examples/rust/get_started/examples/11-attribute-based-authentication-edge-plane.rs
@@ -5,10 +5,10 @@ use hello_ockam::{create_token, import_project};
 use ockam::abac::AbacAccessControl;
 use ockam::identity::authenticated_storage::AuthenticatedAttributeStorage;
 use ockam::identity::credential::OneTimeCode;
-use ockam::identity::{Identity, TrustEveryonePolicy, TrustMultiIdentifiersPolicy};
+use ockam::identity::{Identity, SecureChannelTrustOptions, TrustEveryonePolicy, TrustMultiIdentifiersPolicy};
 use ockam::{route, vault::Vault, Context, Result, TcpTransport};
 use ockam_api::authenticator::direct::{CredentialIssuerClient, RpcClient, TokenAcceptorClient};
-use ockam_api::DefaultAddress;
+use ockam_api::{create_tcp_session, DefaultAddress};
 
 /// This node supports an "edge" server which can connect to a "control" node
 /// in order to connect its TCP inlet to the "control" node TCP outlet
@@ -54,16 +54,23 @@ async fn start_node(ctx: Context, project_information_path: &str, token: OneTime
     //    to retrieve the node credential
 
     // Import the authority identity and route from the information file
-    let project = import_project(project_information_path, &vault, &tcp).await?;
+    let project = import_project(project_information_path, &vault).await?;
+
+    let tcp_authority_session = create_tcp_session(&project.authority_route(), &tcp).await.unwrap(); // FIXME: Handle error
+    let authority_trust_options =
+        SecureChannelTrustOptions::new().with_trust_policy(TrustMultiIdentifiersPolicy::new(vec![
+            project.authority_public_identifier()
+        ]));
+    let trust_options = if let Some((sessions, session_id)) = tcp_authority_session.session {
+        authority_trust_options.with_ciphertext_session(&sessions, &session_id)
+    } else {
+        authority_trust_options
+    };
 
     // create a secure channel to the authority
     // when creating the channel we check that the opposite side is indeed presenting the authority identity
     let secure_channel = edge_plane
-        .create_secure_channel_extended(
-            project.authority_route(),
-            TrustMultiIdentifiersPolicy::new(vec![project.authority_public_identifier()]),
-            Duration::from_secs(120),
-        )
+        .create_secure_channel_extended_trust(tcp_authority_session.route, trust_options, Duration::from_secs(120))
         .await?;
 
     let token_acceptor = TokenAcceptorClient::new(
@@ -100,11 +107,20 @@ async fn start_node(ctx: Context, project_information_path: &str, token: OneTime
 
     // 4. create a tcp inlet with the above policy
 
+    let tcp_project_session = create_tcp_session(&project.route(), &tcp).await.unwrap(); // FIXME: Handle error
+    let project_trust_options = SecureChannelTrustOptions::new()
+        .with_trust_policy(TrustMultiIdentifiersPolicy::new(vec![project.identifier()]));
+    let project_trust_options = if let Some((sessions, session_id)) = tcp_project_session.session {
+        project_trust_options.with_ciphertext_session(&sessions, &session_id)
+    } else {
+        project_trust_options
+    };
+
     // 4.1 first created a secure channel to the project
     let secure_channel_address = edge_plane
-        .create_secure_channel_extended(
-            project.route(),
-            TrustMultiIdentifiersPolicy::new(vec![project.identifier()]),
+        .create_secure_channel_extended_trust(
+            tcp_project_session.route,
+            project_trust_options,
             Duration::from_secs(120),
         )
         .await?;
--- a/examples/rust/get_started/examples/11-attribute-based-authentication-edge-plane.rs
+++ b/examples/rust/get_started/examples/11-attribute-based-authentication-edge-plane.rs
@@ -70,7 +70,7 @@ async fn start_node(ctx: Context, project_information_path: &str, token: OneTime
     // create a secure channel to the authority
     // when creating the channel we check that the opposite side is indeed presenting the authority identity
     let secure_channel = edge_plane
-        .create_secure_channel_extended_trust(tcp_authority_session.route, trust_options, Duration::from_secs(120))
+        .create_secure_channel_extended(tcp_authority_session.route, trust_options, Duration::from_secs(120))
         .await?;
 
     let token_acceptor = TokenAcceptorClient::new(
@@ -118,7 +118,7 @@ async fn start_node(ctx: Context, project_information_path: &str, token: OneTime
 
     // 4.1 first created a secure channel to the project
     let secure_channel_address = edge_plane
-        .create_secure_channel_extended_trust(
+        .create_secure_channel_extended(
             tcp_project_session.route,
             project_trust_options,
             Duration::from_secs(120),
--- a/examples/rust/get_started/src/project.rs
+++ b/examples/rust/get_started/src/project.rs
@@ -1,10 +1,8 @@
 use anyhow::anyhow;
 use ockam::identity::{IdentityIdentifier, PublicIdentity};
 use ockam::vault::Vault;
-use ockam::TcpTransport;
-use ockam_api::multiaddr_to_route;
 use ockam_core::errcode::{Kind, Origin};
-use ockam_core::{Error, Result, Route};
+use ockam_core::{Error, Result};
 use ockam_multiaddr::MultiAddr;
 use serde_json::{Map, Value};
 use std::fs::File;
@@ -16,8 +14,8 @@ use std::str::FromStr;
 pub struct Project {
     pub project_identifier: IdentityIdentifier,
     pub authority_public_identity: PublicIdentity,
-    pub authority_route: Route,
-    pub project_route: Route,
+    pub authority_route: MultiAddr,
+    pub project_route: MultiAddr,
 }
 
 /// Accessors for a Project
@@ -38,19 +36,19 @@ impl Project {
     }
 
     /// Return the authority route
-    pub fn authority_route(&self) -> Route {
+    pub fn authority_route(&self) -> MultiAddr {
         self.authority_route.clone()
     }
 
     /// Return the project route
-    pub fn route(&self) -> Route {
+    pub fn route(&self) -> MultiAddr {
         self.project_route.clone()
     }
 }
 
 /// Import a project identity into a Vault from a project.json path
 /// and return a Project struct
-pub async fn import_project(path: &str, vault: &Vault, tcp: &TcpTransport) -> Result<Project> {
+pub async fn import_project(path: &str, vault: &Vault) -> Result<Project> {
     match read_json(path)? {
         Value::Object(values) => {
             let project_identifier = IdentityIdentifier::from_str(get_field_as_str(&values, "identity")?.as_str())?;
@@ -60,15 +58,12 @@ pub async fn import_project(path: &str, vault: &Vault, tcp: &TcpTransport) -> Re
                 PublicIdentity::import(&hex::decode(authority_identity).unwrap(), vault).await?;
 
             let authority_access_route = get_field_as_str(&values, "authority_access_route")?;
-            let authority_route: Route =
-                multiaddr_to_route(&MultiAddr::from_str(authority_access_route.as_str())?, tcp)
-                    .await
-                    .ok_or_else(|| error("incorrect multi address"))?;
+            let authority_route =
+                MultiAddr::from_str(authority_access_route.as_str()).map_err(|_| error("incorrect multi address"))?;
 
             let project_access_route = get_field_as_str(&values, "access_route")?;
-            let project_route: Route = multiaddr_to_route(&MultiAddr::from_str(project_access_route.as_str())?, tcp)
-                .await
-                .ok_or_else(|| error("incorrect multi address"))?;
+            let project_route =
+                MultiAddr::from_str(project_access_route.as_str()).map_err(|_| error("incorrect multi address"))?;
 
             Ok(Project {
                 project_identifier,
--- a/implementations/rust/ockam/ockam_api/src/cloud/mod.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/mod.rs
@@ -3,10 +3,9 @@ use std::str::FromStr;
 use minicbor::{Decode, Encode};
 
 use crate::error::ApiError;
-use ockam::TcpTransport;
 #[cfg(feature = "tag")]
 use ockam_core::TypeTag;
-use ockam_core::{CowStr, Result, Route};
+use ockam_core::{CowStr, Result};
 use ockam_multiaddr::MultiAddr;
 
 pub mod addon;
@@ -52,12 +51,9 @@ impl<'a, T> CloudRequestWrapper<'a, T> {
         }
     }
 
-    pub async fn route(&self, tcp: &TcpTransport) -> Result<Route> {
-        let maddr = MultiAddr::from_str(self.route.as_ref())
-            .map_err(|_err| ApiError::generic(&format!("Invalid route: {}", self.route)))?;
-        crate::multiaddr_to_route(&maddr, tcp)
-            .await
-            .ok_or_else(|| ApiError::generic(&format!("Invalid MultiAddr: {maddr}")))
+    pub fn multiaddr(&self) -> Result<MultiAddr> {
+        MultiAddr::from_str(self.route.as_ref())
+            .map_err(|_err| ApiError::generic(&format!("Invalid route: {}", self.route)))
     }
 }
 
@@ -82,18 +78,18 @@ mod node {
     use std::time::Duration;
 
     use minicbor::Encode;
-    use ockam_vault::Vault;
     use rust_embed::EmbeddedFile;
 
     use ockam_core::api::RequestBuilder;
-    use ockam_core::{self, route, Address, Result, Route};
-    use ockam_identity::{Identity, IdentityIdentifier, TrustIdentifierPolicy};
+    use ockam_core::{self, route, AsyncTryClone, CowStr, Result};
+    use ockam_identity::IdentityIdentifier;
+    use ockam_multiaddr::MultiAddr;
     use ockam_node::api::request_with_timeout;
     use ockam_node::{Context, DEFAULT_TIMEOUT};
 
     use crate::cloud::OCKAM_CONTROLLER_IDENTITY_ID;
     use crate::error::ApiError;
-    use crate::lmdb::LmdbStorage;
+    use crate::nodes::connection::Connection;
     use crate::nodes::{NodeManager, NodeManagerWorker};
     use crate::StaticFiles;
 
@@ -127,25 +123,6 @@ mod node {
         pub(crate) fn controller_identity_id(&self) -> IdentityIdentifier {
             self.controller_identity_id.clone()
         }
-
-        /// Returns a secure channel between the node and the controller.
-        async fn controller_secure_channel(
-            &mut self,
-            route: impl Into<Route>,
-            identity: Identity<Vault, LmdbStorage>,
-        ) -> Result<Address> {
-            let route = route.into();
-            // Create secure channel for the given route using the orchestrator identity.
-            trace!(target: TARGET, %route, "Creating orchestrator secure channel");
-            let addr = identity
-                .create_secure_channel(
-                    route,
-                    TrustIdentifierPolicy::new(self.controller_identity_id()),
-                )
-                .await?;
-            debug!(target: TARGET, %addr, "Orchestrator secure channel created");
-            Ok(addr)
-        }
     }
 
     impl NodeManagerWorker {
@@ -155,10 +132,10 @@ mod node {
             ctx: &Context,
             label: &str,
             schema: impl Into<Option<&str>>,
-            cloud_route: impl Into<Route>,
+            cloud_multiaddr: &MultiAddr,
             api_service: &str,
             req: RequestBuilder<'_, T>,
-            ident: Identity<Vault, LmdbStorage>,
+            ident: Option<CowStr<'_>>,
         ) -> Result<Vec<u8>>
         where
             T: Encode<()>,
@@ -167,7 +144,7 @@ mod node {
                 ctx,
                 label,
                 schema,
-                cloud_route,
+                cloud_multiaddr,
                 api_service,
                 req,
                 ident,
@@ -182,23 +159,25 @@ mod node {
             ctx: &Context,
             label: &str,
             schema: impl Into<Option<&str>>,
-            cloud_route: impl Into<Route>,
+            cloud_multiaddr: &MultiAddr,
             api_service: &str,
             req: RequestBuilder<'_, T>,
-            ident: Identity<Vault, LmdbStorage>,
+            ident: Option<CowStr<'_>>,
             timeout: Duration,
         ) -> Result<Vec<u8>>
         where
             T: Encode<()>,
         {
-            let mut node_manger = self.get().write().await;
-            let cloud_route = cloud_route.into();
-            let sc = node_manger
-                .controller_secure_channel(cloud_route, ident)
-                .await?;
+            let mut node_manager = self.get().write().await;
+            let tcp_transport = node_manager.tcp_transport.async_try_clone().await?;
+            let connection = Connection::new(&tcp_transport, ctx, cloud_multiaddr)
+                .with_identity_name(ident)
+                .with_authorized_identities(node_manager.controller_identity_id())
+                .with_timeout(timeout);
+            let (sc, _) = node_manager.connect(connection).await?;
             let route = route![&sc.to_string(), api_service];
             let res = request_with_timeout(ctx, label, schema, route, req, timeout).await;
-            ctx.stop_worker(sc).await?;
+            ctx.stop_worker(&sc.to_string()).await?;
             res
         }
     }
--- a/implementations/rust/ockam/ockam_api/src/cloud/mod.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/mod.rs
@@ -199,7 +199,7 @@ mod node {
                     trust_options
                 };
                 identity
-                    .create_secure_channel_trust(cloud_session.route, trust_options)
+                    .create_secure_channel(cloud_session.route, trust_options)
                     .await?
             };
             let route = route![&sc.to_string(), api_service];
--- a/implementations/rust/ockam/ockam_api/src/cloud/mod.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/mod.rs
@@ -82,7 +82,7 @@ mod node {
 
     use ockam_core::api::RequestBuilder;
     use ockam_core::{self, route, AsyncTryClone, CowStr, Result};
-    use ockam_identity::IdentityIdentifier;
+    use ockam_identity::{IdentityIdentifier, TrustIdentifierPolicy};
     use ockam_multiaddr::MultiAddr;
     use ockam_node::api::request_with_timeout;
     use ockam_node::{Context, DEFAULT_TIMEOUT};
@@ -91,7 +91,7 @@ mod node {
     use crate::error::ApiError;
     use crate::nodes::connection::Connection;
     use crate::nodes::{NodeManager, NodeManagerWorker};
-    use crate::StaticFiles;
+    use crate::{local_multiaddr_to_route, StaticFiles};
 
     const TARGET: &str = "ockam_api::nodemanager::service";
 
@@ -168,16 +168,41 @@ mod node {
         where
             T: Encode<()>,
         {
-            let mut node_manager = self.get().write().await;
-            let tcp_transport = node_manager.tcp_transport.async_try_clone().await?;
-            let connection = Connection::new(&tcp_transport, ctx, cloud_multiaddr)
-                .with_identity_name(ident)
-                .with_authorized_identities(node_manager.controller_identity_id())
-                .with_timeout(timeout);
-            let (sc, _) = node_manager.connect(connection).await?;
+            let identity = {
+                let node_manager = self.get().read().await;
+                match &ident {
+                    Some(existing_identity_name) => {
+                        let identity_state = node_manager
+                            .cli_state
+                            .identities
+                            .get(existing_identity_name.as_ref())?;
+                        match identity_state.get(ctx, node_manager.vault()?).await {
+                            Ok(idt) => idt,
+                            Err(_) => {
+                                let vault_state = node_manager.cli_state.vaults.default()?;
+                                identity_state.get(ctx, &vault_state.get().await?).await?
+                            }
+                        }
+                    }
+                    None => node_manager.identity()?.async_try_clone().await?,
+                }
+            };
+            let sc = {
+                let node_manager = self.get().read().await;
+                let cloud_route =
+                    crate::multiaddr_to_route(&cloud_multiaddr, &node_manager.tcp_transport)
+                        .await
+                        .ok_or_else(|| ApiError::generic("Invalid Multiaddr"))?;
+                identity
+                    .create_secure_channel(
+                        cloud_route,
+                        TrustIdentifierPolicy::new(node_manager.controller_identity_id()),
+                    )
+                    .await?
+            };
             let route = route![&sc.to_string(), api_service];
             let res = request_with_timeout(ctx, label, schema, route, req, timeout).await;
-            ctx.stop_worker(&sc.to_string()).await?;
+            ctx.stop_worker(sc).await?;
             res
         }
     }
--- a/implementations/rust/ockam/ockam_api/src/cloud/mod.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/mod.rs
@@ -82,18 +82,15 @@ mod node {
 
     use ockam_core::api::RequestBuilder;
     use ockam_core::{self, route, AsyncTryClone, CowStr, Result};
-    use ockam_identity::{IdentityIdentifier, TrustIdentifierPolicy};
+    use ockam_identity::{IdentityIdentifier, SecureChannelTrustOptions, TrustIdentifierPolicy};
     use ockam_multiaddr::MultiAddr;
     use ockam_node::api::request_with_timeout;
     use ockam_node::{Context, DEFAULT_TIMEOUT};
 
     use crate::cloud::OCKAM_CONTROLLER_IDENTITY_ID;
     use crate::error::ApiError;
-    use crate::nodes::connection::Connection;
     use crate::nodes::{NodeManager, NodeManagerWorker};
-    use crate::{local_multiaddr_to_route, StaticFiles};
-
-    const TARGET: &str = "ockam_api::nodemanager::service";
+    use crate::StaticFiles;
 
     impl NodeManager {
         /// Load controller identity id from file.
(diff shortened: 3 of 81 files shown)


SanjoDeundiak:
# Example

## bob.rs
```rs
use ockam::workers::Echoer;
use ockam_core::sessions::Sessions;
use ockam_core::{route, AllowAll, Result};
use ockam_identity::{Identity, SecureChannelListenerTrustOptions};
use ockam_node::Context;
use ockam_transport_tcp::{TcpListenerTrustOptions, TcpTransport};
use ockam_vault::Vault;
use std::time::Duration;

#[ockam::node]
    ctx.start_worker("echoer", Echoer, AllowAll, AllowAll)
        .await?;

    let vault = Vault::create();

    let bob = Identity::create(&ctx, &vault).await?;

    let sessions = Sessions::default();
    let session_id = sessions.generate_session_id();
    let tcp_trust_options = TcpListenerTrustOptions::new().with_session(&sessions, &session_id);
    tcp.listen("127.0.0.1:4000", tcp_trust_options).await?;

    let secure_channel_trust_options_bob =
        SecureChannelListenerTrustOptions::new().with_session(&sessions, &session_id);
    bob.create_secure_channel_listener("bob_listener", secure_channel_trust_options_bob)
        .await?;

    let connection_to_alice = loop {
        if let Some(connection_to_alice) = tcp.registry().get_all_sender_workers().first() {
            break connection_to_alice.clone();
        } else {
            ctx.sleep(Duration::from_millis(100)).await;
        }
    };

    ctx.send(
        route![connection_to_alice.clone(), "echoer"],
        "Hello, Alice!".to_string(),
    )
    .await?;
    let res = ctx.receive_timeout::<String>(1).await;
    assert!(
        res.is_err(),
        "Messages on Alice side only allowed to go through the Secure Channel"
    );

    let channel_to_alice = loop {
        if let Some(channel_to_alice) = bob.secure_channel_registry().get_channel_list().first() {
            break channel_to_alice.encryptor_messaging_address().clone();
        } else {
            ctx.sleep(Duration::from_millis(100)).await;
        }
    };
    ctx.send(
        route![channel_to_alice, "echoer"],
        "Do you hear me, Alice?".to_string(),
    )
    .await?;
    let msg = ctx.receive::<String>().await?.take().body();
    assert_eq!("Do you hear me, Alice?", msg);

    ctx.sleep(Duration::from_secs(5)).await;

    ctx.stop().await
}
```
## alice.rs
```rs
use ockam::workers::Echoer;
use ockam_core::sessions::Sessions;
use ockam_core::{route, AllowAll, Result};
use ockam_identity::{Identity, SecureChannelTrustOptions, TrustEveryonePolicy};
use ockam_node::Context;
use ockam_transport_tcp::{TcpConnectionTrustOptions, TcpTransport};
use ockam_vault::Vault;
use std::time::Duration;

#[ockam::node]
    ctx.start_worker("echoer", Echoer, AllowAll, AllowAll)
        .await?;

    let vault = Vault::create();

    let alice = Identity::create(&ctx, &vault).await?;

    let sessions = Sessions::default();
    let session_id = sessions.generate_session_id();
    let tcp_trust_options = TcpConnectionTrustOptions::new().with_session(&sessions, &session_id);
    let connection_to_bob = tcp
        .connect("127.0.0.1:4000".to_string(), tcp_trust_options)
        .await?;

    ctx.send(
        route![connection_to_bob.clone(), "echoer"],
        "Hello, Bob!".to_string(),
    )
    .await?;
    let res = ctx.receive_timeout::<String>(1).await;
    assert!(
        res.is_err(),
        "Messages on Bob side are only allowed to go through the Secure Channel"
    );

    let secure_channel_trust_options =
        SecureChannelTrustOptions::new().with_ciphertext_session(&sessions, &session_id);
    let channel_to_bob = alice
        .create_secure_channel(
            route![connection_to_bob.clone(), "bob_listener"],
            secure_channel_trust_options,
        )
        .await?;

    ctx.send(
        route![channel_to_bob, "echoer"],
        "Do you hear me, Bob?".to_string(),
    )
    .await?;
    let msg = ctx.receive::<String>().await?.take().body();
    assert_eq!("Do you hear me, Bob?", msg);

    let res = alice
        .create_secure_channel_extended(
            route![connection_to_bob.clone(), "bob_listener"],
            TrustEveryonePolicy,
            Duration::from_secs(1),
        )
        .await;
    assert!(
        res.is_err(),
        "We can only create 1 secure channel with that connection"
    );

    ctx.sleep(Duration::from_secs(5)).await;

    ctx.stop().await
}
```

implementations/rust/ockam/ockam_identity/src/credential_issuer.rs
```diff
@@ -204,14 +198,12 @@ impl CredentialIssuerClient {
 #[ockam_core::async_trait]
 impl CredentialIssuerApi for CredentialIssuerClient {
     async fn public_identity(&self) -> Result<PublicIdentity> {
-        let channel = self
```
SanjoDeundiak, line 207: @etorreborre FYI moved the channel creation to the calling side

implementations/rust/ockam/ockam_api/src/nodes/service/credentials.rs
```diff
@@ -125,8 +132,9 @@ impl NodeManagerWorker {
         let node_manager = self.node_manager.read().await;
         let request: PresentCredentialRequest = dec.decode()?;
 
+        // TODO: Replace with self.connect?
```
SanjoDeundiak, line 135: Not achievable right now because creates a recursion

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -258,6 +261,10 @@ impl<V: IdentityVault, S: AuthenticatedStorage, F: ForwarderCreator>
             } else {
                 trace!("creating new secure channel to {topic_partition_address}");
 
+                // This route should not use Sessions because we are using tunnel over existing
```
SanjoDeundiak, line 264: FYI @davide-baldo

examples/rust/get_started/examples/11-attribute-based-authentication-edge-plane.rs
```diff
@@ -54,16 +54,23 @@ async fn start_node(ctx: Context, project_information_path: &str, token: OneTime
     //    to retrieve the node credential
 
     // Import the authority identity and route from the information file
-    let project = import_project(project_information_path, &vault, &tcp).await?;
+    let project = import_project(project_information_path, &vault).await?;
+
+    let tcp_authority_session = create_tcp_session(&project.authority_route(), &tcp).await.unwrap(); // FIXME: Handle error
```
SanjoDeundiak, line 59: @etorreborre what do you think about these changes?

examples/rust/get_started/src/project.rs
```diff
@@ -16,8 +14,8 @@ use std::str::FromStr;
 pub struct Project {
     pub project_identifier: IdentityIdentifier,
     pub authority_public_identity: PublicIdentity,
-    pub authority_route: Route,
-    pub project_route: Route,
+    pub authority_route: MultiAddr,
```
etorreborre, line 17: 👍

implementations/rust/ockam/ockam_api/src/cloud/mod.rs
```diff
@@ -182,20 +156,52 @@ mod node {
...
         ) -> Result<Vec<u8>>
         where
             T: Encode<()>,
         {
-            let mut node_manger = self.get().write().await;
-            let cloud_route = cloud_route.into();
-            let sc = node_manger
-                .controller_secure_channel(cloud_route, ident)
-                .await?;
+            let identity = {
```
etorreborre, line 168: Thanks for this refactoring :-).

I would actually go one step further and extract this as a method `get_identity_by_name`

implementations/rust/ockam/ockam_api/src/cloud/mod.rs
```diff
@@ -182,20 +156,52 @@ mod node {
...
+                            Err(_) => {
+                                let vault_state = node_manager.cli_state.vaults.default()?;
+                                identity_state.get(ctx, &vault_state.get().await?).await?
+                            }
+                        }
+                    }
+                    None => node_manager.identity()?.async_try_clone().await?,
+                }
+            };
+            let sc = {
```
etorreborre, line 187: We might as well also extract that part 👇 as `create_secure_channel`, so that the logic in `request_controller` is just: 1. retrieve identity 2. create secure channel to cloud 3. send request

implementations/rust/ockam/ockam_api/src/cloud/mod.rs
```diff
@@ -182,20 +156,52 @@ mod node {
             ctx: &Context,
             label: &str,
             schema: impl Into<Option<&str>>,
-            cloud_route: impl Into<Route>,
+            cloud_multiaddr: &MultiAddr,
             api_service: &str,
             req: RequestBuilder<'_, T>,
-            ident: Identity<Vault, LmdbStorage>,
+            ident: Option<CowStr<'_>>,
```
etorreborre, line 162: I wish we had a newtype for an `IdentityName` and not just a naked string

implementations/rust/ockam/ockam_api/src/nodes/connection.rs
```diff
@@ -0,0 +1,46 @@
+use ockam_core::CowStr;
+use ockam_identity::IdentityIdentifier;
+use ockam_multiaddr::MultiAddr;
+use ockam_node::Context;
+use std::time::Duration;
+
+pub struct Connection<'a> {
```
etorreborre, line 7: I know that this struct gets some of its fields with a lifetime parameter but do we really really need them? Can't we just store strings in that struct and get away to with a new lifetime parameter?

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -732,9 +770,9 @@ impl Worker for NodeManagerWorker {
     type Context = Context;
 
     async fn initialize(&mut self, ctx: &mut Context) -> Result<()> {
-        let mut node_manger = self.node_manager.write().await;
-        if !node_manger.skip_defaults {
-            node_manger.initialize_defaults(ctx).await?;
+        let mut node_manager = self.node_manager.write().await;
```
etorreborre, line 773: Thanks for fixing this typo 😊

implementations/rust/ockam/ockam_api/src/nodes/connection.rs
etorreborre, line 7: A bit of documentation would be nice: what does this struct represent? Why do we need it?
Side-question: do we really need to put a context in there? I would keep the `ctx` parameter in the `AuthorityInfo::connect` function instead so that this struct is pure data specifying how a connection should be made

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -396,41 +414,61 @@ impl NodeManager {
         if let Some(pos1) = starts_with_host_tcp(addr) {
             debug!(%addr, "creating a tcp connection");
             let (a1, b1) = addr.split(pos1);
-            let r1 = multiaddr_to_route(&a1, &self.tcp_transport)
-                .await
-                .ok_or_else(|| ApiError::generic("invalid multiaddr"))?;
-
             return match starts_with_secure(&b1) {
                 Some(pos2) => {
+                    let tcp_session = create_tcp_session(&a1, transport)
```
etorreborre, line 419: How can all this code evolve if we want to use a different transport? I'm afraid that there would be many impacts

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
```diff
@@ -51,12 +52,16 @@ impl NodeManagerWorker {
         let res = match (tt, tm) {
             (Tcp, Listen) => node_manager
                 .tcp_transport
-                .listen(&addr)
+                // We don't use Sessions for listeners and connections created manually
+                // TODO: Add that functionality
```
etorreborre, line 56: Could you please expand a bit on that `TODO`? This is certainly more than just creating a `Sessions` struct here. We then need to pass it around, but where?

implementations/rust/ockam/ockam_core/src/access_control.rs
```diff
@@ -28,6 +28,7 @@ use core::fmt::Debug;
 #[async_trait]
 #[allow(clippy::wrong_self_convention)]
 pub trait IncomingAccessControl: Debug + Send + Sync + 'static {
+    // TODO: Consider &mut self
```
etorreborre, line 31: In which situation would we need to change the access control based on the current message? 
[edit] I see that we are mutating state in the `SessionOutgoingAccessControl`

implementations/rust/ockam/ockam_core/src/sessions.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub fn set_listener_session_id(&self, address: &Address, session_id: &SessionId) {
+        let mut lock = self.internal.write().unwrap();
+        let address_info = lock.get_info_mut(address);
+
+        address_info.listener_session_id = Some(session_id.clone());
+    }
+
+    /// Get listener [`SessionId`] for given [`Address`]
+    pub fn get_listener_session_id(&self, address: &Address) -> Option<SessionId> {
+        match self.internal.write().unwrap().get_info(address) {
```
etorreborre: s/write/read

implementations/rust/ockam/ockam_core/src/sessions.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub fn set_session_id(&self, address: &Address, session_id: &SessionId) {
+        let mut lock = self.internal.write().unwrap();
+        let address_info = lock.get_info_mut(address);
+
+        address_info.session_id = Some(session_id.clone());
+    }
+
+    /// Get [`SessionId`] for given [`Address`]
+    pub fn get_session_id(&self, address: &Address) -> Option<SessionId> {
+        match self.internal.write().unwrap().get_info(address) {
```
etorreborre: s/write/read

implementations/rust/ockam/ockam_core/src/sessions.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub fn get_listener_session_id(&self, address: &Address) -> Option<SessionId> {
+        match self.internal.write().unwrap().get_info(address) {
+            Some(address_info) => address_info.listener_session_id.clone(),
+            None => None,
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+struct AddressInfo {
```
etorreborre, line 68: It would be good to have a comment explaining what that means:

 - can both options be `None`?
 - what does it mean when one in `Some` and the other is `None`?
 - how is a `session_id` used?
- how is a `listener_session_id` used?
  
Maybe developing an example can help

implementations/rust/ockam/ockam_core/src/sessions.rs
```diff
@@ -0,0 +1,109 @@
...
+            address,
+            session_id,
+            listener_session_id,
+        }
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+struct SessionsInternal {
+    data: Vec<AddressInfo>,
```
etorreborre, line 90: Why not a `HashMap` by `Address` since that seems to be the most frequent access use case?

implementations/rust/ockam/ockam_transport_tcp/src/trust_options.rs
```diff
@@ -0,0 +1,103 @@
...
+        Self { session: None }
+    }
+
+    /// Set session for this connection, in this case messages from that connection
+    /// will be only allowed to go to the [`Address`]es with the same [`SessionId`].
+    /// Information of [`Address`]' [`SessionId`] is stored in [`Sessions`] struct.
+    ///
+    /// Also this [`SessionId`] will be added to [`LocalInfo`] of the messages from that
+    /// connection
+    pub fn with_session(mut self, sessions: &Sessions, session_id: &SessionId) -> Self {
```
etorreborre, line 29: I think we should remove `Sessions` out of the trust options. One `TcpConnectionTrustOptions` struct represents the options for one TCP connection and should not point to some global data like `Sessions`. 

If we leave `Sessions` out this means that:

 - we need to explicitly pass it to a function like `tcp.listen`
 - or make it implicit by attaching it to the identity context (for example by having an optional `Sessions` value set on the identity `Context`)

IMO if we leave it outside of `TcpConnectionTrustOptions` this will make the migration to a `Router` support for sessions easier.

implementations/rust/ockam/ockam_api/src/util.rs
```diff
@@ -74,8 +76,110 @@ pub fn multiaddr_to_socket_addr(ma: &MultiAddr) -> Option<String> {
...
+    pub route: Route,
+}
+
+pub async fn create_tcp_session(ma: &MultiAddr, tcp: &TcpTransport) -> Option<TcpSession> {
+    let mut rb = Route::new();
+    let mut it = ma.iter().peekable();
+    let sessions = Sessions::default();
+    let session_id = sessions.generate_session_id();
+
+    let mut trust_options =
```
etorreborre, line 90: We start with `Some` but I don't get when we can end up with `None`, didn't you miss a place where `trust_options = None`? Typically that's a place where a unit test should test this behaviour.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -249,7 +249,11 @@ async fn run_foreground_node(
 
     let tcp = TcpTransport::create(&ctx).await?;
     let bind = cmd.tcp_listener_address;
-    let (socket_addr, listener_addr) = tcp.listen(&bind).await?;
+
+    // This listener gives exclusive access to our node, make sure this is intended
```
etorreborre, line 253: Maybe we should create a shortcut issue right away under the security epic to address this

implementations/rust/ockam/ockam_core/src/sessions/access_control.rs
```diff
@@ -0,0 +1,112 @@
...
+            self.session_id,
+            self.listener_session_id,
+            self.allow_only_1_message_to_the_listener,
+            self.sessions,
+        )
+    }
+
+    /// Allow one message to the listener with given listener session id (this will
+    /// make listener to spawn a new session)
+    pub fn allow_one_message_to_the_listener(mut self, listener_session_id: SessionId) -> Self {
```
etorreborre, line 38: Let's keep the naming consistent and then use `allow_one_message_to_the_listener` in the struct

implementations/rust/ockam/ockam_core/src/sessions/access_control.rs
```diff
@@ -0,0 +1,112 @@
...
+        self.allow_only_1_message_to_the_listener = true;
+
+        self
+    }
+}
+
+/// Session Outgoing Access Control
+///
+/// Allows to send messages only to members of the given [`SessionId`] or message listener
+/// with given [`SessionId`]. Optionally, only 1 message can be passed to the listener.
```
etorreborre, line 49: Can we explain when passing just one message is the desirable behaviour? In particular we should be specific about what kind of listener we are talking about. In this case, this is a secure channel listener if I understand correctly

implementations/rust/ockam/ockam_core/src/sessions/access_control.rs
```diff
@@ -0,0 +1,112 @@
...
+    sessions: Sessions,
+}
+
+impl SessionOutgoingAccessControlBuilder {
+    /// Constructor
+    pub fn new(session_id: SessionId, sessions: Sessions) -> Self {
+        Self {
+            session_id,
+            listener_session_id: None,
+            allow_only_1_message_to_the_listener: true,
```
etorreborre: Shouldn't that be initialized to `false` since the only other builder function sets it to `true`?

implementations/rust/ockam/ockam_core/src/sessions/access_control.rs
```diff
@@ -0,0 +1,112 @@
...
+
+#[async_trait]
+impl OutgoingAccessControl for SessionOutgoingAccessControl {
+    async fn is_authorized(&self, relay_msg: &RelayMessage) -> Result<bool> {
+        let onward_route = relay_msg.onward_route();
+
+        let next = onward_route.next()?;
+
+        // Allow messages to workers with the same session id
+        if let Some(address_session_id) = self.sessions.get_session_id(next) {
```
etorreborre, line 85: Alternatively this could be
```
if self.sessions.get_session_id(next). is_some_and(|s| s == self.session_id) {
...
}
```

implementations/rust/ockam/ockam_core/src/sessions/local_info.rs
```diff
@@ -0,0 +1,74 @@
...
+use crate::sessions::SessionId;
+use crate::{Decodable, Encodable, Error, LocalInfo, LocalMessage, Result};
+use serde::{Deserialize, Serialize};
+
+/// SessionId LocalInfo unique Identifier
+pub const SESSION_ID_IDENTIFIER: &str = "SESSION_ID_IDENTIFIER";
+
+/// Session LocalInfo used for LocalMessage
+#[derive(Serialize, Deserialize)]
+pub struct SessionIdLocalInfo {
```
etorreborre, line 11: I wonder if we try to regroup `SessionId` and `SessionIdLocalInfo` into an opaque type `Session` (or keep `SessionId`) with the an adequate `Eq` instance + serialization / deserialization to/from `LocalInfo`. The only thing we are interested in are:

 - creating random values of that type
 - being able to compare them for equality
 - serialize / deserialize them

Do we need a new struct just to add the serialization?

implementations/rust/ockam/ockam_identity/src/channel/listener.rs
```diff
@@ -30,8 +52,16 @@ impl<V: IdentityVault, S: AuthenticatedStorage> Worker for IdentityChannelListen
         ctx: &mut Self::Context,
         msg: Routed<Self::Message>,
     ) -> Result<()> {
-        let trust_policy = Arc::clone(&self.trust_policy);
         let identity = self.identity.async_try_clone().await?;
-        DecryptorWorker::create_responder(ctx, identity, trust_policy, msg).await
+
+        // Check if there is a session that connection worker added to LocalInfo
+        // If yet - decryptor will be added to that session to be able to receive further messages
```
etorreborre: s/If yet/If not yet

implementations/rust/ockam/ockam_identity/src/channel/trust_options.rs
```diff
@@ -0,0 +1,112 @@
+use crate::{TrustEveryonePolicy, TrustPolicy};
+use ockam_core::compat::sync::Arc;
+use ockam_core::sessions::{SessionId, Sessions};
+
+/// Trust options for a Secure Channel
+pub struct SecureChannelTrustOptions {
```
etorreborre, line 6: Let's remove `_plaintext_session` if it is not used yet

implementations/rust/ockam/ockam_transport_tcp/src/registry.rs
```diff
@@ -80,6 +80,13 @@ impl TcpRegistry {
     }
 }
 
+impl TcpRegistry {
```
etorreborre, line 83: Why not merge this section with the previous `impl TcpRegistry` one? Because of the `pub` annotation?

implementations/rust/ockam/ockam_transport_tcp/src/trust_options.rs
```diff
@@ -0,0 +1,103 @@
+use ockam_core::compat::sync::Arc;
+use ockam_core::sessions::{SessionId, SessionOutgoingAccessControlBuilder, Sessions};
+use ockam_core::{IncomingAccessControl, LocalOnwardOnly, LocalSourceOnly, OutgoingAccessControl};
+
+pub(crate) struct TcpConnectionAccessControl {
+    pub fresh_session_id: Option<SessionId>,
```
etorreborre: The name is a bit weird on the client side when accessing this field. Because the session id is "fresh" when you generate it, not anymore when you use it. I would simply put `session_id`

implementations/rust/ockam/ockam_transport_tcp/src/workers/addresses.rs
```diff
@@ -0,0 +1,38 @@
+use crate::workers::ConnectionRole;
+use ockam_core::Address;
+
+#[derive(Clone, Debug)]
+pub(crate) struct Addresses {
```
etorreborre, line 5: 👍 nice, this helps to keep the consistency of addresses

implementations/rust/ockam/ockam_identity/tests/message_flow_auth.rs
```diff
@@ -0,0 +1,316 @@
...
+        .await?;
+
+    let info = SecureChannelInfo { identity, address };
+
+    Ok(info)
+}
+
+#[allow(non_snake_case)]
+#[ockam_macros::test]
+async fn sessions__secure_channel_over_tcp_without_session__should_pass_messages(
```
etorreborre: That's a stylistic thing but personally I'd rather read about the tests first in a test file, then have a look at the implementation, instead of having to read a bunch of implementation-related code until I get to the context where it is used

implementations/rust/ockam/ockam_identity/tests/message_flow_auth.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    Ok(info)
+}
+
+struct TcpConnectionInfo {
+    address: Address,
+    session: Option<(Sessions, SessionId)>,
+}
+
+async fn create_connection(
```
etorreborre: `create_tcp_connection` maybe?

implementations/rust/ockam/ockam_identity/tests/message_flow_auth.rs
```diff
@@ -0,0 +1,316 @@
...
+use ockam_identity::authenticated_storage::mem::InMemoryStorage;
+use ockam_identity::{
+    Identity, SecureChannelListenerTrustOptions, SecureChannelTrustOptions, TrustEveryonePolicy,
+};
+use ockam_node::Context;
+use ockam_transport_tcp::{TcpConnectionTrustOptions, TcpListenerTrustOptions, TcpTransport};
+use ockam_vault::Vault;
+use rand::random;
+
+async fn check_message_flow(ctx: &Context, route: Route, should_pass: bool) -> Result<()> {
```
etorreborre, line 166: Since we are not using the full route in the test we might as well take just an `Address`. And maybe provide 2 specialized assertions:

 - `message_should_pass`
 - `message_should_not_pass`

to make the tests even more readable

implementations/rust/ockam/ockam_identity/tests/message_flow_auth.rs
```diff
@@ -0,0 +1,316 @@
...
+        self.tcp
+            .registry()
+            .get_all_sender_workers()
+            .first()
+            .unwrap()
+            .clone()
+    }
+}
+
+async fn create_tcp_listener(ctx: &Context, with_session: bool) -> Result<TcpListenerInfo> {
```
etorreborre, line 212: Here I would also provide 2 functions based on that one:

 - `create_tcp_listener_with_session`
 - `create_tcp_listener_without_session`

Alternatively use
```
enum WithOrWithoutSession {
  WithSession,
  WithoutSession,
}
```

to make the tests more readable.

implementations/rust/ockam/ockam_identity/tests/message_flow_auth.rs
```diff
@@ -0,0 +1,316 @@
...
+    check_message_flow(ctx, route![connection_to_alice.clone()], true).await?;
+
+    let channel_to_bob = create_secure_channel(ctx, &connection_to_bob).await?;
+    ctx.sleep(Duration::from_millis(50)).await; // Wait for workers to add themselves to the registry
+    let channel_to_alice = bob_listener_info.get_channel();
+
+    check_message_flow(ctx, route![channel_to_bob.address.clone()], true).await?;
+    check_message_flow(ctx, route![channel_to_alice.clone()], true).await?;
+
+    ctx.stop().await
```
etorreborre: Aside: I don't understand why the `ctx` is not always stopped by the test macro, and not only when there is a panic or an error. In 99% of the cases that's what we want, isn't it?

implementations/rust/ockam/ockam_identity/tests/message_flow_auth.rs
```diff
@@ -0,0 +1,316 @@
...
+}
+
+#[allow(non_snake_case)]
+#[ockam_macros::test]
+async fn sessions__secure_channel_over_tcp_with_alice_session__should_not_pass_messages(
```
etorreborre: The naming of the tests does not make it very obvious what we are testing and if we are testing all the possible combinations. Maybe you can try to refactor the code so that there is function taking: 

 - 4 input parameters
    - `tcp_listener_creation`: `WithSession/WithoutSession`
    - `tcp_connection_creation`: `WithSession/WithoutSession`
    - `secure_channel_listener_creation`: `WithSession/WithoutSession`
    - `secure_channel_creation`: `WithSession/WithoutSession`

 - 4 expected results
    - `outgoing_message_from_bob_tcp_receiver`: `Received/NotReceived`
    - `outgoing_message_from_bob_tcp_sender`: `Received/NotReceived` 
    - `outgoing_message_from_bob_encryptor`: `Received/NotReceived`
    - `outgoing_message_from_bob_decryptor`: `Received/NotReceived` 

What I wrote above is probably incorrect and makes me think that we should make a first separation between the listening and the connecting side and test:

 1. the topology of a TcpListener + SecureChannel listener (bob), with and without a session
       - can a message be sent from the `TcpRecvProcessor` when a connection is created?
       - can a message be sent from the `TcpSender` when a connection is created?
       - can a message be sent from the `Decryptor` (via the the `TcpRcvProcessor`) when a secure channel is created?
       - can a message be sent from the `Encryptor` (via the the `TcpSender`) when a secure channel is created?

 2. the topology of a TcpConnection + SecureChannel connection (alice), with and without a session
       -  similar tests here

etorreborre (dismissed):
I think this does the job of securing the transport <-> secure channel layer but at the price of introducing lots of additional machinery. I hope we can find a way to simplify this in the future :-)

implementations/rust/ockam/ockam_api/src/nodes/service.rs
SanjoDeundiak: I think this part will be refactored in the future. However, I think it's fine that we rely on TCP here. The idea is that for different transport types API for creating connections/initialization could be different, but after that they should have identical messaging API.

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
SanjoDeundiak: I'm not sure in which way we want to support this. Inlet/Outlet use-case has more or less defined topology which we want to secure in the nearest future. I think we'll know more about how to secure arbitrary topology while doing that

implementations/rust/ockam/ockam_core/src/sessions.rs
SanjoDeundiak: I think Vec is better when number of elements is relatively small

implementations/rust/ockam/ockam_transport_tcp/src/trust_options.rs
SanjoDeundiak: I didn't get your suggestion. The idea is to have globally shared `Sessions` storage, so you only need its id to access a specific session. At this point global state is inside a shared `Sessions`, later, that could be embedded in `ockam_node` so that everyone has access to it using `Context`.

implementations/rust/ockam/ockam_api/src/util.rs
SanjoDeundiak: `Option::take()` sets it to `None`

implementations/rust/ockam/ockam_command/src/node/create.rs
SanjoDeundiak: The big shortcut issue covers it. Not sure what would be a suitable way to divide that into smaller issues

implementations/rust/ockam/ockam_api/src/nodes/connection.rs
SanjoDeundiak: @adrianbenavides

implementations/rust/ockam/ockam_api/src/nodes/service.rs
SanjoDeundiak: @adrianbenavides

implementations/rust/ockam/ockam_api/src/nodes/connection.rs
SanjoDeundiak: @adrianbenavides

implementations/rust/ockam/ockam_api/src/cloud/mod.rs
SanjoDeundiak: @adrianbenavides

implementations/rust/ockam/ockam_api/src/cloud/mod.rs
SanjoDeundiak: @adrianbenavides

implementations/rust/ockam/ockam_api/src/cloud/mod.rs
SanjoDeundiak: @adrianbenavides

implementations/rust/ockam/ockam_core/src/sessions/access_control.rs
SanjoDeundiak: I think the listener pattern is abstract and currently used for secure channels and outlets.

implementations/rust/ockam/ockam_core/src/sessions/access_control.rs
SanjoDeundiak: Value of this flag doesn't matter if `listener_session_id` is `None`

implementations/rust/ockam/ockam_core/src/sessions/access_control.rs
SanjoDeundiak: It's not in stable yet

implementations/rust/ockam/ockam_core/src/sessions/local_info.rs
SanjoDeundiak: I think concepts of Session and SessionId should be completely decoupled from the idea of how that information can travel in a node.

implementations/rust/ockam/ockam_identity/src/channel/listener.rs
SanjoDeundiak: That should be `If yes` actually :D

implementations/rust/ockam/ockam_identity/src/channel/trust_options.rs
SanjoDeundiak: I'm going to use it very soon, so would like to indicate what is the future for this API

implementations/rust/ockam/ockam_transport_tcp/src/registry.rs
SanjoDeundiak: I just like to keep multiple sections for things that I think are logically separate. In this case I think it's easier to read the code when you clearly see what is the public API of the struct

implementations/rust/ockam/ockam_identity/tests/message_flow_auth.rs
SanjoDeundiak: My C/C++ mind sometimes protests to declare functions after their usage. But I agree

implementations/rust/ockam/ockam_identity/tests/message_flow_auth.rs
SanjoDeundiak: I remember there was a reason, but I don't remember what it was

SanjoDeundiak: Maybe you could try it, if you want

implementations/rust/ockam/ockam_identity/tests/message_flow_auth.rs
SanjoDeundiak: I don't think it's worth refactoring it even more right now

implementations/rust/ockam/ockam_core/src/sessions.rs
SanjoDeundiak: I'll get back to it in next iterations

implementations/rust/ockam/ockam_api/src/cloud/mod.rs
SanjoDeundiak: @adrianbenavides

implementations/rust/ockam/ockam_transport_tcp/src/trust_options.rs
etorreborre: My suggestion is to _not_ put `Sessions` inside the trust options which concern only one connection/listener and pass it later where it is needed. We will have to do that eventually anyway as you write when we access it from a `Context`. I would suggest to take the hit now, but maybe we'd rather release ASAP.

implementations/rust/ockam/ockam_core/src/sessions/access_control.rs
etorreborre: Damned

implementations/rust/ockam/ockam_api/src/util.rs
etorreborre: Haha! Thanks, I'm not used to this kind of API for `Option` and I missed it

implementations/rust/ockam/ockam_command/src/node/create.rs
etorreborre: Ok, I don't know which issue exactly you are referring to but as long as there is an owner it's all good :-)

implementations/rust/ockam/ockam_core/src/sessions/access_control.rs
etorreborre: I see, thanks

implementations/rust/ockam/ockam_core/src/sessions/access_control.rs
etorreborre: Alright

implementations/rust/ockam/ockam_core/src/sessions/local_info.rs
etorreborre: Ok, I don't see exactly how much we gain from this, but ok

implementations/rust/ockam/ockam_identity/src/channel/trust_options.rs
etorreborre: I would still introduce it when needed, but no big deal

implementations/rust/ockam/ockam_transport_tcp/src/registry.rs
etorreborre: Ok. As a newcomer I find it a bit hard sometimes to know where is *all* the code related to a given struct so my experience is a bit different

implementations/rust/ockam/ockam_transport_tcp/src/trust_options.rs
SanjoDeundiak: But it's already meant to be shared (and cloning actually gives you an object with access to the same storage)

implementations/rust/ockam/ockam_identity/tests/message_flow_auth.rs
etorreborre: Ok, but I'm not sure we cover all the cases we should be covering there. Right now I don't see how the coverage is systematic and exhaustive. Maybe you can cover this in a follow-up PR, or convince me that we're good :-)

implementations/rust/ockam/ockam_identity/tests/message_flow_auth.rs
etorreborre: Ok, I will, at some stage

etorreborre (approved):
I resolved all the conversations. I think I am still a bit confused by the naming and coverage of the message flow auth tests but let's move on.

Issue #3507, opened 2022-09-21
UDP NAT Puncturing


mrinalwadhwa:
# Background

https://www.youtube.com/watch?v=01ajHxPLxAw
https://en.wikipedia.org/wiki/UDP_hole_punching

A naive implementation of UDP NAT puncturing looks like this

```mermaid
sequenceDiagram
    participant A
    participant S as Rendezvous Service (S)
    participant B

    autonumber
    A ->> S: send a udp message
    B ->> S: send a udp message
    S ->> A: internet visible udp address of B
    S ->> B: internet visible udp address of A
    B ->> A: B can send messages to A
    A ->> B: A can send messages to B
```

Wikipedia has a good description of this flow:

```
Let A and B be the two hosts, each in its own private network; NA and NB are the two NAT devices with globally reachable IP addresses EIPA and EIPB respectively; S is a public server with a well-known, globally reachable IP address.

- A and B each begin a UDP conversation with S; the NAT devices NA and NB create UDP translation states and assign temporary external port numbers EPA and EPB.
- S examines the UDP packets to get the source port used by NA and NB (the external NAT ports EPA and EPB).
- S passes EIPA:EPA to B and EIPB:EPB to A.
- A sends a packet to EIPB:EPB.
- NA examines A's packet and creates the following in its translation table: (Source-IP-A, EPA, EIPB, EPB).
- B sends a packet to EIPA:EPA.
- NB examines B's packet and creates the following tuple in its translation table: (Source-IP-B, EPB, EIPA, EPA).
- Depending on the state of NA's translation table when B's first packet arrives (i.e. whether the tuple (Source-IP-A, EPA, EIPB, EPB) has been created by the time of arrival of B's first packet), B's first packet is dropped (no entry in translation table) or passed (entry in translation table has been made).
- Depending on the state of NB's translation table when A's first packet arrives (i.e. whether the tuple (Source-IP-B, EPB, EIPA, EPA) has been created by the time of arrival of A's first packet), A's first packet is dropped (no entry in translation table) or passed (entry in translation table has been made).
- At worst, the second packet from A reaches B; at worst the second packet from B reaches A. Holes have been "punched" in the NAT and both hosts can directly communicate.
```

Let's start with a prototype of this basic protocol.

# Integrating with Ockam

This basic prototype will help us explore how this should integrate with Ockam.

[...]

PR #4381, opened 2023-03-02
feat(rust): use one socket for all client udp messages

neil2468:
## Current Behavior

The UDP transport currently...
- Uses a new local socket for each UDP message sent to a new destination address.
- Can put the sending part of a socket into an error state, for example, by trying to send to an IPv6 address. Once in the error state, the transport is not able to send more messages. 

## Proposed Changes

Related to #3507. For UDP NAT puncturing it is important that UDP messages sent to an external Rendezvous Service and to an external peer 'B' are all sent from the same UDP address & port. 

Changes to UDP transport...
- Try to use a single local socket for all sent UDP messages which were initiated by an entity within the local node (i.e. so called 'client' messages).
- Enable the senders and listeners to inform the router when they experience an error. On knowing an error has occurred, the router tries to stop and replace the errored sender and listener pair.
- Some refeactoring to reduce code.
- Add tests to exercise the new features.

Apologies for submitting this all in one large chunk. But I couldn't quite work out how not to.

## Checks

Thanks!


davide-baldo:
Hi! :wave: 
Thanks for this contribution!

neil2468:
Hi 👋 
I just updated the commit message and rebased with develop.
Thanks!

implementations/rust/ockam/ockam_transport_udp/src/router/handle.rs
```diff
@@ -25,99 +11,38 @@ pub(crate) struct UdpRouterHandle {
...
-            hostnames
-                .iter()
-                .filter_map(|s| UdpAddress::from_str(s).ok())
-                .map(|addr| addr.into()),
-        );
-
-        // TODO: should we send a router request instead
-        // and see if worker is already registered?
-        let response = self
+    /// Request reouter start listening on a local UDP port
```
davide-baldo: typo

implementations/rust/ockam/ockam_transport_udp/src/router/handle.rs
```diff
@@ -25,99 +11,38 @@ pub(crate) struct UdpRouterHandle {
...
+    /// so the local node can act as a server to other nodes
+    pub async fn listen(&self, local_addr: SocketAddr) -> Result<()> {
```
davide-baldo, line 34: nice! :+1:  consistent with TCP side

implementations/rust/ockam/ockam_transport_udp/src/workers/listener.rs
```diff
@@ -61,32 +60,41 @@ impl Processor for UdpListenProcessor {
...
                 Ok((msg, addr)) => (msg, addr),
-                Err(_e) => {
-                    info!("Failed to read message from UDP socket.");
+                Err(e) => {
+                    warn!("Failed to read message from UDP socket: {:?}", e);
+
+                    // Inform router of error
+                    let _ = self
+                        .router_handle
+                        .sender_listener_error(self.local_addr)
```
davide-baldo: I'm not sure re-building the UDP socket when it breaks it's the way forward:
> Can put the sending part of a socket into an error state, for example, by trying to send to an IPv6 address.

Can't we avoid breaking the socket by allowing only ipv4? Are there other cases?

implementations/rust/ockam/ockam_transport_udp/src/workers/listener.rs
```diff
@@ -61,32 +60,41 @@ impl Processor for UdpListenProcessor {
...
-        self.router_handle
-            .register(self.tx_addr.clone(), addr.to_string())
-            .await?;
-
-        msg.return_route.modify().prepend(UdpAddress::from(addr));
-
-        debug!("Message onward route: {}", msg.onward_route);
-        debug!("Message return route: {}", msg.return_route);
+        // Set return route to go directly to paired sender, skipping the UDP router
+        msg.return_route
```
davide-baldo: You can use the `route!` macro to make it a bit more readable:
```rust
msg.return_route = route![
  self.sender_addr.clone(),
  Address::new(UDP, addr.to_string()),
  msg.return_route
];
```

implementations/rust/ockam/ockam_transport_udp/src/workers/sender.rs
```diff
@@ -38,24 +43,54 @@ impl Worker for UdpSendWorker {
...
+                trace!("Successful send to {}", addr);
+                Ok(())
+            }
+            Err(e) => {
+                warn!("Failed send to {}: {:?}", addr, e);
+
+                // Inform router of our error
+                let _ = self
+                    .router_handle
+                    .sender_listener_error(self.local_addr)
```
davide-baldo: Same as before, can we avoid breaking it?

implementations/rust/ockam/ockam_transport_udp/tests/tests.rs
```diff
@@ -0,0 +1,250 @@
...
+use ockam_core::{route, AllowAll, Result, Routed, Worker};
+use ockam_node::Context;
+use ockam_transport_udp::{UdpTransport, UDP};
+use std::net::SocketAddr;
+use std::time::Duration;
+use tokio::net::UdpSocket;
+use tracing::{debug, error, trace};
+
+const TIMEOUT: Duration = Duration::from_secs(5);
+const AVAILABLE_LOCAL_PORTS_ADDR: &str = "0.0.0.0:0";
```
davide-baldo: Can you bind to 127.0.0.1 rather than 0.0.0.0?

davide-baldo (changes requested):
Nice step up! :+1:

davide-baldo:
@neil2468 
I wanted to let you know we are planning to remove auto-connect, and protocol from `Address` to avoid security pitfalls.
So, the router will change heavily yet again, and you'll need to explicitly create outgoing workers as you need to declare listeners.
Unfortunately, I can't give you any fixed time frame yet.

mrinalwadhwa:
> @neil2468
I wanted to let you know we are planning to remove auto-connect, and protocol from Address to avoid security pitfalls.
So, the router will change heavily yet again, and you'll need to explicitly create outgoing workers as you need to declare listeners.
Unfortunately, I can't give you any fixed time frame yet.

@neil2468 I think we can move these two efforts in parallel. It would be best for you to keep building on the current code and for us to start landing the changes. We can refactor for the design changes @davide-baldo mentioned above as we go.

implementations/rust/ockam/ockam_transport_udp/src/router/handle.rs
neil2468: That was lucky 😄 I hadn't noticed.

implementations/rust/ockam/ockam_transport_udp/src/router/handle.rs
neil2468: Done, thanks.

implementations/rust/ockam/ockam_transport_udp/src/workers/listener.rs
neil2468: Done.

implementations/rust/ockam/ockam_transport_udp/tests/tests.rs
neil2468: Okay.

implementations/rust/ockam/ockam_transport_udp/src/workers/listener.rs
neil2468: I've seen the sink part of a socket get in to an error state when (1) sending to an IPv6 address and (2) when sending to "0.0.0.0:0". The transport does not support IPv6, so (1) can no longer happen, but something like (2) or something I've never seen could happen.

My gut feeling is that errors will be rare as UDP is happy to send to peers that have no route or are disconnected. However, I don't know what errors could happen on all platforms and now, as the transport uses a single UDP socket for 'client' messages to all peers, if an error occurs the transport will grind to a halt for 'client' messages unless the transport tries to clear/recover from the error.

I'm happy to implement whatever Ockam prefer. 👍

implementations/rust/ockam/ockam_transport_udp/src/workers/sender.rs
neil2468: See related comments on `listener.rs`.

implementations/rust/ockam/ockam_transport_udp/src/workers/listener.rs
davide-baldo: > as the transport uses a single UDP socket for 'client' messages to all peers, if an error occurs the transport will grind to a halt for 'client' messages unless the transport tries to clear/recover from the error.

I 100% agree, we cannot afford to have a broken socket when the whole node depends on it.
Still, what worries me most, is the scenario where one worker keeps re-trying sending a message and keeps breaking the socket. In this case, the service seems to be working fine and will probably go undetected, but the packet drop is higher than it should be.

> However, I don't know what errors could happen on all platforms

How about fixing errors one step at the time? I believe a broken socket doesn't go unnoticed, and will help identify and fix every failure case in every platform.

If you could please reject and log messages sent toward `0.0.0.0`, we could omit the recover mechanism.

neil2468: Okay, cool 👍 .

I've added the below check in `UdpSendWorker` just before sending, and have removed the recover mechanism.

```
        // Error on conditions that _might_ put the sink
        // into an error state
        if addr.port() == 0 {
            warn!(peer_addr = %peer_addr, "Will not send to address");
            return Err(TransportError::InvalidAddress.into());
        }
```

From testing, it seems sending to a port 0 causes the sink to get into an error state. For example, sending to `0.0.0.0:8080` is okay but `0.0.0.0:0` and `192.168.1.10:0` are not.

davide-baldo (approved):
Nice! You even anticipated the auto-connect removal and simplified the architecture by an order of magnitude, hat off! :+1:

PR #4346, opened 2023-02-27
tests(rust): allow using a custom ockam home on bats tests

adrianbenavides:
Fixes https://github.com/build-trust/ockam/pull/4257#issuecomment-1444276091

PR #4387, opened 2023-03-02
Update rust version to 1.68.0

metaclips:
This PR 
- Updates rust version to 1.68.0
- Fixes clippy issues
- Pins rust nightly version to March 15 2023 as nightly latest is broken
- Installs parallel package so that we can run bats tests concurrently


adrianbenavides:
I found a bug in the setup. It is solved at https://github.com/build-trust/ockam/pull/4346

We can try running the orchestrator tests on CI after rebasing those changes.

SanjoDeundiak (approved):

PR #4407, opened 2023-03-06
fix(rust): fixed compatibility with older kafka protocol versions

davide-baldo:
## Current behavior

When older Kafka clients are used, the decoding of Kafka packets breaks, closing the connection.

## Proposed changes

Fixed a bug during encoding and decoding of request and response headers.
The header version doesn't depend solely on the API version, but it also depends on the type of message and whether it's a request or a response.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3571,12 +3571,11 @@ checksum = "624a8340c38c1b80fd549087862da4ba43e08858af025b236e509b6649fc13d5"
 
 [[package]]
 name = "open"
-version = "3.2.0"
+version = "3.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2078c0039e6a54a0c42c28faa984e115fb4c2d5bf2208f77d1961002df8576f8"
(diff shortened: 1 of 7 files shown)


hairyhum:
> The header version doesn't depend solely on the API version, but it also depends on the type of message and whether it's a request or a response.

From https://github.com/apache/kafka/blob/trunk/generator/src/main/java/org/apache/kafka/message/ApiMessageTypeGenerator.java#L329
and https://github.com/apache/kafka/tree/trunk/clients/src/main/resources/common/message

It seems like it depends on API version and whether it is request or response. Is there something else in there?

hairyhum (approved):

davide-baldo:
> It seems like it depends on API version and whether it is request or response. Is there something else in there?

It also depends on the type of request, from the code you linked:
```java
if (type.equals("response") && apiKey == 18) {
    buffer.printf("// ApiVersionsResponse always includes a v0 header.%n");
```

Other than these exceptions, I don't think there is anything else.

PR #4409, opened 2023-03-06
ci

mrinalwadhwa:
1. add *-ignored files to trigger paths
2. break out rust-lint-cargo-lock so it only runs on PRs / pushes changing cargo lock

### DIFF
--- a/.github/workflows/rust-ignored.yml
+++ b/.github/workflows/rust-ignored.yml
@@ -11,7 +11,7 @@ on:
       - "**.rs"
       - "**.toml"
       - "**/Cargo.lock"
-      - "implementations/rust/ockam/ockam_command/tests/**.bats"
+      - "implementations/rust/ockam/ockam_command/tests/**"
       - "**.gradle"
       - "tools/gradle/**"
       - "gradlew"
@@ -89,12 +89,6 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - run: 'echo "Rust - test_nightly - Ignored"'
-
-  lint_cargo_lock:
-    name: Rust - lint_cargo_lock
-    runs-on: ubuntu-20.04
-    steps:
-      - run: 'echo "Rust - lint_cargo_lock - Ignored"'
 
   test_ockam_command:
     name: Rust - test_ockam_command
--- a/.github/workflows/rust-lint-cargo-lock-ignored.yml
+++ b/.github/workflows/rust-lint-cargo-lock-ignored.yml
@@ -1,0 +1,19 @@
+name: Rust
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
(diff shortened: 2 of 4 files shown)


.github/workflows/elixir-ignored.yml
```diff
@@ -7,6 +7,7 @@ on:
   pull_request:
     paths-ignore:
       - '.github/workflows/elixir.yml'
+      - '.github/workflows/elixir-ignored.yml'
```
metaclips: If we paths-ignore this file, that'll mean that if we create a PR that changes `only` this file, the PR will block as Elixir workflow is required.

mrinalwadhwa: Hmm, really good point

mrinalwadhwa: removed it.

mrinalwadhwa:
The cargo lock lint fails a lot. Its not clear to me why its been succeeding until now
Screenshot 2023-03-06 at 12 02 02 PM

metaclips:
> The cargo lock lint fails a lot. Its not clear to me why its been succeeding until now Screenshot 2023-03-06 at 12 02 02 PM

Looking at the `Cargo.lock` file [here](https://github.com/build-trust/ockam/pull/4414/files), it seems like dependabot is selectively updating packages using `cargo update --package <package_name>` where package name in `PR 4414` is `io-lifetimes` and then `lint_cargo_lock` workflow then calls a `cargo update` when updates all dependencies, and the workflow fails due to the discrepancies.

metaclips:
> The cargo lock lint fails a lot. Its not clear to me why its been succeeding until now Screenshot 2023-03-06 at 12 02 02 PM

My findings 
- Dependabot selectively updates crates per PR by calling `cargo update <crate_name>` to update `Cargo.lock` file
- CI deletes `Cargo.lock` file and then tries to generate a new `Cargo.lock` which adds all pinned dependencies and updates their transitive dependencies if any.

I believe we have a special case here, PR #4413 and #4414 both update transitive dependencies `open v3.4.0`, and `io-lifetimes v1.0.6` simultaneously. Checking `lock lint workflow` for #4413 shows that `cargo check` downloads [open v3.4.0](https://github.com/build-trust/ockam/actions/runs/4347030189/jobs/7593841413#step:6:60) and also downloads [io-lifetimes v1.0.6](https://github.com/build-trust/ockam/actions/runs/4347030189/jobs/7593841413#step:6:179) crates which are added as separate updates in different PRs, so Lint workflow fails (which is intended) as we are having different Cargo.lock files.

I believe this might have slipped in other dependabot PRs as we didn't add Cargo Lock Lint workflow as a required workflow till of recent, so PRs get merged even if linting fails, or we never had scenarios of two PRs updating transitive dependencies.

PR #4422, opened 2023-03-07
fix(rust): fixed the compilation errors with the tag feature

etorreborre:
and removed unused code under that flag

Thanks to @nidnogg for spotting this issue!


mrinalwadhwa:
It would be good to have a test for this in CI
https://github.com/build-trust/ockam/blob/develop/.github/workflows/rust.yml

davide-baldo (dismissed):

davide-baldo (approved):

PR #4427, opened 2023-03-08
Improve the credential exchange example

etorreborre:
This PR removes the dependency on `ockam_core` (and replaces it with a dependency of `ockam`) + changes a log level to clean the execution trace of nodes when stopping the context.


davide-baldo (approved):

etorreborre:
Thanks @davide-baldo

PR #4430, opened 2023-03-08
test(rust): cleanup unused variables from bats tests

adrianbenavides:
This [section here](https://github.com/build-trust/ockam/pull/4430/files#diff-7b9e7353cc1d5bad26ceadd80529b67b6b6585cde34ff581e4789f6bbff687ecR117-R118) also fixes a problem that was causing the bats tests to hang right before exiting even though all the tests where successful.


metaclips (approved):
Thanks!

PR #4440, opened 2023-03-09
docs: add influxdb demo

glenngillen:
Adding in a reusable demo to compliment the [blog post about InfluxDB](https://www.ockam.io/blog/connect_private_influxdb).


glenngillen:
Not sure how to tell it to skip over running the shell linting in the demo folders. The file it's erroring on is a copy of an InfluxDB managed script so not really something I want to be reformatting unnecessarily.

etorreborre:
@glenngillen I guess you can try to tweak the `find` search [here](https://github.com/build-trust/ockam/blob/develop/.github/workflows/shell.yml#L57) so that it skips the `demo` folders.

demos/influxdb-telegraf/bin/setup
```diff
@@ -0,0 +1,21 @@
...
+        Yes ) break;;
+        No ) exit;;
+    esac
+done
+ockam reset
+ockam enroll
+
+PROJECT_NAME="default"
+result=$(ockam project information $PROJECT_NAME &>/dev/null)
+# if [ ! $? -eq 0 ]; then
```
etorreborre: Let's remove the comments if we don't use that code

etorreborre (dismissed):
Looks all good to me

etorreborre (approved):

PR #4445, opened 2023-03-09
fix(elixir): healthcheck improvements

hairyhum:
- support multiple healthcheck targets configuration
- properly cleanup secure channels and connections
- report current status of a target as 1 or 0

Additional changes
- always start TCP client as restart_type: :temporary
- add :disconnect protocol to identity secure channels
- support :host and :port options in tcp client creation (instead of :destination)
- make create function overridable in workers (see TCP.Client for example)

## Checks


hairyhum:
New secure channel service messages protocol.

For messages directed to identity secure channel coming from encryption secure channel side (onward_route terminates on the inner identity secure channel address).
If the message payload can be decoded as CBOR:
```
service_message = {
  0: optional(tag),
  1: command
}
command = 0 ;; disconnect
```

The service message is processed.

implementations/elixir/ockam/ockam/lib/ockam/worker.ex
```diff
@@ -88,7 +88,9 @@ defmodule Ockam.Worker do
       ## handle_message definition may not return {:error, ...} and it shouldn't fail because of that
       @dialyzer {:no_match, handle_info: 2, handle_continue: 2}
 
-      def create(options \\ [], timeout \\ @default_timeout) when is_list(options) do
+      def create(options \\ [], timeout \\ @default_timeout)
+
+      def create(options, timeout) when is_list(options) do
```
mszpakowski, line 93: Just curious, do we need to extract the function head here?

implementations/elixir/ockam/ockam_healthcheck/test/healthcheck/config_test.exs
```diff
@@ -0,0 +1,23 @@
...
+      "host":"localhost",
+      "port": 4000,
+      "api_worker": "api",
+      "healthcheck_worker": "healthcheck"}]
+    """
+
+    {:ok, _targets} = Ockam.Healthcheck.Application.parse_config(config)
+
+    bad_target = "[{\"host\":\"localhost\"}]"
+    {:error, {:invalid_target, _error}} = Ockam.Healthcheck.Application.parse_config(bad_target)
```
mszpakowski: Should we add `assert` here and on line 21?

implementations/elixir/ockam/ockam_healthcheck/test/healthcheck/config_test.exs
```diff
@@ -0,0 +1,23 @@
+defmodule Ockam.Healthcheck.ConfigTests do
```
mszpakowski: This is a nit and maybe just my personal preference, but I always aim to have `test` folder structure resemble the `lib` folder structure for unit tests. So I personally would put this test in the `test/healthcheck/application_test.exs` file and name this module `Ockam.Healthcheck.ApplicationTest`(or alternatively create a `Ockam.Healthcheck.Config` module in the `healthcheck/config.ex` file and put the `parse_config/1` function there).

I find it to be a good way to keep track where unit tests for which module live and which of its interfaces are tested.

implementations/elixir/ockam/ockam_healthcheck/test/healthcheck/healthcheck_test.exs
```diff
@@ -24,41 +24,62 @@ defmodule Ockam.Tests do
...
-    assert :ok = Ockam.Healthcheck.check_node(node_host, node_port, api_worker, ping_worker, 1000)
+  test "healthcheck targets OK" do
+    target = %{
+      host: "localhost",
+      port: 4000,
+      api_worker: "api",
+      healthcheck_worker: "healthcheck"
+    }
+
+    Application.put_env(:ockam_healthcheck, :targets, [target, target])
```
mszpakowski, line 39: Should we restore the original value after this test finishes? Otherwise the config change will during the whole test suite run and could interfere with other tests (and for that reason I also don't think we should use `async: true` option)

implementations/elixir/ockam/ockam_healthcheck/lib/application.ex
```diff
@@ -14,14 +14,52 @@ defmodule Ockam.Healthcheck.Application do
...
+        {:error, {:invalid_config, other}}
+
+      {:error, reason} ->
+        {:error, {:invalid_config, reason}}
+    end
+  end
+
   def healthcheck_schedule() do
     ## TODO: crontab resolution is per minute
     ## maybe we can use some other tool to schedule healthchecks?
```
mszpakowski: Not sure if I understand this TODO correctly, but with `SchedEx` we can schedule jobs at higher frequencies than 1 minute:

> Building on the support provided by the crontab library, SchedEx supports extended crontabs. Such crontabs have 7 segments instead of the usual 5; one is added to the beginning of the crontab and expresses a seconds value, and one added to the end expresses a year value. As such, it's possible to specify a unique instant down to the second

implementations/elixir/ockam/ockam_healthcheck/lib/healthcheck/healthcheck.ex
```diff
@@ -9,35 +9,38 @@ defmodule Ockam.Healthcheck do
...
-           identity: identity,
-           key_exchange_timeout: 10_000
-         ) do
-      {:ok, channel} ->
+  def ping_target(target, timeout) do
+    %{host: host, port: port, api_worker: api_worker, healthcheck_worker: healthcheck_worker} =
+      target
+
+    with_tcp(host, port, fn conn ->
+      with_channel(conn, api_worker, fn channel ->
```
mszpakowski, line 43: Just a note, as I don't have a suggestion on how to simplify this, but it's difficult to follow these wrappers around anonymous functions nested inside each other. Probably a refactoring opportunity for the future (maybe by making it more explicit at a cost of it being less DRY?).

implementations/elixir/ockam/ockam/lib/ockam/worker.ex
hairyhum: Because if we want to override it for a specific worker there will be a warning.

implementations/elixir/ockam/ockam_healthcheck/lib/application.ex
hairyhum: Good point. I didn't see that.

implementations/elixir/ockam/ockam_healthcheck/lib/healthcheck/healthcheck.ex
hairyhum: Explicit wrapping of try/catch blocks doesn't look good at all and we don't really have a good resource cleanup RN. It's less about DRY as it is about readability.

implementations/elixir/ockam/ockam_healthcheck/test/healthcheck/config_test.exs
```diff
@@ -1,23 +0,0 @@
-defmodule Ockam.Healthcheck.ConfigTests do
-  use ExUnit.Case, async: true
-
-  require Logger
-
-  test "parse target config" do
```
mszpakowski: Was it intended for this test to be removed completely?

mszpakowski (dismissed):

mszpakowski (dismissed):

polvorin (approved):

PR #4446, opened 2023-03-10
feat(rust): added `tcp-outlet` and `tcp-inlet` `show` commands

### LINKED ISSUE #4421: Create clap command to show the details of an existing TCP inlet on a node
### Current behavior

Missing command for `tcp-inlet show`.

### Desired behavior

Implement a command that retrieves from a running node the information of a `tcp-inlet` given its alias.

```bash
ockam tcp-inlet show [alias] --at [node-name]
```

Required args:
- alias

Optional args:
- at: defaults to the default node

Output: same as described at [`tcp-outlet list`](https://github.com/build-trust/ockam/tree/develop/implementations/rust/ockam/ockam_command/src/tcp/outlet/list.rs#L34).

Similar command: [`tcp-outlet delete`](https://github.com/build-trust/ockam/tree/develop/implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs#L14)

### LINKED ISSUE #4435: Create clap command to show the details of an existing TCP outlet on a node
### Current behavior

Missing command for `tcp-outlet show`.

### Desired behavior

Implement a command that retrieves from a running node the information of a `tcp-outlet` given its alias.

```bash
ockam tcp-outlet show [alias] --node [node-name]
```

Required args:
- alias

Optional args:
- node: defaults to the default node. Use the `NodeOpts` struct to define this argument.

Output: same as described at [`tcp-outlet list`](https://github.com/build-trust/ockam/tree/develop/implementations/rust/ockam/ockam_command/src/tcp/inlet/list.rs#L34).

Similar command: [`tcp-outlet delete`](https://github.com/build-trust/ockam/tree/develop/implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs#L14)

nidnogg:
## Current behavior

There is no way to directly show a TCP Outlet or Inlet by its `alias` and `node`.

## Proposed changes

- Solves https://github.com/build-trust/ockam/issues/4421 and https://github.com/build-trust/ockam/issues/4435 by introducing the `tcp-outlet show` and `tcp-inlet show` commands

- Adds corresponding bats tests

- Refactored `DeleteInlet` and `DeleteOutlet` into an in-route `alias` param

- Adjusted previous `tcp-inlet`/`tcp-outlet` CRUD tests to use output from `tcp-inlet`/`tcp-outlet` `show` commands in assertions

- Added missing `tcp-outlet list` test

- Fixed a bug where the `tcp-inlet CRUD` test would fail to properly assert the deletion of a non existing inlet

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -657,25 +657,19 @@ impl NodeManagerWorker {
 
             // ==*== Inlets & Outlets ==*==
             (Get, ["node", "inlet"]) => {
-                if req.has_body() {
-                    self.show_inlet(req, dec).await?.to_vec()?
-                } else {
-                    let node_manager = self.node_manager.read().await;
-                    self.get_inlets(req, &node_manager.registry).to_vec()?
-                }
+                let node_manager = self.node_manager.read().await;
+                self.get_inlets(req, &node_manager.registry).to_vec()?
             }
+            (Get, ["node", "inlet", alias]) => self.show_inlet(req, alias).await?.to_vec()?,
             (Get, ["node", "outlet"]) => {
-                if req.has_body() {
-                    self.show_outlet(req, dec).await?.to_vec()?
-                } else {
-                    let node_manager = self.node_manager.read().await;
-                    self.get_outlets(req, &node_manager.registry).to_vec()?
-                }
+                let node_manager = self.node_manager.read().await;
+                self.get_outlets(req, &node_manager.registry).to_vec()?
             }
+            (Get, ["node", "outlet", alias]) => self.show_outlet(req, alias).await?.to_vec()?,
             (Post, ["node", "inlet"]) => self.create_inlet(req, dec, ctx).await?.to_vec()?,
             (Post, ["node", "outlet"]) => self.create_outlet(req, dec).await?.to_vec()?,
-            (Delete, ["node", "outlet"]) => self.delete_outlet(req, dec).await?.to_vec()?,
-            (Delete, ["node", "inlet"]) => self.delete_inlet(req, dec).await?.to_vec()?,
+            (Delete, ["node", "outlet", alias]) => self.delete_outlet(req, alias).await?.to_vec()?,
+            (Delete, ["node", "inlet", alias]) => self.delete_inlet(req, alias).await?.to_vec()?,
             (Delete, ["node", "portal"]) => todo!(),
 
             // ==*== Workers ==*==
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -657,13 +657,19 @@ impl NodeManagerWorker {
 
             // ==*== Inlets & Outlets ==*==
             (Get, ["node", "inlet"]) => {
-                let node_manager = self.node_manager.read().await;
-                self.get_inlets(req, &node_manager.registry).to_vec()?
+                let inlet_registry = {
+                    let node_manager = self.node_manager.read().await;
+                    &node_manager.registry.inlets.clone()
+                };
+                self.get_inlets(req, inlet_registry).to_vec()?
             }
             (Get, ["node", "inlet", alias]) => self.show_inlet(req, alias).await?.to_vec()?,
             (Get, ["node", "outlet"]) => {
-                let node_manager = self.node_manager.read().await;
-                self.get_outlets(req, &node_manager.registry).to_vec()?
+                let outlet_registry = {
+                    let node_manager = self.node_manager.read().await;
+                    &node_manager.registry.outlets.clone()
+                };
+                self.get_outlets(req, outlet_registry).to_vec()?
             }
             (Get, ["node", "outlet", alias]) => self.show_outlet(req, alias).await?.to_vec()?,
             (Post, ["node", "inlet"]) => self.create_inlet(req, dec, ctx).await?.to_vec()?,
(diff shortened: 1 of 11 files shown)


implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -657,18 +657,25 @@ impl NodeManagerWorker {
...
             }
             (Get, ["node", "outlet"]) => {
-                let node_manager = self.node_manager.read().await;
-                self.get_outlets(req, &node_manager.registry).to_vec()?
+                if req.has_body() {
+                    self.show_outlet(req, dec).await?.to_vec()?
+                } else {
+                    let node_manager = self.node_manager.read().await;
+                    self.get_outlets(req, &node_manager.registry).to_vec()?
+                }
```
adrianbenavides: The endpoints should be:

- `node/inlet/{alias}`: show
- `node/inler`: list

And the same for `outlet`. Given that the `show` handlers only need the `alias` parameter, you can simplify them by removing the request body altogether.

Another little improvement you could do here: the `node_manager` lock is being held unnecessarily during the execution of the get/show function calls. This could be refactored as such:

```rust
let registry = {
    &node_manager.registry
};
```

adrianbenavides (changes requested):
Looks pretty great, thank you very much @nidnogg ! I've left a small comment on something that could be simplified. Other than that, looks pretty solid.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
nidnogg: Thanks for the suggestions, I think they made for way more straightforward endpoints. I refactored the `PortalAlias` struct out entirely from the `show` and `delete` commands using that first tip.

However, regarding the second improvement:
> Another little improvement you could do here: the `node_manager` lock [...]

After trying out several attempts at implementing it, not only did I get halted by the compiler with the same `borrow does not live long enough`, `lifetime ends here` errors, it also occurred to me this might possibly lead to a race condition if freed up before the request ends anyway - the registry could be modified before being read (or vice-versa), or modified improperly.

I tried:
- adding the block outside the `portals.rs` methods right after reading the request, replacing what's currently done;
- adding the `let registry { [...] }` block inside the `portals.rs` methods themselves;
- changing the `Registry` lifetime to `'static` (although I felt this would be a drawback in itself), and the fix you suggested

Let me know if you think it's still worth going down that route or if I misunderstood that second one! ✌️

adrianbenavides: Oh sorry, you are right, I didn't see the problem when I wrote the comment. The issue is that you can't get a reference to the data that is held behind the lock after the lock has been dropped. The solution to making that approach work would be to clone the whole registry. To mitigate the cost of cloning the entire registry, you could refactor the `get_inlets/outlets` parameters to receive only the part of the registry that is needed. There is another solution to avoid cloning data at the cost of a bigger memory footprint, which is having the registry behind an `Arc`. 

I'd go for the second option: cloning the part of the registry you need and refactoring the method's signature accordingly.

> the registry could be modified before being read (or vice-versa), or modified improperly.

About this, note that we are only reading the registry, so a modification should not be possible or needed.

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
```diff
@@ -405,7 +427,34 @@ impl NodeManagerWorker {
...
+            )))
+        }
+    }
+
+    pub(super) async fn show_outlet<'a>(
+        &mut self,
+        req: &Request<'_>,
+        alias: &'a str,
+    ) -> Result<ResponseBuilder<OutletStatus<'a>>> {
+        let node_manager = self.node_manager.write().await;
```
adrianbenavides: You could use a read lock here: `self.node_manager.read().await;`

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
```diff
@@ -288,7 +284,36 @@ impl NodeManagerWorker {
...
+            )))
+        }
+    }
+
+    pub(super) async fn show_inlet<'a>(
+        &mut self,
+        req: &Request<'_>,
+        alias: &'a str,
+    ) -> Result<ResponseBuilder<InletStatus<'a>>> {
+        let node_manager = self.node_manager.write().await;
```
adrianbenavides: Same here: `self.node_manager.read().await;`

nidnogg: Added in latest commit

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
nidnogg: Added in latest commit

implementations/rust/ockam/ockam_api/src/nodes/service.rs
nidnogg: > About this, note that we are only reading the registry, so a modification should not be possible or needed.

A question for the strategy of cloning part of the registry:

The `show` and `list` commands are read only but the `delete` commands also use these registries and with `write` locks - if we had a `delete` write lock assume context after the cloning that part of the registry, wouldn't the output be inaccurate at this point (i.e. not correspond to the actual registry)?

If not, I've added the workaround with a partial registry `clone()` modifying the corresponding signatures in `get_outlets()` and `get_inlets()` from `portals.rs`. Otherwise I could roll these back no problem. The rest of the unrelated fixes are there as well.

adrianbenavides: > but the delete commands also use these registries and with write locks - if we had a delete write lock assume context after the cloning that part of the registry, wouldn't the output be inaccurate at this point (i.e. not correspond to the actual registry)?

Exactly, you can't apply the same trick to the endpoints that modify the registry. In those cases, it makes total sense to hold the lock until you are finished, so other concurrent calls to the NodeManager that affects the registry do not interfere with the current execution. As you can imagine, the problem here is the actual design of the registry: having all those sub-registries behind the same lock is not the most efficient idea because we shouldn't block modifying the tcp-connections registry when we know we only want to work with the tcp-outlets registry, for example. But right now is not critical either, so it's just fine.

adrianbenavides (approved):
LGTM! Awesome job @nidnogg , thank you so much for contributing to Ockam 💪

PR #4453, opened 2023-03-12
More Sessions

SanjoDeundiak:


implementations/rust/ockam/ockam_core/src/sessions.rs
```diff
@@ -1,28 +1,50 @@
...
-//! for security reasons.
+//! Allows limiting messaging between workers for security reasons (Message Flow Authorization).
+//!
+//! Sessions imply 3 roles:
+//!
+//! - Producers, that produce messages (that usually originate on other nodes), which are
+//!     potentially malicious and should be limited in terms of which workers they can reach.
+//!     Tcp Receiver is an examples of a Producer.
+//! - Consumers, that are allowed to consume potentially malicious messages from Producers.
+//!     Secure Channel Decryptor is an examples of Consumer.
```
etorreborre: s/examples/example

implementations/rust/ockam/ockam_core/src/sessions.rs
```diff
@@ -31,79 +53,92 @@ impl Sessions {
...
     }
 
-    /// Mark that given [`Address`] belongs to the given [`SessionId`]
-    pub fn set_session_id(&self, address: &Address, session_id: &SessionId) {
-        let mut lock = self.internal.write().unwrap();
-        let address_info = lock.get_info_mut(address);
+    /// Mark that given [`Address`] is a Consumer for Producer or Spawner with the given [`SessionId`]
+    pub fn add_consumer(&self, address: &Address, session_id: &SessionId, policy: SessionPolicy) {
+        let mut consumers = self.consumers.write().unwrap();
+        if !consumers.contains_key(session_id) {
```
etorreborre, line 59: You could possibly use a [default map](https://docs.rs/defaultmap/latest/defaultmap/) here.

implementations/rust/ockam/ockam_core/src/sessions.rs
```diff
@@ -31,79 +53,92 @@ impl Sessions {
...
 
-impl SessionsInternal {
-    fn get_info_mut(&mut self, address: &Address) -> &mut AddressInfo {
-        if !self.data.iter().any(|x| &x.address == address) {
-            let info = AddressInfo::new(address.clone(), None, None);
-            self.data.push(info);
-        }
+/// Known Consumers for the given [`SessionId`]
+#[derive(Default, Clone, Debug)]
+pub struct SessionConsumersInfo(BTreeMap<Address, SessionPolicy>);
```
etorreborre: In terms of naming I think that `ConsumersInfo` is enough, and symmetrical with `ProducerInfo`

implementations/rust/ockam/ockam_core/src/sessions.rs
```diff
@@ -31,79 +53,92 @@ impl Sessions {
...
+
+/// Set of Consumers and Spawners for the given [`SessionId`]
```
etorreborre: The comment needs to be adjusted

implementations/rust/ockam/ockam_core/src/sessions.rs
```diff
@@ -1,28 +1,50 @@
...
 mod local_info;
+mod policy;
 mod session_id;
 
 pub use access_control::*;
 pub use local_info::*;
+pub use policy::*;
 pub use session_id::*;
 
 // TODO: Consider integrating this into Routing for better UX + to allow removing
```
etorreborre, line 37: I'm starting to get concerned about the "removing" part, and mid-term issues with memory :-)

implementations/rust/ockam/ockam_identity/tests/plaintext_message_flow_auth.rs
```diff
@@ -0,0 +1,162 @@
...
+
+// Alice: Secure Channel. With session
+// Bob: Secure Channel listener. With session
+#[ockam_macros::test]
+async fn test1(ctx: &mut Context) -> Result<()> {
+    let sessions_alice = Sessions::default();
+    let session_id_alice_channel = sessions_alice.generate_session_id();
+
+    let sessions_bob = Sessions::default();
+    let session_id_bob_channel = SessionId("bob".to_string()); //sessions_bob.generate_session_id();
```
etorreborre: Is the comment a result of some debugging and should be uncommented now?

implementations/rust/ockam/ockam_identity/tests/plaintext_message_flow_auth.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    let alice = Identity::create(ctx, Vault::create()).await?;
+    let bob = Identity::create(ctx, Vault::create()).await?;
+
+    bob.create_secure_channel_listener(
```
etorreborre, line 27: A builder API would be a bit safer here:
```rust
let listener = bob.make_secure_channel_listener()
  .with_name("listener")
  .with_session(&sessions_bob, &session_id_bob_channel);
listener.start().await?;
```

This way there is no need to remember that a listener should use the `as_spawner` trust option and a connection should use `as_producer`.

implementations/rust/ockam/ockam_identity/tests/plaintext_message_flow_auth.rs
```diff
@@ -0,0 +1,162 @@
...
+    let channel_to_bob = alice
+        .create_secure_channel(
+            route!["listener"],
+            SecureChannelTrustOptions::new()
+                .as_producer(&sessions_alice, &session_id_alice_channel),
+        )
+        .await?;
+
+    ctx.sleep(Duration::from_millis(50)).await; // Wait for workers to add themselves to the registry
+    let channel_to_alice = bob
```
etorreborre, line 42: This code could go closer to L.60

implementations/rust/ockam/ockam_identity/tests/plaintext_message_flow_auth.rs
```diff
@@ -0,0 +1,162 @@
...
+            .as_consumer(
+                &sessions_bob,
+                &session_id_bob_tcp,
+                SessionPolicy::SpawnerAllowOnlyOneMessage,
+            )
+            .as_spawner(&sessions_bob, &session_id_bob_plaintext),
+    )
+    .await?;
+
+    let channel_to_bob = alice
```
etorreborre, line 124: Possibly if we could also use a builder pattern to create then start the TCP listener then we could write
```rust
let tcp_listener = tcp_bob
        .listen_to("127.0.0.1:0")
        .with_session(&sessions_bob, &session_id_bob_tcp));
let (socket_addr, _) = tcp_listener.start().await?;

let secure_channel_listener = 
       bob.create_secure_channel_listener()
          .with_name("listener")
          .with_session(&sessions_bob, &session_id_bob_plaintext)
          .consuming_one_message_from(tcp_listener.get_session_id());

secure_channel_listener.start().await?;
```

I don't know what is the best API but we can probably reduce the risk of misuse like this. And you see where this is going :-). What is stopping us from writing:
```rust
// for a secure channel connection using a dedicated TCP Listener
let secure_channel_listener = 
       bob.create_secure_channel_listener()
          .with_name("listener")
          // this generates the session_id
          // it is available via secure_channel_listener.get_session_id()
          .with_sessions(&sessions) 
          .with_transport(TCP, "127.0.0.1:0");

// for a secure channel connection using a TCP Listener shared with 
// other secure channel listeners
let secure_channel_listener = 
       bob.create_secure_channel_listener()
          .with_name("listener")
          .with_sessions(&sessions)
          .with_shared_transport(tcp_listener);
```

implementations/rust/ockam/ockam_node/Cargo.toml
```diff
@@ -25,7 +25,7 @@ publish = true
 rust-version = "1.56.0"
 
 [features]
-default = ["std"]
+default = ["std", "debugger"]
```
etorreborre: Out of curiosity, why did you need this?

implementations/rust/ockam/ockam_transport_tcp/src/trust_options.rs
```diff
@@ -1,8 +1,11 @@
...
-use ockam_core::sessions::{SessionId, SessionOutgoingAccessControlBuilder, Sessions};
-use ockam_core::{IncomingAccessControl, LocalOnwardOnly, LocalSourceOnly, OutgoingAccessControl};
+use ockam_core::sessions::{SessionId, SessionOutgoingAccessControl, Sessions};
+use ockam_core::{
+    Address, AllowAll, IncomingAccessControl, LocalOnwardOnly, LocalSourceOnly,
+    OutgoingAccessControl,
+};
 
-pub(crate) struct TcpConnectionAccessControl {
+pub(crate) struct TcpConnectionTrustOptionsProcessed {
```
etorreborre, line 8: I think I preferred the old name (with a plural) because it explains better what it is, rather than where it came from.

implementations/rust/ockam/ockam_transport_tcp/src/trust_options.rs
```diff
@@ -11,39 +14,43 @@ pub(crate) struct TcpConnectionAccessControl {
...
-        match self.session {
-            Some((sessions, session_id)) => TcpConnectionAccessControl {
-                session_id: Some(session_id.clone()),
-                sender_incoming_access_control: Arc::new(LocalSourceOnly),
-                receiver_outgoing_access_control: Arc::new(
-                    SessionOutgoingAccessControlBuilder::new(session_id, sessions).build(),
-                ),
-            },
-            None => TcpConnectionAccessControl {
+    pub(crate) fn process(self, address: &Address) -> TcpConnectionTrustOptionsProcessed {
```
etorreborre, line 37: I would call this function `create_access_controls`

implementations/rust/ockam/ockam_identity/src/channel/trust_options.rs
```diff
@@ -15,19 +18,34 @@ impl Default for SecureChannelTrustOptions {
     }
 }
 
+pub(crate) struct SecureChannelTrustOptionsProcessed {
```
etorreborre, line 21: See my comment on the `Processed` naming. I also thing that `SecureChannelAccessControls` would be better here.

implementations/rust/ockam/ockam_identity/src/channel/trust_options.rs
```diff
@@ -36,11 +54,55 @@ impl SecureChannelTrustOptions {
         self.trust_policy = Arc::new(trust_policy);
         self
     }
+
+    pub(crate) fn process(self, addresses: &Addresses) -> SecureChannelTrustOptionsProcessed {
```
etorreborre, line 58: I think we are cramming too many things in there:

 - the trust policy is copied in the `SecureChannelTrustOptionsProcessed` but could also be passed as an independent parameter to `Decryptor::create_initiator`. Then we can call `SecureChannelTrustOptionsProcessed` `SecureChannelAccessControls` to mirror my suggestion of `TcpConnectionAccessControls` instead of `TcpConnectionTrustOptionsProcessed`

Then:

 - we are having 2 side-effects here
   - creating a session id
   - modifying the global `Sessions` state

I think it would be better to separate that in 2 functions:

 - `create_secure_channel_session(&self, sessions: &Sessions, addresses: &Addresses) -> Option<SessionId>`
    to update the sessions and return the session id
 -  `create_access_controls(&self, session_id) -> SecureChannelAccessControls`

I know that this entails some of the same checks (is there a `producer_session` etc...) but the data flow of what is modified and what is created would be clearer IMO.

etorreborre (dismissed):
I think this all works but I feel concerned about the level of complexity we add internally and that we expose to users. IMO it is worth exploring some builder patterns on the user side to encapsulate the major use cases for setting up secure channels

implementations/rust/ockam/ockam_core/src/sessions.rs
SanjoDeundiak: I don't think it's worth introducing a new dependency

implementations/rust/ockam/ockam_core/src/sessions.rs
SanjoDeundiak: My plan was if we decide this is a good approach - we can embed it inside the ockam_node and all resources will be cleaned when a Worker is stopped

implementations/rust/ockam/ockam_identity/tests/plaintext_message_flow_auth.rs
SanjoDeundiak: I think both approaches are fine. For know I would like to avoid breaking backwards compatibility and updating every documentation snippet again :)

implementations/rust/ockam/ockam_identity/tests/plaintext_message_flow_auth.rs
SanjoDeundiak: I agree, but I also see sense in keeping "creation of secure channel" and "getting its address on listener side" close.

implementations/rust/ockam/ockam_identity/tests/plaintext_message_flow_auth.rs
SanjoDeundiak: I think there is a disadvantage of coupling transports and secure channels here, which should not depend on each other. But I think adding some sort of higher lever API that would couple things but introduce better UX may be a good idea in the future. For now, definitely out of scope of this PR

implementations/rust/ockam/ockam_node/Cargo.toml
SanjoDeundiak: Nice catch, forgot to remove

implementations/rust/ockam/ockam_transport_tcp/src/trust_options.rs
SanjoDeundiak: I'm not saying new names are perfect, but the reason of renaming is that `TcpConnectionAccessControl` and `SecureChannelAccessControl` aren't actually what those structs are, they have more that AccessControl in them

implementations/rust/ockam/ockam_identity/src/channel/trust_options.rs
SanjoDeundiak: The names would be somewhat clearer, on the other side I don't see a point of having 2 separate functions if they're not used separately. Or am I missing something from your point?

implementations/rust/ockam/ockam_core/src/sessions.rs
etorreborre: yes, I'm not sure if it's worth it either :-)

implementations/rust/ockam/ockam_core/src/sessions.rs
etorreborre: ok

implementations/rust/ockam/ockam_transport_tcp/src/trust_options.rs
etorreborre: Well if you move the `TrustPolicy` out of the `SecureChannelTrustOptionsProcessed` that's mostly it. Then there is an optional `session_id` but I don't see where it is actually used, is it?

SanjoDeundiak: It's used in Decryptor. So only 1 of 3 things is AccessControl, that's why I didn't want to call it AccessControl

implementations/rust/ockam/ockam_identity/src/channel/trust_options.rs
etorreborre: Yes those 2 functions can not be used separately. But that function looks innocuous and is actually modifying _shared global state_ in a non-trivial way so I would be more comfortable highlighting this aspect with a dedicated function and associated documentation (call that scars of years of Java code :-)). It's more about future readers / debuggers that I am thinking.

SanjoDeundiak: Ok, I can try to clean it in a following PR

implementations/rust/ockam/ockam_transport_tcp/src/trust_options.rs
etorreborre: I see. My IDE was not showing a usage for `plaintext_session_id`, I don't know why. Ok, let's keep it like that then.

etorreborre (approved):

PR #4467, opened 2023-03-15
feat(rust): create an authority node

etorreborre:
The purpose of this PR is to create a small binary representing an authority node (a node which authenticates identity attributes and issues credentials).

# Current situation

At the moment when we start an authority node we actually start a full node with many services. In particular, there is an unrestricted TCP listener started just to receive commands (to create inlets, outlets etc...).

# Proposed solution

This PR provides a node:

 - with a TCP listener which can only be used to create secure channels
 - starting only services related to the role of an authority

The authority node is started exactly as before in order to keep the Orchestrator deployment working
```
ockam node create \
  --tcp-listener-address=0.0.0.0:4200 \ 
  --launch-config=/mnt/storage/authority-conf.json \
  --reload-from-trusted-identities-file=/mnt/storage/trust-anchors.json \
  authority
```

In the command above the specific presence of `--launch-config` and `authority` as a node name will start a restricted authority node. The next step is to provide an independent `ockam authority create` command specifically for the authority node use case.


SanjoDeundiak (dismissed):
Looks great. The only thing that was hard to figure out from just reading the code - is this node and its state compatible with an only Authority node? Will it be able (with proper configuration file) to pick up the state from the Authority node that was already running before? Or we don't need that?

SanjoDeundiak:
@etorreborre could you please split your commit into few more granular if possible? Also, could you please use imperative-style commit messages?

etorreborre:
> Or we don't need that?

Yes this should work with the current `node create` command because I am reusing [the global paths to initialize the Authority vault and storage](https://github.com/build-trust/ockam/pull/4467/files#diff-afac108b9e4f2377c7932bef242708078b72d74836268ccc0ff4a09a24f67a53R564) (unless I misunderstood something of course). 

And this should also work with the upcoming `authority create` command.

PR #4468, opened 2023-03-15
kafka auto reconnect 

davide-baldo:
## Current behavior

Currently, whenever a project is relocated, or a connection simply goes down, there is no recovery of any kind and consumer and producer remain broken indefinitely.

[Superseding the old PR where ad-hoc solution was created](https://github.com/build-trust/ockam/pull/4375).

## Proposed changes

Instead of creating low-level inlet/outlets and forwarders, we leverage the `NodeManager` which comes with the connection supervisor (called session for now).
I also changed the topology: instead of having one connection for everything, each TCP inlet/outlet has its own encrypt or/decryptor and instantiate a `MultiAddr`. 
Some extra refactoring was done in the `NodeManager::connect()` method to generalize the `MultiAddr` as much as possible.


implementations/rust/ockam/ockam_api/src/session/mod.rs
```diff
@@ -25,7 +22,7 @@ pub struct Medic {
     delay: Duration,
     sessions: Arc<Mutex<Sessions>>,
     pings: JoinSet<(Key, Result<(), Error>)>,
-    replacements: JoinSet<(Key, Result<MultiAddr, Error>)>,
```
davide-baldo, line 29: The `MultiAddr` was actually used as a Route, I just made it clearer

implementations/rust/ockam/ockam_api/src/session/sessions.rs
```diff
@@ -21,16 +22,12 @@ pub struct Sessions {
 
 pub struct Session {
     key: Key,
-    addr: MultiAddr,
-    data: Data,
```
davide-baldo, line 25: data was a shared map where all the session re-creation information was stored in an encoded form.
I've simply moved the state, avoiding the map.

implementations/rust/ockam/ockam_api/src/util.rs
```diff
@@ -48,11 +48,19 @@ pub struct TcpSession {
...
 
 pub async fn create_tcp_session(ma: &MultiAddr, tcp: &TcpTransport) -> Option<TcpSession> {
-    let mut rb = Route::new();
-    let mut it = ma.iter().peekable();
     let sessions = Sessions::default();
     let session_id = sessions.generate_session_id();
+    create_tcp_session_ext(ma, tcp, sessions, session_id).await
+}
 
+pub async fn create_tcp_session_ext(
```
davide-baldo: I needed a way to keep using the same session across multiple secure channels

implementations/rust/ockam/ockam_core/src/routing/address.rs
```diff
@@ -28,11 +31,16 @@ use serde::{Deserialize, Serialize};
 /// * `"0#alice"` represents a local worker with the address: `alice`.
 /// * `"1#carol"` represents a remote worker with the address `carol`, reachable over TCP transport.
 ///
-#[derive(Serialize, Deserialize, Clone, Hash, Ord, PartialOrd, Eq, PartialEq)]
+#[derive(Serialize, Deserialize, Decode, Encode, Clone, Hash, Ord, PartialOrd, Eq, PartialEq)]
+#[rustfmt::skip]
+#[cbor(map)]
 pub struct Address {
```
davide-baldo, line 37: In order to use `Route` and **not** `MultiAddr` I added the Route ad as a cbor encodable.
We can express a `Route` as `MultiAddr` but we need to double cast it, i think sometimes a `Route`  is really what we want.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -379,119 +377,69 @@ impl NodeManager {
     }
 
     /// Resolve project ID (if any), create secure channel (if needed) and create a tcp connection
-    ///
-    /// Returns the secure channel worker address (if any) and the remainder
-    /// of the address argument.
+    /// Returns [`ConnectionInstance`]
     pub(crate) async fn connect(
```
davide-baldo, line 444: This method was completely rewritten to further generalize the `MultiAddr` handling

implementations/rust/ockam/ockam_api/src/nodes/connection/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    pub fn with_timeout(mut self, timeout: Duration) -> Self {
+        self.timeout = Some(timeout);
+        self
+    }
+}
+
+#[derive(Clone)]
+pub struct ConnectionInstance {
+    /// Transport route consist of only transport addresses,
```
etorreborre: s/consist/consists

implementations/rust/ockam/ockam_api/src/nodes/connection/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+    /// A fully normalized [`MultiAddr`] contains only Service entries.
+    pub normalized_addr: MultiAddr,
+    /// The original provided [`MultiAddr`]
+    pub original_addr: MultiAddr,
+    /// A list of secure channel encryptors created for the connection.
+    /// Needed to cleanup the connection resources when it must be closed.
+    pub secure_channel_encryptors: Vec<Address>,
+    /// A TCP worker address if used when instantiating the connection
+    pub tcp_worker: Option<Address>,
+    /// If a tcp session was created
```
etorreborre: Is it TCP specific?

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
```diff
@@ -756,7 +663,7 @@ impl NodeManagerWorker {
         req: &'a Request<'_>,
         dec: &mut Decoder<'_>,
     ) -> Result<Vec<u8>> {
-        let mut node_manager = self.node_manager.write().await;
+        // let mut node_manager = self.node_manager.write().await;
```
etorreborre: Let's remove that commented out code

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
```diff
@@ -769,22 +681,122 @@ impl NodeManagerWorker {
...
+        bind_ip: String,
+        server_bootstrap_port: u16,
+        brokers_port_range: (u16, u16),
+        project_multiaddr: MultiAddr,
+        kind: KafkaServiceKind,
+    ) -> Result<()> {
+        // interceptor_multiaddr:
+        //   local interceptor => project route => interceptor => (+ broker or bootstrap address)
+
+        let mut interceptor_multiaddr = project_multiaddr.clone();
```
etorreborre: A macro for `MultiAddress`, like we have for `Route` would be nice. Then we could write
```rust
let interceptor_multiaddr = multiaddr![
    Service::new(local_interceptor_address.to_string()),
    project_multiaddr.clone(),
    Service::new(ORCHESTRATOR_KAFKA_INTERCEPTOR_ADDRESS),
    ];

let bootstrap_multiaddr = multiaddr![
    interceptor_multiaddr,
    Service::new(ORCHESTRATOR_KAFKA_BOOTSTRAP_ADDRESS),
    ]
```

Which, I think, helps better visualizing what's going on.

implementations/rust/ockam/ockam_api/src/nodes/service/services.rs
```diff
@@ -769,22 +681,122 @@ impl NodeManagerWorker {
...
+                .map_err(|_| ApiError::message("invalid port range"))?,
+        );
+
+        //HORRIBLE HACK:
+        // since we cannot call APIs of node manager via message
+        // we must encode requests as an external
+
+        let mut buffer = Vec::new();
+        Encoder::new(&mut buffer).encode(CreateInlet::to_node(
+            SocketAddr::from_str(&format!("{bind_ip}:{server_bootstrap_port}"))
```
etorreborre: We shouldn't have to parse again this address at this stage. We parsed a valid `Ipv4Addr` address from the beginning, then lost its validation by transmitting it as a string and then parsing it again. We should keep it as `IpV4Addr` all the way down (why not `IpAddr` by the way?)

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -73,22 +83,55 @@ pub(crate) trait ForwarderCreator: Send + Sync + 'static {
     async fn create_forwarder(&self, context: &Context, alias: String) -> Result<()>;
 }
 
-pub(crate) struct RemoteForwarderCreator {
-    hub_route: Route,
+pub(crate) struct NodeManagerForwarderCreator {
+    hub_multiaddr: MultiAddr,
```
etorreborre: We should stop using `hub` now and use `orchestrator` instead

implementations/rust/ockam/ockam_api/src/kafka/inlet_controller.rs
```diff
@@ -0,0 +1,143 @@
...
+use crate::port_range::PortRange;
+use crate::route_to_multiaddr;
+
+type BrokerId = i32;
+
+/// Shared structure for every kafka worker to keep track of which brokers are being proxied
+/// with the relative inlet listener socket address.
+/// Also takes care of creating inlets dynamically when they are not present yet.
+#[derive(Debug, Clone)]
+pub(crate) struct KafkaInletController {
```
etorreborre, line 27: The name makes more sense, thanks 🎸

implementations/rust/ockam/ockam_api/src/nodes/connection.rs
```diff
@@ -44,3 +48,203 @@ impl<'a> Connection<'a> {
...
+        ockam_core::sessions::Sessions,
+        ockam_core::sessions::SessionId,
+    )>,
+    pub current_multiaddr: MultiAddr,
+    pub secure_channel_encryptors: Vec<Address>,
+}
+
+pub type Normalized =
+    Pin<Box<dyn Future<Output = Result<ReplacerChanges, ockam_core::Error>> + Send>>;
+pub type Normalizer<'a> =
```
etorreborre: Wouldn't it be simpler to define a `Normalizer` async trait?

implementations/rust/ockam/ockam_api/src/nodes/connection.rs
```diff
@@ -44,3 +48,203 @@ impl<'a> Connection<'a> {
...
+            secure_channel_encryptors: self.secure_channel_encryptors,
+            session: self.session,
+        }
+    }
+
+    pub async fn normalize<'a, I>(
+        mut self,
+        codes: I,
+        mut normalizer: Normalizer<'a>,
+    ) -> Result<Self, ockam_core::Error>
```
etorreborre: If you import `ockam_core::Result` then this is just `Result<Self>`

etorreborre (commented):
Just releasing a few comments that I left when I started doing the review

implementations/rust/ockam/ockam_api/src/nodes/connection/mod.rs
davide-baldo: Yes, it's specific when creating TCP connection/worker

implementations/rust/ockam/ockam_core/src/flow_control/flow_controls.rs
```diff
@@ -122,6 +122,37 @@ impl FlowControls {
             .map(|x| x.0.clone())
             .collect()
     }
+
+    /// Prints debug information regarding Flow Control for the provided addresses
+    #[allow(dead_code)]
+    pub fn debug_addresses(&self, addresses: Vec<Address>) {
```
SanjoDeundiak: Do you think it may be cleaner to accept only 1 Address here?

implementations/rust/ockam/ockam_core/src/flow_control/access_control.rs
```diff
@@ -92,6 +92,16 @@ impl OutgoingAccessControl for FlowControlOutgoingAccessControl {
             }
         }
 
+        #[cfg(feature = "debugger")]
+        self.flow_controls.debug_addresses(
```
SanjoDeundiak: I think it only makes sense to print debug info for current and next onward address, what is the point to debug the whole onward and return route? It's not even guaranteed that those addresses belong to our node

implementations/rust/ockam/ockam_core/src/routing/transport_type.rs
```diff
@@ -1,10 +1,13 @@
 use core::fmt::{self, Debug, Display};
+use minicbor::{Decode, Encode};
 use serde::{Deserialize, Serialize};
 
 /// The transport type of an [`Address`].
-#[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(
+    Serialize, Deserialize, Decode, Encode, Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash,
+)]
```
SanjoDeundiak, line 8: #[cbor(transparent)]

implementations/rust/ockam/ockam_core/src/routing/transport_type.rs
```diff
@@ -1,10 +1,13 @@
...
 #[serde(transparent)]
-pub struct TransportType(u8);
+pub struct TransportType(#[n(1)] u8);
```
SanjoDeundiak: I think index should be equal to `0`

implementations/rust/ockam/ockam_core/src/type_tag.rs
```diff
@@ -20,6 +22,28 @@ impl<const N: usize> fmt::Debug for TypeTag<N> {
     }
 }
 
+impl<const N: usize> Hash for TypeTag<N> {
```
SanjoDeundiak, line 25: Could you please add functional implementations to this file?

implementations/rust/ockam/ockam_api/src/util.rs
```diff
@@ -308,6 +319,20 @@ pub fn is_local_node(ma: &MultiAddr) -> anyhow::Result<bool> {
     }
 }
 
+/// Tells whether the input [`Code`] references a local worker.
+pub fn local_worker(code: &Code) -> Result<bool> {
+    match *code {
+        Node::CODE | Space::CODE | Project::CODE | DnsAddr::CODE | Ip4::CODE | Ip6::CODE => {
+            Ok(false)
+        }
+        Secure::CODE | Worker::CODE | Service::CODE => Ok(true),
```
SanjoDeundiak: Is `Secure` considered as a local worker in this context?

implementations/rust/ockam/ockam_api/src/session/sessions.rs
```diff
@@ -1,18 +1,24 @@
...
 use ockam_core::compat::rand;
-use ockam_core::Error;
-use ockam_multiaddr::MultiAddr;
-use std::sync::{Arc, Mutex};
-use tracing as log;
+use ockam_core::{Error, Route};
 
-pub type Replacement = Pin<Box<dyn Future<Output = Result<MultiAddr, Error>> + Send>>;
-pub type Replacer = Box<dyn FnMut(MultiAddr) -> Replacement + Send>;
+//most sessions replacer are dependent on the node manager, if many session
```
SanjoDeundiak, line 14: Off-topic: do you think instead of having a lock for the NodeManager state we should have individual locks for its fields?

implementations/rust/ockam/ockam_api/src/session/sessions.rs
```diff
@@ -87,11 +89,10 @@ impl Sessions {
 }
 
 impl Session {
-    pub fn new(addr: MultiAddr) -> Self {
+    pub fn new(addr: Route) -> Self {
```
SanjoDeundiak: Let's also rename it to `ping_route`

implementations/rust/ockam/ockam_api/src/session/mod.rs
```diff
@@ -163,25 +147,25 @@ impl Medic {
...
                             log::info!(key = %k, "replacing session");
                             self.replacements.spawn(async move { (k, f.await) });
                         }
                     }
-                    Some(Ok((k, Ok(a)))) => {
+                    Some(Ok((k, Ok(ping_route)))) => {
                         let mut sessions = self.sessions.lock().unwrap();
                         if let Some(s) = sessions.session_mut(&k) {
-                            log::info!(key = %k, addr = %a, "replacement is up");
+                            log::info!(key = %k, addr = %ping_route, "replacement is up");
```
SanjoDeundiak: ping_route = %ping_route

implementations/rust/ockam/ockam_api/src/session/mod.rs
```diff
@@ -78,32 +74,20 @@ impl Medic {
...
-                                } else {
-                                    log::error! {
-                                        key  = %key,
-                                        addr = %session.ping_address(),
-                                        "failed to convert address to route"
-                                    }
-                                    continue;
-                                };
-                            log::debug! {
+                            let echo_route: Route =
```
SanjoDeundiak: `: Route` is not needed anymore

implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
```diff
@@ -405,25 +409,38 @@ impl NodeManagerWorker {
             None => None,
         };
 
+        let flow_controls;
+        {
+            let node_manager = self.node_manager.write().await;
```
SanjoDeundiak: Can it be a read lock?

implementations/rust/ockam/ockam_core/src/flow_control/flow_controls.rs
davide-baldo: Yes, having a `Vec<>` doesn't improve the output anyway

implementations/rust/ockam/ockam_core/src/flow_control/access_control.rs
davide-baldo: Makes sense :+1:

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
```diff
@@ -93,50 +99,46 @@ impl NodeManagerWorker {
             "Creating inlet portal"
         }
 
+        let flow_controls;
+        {
+            let node_manager = self.node_manager.write().await;
```
SanjoDeundiak: should be read as well

implementations/rust/ockam/ockam_api/src/session/sessions.rs
davide-baldo: Splitting for each field is most likely an overkill, but we can definitively split it up and separate read-only from writable fields. I think most usages are actually derived from some kind of registry and don't even need an async lock, a sync lock or an atomic update will probably suffice.

implementations/rust/ockam/ockam_api/src/util.rs
davide-baldo: Yes, the reasoning was that `secure` never changes the scope to another node by itself, but it's always used in conjunction with `tcp` or other transports. That said, I can't find any reasonable use case to create a secure channel within the node.

implementations/rust/ockam/ockam_core/src/type_tag.rs
davide-baldo: Since only one type of instance can be created, it shouldn't be an issue if the default hash is always emitted.
That said, it's faster and safer to just implement it:
```
state.write_usize(N)
```

implementations/rust/ockam/ockam_api/src/util.rs
davide-baldo: It's probably better to just consider secure remote, since there are many scenarios where this makes sense and zero scenarios where reverse makes sense.

etorreborre (approved):

PR #4475, opened 2023-03-16
handle CTRL-C on foreground node

### LINKED ISSUE #4372: Terminate nodes on SIGTERM
The `SIGTERM` signal is a way to politely ask a program to terminate.

What needs to be done:

Handle the `SIGTERM`  signal on ockam nodes  (both foreground running nodes and background running nodes),  and immediately terminate the process once receiving it.
This will make ockam a better citizen when running in managed environment like docker/kubernetes or as part of some supervised tree.

Using https://crates.io/crates/ctrlc is likely the easiest way to do it,  but feel free to explore other options.  

Note that a better implementation would be to wait for all in-flight work to terminate before exiting,  but that's out of scope for this issue.

---

phillyphil91:
## Current behavior
Foreground nodes are not properly terminated on `CRTL-C`. They remain in `ockam node list`, even after `CTRL-C`.

## Proposed changes
A listener for `CTRL-C` is introduced and handles gracefully shutting down the foreground node including removing the node state.
Addresses Issue: #4372

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -1,12 +1,13 @@
 use clap::Args;
+use colorful::Colorful;
 use ockam_identity::PublicIdentity;
 use ockam_multiaddr::MultiAddr;
 use ockam_vault::Vault;
 use rand::prelude::random;
-use tokio::io::AsyncBufReadExt;
 use tokio::time::{sleep, Duration};
 
 use anyhow::{anyhow, Context as _};
+use std::io::{self, Read};
 use std::{
     net::{IpAddr, SocketAddr},
     path::PathBuf,
@@ -214,7 +215,7 @@ fn create_foreground_node(opts: &CommandGlobalOpts, cmd: &CreateCommand) -> crat
 
 async fn run_foreground_node(
     mut ctx: Context,
-    (opts, cmd, addr): (CommandGlobalOpts, CreateCommand, SocketAddr),
+    (mut opts, cmd, addr): (CommandGlobalOpts, CreateCommand, SocketAddr),
 ) -> crate::Result<()> {
     let cfg = &opts.config;
     let node_name = parse_node_name(&cmd.node_name)?;
@@ -351,8 +352,44 @@ async fn run_foreground_node(
         }
     }
 
+    // Create a channel for communicating back to the main thread
+    let (tx, mut rx) = tokio::sync::mpsc::channel(2);
+
+    // Register a handler for SIGINT, SIGTERM, SIGHUP
+    let tx_clone = tx.clone();
+    let mut opts_clone = opts.clone();
+    ctrlc::set_handler(move || {
+        let _ = tx_clone.blocking_send(());
+        let _ = opts_clone
+            .shell
+            .write_line(format!("{} Ctrl+C signal received", "!".light_yellow()).as_str());
+    })
+    .expect("Error setting Ctrl+C handler");
+
+    // Spawn a thread to monitor STDIN for EOF
     if cmd.exit_on_eof {
-        stop_node_on_eof(&mut ctx, &opts, &node_name).await?;
+        let tx_clone = tx.clone();
+        let mut opts_clone = opts.clone();
+        std::thread::spawn(move || {
+            let mut buffer = Vec::new();
+            let mut handle = io::stdin().lock();
+            handle
+                .read_to_end(&mut buffer)
+                .expect("Error reading from stdin");
+            let _ = tx_clone.blocking_send(());
+            let _ = opts_clone
+                .shell
+                .write_line(format!("{} EOF received", "!".light_yellow()).as_str());
+        });
+    }
+
+    // Shutdown on SIGINT, SIGTERM, SIGHUP or EOF
+    if rx.recv().await.is_some() {
+        opts.state.nodes.get(&node_name)?.kill_process(false)?;
+        ctx.stop().await?;
+        opts.shell
+            .write_line(format!("{}Node stopped successfully", "✔︎".light_green()).as_str())
+            .unwrap();
     }
 
     Ok(())
@@ -371,28 +408,6 @@ pub fn load_pre_trusted_identities(cmd: &CreateCommand) -> Result<Option<PreTrus
         _ => None,
     };
     Ok(pre_trusted_identities)
-}
-
-// Read STDIN until EOF is encountered and then stop the node
-async fn stop_node_on_eof(
-    ctx: &mut Context,
-    opts: &CommandGlobalOpts,
-    node_name: &str,
-) -> crate::Result<()> {
-    let reader = tokio::io::BufReader::new(tokio::io::stdin());
-    let mut lines = reader.lines();
-
-    loop {
-        match lines.next_line().await {
-            Ok(Some(_)) => (),
-            Ok(None) => break,
-            Err(_) => unreachable!(),
-        }
-    }
-
-    ctx.stop().await?;
-    opts.state.nodes.get(node_name)?.kill_process(false)?;
-    Ok(())
 }
 
 async fn start_services(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,7 +176,7 @@ checksum = "0e97ce7de6cf12de5d7226c73f5ba9811622f4db3a5b91b55c53e987e5f91cba"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.11",
+ "syn 2.0.13",
 ]
 
 [[package]]
@@ -187,7 +187,7 @@ checksum = "b9ccdd8f2a161be9bd5c023df56f1b2a0bd1d83872ae53b71a84a12c9bf6e842"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.11",
+ "syn 2.0.13",
 ]
 
 [[package]]
@@ -961,7 +961,7 @@ dependencies = [
  "heck 0.4.1",
  "proc-macro2",
  "quote",
- "syn 2.0.11",
+ "syn 2.0.13",
 ]
 
 [[package]]
@@ -1387,6 +1387,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "049bb91fb4aaf0e3c7efa6cd5ef877dbbbd15b39dad06d9948de4ec8a75761ea"
 dependencies = [
  "cipher",
+]
+
+[[package]]
+name = "ctrlc"
+version = "3.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bbcf33c2a618cbe41ee43ae6e9f2e48368cd9f9db2896f10167d8d762679f639"
+dependencies = [
+ "nix",
+ "windows-sys 0.45.0",
 ]
 
 [[package]]
@@ -1834,22 +1844,22 @@ checksum = "c34f04666d835ff5d62e058c3995147c06f42fe86ff053337632bca83e42702d"
 
 [[package]]
 name = "enumflags2"
-version = "0.7.5"
+version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e75d4cd21b95383444831539909fbb14b9dc3fdceb2a6f5d36577329a1f55ccb"
+checksum = "0044ebdf7fbb2a772e0c0233a9d3173c5cd8af8ae7078d4c5188af44ffffaa4b"
 dependencies = [
  "enumflags2_derive",
 ]
 
 [[package]]
 name = "enumflags2_derive"
-version = "0.7.4"
+version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f58dc3c5e468259f19f2d46304a6b28f1c3d034442e14b322d2b850e36f6d5ae"
+checksum = "9d2c772ccdbdfd1967b4f5d79d17c98ebf92009fdcc838db7aa434462f600c26"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 1.0.109",
+ "syn 2.0.13",
 ]
 
 [[package]]
@@ -1957,12 +1967,12 @@ dependencies = [
 
 [[package]]
 name = "fd-lock"
-version = "3.0.10"
+version = "3.0.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8ef1a30ae415c3a691a4f41afddc2dbcd6d70baf338368d85ebc1e8ed92cedb9"
+checksum = "9799aefb4a2e4a01cc47610b1dd47c18ab13d991f27bbcaed9296f5a53d5cbad"
 dependencies = [
  "cfg-if",
- "rustix 0.36.11",
+ "rustix 0.37.6",
  "windows-sys 0.45.0",
 ]
 
@@ -2148,7 +2158,7 @@ checksum = "89ca545a94061b6365f2c7355b4b32bd20df3ff95f02da9329b34ccc3bd6ee72"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.11",
+ "syn 2.0.13",
 ]
 
 [[package]]
@@ -2604,7 +2614,7 @@ checksum = "256017f749ab3117e93acb91063009e1f1bb56d03965b14c2c8df4eb02c524d8"
 dependencies = [
  "hermit-abi 0.3.1",
  "io-lifetimes",
- "rustix 0.37.4",
+ "rustix 0.37.6",
  "windows-sys 0.45.0",
 ]
 
@@ -2736,9 +2746,9 @@ checksum = "f051f77a7c8e6957c0696eac88f26b0117e54f52d3fc682ab19397a8812846a4"
 
 [[package]]
 name = "linux-raw-sys"
-version = "0.3.0"
+version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cd550e73688e6d578f0ac2119e32b797a327631a42f9433e59d02e139c8df60d"
+checksum = "d59d8c75012853d2e872fb56bc8a2e53718e2cafe1a4c823143141c6d90c322f"
 
 [[package]]
 name = "lmdb-rkv"
@@ -3305,6 +3315,7 @@ dependencies = [
  "console",
  "const-str",
  "crossbeam-channel",
+ "ctrlc",
  "dialoguer",
  "directories",
  "dirs",
@@ -3674,9 +3685,9 @@ dependencies = [
 
 [[package]]
 name = "openssl"
-version = "0.10.48"
+version = "0.10.49"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "518915b97df115dd36109bfa429a48b8f737bd05508cf9588977b599648926d2"
+checksum = "4d2f106ab837a24e03672c59b1239669a0596406ff657c3c0835b6b7f0f35a33"
 dependencies = [
  "bitflags 1.3.2",
  "cfg-if",
@@ -3689,13 +3700,13 @@ dependencies = [
 
 [[package]]
 name = "openssl-macros"
-version = "0.1.0"
+version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b501e44f11665960c7e7fcf062c7d96a14ade4aa98116c004b2e37b5be7d736c"
+checksum = "a948666b637a0f465e8564c73e89d4dde00d72d4d473cc972f390fc3dcee7d9c"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 1.0.109",
+ "syn 2.0.13",
 ]
 
 [[package]]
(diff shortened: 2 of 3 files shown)


phillyphil91:
Hi.
A couple of notes/questions from my side.

Instead of using the `ctrlc` crate, I found that `tokio` already provides similar functionality with `tokio::signal::ctrl_c` if I understood the workings of both solutions correctly. Both allow for a `CTRL-C` handler to be created. `ctrlc` apparently does this via a dedicated thread. So maybe using the `tokio` version is more close to what we want and part of the dependencies already anyways.

I hope I understood the issue correct that `CTRL-C` previously caused the node to remain in the `ockam node list` ouput even though it was run as a foreground node and thus should not? Now if you start a foreground node and do `CTRL-C` is disappears and leaves no trace in `ockam node list`. This was not the case before. Of course this only handles `CTRC-C` which I think is `SIGINT` at least on unix system. So it probably doesn't handle `SIGTERM` correctly in general.

@adrianbenavides  you mentioned to call the `NodeState::delete()` function after catching `CTRL-C`. I believe `opts.state.nodes.delete(&node_name, false)?` does this.

Regarding `let _ = tokio::spawn(stop_foreground_node(ctx, opts, node_name)).await?;` I was not sure whether `stop_foreground_node()` really needs to be passed to `tokio::spawn()`. On the other hand if I just `await` `stop_foreground_node()` directly or even move the function body directly into `run_foreground_node()` then this would block until `CTRL-C` is caught and thus never reach the `OK(())` part right? Not sure if that's actually what happens and if that's what we want. 
Thanks for help on this. I know it's not much of a great change yet.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -351,6 +351,20 @@ async fn run_foreground_node(
         stop_node_on_eof(&mut ctx, &opts, &node_name).await?;
     }
 
+    let _ = tokio::spawn(stop_foreground_node(ctx, opts, node_name)).await?;
```
adrianbenavides: Whether you use tokio or ctrlc to handle the signals, they can be handled in a separate thread and run in the background... or handled in the main thread blocking its execution. What approach to use will depend on the use case.

With tokio, if you `await` the spawned task, you will stop the execution there until the inner future completes. So, you just want to spawn the thread and continue: `tokio::spawn(stop_foreground_node(ctx, opts, node_name));`. In this case, you can get away without communicating back to the main thread because you can stop the context from wherever you want as long as you are in an async runtime.

With `ctrlc`, the handler is a simple closure (non-async), so you have to communicate back to the main thread where you will be able to stop the context. A side effect of this is that you should run the `stop_node_on_eof` in a background tokio task and set up the ctrlc signal in the main thread, similar to how it's shown in the [docs example](https://docs.rs/ctrlc/latest/ctrlc/#example). The reason of this is explained on the other comment.

And yeah, this issue doesn't require too many lines, but it's full of hidden details and traps 😆

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -351,6 +351,20 @@ async fn run_foreground_node(
         stop_node_on_eof(&mut ctx, &opts, &node_name).await?;
     }
```
adrianbenavides, line 384: This `stop_node_on_eof` function runs a loop and blocks the execution until it receives an EOF character, so the signal handler should be registered before this block.

We have to support these two loops at the same time and exit on the first one that happens, so we can only block the main thread with only one of the loops.

adrianbenavides:
First off, thank you so much for taking a look into this!


> Instead of using the `ctrlc` crate, I found that `tokio` already provides similar functionality with `tokio::signal::ctrl_c`

Yeah, they are equivalent but `ctrlc` support SIGINT, SIGTERM and SIGHUP by default. I think it w

> I hope I understood the issue correct that `CTRL-C` previously caused the node to remain in the `ockam node list` ouput even though it was run as a foreground node and thus should not? Now if you start a foreground node and do `CTRL-C` is disappears and leaves no trace in `ockam node list`. This was not the case before. Of course this only handles `CTRC-C` which I think is `SIGINT` at least on unix system. So it probably doesn't handle `SIGTERM` correctly in general.

It might worked on your example but the signal handler is not set up correctly. I'll leave a comment explaining how to use it properly.

> @adrianbenavides you mentioned to call the `NodeState::delete()` function after catching `CTRL-C`. I believe `opts.state.nodes.delete(&node_name, false)?` does this.

Actually, now that I think of it, I think the correct thing to do is to stop the node, rather than delete it. You can do that with:

```rust
ctx.stop().await?;
opts.state.nodes.get(node_name)?.kill_process(false)?;
```

> Regarding `let _ = tokio::spawn(stop_foreground_node(ctx, opts, node_name)).await?;` I was not sure whether `stop_foreground_node()` really needs to be passed to `tokio::spawn()`. On the other hand if I just `await` `stop_foreground_node()` directly or even move the function body directly into `run_foreground_node()` then this would block until `CTRL-C` is caught and thus never reach the `OK(())` part right? Not sure if that's actually what happens and if that's what we want. Thanks for help on this. I know it's not much of a great change yet.

I'll leave a comment about this.

adrianbenavides:
Finally, one way to test this would be to run a foreground node with the `--exit-on-eof` argument so you set up the two handlers. Then you can try stopping the node with either of the signals supported and verify that the node is stopped correctly.

phillyphil91:
Hi @adrianbenavides,

once again, thanks for the helpful comments. I thought I had it, but then realized I didn't :smile: 

So the `ctrlc` crate is more useful for us here because it handles `SIGTERM`, `SIGINT`, etc. But other than that has no advantage over `tokio::signal::ctrl_c`?

I tried to implement everything according to the `ctrlc` example which includes running a loop that checks if `signal_terminate` is set to `true` and then run the node stop/ kill process.

This seems to work when only `--foreground` is set. Once `--foreground` and `--exit-on-eof` are set this does not work anymore. The node is still stopped, but you have to manually confirm `CTRL-C`. Sending `SIGTERM` also stops the node but the process is still running etc.

I figured since the monitoring of `EOF` and `CTRL-C`, etc. requires looping it would make sense to spawn both them in a new `tokio task`. And then either could run the `ctx.stop()` and `kill_process()` functions and then everything would terminate. 
But this doesn't seem to be the case and I am not sure why. It seems like either one of the two or the main thread is still running even after one of them has successfully executed. Maybe it's neither of the two tasks can actually kill the entire process if the other one is still running (possibly in another thread)?

Tried just sending one of to a new spawned tasks and await the other at the end, but this didn't seem to really help either.

phillyphil91:
Maybe the solution is to use `tokio::task::spwan_blocking` or `std::threat::spawn` because both functions are blocking indeed.

With `tokio::task::spwan_blocking` `CTRL-C` at least seems to work as expected, but `EOF` does not.

I have to take another look.

phillyphil91:
Okay so I think the current version works, even though I am not entirely sure how.

`stop_foreground_node()` now is in wrapped in `spawn_blocking` and `stop_node_on_eof()` now just blocks the main thread at the end. 

I used the approach with the `Arc` because `ctx` doesn't implement `Clone`. If it did I could have just cloned the arguments to `stop_foreground_node()` in the closure I believe.

adrianbenavides:
I haven't gone through the PR in detail but given how complex is this getting I'd recommend using an mpsc channel so you can send the `stop node` signal back to the main thread from background tasks (i.e. the ctrlc and the eof functions). The main advantage of this approach is that you won't have to pass the `Context` to the spawned threads, and therefore, you won't have to deal with the side effects of cloning the context.

Let me know if you need more context on how to do that and I'll provide a detailed example.

phillyphil91:
> I haven't gone through the PR in detail but given how complex is this getting I'd recommend using an mpsc channel so you can send the `stop node` signal back to the main thread from background tasks (i.e. the ctrlc and the eof functions). The main advantage of this approach is that you won't have to pass the `Context` to the spawned threads, and therefore, you won't have to deal with the side effects of cloning the context.
> 
> Let me know if you need more context on how to do that and I'll provide a detailed example.

Thanks. I will try to solve this using an mpsc channel. I will get back to you if I have any questions.

The current version is not suitable yet?

adrianbenavides:
> The current version is not suitable yet?

It's a valid solution, but I think it's worth exploring other simpler alternatives. The code in the `ockam_command` crate is usually the first contact with the ockam repository for most of the people that contributes to it for the first time. It's also common that those contributors have never written any rust code before. If I put myself in that situation, I know for sure this code would blow my mind 😆

phillyphil91:
> > The current version is not suitable yet?
> 
> It's a valid solution, but I think it's worth exploring other simpler alternatives. The code in the `ockam_command` crate is usually the first contact with the ockam repository for most of the people that contributes to it for the first time. It's also common that those contributors have never written any rust code before. If I put myself in that situation, I know for sure this code would blow my mind :laughing: 

That makes sense. Plus to be fair I am not even sure how this one works but sending both to `tokio::task::spawn_blocking` does not. :confused: 
I will have a look at the mpsc channel alternative.

phillyphil91:
Hi @adrianbenavides 

Over the weekend I tried to implement this in an `async` fashion in every possible way under the sun, I believe :sweat_smile: 
But any combination with `tokio::spawn` or `tokio::task::spawn_blocking` did not work and always left the program hanging. Even when receiving the appropriate message on the channel or reaching the final `Ok(())`.

It is working with dedicated OS threads as in the current version. I am not sure if that's okay giving the environments that Ockam is supposed to run in. I still used `tokio::sync::mpsc::channel()` to not block the `run_foreground_node` function.

I removed the `stop_node_on_eof` function and integrated that straight into `run_foreground_node` in a slightly different fashion. Let me know what you think.

P.s. I still have to clean up the other files that I changed for a previous approach

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -347,32 +346,34 @@ async fn run_foreground_node(
...
+
+    // Spawn a thread to monitor STDIN for EOF
+    let tx_clone = tx.clone();
+    std::thread::spawn(move || {
+        let mut buf = String::new();
+        match io::stdin().read_to_string(&mut buf) {
+            Ok(_) => tx_clone.blocking_send(1).unwrap(),
+            Err(e) => eprintln!("Error reading STDIN: {}", e),
+        };
+    });
```
adrianbenavides: This should only be executed `if cmd.exit_on_eof`. Also, the logic is wrong as it will exit upon any read line.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -347,32 +346,34 @@ async fn run_foreground_node(
...
-
-    Ok(())
-}
+    // Create a channel for communicating back to the main thread
+    let (tx, mut rx) = tokio::sync::mpsc::channel(2);
+
+    // Register a handler for the Ctrl+C signal
+    let tx_clone = tx.clone();
+    ctrlc::set_handler(move || {
+        tx_clone.blocking_send(1).unwrap();
```
adrianbenavides: If you want to return a dummy value you can return the unit `()` value.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -347,32 +346,34 @@ async fn run_foreground_node(
...
-    let mut lines = reader.lines();
-
-    loop {
-        match lines.next_line().await {
-            Ok(Some(_)) => (),
-            Ok(None) => break,
-            Err(_) => unreachable!(),
+    if cmd.foreground || cmd.exit_on_eof {
+        // Shutdown on SIGINT, SIGTERM, SIGHUP or EOF
+        if let Some(_) = rx.recv().await {
```
adrianbenavides: The `rx.recv().await` must be always checked. In the default case, you will register the ctrlc signal and you want to wait for that. So, the `if cmd.foreground || cmd.exit_on_eof` should not be there.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -347,32 +346,34 @@ async fn run_foreground_node(
...
+            ctx.stop().await?;
+            opts.state.nodes.get(&node_name)?.kill_process(false)?;
```
adrianbenavides: I'd switch these two lines. Stopping the context should be the last operation to run before exiting an ockam runtime.

adrianbenavides (commented):
I've been playing around a little with this and I've managed to make it work with the changes I've described below. I can push my changes if you get stuck, but I think you are pretty close now! 💪

implementations/rust/ockam/ockam_command/src/node/create.rs
adrianbenavides: Here is an example of how it could look:

```rust
let mut buffer = Vec::new();
let mut handle = io::stdin().lock();
handle
    .read_to_end(&mut buffer)
    .expect("Error reading from stdin");
info!("! EOF received");
let _ = tx_clone.blocking_send(());
```

adrianbenavides: If it does what it has to do, pressing `ctrl+d` should trigger the EOF handler.

implementations/rust/ockam/ockam_command/src/node/create.rs
phillyphil91: I thought the default case was to not watch either: `CTRL-C `and `EOF`. but only do so when `--foreground` is set (`CTRL-C`) and of course when `--exit-on-eof` is set (`EOF`).

So my idea was to run both transmission channels and then run `recv()` only when either `--foreground` or `--exit-on-eof` is set. 
But I am just realizing that then `EOF` would also always be monitored even if just `--foreground` is set. Which is wrong.

If `ctrlc` should always be handled, I will change it to that.

implementations/rust/ockam/ockam_command/src/node/create.rs
phillyphil91: I can change that to your version, but I thought since `read_to_string()` says that it 

```
Read(s) all bytes until EOF in this source, appending them to buf.

If successful, this function returns the number of bytes which were read and appended to buf.
```

I figured that it would jut continue to read `STDIN` until it receives `EOF` from somewhere and not exit on any line?
Or is not calling `.lock()` the issue here?

implementations/rust/ockam/ockam_command/src/node/create.rs
adrianbenavides: Note that we are already in the `run_foreground_node` function, so we can assume `--foreground` is passed. Does that make sense?

phillyphil91: of course it does. I totally overlooked the context of the function. sorry, that makes total sense

implementations/rust/ockam/ockam_command/src/node/create.rs
adrianbenavides: mm good point. To be honest I'm not sure of the differences between the two methods for our use-case. You can keep your version and try if it works as expected. If it reacts to `ctrl+d` then we are good.

implementations/rust/ockam/ockam_command/src/node/create.rs
adrianbenavides: no worries, I also had to double-check this one 😆

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -347,32 +346,36 @@ async fn run_foreground_node(
         }
     }
 
+    // Create a channel for communicating back to the main thread
+    let (tx, mut rx) = tokio::sync::mpsc::channel(2);
+
+    // Register a handler for SIGINT, SIGTERM, SIGHUP
+    let tx_clone = tx.clone();
+    ctrlc::set_handler(move || {
+        tx_clone.blocking_send(()).unwrap();
```
adrianbenavides: Add message here:

```rust
opts.shell.write_line(format!(
    "{} Ctrl+C signal received",
    "!".light_yellow(),
    &cmd.node_name
))?;
```

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -347,32 +346,36 @@ async fn run_foreground_node(
...
     if cmd.exit_on_eof {
-        stop_node_on_eof(&mut ctx, &opts, &node_name).await?;
+        let tx_clone = tx.clone();
+        std::thread::spawn(move || {
+            let mut buffer = Vec::new();
+            let mut handle = io::stdin().lock();
+            handle
+                .read_to_end(&mut buffer)
+                .expect("Error reading from stdin");
+            info!("EOF received!");
```
adrianbenavides: Change this message to:

```rust
opts.shell.write_line(format!(
    "{} EOF received",
    "!".light_yellow(),
    &cmd.node_name
))?;
```

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -347,32 +346,36 @@ async fn run_foreground_node(
...
-    loop {
-        match lines.next_line().await {
-            Ok(Some(_)) => (),
-            Ok(None) => break,
-            Err(_) => unreachable!(),
-        }
+    // Shutdown on SIGINT, SIGTERM, SIGHUP or EOF
+    if let Some(_) = rx.recv().await {
+        opts.state.nodes.get(&node_name)?.kill_process(false)?;
+        ctx.stop().await?;
```
adrianbenavides, line 389: Add message:

```rust
opts.shell.write_line(format!(
    "{}Node stopped successfully",
    "✔︎".light_green(),
    &cmd.node_name
))?;
```

adrianbenavides (commented):
Looks like it's working now perfectly! There are a couple of minor things before we can merge it:
- Rebase on top of latest develop
- Print some messages to inform the user of what's going on. Once you have the changes from develop, you will be able to print the messages as pointed out in the comments

Once you are done with that, you will need to:
- squash all commits
- write a valid commit message, for example: `feat(rust): handle ctrlc signal on foreground nodes`

implementations/rust/ockam/ockam_command/src/node/create.rs
phillyphil91: Had to modify this a bit, but works now, hopefully as you would expect. I wasn't sure if we want the node_name printed as well. let me know.

Also I just unwraped the `shell.write_line()`. Is this okay? Or rather `expect()` with a more meaningful message?

implementations/rust/ockam/ockam_command/src/node/create.rs
phillyphil91: I think mine was working too, but I changed it to yours because  the function name `read_to_end()` is more closely related to what we actually want to do :smile:

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -352,8 +353,46 @@ async fn run_foreground_node(
...
+            let _ = tx_clone.blocking_send(());
+        });
+    }
+
+    // Shutdown on SIGINT, SIGTERM, SIGHUP or EOF
+    if let Some(_) = rx.recv().await {
+        opts.state.nodes.get(&node_name)?.kill_process(false)?;
+        ctx.stop().await?;
+        opts.shell
+            .write_line(format!("{} Node stopped successfully", "✔︎".light_green()).as_str())
```
adrianbenavides: The "✔︎" includes a trailing space at the end, so it should be `{}Node`.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -352,8 +353,46 @@ async fn run_foreground_node(
...
+            let mut buffer = Vec::new();
+            let mut handle = io::stdin().lock();
+            handle
+                .read_to_end(&mut buffer)
+                .expect("Error reading from stdin");
+            opts_clone
+                .shell
+                .write_line(format!("{} EOF received", "!".light_yellow()).as_str())
+                .unwrap();
+            let _ = tx_clone.blocking_send(());
```
adrianbenavides: Let's send the message first and then log the message. In this context, we can ignore the error as it can't be recovered and we don't care if we can't write to stderr:

```rust
let _ = tx_clone.blocking_send(());
let _ = opts_clone.shell.write(...);
```

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -352,8 +353,46 @@ async fn run_foreground_node(
...
+
+    // Register a handler for SIGINT, SIGTERM, SIGHUP
+    let tx_clone = tx.clone();
+    let mut opts_clone = opts.clone();
+    ctrlc::set_handler(move || {
+        tx_clone.blocking_send(()).unwrap();
+        opts_clone
+            .shell
+            .write_line(format!("{} Ctrl+C signal received", "!".light_yellow()).as_str())
+            .unwrap();
```
adrianbenavides: Same here about ignoring the error.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -352,8 +353,46 @@ async fn run_foreground_node(
...
+            opts_clone
+                .shell
+                .write_line(format!("{} EOF received", "!".light_yellow()).as_str())
+                .unwrap();
+            let _ = tx_clone.blocking_send(());
+        });
+    }
+
+    // Shutdown on SIGINT, SIGTERM, SIGHUP or EOF
+    if let Some(_) = rx.recv().await {
```
adrianbenavides: There is a clippy warning left to be fixed here.

implementations/rust/ockam/ockam_command/src/node/create.rs
adrianbenavides: >  I wasn't sure if we want the node_name printed as well. let me know.

Sorry, copy-paste error. No need to show the node_name. I added a separate comment about the unwrapped `write_line()`.

adrianbenavides (approved):
Ready to merge, thanks @phillyphil91 for another great contribution! 💪

metaclips (approved):

phillyphil91:
thank you @adrianbenavides for your help again.

PR #4482, opened 2023-03-17
feat(rust): add examples and about sections to markdown generated docs

adrianbenavides:
This PR reorganizes the commands' documentation so we can better leverage clap tools to handle docs. The main introduced change is that we now have the command documentation split into two sections: `long about` to give an extended description of the command and `after_long_help` to show how to use the command (the examples section).

The documentation has the following structure now:
```
{about}
{usage}
{subcommands, arguments}
{options, global options}
{examples}
{learn more, feedback}
```

Another important change that this PR introduces is that now the markdown-generated docs have a more enriched structure, rendering the examples section as well as both the long and short "about" sections. The markdown version doesn't include the `{learn more, feedback}` section.


mrinalwadhwa:
@adrianbenavides this looks great.
Since this is touching all the help txt files. Let's move them out of the constant folder into their respective command folders. There is no reason to centralize these in one folder.

adrianbenavides:
> @adrianbenavides this looks great. Since this is touching all the help txt files. Let's move them out of the constant folder into their respective command folders. There is no reason to centralize these in one folder.

We are duplicating the whole commands folder structure, so that makes total sense.

davide-baldo (approved):

Issue #2496, opened 2022-02-11
Changing the binary message format?


twittner:
Currently Ockam messages are encoded with [BARE][BARE-RFC], a binary data format that optimises for compactness and simplicity. It does that by requiring external schema information which allows it to not encode a lot of structural information. Apart from actual data such as string values or numbers it only adds some length information where necessary, but not much else. Insofar the format has only minimal overhead. The flip side of this approach is that without the correct schema, BARE messages can not be decoded. This prevents generic message processing and makes it difficult to have support for forward and backward compatibility[^1] as it is not possible to skip over message parts without knowing their meaning. The ambiguities that may arise surface in functions like [`ockam_node::parser::message`][fallback] which retries message decoding should the first decoding attempt fail, by *modifying* the message. The appendix below contains further examples that highlight the importance of accurate schema information.

[^1]: Section 4 of the [RFC draft][BARE-RFC] of BARE mentions schema evolution in a backward and forward compatible way by using tagged unions with explicit tag numbers. While adding another tagged branch is backward compatible because newer software understands the messages of older software, the reverse is not true: a message using this new tag will be rejected by older software, hence forward compatibility is not realised.

Many formats work with schema information and it is a great way to save precious bytes, however it would be nice if in addition, messages have enough embedded structure so that forward and backward compatibility is possible and maybe also generic processing which can help with debugging or logging.

Are there alternatives to BARE which offer this without sacrificing any of the other important requirements, i.e.

1. a compact, simple and language agnostic format, with
2. simple codecs that have minimal resource requirements, which are capable of
3. safe message processing that handles malformed input gracefully without causing resource exhaustion or violating memory safety?

I would like to argue for [CBOR][CBOR] here which is not as compact as BARE, but should otherwise offer all of this.

In contrast to BARE, CBOR does encode more structural information which increases message size. Every CBOR data item is prefixed with a byte that encodes the item type as well as additional information (mostly about how to load the data, e.g. length information). This enables generic processing of arbitrary data items, for example to map them to a human readable format. This also makes it possible to skip over message parts which in turn allows for forward and backward compatibility. The interesting question is how much overhead does this cost in practice? In https://github.com/twittner/encoding-size I attempt to do a comparison between BARE and CBOR. As data I use randomly generated trees of structs and enums with additional data items such as strings and numbers. Here is an example run:

```
>>> TINY
BARE = 53 bytes ( 82.81%)
CBOR = 64 bytes (120.75%)
>>> SMALL
[...]

PR #4489, opened 2023-03-18
feat(rust): add rendezvous service for nat hole punching

neil2468:
## Proposed changes

This is part of the prototype for UDP NAT Hole Punching (#3507).

- Add a 'Rendezvous service' to the UDP transport.
- Add tests to exercise new code.

Thanks!

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_transport_udp/src/rendezvous_service/messages.rs
+++ b/implementations/rust/ockam/ockam_transport_udp/src/rendezvous_service/messages.rs
@@ -1,0 +1,24 @@
+use std::net::SocketAddr;
+
+use ockam_core::{Message, Result};
+use serde::{Deserialize, Serialize};
+
+/// Request type for UDP Hole Punching Rendezvous service
+#[derive(Serialize, Deserialize, Debug, Message)]
+pub enum RendezvousRequest {
+    Update {
+        /// Name of sending node
+        node_name: String,
+    },
+    Query {
+        /// Name of node to lookup
+        node_name: String,
+    },
+}
+
+/// Response type for UDP Hole Punching Rendezvous service
+#[derive(Serialize, Deserialize, Debug, Message)]
+pub enum RendezvousResponse {
+    Update(Result<()>),
+    Query(Result<SocketAddr>),
+}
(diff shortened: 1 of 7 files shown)


implementations/rust/ockam/ockam_transport_udp/src/rendezvous_service/messages.rs
```diff
@@ -0,0 +1,24 @@
...
+        node_name: String,
+    },
+}
+
+/// Response type for UDP Hole Punching Rendezvous service
+#[derive(Serialize, Deserialize, Debug, Message)]
+pub enum RendezvousResponse {
+    Update(Result<()>),
+    Query(Result<SocketAddr>),
+}
```
mrinalwadhwa, line 24: This looks like a tiny protocol.

Leaving this comment just to note that we don't want the message serialization to be coupled with Rust's default binary encoding, so we should turn this into CBOR messages.

mrinalwadhwa:
@neil2468 This looks awesome. Did you happen to test it across NATs somehow?

neil2468:
@mrinalwadhwa Thanks. I've not yet tried setting up an evironment for NAT testing, but now that you mention it maybe I should do that next. 

This PR does not implement hole punching yet, it's only adds the Rendezvous service which will maintain a table of the public IPs of remote nodes.

davide-baldo (dismissed):
Looks good!

implementations/rust/ockam/ockam_transport_udp/src/rendezvous_service/messages.rs
neil2468: @mrinalwadhwa Are you saying that this should change to use the `minicbor` encoding and `ockam_core::api::Request` like `CredentialExchangeWorker` does?

So I can learn, can I ask why would we like this to use CBOR while other messages (e.g. `PortalMessage` used by `TcpOutletListenWorker`) use Rust's default encoding?

Thanks!

davide-baldo (approved):

implementations/rust/ockam/ockam_transport_udp/src/rendezvous_service/messages.rs
etorreborre: @neil2468 I think that we are supposed to move everything to minicbor now. I am going to raise this issue.

etorreborre: @neil2468 I actually still need to come back to you with a better answer. The question of migrating ockam messages from BARE to CBOR [is still open](https://github.com/build-trust/ockam/issues/2496). I think that what @mrinalwadhwa meant was that for workers implementing a request/response protocol, they should follow the pattern established by workers like the `CredentialExchangeWorker` and use the `Request` API present in `ockam_core`. I will come back with a confirmation.

PR #4493, opened 2023-03-20
Add an `ockam authority create` command

etorreborre:
This PR adds a new command `ockam authority create` to create an Authority node.

An authority node is a node which only starts authentication services (credentials, enrollment tokens, authentification).
Those services are only accessible by first creating a secure channel.

When the Orchestrator has migrated to this command it will be possible to remove the [special casing of the current `ockam node create` command for authority nodes](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/node/create.rs#L220).

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/authority/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/authority/create.rs
@@ -79,6 +79,7 @@ impl CreateCommand {
     }
 }
 
+/// Given a Context start a node in a new OS process
 async fn create_background_node(
     ctx: Context,
     (opts, cmd): (CommandGlobalOpts, CreateCommand),
@@ -87,6 +88,8 @@ async fn create_background_node(
     spawn_background_node(&ctx, &opts, &cmd).await
 }
 
+/// Start an authority node by calling the `ockam` executable with the current command-line
+/// arguments
 async fn spawn_background_node(
     ctx: &Context,
     opts: &CommandGlobalOpts,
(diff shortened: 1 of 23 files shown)


implementations/rust/ockam/ockam_command/src/authority/create.rs
```diff
@@ -0,0 +1,261 @@
...
+            )?),
+    )?;
+
+    let configuration = authority_node::Configuration {
+        identity: public_identity,
+        storage_path: options.state.identities.authenticated_storage_path()?,
+        vault_path: options.state.vaults.default()?.vault_file_path()?,
+        project_identifier: command.project_identifier,
+        tcp_listener_address: command.tcp_listener_address.clone(),
+        secure_channel_listener_name: None,
```
SanjoDeundiak, line 251: We run a secure channel listener, is it intended to have `None` here?

etorreborre: This is just for the name of the secure channel listener service. I haven't left the possibility to customize it on the command line, so this is why there is `None` here which means the authority node will use `DefaultAddress::SECURE_CHANNEL_LISTENER` when starting the listener. If we decide to make this configurable from the command line, this will become a `Some(_)`

SanjoDeundiak (dismissed):

polvorin (approved):
Good one!.

There are a couple things but that can be addressed separately in future  when/if needed:

* We have a way to start/not start the okta addon.      Same should be done for the "direct"  and "token enrollment"  ones,   one might decide for example that doesn't want the token enrollment to be available on his authority.    Even no authenticator at all should be possible  (authority would only produce credentials for the identities on the trust anchors, or identities _already_ added but no new one will be added).   The only thing that should always be there to be a proper authority node is the credential issuer

* The format for trust anchors is different if passed as a file  (a map)  than if passed as direct argument  (an array of objects).   I feel a map makes more sense here,  don't need to worry about same identity appearing multiple times.

PR #4498, opened 2023-03-20
refactor(rust): refactored channel decryptor state to the most normal…

davide-baldo:
## Current behavior

From the decryptor state, it isn't clear which fields are used in which phase. If an error occurs during secure channel construction, a half-modified state is kept.

## Proposed changes

Every step has explicit field dependencies expressed in the structures, if an error occurs during secure channel construction, the worker state is lost, becoming inoperable.
Some of the logic was also made more explicit by splitting the identity exchange.

### DIFF
--- a/implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
+++ b/implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
@@ -2,13 +2,16 @@ use crate::api::{DecryptionRequest, DecryptionResponse};
 use crate::channel::addresses::Addresses;
 use crate::channel::common::{AuthenticationConfirmation, CreateResponderChannelMessage, Role};
 use crate::channel::decryptor::Decryptor;
-use crate::channel::decryptor_state::{ExchangeIdentity, Initialized, KeyExchange, State};
+use crate::channel::decryptor_state::{
+    IdentityExchangeState, InitializedState, KeyExchangeState, State,
+};
 use crate::channel::encryptor::Encryptor;
 use crate::channel::encryptor_worker::EncryptorWorker;
 use crate::channel::messages::IdentityChannelMessage;
 use crate::{
-    to_symmetric_vault, to_xx_vault, Identity, IdentityError, IdentitySecureChannelLocalInfo,
-    PublicIdentity, SecureChannelRegistryEntry, SecureChannelTrustInfo, TrustPolicy,
+    to_symmetric_vault, to_xx_vault, Identity, IdentityError, IdentityIdentifier,
+    IdentitySecureChannelLocalInfo, PublicIdentity, SecureChannelRegistryEntry,
+    SecureChannelTrustInfo, TrustPolicy,
 };
 use core::time::Duration;
 use ockam_core::compat::vec::Vec;
@@ -28,17 +31,7 @@ use ockam_node::{Context, MessageReceiveOptions, WorkerBuilder};
 use tracing::{debug, info, warn};
 
 pub(crate) struct DecryptorWorker {
-    role: Role,
-    addresses: Addresses,
-    // Route to the other side of the channel
-    remote_route: Route,
-    remote_backwards_compatibility_address: Option<Address>,
-    init_payload: Option<Vec<u8>>,
-    identity: Identity,
-    trust_policy: Arc<dyn TrustPolicy>,
-    state_key_exchange: Option<KeyExchange>,
-    state_exchange_identity: Option<ExchangeIdentity>,
-    state_initialized: Option<Initialized>,
+    state: Option<State>,
 }
 
 impl DecryptorWorker {
@@ -67,18 +60,16 @@ impl DecryptorWorker {
         let mailboxes = Self::mailboxes(&addresses, decryptor_outgoing_access_control);
 
         let worker = DecryptorWorker {
-            role: Role::Initiator,
-            addresses: addresses.clone(),
-            remote_route,
-            remote_backwards_compatibility_address: None,
-            init_payload: None,
-            identity,
-            trust_policy,
-            state_key_exchange: Some(KeyExchange {
-                key_exchanger: Box::new(key_exchanger),
-            }),
-            state_exchange_identity: None,
-            state_initialized: None,
+            state: Some(State::new(
+                Role::Initiator,
+                identity,
+                addresses.clone(),
+                Box::new(key_exchanger),
+                remote_route,
+                trust_policy,
+                None,
+                None,
+            )),
         };
 
         WorkerBuilder::with_mailboxes(mailboxes, worker)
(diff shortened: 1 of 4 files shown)


implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
```diff
@@ -670,31 +627,57 @@ impl Worker for DecryptorWorker {
...
+
+        // if any error occurs during key and identity exchange the state will become
+        // invalid, this enforces atomicity reducing the possibility of attacks
+
+        let new_state = match state {
+            State::KeyExchange(state) => {
+                if msg_addr == state.addresses.decryptor_remote {
+                    state.handle_key_exchange_msg(ctx, msg).await?
                 } else {
+                    // let's ignore invalid address messages errors
```
etorreborre, line 658: Let's log them instead?

implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
```diff
@@ -670,31 +627,57 @@ impl Worker for DecryptorWorker {
...
-        match self.state()? {
-            State::KeyExchange => {
-                if msg_addr == self.addresses.decryptor_remote {
-                    self.handle_key_exchange_msg(ctx, msg).await?;
-                } else {
-                    return Err(IdentityError::UnknownChannelMsgDestination.into());
-                }
+        // since once initialized, the state can't change anymore we just borrow a mutable,
+        // doing so we avoid extra copies (in case the compiler don't optimize them away),
+        // and we make sure the state remain initialized even in the case of an error
```
etorreborre, line 622: The drawaback is that now we don't have a full pattern match on [L.676](https://github.com/build-trust/ockam/pull/4498/files#diff-9faefb775a440129e0a51b66c188674bd94071bed6d24095bdf7439be705d90dR676) :-(.

implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
```diff
@@ -670,31 +627,57 @@ impl Worker for DecryptorWorker {
...
+            return Ok(());
+        }
+
+        let state = self
+            .state
+            .take()
+            .ok_or_else(|| IdentityError::InvalidSecureChannelInternalState)?;
+
+        // if any error occurs during key and identity exchange the state will become
+        // invalid, this enforces atomicity reducing the possibility of attacks
```
etorreborre, line 645: Shouldn't we close the context and kill the worker in that case?

implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
```diff
@@ -29,18 +32,7 @@ use ockam_node::{Context, WorkerBuilder};
...
-    remote_route: Route,
-    remote_backwards_compatibility_address: Option<Address>,
-    init_payload: Option<Vec<u8>>,
-    identity: Identity,
-    trust_policy: Arc<dyn TrustPolicy>,
-    state_key_exchange: Option<KeyExchange>,
-    state_exchange_identity: Option<ExchangeIdentity>,
-    state_initialized: Option<Initialized>,
-    plaintext_session_id: Option<SessionId>,
+    state: Option<State>,
```
etorreborre, line 34: Isn't the `state` always `Some`? A `Decryptor` should always have some state, isn't it?

implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
```diff
@@ -220,51 +208,41 @@ impl DecryptorWorker {
...
         // Key exchange hasn't been completed -> generate and send next request
-        if !state.key_exchanger.is_complete().await? {
+        if !self.key_exchanger.is_complete().await? {
             request_was_sent = true;
-            let payload = state.key_exchanger.generate_request(&[]).await?;
+            let payload = self.key_exchanger.generate_request(&[]).await?;
 
             // We should send first_responder_address only with first message from the initiator
-            let custom_payload = if self.role.is_initiator() && first_run {
+            let custom_payload = if self.role.is_initiator() && initialization_run {
```
etorreborre: I think that we should even be able to move the `initialization_run` piece of state into the `KeyExchangeState` and not pass it around.

etorreborre (dismissed):
💯 this is excellent! That's actually what I had in mind where `State` becomes the whole state machine

implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
davide-baldo: This is actually needed in order to allow a "switch" from one state to the other: during the passage we need to move some instances and a "invalid state" need to exist, either inside the state or as a None

SanjoDeundiak: @davide-baldo I encountered the same problem when implementing it previous time

etorreborre: Why not encode an explicit `Invalid` state in the `State` enum then?

SanjoDeundiak: I suspect that would be hard

etorreborre: I'll give it a go at some stage. I still hope that we can make the states and state transitions self-contained in some way.

davide-baldo: I did implement it with an `Invalid` state within State in the beginning, I simply thought I was more elegant using an Option.
I can use `alloc::mem::replace(state,Invalid)`, create the new state by moving instances, and then update the state.

implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
davide-baldo: The error returned by `handle_message` is already logged by the relay

implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
davide-baldo: We could avoid it, this is just an optimization. We could compensate for it by boxing the states.

implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
davide-baldo: The second `handle_message()` after an error may or may not happen, when you are debugging, you want a consistent behavior: either the worker stops on errors or it doesn't.
maybe it's better to stop the worker **before** it returns the error, so this state should become an `unreachable()`

davide-baldo: I've made sure the worker stops when it reaches a broken state

implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
davide-baldo: To make it more explicit, I've replaced the returned error with an `unreachable!`

implementations/rust/ockam/ockam_identity/src/channel/decryptor_worker.rs
etorreborre: I don't really understand that part "create the new state by moving instances, and then update the state". What is 'moving' and 'instance' in that context? But let's move ahead, we can discuss it online later.

davide-baldo: I've written a small sample of what I did with the `Invalid`, as you can see, it's far from being elegant :(

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f0896128920c1aa65f296be85ddd8459

etorreborre (approved):

PR #4505, opened 2023-03-21
Refactor ockam_node, extend Routed<M>

SanjoDeundiak:


implementations/rust/ockam/ockam_node/src/context/receive_message.rs
```diff
@@ -39,26 +39,22 @@ impl Context {
...
     /// This function will block and re-queue messages into the
     /// mailbox until it can receive the correct message payload.
     ///
     /// WARNING: this will temporarily create a busyloop, this
     /// mechanism should be replaced with a waker system that lets the
     /// mailbox work not yield another message until the relay worker
     /// has woken it.
-    async fn next_from_mailbox<M: Message>(&mut self) -> Result<(M, LocalMessage, Address)> {
+    /// A convenience function to get a Routed message from the Mailbox
+    async fn next_from_mailbox<M: Message>(&mut self) -> Result<Routed<M>> {
```
etorreborre, line 43: Nice!

etorreborre (approved):
This is great! 

One thing I would be tempted to do is to put `Context` into its own file as well and not in `mod.rs`. This way we could normalize `mod.rs` being a pure declaration of modules and uses. Having some additional code, types, constants is an open-door in my mind to adding more stuff in there because the limit of what should go there is fuzzy. 

If we just put the module documentation, `mod` and `use` statements the end result will be more helpful to navigate the crates IMO.

PR #4507, opened 2023-03-21
make authority node to exit with error code on startup failure

polvorin:
We need to make the authority to exit with an error if it can't startup all services.

(to test,  one easy way would be to put something like  `options.state.nodes.get("I_DO_NOT_EXISTS")?`  [here](https://github.com/build-trust/ockam/blob/a020f04e513e2c71ce200967712ed9b05862e4af/implementations/rust/ockam/ockam_command/src/authority/create.rs#L192)  , that make the startup to fail  and _must_ result in the process being terminated)

I don't know if a better solution than this would be to make [embedded_node_that_is_not_stopped](https://github.com/build-trust/ockam/blob/a020f04e513e2c71ce200967712ed9b05862e4af/implementations/rust/ockam/ockam_command/src/util/mod.rs#L416) to actually exit if the underling function return error. 

A nuance is the sleep before termination (see comment on code),  sadly right now that's still necessary.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/bootstrapped_identities_store.rs
+++ b/implementations/rust/ockam/ockam_api/src/bootstrapped_identities_store.rs
@@ -9,6 +9,7 @@ use ockam_identity::authenticated_storage::{
 };
 use ockam_identity::credential::Timestamp;
 use ockam_identity::IdentityIdentifier;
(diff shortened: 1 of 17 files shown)


etorreborre:
This PR can be closed now.

PR #4513, opened 2023-03-22
Sessions update

SanjoDeundiak:


examples/rust/get_started/examples/09-streams-initiator.rs
```diff
@@ -34,7 +34,9 @@ async fn main(mut ctx: Context) -> Result<()> {
     .await?;
 
     // Receive a message from the "responder-to-initiator" stream
-    let reply = ctx.receive_block::<String>().await?;
+    let reply = ctx
```
etorreborre, line 37: I know that we have less syntactic options in Rust than in other languages but the various `_ext` or `_extension` methods are not great IMO. I think we might be able to migrate to builder patterns instead. Here is an example:
```rust
use ockam::Context;
use ockam_core::{Result, Route};

/// Builder to send messages with a Context
pub struct SendMessage {
    route: Route,
    message: String,
    timeout: Option<u8>,
}

impl SendMessage {
    pub fn new(route: Route, message: String) -> SendMessage {
        SendMessage {
            route,
            message,
            // no timeout by default
            timeout: None,
        }
    }
    pub fn with_timeout(&mut self, t: u8) -> &mut SendMessage {
        self.timeout = Some(t);
        self
    }

    pub async fn run(&self, ctx: &Context) -> Result<()> {
        match &self.timeout {
            // TODO use the timeout
            Some(_t) => ctx.send(self.route.clone(), self.message.clone()).await,
            None => ctx.send(self.route.clone(), self.message.clone()).await,
        }
    }
}

/// in send_message.rs
fn send_message(route: impl Into<Route>, message: &str) -> SendMessage {
    SendMessage::new(route.into(), message.to_string())
}

/// in some user code
send_message("address", "hello")
  .with_timeout(10)
  .run(ctx)
  .await?
```
The main difference is that we don't start from `ctx` but we build an expression _then_ we execute it in a given context.

I think that this approach:

 - is applicable in other places in our APIs where we currently resort to some variations when naming functions with the `ext` suffix so we could make this approach very consistent
 
 - is scalable like the options pattern that you introduced

 - is concise, more than the options pattern since `MessageReceiveOptions` stays essentially hidden: no import and the expression takes less place

What do you think?

etorreborre (approved):
All good. I think we should consider my syntactic suggestion in other PRs

examples/rust/get_started/examples/09-streams-initiator.rs
SanjoDeundiak: I also don't like `_ext` and other suffixes, but builder pattern that should be run in the end looks overkill to me at the moment.

PR #4514, opened 2023-03-22
RemoteForwarder Sessions

SanjoDeundiak:


implementations/rust/ockam/ockam/src/remote/mod.rs
```diff
@@ -2,13 +2,28 @@
...
 pub use info::*;
 pub use trust_options::*;
+
+use crate::remote::addresses::Addresses;
+use core::time::Duration;
+use ockam_core::compat::{string::String, vec::Vec};
+use ockam_core::Route;
+use ockam_node::DelayedEvent;
+
+/// This Worker is responsible for registering on Ockam Hub and forwarding messages to local Worker
```
etorreborre: s/Ockam Hub/Ockam Orchestrator

implementations/rust/ockam/ockam/src/remote/addresses.rs
```diff
@@ -0,0 +1,34 @@
...
+
+#[derive(Clone, Debug)]
+pub(super) struct Addresses {
+    // Used to talk to the service
+    pub(super) main_remote: Address,
+    // Used to forward messages inside the node
+    pub(super) main_internal: Address,
+    // Used to receive heartbeats
+    pub(super) heartbeat: Address,
+    // Used to receive completion callback
```
etorreborre, line 12: Maybe "Used to _send_ a completion callback"?

implementations/rust/ockam/ockam/src/remote/lifecycle.rs
```diff
@@ -0,0 +1,224 @@
...
+};
+use ockam_core::{
+    route, Address, AllowAll, AllowSourceAddress, DenyAll, Mailbox, Mailboxes,
+    OutgoingAccessControl, Result, Route,
+};
+use ockam_node::{DelayedEvent, WorkerBuilder};
+use tracing::debug;
+
+#[derive(Clone, Copy)]
+pub(super) enum Ftype {
```
etorreborre: I would rather name this `ForwardType` instead.

implementations/rust/ockam/ockam/src/remote/info.rs
```diff
@@ -0,0 +1,27 @@
+use crate::Message;
+use ockam_core::compat::string::String;
+use ockam_core::{Address, Route};
+use serde::{Deserialize, Serialize};
+
+/// Information about a remotely forwarded worker.
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, Debug, Message)]
+pub struct RemoteForwarderInfo {
+    pub(super) forwarding_route: Route,
+    pub(super) remote_address: String,
```
etorreborre, line 10: Can we use an `Address` instead of a String there?

implementations/rust/ockam/ockam/src/remote/lifecycle.rs
```diff
@@ -0,0 +1,224 @@
...
+            addresses,
+            completion_msg_sent: false,
+            registration_route,
+            registration_payload,
+            heartbeat,
+            heartbeat_interval,
+        }
+    }
+
+    /// Create and start static RemoteForwarder at predefined address with given Ockam Hub route
```
etorreborre: s/Ockam Hub/Ockam Orchestrator

implementations/rust/ockam/ockam/src/remote/lifecycle.rs
```diff
@@ -0,0 +1,224 @@
...
+            .await?;
+
+        let resp = callback_ctx.receive::<RemoteForwarderInfo>().await?.body();
+
+        Ok(resp)
+    }
+
+    /// Create and start new static RemoteForwarder without heart beats
+    // This is a temporary kind of RemoteForwarder that will only run on
+    // rust nodes (hence the `forwarding_service` addr to create static forwarders).
```
etorreborre: That comment is not clear to me. Maybe it did not age well?

implementations/rust/ockam/ockam/tests/forwarder.rs
```diff
@@ -0,0 +1,165 @@
...
+    ForwardingService::create(ctx, "forwarding_service", AllowAll, AllowAll).await?;
+
+    ctx.start_worker("echoer", Echoer, AllowAll, AllowAll)
+        .await?;
+
+    let remote_info =
+        RemoteForwarder::create(ctx, route![], RemoteForwarderTrustOptions::new()).await?;
+
+    let resp = ctx
+        .send_and_receive_extended::<String>(
```
etorreborre: `send_and_receive` without the `MessageSendReceiveOptions` should be enough here

implementations/rust/ockam/ockam/tests/forwarder.rs
```diff
@@ -0,0 +1,165 @@
...
+    let remote_info = RemoteForwarder::create(
+        ctx,
+        cloud_connection.clone(),
+        RemoteForwarderTrustOptions::new()
+            .as_consumer_and_producer(&server_sessions, &server_tcp_session_id),
+    )
+    .await?;
+
+    let mut child_ctx = ctx.new_detached("ctx", AllowAll, AllowAll).await?;
+    ctx.send(
```
etorreborre, line 133: Why not group both statements with `send_and_receive`?

implementations/rust/ockam/ockam/tests/forwarder.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    assert!(res.is_err(), "Should not pass outgoing access control");
+
+    ctx.send(
+        route![remote_info.remote_address(), "ctx"],
+        "Hello".to_string(),
+    )
+    .await?;
+
+    server_sessions.add_consumer(
```
etorreborre, line 147: Did you choose on purpose to `add_consumer` in between `send` and `receive`? Is it possible that the message fails to be sent and gets dropped before we even have the chance to add a consumer?

implementations/rust/ockam/ockam/tests/forwarder.rs
```diff
@@ -0,0 +1,165 @@
...
+        )
+        .await?;
+
+    assert_eq!(resp, "Hello");
+
+    ctx.stop().await
+}
+
+// Server: Connects to a Cloud using tcp and creates a dynamic Forwarder. Using session
+// Cloud: Hosts a Forwarding service and sends a message to the Client with and without a session
```
etorreborre: Maybe `send replies from the Server to the Client` instead of `sends a message` because that gives the impression that it is sending a message to the client on its own will.

implementations/rust/ockam/ockam/src/remote/lifecycle.rs
```diff
@@ -0,0 +1,224 @@
...
+            registration_payload,
+            heartbeat,
+            heartbeat_interval,
+        }
+    }
+
+    /// Create and start static RemoteForwarder at predefined address with given Ockam Hub route
+    pub async fn create_static(
+        ctx: &Context,
+        hub_route: impl Into<Route>,
```
etorreborre, line 87: We probably need to do a full pass on `Hub`. There are many places where it is still used.

implementations/rust/ockam/ockam/tests/forwarder.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    let server_tcp = TcpTransport::create(ctx).await?;
+    let cloud_connection = server_tcp
+        .connect(
+            socket_addr.to_string(),
+            TcpConnectionTrustOptions::new().as_producer(&server_sessions, &server_tcp_session_id),
+        )
+        .await?;
+
+    let remote_info = RemoteForwarder::create(
```
etorreborre, line 67: I think it would be nice to have an additional test where we create a Forwarder via a secure channel.

etorreborre (dismissed):
That looks good. I just wonder if we should add some more tests where the server uses a secure channel to create its forwarder and the client also use a secure channel to send messages.

implementations/rust/ockam/ockam/src/remote/addresses.rs
SanjoDeundiak: No, it's used to receive

implementations/rust/ockam/ockam/src/remote/info.rs
SanjoDeundiak: I think it makes sense to keep it as a String, because it's only used to print it and transfer to the other node. It's not used as an `Address` on the node that operates this struct

implementations/rust/ockam/ockam/src/remote/lifecycle.rs
SanjoDeundiak: It's actually an Ockam Project node I think, but more specifically it's the route to a node with Forwarding Service, can be any. But for simplicity let's call it Ockam Orchestrator

implementations/rust/ockam/ockam/src/remote/lifecycle.rs
SanjoDeundiak: Hm, I think it's correct. Not very clear, but it's a workaround function unfortunately that we would like to remove in the future

implementations/rust/ockam/ockam/tests/forwarder.rs
SanjoDeundiak: Because we send and receive from different Context instances

implementations/rust/ockam/ockam/tests/forwarder.rs
SanjoDeundiak: Nice catch. That would be a flaky test!

implementations/rust/ockam/ockam/src/remote/lifecycle.rs
SanjoDeundiak: I'll create an issue for that

SanjoDeundiak: https://github.com/build-trust/ockam/issues/4520

etorreborre (approved):
🏆 let's merge!

PR #4532, opened 2023-03-24
fix(rust): keep migrating legacy storage data but not on startup

etorreborre:
This is an attempt to fix some restart issues that we observed with Ockam Orchestrator. In this PR:

 - the code for migrating legacy data has been put back
 - however it is only activated once when we enroll a member and not at start-up in order to avoid triggering a LMDB exception

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3553,6 +3553,7 @@ dependencies = [
  "ockam_core",
  "ockam_macros",
  "ockam_node",
+ "openssl",
  "p256",
  "rand",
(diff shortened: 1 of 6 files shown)


davide-baldo:
There is no guarantee that a new member will be added so we don't have any guarantee that the storage will be migrated.
Is there a certain trigger we can add?

etorreborre:
@davide-baldo yes, this is actually silly. I think I need to better understand exactly what is not working at the moment. Closing this for now

PR #4535, opened 2023-03-24
Sessions support for Portals

SanjoDeundiak:


implementations/rust/ockam/ockam/src/lib.rs
```diff
@@ -129,4 +129,7 @@ pub mod authenticated_storage {
 }
 
 #[cfg(feature = "ockam_transport_tcp")]
-pub use ockam_transport_tcp::{TcpConnectionTrustOptions, TcpListenerTrustOptions, TcpTransport};
+pub use ockam_transport_tcp::{
```
SanjoDeundiak, line 132: @etorreborre do you think we could create submodule `ockam::tcp`?

etorreborre: For consistency that would be great!

implementations/rust/ockam/ockam_identity/src/channel/trust_options.rs
```diff
@@ -102,7 +102,7 @@ impl Default for SecureChannelListenerTrustOptions {
 }
 
 impl SecureChannelListenerTrustOptions {
-    /// Constructor
+    /// Constructor without Consumer and Producer sessions, with [`TrustEveryonePolicy
```
etorreborre: The end of line is missing here

implementations/rust/ockam/ockam_transport_tcp/src/transport/portals.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// Stop outlet at addr
+    /// ```rust
+    /// use ockam_transport_tcp::TcpTransport;
+    /// # use ockam_node::Context;
+    /// # use ockam_core::{AllowAll, Result};
+    /// # async fn test(ctx: Context) -> Result<()> {
+    /// const TARGET_PEER: &str = "127.0.0.1:5000";
+    ///
+    /// let tcp = TcpTransport::create(&ctx).await?;
+    /// tcp.create_outlet("outlet", TARGET_PEER, AllowAll).await?;
```
etorreborre: That comment needs to be fixed

etorreborre (dismissed):
🏎️ that's quite straightforward now :-)

etorreborre (approved):

PR #4542, opened 2023-03-24
refactor(rust): extract identity as an entity

etorreborre:
## Introduction 

This PR aims at simplifying the current handling of identities to provide:

 - a better notion of `Identity`: at the moment an `Identity` is not only a set of key changes representing a specific actor in the system but also a set of services shared by all the actors (to persist keys, sign them, store attributes etc...). As a result a redundant notion of `PublicIdentity` had to be introduced. Moreover it is not even possible to simply `clone` an `Identity` since it contains a `Context`

 - a more consistent and navigable API for the available services which can be used to
   - create / import identities
   - store attributes / issue credentials
   - create secure channels

Since the notion of `Identity` is tied to most of the library, this PR is necessarily touching many files. The following sections describe the main changes:

## The `ockam_identity` crate

The top-level folders in the `ockam_identity` crate are organized around "entity" (singular) and "entities" (plural, services for 'entity') modules:

 - `credential`: `CredentialData`, `Credential`
 - `credentials`: `Credentials` service
 - `identity`: `Identity`, `IdentityChangeHistory`, `IdentityIdentifier`
 - `identities`: `Identities`, `IdentitiesCreation` services
 - `secure_channel`: `TrustOptions`, `Addresses` etc...
 - `secure_channels`: `SecureChannels` services

### Data types

The main changes are:

 - `Identity` is now just a set of key changes and an identifier, no more services, no more `Context`
 - `PublicIdentity` (and corresponding functions) has been removed
 - `CredentialBuilder` had to have some modifications to incorporate the issuer (it was implicit before when doing `identity.issue_credential`)

### Node Services

**`Identities`**  is a set of services to manage identities
   - `IdentitiesCreation` to create / import identities
   - `IdentitiesKeys` to manage identities keys
   - `Credentials` to manage attributes and credentials
   - `CredentialsServer` to start a worker serving credentials

The `Identities` services are backed up by
   - an `IdentitiesVault` to store keys and provide cryptographic information
   - an `IdentitiesRepository` to persist identities data (attributes etc...). It has 4 interfaces
     - `IdentityAttributesWriter` to persist the attributes of an identity
     - `IdentityAttributesReader` to retrieve the attributes of an identity
     - `IdentitiesWriter` to persist a known identity
     - `IdentitiesReader` to get a persisted known identity

**`SecureChannels`** is a set of services used to create secure channels
  - it uses `Identities` services (so it uses the same underlying vault)
  - it has a `SecureChannelsRegistry`

Note on vaults: strictly speaking there is one less vault interface needed to manage identities (no need for `SymmetricVault`). However introducing a distinction between `IdentitiesVault` and `SecureChannelsVault` led to more difficulties than necessary for no obvious modularity gain.

**Additional note** the `Context` which was stored before inside `Identity` is now explicitly passed as an argument when running operations requiring it, for example to create a secure channel. This makes that part slightly more verbose but now it also dispense of the creation of a full `Context` when doing simple identity operations like storing attributes.

### Builders

`Identities` and `SecureChannels` can be created using a builder pattern. For example 
```rust
let identities = identities::builder().with_identities_storage(FileStorage::new("identities")).build()
```

In the user guide examples a `Services` struct is available, and gives access to all the required services to create identities, secure channel, access credentials etc... (see the modified examples). This leaves the room open for other types of services (projects, spaces?). It also opens the possibility to encapsulate the creation of transports in the future (`services.create_tcp_transport().await?`) but I haven't explored that yet.

### `NodeManager`

The `NodeManager` now has

 - an `Identity`
 - an optional `credential`
 - `SecureChannels` services (all sub-services like `identities_creation` can be directly accessed via `NodeManager::self`

The independent `Vault` and `AttributesStorage` have been removed since they are now implementation details of `SecureChannels`. [The `NodeManager` uses a `SecureChannels` builder](https://github.com/build-trust/ockam/pull/4542/files#diff-fee061b9c2629c91dcb8dcd8412526980626c18ba53562340f2040cab470469aR315).

### `NodeIdentities`

A `NodeIdentities` struct has been introduced to help implement the `IdentitiesService` endpoint, where a default vault is used unless a vault name is specified. I suspect that there could be better/more refactorings there.

## The `ockam_command` crate

On the `ockam_command` side we are mostly adapting to the changes in `ockam_api`. One thing to notice is that it is less frequent to require a `Context` to run operations.
 
## Examples

At the top-level, since we separate an `Identity` from various capabilities (creation, secure channels), we need to change the user-facing API. The proposal is to do the following:
```rust
// Create an Identity to represent Alice.
let node = node(&ctx);
let alice = node.create_identity().await?;
...
// Connect to a secure channel listener and perform a handshake.
let r = route![connection_to_middle_node, "forward_to_bob", "bob_listener"];
let channel = node
      .create_secure_channel(&alice, r, TrustEveryonePolicy).await?;
```
In the code above:

 - we instantiate a default set of `ockam` services as a `node`. The default implementations of all services is storing data in-memory but there is also a `node::builder()` to configure exactly how we want implementations to behave

 - we obtain various capabilities: create identities, or start secure channels from the `services` variable. One advantage of this approach is that it makes the top level api discoverable in and IDE by simply asking the IDE to propose completions

## TO BE DONE

The `CHANGELOG` files for each crate have not been updated with all the changes yet.
Examples in the documentation also need to be updated.


implementations/rust/ockam/ockam_identity/src/credential_issuer.rs
```diff
@@ -1,256 +0,0 @@
-use crate::alloc::string::ToString;
```
SanjoDeundiak, line 1: It's now an empty file

SanjoDeundiak: There are others, could you please check for empty files?

implementations/rust/ockam/ockam_identity/src/identities/identities.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    /// Return the identities credentials server
+    pub fn credentials_server(&self) -> Arc<dyn CredentialsServer> {
+        Arc::new(CredentialsServerModule::new(self.credentials()))
+    }
+}
+
+/// This struct supports all the services related to identities
+#[derive(Clone)]
+pub struct Identities {
```
SanjoDeundiak: Is it intended that struct declaration is at the bottom?

implementations/rust/ockam/ockam_identity/src/identities/identities_creation.rs
```diff
@@ -0,0 +1,118 @@
...
+        let identity_keys = IdentitiesKeys::new(self.vault.clone());
+        identity_keys
+            .verify_all_existing_changes(&change_history)
+            .await?;
+
+        let identifier = self.compute_identity_identifier(&change_history).await?;
+        Ok(Identity::new(identifier, change_history))
+    }
+
+    /// Create an identity with a vault initialized with a specific private key
```
SanjoDeundiak: Let's stick to the "secret key" term

implementations/rust/ockam/ockam_identity/src/credential_issuer.rs
etorreborre: That looks like the result of merge conflicts resolution. Done.

implementations/rust/ockam/ockam_identity/src/identities/identities.rs
etorreborre: No particular reason. Moved up now

implementations/rust/ockam/ockam_identity/src/identities/identities_creation.rs
etorreborre: Done

implementations/rust/ockam/ockam/src/node.rs
```diff
@@ -0,0 +1,329 @@
...
+    Node {
+        context: ctx,
+        secure_channels: secure_channels(),
+    }
+}
+
+/// Return a new builder for top-level services
+pub fn builder() -> NodeBuilder {
+    NodeBuilder::new()
+}
```
adrianbenavides: Why have these two methods separated from the `Node` struct? Especially the `node` method, it feels unnatural to not have it as `Node::new(ctx)`. Also, we have been attaching the `builder` methods to the built struct in several places of the codebase.

etorreborre: `node` is exposed like that because for now it is used directly in examples as 
```rust
#[ockam_macros::node]
fn main(ctx: Context) -> Result<()> {
  let node = node(&ctx);
...
}
```

With the idea that later on we could simply have
```
#[ockam_macros::node]
fn main(node: Node) -> Result<()> {
...
}
```

> Also, we have been attaching the builder methods to the built struct in several places of the codebase.

Ok, I can do this.

etorreborre: Actually I think my reason was to not have a recursive dependency between the builder and the "buildee" so that `builder -> buildee` and not the other way around. I prefer to avoid circular dependencies as much as I can but in that case they are likely to evolve together so this can be an exception to the rule.

examples/rust/get_started/examples/old/12-secure-channel-over-cloud-node-forwarding-initiator.rs
```diff
@@ -11,25 +11,25 @@ async fn main(mut ctx: Context) -> Result<()> {
         "Paste the forwarding address of the secure channel here.";
 
     // Initialize the TCP Transport.
-    let _tcp = TcpTransport::create(&ctx).await?;
+    let mut node = node(ctx);
+    let _tcp = node.create_tcp_transport().await?;
```
mrinalwadhwa, line 15: Does this couple nodes to the std TCP transport?
We want transport to remain pluggable

examples/rust/ockam_kafka/examples/ockam_kafka_alice.rs
```diff
@@ -1,20 +1,15 @@
...
-    let tcp = TcpTransport::create(&ctx).await?;
-
-    // Create a Vault to safely store secret keys for Alice.
-    let vault = Vault::create();
+    let mut node = node(ctx);
+    let tcp = node.create_tcp_transport().await?;
 
     // Create an Identity to represent Alice.
-    let alice = Identity::create(&ctx, vault).await?;
+    let alice = node.create_identity().await?;
```
mrinalwadhwa, line 13: How would an alternate vault impl be plugged in?

etorreborre: Instead of creating a default `node` with `node(&ctx)`, we can use a builder:
```rust
let node = Node::builder().with_identities_vault(Vault::create()).build(&ctx);
```

examples/rust/get_started/examples/old/12-secure-channel-over-cloud-node-forwarding-initiator.rs
etorreborre: `create_tcp_transport` is some syntactic sugar on top of `TcpTransport::create(node.context())`. 
So other transports can still be started with `XxxTransport::create(node.context())`

implementations/rust/ockam/ockam_api/src/bootstrapped_identities_store.rs
```diff
@@ -87,17 +85,46 @@ impl IdentityAttributeStorageWriter for BootstrapedIdentityStore {
...
+    }
+}
+
+#[async_trait]
+impl IdentitiesWriter for BootstrapedIdentityStore {
+    async fn put_identity(&self, identity: &Identity) -> Result<()> {
+        self.repository.put_identity(identity).await
+    }
+
+    async fn update_known_identity(&self, identity: &Identity) -> Result<()> {
```
adrianbenavides: In this particular case, I found confusing the put/update terms. Is the `put` method overriding the identity if already exists?

mrinalwadhwa:
@etorreborre love the direction of this. After a quick skim, note that we want to allow users to be able to plug in at least -  async executor, hardware key stores, transports.

We want this to be possible without them talking to us or without the deps of TCP leaking into into a program that uses some other transport say LoRaWAN.

Similarly for vaults like AWS KMS or Yubikey etc.. Our current design for the KMS vault need some fixing.

etorreborre:
@mrinalwadhwa we can plug-in vaults and transports with this PR. You also wrote "async executor". How is it possible at the moment since the `node` macro gives you a `Context` fully wired to an executor? This is done via the compilation configuration, right?

implementations/rust/ockam/ockam_api/src/bootstrapped_identities_store.rs
etorreborre: Yes it does. I agree this is not fantastic naming. I exposed `put_identity` because it looked like a more fundamental operation than `update_known_identity` which can be implemented by any client accessing `put_identity`. 
What I can do is hide `put_identity` which is actually not used anywhere in the codebase except for the implementation of `update_known_identity`.

etorreborre: Done

implementations/rust/ockam/ockam/src/node.rs
etorreborre: Done (the `builder` move)

etorreborre: BTW the `node` function is mostly meant to be used in examples to keep them short. If you look at the changed code in `ockam_api` it is using `builder`.

examples/rust/file_transfer/examples/receiver.rs
```diff
@@ -82,19 +79,15 @@ impl Worker for FileReception {
 
 #[ockam::node]
 async fn main(ctx: Context) -> Result<()> {
-    // Initialize the TCP Transport.
-    let tcp = TcpTransport::create(&ctx).await?;
-
-    // Create a Vault to safely store secret keys for Receiver.
-    let vault = Vault::create();
+    let node = node(ctx);
```
SanjoDeundiak, line 83: I'm a little bit concerned with how many decisions are made inside this function that are not visible to a user, and even if you dig inside this function, it's not really obvious what those decisions are and what other options you had.
For example, any real-world use-case would not want to use a `Vault` without persistence. I even think that the current implementation should have looked like this `Vault::create_without_persistence()`. But now it's created for you somewhere under the hood, and it will be super hard for a user to realize that their keys are gone after a restart and what should be done to fix that.
Should we explore options to make the user make this decision explicitly?

SanjoDeundiak: We can leave this discussion for later and merge as it is, if you wish

implementations/rust/ockam/ockam/src/node.rs
```diff
@@ -0,0 +1,329 @@
...
+            .await
+    }
+
+    /// Signal to the local runtime to shut down
+    pub async fn stop(&mut self) -> Result<()> {
+        self.context.stop().await
+    }
+
+    /// Send a message to an address or via a fully-qualified route
+    pub async fn send<R, M>(&self, route: R, msg: M) -> Result<()>
```
SanjoDeundiak, line 193: I think some duplication of API to this struct makes sense, but sending and receiving don't make sense to me. Those things a related very tight with an `Address` assigned to the specific `Context` instance, I'm concerned that this relation is now kinda of lost. `send_and_receive` though, creates a new detached `Context`, so it's fine.

implementations/rust/ockam/ockam_identity/src/secure_channels/secure_channels.rs
```diff
@@ -1,60 +1,89 @@
...
 use core::time::Duration;
-use ockam_core::{Address, AsyncTryClone, Result, Route};
+use ockam_core::compat::sync::Arc;
+use ockam_core::Result;
+use ockam_core::{Address, Route};
+use ockam_node::Context;
+
+/// Identity implementation
+#[derive(Clone)]
+pub struct SecureChannels {
```
SanjoDeundiak, line 17: Do you think we should wrap `SecureChannels` and other services inside an `Arc` if we can simply clone them instead?

SanjoDeundiak (dismissed):
I didn't go through the whole PR due to its size, but the structure overall looks good to me. We can merge and discuss later

examples/rust/file_transfer/examples/receiver.rs
etorreborre: What I am planning to do is to add some examples showing how to go from this simple form of example to more complex examples where the implementation can be selected by the user. 

I can also:

 - add comments on top of this instantiation, to make it clear that it is in-memory only (like we we doing with `Vault::create()`
 - add rust comments in `fn node` to explain how to build more complex implementations for production use cases

implementations/rust/ockam/ockam/src/node.rs
etorreborre: Let's discuss this further because I think that we should rework the notion of `Context`.

implementations/rust/ockam/ockam_identity/src/secure_channels/secure_channels.rs
etorreborre: I did it mostly to stay flush with other services which were provided as `dyn TraitXxx` with the idea that we might extract an interface in the future.

examples/rust/get_started/examples/old/12-secure-channel-over-cloud-node-forwarding-initiator.rs
etorreborre: I added extension methods for all transports on top of `Node` so they are all now properly pluggable (any transport can be defined independently of `Node` and can be integrated with a `node.create_xxx_transport()`)

examples/rust/file_transfer/examples/receiver.rs
etorreborre: done

SanjoDeundiak (approved):

PR #4543, opened 2023-03-25
fix(rust): make ockam reset delete specific state files

### LINKED ISSUE #4500: `ockam reset` deletes the latest release from system
### Observed behavior

The new installer script installs the binary into `~/.ockam/bin`. Running `ockam reset` will delete all of `~/.ockam`, including the sourced shell scripts and the binary. You'll be unable to run `ockam` at all after that and will have to re-install it.

### Steps to reproduce

1. Use the install script:

```bash
curl --proto '=https' --tlsv1.2 -sSf \
    https://raw.githubusercontent.com/build-trust/ockam/develop/install.sh | bash
```

2. Run `ockam reset`
3. Running `ockam` will now fail

### Expected behavior

Don't have `ockam reset` remove the installed binary

### Ockam Version

0.82.0

mrinalwadhwa:
Fixes #4500

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state.rs
@@ -11,11 +11,11 @@ use ockam_identity::{
 use ockam_vault::{storage::FileStorage, Vault};
 use rand::random;
 use serde::{Deserialize, Serialize};
-use std::fmt;
 use std::fmt::{Display, Formatter};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::time::SystemTime;
+use std::{fmt, fs};
 use sysinfo::{Pid, System, SystemExt};
 
 use crate::lmdb::LmdbStorage;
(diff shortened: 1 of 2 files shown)


glenngillen:
@mrinalwadhwa I'm wondering if purely from a safety and least surprise perspective if it's possible to be explicit about the files to delete rather than the exceptions? My spidey-sense is going off on this one as something that will bite us or a user in the backside again in the future.

mrinalwadhwa:
@glenngillen makes a lot of sense. I changed it to delete specific directories and files, although I think the code 6c4214eed98ec70a942761c792d64dc80ac402f0 is still brittle but better than e002c8d421684b9718f3a3680acc36f12ddc57b2. For example I had to know to delete `config.json` even though there is no mention of it in the CliState structure cc: @CAOakleyII @adrianbenavides

adrianbenavides:
I guess we should consider storing the state in a specific folder like `~/.ockam/state`. This will also be necessary when adding support for plugins anyways, as we'll want to install them under `~/.ockam/plugins` or similar. This change, however, would have to be done carefully to add some backward compatibility for the nodes deployed on the cloud side.

etorreborre:
I am going to add a bats test to avoid regressions

etorreborre (dismissed):

davide-baldo (approved):

mrinalwadhwa:
@etorreborre thank you for adding tests 🙏

PR #4551, opened 2023-03-27
fix(rust): when deleting the default vault/identity/project the data and the link are deleted

### LINKED ISSUE #4541: Default project delete+create fails to refresh `project/default.json`
### Observed behavior

When deleting the default project and re-creating it, the old default.json is kept around, causing major failures.

### Steps to reproduce

ockam project delete "${SPACE_ID}" default
ockam project create "${SPACE_ID}" default
ockam project enroll --attribute role=member

### Expected behavior

the member is enrolled

### Ockam Version

ockam 0.82.0

adrianbenavides:
Fixes https://github.com/build-trust/ockam/issues/4541 as well as other related issues:
- When deleting the default vault/identity/project the link under `$OCKAM_HOME/defaults` is also deleted.
- When a space is deleted via `ockam space delete` the stored projects in `$OCKAM_HOME/projects` are deleted as well
- When a project is created via `ockam project create` it will be stored in `$OCKAM_HOME/projects`. Previously, projects were stored in the state only when creating the default project in `ockam enroll`
- When a project is deleted via `ockam project delete` it will be deleted from `$OCKAM_HOME/projects`

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state.rs
@@ -182,14 +182,21 @@ impl VaultsState {
 
     pub async fn delete(&self, name: &str) -> Result<()> {
         // Retrieve vault. If doesn't exist do nothing.
-        let vault_state = match self.get(name) {
+        let vault = match self.get(name) {
             Ok(v) => v,
             Err(CliStateError::NotFound(_)) => return Ok(()),
             Err(e) => return Err(e),
         };
 
+        // If it's the default, remove link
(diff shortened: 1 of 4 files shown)


etorreborre (approved):
That looks all good to me. We could add bats tests to it to avoid regressions, maybe?

adrianbenavides:
> That looks all good to me. We could add bats tests to it to avoid regressions, maybe?

I thought about it but I couldn't come up with a good strategy to test it. I'll create an issue with the idea I have in mind.

PR #4552, opened 2023-03-27
tests(rust): generate the project.json file before every test instead of reusing it

adrianbenavides:
Fixes the following problems

### bats tests setup

Currently, the `project.json` file is generated once before running the first test and then it's shared among the other tests. This can cause problems if one of the tests modifies that file (deleting the default project for example). 

This PR changes that strategy and generates/stores the file in the test environment before running it. This will make orchestrator-based tests a little bit slower but more robust.

### project deletion from CliState

Two different fixes have been applied here:
- `ockam space delete` no longer deletes the projects from the CliState. Details [here](https://github.com/build-trust/ockam/pull/4552/files#diff-cc6c2d924ad6a110a7e92821050d827ba16ca0a5eb742b314d3a72ffee4a06d4R73)
- deleting a project from CliState now works as expected: the `ProjectState` is just a json file with the project details and we were trying to delete it as a directory, causing an error. Details [here](https://github.com/build-trust/ockam/pull/4552/files#diff-a17a5a1da1448cd1d92017a87dd5cd59c347565441dea82b61b4903d814776d7R1187) and [here](https://github.com/build-trust/ockam/pull/4552/files#diff-a17a5a1da1448cd1d92017a87dd5cd59c347565441dea82b61b4903d814776d7L1250)

### DIFF
--- a/implementations/rust/ockam/ockam_command/tests/bats/load/orchestrator.bash
+++ b/implementations/rust/ockam/ockam_command/tests/bats/load/orchestrator.bash
@@ -22,15 +22,8 @@ function skip_if_long_tests_not_enabled() {
 
 function load_orchestrator_data() {
   if [ ! -z "${ORCHESTRATOR_TESTS}" ]; then
-    if [ ! -f "$OCKAM_HOME_BASE/project.json" ]; then
-      OCKAM_HOME=$OCKAM_HOME_BASE $OCKAM project information --output json >"$OCKAM_HOME_BASE/project.json"
-    fi
-    export PROJECT_JSON_PATH="$OCKAM_HOME_BASE/project.json"
-  fi
-}
-
-function copy_orchestrator_data() {
-  if [ ! -z "${ORCHESTRATOR_TESTS}" ]; then
     cp -a $OCKAM_HOME_BASE $OCKAM_HOME
+    export PROJECT_JSON_PATH="$OCKAM_HOME/project.json"
+    $OCKAM project information --output json >"$PROJECT_JSON_PATH"
   fi
 }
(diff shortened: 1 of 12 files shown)


implementations/rust/ockam/ockam_command/tests/bats/load/orchestrator.bash
```diff
@@ -21,16 +21,9 @@ function skip_if_long_tests_not_enabled() {
...
-      OCKAM_HOME=$OCKAM_HOME_BASE $OCKAM project information --output json >"$OCKAM_HOME_BASE/project.json"
-    fi
-    export PROJECT_JSON_PATH="$OCKAM_HOME_BASE/project.json"
-  fi
-}
-
-function copy_orchestrator_data() {
   if [ ! -z "${ORCHESTRATOR_TESTS}" ]; then
     cp -a $OCKAM_HOME_BASE $OCKAM_HOME
+    $OCKAM project information --output json >"$PROJECT_JSON_PATH"
```
etorreborre: Shouldn't those 2 statements be switched?

adrianbenavides: `cp -a $OCKAM_HOME_BASE $OCKAM_HOME` will include everything we have in the `OCKAM_HOME_BASE` directory, but the idea is to override the project.json file before running a test so we have an updated version. Does that make sense to you?

Actually, I found a bug there, the second line should be `$OCKAM project information --output json >"$OCKAM_HOME/project.json", I'll push the change now.

etorreborre: Yes I was talking about L.26 and L27, I should have been more precise. Maybe now I would write this instead to avoid the repetition:
```sh
export PROJECT_JSON_PATH="$OCKAM_HOME/project.json"
$OCKAM project information --output json >"$PROJECT_JSON_PATH"
```

adrianbenavides: 🤦 of course, fixed!

etorreborre (approved):

PR #4553, opened 2023-03-27
Sessions usage

SanjoDeundiak:
Currently, `Sessions` are not used in:
1. Most of the code examples and most of the tests
2. Node's main tcp listener (the one that listens on 127.0.0.1:something and is used only locally from ockam_command)

Everything else is using Sessions

Further steps:
1. Pick a better name for `Sessions`
2. Integrate `Sessions` into `Context`, in a way that you can access it from any place + `Sessions` info is tied to the `Address` lifecycle (right now, we don't clean resources when the `Worker` is shut down)
4. Use `Sessions` for the node's localhost tcp listener
5. Integrate `Sessions` into commands
6. Clean `::new()` usage in tests and examples
7. Revisit the need for Hop worker

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
+++ b/implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
@@ -2,6 +2,7 @@ use bytes::{Bytes, BytesMut};
 use core::sync::atomic::AtomicBool;
 use core::sync::atomic::Ordering;
 use ockam_core::compat::sync::Arc;
+use ockam_core::sessions::{SessionId, SessionPolicy, Sessions};
 use ockam_core::{
     errcode::{Kind, Origin},
     Address, AllowAll, AsyncTryClone, Encodable, Error, LocalInfo, LocalMessage, Route, Routed,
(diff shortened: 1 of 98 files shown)


implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
```diff
@@ -275,6 +277,13 @@ impl KafkaPortalWorker {
             .start_worker(inlet_address.clone(), inlet_worker, AllowAll, AllowAll)
             .await?;
 
+        if let Some((sessions, session_id)) = session {
+            sessions.add_consumer(
+                &outlet_address,
```
davide-baldo, line 282: I don't understand why it is working: we are only allowing outlet worker (aka worker that handles responses from kafka) but we are not explicitly allowing the inlet (aka worker that handles requests to kafka)

SanjoDeundiak: @davide-baldo because any worker on our node can send a message to a secure channel, but you need to be a consumer to receive messages from a secure channel. Does that make sense now?

davide-baldo (approved):

PR #4568, opened 2023-03-28
feat(rust): serialize keys using hex while keeping back-compatibility

davide-baldo:
## Current behavior

Keys within vaults are stored as a list of integers: 
```
"Key":[23,251,34,116,121,124,85,11,253,75,129,109,88,235,189,88,226,150,43,55,187,149,247,137,163,166,131,44,126,133,229,139]
```

## Proposed changes

Keys are stored in hex while keeping back-compatibility for the list of integers.

```
"Key":"17fb2274797c550bfd4b816d58ebbd58e2962b37bb95f789a3a6832c7e85e58b"
```
No explicit migration is present, at the first edit the new hex serialization will be used.

### DIFF
--- a/implementations/rust/ockam/ockam_core/src/hex_encoding.rs
+++ b/implementations/rust/ockam/ockam_core/src/hex_encoding.rs
@@ -1,0 +1,57 @@
+use alloc::vec::Vec;
+use core::fmt;
+use serde::de::{SeqAccess, Unexpected};
+use serde::{Deserializer, Serializer};
+
+/// By default, serde serializes using a sequence of integers.
+/// We rather serialize a using hex string.
+pub fn serialize<S: Serializer>(value: &Vec<u8>, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    serializer.serialize_str(&hex::encode(value.as_slice()))
+}
+
+/// To keep back-compatibility we parse both sequence of integer or a hex string
+pub fn deserialize<'de, D>(deserializer: D) -> Result<Vec<u8>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    struct Visitor;
+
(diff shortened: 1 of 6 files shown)


mrinalwadhwa:
@davide-baldo this is great! I have one suggestion. When we serialize keys let's use hex, base64 is opaque to human readers and we don't need the small size win here. We use hex in other places where we serialize keys.

implementations/rust/ockam/ockam_core/src/vault/types.rs
```diff
@@ -102,6 +103,62 @@ impl AsRef<[u8]> for SecretKey {
     }
 }
 
+/// By default serde serialize using a vector of integer, we rather serialize using base64
+impl Serialize for SecretKey {
```
SanjoDeundiak: Do you think we can simplify it like this? I know that it's more complicated case because of backwards comp
https://users.rust-lang.org/t/serialize-a-vec-u8-to-json-as-base64/57781/2?u=sanjodeundiak

davide-baldo: sure!

davide-baldo: I've moved the implementation to a public module named `hex_encoding` in `ockam_core`, there is a good chance we may need it in other context as well.

implementations/rust/ockam/ockam_core/src/hex_encoding.rs
```diff
@@ -0,0 +1,55 @@
+use core::fmt;
+use serde::de::{SeqAccess, Unexpected};
+use serde::{Deserializer, Serializer};
+
+/// By default serde serialize using a sequence of integers, we rather serialize a using hex string
```
etorreborre: s/serialize/serializes

etorreborre (dismissed):

etorreborre (approved):

PR #4574, opened 2023-03-29
feat(rust): add trust context and implementation over project authority

caoakleyii:
## Proposed changes

Implements: https://github.com/build-trust/codebase/issues/513

> 3. Modify CredentialExchangeWorker as described, based on this TrustContext entity
> 5. There shouldn't be anything else left using identity.credential()?, remove that. If there is something else using it.. review why and review the plan
> 

This currently is unable to be fully implemented because of the dependencies that exist within a credential retriever.

Currently the CredentialRetriever implementation for an issuer, relies on a multiaddr, which is used to create a TCP session.

This requires the implementation to reside within the `ockam_api` crate, preventing the `CredentialExchangeWorker` and `Identity` implementations to use them as a dependency. 

Currently the callers of those structs, utilize the `TrustContext` to retrieve and set the credential.

----

Still working through the many test cases and bats test that need to be updated and added; however I wanted to get the PR open to have eyes on it early.

And potentially reopen the discussion of handling the MultiAddr dependency.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
+++ b/implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
@@ -32,6 +32,7 @@ const DEFAULT_CLIENT_TIMEOUT: Duration = Duration::from_secs(30);
 /// - `role`: b"member"
 pub const PROJECT_MEMBER_SCHEMA: SchemaId = SchemaId(1);
 pub const PROJECT_ID: &str = "project_id";
+pub const TRUST_CONTEXT_ID: &str = "trust_context_id";
 pub const LEGACY_MEMBER: &str = "member";
 
 // This acts as a facade, modifying and forwarding incoming messages from legacy clients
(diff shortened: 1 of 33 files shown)


polvorin:
> Currently the CredentialRetriever implementation for an issuer, relies on a multiaddr, which is used to create a TCP session.

>This requires the implementation to reside within the ockam_api crate, preventing the CredentialExchangeWorker and Identity implementations to use them as a dependency.

True,   the concept of  `TrustContext`  must be moved inner in the ockam crate trees,  likely into `ockam_identity`.     _Some_   of  `CredentialRetriever`  implementation can be there  (like the from memory one),   while others (the one that reads from disk for example)  can't be there and need to be on `ockam_api`.     This implies using a `trait`  for this rather than an `enum` .

caoakleyii:
> > Currently the CredentialRetriever implementation for an issuer, relies on a multiaddr, which is used to create a TCP session.
> 
> > This requires the implementation to reside within the ockam_api crate, preventing the CredentialExchangeWorker and Identity implementations to use them as a dependency.
> 
> True, the concept of `TrustContext` must be moved inner in the ockam crate trees, likely into `ockam_identity`. _Some_ of `CredentialRetriever` implementation can be there (like the from memory one), while others (the one that reads from disk for example) can't be there and need to be on `ockam_api`. This implies using a `trait` for this rather than an `enum` .

Okay. I'm happy with that sort of an implementation. 

Another thought process I was running down was how feasible it would be to use `route` instead of `multiaddr` and strictly `PathBuf` of `CredentialState`


I'll definitely start moving the implementation back over into the `ockam_identity` crate (where I once had it), and go from there.

polvorin:
> Another thought process I was running down was how feasible it would be to use route instead of multiaddr and strictly PathBuf 

that would be simpler but not possible due to the core part not having access to file system  (`no_std`)

caoakleyii:
Closing in favor of this PR: https://github.com/build-trust/ockam/pull/4588

Which has a better commit history and passing tests.

PR #4583, opened 2023-03-30
refactor(rust): implement custom get_env

### LINKED ISSUE #4549: Refactor the processing of environment variables
### Current behavior

Environment variables are processed in a dozen of places and could benefit from a more consistent treatment (see [this PR comment](https://github.com/build-trust/ockam/pull/4548#discussion_r1148944486) for an example)

### Desired behavior

Refactor the access to environment variables as a function 
```rust
pub fn get_env<T: FromString>(var_name: &str, default_value: T): Result<T>
```

where

 - `FromString` is a trait used to parse strings into values, with implementations for `String`, comma-separated list of strings, `bool`, `u8`, etc... (it cannot be `FromStr` because we want to accept values like `TRUE` or `1` for booleans)
 - `default_value` is returned if the variable is missing
 - an `Err` is returned if the value cannot be parsed

---

JungleTryne:
## Current behavior

Environment variables are used everywhere and being parsed in different manners. This requires boilerplate code to handle them.

## Proposed changes
Implement convinient custom `get_env` and `get_env_with_default` which are going to parse env variables and give the results right away.

fixes https://github.com/build-trust/ockam/issues/4549

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state.rs
@@ -18,6 +18,7 @@ use std::time::SystemTime;
 use sysinfo::{Pid, System, SystemExt};
 
 use crate::lmdb::LmdbStorage;
+use ockam_core::env::get_env_with_default;
 use ockam_identity::authenticated_storage::AuthenticatedStorage;
 use ockam_identity::credential::Credential;
 use thiserror::Error;
@@ -130,12 +131,12 @@ impl CliState {
 
     /// Returns the default directory for the CLI state.
     pub fn default_dir() -> Result<PathBuf> {
-        Ok(match std::env::var("OCKAM_HOME") {
-            Ok(dir) => PathBuf::from(&dir),
-            Err(_) => dirs::home_dir()
+        Ok(get_env_with_default::<PathBuf>(
+            "OCKAM_HOME",
+            dirs::home_dir()
                 .ok_or_else(|| CliStateError::NotFound("home dir".to_string()))?
                 .join(".ockam"),
-        })
+        )?)
     }
 
     /// Returns the directory where the default objects are stored.
--- a/implementations/rust/ockam/ockam_api/src/cloud/mod.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/mod.rs
@@ -73,7 +73,6 @@ impl<'a> BareCloudRequestWrapper<'a> {
 }
 
 mod node {
-    use std::env;
     use std::str::FromStr;
     use std::sync::Arc;
     use std::time::Duration;
@@ -82,6 +81,7 @@ mod node {
     use rust_embed::EmbeddedFile;
 
     use ockam_core::api::RequestBuilder;
+    use ockam_core::env::get_env;
     use ockam_core::{self, route, CowStr, Result};
     use ockam_identity::{IdentityIdentifier, SecureChannelTrustOptions, TrustIdentifierPolicy};
     use ockam_multiaddr::MultiAddr;
@@ -99,9 +99,9 @@ mod node {
         /// If the env var `OCKAM_CONTROLLER_IDENTITY_ID` is set, that will be used to
         /// load the identity instead of the file.
         pub(crate) fn load_controller_identity_id() -> Result<IdentityIdentifier> {
-            if let Ok(s) = env::var(OCKAM_CONTROLLER_IDENTITY_ID) {
-                trace!(idt = %s, "Read controller identity id from env");
-                return IdentityIdentifier::from_str(&s);
+            if let Ok(Some(idt)) = get_env::<IdentityIdentifier>(OCKAM_CONTROLLER_IDENTITY_ID) {
+                trace!(idt = %idt, "Read controller identity id from env");
+                return Ok(idt);
             }
             match StaticFiles::get("controller.id") {
                 Some(EmbeddedFile { data, .. }) => {
--- a/implementations/rust/ockam/ockam_command/src/terminal/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/terminal/mod.rs
@@ -6,6 +6,8 @@ use anyhow::{anyhow, Context as _};
 use indicatif::{ProgressBar, ProgressDrawTarget, ProgressStyle};
 
 use mode::*;
+use ockam_core::env::{get_env, get_env_with_default, FromString};
+use ockam_core::errcode::Kind;
 
 use crate::{OutputFormat, Result};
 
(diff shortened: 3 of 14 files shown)


implementations/rust/ockam/ockam_command/src/terminal/mod.rs
```diff
@@ -151,13 +152,13 @@ impl<W: TerminalWriter> Terminal<W> {
     fn should_disable_color(no_color: bool) -> bool {
         // If global argument `--no-color` is passed or the `NO_COLOR` env var is set, colors
         // will be stripped out from output messages. Otherwise, let the terminal decide.
-        no_color || std::env::var("NO_COLOR").is_ok()
+        no_color || get_env("NO_COLOR", false).unwrap_or_else(|e| false)
```
etorreborre: I think I would just error out in that case. If a variable is defined but cannot be parsed when starting the command line I think it is better to inform the user right away.

JungleTryne: I didn't want to change the behavior of the function, as it always returns bool. But if you think that it would be better to make this function return error, then I could do it

etorreborre: You're right we should leave it as it is for now because it is used in the building of a `Terminal` and we don't even have yet access to `stderr` to report it.

implementations/rust/ockam/ockam_api/src/cli_state.rs
```diff
@@ -130,9 +131,9 @@ impl CliState {
 
     /// Returns the default directory for the CLI state.
     pub fn default_dir() -> Result<PathBuf> {
-        Ok(match std::env::var("OCKAM_HOME") {
-            Ok(dir) => PathBuf::from(&dir),
-            Err(_) => dirs::home_dir()
+        Ok(match get_env::<String>("OCKAM_HOME") {
+            Ok(Some(dir)) => PathBuf::from(&dir),
```
etorreborre: Instead of going `PathBuf::from` here we can use a `FromString` instance and provide
```
dirs::home_dir().ok_or_else(|| CliStateError::NotFound("home dir".to_string()))?.join(".ockam")
```
as a default value

implementations/rust/ockam/ockam_api/src/cloud/mod.rs
```diff
@@ -99,7 +99,7 @@ mod node {
         /// If the env var `OCKAM_CONTROLLER_IDENTITY_ID` is set, that will be used to
         /// load the identity instead of the file.
         pub(crate) fn load_controller_identity_id() -> Result<IdentityIdentifier> {
-            if let Ok(s) = env::var(OCKAM_CONTROLLER_IDENTITY_ID) {
+            if let Ok(Some(s)) = get_env::<String>(OCKAM_CONTROLLER_IDENTITY_ID) {
                 trace!(idt = %s, "Read controller identity id from env");
                 return IdentityIdentifier::from_str(&s);
```
etorreborre: We can also provide a `FromString` instance for `IdentityIdentifier`

implementations/rust/ockam/ockam_command/src/terminal/mod.rs
```diff
@@ -41,8 +42,8 @@ impl TerminalBackground {
     ///
     /// Reference: https://stackoverflow.com/a/54652367
     pub fn detect_background_color() -> TerminalBackground {
-        match std::env::var("COLORFGBG") {
-            Ok(v) => {
+        match get_env::<String>("COLORFGBG") {
```
etorreborre: Here I think we can create a `TerminalColors` struct:
```rust
type Color = u8;

struct TerminalColors {
   foreground: Color,
   background: Color,
}

impl TerminalColors {
    pub fn terminalBackground(&self) -> TerminalBackground {
        if (0..8).contains(&self.backgroundColor) {
            TerminalBackground::Dark
        } else {
            TerminalBackground::Light
        }
    } 
}
```

And provide a `FromString` instance doing the parsing of the 2 colors with `;`

implementations/rust/ockam/ockam_command/src/util/api.rs
```diff
@@ -328,11 +329,11 @@ pub struct ProjectOpts {
 
 impl CloudOpts {
     pub fn route(&self) -> MultiAddr {
-        let route = if let Ok(s) = std::env::var(OCKAM_CONTROLLER_ADDR) {
-            s
-        } else {
-            DEFAULT_CONTROLLER_ADDRESS.to_string()
-        };
+        let route = get_env_with_default::<String>(
```
etorreborre: Same thing here, we can use a `FromString<MultiAddr>`

implementations/rust/ockam/ockam_core/src/env.rs
```diff
@@ -0,0 +1,182 @@
...
+        assert!(u8::from_string("256").is_err());
+        assert!(u16::from_string("65536").is_err());
+        assert!(u32::from_string("4294967296").is_err());
+        assert!(u64::from_string("18446744073709551616").is_err());
+    }
+
+    #[test]
+    fn test_option_ok() {
+        let result = Option::<u8>::from_string("1");
+        if let Ok(Some(value)) = result {
```
etorreborre: `assert_eq!(result, Some(1))` should work here.

implementations/rust/ockam/ockam_node/src/router/start_processor.rs
```diff
@@ -52,7 +53,7 @@ async fn start(
     router.map.internal.insert(addr.clone(), record);
 
     #[cfg(feature = "std")]
-    if std::env::var("OCKAM_DUMP_INTERNALS").is_ok() {
+    if let Ok(Some(_)) = get_env::<String>("OCKAM_DUMP_INTERNALS") {
```
etorreborre: It seems to me that this one should be a `bool`

etorreborre (commented):
@JungleTryne you made some good progress! I left some comments because I think that we can now leverage the parsing of some specific types a bit more using the `FromString` trait

etorreborre:
Hi @JungleTryne, this is excellent. You now only have to go through the errors for the `check_no_std` job and in most places follow the recommendation (like `add use crate::alloc::string::ToString;`) and (after a quick rebase) we should be ready to merge.

JungleTryne:
Thanks for the review! I don't really know how to fix
```
use of undeclared crate or module `std`
```
It doesn't really give me a hint about `std::env` :(

etorreborre:
Right we need a bunch of config flags to do some conditional compilation. I have pushed a commit which should work.

davide-baldo (approved):
nice, beautiful simplification!

etorreborre:
Thanks Dan for your contribution, it really makes the library better!

JungleTryne:
Thanks for the opportunity to contribute!

PR #4588, opened 2023-04-01
feat(rust): add and implement trust context

caoakleyii:
## Proposed changes

With this PR it introduces the Trust Context struct.

Which allows for nodes to support orchestrator authorities, local authorities, and other variations.

Trust Contexts are not tied to a project, however backwards compatibility for the --project-path options within command are supported here. 

They create trust contexts based on the project auth information.

Trust Context ID is now used for ABAC

Credential Exchange Worker uses the Trust Context to verify a credential, and can also use it to retrieve it's own credential and present back.

Ockam API's Secure Channel creation uses the trust context stored on the node manager to present credentials
 
Implementation of this issue: https://github.com/build-trust/codebase/issues/513

## Whats Left

I have manually done test cases to check this behavior.

1. Updating our tests and documentation to show these scenarios:
* Start two nodes, configure a trust_context on them without any authority, add each identity to the other' pre-trusted (trust-anchors) set when starting, outlet from one to the other => OK
* Start two nodes, configure a trust_context on them with an offline authority. Issue credentials for each, and start each node with a trust_context using this info, outlet from one to the other => OK
* Start an authority node and two nodes, configure a trust_context on them with an online authority, pointing to the one started. outlet from one to the other => OK (nodes will need to be enrolled with the authority, use the ockam authority enroll for that)

2. Add a way to configure "routes" at ockam node creation (and other commands like "ockam message", etc).

3. Add new commands to the cli. Like ockam project trust_context --output json to output a .json that can be ingested as trust context, and ockam project route aliases to output the routes.json one. For now, implement both of these based on the existing "project info" RPC call that returns these two things together.

4. Update bats test to not use project-path
## Checks


caoakleyii:
Working secure channel and auth exchange through orchestrator using only trust context info.


Currently these two identities were authenticated using a project-json.

Updating to allow for auth and enroll to be done fully through trust context next.

```❯ ockam project authenticate --identity m1  --project-path project.json --token $(cat m1.token)
---
Schema:     1
Subject:    Pba68e3ce63d258a559cc147e46e92f322d47f6639ad3ad156d3179dfc0aaeda2
Issuer:     P62e4297080fd03461787842e44d6f06ce548f60acbf58887d893f18e14bdf8c7 (OCKAM_RK)
Created:    2023-04-02T23:35:43Z
Expires:    2023-05-02T23:35:43Z
Attributes: {"ockam-role": "enroller", "project_id": "3f230f78-98c2-4674-ad21-95d733f2a2c6"}
Signature:  620bbbbe2d300e338dde7e608bcc8dbce4c36ded1f090ac288309e3c02792f0939e561aa4e34058ed0dc4c4070fd1b52c2be5cbfee90f469e7493d83fa847a0b

---
❯ ockam project authenticate --identity m2  --project-path project.json --token $(cat m2.token)
---
Schema:     1
Subject:    Pe7d0c0d453a98b6aaa71bd9357520d43946f59267e9d262b055bea3819c2a020
Issuer:     P62e4297080fd03461787842e44d6f06ce548f60acbf58887d893f18e14bdf8c7 (OCKAM_RK)
Created:    2023-04-02T23:35:50Z
Expires:    2023-05-02T23:35:50Z
Attributes: {"project_id": "3f230f78-98c2-4674-ad21-95d733f2a2c6"}
Signature:  b84f2a816b15aee058576b1f108b1426ed4c979f94547a8ade82e339eeb0cb26c5c1dcccac8e9633349704db1b784f55448492c9e4ae78bb8bededbc12cc0d02

---
❯ opr node create n1 --identity m1 --trust-context ~/trust_context_issuer.json

Node:
  Name: n1 (Default)
  Status: UP
  Route To Node:
    Short: /node/n1
    Verbose: /dnsaddr/localhost/tcp/59462
  Identity: Pba68e3ce63d258a559cc147e46e92f322d47f6639ad3ad156d3179dfc0aaeda2
  Transports:
    Transport:
      Type: TCP
      Mode: Listening
      Socket: 127.0.0.1:59462
      Worker: 0#5613b2edd6cd20d27f8b0d89e5fc7404
  Secure Channel Listeners:
    Listener:
      Address: /service/api
  Inlets:
  Outlets:
  Services:
    Service:
      Type: vault_service
      Address: /service/vault_service
    Service:
      Type: identity_service
      Address: /service/identity_service
    Service:
      Type: authenticated
      Address: /service/authenticated
    Service:
      Type: uppercase
      Address: /service/uppercase
    Service:
      Type: echo
      Address: /service/echo
    Service:
      Type: hop
      Address: /service/hop
    Service:
      Type: credentials
      Address: /service/credentials
❯ opr node create n2 --identity m2 --trust-context ~/trust_context_issuer.json

Node:
  Name: n2
  Status: UP
  Route To Node:
    Short: /node/n2
    Verbose: /dnsaddr/localhost/tcp/59470
  Identity: Pe7d0c0d453a98b6aaa71bd9357520d43946f59267e9d262b055bea3819c2a020
  Transports:
    Transport:
      Type: TCP
      Mode: Listening
      Socket: 127.0.0.1:59470
      Worker: 0#66e3883c2258073fbd507035685c9c0c
  Secure Channel Listeners:
    Listener:
      Address: /service/api
  Inlets:
  Outlets:
  Services:
    Service:
      Type: vault_service
      Address: /service/vault_service
    Service:
      Type: identity_service
      Address: /service/identity_service
    Service:
      Type: authenticated
      Address: /service/authenticated
    Service:
      Type: uppercase
      Address: /service/uppercase
    Service:
      Type: echo
      Address: /service/echo
    Service:
      Type: hop
      Address: /service/hop
    Service:
      Type: credentials
      Address: /service/credentials
❯ ockam project information --output json > project.json
❯ opr secure-channel create --from /node/n1 --to /node/n2/service/api --verbose
2023-04-02T23:36:46.438332Z  INFO ockam_node::node: Initializing ockam node
2023-04-02T23:36:46.441858Z  INFO ockam_node::worker_builder: Initializing ockam worker '0#bcc49e19f8105b36deba840344ffb9e2' with access control in:AllowAll out:DenyAll
2023-04-02T23:36:46.441990Z  INFO ockam_node::processor_builder: Initializing ockam processor '0#b040aa6740537b11aac451723ef06824' with access control in:DenyAll out:AllowAll

  Created Secure Channel:
  • From: /node/n1
  •   To: /node/n2/service/api (/ip4/127.0.0.1/tcp/59470/service/api)
  •   At: /service/debaec69d47ecdf635c65b274cf0a0f6
2023-04-02T23:36:47.579668Z  INFO ockam_node::router::shutdown: Initiate graceful node shutdown
2023-04-02T23:36:47.580062Z  INFO ockam_node::router: No more workers left.  Goodbye!
```

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -179,7 +180,8 @@ impl CredentialIssuer {
                         Credential::builder(from.clone()).with_schema(PROJECT_MEMBER_SCHEMA),
                         |crd, (a, v)| crd.with_attribute(a, v),
                     )
-                    .with_attribute(PROJECT_ID, &self.project);
+                    .with_attribute(PROJECT_ID, &self.trust_context)
```
etorreborre: Is this for backward compatibility? Will this be removed at some stage? Maybe this deserves a comment explaining the situation

implementations/rust/ockam/ockam_api/src/config/cli.rs
```diff
@@ -59,6 +69,116 @@ impl OckamConfig {
...
+pub enum CredentialRetrieverType {
+    /// Credential is stored in memory
+    FromMemory(Credential),
+    /// Path to credential file
+    FromPath(CredentialState),
+    /// MultiAddr to Credential Issuer
+    FromCredentialIssuer(CredentialIssuerInfo),
+}
+
+impl CredentialRetrieverType {
```
etorreborre, line 250: I would invert the dependency here. If `CredentialRetrieverType` is a piece of configuration denoting how we should retrieve credentials I would have a factory function on the `CredentialRetriever` side using a `CredentialRetrieverType`:
```rust
fn make_credential_retriever(retriever_type: CredentialRetrieverType) -> Result<Arc<dyn CredentialRetriever>>
```
This way we have a dependency from something that can potentially change, the implementation and behaviour of a `CredentialRetriever` towards a piece of data that is less likely to change, the `CredentialRetrieverType` (almost no associated behaviour except serialization maybe).

implementations/rust/ockam/ockam_identity/src/trust_context.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    /// Retrieve the credential for an identity within this authority
+    pub async fn credential(
+        &self,
+        for_identity: &Identity,
+    ) -> Result<Credential, ockam_core::Error> {
+        let retriever = self
+            .own_credential()
+            .ok_or(IdentityError::UnknownAuthority)?;
+        let credential = retriever.retrieve(self.identity(), for_identity).await?;
```
etorreborre: It seems to be that there is an unnecessary indirection here with the `retrieve` method. The first parameter is actually (if I understand correctly) the authority identity used to create the `AuthorityInfo`. It is used in only one implementation of a `CredentialRetriever` which is the `CredentialIssuerRetriever`.
When we create an `AuthorityInfo` we always create a `CredentialIssuer`. In the case of a `CredentialIssuerRetriever` we could already load it with the authority identity and not have to pass this parameter to the `retrieve` method.

implementations/rust/ockam/ockam_api/src/config/cli.rs
```diff
@@ -59,6 +69,116 @@ impl OckamConfig {
     }
 }
 
+/// A configuration struct to serialize and deserialize a trust context
+/// used within the ockam CLI and ockam node
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TrustContextConfig {
+    id: String,
+    authority: Option<TrustAuthorityConfig>,
```
etorreborre, line 82: Why do we need an `Option` here? Of all the usages of `TrustContextConfig` I can always see `Some` being passed.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -239,27 +244,102 @@ async fn run_foreground_node(
...
             let p: ProjectInfo = serde_json::from_str(&s)?;
             let project_id = p.id.to_string();
             project::config::set_project(cfg, &(&p).into()).await?;
-            add_project_authority_from_project_info(p, &node_name, cfg).await?;
             Some(project_id)
         }
         None => None,
     };
 
+    let mut trust_context_config = if let Some(authority_identity) = &cmd.authority_identity {
```
etorreborre: I would extract all that code to a dedicated function. Inside that function I might even call 3 sub-functions returning an `Option<TrustContextConfig` (not sure what are the best names here):

 1. `get_trust_context_config_from_authority`
 2. `get_trust_context_config_from_project`
 3. `get_trust_context_config_from_persisted`

And show what are the levels of precedence:

 - `persisted` trumps `project`
 - `project` trumps `authority`

This matters in particular for the error cases. In the current code we are trying to build all 3 cases and dismissing previous values if there's one with a higher priority. But at each there might be errors aborting the whole process.
If we make the priorities explicit, we don't even build other cases, and there are less possibilities for errors.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -239,27 +244,102 @@ async fn run_foreground_node(
...
+        let vault = Vault::create();
+        let authority_public_identity =
+            PublicIdentity::import(&hex::decode(authority_identity)?, vault).await?;
+
+        let own_cred = match &cmd.credential {
+            Some(cred_name) => {
+                let state = opts.state.credentials.get(cred_name)?;
+                Some(CredentialRetrieverType::FromPath(state))
+            }
+            None => None,
```
etorreborre: This seems to be the only case where we can end-up with a `TrustAuthorityConfig` which would have no `own_credential` setup. Is it a legitimate case? Is that going to fail anyway later in the process? If that is the case, maybe we should throw an error right away to let the user fix the configuration (and then we can remove the `Option` in `TrustAuthorityConfig`)

implementations/rust/ockam/ockam_api/src/config/cli.rs
etorreborre: I am also wondering if we really need to model a `TrustAuthorityConfig`. It seems to me that `TrustContextConfig` and `TrustAuthorityConfig` are always built and used together. Could we just have instead:
```rust
struct TrustContextConfig {
  id: String,
  authority_identity: PublicIdentity,
  credential_retriever_type: CredentialRetrieverType,
}
```

implementations/rust/ockam/ockam_api/src/credential_retriever.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+
+#[async_trait]
+impl CredentialRetriever for CredentialIssuerRetriever {
+    async fn retrieve(
+        &self,
+        identity: &PublicIdentity,
+        for_identity: &Identity,
+    ) -> Result<Credential, ockam_core::Error> {
```
etorreborre: One side question (because it's probably outside this PR scope): we return a `Result<Credential>`. Should we return instead a `Result<Option<Credential>>` where `Ok(None)` would mean I don't know anything about that identity? `Ok(Credential::default())` would mean `I know that identity and it has 0 attributes`. And finally `Err(...)` would mean `I can't produce a Credential because ...`.

implementations/rust/ockam/ockam_api/src/nodes/authority_node/configuration.rs
```diff
@@ -56,6 +59,11 @@ impl Configuration {
         self.project_identifier.as_bytes().to_vec()
     }
 
+    /// Return the trust context identifier as bytes
+    pub(crate) fn trust_context_identifier(&self) -> Vec<u8> {
```
etorreborre, line 63: Could we please stop using `Vec<u8>`, arbitrary bytes, when we know that we have a perfectly typed `String` to start with :-)? Why not create a `TrustContextIdentifier` instead? Newtypes can be made convenient in Rust with many libraries like [`derive_more`](https://jeltef.github.io/derive_more/derive_more) (which is quite old) or [`nutype`](https://www.greyblake.com/blog/nutype-the-newtype-with-guarantees) which is more recent.

implementations/rust/ockam/ockam_identity/src/trust_context.rs
```diff
@@ -0,0 +1,119 @@
+use async_trait::async_trait;
+
+use crate::{credential::Credential, error::IdentityError, Identity, PublicIdentity};
+use ockam_core::compat::{boxed::Box, string::String, sync::Arc};
+
+/// Trust Context is a set of information about a trusted authority
+#[derive(Clone)]
+pub struct TrustContext {
+    id: String,
```
etorreborre: It seems to me that we should be more specific here and write `id: IdentityIdentifier` or `id: PublicIdentity`. Because we are not using arbitrary strings to create a trust context, it looks like we are always using a specific identity. I think that this is important when we try to understand what exact ABAC rule applies when we then use a `subject.trust_context_id` and `resource.trusted_context_id`. Maybe if the `trusted_context_id` field was named `trusted_authority_id` or something like that, it would make the data provenance clearer.

implementations/rust/ockam/ockam_command/src/credential/issue.rs
```diff
@@ -1,24 +1,25 @@
...
+use ockam_identity::{credential::CredentialBuilder, PublicIdentity};
 
 #[derive(Clone, Debug, Args)]
 pub struct IssueCommand {
     #[arg(long = "as", default_value_t = default_identity_name())]
     pub as_identity: String,
 
     #[arg(long = "for", value_name = "IDENTITY_ID")]
-    pub for_identity: IdentityIdentifier,
+    pub for_identity: String,
```
etorreborre, line 22: I would suggest to put instead `for_identity: PublicIdentity` here and specify a `value_parser=parse_public_identity` in the `clap` macro attribute. This way the parser can be reused everywhere we pass a `PublicIdentity` on the command line and we can avoid the lines 53 to 59 and similar lines in `store.rs`, `verify.rs`.

This would also make the possibility to provide a clear error message like "expecting a hex-encoded public identity". It can be produced with `ockam identity show --full --output hex`. Since we are modifying the API of the command line that can help people who were used to pass a simple identity identifier.

etorreborre (commented):
@CAOakleyII I left lots of comments :-). I'm happy to discuss them live if anything is unclear (and be corrected if I am wrong in any of those)

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
polvorin: yes it's for backwards compatibility,  this way we would be able to deploy this on the cloud  and test _before_  ockam binary release.    Credentials will still have the `project_id` field that existing clients at that point will expect. 
I'd suggest changing `PROJECT_ID`  to `LEGACY_ID` or similar and add a comment

implementations/rust/ockam/ockam_api/src/config/cli.rs
polvorin: yes authority field is optional.   A node can have trust context and _no_ authority.   This means the node won't accept nor present credentials  (it doesn't trust any _authority_ to do sign that),  but still will have the default ABAC rules in place.   So only information that is explicity configured  at --trusted-identities is  used.   See test [here](https://github.com/build-trust/ockam/blob/725921dece54dc29727c39264bbf55debc204304/implementations/rust/ockam/ockam_command/tests/bats/trust.bats#L33)   (it's likely a bit different in this PR,  but the idea is the same).

This is different than having _no_  trust_context at all:  [in that case ](https://github.com/build-trust/ockam/blob/725921dece54dc29727c39264bbf55debc204304/implementations/rust/ockam/ockam_command/tests/bats/trust.bats#L22)no default ABAC rules are attached to any service,  anyone can call anyone.

Having an  authority _without_  a credential retriever is also possible.    It means the node can accept credentials signed by that authority,  but won't have a credential for his own.    So no MUTUAL credential exchange, only be on the receiving side of ONEWAY  ones.    This is a corner case and  I'd agree not a super useful one (but is what we currently have on project nodes:    they require credentials but do not present own, as they don't have right now).

implementations/rust/ockam/ockam_command/src/node/create.rs
polvorin: IIUC I agree here.    Basically  we want to transition away from "project"   to "trust_context".    On the command side,  we need to still accept  a project.json to not break things right now,   while alternatively _also_ accepting trust_context.    We also have even _another_  way on the command that is by _naming_  the project  (and assuming there is info for a project with that name on OCKAM_HOME).      Note in every case only one of the three options should be specified.  
I think to map that a nice way is to put these[ three argument together](https://github.com/build-trust/ockam/blob/725921dece54dc29727c39264bbf55debc204304/implementations/rust/ockam/ockam_command/src/util/api.rs#L355-L370) ,  then on _that_  struct,   have a[ function that returns a trust context config](https://github.com/build-trust/ockam/blob/725921dece54dc29727c39264bbf55debc204304/implementations/rust/ockam/ockam_command/src/util/api.rs#L473).   If something else was passed (lets say the legacy project.json),   we convert it to trust context.     *All*   commands that accepted --project or --project-path  before,   now accept this struct of either `--project | --project-path  or --trust-context`


note the transition from  project ->  trust context  _needs_ to be a  _bit_  leaky at the moment.    The project has the project'  address,  and we don't have a good place to put that yet,  it definitely won't go on trust_context,  it's more like a service list,  or alias list.

implementations/rust/ockam/ockam_api/src/nodes/authority_node/configuration.rs
polvorin: Agree,  but I see where this is coming from,  I'd let this to another PR.  There is a mix of Vec<u8>  and CowBytes everywhere,   this is transferred on Credentials  that define this to be bytes,   and stored on authenticated_table  that also has its own opinions.   
A futher PR that at least use String here  and convert when serialized/deserialized would be a good first step.  But I think long term we should also change these other places.. I don't think there is a point on keeping it open to bytes as we don't really support them anyway _latter_  (for example to use on an ABAC rule,  they _must_ be strings right now..)

implementations/rust/ockam/ockam_identity/src/trust_context.rs
```diff
@@ -0,0 +1,119 @@
+use async_trait::async_trait;
+
+use crate::{credential::Credential, error::IdentityError, Identity, PublicIdentity};
+use ockam_core::compat::{boxed::Box, string::String, sync::Arc};
+
+/// Trust Context is a set of information about a trusted authority
```
mrinalwadhwa: A trust context defines which authorities are trusted to attest to which attributes, within a context.

Our first implementation assumes that there is only one authority and it is trusted to attest to all attributes within this context.

implementations/rust/ockam/ockam_api/src/config/cli.rs
etorreborre: Removing the intermediate layer of `TrustAuthorityConfig` could still be a good idea. Maybe by introducing a `NoCredential` variant in `CredentialRetrieverType` and have `CredentialRetriever::retrieve` return an `Option`?

implementations/rust/ockam/ockam_command/src/node/create.rs
etorreborre: Yes @polvorin what you have in your PR is what I had in mind

implementations/rust/ockam/ockam_api/src/nodes/authority_node/configuration.rs
etorreborre: Yes this can be done in another PR. I was just hoping that when we introduce something new we do it the better way and then there is less work to retrofit all the rest. But maybe a full sweeping PR taking care of similar cases will be better.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
caoakleyii: Correct, backward compatibility, we do plan on removing it at some stage. 

I'll add comments and todo's around those areas for when we are ready.

implementations/rust/ockam/ockam_api/src/config/cli.rs
caoakleyii: Correct, this was mainly used for serialization & deserialization; I can still move this over to a factory implementation.

implementations/rust/ockam/ockam_api/src/config/cli.rs
caoakleyii: > Why do we need an Option here? Of all the usages of TrustContextConfig I can always see Some being passed.

I agree, with this current usage I don't believe there is a reason for it to be an option. I think there was something tied to the ABAC requirements; however I'm unsure exactly. @polvorin  do you remember why this was defined to be an Option?

caoakleyii: > I am also wondering if we really need to model a TrustAuthorityConfig. It seems to me that TrustContextConfig and TrustAuthorityConfig are always built and used together. Could we just have instead:


We will eventually have more than one TrustAuthority per TrustContext.

implementations/rust/ockam/ockam_command/src/node/create.rs
caoakleyii: No, it's an expected use case.

There's the mutual and one way credential exchange.

I know currently the orchestrator implementations do not have a credential to return.

implementations/rust/ockam/ockam_api/src/credential_retriever.rs
caoakleyii: I can understand that being better in the case of error handling and messaging.

However, I don't know if there's a current use case for it?

implementations/rust/ockam/ockam_api/src/config/cli.rs
caoakleyii: edit: apparently my github did not refresh with Pablo's responses :) excuse the ping.

implementations/rust/ockam/ockam_identity/src/trust_context.rs
caoakleyii: I would love a stricter definition on this field;

However, it's also meant to be a generic field. For example we use the project id, in the use case of an orchestrator based trust context.

In the cases when that isn't available, I've been using IdentityIdentifier as a string, however, it could also just as simple be a uuid/guid.

**The only usage of it right now is for ABAC rules.**

implementations/rust/ockam/ockam_command/src/credential/issue.rs
caoakleyii: I would love to do this; and that's how I originally started it.

However, I believe the only way to parse from a string to a Public Identity requires async calls and I was unable to see a way to use clap parsers with async functions.

etorreborre: You're absolutely right, that is not possible out of the box at the moment. _However_ the parsing of a `hex` identity as a `PublicIdentity` should be possible without needing any async. We only need to use `async` at the moment because of how we decided to design our interfaces to the `Vault`. This is something I would like to revisit after the "identity PR" is merged.

implementations/rust/ockam/ockam_command/src/node/create.rs
etorreborre: Ok, thanks for the confirmation

implementations/rust/ockam/ockam_command/src/credential/issue.rs
etorreborre: I think in the meantime it could still be a good idea to refactor a bit to avoid duplicating `pub async fn public_identity(&self) -> Result<PublicIdentity>` in several `credential/` files and also make sure we provide a good error message in case the `hex` decoding or the identity import fails (ideally with a unit test, I know I'm piling work onto you :-))

implementations/rust/ockam/ockam_api/src/credential_retriever.rs
etorreborre: I don't have enough background to weigh-in, maybe @polvorin can chime in. It's more a gut feeling that an API providing a `Credential` for an `Identity` might have to be able to say "no" without this being an error and the client having to know about error types. But we can think about this later, it's outside the scope of this PR.

implementations/rust/ockam/ockam_identity/src/trust_context.rs
etorreborre: Well it looks like I was wrong anyway to suggest this if a trust context can eventually encompass several authorities. But then the semantics of the ABAC rules are less clear to me. If a server has a resource with attribute trust_context_1 connected to authority 1 and 2, can a client still access the resource if it has a trust_context only connected to authority 2?

caoakleyii: > The semantics of the ABAC rules are less clear to me. If a server has a resource with attribute trust_context_1 connected to authority 1 and 2, can a client still access the resource if it has a trust_context only connected to authority 2?

That is a really good point, I would think the ID should be tied to the authority specifically, which changes some things.

I'll write a note/todo about that, in the structs so when we get to that point we will know to change it.

implementations/rust/ockam/ockam_api/src/credential_retriever.rs
```diff
@@ -0,0 +1,120 @@
...
+        debug!("Getting credential from : {}", &self.issuer.maddr);
+
+        let allowed = vec![identity.identifier().clone()];
+
+        let authority_tcp_session = match create_tcp_session(&self.issuer.maddr, &self.transport)
```
shanesveller: Is [let-else](https://doc.rust-lang.org/reference/statements.html#let-statements) with a destructuring left-hand-side any more attractive here and a few other spots? Does our MSRV permit us to use that?  It was introduced in 1.65.

implementations/rust/ockam/ockam_command/src/enroll.rs
```diff
@@ -47,7 +47,7 @@ async fn rpc(ctx: Context, (opts, cmd): (CommandGlobalOpts, EnrollCommand)) -> R
 }
 
 async fn run_impl(ctx: &Context, opts: CommandGlobalOpts, cmd: EnrollCommand) -> Result<()> {
-    let node_name = start_embedded_node(ctx, &opts, None).await?;
+    let node_name = start_embedded_node(ctx, &opts, None, None).await?;
```
shanesveller: Hopefully this function signature doesn't grow much wider - maybe an opportunity for builder pattern in a later refactor?

implementations/rust/ockam/ockam_command/src/node/start.rs
```diff
@@ -67,6 +67,7 @@ async fn run_impl(
         None,
         None, // No launch config available
         None,
+        None,
```
shanesveller: Any chance we could annotate more of the Nones with a quick comment, at least the newly introduced one?

implementations/rust/ockam/ockam_command/src/enroll.rs
caoakleyii: Agreed

implementations/rust/ockam/ockam_api/src/config/cli.rs
caoakleyii: Hmm, coming back to this -

this is probably something we can rehash later; The `CredentialRetriever` is defined with the `identity` crate purposefully, and `CredentialRetrieverType` is defined within the `api` crate purposefully to allow for `CredentialState`, `PathBuf`, and `MultiAddr` retrievers.

Due to the dependencies on allowing for no_std and not using multi_addr within `identity`

implementations/rust/ockam/ockam_command/src/node/create.rs
caoakleyii: I just want to note, with the `project_opts` approach this does change the function of `--project` on `ockam node create`

from being a path or a name. (and adds `--project-path`)

Unifying this, even though it's being deprecated is obviously ideal, but something to keep in mind for docs.

I'll be sure to update all instances within this repo/commit.

polvorin: yes :+1:  , lets unify on  `--project-path`   for that,    as you noted now in some places  `--project` is used,  in others `--project-path`,    and in other   `--project` refer to a different thing  (project by name).    It's all confusing,  I'll say it's a worthy change to have now,   and not *that*  bad at breaking things

caoakleyii:
```sh
# Example using Orchestrator to create a secure channel between two nodes
ockam project information --as-trust-context > trust_context.json

ockam identity create m1
ockam project enroll > m1.token
ockam project authenticate --identity m1 --trust-context ./trust_context.json --token $(cat m1.token)

ockam identity create m2
ockam project enroll > m2.token
ockam project authenticate --identity m2 --trust-context ./trust_context.json --token $(cat m2.token)

ockam node create n1 --identity m1 --trust-context ./trust_context.json

ockam node create n2 --identity m2 --trust-context ./trust_context.json

ockam secure-channel create --from /node/n1 --to /node/n2/service/api \
    | ockam message send hello --from /node/n1 --to -/service/uppercase
```

caoakleyii:
```sh
    # Example using a local identity and trust context to create a secure channel 

    ockam identity create i1
    ockam identity show i1 --full --encoding hex > i1.id

    ockam identity create i2
    ockam identity show i2 --full --encoding hex > i2.id

    ockam identity create identity_authority
    ockam identity show identity_authority --full --encoding hex > authority.id

    ockam credential issue --as identity_authority --for $(cat i1.id) --attribute city="New York" --encoding hex > i1.cred
    ockam credential store i1-cred --issuer $(cat authority.id) --credential-path i1.cred
    ockam credential show i1-cred --as-trust-context > i1-trust-context.json

    ockam credential issue --as identity_authority --for $(cat i2.id) --attribute city="Boston" --encoding hex > i2.cred
    ockam credential store i2-cred --issuer $(cat authority.id) --credential-path i2.cred
    ockam credential show i2-cred --as-trust-context > i2-trust-context.json

    ockam node create n3 --identity i1 --trust-context ./i1-trust-context.json

    ockam node create n4 --identity i2 --trust-context ./i2-trust-context.json

    ockam secure-channel create --from /node/n3 --to /node/n4/service/api \
        | ockam message send hello --from /node/n3 --to -/service/echo
```

polvorin (dismissed):

polvorin (approved):

PR #4591, opened 2023-04-03
test(rust): move bats tests artifacts to a non-temporary directory to keep them in case of failure

adrianbenavides:


implementations/rust/ockam/ockam_command/tests/bats/load/base.bash
```diff
@@ -65,7 +65,14 @@ setup_home_dir() {
   export OCKAM_HOME="$dir"
 }
 
+mkdir -p "$HOME/.bats-tests"
 teardown_home_dir() {
+  # If BATS_TEST_COMPLETED is not set, the test failed.
+  if [[ -z "$BATS_TEST_COMPLETED" ]]; then
+    # Copy the CLI directory to $HOME/.bats-tests so it can be inspected.
+    # For some reason, if the directory is moved, the teardown function gets stuck.
+    cp -a "$OCKAM_HOME" "$HOME/.bats-tests"
```
adrianbenavides, line 74: Would this do the trick on CI? @metaclips

etorreborre (approved):
Let's wait for @metaclips validation but that looks good to me. 
Question: how do we access those files once they have been created on CI?

implementations/rust/ockam/ockam_command/tests/bats/load/base.bash
metaclips: It should do, thanks.

metaclips:
> Let's wait for @metaclips validation but that looks good to me. Question: how do we access those files once they have been created on CI?

We could use GitHub artifacts https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

PR #4597, opened 2023-04-04
Suggest Rust contributors disable upgrade check via environment variable

shanesveller:
## Current behavior

Contributors may be unaware they can disable the built-in upgrade checks for the `ockam` CLI, which is a small opt-in optimization for developers during frequent iteration.

## Proposed changes

The Gradle task wrappers for `rust:build` and `rust:lint` will check for the presence of the environment variable `OCKAM_DISABLE_UPGRADE_CHECK` and that its value is the string form of `1`, `yes`, or `true`. If it is absent or set to another value, Gradle will emit a warning suggesting that the environment variable be set.

In the current form, no color highlights or delays are included so as not to introduce friction for an optional suggestion. I am very new to use of Gradle and would welcome any suggested improvements.

I feel this in-line suggestion would benefit from a permalink to an entry on the [docs site](https://docs.ockam.io/reference/command) or the [command manual](https://command.ockam.io/manual/) but I don't currently see any content around this envvar.

## Checks

### DIFF
--- a/implementations/rust/build.gradle
+++ b/implementations/rust/build.gradle
@@ -10,6 +10,13 @@ ext {
   // Are we running as part of Continuous Integration?
   ci = System.getenv('CI') == 'true'
 
+  // Upgrade checks are not necessary during CLI development
+  // Ockam will also receive anonymized statistics from GitHub based on these checks
+  upgradeCheckEnvVar = 'OCKAM_DISABLE_UPGRADE_CHECK'
+  upgradeCheckDisabledValues = ['1', 'true', 'yes']
+  upgradeCheckValue = System.getenv(upgradeCheckEnvVar)
+  upgradeCheckDisabled = upgradeCheckDisabledValues.contains(upgradeCheckValue)
+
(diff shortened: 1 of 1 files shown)


implementations/rust/build.gradle
```diff
@@ -65,11 +70,16 @@ task lint_cargo_deny {
   }
 }
 
+task lint_environment_variables {
+  logger.warn("Contributors: Please set the environment variable {} to {}", upgradeCheckEnvVar, upgradeCheckDisabledValue)
```
shanesveller: This could probably benefit from printing out some further rationale or a link to outside documentation that gives more context (like the numbered compiler error codes in `rustc`).

etorreborre (dismissed):

implementations/rust/build.gradle
etorreborre: Yes a small explanation could be a good idea

mrinalwadhwa (approved):

PR #4604, opened 2023-04-04
feat(elixir): api to show node information, currently only version

hairyhum:
API worker should be initialized with the information to show This allows the managing application to chose which value to use for version

A step towards exposing current node version from Orchestrator nodes.

## Checks


implementations/elixir/ockam/ockam_services/test/services/node_info_api_test.exs
```diff
@@ -0,0 +1,20 @@
...
+
+  test "node info with version" do
+    {:ok, address} = NodeInfo.create(info: %NodeInfo.Info{version: "foo"})
+
+    {:ok, response} = ApiClient.sync_request(:get, "", "", [address])
+
+    assert {:ok, %NodeInfo.Info{version: "foo"}, ""} = NodeInfo.Info.decode(response.body)
+  end
+
+  test "node info without version crashes" do
```
shanesveller, line 15:
```suggestion
  test "node info without version returns error" do
```

Or "does not crash"

implementations/elixir/ockam/ockam_services/lib/services/api/node_info_api.ex
```diff
@@ -0,0 +1,63 @@
...
+
+  alias Ockam.API.Request
+  alias Ockam.Services.API.NodeInfo.Info
+
+  def node_info(worker, timeout \\ 5000) do
+    Ockam.Worker.call(worker, :get, timeout)
+  end
+
+  @impl Ockam.Worker
+  def setup(options, state) do
```
shanesveller, line 32: Zero functionality difference and no change necessary, but stylistically I probably would've written these as separate function heads and pattern match there. Do we have an internal style guide for that kind of super subjective stuff?

implementations/elixir/ockam/ockam_services/test/services/node_info_api_test.exs
hairyhum: The process does crash though. `create/1` function returns error. I can validate that it crashes in the test.

implementations/elixir/ockam/ockam_services/lib/services/api/node_info_api.ex
hairyhum: Not sure I follow. What can you pattern match in the function head? I don't think it's possible to pattern match a proplist to get `:info` from the options.

implementations/elixir/ockam/ockam_services/test/services/node_info_api_test.exs
shanesveller: Oh! That's an unusual design, then, but if we want to uphold that behavior as an important implementation detail, is there any way we can get a pid out somewhere in these the steps and then check for `Process.alive?` at the end?

implementations/elixir/ockam/ockam_services/lib/services/api/node_info_api.ex
shanesveller: 🤦🏻  I totally overlooked it was a Keyword, sorry!

shanesveller (approved):
LGTM, but please seek a second opinion if you have time / feel it's worthwhile. I also probably don't count for an approver yet.

mszpakowski (approved):

Issue #4479, opened 2023-03-16
ockam run - a single command to run many ockam “create” commands


mrinalwadhwa:
To remove the need for typing many commands in an exact order

ockam run - a single command to run many ockam “create” commands where what is created is specified in a declarative file.

This example series of commands becomes:
```
ockam node create telegraf --project project.json --enrollment-token $OCKAM_TELEGRAF_TOKEN
ockam policy set --at telegraf --resource tcp-inlet --expression '(= subject.component "influxdb")'
ockam tcp-inlet create --at /node/telegraf --from 127.0.0.1:8087 --to /project/default/service/forward_to_influxdb/secure/api/service/outlet


ockam node create influxdb --project project.json --enrollment-token $OCKAM_INFLUXDB_TOKEN
ockam policy set --at influxdb --resource tcp-outlet --expression '(= subject.component "telegraf")'
ockam tcp-outlet create --at /node/influxdb --from /service/outlet --to 127.0.0.1:8086
ockam forwarder create influxdb --at /project/default --to /node/influxdb
```

becomes `ockam.yaml`

```yaml
projects:
  default:
    # ...
  
nodes:
  telegraf:
    enrollment-token: $OCKAM_TELEGRAF_TOKEN
    tcp-inlets:
      telegraf:
        from: 127.0.0.1:8087
        to: /project/default/service/forward_to_influxdb/secure/api/service/outlet
        access_control: '(= subject.component "influxdb")'

  influxdb:
    enrollment-token: $OCKAM_INFLUXDB_TOKEN
    tcp-outlets:
      influxdb:
        from: /service/outlet
        to: 127.0.0.1:8086
        access_control: '(= subject.component "telegraf")'
    forwarders:
      influxdb:
        at: /project/default
```

you write that file and then run:

```
$ ockam run
```

We may allow multiple yaml files to be combined - for example project.yml + ockam.yml in some order of presidence.

The YAML may reference the state already in ~/.ockam - for example /project/default or an identity already defined.

Many people don’t like YAML because of its ambiguity so we may ~consider things like cue https://cuelang.org/~ - turns out cue doesn’t have a rust impl yet. TOML and JSON may also be supported as options, but I think the succinctness of YAML has some advantages here.

mrinalwadhwa:
This is similar to docker compose files
https://github.com/compose-spec/compose-spec/blob/master/spec.md

The `depends_on` constraint would be key to deciding on the order in which we create this.
[...]

PR #4605, opened 2023-04-05
feat(rust): add a limited version of the `ockam run` command

adrianbenavides:
Ref: https://github.com/build-trust/ockam/issues/4479

Limitations:
- Supports creation of nodes, tcp-inlets/outlets, policies on tcp-inlets/outlets, and forwarders
- Supports dependencies only between nodes
- The config file is mapped to a Hashmap, which doesn't keep the file's explicit structure. We'd like to honor the explicit order of the commands
- Assumes `enroll` is done and the needed projects/spaces exist

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/run/parser.rs
+++ b/implementations/rust/ockam/ockam_command/src/run/parser.rs
@@ -1,0 +1,418 @@
+use crate::Result;
+use duct::Expression;
+use ockam_api::cli_state::CliState;
+use ockam_core::compat::collections::HashMap;
+use once_cell::sync::Lazy;
+use serde::Deserialize;
+use std::collections::{BTreeMap, HashSet, VecDeque};
+use std::fmt::Debug;
+use std::path::Path;
+use tracing::debug;
+
+pub struct ConfigRunner {
+    commands_sorted: Vec<ParsedCommand>,
+    commands_index: BTreeMap<String, usize>,
+}
+
+#[derive(Clone)]
+pub struct ParsedCommand {
+    pub id: String,
+    pub depends_on: Option<String>,
+    pub cmd: Expression,
+}
+
+impl ConfigRunner {
+    fn new() -> Self {
+        Self {
+            commands_sorted: vec![],
+            commands_index: Default::default(),
+        }
+    }
+
+    pub fn go(cli_state: &CliState, path: &Path) -> Result<()> {
+        let mut cr = Self::new();
+        cr.parse(cli_state, path)?;
+        cr.run()?;
+        Ok(())
+    }
+
+    fn parse(&mut self, cli_state: &CliState, path: &Path) -> Result<()> {
+        let config = std::fs::read_to_string(path)?;
+        let config: Config = serde_yaml::from_str(&config)?;
+        let mut visited = HashSet::new();
+        let mut nodes = VecDeque::new();
+        for (name, node) in config.nodes {
+            nodes.push_back((name, node));
+        }
+        while let Some((name, node)) = nodes.pop_front() {
+            // If the node depends on another node, check if that node has been parsed.
+            if let Some(depends_on) = &node.depends_on {
+                // If the dependency has been visited already but not
+                // parsed, we have a circular dependency.
+                if visited.contains(depends_on) {
+                    return Err(anyhow::anyhow!(
+                        "Circular dependency detected: {} -> {}",
+                        depends_on,
+                        name
+                    )
+                    .into());
+                }
+                // If the dependency has been parsed, remove it from the control
+                // vector and proceed with the current node.
+                if self
+                    .commands_index
+                    .contains_key(&format!("node/{depends_on}"))
+                {
+                    visited.remove(depends_on);
+                }
+                // If the dependency has not been parsed, push the current
+                // node back to the queue and continue with the next one.
+                if !self
+                    .commands_index
+                    .contains_key(&format!("node/{depends_on}"))
+                {
+                    visited.insert(name.clone());
+                    nodes.push_back((name, node));
+                    continue;
+                }
+            }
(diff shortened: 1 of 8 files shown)


implementations/rust/ockam/ockam_command/src/run/parser.rs
```diff
@@ -0,0 +1,259 @@
...
+}
+
+/// The config structure will be a yml file with the following structure:
+/// ```yml
+/// nodes:
+///   telegraf:
+///     enrollment-token: $OCKAM_TELEGRAF_TOKEN
+///     tcp-inlets:
+///       telegraf:
+///         from: 127.0.0.1:8087
```
shanesveller: Based on how often YAML parsing has injured itself in its confusion for string-like values that start with an integer, should we document these as:

```suggestion
///         from: '127.0.0.1:8087'
```

I've had some languages' parsers try to give me a float from IP addresses before.

adrianbenavides: Good point, I'll do that, thanks for the heads up!

implementations/rust/ockam/ockam_command/src/run/parser.rs
```diff
@@ -0,0 +1,323 @@
...
+
+/// Defines the structure of a forwarder in the config file.
+#[derive(Debug, Deserialize)]
+pub struct ForwarderConfig {
+    pub at: String,
+}
+
+static BINARY_PATH: Lazy<String> = Lazy::new(|| {
+    if let Ok(p) = std::env::var("OCKAM") {
+        p
```
davide-baldo: I think you were looking for the path of the ockam binary, in this case we can look for the first argument instead of using an environment variable.
```
    if let Some(zero_arg) = std::env::args().next() {
        println!("zero_arg: {}", zero_arg);
    }
```
It can be empty in a very screwed scenario, when the caller uses low-level APIs to run the process and overwrites the first argument, IMHO we can just fail if the argument is not available.

implementations/rust/ockam/ockam_command/src/run/parser.rs
```diff
@@ -0,0 +1,323 @@
...
+        }
+        Ok(())
+    }
+}
+
+/// The config structure will be a yml file with the following structure:
+/// ```yml
+/// nodes:
+///   telegraf:
+///     enrollment-token: $OCKAM_TELEGRAF_TOKEN
```
davide-baldo, line 103: I think `$OCKAM_TELEGRAF_TOKEN` is a bit misleading since we are not resolving secrets and/or env. variables yet

implementations/rust/ockam/ockam_command/src/run/parser.rs
```diff
@@ -0,0 +1,323 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_parse_config_with_depends_on() {
+        let config = r#"
+            nodes:
```
davide-baldo, line 304: I can see HCL as another suitable candidate for commands representation

implementations/rust/ockam/ockam_command/src/run/parser.rs
```diff
@@ -0,0 +1,323 @@
...
+        let config: Config = serde_yaml::from_str(&config)?;
+        let mut nodes = VecDeque::new();
+        for (name, node) in config.nodes {
+            nodes.push_back((name, node));
+        }
+        while let Some((name, node)) = nodes.pop_front() {
+            // If the node depends on another node, check if that node has been parsed.
+            if let Some(depends_on) = &node.depends_on {
+                // If the node hasn't been parsed, push it back to the queue.
+                if !self
```
davide-baldo, line 70: What about reciprocal dependencies?
I would at least panic or better return an error

davide-baldo (changes requested):

implementations/rust/ockam/ockam_command/src/run/parser.rs
adrianbenavides: I think I have an idea to control this without complicating the logic too much... I'll try it out.

implementations/rust/ockam/ockam_command/src/run/parser.rs
adrianbenavides: The current structure is going to change in the next iteration, but as soon as we have that part more stabilized, we could add support to any format we want I think.

implementations/rust/ockam/ockam_command/src/run/parser.rs
adrianbenavides: True, I just copied from the issue. I'll note this for the next iteration, thanks!

implementations/rust/ockam/ockam_command/src/run/parser.rs
adrianbenavides: oh, I didn't know about this, I'll integrate it 🙏

implementations/rust/ockam/ockam_command/src/run/parser.rs
adrianbenavides: I've added a recursive dependency detection mechanism. You can check the tests for more details.

implementations/rust/ockam/ockam_command/src/run/parser.rs
adrianbenavides: I'll try to fix this in a separate PR this week. Let's keep it for now in the docs, as it's going to be used internally only.

implementations/rust/ockam/ockam_command/src/run/parser.rs
davide-baldo: nice!

davide-baldo (dismissed):

mrinalwadhwa (approved):

PR #4617, opened 2023-04-06
feat: prototype command to check orchestrator nodes versions

hairyhum:
Add `ockam orchestrator info` command which requires `--tcp-address` and `--identity-id` arguments and prints version from `node_info` api.

Ideally we should use orchestrator and project information from the cli environment.

@adrianbenavides @CAOakleyII can you help me with that?

## Checks


adrianbenavides (commented):
It looks good. I'll add a couple of changes to handle the default values. Other than that, should we reconsider the command name? We have to keep in mind that whatever we choose, it must not collide with or resemble the `ockam --version` command.

etorreborre:
@adrianbenavides @CAOakleyII @hairyhum I can pick-up this PR to finish it. I think there are a few remaining questions:

 - what the command name should be. I propose `ockam orchestrator show-versions` instead to somehow keep the `noun/verb` structure (I also wonder if `show-versions` should be a `project` sub-command instead of creating a brand new command)
 - the output as json data. I propose to do a normal output using the terminal functions and output the json equivalent if `--output json` is enabled
 - the `version` field in the json data. Should it be `orchestrator_version` instead to be more specific?

hairyhum:
>     * what the command name should be. I propose `ockam orchestrator show-versions` instead to somehow keep the `noun/verb` structure (I also wonder if `show-versions` should be a `project` sub-command instead of creating a brand new command)

I think we agreed to `ockam show` and keep it hidden from the `help`

>     * the output as json data. I propose to do a normal output using the terminal functions and output the json equivalent if `--output json` is enabled

Fine with me as long as it can output json
 
>     * the `version` field in the json data. Should it be `orchestrator_version` instead to be more specific?

It should be a `controller_version` then.

caoakleyii:
Just as an FYI; I created a rough draft PR, to how to do this with the OrchestratorAPIBuilder - https://github.com/build-trust/ockam/pull/4847

There's a couple smaller changes that need to be made to the builder to allow this - but it cleans up the command code


```rs
    /// ockam show
    let mut orchestrator_api_builder =
        OrchestratorApiBuilder::new(&ctx, &opts, &cmd.trust_context_opts);

    if let Some(identity_id) = cmd.remote_identity_id {
        orchestrator_api_builder.with_trusted_identities(vec![identity_id]);
    }

    let mut api = orchestrator_api_builder
        .as_identity(cmd.cloud_opts.identity.clone())
        .with_endpoint(OrchestratorEndpoint::Controller)
        .with_new_embbeded_node()
        .await?
        .build(&MultiAddr::from_str("/service/version_info")?)
        .await?;

    let resp: VersionInfo = api.request_with_response(Request::get("")).await?;

    println!(
        "{{\"version\":\"{}\",\"project_version\":\"{}\"}}",
        resp.version, resp.project_version
    );
```

adrianbenavides:
I've pushed the needed changes to retrieve the controller's versions.
```sh
$ ockam orchestrator show
     > Controller version: c91c77cea12ea154871e6673c67a0c0419f7fc81-2023-06-30
     > Project version: bfa63a61e928f22337fe4b2caf254371bdac3ac3-2023-06-28

$ ockam project show --output json
{"controller_version":"c91c77cea12ea154871e6673c67a0c0419f7fc81-2023-06-30","project_version":"bfa63a61e928f22337fe4b2caf254371bdac3ac3-2023-06-28"}
```

Note that I renamed the command, adding the `orchestrator` subject. It was very confusing to have a `show` command in the root.

implementations/rust/ockam/ockam_command/src/message/send.rs
```diff
@@ -38,7 +38,7 @@ pub struct SendCommand {
     #[arg(long)]
     pub hex: bool,
 
-    /// Override Default Timeout
+    /// Override default timeout (in seconds)
```
etorreborre, line 41: One finesse we could implement at some stage would be to have a value parser to `Duration` so that we can pass `10s`, `1m`, `500ms` on the command line

implementations/rust/ockam/ockam_command/src/orchestrator/show.rs
```diff
@@ -0,0 +1,104 @@
...
+            .with_identities_vault(identities_vault)
+            .with_identities_repository(identities_repository)
+            .build(ctx)
+            .await
+            .into_diagnostic()?
+    };
+    let identifier = opts.state.identities.default()?.identifier();
+    let secure_channel_options = SecureChannelOptions::new()
+        .with_trust_policy(TrustIdentifierPolicy::new(controller_identifier))
+        .with_timeout(Duration::from_secs(50));
```
etorreborre: Shouldn't the timeout argument be used here as well? The user interpretation of the `timeout` argument is probably going to be "how long do I give this command to execute before giving up?". If we reuse the timeout there we potentially double the max execution time but at least there is an arg to play with.

implementations/rust/ockam/ockam_command/src/orchestrator/show.rs
```diff
@@ -0,0 +1,104 @@
...
+        .await
+        .into_diagnostic()?
+        .body();
+    let mut dec = Decoder::new(&buf);
+    let hdr = dec.decode::<Response>().into_diagnostic()?;
+    if hdr.status() == Some(Status::Ok) {
+        let info = dec.decode::<VersionInfo>().into_diagnostic()?;
+        opts.terminal
+            .stdout()
+            .plain(
```
etorreborre: This double output, text + json, is great!

etorreborre (dismissed):
All good. My only concern is the hardcoded timeout to establish a secure channel with the controller.

implementations/rust/ockam/ockam_command/src/orchestrator/show.rs
adrianbenavides: Yeah, I found this fixed timeout weird. If we reuse it here, perhaps we should increment the default value of 10 seconds up to 30.

implementations/rust/ockam/ockam_command/src/message/send.rs
adrianbenavides: That would be really nice! I'll create an issue for this.

etorreborre (dismissed):
🐰

mrinalwadhwa (dismissed):
Please don't merge this yet - I have a few points we should discuss before that.

adrianbenavides:
I've moved the orchestrator version data as part of the `show` command output.

```sh
$ ockam status
$ ockam status --output json
```

etorreborre (dismissed):

etorreborre (approved):

PR #4623, opened 2023-04-07
fixes from running bats test

polvorin:
bats tests on `message`,  `portals`,  `projects` ,  `trust_context` and `use-cases` suites where broken,  this fix them.  

* Use the default project configured on the `TrustContextConfigBuilder` , if no other was given.
* Do not enforce  `trust_context_id`  on abac rules  _yet_   , as that requires coordination to update existing authorities that don't return that until this get merged
*  `/project/name`  multiaddresses are  resolved from some lookup mechanism.    This data must be obtained from the project.json file  (it won't be there on  OCKAM_HOME   if the node is started on a different OCKAM_HOME,  or different machine than the one where the project was created)

TODO:
one trust_context case nees :eyes:  as it's not fixed.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -148,10 +148,15 @@ impl NodeManager {
                     Some(e) => e.clone(),
                     None => and([
                         eq([ident("resource.project_id"), ident("subject.project_id")]), // TODO: DEPRECATE - Removing PROJECT_ID attribute in favor of TRUST_CONTEXT_ID
-                        eq([
-                            ident("resource.trust_context_id"),
-                            ident("subject.trust_context_id"),
-                        ]),
+                                                                                         /*
+                                                                                         * TODO: replace the project_id check for trust_context_id.  For now the
+                                                                                         * existing authority deployed doesn't know about trust_context so this is to
+                                                                                         * be done after updating deployed authorities.
+                                                                                         eq([
+                                                                                             ident("resource.trust_context_id"),
+                                                                                             ident("subject.trust_context_id"),
+                                                                                         ]),
+                                                                                         */
                     ]),
                 };
                 self.policies.set_policy(r, a, &fallback).await?
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -13,18 +13,18 @@ use std::{
 };
 use tracing::error;
 
+use crate::node::util::{add_project_info_to_node_state, init_node_state, spawn_node};
 use crate::secure_channel::listener::create as secure_channel_listener;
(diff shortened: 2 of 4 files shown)


mrinalwadhwa:
Before `node create` we seem to always do `project authenticate`

```
ockam project authenticate --identity influxdb --token $OCKAM_INFLUXDB_TOKEN --project-path project.json
ockam node create influxdb --project project.json --identity influxdb
```

could `project authenticate` setup the default project so its not needed in `node create`

caoakleyii:
> could project authenticate setup the default project so its not needed in node create

This should already be implemented.

Edit: correction - it's only being done for `ockam enroll` commands.

caoakleyii:
> one trust_context case nees 👀 as it's not fixed.

I believe this is because the `ockam project enroll` is not creating a credential with the trust_context_id attributes

polvorin:
> Before `node create` we seem to always do `project authenticate`
> 
> ```
> ockam project authenticate --identity influxdb --token $OCKAM_INFLUXDB_TOKEN --project-path project.json
> ockam node create influxdb --project project.json --identity influxdb
> ```
> 
> could `project authenticate` setup the default project so its not needed in `node create`

It can be done on _some_  situations,  but think it's not a good idea:

1)  `node create`  *still*  need to be given the project.   If we import the project information on some previous command (I think a bad idea, see next),    what we gain is we can give the project name  in    `--project`  , instead of  `--project-path`, but not that we use it by default on `node create`.   Making a plain `node create`   _use_  a project  if there is one on OCKAM_HOME (this means setting the default ABAC rules,  exchange credentials, etc)  breaks all the simpler cases where a node is started and don't want to do that.

2)  `project authenticate`   is not always needed.    It's needed when the member do the final step of authenticating itself through an enrollment token or okta for example,   but it is not needed for the cases where the member' identity was added _directly_  the the enroller.

3) there is no "project"  in  `project authenticate`  really :upside_down_face: .    `project authenticate`   authenticates a given identity with some  _authenticator|online authority_  .  Might be a good idea to rename that `project authenticate`  to something else, and we had that on the plan  (same thing for  `project enroll`).    For now and for backwards compatibility  this was not renamed plus also  keep accepting the project.json  (just to extract the authority' info from it)

4) In more broad terms,  the whole  `/project/xxx`  multiaddr is confusing,  we need to get rid of it at some point.   If the node is started using projectA'  info, then you _always_  use `project/projectA`  ,   even in the case where there could exist a projectB  that the identity also has access to,   using `/project/projectB`  won't work  as the node code is not prepared to that  (to know what credential to present,  to know how to split  authenticated identities by project,   to know what default rules to put on  portals  (require projA or project B? membership, etc).  All is single-project  (I think that's  ok).     What the `/project/xxx`  does is to define an alias to some address.    We could perhaps  (not now)  change that to  /alias/project  for example for this case.   And other alias be added.     Or perhaps to specific _services_  at some address

PR #4629, opened 2023-04-09
fix(rust): add `verbose` argument when creating an authority node

adrianbenavides:
Nodes crated with the `ockam authority create` command were not passing the `verbose` argument to the subprocess, creating a node with no logs.


mrinalwadhwa (dismissed):

mrinalwadhwa:
This failed in the queue https://github.com/build-trust/ockam/actions/runs/4651132636/jobs/8230547230

adrianbenavides:
> This failed in the queue https://github.com/build-trust/ockam/actions/runs/4651132636/jobs/8230547230

Looks like a problem with how the test is written. It only fails when running it in parallel.

Update: seems like the name of the `PROJECT_JSON_PATH` file has some weird effect on the command.

mrinalwadhwa (approved):

PR #4635, opened 2023-04-10
refactor(rust): format `Cargo.toml` files consistently

### LINKED ISSUE #2129: Fix formatting of dependencies in our Cargo.toml files
Many of our Cargo.toml files have some slightly odd formatting for dependency lists. Often this is either:

1. Spaces prior to a `,` for example https://github.com/ockam-network/ockam/blob/17772ae654d7d86acd5ea3e3e359225c7f93e81d/implementations/rust/ockam/ockam/Cargo.toml#L77 has `version = "0.35.0" ,`

2. Trailing whitespace before the closing `}`, for example https://github.com/ockam-network/ockam/blob/17772ae654d7d86acd5ea3e3e359225c7f93e81d/implementations/rust/ockam/ockam/Cargo.toml#L100-L101

This bugs me a very little bit whenever I notice it, but I generally don't touch it as to not include unrelated things in the diff and whatnot.

It would be nice if someone went through our Cargo.toml files and fixed these kinds of issues. (Eventually maybe we should check toml formatting in CI, but for now that seems like overkill)

If you have `ripgrep` installed and have ockam located in `~/src/ockam`, you can find the set of issues using a command like:

```
rg -e '("\s+,|\s{2,}\})' ~/src/ockam/implementations/rust/ockam/*/Cargo.toml
```

This doesn't include the Cargo.tomls of example crates, but none of them seem to have this issue.

adrianbenavides:
Fixes https://github.com/build-trust/ockam/issues/2129

The formatting was done with `dprint`, and we could use it to enforce a consistent format of Cargo.toml files in CI as well: https://dprint.dev/ci/. This integration could be done separately.


adrianbenavides:
@metaclips what do you think about integrating dprint? It also supports markdown files. I did a quick test and it fixes a lot of problems.

metaclips:
> @metaclips what do you think about integrating dprint? It also supports markdown files. I did a quick test and it fixes a lot of problems.

Nice, we should integrate it into CI. I believe EditorConfig CI can also solve this (might need some few tweaks)

metaclips (approved):

metaclips:
Dprint also ensures alphabetic ordering of dependencies, adding that with formating is a +1.  I created an issue on using Dprint in CI https://github.com/build-trust/ockam/issues/4637

shanesveller:
I've not used either `dprint` or this one, but I came across [Taplo](https://taplo.tamasfe.dev/cli/usage/formatting.html) ([options](https://taplo.tamasfe.dev/configuration/formatter-options.html)) while doing some side-project research and found it quite interesting.

PR #4636, opened 2023-04-10
test(rust): reuse default space/project and default arguments to simplify test

adrianbenavides:
This test doesn't exit properly when it fails, blocking the CI pipeline. As we know, this can happen if the test resources are not freed up. I was able to consistently reproduce the blocking issue making the `project delete` command fail.

This PR simplifies the test so it doesn't have to create/delete new space/project. It also updates the test to make use of the new default values available in some of the used commands.


metaclips (approved):
Thanks for the quick fix @adrianbenavides

PR #4646, opened 2023-04-10
test(elixir): update ports and process flags to prevent clashes

hairyhum:
reduce possibility of port clash in tests

## Checks

### DIFF
--- a/implementations/elixir/ockam/ockam_kafka/test/interceptor/outlet_manager_test.exs
+++ b/implementations/elixir/ockam/ockam_kafka/test/interceptor/outlet_manager_test.exs
@@ -67,29 +67,29 @@ defmodule Ockam.Kafka.Interceptor.OutletManager.Test do
     ssl = false
     ssl_options = []
 
-    base_port = 7000
+    base_port = 11_000
     allowed_ports = 10
     base_route = []
 
     start_supervised!({InletManager, [base_port, allowed_ports, base_route, outlet_prefix]})
     start_supervised!({OutletManager, [outlet_prefix, ssl, ssl_options]})
 
-    Ockam.Transport.TCP.start(listen: [port: 4000])
+    Ockam.Transport.TCP.start(listen: [port: 12_000])
 
     InletManager.set_inlets(InletManager, [1])
 
-    ## Configure outlet to connect to ockam port 4000
+    ## Configure outlet to connect to ockam port 12_000
     OutletManager.set_outlets(OutletManager, [
       %Outlet{
         outlet_prefix: outlet_prefix,
         node_id: "1",
         target_host: "localhost",
-        target_port: 4000
+        target_port: 12_000
       }
     ])
 
     ## Connect to inlet port
-    {:ok, client} = Ockam.Transport.TCP.Client.create(destination: {"localhost", 7001})
+    {:ok, client} = Ockam.Transport.TCP.Client.create(destination: {"localhost", 11_001})
 
     {:ok, "echo"} = Ockam.Services.Echo.create(address: "echo")
 
--- a/implementations/elixir/ockam/ockam/mix.exs
+++ b/implementations/elixir/ockam/ockam/mix.exs
@@ -86,7 +86,7 @@ defmodule Ockam.MixProject do
     [
       credo: "credo --strict",
       docs: "docs --output _build/docs --formatter html",
-      "test.cover": "test --no-start --cover",
+      "test.cover": "test --cover",
       "lint.format": "format --check-formatted",
       "lint.credo": "credo --strict",
       "lint.dialyzer": "dialyzer --format dialyxir",
--- a/implementations/elixir/ockam/ockam_healthcheck/mix.exs
+++ b/implementations/elixir/ockam/ockam_healthcheck/mix.exs
@@ -80,7 +80,7 @@ defmodule Ockam.Healthcheck.MixProject do
       "lint.credo": "credo --strict",
       "lint.dialyzer": "dialyzer --format dialyxir",
       lint: ["lint.format", "lint.credo"],
-      "test.cover": "test --no-start --cover"
+      "test.cover": "test --cover"
     ]
   end
 end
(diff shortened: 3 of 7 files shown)


implementations/elixir/build.gradle
```diff
@@ -85,7 +85,7 @@ apps.each { app ->
     dependsOn checkLocalHexAndRebar
     doLast {
       exec {
-        commandLine mix('do deps.get, test')
+        commandLine mix('do deps.get, test --max-cases=1')
```
shanesveller: Is it worth it to try with just the port numbers changes and the rest of the files in this PR first before introducing this one? I'd usually try to localize which test modules use some kind of global state that can be race-y and mark them `async: false` with a doc comment about why.

Does gradle itself parallelize the different sub-packages' test suites at all? If not, I'm not sure how we would get collisions across separate test suites, and when I spot checked the changed modules here, already none of them are using `use ExUnit.Case, async: true` so they shouldn't run parallel to each other within the suite either.

hairyhum: I don't think the main issue is with `async: true` tests, rather different test cases impacting each other.
Gradle runs different app tests in different containers (in parallel), so the only overlap is between different test modules in the same app.

implementations/elixir/ockam/ockam_kafka/test/interceptor/outlet_manager_test.exs
```diff
@@ -67,29 +67,29 @@ defmodule Ockam.Kafka.Interceptor.OutletManager.Test do
...
-    base_port = 7000
+    base_port = 11_000
     allowed_ports = 10
     base_route = []
 
     start_supervised!({InletManager, [base_port, allowed_ports, base_route, outlet_prefix]})
     start_supervised!({OutletManager, [outlet_prefix, ssl, ssl_options]})
 
-    Ockam.Transport.TCP.start(listen: [port: 4000])
+    Ockam.Transport.TCP.start(listen: [port: 12_000])
```
polvorin, line 77: not for now but if the problem is the port collision,  we could make the  port optional and let the OS choose one for us, returning it from the start() function? .  So the rest of the test can refer to it.    Seems useful only for tests but could be an easy win?

polvorin (dismissed):

implementations/elixir/build.gradle
mszpakowski: I too have trouble understanding what is the reason for setting `max_cases` to 1. Tests within one module will still be executed the same way as before. AFAIK `max_cases` only impacts `async: true` tests (i.e. how many `async: true` test modules run in parallel)

implementations/elixir/ockam/ockam_kafka/test/interceptor/outlet_manager_test.exs
hairyhum: For this one we can do that, for inlet-outlet manager test we set up port range, which is more complex.
But this change is a quick-fix so we don't get the flaky tests so often.

implementations/elixir/build.gradle
hairyhum: > AFAIK max_cases only impacts async: true tests (i.e. how many async: true test modules run in parallel)

According to the `mix test` docs, it runs different modules in parallel `--max-cases - sets the maximum number of tests running asynchronously. Only tests from different modules run in parallel. Defaults to twice the number of cores`

Which is what we've seen. We don't use `async: :true` in the tests which had clashes.

mszpakowski: I apologize if I'm being annoying, it's not my intention. I just want to make sure we're on the same page.

> According to the `mix test` docs, it runs different modules in parallel `--max-cases - sets the maximum number of tests running asynchronously. Only tests from different modules run in parallel. Defaults to twice the number of cores`
>
> We don't use `async: :true` in the tests which had clashes.

I think it's safe to assume that the modules that run in parallel are only `async: true` modules (despite the docs not mentioning it explicitly). If `async: false` modules were to be run in parallel it would defeat their purpose.

And a glance at the control flow for the ExUnit runner seems to support my assumption:
https://github.com/elixir-lang/elixir/blob/7e4fbe657dbf9c3e19e3d2bd6c17cc6d724b4710/lib/ex_unit/lib/ex_unit/runner.ex#L95-L128
It first takes all `async: true` modules running up to the number of `max_cases` of them in parallel. And when they're finished it runs `async: false` modules one by one.

shanesveller: What Michał raises is my past experience too. The only thing that may contradict this is any global mutable state (such as spawned supervised processes for the `Ockam.Worker`s) that aren't sandboxed per-test-process like Ecto or `spawn_supervised` do on our behalf, and aren't being exhaustively cleaned away after the test bodies. I haven't seen enough yet to know if that exception is likely in our case.

hairyhum: That's an interesting point. The tests where port clash happened did use `spawn_supervised` and did not use `async: true` but still got port clash.
If that's not because tests running in parallel, I'm not sure how that's possible.
I will roll back `--max-cases` for now.

shanesveller (dismissed):
Should we revise the PR body or the merge message before it goes through, since the max-cases got moved out for now?

shanesveller (approved):

mrinalwadhwa (approved):

implementations/elixir/build.gradle
mszpakowski: > That's an interesting point. The tests where port clash happened did use `spawn_supervised` and did not use `async: true` but still got port clash. If that's not because tests running in parallel, I'm not sure how that's possible.

In this test the listener is not supervised: 
https://github.com/build-trust/ockam/blob/060ab64adeac8109c9f4d0911ebff90865c7b2fc/implementations/elixir/ockam/ockam_kafka/test/interceptor/interceptor_test.exs#L127-L131
And I'm pretty sure it not terminating before we called to set inlets in this test https://github.com/build-trust/ockam/blob/060ab64adeac8109c9f4d0911ebff90865c7b2fc/implementations/elixir/ockam/ockam_kafka/test/interceptor/inlet_manager_test.exs#L43 was the culprit as they both would use port 9000

PR #4648, opened 2023-04-10
feat(rust): combined `bootstrap_server_ip`/`port` into `bootstrap_server`

### LINKED ISSUE #4614: update `ockam service start kafka-producer/consumer` clap command to combine arguments `bootstrap-server-ip` and `bootstrap-server-port`
### Current behavior

Currently when starting a kafka producer and consumer service, the [arguments to indicate the ip and port](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/service/start.rs#L62-L98) are separate such as:

```sh
ockam service start kafka-producer --node producer1 --project-route /project/default \
  --bootstrap-server-ip 127.0.0.1 --bootstrap-server-port 5000 \
  --brokers-port-range 5001-5100
  ```
  
  ```
  ockam service start kafka-consumer --node consumer --project-route /project/default \
  --bootstrap-server-ip 127.0.0.1 --bootstrap-server-port 4000 \
  --brokers-port-range 4001-4100
  ```

### Desired behavior

Update this command to allow for one singular argument, named `bootstrap-server` that accepts an `{IP}:{PORT}` or just `{PORT}` 

In the case of just a `PORT`, the ip address will should be the loopback address, `127.0.0.1`

```sh
ockam service start kafka-producer --node producer1 --project-route /project/default \
  --bootstrap-server 192.168.0.51:5000 \
  --brokers-port-range 5001-5100
```

```sh
ockam service start kafka-producer --node producer1 --project-route /project/default \
  --bootstrap-server 5000 \
  --brokers-port-range 5001-5100
```

```sh
ockam service start kafka-consumer --node consumer --project-route /project/default \
  --bootstrap-server 192.168.0.50:4000 \
  --brokers-port-range 4001-4100
```

```sh
ockam service start kafka-consumer --node consumer --project-route /project/default \
  --bootstrap-server 4000 \
  --brokers-port-range 4001-4100
```

Dev Notes
---

The commands implementation to use the [arguments](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/service/start.rs#L62-L98) are found [here](https://github.com/build-trust/ockam/blob/af78ca3f6a971aa24ea8e5ff544bc17388f0dfab/implementations/rust/ockam/ockam_command/src/service/start.rs#L182-L233).

---

nidnogg:
## Current behavior

Currently, for the `kafka-producer/consumer` commands, the `bootstrap_server_ip` and `bootstrap_server_port` arguments are separate.

## Proposed changes
This PR combines both of those arguments into a single  `bootstrap_server` argument, to solve https://github.com/build-trust/ockam/issues/4614

## Checks


adrianbenavides:
> - The [parsing routine](https://github.com/build-trust/ockam/pull/4648/files#diff-4264bbada18a8ed0c1e0f34f460b015c66eec89a9499bf758d703188f25708ecR190-R228) could perhaps be abstracted into a BootstrapServer struct defined under /util.rs. It'd likely only be used within the ockam_command/src/service/start.rs routines, however. Would this be an OK abstraction?
> - I've had troubles testing it as to not having Kafka installed atm. But if there are any glaring issues before I go through the docs examples, I'd be glad to work on having them fixed ASAP.

Since it's just a processing function that converts A -> B,C, I'd just extract that logic into a private function on the command's file and write a unit test for that. You only have to ensure that the argument is parsed correctly.

implementations/rust/ockam/ockam_command/src/service/start.rs
```diff
@@ -64,13 +64,11 @@ pub enum StartSubCommand {
...
-        /// The address where to bind and where the client will connect to
-        #[arg(long, default_value_t = [127, 0, 0, 1].into())]
-        bootstrap_server_ip: Ipv4Addr,
-        /// The port to bind to and where the client will connect to
+        /// The address where to bind and where the client will connect to alongside its port, <address>:<port>.
+        /// In case just a port is specified, the default loopback address will be used
         #[arg(long)]
-        bootstrap_server_port: u16,
-        /// Local port range dynamically allocated to kafka brokers, mut not overlap with the
+        bootstrap_server: String,
```
caoakleyii: We should still be able to keep this value an Ipv4Addr

in order to handle this and wrap the logic in a reusable way clap provides an arg attribute `value_parser = parse_ip_address`

The fn signature would look like this
`parse_ip_address(input: &str) -> Result<Ipv4Addr>`

nidnogg: I thought the `Ipv4Addr` type supported only the IP address itself and not the port as it's just an array of octets. Wouldn't this end up requiring the port separately anyway?

nidnogg: > Since it's just a processing function that converts A -> B,C, I'd just extract that logic into a private function on the command's file and write a unit test for that. You only have to ensure that the argument is parsed correctly.

I've gone ahead with this approach because of the situation I mentioned in the last comment.
I've also undrafted this PR since I've managed to test the parameter parsing with a local kafka installation and it seems to be working OK as of today. 

Let me know if you need any further touch-ups ✌️!

adrianbenavides: @nidnogg I think @CAOakleyII was referring to `SocketAddr`, which represents an `Ipv4Addr` + a port. Using that you can avoid returning the ip/port split in two variables at `parse_bootstrap_server`.

caoakleyii: Thank you for that catch  @adrianbenavides  :)

I was indeed referring to `SocketAddr`

nidnogg: Makes sense to me - I've modified it to use a `SocketAddr` as well as in the unit test, although the `start_kafka_service_impl` routine in `ockam_api/src/nodes/service/services.rs` [does use the port and IP in separate ways](https://github.com/build-trust/ockam/pull/4648/files#diff-9ec90b1c70033c4cc7a9648c1dfba47cdd60a45dce9b698e02de4c9ee1f32e3dR723-R724), so I kept it split there deeper in the flow. 

Let me know if you want to just use one parameter there as well!

adrianbenavides: > Let me know if you want to just use one parameter there as well!

I think we can let that as it is 👍

adrianbenavides:
@nidnogg I've added a commit to simplify a little the `parse_bootstrap_server` function. Let me know what you think ([here is the diff](https://github.com/build-trust/ockam/compare/850239600729676d520d5eff14172edf08bcfb91..5991822c2e4b5289566da37cd6549b3e43413928)):
- We can parse the IP string directly without iterating the octets using `.parse()`
- We can implicitly parse a Ipv4Addr into a IpAddr::V4 using `.into()`
- We can eliminate the indirection produced by the `if/else` blocks and give more context about what those `len == 2 / len < 2` mean using a `match`

 We can proceed with the final review after that.

nidnogg:
> @nidnogg I've added a commit to simplify a little the `parse_bootstrap_server` function. Let me know what you think ([here is the diff](https://github.com/build-trust/ockam/compare/850239600729676d520d5eff14172edf08bcfb91..5991822c2e4b5289566da37cd6549b3e43413928)):
> 
>     * We can parse the IP string directly without iterating the octets using `.parse()`
> 
>     * We can implicitly parse a Ipv4Addr into a IpAddr::V4 using `.into()`
> 
>     * We can eliminate the indirection produced by the `if/else` blocks and give more context about what those `len == 2 / len < 2` mean using a `match`
> 
> 
> We can proceed with the final review after that.

Looks good to me! I often forget about `.into()` like in these cases. I also agree that the `match` is clearer in this regard.

adrianbenavides (approved):
Great, ready to merge then! Thanks @nidnogg for another great contribution 🙏

etorreborre (approved):

PR #4652, opened 2023-04-11
feat(rust): add new output formats to create/default/delete vault commands

adrianbenavides:
Main changes:
- modify the output of the create/default/delete vault commands
- add interactivity on tty terminals
- add macros to easily format tty messages

## plain/tty output examples

First vault:

```bash
➜ ockam vault create     
> This is the first vault to be created in this environment. It will be set as the default vault
✔ Vault created with name '2d003265'!
! You can fetch its details running `ockam vault show 2d003265` or check all the available vaults running `ockam vault list`
```

Another vault:

```bash
➜ ockam vault create myvault
✔ Vault created with name 'myvault'!
! You can fetch its details running `ockam vault show myvault` or check all the available vaults running `ockam vault list`
```

## machine/json output

If we redirect the stderr through a pipe:

```bash
➜ ockam vault create thevault 2>&1 | cat
thevault
```

And, with the `--output` flag, we can check the json format:

```bash
➜ ockam vault create jsonvault --output json
{"vault":{"name":"jsonvault"}}
```


mrinalwadhwa:
> ask for confirmation if all the arguments are default

This feels unnecessary. We should pause for confirmation only if something destructive and irreversible is happening.

implementations/rust/ockam/ockam_command/src/vault/create.rs
```diff
@@ -38,8 +44,35 @@ async fn run_impl(
...
+        .name
+        .clone()
+        .unwrap_or_else(|| hex::encode(random::<[u8; 4]>()));
+    if cmd.has_default_args() {
+        if let ConfirmResult::No = opts.terminal.confirm(&fmt_info!(
+            "This will create a vault with name '{name}'. Do you want to continue?"
+        ))? {
+            return Ok(());
+        }
+    }
```
adrianbenavides: I'm not 100% convinced about relying on whether the command is being used with all its defaults or not to show the confirmation prompt. Especially because it forces us to remove the clap's attributes. 

How do you think we should handle this? Show it always no matter the arguments being used?

adrianbenavides: I just saw your comment above @mrinalwadhwa (https://github.com/build-trust/ockam/pull/4652#issuecomment-1504510714). I'll leave this open to bring the reviewers' attention to it in the mean time.

caoakleyii:
```sh
➜ ockam vault create     
> This is the first vault to be created in this environment. It will be set as the default vault
✔ Vault created with name '2d003265'!
! You can fetch its details running `ockam vault show 2d003265` or check all the available vaults running `ockam vault list`
```

My 2c -

>  `> This is the first vault to be created in this environment. It will be set as the default vault`

I like this. Having some of the actions that are done for the user present and visible seems to be a good idea, to allow the user to understand what's going on and give feedback.

> `✔ Vault created with name '2d003265'!`

Also, like this, wonder if we want some coloring or highlighting around the `'2d003265'` to allow it to stand out.


>  `! You can fetch its details running `ockam vault show 2d003265` or check all the available vaults running `ockam vault list``

Seeing this at the end of every command I think can be tiring, and also detract from the user being able to readily get to the positive response from an action.

I do like the idea of having `tips` for the user, just how often and where it shows within the output should be determined, or possibly a way for the user to turn it off seems needed.

adrianbenavides:
> I do like the idea of having tips for the user, just how often and where it shows within the output should be determined, or possibly a way for the user to turn it off seems needed.

I have some ideas about this. ~Should we remove that part for now until we figure it out properly?~ I've removed that in case that's enough to get the PR merged.

implementations/rust/ockam/ockam_command/src/terminal/fmt.rs
```diff
@@ -0,0 +1,39 @@
...
+
+#[macro_export]
+macro_rules! fmt_err {
+    ($input:expr) => {
+        format!("{} {}", "×".light_red(), format!($input))
+    };
+    ($input:expr, $($args:expr),+) => {
+        format!("{} {}", "×".light_red(), format!($input, $($args),+))
+    };
+}
```
caoakleyii, line 39: I really like this :+1:

implementations/rust/ockam/ockam_command/src/terminal/fmt.rs
```diff
@@ -0,0 +1,39 @@
...
+    };
+    ($input:expr, $($args:expr),+) => {
+        format!("{} {}", "✔".light_green(), format!($input, $($args),+))
+    };
+}
+
+#[macro_export]
+macro_rules! fmt_info {
+    ($input:expr) => {
+        format!("{} {}", ">".light_green(), format!($input))
```
caoakleyii, line 14: We can probably work more on the colors we choose in the future.

I do think light green should be for `fmt_ok` and there should be a completely different color for `fmt_info`

adrianbenavides: Yeah agree, I just took it from the output we currently have in the `ockam enroll` command. The good thing is that we'll only have to change it here.

implementations/rust/ockam/ockam_command/src/terminal/fmt.rs
adrianbenavides: And I found it pretty comfortable to use. With these macros + the use of `indoc` we can create all sorts of output structures (single/multi line) and with lint support (just like if we were using `format!()`).

glenngillen:
I agree with Chris, and love the touch of explaining that it's setup the default project. It's a good example of just doing the right thing to provide a good experience, but explaining how/what is done so people understand it. 

LGTM!

implementations/rust/ockam/ockam_command/src/terminal/fmt.rs
shanesveller: The most recent Clap release has some interesting theming stuff included behind an unstable flag - maybe we can offer them feedback or adopt it in the future. Docs are quite sparse as of today.

https://github.com/clap-rs/clap/issues/3234
Dependabot PR for us: https://github.com/build-trust/ockam/pull/4748

adrianbenavides: That's cool! We could leverage that feature to integrate the clap native errors/output (https://docs.rs/clap_builder/4.2.3/src/clap_builder/builder/styled_str.rs.html#233-242).

mrinalwadhwa (dismissed):

mrinalwadhwa:
@adrianbenavides this one needs a rebase after #4703 merged.

adrianbenavides:
> @adrianbenavides this one needs a rebase after #4703 merged.

Ready now, thanks for the heads-up!

metaclips (approved):

PR #4656, opened 2023-04-11
feat(rust): create a background default node and identity on demand

caoakleyii:
## Current behavior

When a user runs a command without supplying a node name, and there is no default node, those commands will fail.

When a user runs a command without supplying an identity name, and there is no default node, those commands will assume no identity.

## Proposed changes

When a user runs a command without supplying a node name, and there is no default node, a default node is created and used.

When a user runs a command without supplying an identity name, and there is no default identity, a default identity is created and used.

```sh
ockam enroll
ockam forwarder create
```

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/mod.rs
@@ -106,8 +106,10 @@ pub fn node_name_parser(node_name: &str) -> Result<String> {
 
 pub fn spawn_default_node(node_name: &str) -> String {
     let config = OckamConfig::load().expect("Failed to load config");
-    let mut args = GlobalArgs::default();
-    args.quiet = true;
+    let args = GlobalArgs {
+        quiet: true,
+        ..Default::default()
+    };
     let global_opts = CommandGlobalOpts::new(args, config);
     let mut create_command = CreateCommand::default();
     create_command.node_name = node_name.to_string();
--- a/implementations/rust/ockam/ockam_command/src/node/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/mod.rs
@@ -106,7 +106,9 @@ pub fn node_name_parser(node_name: &str) -> Result<String> {
 
 pub fn spawn_default_node(node_name: &str) -> String {
     let config = OckamConfig::load().expect("Failed to load config");
-    let global_opts = CommandGlobalOpts::new(GlobalArgs::default(), config);
+    let mut args = GlobalArgs::default();
+    args.quiet = true;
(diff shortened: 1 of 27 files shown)


implementations/rust/ockam/ockam_command/src/node/mod.rs
```diff
@@ -95,3 +95,21 @@ pub fn default_node_name() -> String {
...
+        return Ok(spawn_default_node(node_name));
+    }
+
+    Ok(node_name.to_string())
+}
+
+pub fn spawn_default_node(node_name: &str) -> String {
+    let config = OckamConfig::load().expect("Failed to load config");
+    let global_opts = CommandGlobalOpts::new(GlobalArgs::default(), config);
+    let mut create_command = CreateCommand::default();
```
adrianbenavides, line 114: So this actually works 🤯

adrianbenavides: I don't know why I thought this didn't work with clap, but I'm glad it does. This will help us a lot in our quest to simplify how we compose commands.

implementations/rust/ockam/ockam_command/tests/bats/unit.bats
```diff
@@ -163,16 +163,14 @@ teardown() {
   # Create with random name
   run "$OCKAM" identity create
   assert_success
-  assert_output --partial "Identity created"
 
   # Create a named identity and delete it
   i=$(random_str)
   run "$OCKAM" identity create "${i}"
   assert_success
-  assert_output --partial "Identity created"
```
caoakleyii, line 172: These were removed because of the way `run` works within bats. 

The command and output interprets it as a tty, and then prints the machine output.

adrianbenavides (commented):
Looks good to me, we can merge it after the conflict is solved 👍

implementations/rust/ockam/ockam_command/tests/bats/unit.bats
adrianbenavides: Something that we should discuss is that we might need to change the output to show the identity name instead of the identifier. That would make things easier. Right now we can't do anything after creating an unnamed identity. If we do this change we could:

```bash
name = $(ockam identity create)
identifier = $(ockam identity show $name)
```

caoakleyii: I agree, I also think we want to figure out how we can still use bats to test plain output, in addition to machine and json.

Especially as we move towards improving our outputs.

adrianbenavides (approved):

PR #4660, opened 2023-04-11
Permit noreply outcomes from API Endpoint callback modules

### LINKED ISSUE #4645: Elixir: expand `Ockam.Services.API.Endpoint` to handle multiple API requests concurrently
### Current behavior

Currently `Ockam.Services.API.Endpoint.dispatch/5` always waits for handler to complete before returning `{:reply ...`.

This means that only one request can run on the same API worker at a time. While functions defined in handler may take some time.

### Desired behavior

If handler functions don't have to block each other (as it usually is in API workers), they can instead run concurrently.

Base API worker `Ockam.Services.API.handle_request/2` supports returning `{:noreply, ...}` and replying later. This can be used in `dispatch/5` or `handle_request/2` callback implementations in `Ockam.Services.API.Endpoint`

shanesveller:
## Current behavior

Elixir modules that `use Ockam.Services.API.Endpoint` currently must return an ok-tuple or error-tuple from their request handler functions in order to satisfy the `case` statement of `dispatch/5` [here](https://github.com/build-trust/ockam/blob/47a03863942682b24a1d47e8f4e7862c21a7ffde/implementations/elixir/ockam/ockam_services/lib/services/api/endpoint.ex#L120).

## Proposed changes

In order to allow such downstream modules to perform asynchronous replies out-of-band, it must be allowed to return `:noreply` or `{:noreply, new_state}`. Implementation details to conduct those replies in practice are not included here, though they would likely be built atop `Ockam.Services.API.reply` and `reply_error`.

## Checks

### DIFF
--- a/implementations/elixir/ockam/ockam_services/lib/services/api/endpoint.ex
+++ b/implementations/elixir/ockam/ockam_services/lib/services/api/endpoint.ex
@@ -125,6 +125,12 @@ defmodule Ockam.Services.API.Endpoint do
           {:ok, body, new_endpoint_state} ->
             {:reply, :ok, body, %{state | endpoint_state: new_endpoint_state}}
 
+          :noreply ->
+            {:noreply, state}
+
+          {:noreply, new_endpoint_state} ->
+            {:noreply, %{state | endpoint_state: new_endpoint_state}}
+
           {:error, reason} ->
             {:error, reason}
         end


implementations/elixir/ockam/ockam_services/lib/services/api/endpoint.ex
```diff
@@ -125,6 +125,12 @@ defmodule Ockam.Services.API.Endpoint do
           {:ok, body, new_endpoint_state} ->
```
hairyhum, line 125: Should we change those to `{:reply, :ok, body}` and `{:reply, :ok, body, state}` for consistency?
@polvorin any particular reason why it was changed to `{:ok, body}`? That prevents handlers from returning other status codes btw.

hairyhum (commented):
I think we can make it more consistent and more flexible by allowing different status codes by changing the `{:ok, body}` result as well

implementations/elixir/ockam/ockam_services/lib/services/api/endpoint.ex
polvorin: right, as is on the success case the handlers returns the  response body  so no way to indicate a success code other than 200.    So if we want to do things like 201 "Created"  we will need to change this.
Other quirk is that the handler accept a Request{},   but don't return a Response{}  (returns a response body,   or on a upper level,   a status code and response body).
It feels that  a Request{}  -> Response{}  here would make more sense.
 
Regarding the async response.. in experience so far it looks like most things that are implemented behind an endpoint like this *don't*  mutate any internal state in any way and are fine to be run in parallel.   Would it make sense to keep the handlers unaware of all this,  and just be a configuration on the routes?   (run the handler 
 in an `one-at-a-time way` like now,   run the handler in a `freshly spawned fun`,  run the handler in a spawned fun but up to N at a given time at max,  etc).   So the handler remains just as a function from req -> resp.     (of course this won't apply for handlers that _do_ change worker state)

shanesveller: @hairyhum @polvorin I haven't had enough time to develop the context I would need to speak to or contribute a larger reimagining of this code's design. Would it be okay if we treated those as two separate conversations and discuss the merits or concerns of this change as-is, and make separate time to discuss your points above?

hairyhum: @polvorin @shanesveller Let's make a follow-up issue with the proposal to support "async handlers" and maybe handler pools as this grows the design scope of this issue.

hairyhum (dismissed):
@shanesveller since we want to review that in the follow-up, it's fine with me if we don't switch to `{:reply, :ok` now.

mszpakowski (dismissed):

mszpakowski:
bats tests are currently failing, and that's the reason I'm waiting with merging

mszpakowski (approved):

PR #4666, opened 2023-04-12
fix(rust): fix okta addon attributes argument on authority command

polvorin:
clap parses Vec<_> as multiple --ARG  rather than comma separate list of things.

Also small addition on okta' authenticator to include appropriate attribute.


etorreborre (approved):
Thanks for fixing this, I should write some unit tests for the command-line parsing of arguments when I can.

PR #4670, opened 2023-04-13
`FlowControls` improvements

SanjoDeundiak:
1. Make FlowControls mandatory (no possibility to opt out)
2. Improve rust `ForwardingService` (e.g., support FlowControls)
3. Improve FlowControls API towards simplicity and mistake-resistance
4. Remove Hop Worker
5. Improve TCP listen/connect API
6. Improve Secure Channel listen/create API
7. Extend the `ockam_transport_tcp` registry
8. Remove the NodeManager Transport registry in favor of the `ockam_transport_tcp` registry
9. Add the `ockam flow-control add-consumer` command
10. Improve `ockam_command` output for relays and tcp
11. Make Hop worker available when manually started for example purposes

TODO in follow up:
1. Clean FlowControls data storage when Workers are shut down
2. Print a warning while starting a Hop worker from command line


adrianbenavides (commented):
- ✅ ockam_comand: checked, nothing to say, it looks good to me
- ✅ ockam_api: checked, left some minor comments

I love the simplifications derived from having the flow control state behind the `Context`, it looks great!

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -399,22 +398,41 @@ impl NodeManager {
...
         // Start services
+        ctx.flow_controls().add_consumer(
+            DefaultAddress::IDENTITY_SERVICE,
+            api_flow_control_id,
+            FlowControlPolicy::SpawnerAllowMultipleMessages,
+        );
         self.start_identity_service_impl(ctx, DefaultAddress::IDENTITY_SERVICE.into())
             .await?;
         self.start_authenticated_service_impl(ctx, DefaultAddress::AUTHENTICATED_SERVICE.into())
             .await?;
```
adrianbenavides, line 407: Not adding these two services as consumers means that they are not accessible from the outside?

implementations/rust/ockam/ockam_api/src/nodes/service/node_services.rs
```diff
@@ -184,6 +185,12 @@ impl NodeManager {
             return Err(ApiError::generic("Hop service exists at this address"));
         }
 
+        ctx.flow_controls().add_consumer(
+            addr.clone(),
+            &self.api_transport.flow_control_id,
+            FlowControlPolicy::SpawnerAllowMultipleMessages,
+        );
```
adrianbenavides, line 192: What happens if this is called twice? Should it fail or ignore that it already exists?

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
```diff
@@ -1,185 +1,351 @@
...
-            }
-            (Tcp, Connect) => {
-                let options = if let Some(flow_control_id) = flow_control_id {
-                    TcpConnectionOptions::as_producer(&node_manager.flow_controls, &flow_control_id)
-                } else {
-                    TcpConnectionOptions::insecure()
-                };
+        // Add all Hop workers as consumers for Demo purposes
+        // Production nodes should not run any Hop workers
+        for hop in node_manager.registry.hop_services.keys() {
```
adrianbenavides, line 198: Probably out of the scope of this PR, but maybe we should move all the `hop` service related code behind a feature flag to avoid leaking this funcionality into production.

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
```diff
@@ -1,185 +1,351 @@
...
+                    None => return Ok(Response::bad_request(req.id())),
+                    Some(addr) => addr,
+                }
             }
-            Some(t) => {
-                node_manager
+            Err(_err) => body.address.into(),
+        };
+
+        match node_manager.tcp_transport.disconnect(&sender_address).await {
```
adrianbenavides, line 310: Is it OK not to remove the tcp transport from the registry?

SanjoDeundiak: It will be removed automatically when Workers are shutdown

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
SanjoDeundiak: That's discussed

implementations/rust/ockam/ockam_api/src/nodes/service/node_services.rs
SanjoDeundiak: It's idempotent

implementations/rust/ockam/ockam_api/src/nodes/service.rs
SanjoDeundiak: I think IDENTITY_SERVICE is added above. Not sure about AUTHENTICATED_SERVICE, is it used? Should it be exposed via api transport?

SanjoDeundiak: At least none of the tests/bats tests expect that

examples/rust/get_started/src/forwarder.rs
```diff
@@ -21,9 +22,27 @@ impl Worker for Forwarder {
             .pop_front() // Remove my address from the onward_route
             .prepend(self.0.clone()); // Prepend predefined address to the onward_route
 
+        let prev_hop = transport_message.return_route.next()?.clone();
+
         // Wipe all local info (e.g. transport types)
         let message = LocalMessage::new(transport_message, vec![]);
 
+        if let Some(info) = ctx.flow_controls().find_flow_control_with_producer_address(&self.0) {
```
etorreborre, line 30: Would it be a good idea to make just one function declaring that `self.0` and `prev_hop` and producer/consumer of each other?

examples/rust/tcp_inlet_and_outlet/examples/03-outlet.rs
```diff
@@ -13,7 +14,15 @@ async fn main(ctx: Context) -> Result<()> {
     //   2. A Secure Channel Listener at Worker address - secure_channel_listener
     //      that will wait for requests to start an Authenticated Key Exchange.
     let e = node.create_identity().await?;
-    node.create_secure_channel_listener(&e, "secure_channel_listener", SecureChannelListenerOptions::new())
+
+    let tcp_listener_options = TcpListenerOptions::new();
+
+    let secure_channel_listener_options = SecureChannelListenerOptions::new().as_consumer(
```
etorreborre, line 20: I am thinking that we might be able to tighten this consumer/producer DSL even a bit more by introducing some traits. For example if `tcp_listener_options` implements a `Spawner` trait. This might not be trivial since there could a bit of dependent typing in there with the relationship with the type of flow control policy. But that could be an interesting follow-up to try to completely hide the notion of a flow id

implementations/rust/ockam/ockam/src/forwarding_service/forwarder.rs
```diff
@@ -0,0 +1,121 @@
...
+        incoming_access_control: Arc<dyn IncomingAccessControl>,
+    ) -> Result<()> {
+        info!("Created new alias {} for {}", address, forward_route);
+
+        // Should be able to reach last and second last hops
+        let outgoing_access_control: Arc<dyn OutgoingAccessControl> = if forward_route.len() == 1 {
+            // We are accessed with our node, no transport is involved
+            Arc::new(AllowAll)
+        } else {
+            let next_hop = forward_route.next()?.clone();
```
etorreborre, line 35: We could return a more explicit error here if the forward_route is empty (maybe write the whole expression as a pattern match on the first 1 or 2 elements of addresses of the route?)

implementations/rust/ockam/ockam/src/forwarding_service/forwarder.rs
```diff
@@ -0,0 +1,121 @@
...
+};
+use ockam_node::WorkerBuilder;
+use tracing::info;
+
+pub(super) struct Forwarder {
+    forward_route: Route,
+    // this option will be `None` after this worker is initialized, because
+    // while initializing, the worker will send the payload contained in this
+    // field to the `forward_route`, to indicate a successful connection
+    payload: Option<Vec<u8>>,
```
etorreborre, line 17: I'm not necessarily a big fan of this suggestion but I wonder if we could just send the payload using a function directly on `Forwarder` right after it has been started as a worker. Then we don't have to store anything in that field.

implementations/rust/ockam/ockam/src/forwarding_service/options.rs
```diff
@@ -0,0 +1,129 @@
+use ockam_core::compat::sync::Arc;
+use ockam_core::compat::vec::Vec;
+use ockam_core::flow_control::{FlowControlId, FlowControlPolicy, FlowControls};
+use ockam_core::{Address, AllowAll, IncomingAccessControl};
+
+/// Trust Options for a Forwarding Service
```
etorreborre, line 6: A bit more documentation about the role of each field would be nice here

etorreborre (approved):

PR #4703, opened 2023-04-14
refactor(rust): create abstraction for the cli state directories and applies it to the vaults state

adrianbenavides:
The `CliState` has similar logic for all the existing kinds of states (vaults, identities, nodes, projects, trust_contexts). The only differences among those states are how are they created/deleted. The rest can be abstracted away.

This PR introduces a set of traits that will abstract a good amount of code with only small changes in the client code. There is still room to abstract more common functions/boilerplate, but that can be done later without further altering the client code, probably through macros.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
@@ -1,0 +1,123 @@
+use crate::cli_state::{file_stem, CliState, CliStateError};
+use ockam_core::async_trait;
+use serde::{Deserialize, Serialize};
+use std::path::{Path, PathBuf};
+
+use super::Result;
+
+/// Represents the directory of a type of state and contains
+/// all the data related to that type.
+#[async_trait]
+pub trait StateTrait: Sized {
+    type ItemDir: StateItemDirTrait;
+    type ItemConfig: StateItemConfigTrait + Serialize + for<'a> Deserialize<'a>;
+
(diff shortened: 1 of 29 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
```diff
@@ -0,0 +1,123 @@
...
+    fn dir(&self) -> &PathBuf;
+    fn dir_as_string(&self) -> String {
+        self.dir().to_string_lossy().to_string()
+    }
+
+    fn path(&self, name: &str) -> PathBuf {
+        self.dir().join(format!("{name}.json"))
+    }
+
+    async fn create(&self, name: &str, config: Self::ItemConfig) -> Result<Self::ItemDir>;
```
caoakleyii, line 40: I have some small concerns on the need for async here and in `delete` fns. Especially when we very rarely use our async calls for any concurrent performance gains.

It has impact on our usage the state within clap parsers, and other implementations.

adrianbenavides: There are some cases where we need the async. Maybe we could have `create` and `create_async` to use async just when necessary?

caoakleyii: perhaps, I didn't realizes there were some cases where we needed the async here.

We can keep it as is then, and then update to add a sync `fn` when we need it.

mrinalwadhwa (approved):

PR #4708, opened 2023-04-14
feat(rust): disable `FlowControl` for loopback tcp connections and listeners

SanjoDeundiak:


implementations/rust/ockam/ockam_api/src/util.rs
```diff
@@ -54,56 +54,75 @@ pub async fn multiaddr_to_route(
...
-        flow_controls,
-        &flow_control_id,
-    ));
+    let mut flow_control_id = None;
+    let mut number_of_tcp_hops = 0;
 
     while let Some(p) = it.next() {
         match p.code() {
             Ip4::CODE => {
+                if number_of_tcp_hops >= 1 {
```
etorreborre, line 64: This condition could be lifted out of the `match`

etorreborre (approved):
👍

implementations/rust/ockam/ockam_api/src/util.rs
SanjoDeundiak: Not sure

davide-baldo: We want the condition only for dns,ipv4 and ipv6

etorreborre: Sorry I skipped the other cases with code folding. I wish we had some common abstraction for those 3 cases but maybe there's nothing that really fits.

PR #4727, opened 2023-04-17
feat(rust): remove the vault service endpoint for getting secret data

etorreborre:
A node currently exposes an operation to export secret data. This creates a risk of exposing sensitive data to the outside world. That risk is currently being mitigated by controlling the flow of messages between workers in a node but the best defense is simply to stop exposing this endpoint.


etorreborre:
@hairyhum are we in any way using this endpoint from the Orchestrator controller?

SanjoDeundiak (approved):

PR #4728, opened 2023-04-17
fix(rust): fixed 2 comments giving the name of kafka forwarders

etorreborre:
It should be `{topic_name}_{partition}`


mrinalwadhwa:
The commit message here would be better as `docs:` .. `fix` and `feat` are added to the generated changelogs

etorreborre:
@mrinalwadhwa indeed, I changed it.

davide-baldo (dismissed):

davide-baldo (approved):

PR #4747, opened 2023-04-19
fix(rust): move the multiaddr resolution for a credential retriever

etorreborre:
**UPDATE** This PR is now closed in favor of https://github.com/build-trust/ockam/pull/4752

This is a proposal of a fix for the broken bats portals tests.

# Diagnostic

 - the previous version had a `CredentialRetriever` possibly configured with a `Route` to the authority
 - the code was moved from the `ockam_api` crate to the `ockam_identity` along with other `TrustContext` and `Credential` concerns (this allowed to remove some redundant code around `CredentialIssuer`)
 - in order to do the move the multiaddr in the original configuration of a `CredentialRetriever` was resolved to a `Route` and that route used as the configuration for the `CredentialRetriever`

My understanding of the current issue is that: 

 - unfortunately that route is not "stable". It contains the addresses of some TCP workers which might be dropped, and then the authority can not be contacted anymore in a portals via Orchestrator case

# Fix

The fix applied here consists in keeping the `MultiAddr` configuration on the `CredentialRetriever` and do the resolution inside the `ockam_identity` crate. This now passes all the tests.

Unfortunately this changes the dependencies of the crates and now we have a dependency between `ockam_identity` and `ockam_multiaddr` and this may not be what we want.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3442,6 +3442,7 @@ dependencies = [
  "ockam_core",
  "ockam_key_exchange_xx",
  "ockam_macros",
+ "ockam_multiaddr",
  "ockam_node",
  "ockam_transport_tcp",
  "ockam_vault",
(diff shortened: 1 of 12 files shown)


davide-baldo:
I think it would be a cleaner solution to just make `CredentialRetriever` a dyn Trait and implement it inside `ockam_api` rather than leaking dependencies within `ockam_identity`

PR #4752, opened 2023-04-21
fix(rust): resolve transport addresses as a separate step

etorreborre:
This fixes a connection issue with the `CredentialRetriever` where the route that is configured to retrieve credentials must be resolved to actual TCP workers when the route is accessed.

The `CredentialRetriever` is now configured with a route which has been resolved from the original multiaddr configuration.

However, instead of creating TCP workers during the multiaddr conversion, IP addresses are replaced with `TCP` transport addresses. Then, on credential retrieval, the TCP addresses are used to create a TCP connection and TCP addresses are replaced with the local workers addresses (what was done during multiaddr resolution before).

I went the extra-step of making this kind of resolution generic across transports so that resolving `UDP` addresses into local UDP workers should be easy to do in the future (whereas the current multiaddr resolution does not support UDP).

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/util.rs
+++ b/implementations/rust/ockam/ockam_api/src/util.rs
@@ -2,7 +2,7 @@ use crate::error::ApiError;
 use anyhow::anyhow;
 use ockam::TcpTransport;
 use ockam_core::flow_control::{FlowControlId, FlowControls};
-use ockam_core::{Address, Error, Result, Route, LOCAL};
+use ockam_core::{Address, Error, Result, Route, TransportType, LOCAL};
 use ockam_multiaddr::proto::{DnsAddr, Ip4, Ip6, Node, Project, Secure, Service, Tcp, Worker};
 use ockam_multiaddr::{MultiAddr, Protocol};
 use ockam_transport_tcp::TcpConnectionOptions;
@@ -152,6 +152,61 @@ pub async fn multiaddr_to_route(
         flow_control_id,
         route: rb.into(),
     })
+}
+
+/// Resolve all the multiaddresses which represent transport addresses
+/// For example /tcp/127.0.0.1/port/4000 is transformed to the Address (TCP, "127.0.0.1:4000")
+/// The creation of a TCP worker and the substitution of that transport address to a worker address
+/// is done later with `context.resolve_transport_route(route)`
+pub fn multiaddr_to_transport_route(ma: &MultiAddr) -> Option<Route> {
+    let mut route = Route::new();
+    let mut it = ma.iter().peekable();
+
+    while let Some(p) = it.next() {
+        match p.code() {
+            Ip4::CODE => {
+                let ip4 = p.cast::<Ip4>()?;
+                let port = it.next()?.cast::<Tcp>()?;
+                let socket_addr = SocketAddrV4::new(*ip4, *port);
+                route = route.append(Address::new(TransportType::new(1), socket_addr.to_string()))
+            }
+            Ip6::CODE => {
+                let ip6 = p.cast::<Ip6>()?;
+                let port = it.next()?.cast::<Tcp>()?;
+                let socket_addr = SocketAddrV6::new(*ip6, *port, 0, 0);
+                route = route.append(Address::new(TransportType::new(1), socket_addr.to_string()))
+            }
+            DnsAddr::CODE => {
+                let host = p.cast::<DnsAddr>()?;
+                if let Some(p) = it.peek() {
+                    if p.code() == Tcp::CODE {
+                        let port = p.cast::<Tcp>()?;
+                        let addr = format!("{}:{}", &*host, *port);
+                        route = route.append(Address::new(TransportType::new(1), addr));
+                        let _ = it.next();
+                        continue;
+                    }
+                }
+            }
+            Worker::CODE => {
+                let local = p.cast::<Worker>()?;
+                route = route.append(Address::new(LOCAL, &*local))
+            }
+            Service::CODE => {
+                let local = p.cast::<Service>()?;
+                route = route.append(Address::new(LOCAL, &*local))
+            }
+            Secure::CODE => {
+                let local = p.cast::<Secure>()?;
+                route = route.append(Address::new(LOCAL, &*local))
+            }
+            other => {
+                error!(target: "ockam_api", code = %other, "unsupported protocol");
+                return None;
+            }
+        }
+    }
+    Some(route.into())
 }
 
 /// Try to convert a multiaddr to an Ockam Address
--- a/implementations/rust/ockam/ockam_identity/src/credentials/credentials_retriever.rs
+++ b/implementations/rust/ockam/ockam_identity/src/credentials/credentials_retriever.rs
@@ -5,11 +5,12 @@ use crate::{
 use core::time::Duration;
 use ockam_core::compat::boxed::Box;
 use ockam_core::compat::sync::Arc;
+use ockam_core::errcode::{Kind, Origin};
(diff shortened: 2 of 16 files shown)


adrianbenavides (commented):
Awesome, delaying multiaddr resolution is definitely a big win! Do you want to wait for @SanjoDeundiak to check this one out? Otherwise we could merge it ASAP

etorreborre:
@adrianbenavides given that this has been blocking develop for a few days now, I think we should merge. @SanjoDeundiak I will follow-up with any comments/suggestions in a subsequent PR if necessary.

adrianbenavides (approved):

implementations/rust/ockam/ockam_transport_tcp/src/transport/lifecycle.rs
```diff
@@ -31,3 +43,138 @@ impl TcpTransport {
...
+            )
+            .await?;
+
+        ctx.stop().await
+    }
+
+    #[ockam_macros::test]
+    async fn test_resolve_route_only_single_hop_is_allowed(ctx: &mut Context) -> Result<()> {
+        let tcp = TcpTransport::create(ctx).await?;
+        let tcp_address = "127.0.0.1:5002";
```
SanjoDeundiak, line 144: let tcp_address = "127.0.0.1:0";

implementations/rust/ockam/ockam_transport_tcp/src/transport/lifecycle.rs
```diff
@@ -31,3 +43,138 @@ impl TcpTransport {
...
+mod tests {
+    use super::*;
+    use ockam_core::{route, TransportType, LOCAL};
+    use ockam_transport_core::TransportError;
+    use std::net::TcpListener;
+
+    #[ockam_macros::test]
+    async fn test_resolve_route(ctx: &mut Context) -> Result<()> {
+        let tcp = TcpTransport::create(ctx).await?;
+        let tcp_address = "127.0.0.1:5001";
```
SanjoDeundiak, line 102: let tcp_address = "127.0.0.1:0";

implementations/rust/ockam/ockam_node/src/node.rs
```diff
@@ -58,6 +60,7 @@ impl NodeBuilder {
                 vec![],
             ),
             None,
+            Arc::new(Mutex::new(HashMap::new())),
```
SanjoDeundiak, line 63: `Default::default()`

implementations/rust/ockam/ockam_identity/src/credentials/credentials_retriever.rs
```diff
@@ -63,7 +64,22 @@ impl RemoteCredentialsRetriever {
 #[async_trait]
 impl CredentialsRetriever for RemoteCredentialsRetriever {
     async fn retrieve(&self, ctx: &Context, for_identity: &Identity) -> Result<Credential> {
+        if !ctx.is_transport_registered(TransportType::new(1)) {
```
SanjoDeundiak, line 67: What is the purpose of this check? Isn't it simpler to ask the `Context` to resolve the route and see if it succeeds?

implementations/rust/ockam/ockam_api/src/util.rs
```diff
@@ -154,6 +154,61 @@ pub async fn multiaddr_to_route(
...
+    let mut route = Route::new();
+    let mut it = ma.iter().peekable();
+
+    while let Some(p) = it.next() {
+        match p.code() {
+            Ip4::CODE => {
+                let ip4 = p.cast::<Ip4>()?;
+                let port = it.next()?.cast::<Tcp>()?;
+                let socket_addr = SocketAddrV4::new(*ip4, *port);
+                route = route.append(Address::new(TransportType::new(1), socket_addr.to_string()))
```
SanjoDeundiak, line 171: Could you please use TCP constant instead of `1` in this function?

implementations/rust/ockam/ockam_transport_tcp/src/transport/lifecycle.rs
```diff
@@ -31,3 +43,138 @@ impl TcpTransport {
...
+                let addr = self.connect(address.address().to_string(), options).await?;
+                result = result.append(addr)
+            } else {
+                result = result.append(address.clone());
+            }
+        }
+
+        let resolved = result.into();
+        debug!("resolved route {} to {}", route, resolved);
+        Ok(resolved)
```
SanjoDeundiak, line 88: Should we close the opened connection if this function returns an Error?

implementations/rust/ockam/ockam_transport_tcp/src/transport/lifecycle.rs
etorreborre: Why this? If I use that port I am getting "generic I/O failure"

implementations/rust/ockam/ockam_api/src/util.rs
etorreborre: Absolutely

implementations/rust/ockam/ockam_identity/src/credentials/credentials_retriever.rs
etorreborre: The way the `Context` resolves the route is by passing the route to each transport capable of resolving it. If no transport is registered then the route is unchanged and the connection will timeout. This way we get an indication of what went wrong. This is quite similar to the check that was previously done in the `ockam_api` crate before doing the multiaddr resolution for the `CredentialRetriever`

implementations/rust/ockam/ockam_transport_tcp/src/transport/lifecycle.rs
etorreborre: That's a good suggestion indeed. However when I try that the second unit test (checking the number of hops) hangs and I can't see why yet. Any idea?

etorreborre: I fixed the issue by simply checking the number of hops first and only creating a connection if the route is correct.

etorreborre:
I have addressed the review comments in: https://github.com/build-trust/ockam/pull/4779

implementations/rust/ockam/ockam_transport_tcp/src/transport/lifecycle.rs
SanjoDeundiak: You should use it in the listener so that OS assigns you a random (guaranteed free) port. That listener will return you a socket_addr it's bound to, so you can connect

implementations/rust/ockam/ockam_identity/src/credentials/credentials_retriever.rs
SanjoDeundiak: How about changing that, then? It makes sense to me if you ask the Context to resolve your route, and you get either the correct resolved route or an error returned. Why does CredentialsRetriever implementation need to know what is TCP transport type and if it was registered on the node? Why can't the Context check everything itself?

etorreborre: At first I thought it was not possible without the `Context` knowing about which transports were registered or not. But it is possible to have the `Context` check if eventually all addresses have been resolved to local addresses which is exactly our intention :-). So thanks for the suggestion that's an excellent one :-).

implementations/rust/ockam/ockam_transport_tcp/src/transport/lifecycle.rs
etorreborre: Thanks, I see what you mean now, and it works fine

PR #4756, opened 2023-04-22
Add support for secure channel rekey and improve out-of-order message handling

polvorin:
Add secure channel rekey implementation


implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
```diff
@@ -1,39 +1,137 @@
...
-    pub async fn decrypt(&self, payload: &[u8]) -> Result<Vec<u8>> {
+    pub async fn decrypt(&mut self, payload: &[u8]) -> Result<Vec<u8>> {
         if payload.len() < 8 {
             return Err(IdentityError::InvalidNonce.into());
         }
 
-        let nonce = Self::convert_nonce_from_small(&payload[..8])?;
+        let (nonce, nonce_buffer) = Self::convert_nonce_from_small(&payload[..8])?;
+
+        self.nonce_tracker.mark(nonce)?;
```
polvorin: the actual marking should be done only after decryption succeed

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
```diff
@@ -1,39 +1,137 @@
...
+
+        if nonce >= self.current_key_nonce + KEY_RENEWAL_INTERVAL {
```
SanjoDeundiak, line 42: I think the sequence should be the following:
1. Check if nonce is correct
2. Obtain the key that should be used to decrypt the message (previous, current, or derive a new one)
3. Try to decrypt the message
  - If decryption fails - do nothing with the state of the Decryptor. Destroy derived keys if there were any
  - If succeeds - replace old keys with the new keys if needed, update nonce&NonceTracker

What do you think?

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
```diff
@@ -1,39 +1,137 @@
...
-    vault: Arc<dyn SymmetricVault>,
+    current_key: KeyId,
+    current_key_nonce: u64,
+
+    previous_key: Option<KeyId>,
+    vault: Arc<dyn XXInitializedVault>,
+    nonce_tracker: NonceTracker,
+}
+
+pub(crate) struct NonceTracker {
```
SanjoDeundiak: Could you please move it either to the separate file or closer to the implementation?

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
```diff
@@ -1,39 +1,137 @@
...
+            let relative_shift: u64 = nonce - self.current_nonce;
+            if relative_shift > KEY_RENEWAL_INTERVAL {
+                return Err(IdentityError::InvalidNonce.into());
+            }
+
+            self.nonce_bitmap <<= relative_shift;
+            self.nonce_bitmap |= 1;
+            self.current_nonce = nonce;
+        } else {
+            // we received a message from the past, out of order
```
SanjoDeundiak: What if nonce == 0 and current_nonce == 0?

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
```diff
@@ -1,39 +1,137 @@
...
+        Self {
+            nonce_bitmap: 0,
+            current_nonce: 0,
+        }
+    }
+
+    /// Mark a nonce as received, reject all invalid nonce values
+    pub(crate) fn mark(&mut self, nonce: u64) -> Result<()> {
+        if nonce >= u64::MAX - KEY_RENEWAL_INTERVAL {
+            return Err(IdentityError::InvalidNonce.into());
```
SanjoDeundiak: Could you please elaborate on why it's an invalid nonce?

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
```diff
@@ -1,39 +1,137 @@
...
+    current_key: KeyId,
+    current_key_nonce: u64,
+
+    previous_key: Option<KeyId>,
+    vault: Arc<dyn XXInitializedVault>,
+    nonce_tracker: NonceTracker,
+}
+
+pub(crate) struct NonceTracker {
+    nonce_bitmap: u16,
```
SanjoDeundiak: I don't get this part, could you please elaborate? We have KEY_RENEWAL_INTERVAL = 10. Imagine Secure Channel was just created, we can accept nonces from 0 to 19 inclusively. How can we track which values were already used with just 16 bits of state?

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor.rs
```diff
@@ -23,15 +29,37 @@ impl Encryptor {
...
+        let current_nonce = self.nonce;
+        if current_nonce == u64::MAX {
             return Err(IdentityError::NonceOverflow.into());
         }
 
         self.nonce += 1;
 
-        let (small_nonce, nonce) = Self::convert_nonce_from_u64(old_nonce);
+        if current_nonce > 0 && current_nonce % KEY_RENEWAL_INTERVAL == 0 {
+            self.key = Self::rekey(&self.vault, &self.key).await?;
```
SanjoDeundiak: Please use `vault.secret_destroy()` to delete the old secret

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
davide-baldo: This makes sense, it would protect against connection disruption attack, granted we can validate the decryption.
Basically, making the decryption an atomic operation.

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
davide-baldo: > we can accept nonces from 0 to 19 inclusively

True, but we are keeping track only of the last 10, when an increased nonce is received we move the window to that nonce.

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
davide-baldo: Makes sense

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
davide-baldo: The code still works, but the comment is misleading, it'll update it

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
polvorin: btw there is a test case testing this https://github.com/build-trust/ockam/blob/65979929bf310f84c62df2fe997b908987db077f/implementations/rust/ockam/ockam_identity/src/secure_channel/mod.rs#L150-L186

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
polvorin: I'm a bit concerned on the dependency of a specific size.    I mean,   the 10 value was taked out of thin air just as a number to put there and check things works,   but at least until getting a :heavy_check_mark:  that we are going to stick to that number,   I'd have expected the code to be ok to change to  20,  50, 100, whatever by changing the constant. 

The out-of-order window doesn't affect interoperability,   just how much tolerant to out-of-order messages a receiver is.. so I think this is not a big blocker  (I mean,  if we decide to rekey each 100 messages,   and this code still works only that with a nonce validity window of 10,   i would say its ok.  And we can work on how to make this window modifiable in some future work).

SanjoDeundiak: Maybe we could use https://docs.rs/bitmaps/latest/bitmaps/ to make things cleaner?

davide-baldo: I'm not convinced introducing an extra dependency for a simple bitmap is worth the trouble, the implementation works as long `KEY_RENEWAL_INTERVAL` is smaller than the number of bits in the `nonce_bitmap` (this would be true even if we were using the library). I'm adding a check for it, it fails at compile time if that's not the case.
It should work up to 128 messages, after that, we would probably need to switch to another implementation.

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
davide-baldo: It was a leftover when the implementation was slightly different, I couldn't think of any useful scenarios. removed.

davide-baldo (approved):

PR #4779, opened 2023-04-24
CredentialRetriever fix review comments

etorreborre:
This PR addresses some comments left on #4752 .


implementations/rust/ockam/ockam_node/src/context/transports.rs
```diff
@@ -30,14 +32,26 @@ impl Context {
         for transport in transports.values() {
```
SanjoDeundiak: Do you think it's possible to simplify some logic considering the fact that there can be only 1 hop resolved here?

etorreborre: yes we could check if the returned route has been fully resolved or not. Let me try that

SanjoDeundiak: For example, there is a check in TCPs `resolve_route` to prevent more than 1 hop, but can't we instead look at the route in this function, find the first transport hop, and then ask the corresponding transport to resolve it?

SanjoDeundiak: This way, we even add support for a more complex topology and let the intermediate node to resolve its transport routes if required. Maybe in the future, we could also use that logic for secure channels (what is it's a transport type as well?)

implementations/rust/ockam/ockam_node/src/node.rs
```diff
@@ -60,7 +59,7 @@ impl NodeBuilder {
                 vec![],
             ),
             None,
-            Arc::new(Mutex::new(HashMap::new())),
+            Arc::new(Mutex::new(Default::default())),
```
SanjoDeundiak: Have you tried `Default::default()` instead of this whole line? Is the compiler able to figure it out?

implementations/rust/ockam/ockam_transport_tcp/src/transport/lifecycle.rs
```diff
@@ -51,19 +51,25 @@ impl Transport for TcpTransport {
     }
 
     async fn resolve_route(&self, flow_controls: &FlowControls, route: Route) -> Result<Route> {
-        let mut result = Route::new();
         let mut number_of_tcp_hops = 0;
-
-        for address in route.iter() {
+        for address in route.clone().iter() {
```
SanjoDeundiak: Is `.clone()` required here?

SanjoDeundiak: Also, you should be able to do something like
`route.iter().filter(...).count() > 1`

implementations/rust/ockam/ockam_node/src/context/transports.rs
SanjoDeundiak: Nevermind, we don't want more than 1 transport hop in the route.

implementations/rust/ockam/ockam_node/src/context/transports.rs
```diff
@@ -1,9 +1,11 @@
...
 use ockam_core::flow_control::FlowControls;
-use ockam_core::{Result, Route, TransportType};
+use ockam_core::{Error, Result, Route, TransportType};
 use ockam_transport_core::Transport;
 
+use crate::Context;
+
 impl Context {
     /// Return the list of supported transports
     pub fn register_transport(&self, transport: Arc<dyn Transport>) {
```
SanjoDeundiak, line 11: Could you please consider replacing `Mutex` with `RwLock`, since writing access to this field is only required a few times, while all other interactions are immutable?

implementations/rust/ockam/ockam_node/src/node.rs
etorreborre: No, that didn't work unfortunately.

etorreborre: It works with `RwLock`.

implementations/rust/ockam/ockam_node/src/context/transports.rs
etorreborre: Done

implementations/rust/ockam/ockam_node/src/node.rs
SanjoDeundiak: Works with `Mutex` as well, just in case

examples/rust/file_transfer/tests/100_32_test.bin
```diff
@@ -0,0 +1,2 @@
+�_��	��Vr����P���K�m���4�k�H�y\r\��g�c��"�_��0���}Lo����zi���ڠ;"N��Y�E��$ɥ���|
```
SanjoDeundiak: Looks like elfish

etorreborre: Grrr, those files snuck in because I was running the tests in parallel....

SanjoDeundiak (approved):
Thank you

PR #4785, opened 2023-04-25
feat(rust): rotate cli logs

adrianbenavides:
Add rotation to the node logs with the following default parameters:
- Rotate daily or when reaching 10 MB
- Keep up to 60 files

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/logs/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/logs/mod.rs
@@ -1,0 +1,95 @@
+use crate::logs::rolling::{RollingConditionBasic, RollingFileAppender};
+
+use ockam_core::env::{get_env, get_env_with_default};
+use std::io::stdout;
+use std::path::PathBuf;
+use termimad::crossterm::tty::IsTty;
+use tracing::level_filters::LevelFilter;
+use tracing_appender::non_blocking::WorkerGuard;
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::util::SubscriberInitExt;
+use tracing_subscriber::EnvFilter;
+
+#[allow(unused, clippy::enum_variant_names)]
+mod rolling;
+
+fn log_max_size() -> u64 {
+    let default = 100;
+    get_env_with_default("OCKAM_LOG_MAX_SIZE_MB", default).unwrap_or(default)
+}
+
+fn log_max_files() -> usize {
+    let default: u64 = 60;
+    get_env_with_default("OCKAM_LOG_MAX_FILES", default).unwrap_or(default) as usize
+}
+
+pub fn setup_logging(
+    verbose: u8,
+    no_color: bool,
+    log_path: Option<PathBuf>,
+) -> Option<WorkerGuard> {
+    let ockam_crates = [
+        "ockam",
+        "ockam_node",
+        "ockam_core",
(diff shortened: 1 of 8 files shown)


implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -496,15 +503,35 @@ pub fn setup_logging(verbose: u8, no_color: bool) {
...
+                log_path,
+                RollingConditionBasic::new()
+                    .daily()
+                    .max_size(100 * 1024 * 1024), // Rotate every 100 MB or day
+                60, // Up to 6 GB or 60 days
+            )
+            .expect("Failed to create rolling file appender");
+            let (n, guard) = tracing_appender::non_blocking(r);
+            let fmt = tracing_subscriber::fmt::Layer::default()
+                .with_ansi(!no_color)
```
shanesveller: Just for my own curiosity - when we spawn new subprocesses, are we setting no_color ourselves, or forwarding whatever setting was chosen in the original CLI command?

adrianbenavides: We set it ourselves ([here for example](https://github.com/build-trust/ockam/blob/faaf3eb8dfa6f6c47041692a634d3b6dbf56d7b6/implementations/rust/ockam/ockam_command/src/node/util.rs#L221)) because we know at that point that we are going to redirect stdout to the node's stdout.log file. Does that make sense to you?

shanesveller: Yes it does!

implementations/rust/ockam/ockam_command/src/logs/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+) -> Option<WorkerGuard> {
+    let ockam_crates = [
+        "ockam",
+        "ockam_node",
+        "ockam_core",
+        "ockam_command",
+        "ockam_identity",
+        "ockam_transport_tcp",
+        "ockam_vault",
+        "ockam_vault_sync_core",
```
etorreborre: This crate does not exist anymore

implementations/rust/ockam/ockam_command/src/logs/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+                .with_ansi(color)
+                .with_writer(n);
+            (subscriber.with(fmt).try_init(), Some(guard))
+        }
+        (_, Some(log_path)) => {
+            let r = RollingFileAppender::new(
+                log_path,
+                RollingConditionBasic::new()
+                    .daily()
+                    .max_size(100 * 1024 * 1024), // Rotate every 100 MB or day
```
etorreborre: Shouldn't we put those 2 values in the global args in order to be able to fix them at deployment time?

etorreborre (dismissed):

implementations/rust/ockam/ockam_command/src/logs/mod.rs
adrianbenavides: I've added env variables for those values. Would you also add them as CLI arguments?

etorreborre: Environment variables is probably good enough

etorreborre (approved):

PR #4786, opened 2023-04-25
fix(rust): log creation of default node/identity to stderr

adrianbenavides:


caoakleyii:
Just curious why this should be going to `stderr` as the title suggests?

~~As well as, how this achieves this - my understanding is this would go to `stdout`.~~
Ah I see the `shell.terminal().write_line()` goes to stdout, but shell.write_line()` goes to stderr.

I think we should make that a bit more clear; though i'm still curious on the above question.

caoakleyii (approved):
My above comment isn't a blocker.

I think that's something more apt to discuss surrounding the interface of our shell/terminal structs.

adrianbenavides:
The terminal abstraction is built around the concepts explained at https://clig.dev/ (ctrl-f “Send messaging to stderr” to go to the relevant section).

As for the API, when you want to write to stdout, it consumes the instance to try to give a hint that it's a final operation. The terminal has two writing modes (ToStdErr and ToStdOut), each of them with its own available set of methods. Current code is [here](https://github.com/build-trust/ockam/blob/faaf3eb8dfa6f6c47041692a634d3b6dbf56d7b6/implementations/rust/ockam/ockam_command/src/terminal/mod.rs#L18) although a small rename has been done at https://github.com/build-trust/ockam/pull/4652

PR #4795, opened 2023-04-26
refactor(rust): apply cli_state abstraction

### LINKED ISSUE #4410: Create traits to abstract the configuration files handled by the CLI state
### Current behavior

Currently the different resources within the CLI State are all implemented independently but share a very similar structure.

Those CLI State resources are used and defined within [here](https://github.com/build-trust/ockam/blob/1ef02c9c07603b5de776563e6861cfdf1aae3dce/implementations/rust/ockam/ockam_api/src/cli_state.rs#L60-L67).

### Desired behavior

Abstract many of the existing systems and create a common structure for the way CLI State resources are persisted and accessed.

adrianbenavides:
Fixes https://github.com/build-trust/ockam/issues/4410

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
@@ -4,7 +4,7 @@ use crate::cli_state::traits::{StateDirTrait, StateItemTrait};
 use crate::cli_state::CliStateError;
 use ockam_identity::{
     Identities, IdentitiesRepository, IdentitiesStorage, IdentitiesVault, Identity,
-    IdentityChangeHistory, IdentityHistoryComparison, IdentityIdentifier, LmdbStorage,
+    IdentityHistoryComparison, IdentityIdentifier, LmdbStorage,
 };
 use serde::{Deserialize, Serialize};
 use std::fmt::{Display, Formatter};
@@ -33,7 +33,7 @@ impl IdentitiesState {
 
         let identity_state = identities
             .into_iter()
-            .find(|ident_state| &ident_state.config.identifier == identifier);
+            .find(|ident_state| &ident_state.config.identity.identifier() == identifier);
 
         match identity_state {
             Some(is) => Ok(is),
@@ -65,7 +65,7 @@ pub struct IdentityState {
 
 impl IdentityState {
     pub async fn get(&self, vault: Arc<dyn IdentitiesVault>) -> Result<Identity> {
-        let data = self.config.change_history.export()?;
+        let data = self.config.identity.export()?;
         Ok(self
             .make_identities(vault)
             .await?
@@ -100,7 +100,9 @@ impl IdentityState {
 
     fn in_use_by(&self, nodes: &[NodeState]) -> Result<()> {
         for node in nodes {
-            if node.config().identity_config()?.identifier == self.config.identifier {
+            if node.config().identity_config()?.identity.identifier()
+                == self.config.identity.identifier()
+            {
                 return Err(CliStateError::Invalid(format!(
                     "Can't delete identity '{}' because is currently in use by node '{}'",
                     &self.name,
@@ -120,7 +122,11 @@ impl Display for IdentityState {
             self.path.as_path().file_stem().unwrap().to_str().unwrap()
         )?;
         writeln!(f, "State Path: {}", self.path.clone().to_str().unwrap())?;
-        writeln!(f, "Config Identifier: {}", self.config.identifier)?;
+        writeln!(
+            f,
+            "Config Identifier: {}",
+            self.config.identity.identifier()
+        )?;
         match &self.config.enrollment_status {
             Some(enrollment) => {
                 writeln!(f, "Enrollment Status:")?;
@@ -136,32 +142,26 @@ impl Display for IdentityState {
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct IdentityConfig {
-    pub identifier: IdentityIdentifier,
-    pub change_history: IdentityChangeHistory,
+    pub identity: Identity,
     pub enrollment_status: Option<EnrollmentStatus>,
 }
 
 impl IdentityConfig {
     pub async fn new(identity: &Identity) -> Self {
-        let identifier = identity.identifier();
-        let change_history = identity.change_history();
         Self {
-            identifier,
-            change_history,
+            identity: identity.clone(),
             enrollment_status: None,
         }
     }
 
     pub fn identity(&self) -> Identity {
-        Identity::new(self.identifier.clone(), self.change_history.clone())
+        self.identity.clone()
     }
 }
 
 impl PartialEq for IdentityConfig {
     fn eq(&self, other: &Self) -> bool {
-        self.identifier == other.identifier
-            && self.change_history.compare(&other.change_history)
-                == IdentityHistoryComparison::Equal
+        self.identity.compare(&other.identity) == IdentityHistoryComparison::Equal
     }
 }
 
--- a/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
@@ -1,6 +1,7 @@
 use super::Result;
+use crate::cli_state::nodes::NodeState;
 use crate::cli_state::traits::{StateDirTrait, StateItemTrait};
-use crate::cli_state::{CliStateError, NodeState};
+use crate::cli_state::CliStateError;
 use ockam_identity::{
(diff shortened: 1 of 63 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
```diff
@@ -0,0 +1,312 @@
...
+            }
+            None => (),
+        }
+        Ok(())
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct IdentityConfig {
+    pub identifier: IdentityIdentifier,
```
etorreborre: Note: those 2 fields could be replaced with `pub identity: Identity`

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
```diff
@@ -8,9 +8,12 @@ use super::Result;
 /// Represents the directory of a type of state and contains
 /// all the data related to that type.
```
etorreborre: It would nice to extend the description a bit to give a better mental model of what's going on:

 - the `StateDirTrait` represents a list of items, uniquely identified by a name and represented by the same `Item` type
 - all the items are persisted under the same directory
 - one item is marked as being the "default" item. It is either
   - the first created item
   - one of the existing items, marked as being the default one with the `set_default` method

Additionally some unit test would show what happens when you create/list/set_default/overwrite etc...

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
```diff
@@ -37,49 +40,94 @@ pub trait StateTrait: Sized {
...
+            Err(e) => return Err(e),
+        };
+        // If it's the default, remove link
+        if let Ok(default) = self.default() {
+            if default.path() == s.path() {
+                let _ = std::fs::remove_file(self.default_path()?);
+            }
+        }
+        // Remove state data
+        s.delete()?;
```
etorreborre: Just `s.delete()` without the final `Ok(())` should work here.

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
```diff
@@ -100,24 +150,41 @@ pub trait StateTrait: Sized {
     fn has_default(&self) -> Result<bool> {
```
etorreborre: I am not sure if this method is absolutely needed. The existence of a default should be guaranteed by the fact that at least one item has been created. So `is_empty` would be a more direct way to test for this condition IMO.

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
```diff
@@ -37,49 +40,94 @@ pub trait StateTrait: Sized {
...
+        }
+        let state = Self::Item::new(self.path(name), config)?;
+        if !self.default_path()?.exists() {
+            self.set_default(name)?;
+        }
+        Ok(state)
     }
 
-    fn list(&self) -> Result<Vec<Self::ItemDir>> {
+    async fn create_async(
```
etorreborre: Can we move this function to `VaultsState` if it is only used there for now?

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
```diff
@@ -100,24 +150,41 @@ pub trait StateTrait: Sized {
...
 
 /// This trait defines the methods to retrieve an item from a state directory.
 /// The details of the item are defined in the `Config` type.
 #[async_trait]
-pub trait StateItemDirTrait: Sized {
-    type Config: StateItemConfigTrait + Serialize + for<'a> Deserialize<'a>;
+pub trait StateItemTrait: Sized {
+    type Config: Serialize + for<'a> Deserialize<'a> + Send;
+
+    fn cli_state(&self) -> Result<CliState> {
```
etorreborre, line 170: This is fishy. An item shouldn't have to be able to access the global state. If we have a look at where it used, it is used to check a dependency between identity and node when trying to delete an identity. 

I propose that:

 - the code for deleting an identity state does just that, and does not try to check if a running node uses it with `identity.in_use` (which btw throws an error while the name implies that it should just return a boolean)
 
 - the `DeleteCommand` which has access to the global `CliState` calls a function on that state to check the dependency between the identity state and the nodes state, then decides if it's okay to call `identity_state.delete()`

This makes the surface of the the `StateItemTrait` smaller and makes each state code more independent of other state code.

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
```diff
@@ -100,24 +150,41 @@ pub trait StateTrait: Sized {
...
+        let contents = serde_json::to_string(self.config())?;
+        std::fs::write(self.path(), contents)?;
+        Ok(())
+    }
+    fn delete(&self) -> Result<()> {
+        std::fs::remove_file(self.path())?;
+        Ok(())
+    }
     fn name(&self) -> &str;
     fn path(&self) -> &PathBuf;
```
etorreborre, line 196: We shouldn't have to expose the `path` and `data_path` functions. The purpose of the `StateItem` trait is to manage the persistence of data somewhere and return some well-formed data as `Config`. 

For example in the case of a `VaultState` I think that `Config` should be a `VaultConfig` to be fed directly into a `Vault::create_from_config(vault_config)` function and the implementation of `VaultState` should sort out that the persisted data needs to a be a json file + some storage file. By exposing `path` and `data_path` we are breaking the encapsulation that we are looking for with those traits because a user of that path needs to know exactly what they can or cannot do with it.

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
```diff
@@ -100,24 +150,41 @@ pub trait StateTrait: Sized {
...
+            self.path()
+                .parent()
+                .expect("no parent dir")
+                .parent()
+                .expect("no parent dir"),
+        )
+    }
 
     fn new(path: PathBuf, config: Self::Config) -> Result<Self>;
     fn load(path: PathBuf) -> Result<Self>;
```
etorreborre, line 184: We could get rid of the `load` method by caching the `Config` on the first access to the `config` method.

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
```diff
@@ -8,9 +8,12 @@ use super::Result;
 /// Represents the directory of a type of state and contains
 /// all the data related to that type.
 #[async_trait]
-pub trait StateTrait: Sized {
-    type ItemDir: StateItemDirTrait;
-    type ItemConfig: StateItemConfigTrait + Serialize + for<'a> Deserialize<'a>;
+pub trait StateDirTrait: Sized {
+    type Item: StateItemTrait;
+
+    fn cli_state(&self) -> Result<CliState> {
```
etorreborre: This function is not used.

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
```diff
@@ -8,9 +8,12 @@ use super::Result;
...
+        CliState::new(self.dir().parent().expect("no parent dir"))
+    }
 
     fn new(dir: PathBuf) -> Self;
     fn default_filename() -> &'static str;
```
etorreborre: I think we should make a clear distinction between:

 - a template / some helper functions to load / persist a `StateDir` and not expose the various `dir` and `path` functions. If there is a `data` dir for a given implementation, the `load` function can do the `std::fs::create_dir_all(dir.join("data"))?;` by itself and maybe use `DATA: &str = "data"` as a constant so we make sure that we are consistent in terms of naming
  
 - the `StateDir` interface which is basically
   - `create(name, config)`
   - `overwrite(name, config)`  
   - `list()`
   - `get(name)`
   - `delete(name)`

etorreborre (commented):
I think that this is step in the right direction but I also think that we should absolutely go one step further:

 - keep all these paths hidden as implementation details
 - use well-typed interfaces to `list, get, delete,...` well-typed configs 

@adrianbenavides what I wrote above still implies a lot of work so feel free to leave it to further PRs

adrianbenavides:
Thanks @etorreborre for the review, that was very helpful 🙏 ! I'll apply the most superficial changes, and I'll tackle the rest in a separate PR so we can also unblock the PRs that depend on this one.

adrianbenavides:
I've addressed a bunch of the comments (closed now) and created [an issue](https://github.com/build-trust/ockam/issues/4829) with all the comments I didn't fix. Feel free to close the comments if you think that's okey.

etorreborre (approved):
Great, let's tackle improvements in some follow-up PRs

Issue #4796, opened 2023-04-26
Auto-reconnect retry without waiting


davide-baldo:
### Observed behavior

When an inlet or a forwarder route timeouts, the sessions checker will invoke a replacer to re-create a new route, but if it fails right away it'll try again right away, without waiting a fixed amount of time, spamming the node log.

### Steps to reproduce

1. ockam reset
2. ockam node create blue
3. ockam node create green
4. ockam tcp-outlet create --at /node/green --to 127.0.0.1:2048
5. ockam tcp-inlet create --at /node/blue --to /node/green/service/outlet
6. ockam node delete green

in the ~/.ockam/nodes/blue/stdout.log

```
DEBUG ockam_api::session: failed to send ping...
DEBUG ockam_api::session: failed to send ping...
DEBUG ockam_api::session: failed to send ping...
 WARN ockam_api::session: session unresponsive key=219c5e5aee2ad26744c47988948a15d6fa6d344b065ea8b8
 INFO ockam_api::session: replacing session key=219c5e5aee2ad26744c47988948a15d6fa6d344b065ea8b8
 ....
 WARN ockam_api::session: replacing session failed key=219c5e5aee2ad26744c47988948a15d6fa6d344b065ea8b8
 INFO ockam_api::session: replacing session key=219c5e5aee2ad26744c47988948a15d6fa6d344b065ea8b
 ...
```

The connection is attempted right away, using 100% of the CPU and spamming the log

### Expected behavior

Before attempting to reconnect, a minimum amount of time must pass, a minimum of 15 seconds could be a good enough compromise.

### Ockam Version

f9ef9249de41e4f172452cbd478c5c716a675a84

jjdoar:
I can help with this

mrinalwadhwa:
@jjdoar that's awesome. Let us know if you have any questions

jjdoar:
I opened a PR with a first attempt, #4805 , however I was unable to recreate the original issue using the **Steps to Reproduce**. I see an error on step 4:

`error: the following required arguments were not provided:
  --from <OUTLET_ADDRESS>`
  
  I tried passing a different local address like `--from 127.0.0.1:8080` and following the rest of the steps, but this did not produce the **Observed behavior**

PR #4805, opened 2023-04-27
feat(rust): Add delay before attempting to reconnect when replacing sessions

jjdoar:
## Current behavior

When an inlet or a forwarder route timeouts, the sessions checker will invoke a replacer to re-create a new route, but if it fails right away it'll try again right away, without waiting a fixed amount of time, spamming the node log.

## Proposed changes

Add a 15 second delay between attempts to replace the session as suggested in [https://github.com/build-trust/ockam/issues/4796](url)

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/session/mod.rs
+++ b/implementations/rust/ockam/ockam_api/src/session/mod.rs
@@ -6,7 +6,9 @@ use minicbor::{Decode, Encode};
 use ockam::{LocalMessage, Route, TransportMessage, Worker};
 use ockam_core::compat::sync::{Arc, Mutex};
 use ockam_core::flow_control::{FlowControlPolicy, FlowControls};
-use ockam_core::{route, Address, AllowAll, Decodable, DenyAll, Encodable, Error, Routed, LOCAL};
+use ockam_core::{
+    route, Address, AllowAll, Decodable, DenyAll, Encodable, Error, Result, Routed, LOCAL,
+};
 use ockam_node::tokio;
 use ockam_node::tokio::sync::mpsc;
 use ockam_node::tokio::task::JoinSet;
@@ -15,7 +17,7 @@ use ockam_node::Context;
 use tracing as log;
 
 const MAX_FAILURES: usize = 3;
-const RETRY_DELAY: Duration = Duration::from_secs(15);
+const RETRY_DELAY: Duration = Duration::from_secs(5);
 const DELAY: Duration = Duration::from_secs(3);
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_api/src/session/mod.rs
```diff
@@ -118,6 +121,7 @@ impl Medic {
                                 session.set_status(Status::Down);
                                 log::info!(%key, "replacing session");
                                 self.replacements.spawn(async move { (key, f.await) });
+                                let _ = sleep(self.retry_delay).await;
```
davide-baldo: If we sleep here, the whole `Medic` instance would be frozen for 15 seconds, since the async context is shared for every session.
If we move the sleep inside the spawned async context on line 123 that would block only that specific session.

davide-baldo (dismissed):
Hi! Thank you for your contribution!
There is just a minor issue

implementations/rust/ockam/ockam_api/src/session/mod.rs
```diff
@@ -117,7 +120,11 @@ impl Medic {
                                 let f = session.replacement(session.ping_route().clone());
                                 session.set_status(Status::Down);
                                 log::info!(%key, "replacing session");
-                                self.replacements.spawn(async move { (key, f.await) });
+                                self.replacements.spawn(async move {
+                                    let replacement = (key, f.await);
```
davide-baldo: The replacer starts from the moment you start waiting for the future, `f.await`.
Here you wait **after** the connection has been attempted, it would be more beneficial to wait **before** the connection attempt, to give time to external services to recover (like a service or a connection restarting).

etorreborre:
Hi @jjdoar. I proposed an evolution of your fix in the latest commit:

 - even with @davide-baldo's suggestion your fix was not fully working because there was another attempt to replace a session in `get_results`
 
 - not only that but the replacement was started in a loop which was trying to collect the results of all active replacements
 
 - so I kept the `get_results` loop merely as a way to collect results and change local state
 
 - and introduced a new `Status::DEGRADED` with the idea that this status represents a session that is unresponsive but currently trying to reconnect again. If, in the main loop, we end-up with such a session we don't try to replace one more time, we just wait on `get_results` to tell us if the current attempt was successful or not

Please tell me what you think (you to @davide-baldo). 

I also suggest to add a bats test in `implementations/rust/ockam/ockam_command/tests/bats/unit.bats` showing that it is possible to reconnect after a while. Something like
```
ockam reset -y
ockam node create blue
ockam node create green -t 127.0.0.1:4001
ockam tcp-outlet create --at /node/green --to 127.0.0.1:2048
ockam tcp-inlet create --at /node/blue --to /node/green/service/outlet

# stop and restart the green node
ockam node delete green
sleep 3 # wait a bit to make sure that the blue node registers that the green node is down
ockam node create green -t 127.0.0.1:4001

# check that the blue node has reconnected to the green node
run curl --fail --head --max-time 10 "127.0.0.1:2048"
assert_success
```

davide-baldo (approved):
We might get an extra delay (wait until the next loop), but I think it's a non-issue :+1:

etorreborre:
@jjdoar thanks for working on this! If you want please have a go at writing some bats tests for the reconnection. I think this is a domain where we could be better covered

PR #4823, opened 2023-04-28
fix(elixir): fix error result on the nif call

polvorin:
The nif code defined a function named `err`,  called to return an error tuple  `{:error, reason}`   to the caller.  
But (at least on linux) that function was _not_ what ended up being called,  as https://man7.org/linux/man-pages/man3/err.3.html took precedence.
As as result,  on error conditions when using a vault _from elixir code_ ,   we ended up with text being printed on the console like

```
beam.smp: failed to aead_aes_gcm_decrypt
``` 
(note it's exactly what the err function would print,  the program name and colon) followed by undefined behavior as the return is not a valid erl term. 

I didn't try to dig into the name collision,  just renamed the function to `error_tuple`.   Also use the defined constant [`ERL_NIF_LATIN1` for string types](https://www.erlang.org/doc/man/erl_nif.html#ErlNifCharEncoding).    (since we use this mainly from elixir,  at some point might be better to change these to elixir string that are utf8 binaries,   or just predefine a set of possible error atoms)

### DIFF
--- a/implementations/elixir/ockam/ockam_vault_software/native/vault/software/common.c
+++ b/implementations/elixir/ockam/ockam_vault_software/native/vault/software/common.c
@@ -20,9 +20,9 @@ ERL_NIF_TERM ok(ErlNifEnv *env, ERL_NIF_TERM result) {
     return enif_make_tuple2(env, id, result);
 }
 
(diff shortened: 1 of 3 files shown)


davide-baldo (approved):
nice catch!

PR #4837, opened 2023-05-02
fix(rust): parsing `GlobalArgs` from input

adrianbenavides:
Fixes https://github.com/build-trust/ockam-artifacts/actions/runs/4855334561/jobs/8653787007


etorreborre (approved):
A quick unit test would be nice, if you have the time for it :-)

PR #4838, opened 2023-05-02
refactor(rust): update how we handle user confirmation on `reset` command

adrianbenavides:


implementations/rust/ockam/ockam_command/src/reset.rs
```diff
@@ -11,33 +13,32 @@ pub struct ResetCommand {
...
+            std::process::exit(e.code());
         }
     }
 }
 
-fn run_impl(opts: CommandGlobalOpts) -> crate::Result<()> {
+fn run_impl(opts: CommandGlobalOpts, cmd: ResetCommand) -> crate::Result<()> {
+    match opts
+        .terminal
+        .confirm("This will delete the local CLI state. Are you sure?")?
```
mrinalwadhwa: We need a better message here. CLI state has no meaning to end users.

adrianbenavides: I changed it to "local Ockam configuration". The point I want to make is that the command only affects local stuff.

etorreborre: @mrinalwadhwa are you ok with that wording? Can we merge?

mrinalwadhwa: That change is good but there is another reference to CLI state in this PR

etorreborre: Right, in the last commit I changed the wording everywhere CLI state was used

adrianbenavides: Thanks @etorreborre ! I didn't see the other reference and forgot to follow up on this

metaclips (approved):

PR #4848, opened 2023-05-03
Fix links and missing sections in READMEs

etorreborre:
This PR fixes some broken links in the `ockam` crate README since all the documentation has now moved over to `docs.ockam.io`. 

The PR also adds some missing badges and sections (usage, licence) for some of the other READMEs.

### DIFF
--- a/implementations/rust/ockam/ockam_command/README.md
+++ b/implementations/rust/ockam/ockam_command/README.md
@@ -8,7 +8,7 @@
 Orchestrate end-to-end encryption, mutual authentication, key management,
 credential management, and authorization policy enforcement — at scale.
 
-This crates provides the `ockam` command line application to start Ockam nodes and interact with them.
+This crate provides the `ockam` command line application to start Ockam nodes and interact with them.
 
 ## Usage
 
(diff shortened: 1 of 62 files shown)


implementations/rust/ockam/ockam_command/README.md
```diff
@@ -8,6 +8,8 @@
 Orchestrate end-to-end encryption, mutual authentication, key management,
 credential management, and authorization policy enforcement — at scale.
 
+This crates provides the `ockam` command line application to start Ockam nodes and interact with them.
```
adrianbenavides: Since the CLI is not limited to work with nodes, maybe we could add more information here, similar to what we have here: https://docs.ockam.io/reference/command

adrianbenavides (commented):
Looks great! Just a small comment on the ockam_command readme

implementations/rust/ockam/ockam_command/README.md
etorreborre: Thanks @adrianbenavides I expanded slightly the description and then put a link the the command doc directly, since I don't want to duplicate the information that is already there.

I'm putting back this PR as a draft until I can fix the failing CI job(s) and I have added a CI check to make sure that the readmes are up to date.

adrianbenavides (approved):

PR #4849, opened 2023-05-03
build: publish the ockam_transport_udp crate

etorreborre:


implementations/rust/ockam/ockam_transport_udp/Cargo.toml
```diff
@@ -15,7 +15,6 @@ edition = "2021"
 homepage = "https://github.com/ockam-network/ockam"
 keywords = ["ockam", "crypto", "network", "networking", "udp"]
 license = "Apache-2.0"
-publish = false
```
metaclips, line 18: Publish should be true for it to be releasable

metaclips (dismissed):

implementations/rust/ockam/ockam_transport_udp/Cargo.toml
etorreborre: thanks @metaclips

metaclips (approved):

PR #4851, opened 2023-05-03
fix(elixir): report duration metrics for asynchronous APIs

hairyhum:
## Current behavior

Currently we only report request duration when the api handler finishes. With asynchronous APIs we might have api handlers returning :noreply and then calling `API.reply`.

## Proposed changes

To support that added request duration reporting in `reply` function

## Checks

### DIFF
--- a/implementations/elixir/ockam/ockam_services/lib/services/api.ex
+++ b/implementations/elixir/ockam/ockam_services/lib/services/api.ex
@@ -50,13 +50,16 @@ defmodule Ockam.Services.API do
   """
   @callback path_group(String.t()) :: String.t()
 
+  def path_group(_path), do: "all"
+
   @doc """
   Send Ockam.API.Response to the from_route of the reply using address as a return_route
   """
-  def reply(%Request{} = request, status, body, address) do
+  def reply(%Request{} = request, status, body, address, module \\ __MODULE__) do
     response = Response.reply_to(request, status_code(status), body)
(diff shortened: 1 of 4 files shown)


implementations/elixir/ockam/ockam_services/lib/services/api.ex
```diff
@@ -92,20 +94,22 @@ defmodule Ockam.Services.API do
   def handle_request(module, request, state) do
     case module.handle_request(request, state) do
       {:reply, status, body, state} ->
-        reply = reply(request, status, body, state.address)
+        reply = reply(request, status, body, state.address, module)
         {reply, state}
 
       {:noreply, state} ->
+        ## No response, but still emit reply=false stop event
+        emit_request_stop(module, request, :noreply, state.address)
```
mszpakowski, line 102: Do we use this `{:noreply, state}` return value for the API worker, other than in the async API handlers?

I'm wondering if we could simply not emit the stop event here? Or differentiate between situations where we want to do it and those where we don't

hairyhum: I think it still makes sense to report how long does it take to handle the request outside of async ops. We can differentiate these durations by `reply=true/false` tag.

mszpakowski (approved):

SanjoDeundiak (approved):

polvorin (approved):

PR #4875, opened 2023-05-05
docs: update readme from `websocket` crate

adrianbenavides:
Looks like there was some (automatic) changes after the last release causing the [CI check](https://github.com/build-trust/ockam/actions/runs/4889298233/jobs/8727799339?pr=4874) to fail.

### DIFF
--- a/implementations/rust/ockam/ockam_transport_websocket/README.md
+++ b/implementations/rust/ockam/ockam_transport_websocket/README.md
@@ -38,7 +38,7 @@ use ockam_core::{AllowAll};
 use ockam_node::NodeBuilder;
 use ockam_macros::node;
 
-#[ockam_macros::node(crate = "0.72.0")]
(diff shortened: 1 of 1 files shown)


etorreborre (approved):
I wonder how that happened. This can be the result of the `cargo release` task doing a version replacement where it shouldn't

adrianbenavides:
I suspect is something related to that. @metaclips could you check that please? I guess we could remove those docs examples if there is no easy fix (we don't have docs examples in any other crate anyways).

etorreborre:
I hope we can find a fix because the ability to add one or two typechecked examples is nice IMO. 
I suppose that we can avoid doing any regex replacement on READMEs since version numbers are filled in with the `cargo readme` task.

etorreborre:
Here is a fix: https://github.com/build-trust/ockam/pull/4877

PR #3841, opened 2022-11-17
chore(rust) adding rotate-key to CLI

noyez:
This commit adds the rotate-key command to the CLI. It can be used via `ockam identity rotate-key --node n1 --label OCKAM_RK` or the label argument can be omitted to use the default label of "OCKAM_RK", e.g. `ockam identity rotate-key --node n1`.

In response to issue: https://github.com/build-trust/ockam/issues/3685

Note: There is a known issue with this command. it will fail after it is executed after `ockam identity show` command. I'm not sure if that error is related to this PR, however i'm sure it's not supposed to fail. 

## Current Behavior

`ockam identity rotate-key` did not exists.

## Proposed Changes

create cli command `ockam identity rotate-key`

## Checks

- [x] All commits in this Pull Request are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and Verified by Github.
- [x] All commits in this Pull Request follow the Ockam [commit message convention](https://github.com/build-trust/.github/blob/main/CONTRIBUTING.md#commit-messages).
- [x] I accept the Ockam Community [Code of Conduct](https://github.com/build-trust/.github/blob/main/CODE_OF_CONDUCT.md).
- [x] I have accepted the Ockam [Contributor License Agreement](https://github.com/build-trust/ockam-contributors/blob/main/CLA.md) by adding my Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/blob/main/CONTRIBUTORS.csv) file in a separate pull request to the [build-trust/ockam-contributors](https://github.com/build-trust/ockam-contributors) repository.

Issue #3685, opened 2022-10-19
Add ockam sub command to rotate identity keys


mrinalwadhwa:
# Currently

The following will create and identity

```
ockam node create n1
ockam identity show --node n1
```

The code for it is
- https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/node/create.rs#L278-L284
- https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/nodes/service/services.rs#L42
- https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/nodes/service/identity.rs#L12

# Desired

It is possible to rotate identity keys
https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_identity/src/identity.rs#L181

We want to support this as a command.

---

<i>We love helping new contributors! ❤️
If you have questions or need help as you explore, please join us on [Discord](https://discord.gg/bewvnm6zqS). If you're looking for other issues to contribute to, please checkout our [good first issues](https://github.com/build-trust/ockam/labels/good%20first%20issue).</i>

noyez:
Hello -- Do you envision the CLI to be something along the lines of `ockam identity rotate --node n1 --key key-label`? Other ideas: 
 * have rotate-key be a subcommand of `ockam node ... `
 * have roate-key be a top-level command, like `ockam rotate ... `

mrinalwadhwa:
@noyez Thank you for looking into this!!
I was thinking of something like this:

```
ockam identity rotate --node n1 --key key-label
```

noyez:
Hello -- I've been looking at this issue and I have a few questions.

 * What is the end result of rotating the keys? should I be seeing a different set of keys with the `ockam identity show --full` output? Or is the only observable effect in the `default_value.json` file? 
 * To rotate the keys, it seems I need to create an `Identity` struct. I think i can piece that together by getting the `IdentityChangeHistory`, and the `Vault` from the `~/.config/ockam-cli/default_vault.json`. But there appears to be done two different ways to create the `Identity` struct.
     1. `Identity::new( )`
     2. `IdentityBuilder::build( )`
     
Is one preferred over the other? It seems that `IdentityBuilder::build()` does a lot more then eventually calls `Identity::new()`. 

Thanks for any input and clarifications!

SanjoDeundiak:
[...]

PR #4876, opened 2023-05-05
feat(rust): adding rotate-key to ockam_command cli

noyez:
This commit adds the rotate-key command to the CLI. It can be used via `ockam identity rotate-key --node n1 --label OCKAM_RK` or the label argument can be omitted to use the default label of "OCKAM_RK", e.g. `ockam identity rotate-key --node n1`.

In response to issue: https://github.com/build-trust/ockam/issues/3685

This is a replacement for https://github.com/build-trust/ockam/pull/3841

## Current Behavior

`ockam identity rotate-key` did not exists.

## Proposed Changes

create cli command `ockam identity rotate-key`

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam/src/node.rs
+++ b/implementations/rust/ockam/ockam/src/node.rs
@@ -254,6 +254,13 @@ impl Node {
         self.identities().identities_keys()
     }
 
+    /*
+    pub fn rotate_keys(&self) -> Arc<IdentitiesKeys> {
+        todo!();
+        self.identities().identities_keys().rotate_key();
+    }
+    */
+
     /// Return services to manage credentials
     pub fn credentials(&self) -> Arc<dyn Credentials> {
         self.identities().credentials()
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -529,11 +529,11 @@ impl NodeManager {
 impl NodeManagerWorker {
     //////// Request matching and response handling ////////
 
-    async fn handle_request(
+    async fn handle_request<'a>(
         &mut self,
         ctx: &mut Context,
-        req: &Request<'_>,
-        dec: &mut Decoder<'_>,
+        req: &Request<'a>,
+        dec: &mut Decoder<'a>,
     ) -> Result<Vec<u8>> {
         debug! {
             target: TARGET,
@@ -566,6 +566,10 @@ impl NodeManagerWorker {
                         node_manager.transports.len() as u32,
                     ))
                     .to_vec()?
+            }
+
+            (Post, ["node", "identity", "actions", "rotate_key"]) => {
+                self.rotate_key(req, dec).await?.to_vec()?
             }
 
             // ==*== Tcp Connection ==*==
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
@@ -8,6 +8,7 @@ use ockam::Result;
 use ockam_core::api::{Request, Response, ResponseBuilder};
 use ockam_transport_tcp::{TcpConnectionOptions, TcpListenerOptions};
 use std::net::{AddrParseError, SocketAddr};
+use crate::nodes::models::identity::{RotateKeyResponse, RotateKeyRequest};
 
 use super::NodeManagerWorker;
 
@@ -53,6 +54,23 @@ impl NodeManagerWorker {
             }
         };
         Ok(res)
+    }
+
+    pub(super) async fn rotate_key<'a>(
+        &mut self,
+        req: &Request<'a>,
+        dec: &mut Decoder<'a>,
+    ) -> Result<ResponseBuilder<RotateKeyResponse<'_>>> {
+        let node_manager = self.node_manager.read().await;
+        let identities = node_manager.identities();
+        let rotate_args = dec.decode::<RotateKeyRequest>()?;
+        let identities_keys = identities.identities_keys();
+        let identity = &mut node_manager.identity();
+        identities_keys.rotate_key(identity, &rotate_args.label).await?;
+
+        let response =
+            Response::ok(req.id()).body(RotateKeyResponse::new(rotate_args.label.to_string()));
+        Ok(response)
     }
 
     pub(super) async fn add_transport<'a>(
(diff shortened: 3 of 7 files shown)


implementations/rust/ockam/ockam_command/src/identity/rotate_key.rs
```diff
@@ -0,0 +1,40 @@
...
+fn default_key_label() -> String {
+    "OCKAM_RK".to_string()
+}
+
+#[derive(Clone, Debug, Args)]
+pub struct RotateKeyCommand {
+    #[command(flatten)]
+    node_opts: NodeOpts,
+    #[arg(short, long, default_value_t=default_key_label())]
+    label: String,
```
SanjoDeundiak, line 18: Do you think this argument could be optional so we can use the default value where we handle this request?

implementations/rust/ockam/ockam_command/src/identity/rotate_key.rs
```diff
@@ -0,0 +1,40 @@
...
+        node_rpc(run_impl, (options, self))
+    }
+}
+
+async fn run_impl(
+    ctx: Context,
+    (opts, cmd): (CommandGlobalOpts, RotateKeyCommand),
+) -> crate::Result<()> {
+    let mut rpc = Rpc::background(&ctx, &opts, &cmd.node_opts.api_node)?;
+    // XXX Should this be post?
```
SanjoDeundiak, line 32: I think that's correct

implementations/rust/ockam/ockam_command/tests/bats/unit.bats
```diff
@@ -81,6 +81,23 @@ teardown() {
   assert_output --partial "/service/vault_service"
 }
 
+@test "create a node and show its identity then rotate keys" {
```
SanjoDeundiak, line 84: Could you please look why this test doesn't pass?

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
```diff
@@ -55,6 +56,23 @@ impl NodeManagerWorker {
         Ok(res)
     }
 
+    pub(super) async fn rotate_key<'a>(
```
SanjoDeundiak, line 59: I think this function should be placed inside the `node_identities.rs` or have a dedicated file, like `identity.rs`

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
```diff
@@ -55,6 +56,23 @@ impl NodeManagerWorker {
...
+        &mut self,
+        req: &Request<'a>,
+        dec: &mut Decoder<'a>,
+    ) -> Result<ResponseBuilder<RotateKeyResponse<'_>>> {
+        let node_manager = self.node_manager.read().await;
+        let identities = node_manager.identities();
+        let rotate_args = dec.decode::<RotateKeyRequest>()?;
+        let identities_keys = identities.identities_keys();
+        let identity = &mut node_manager.identity();
```
SanjoDeundiak, line 68: `NodeManager::identity()` returns a clone of the `Identity` struct, so this rotate call won't actually change anything on NodeManager's Identity

SanjoDeundiak: Do you think we could access the field directly here?

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -529,11 +529,11 @@ impl NodeManager {
 impl NodeManagerWorker {
     //////// Request matching and response handling ////////
 
-    async fn handle_request(
+    async fn handle_request<'a>(
```
SanjoDeundiak, line 532: Why is this change required?

implementations/rust/ockam/ockam/src/node.rs
```diff
@@ -254,6 +254,13 @@ impl Node {
         self.identities().identities_keys()
     }
 
+    /*
```
SanjoDeundiak, line 257: Could you please remove it?

SanjoDeundiak (changes requested):
Nice job! Have few comments

SanjoDeundiak (changes requested):
Nice job! Have few comments

etorreborre:
Hi @noyez! Do you need any help for completing this PR?

noyez:
I think i need to pull in some code from master to get the tests passing. My dev machine is on the fritz and so i've been trying to get to get it to boot again.

davide-baldo:
Your work is much appreciated! Is there anything we can do to support you in completing this task?

implementations/rust/ockam/ockam_command/src/identity/rotate_key.rs
etorreborre: I am actually not concerned that we could infer a default value. I am more concerned by the fact that there is no argument to select a specific identity :-).

I would actually not use `node` as an argument for the command but rather `identity` since when you create an identity you don't specify a node. The other argument which should also be added IMO is `vault` since you can specify a specific vault when creating an identity.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
noyez: This was puzzling to me too. w/o it i received this compile error. 

```error: lifetime may not live long enough
   --> implementations/rust/ockam/ockam_api/src/nodes/service.rs:572:17
    |
535 |         req: &Request<'_>,
    |                       -- let's call the lifetime of this reference `'1`
536 |         dec: &mut Decoder<'_>,
    |                           -- let's call the lifetime of this reference `'2`
...
572 |                 self.rotate_key(req, dec).await?.to_vec()?
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^ argument requires that `'1` must outlive `'2`
    ```

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
noyez: i rebased on develop, It looks like NodeManager::Identity() no long exists, so this will have to be reworked anyway.

implementations/rust/ockam/ockam_command/src/identity/rotate_key.rs
noyez: @SanjoDeundiak which argument do you mean? 

@etorreborre Oh, ok. I thought that a node always has to be running and that held the identity.

etorreborre:
@SanjoDeundiak what is the impact of the identity v2 work for this PR? Just a minor change to use a new API for identities?

SanjoDeundiak:
@etorreborre should not be a significant change. It's OK to merge this PR when it's ready

PR #4884, opened 2023-05-08
refactor: vault kms

etorreborre:
This PR addresses a number of issues with the current `ockam_vault` design:

 1. the code for using an AWS KMS to create and store secrets is intertwined with the generic vault code and the Ockam implementation. This makes it impossible to add another implementation without even more intrusive modifications
 
 2. the storage for secrets mixes several concerns: storing persistent data on file, caching it, ephemeral vs persistent storage (left as attributes on a secret). Part of the storage code is not vault-specific and should be in `ockam_node`
 
 3. the initial interfaces for a `Vault` are not really used independently and it is not actually possible to just replace one of those implementations (say, for `AsymmetricVault` encryption for example) 

# Changes

Several changes were introduced:

 1. code regrouping
 2. generic storage extraction
 3. redefinition of a secret
 4. extraction of the aws code in a separate crate
 5. redefinition of the `Vault` traits (and the test support)
 6. addition of a builder to create a Vault and select implementations

# Code regrouping

The `vault` traits (`Signer`, `AsymmetricVault` ...) are regrouped in the `ockam_vault` crate since the _only_ thing they share with concepts in the `ockam_core` crate is the notion of a `KeyId` (effectively a `String`), used for the interface of the `KeyExchanger` trait

# Generic storage

Generic support for storing generic data in memory or in a file have been moved to `ockam_core` as traits, with various implementations:

 - `dyn KeyValueStorage<K, V>` represents stored key/values. There are 2 implementations, in-memory and file. The in-memory implementation is used to store ephemeral secrets in the `Vault`
      
 - `dyn ValueStorage<V>` represent a single stored value. There are 2 implementations, in-memory and file. The file implementation is used to store persistent secrets in the `Vault` (the in-memory implementation can be used for tests). The important part about the `FileStorage` is that is transactional and concurrent w.r.t external processes
  
# Redefinition of a secret

## Ephemeral vs Persistent

Related to how data is stored in a `Vault`, a strong distinction has been made between:

 - ephemeral storage: ephemeral secrets can only be stored by an in-memory implementation, so they won't persist after a restart
 - persistent storage: persistent secrets can use an in-memory or a file implementation, and can be retrieved after a restart with the file implementation

Both storage share some common operations but also some major distinctions:

 - the persistent storage can be backed-up by a KMS, like the AWS KMS. A `Kms` trait has been introduced to represent these capabilities
 - an ephemeral can import and export private secrets, a persistent storage can only generate them (via its KMS)

As a result of this interface redefinition for secret storage the `SecretPersistence` attribute has been removed since now this is a property of the code calling the vault to decide if the ephemeral or the persistent API must be used.

## Key vs KeyId

A stored secret used to be either a `KeyId` for AWS or some binary secret. Since a KMS knows the key ids it created there is no need to encode this knowledge in a secret.

## `SecretType`

The other secret attributes were `secret_type` and `length`. It turns out that those 2 values are strongly related so `length` has been removed in most case and `SecretType` is now an enum listing all the possible secret types and in the case of a `Buffer` variant a length can be specified.

## `StoredSecret`

A secret and its metadata becomes a `StoredSecret` = binary secret + secret attributes. `SecretAttributes` is an enum implying both a `SecretType` and a `length`. For example the `Aes256` variant has type `Aes` and length `32`. As a result the code is more typesafe and several redundant length checks were removed.

## Backward compatibility

The new data has been made compatible with the old data by adding backward compatible deserializers for a `Secret` (with the removal of the `Aws` case) and `StoredSecrets` (renaming of the data stored on disk for a Vault). Tests have been added to read the old and new formats. 

#  Extraction of the aws code in a separate crate

Since there is a new `Kms` trait it becomes easy to introduce a new `ockam_vault_aws` crate with dependencies on `aws-*` libraries. That crate uses an `AwsKmsClient` to access an AWS KMS.

Note: most of the configuration for that client is done via environment variables.

# Redefinition of the `Vault` traits (and the test support)

The traits previously defined are rarely used independently. But we would still like to be able to replace some of them with a different implementation. In order to simplify this design the traits have been simplified to 4 traits:

 - `AsymmetricVault` for asymmetric encryption with the creation of Diffie-Hellman keys + hkdf compution since those 2 functions are used together
 - `SymmetricVault` for symmetric encryption
 - `Signer` which includes signing _and_ verifying. This makes the definition of this trait more self-contained instead of having 2 inter-dependent traits
 - `StoredSecrets` is the general trait to create/retrieve secrets and keys. It is further subdivided into
   - `EphemeralSecrets` to manage secrets which are not committed to durable memory
   - `PersistentSecrets` to manage secrets which are committed to durable memory. Those secrets cannot be imported / exported

The `Hasher` trait has been removed since the only remaining function is `sha256` which is directly taken from the Rust library implementation and which needs to stay consistent everywhere it is used.

## Test support

Instead of having separate files for the test code related to each trait, the test code has been moved to the trait file which makes the trait better documented with its own doc + specification as tests.

# Addition of a `builder` to create a `Vault` and select implementations

We might want to change several things for a `Vault`:

 - the low level use of a storage for persistent secrets
 - the use of an entirely different KMS to generate and store them
 - some alternate implementation of the high-level functionality for encrypting messages for example

This is all supported by a `Vault::builder()` and several `with_xxx` methods to select implementations. It must be noted though that since those implementations are on different levels of abstraction it is possible to introduce implementations which won't play well together, for example an implementation of the `AsymmetricVault` trait might not retrieve secrets from the right storage.


implementations/rust/ockam/ockam_command/src/vault/attach_key.rs
```diff
@@ -42,14 +42,13 @@ async fn run_impl(opts: CommandGlobalOpts, cmd: AttachKeyCommand) -> crate::Resu
...
-        let kid = vault.secret_import(Secret::Aws(cmd.key_id), attrs).await?;
-        let attrs = KeyAttributes::new(IdentityChangeConstants::ROOT_LABEL.to_string(), attrs);
+        let attrs = SecretAttributes::NistP256;
+        let key_attrs = KeyAttributes::new(IdentityChangeConstants::ROOT_LABEL.to_string(), attrs);
         opts.state
             .get_identities(vault)
             .await?
             .identities_creation()
-            .create_identity_with_external_key(&kid, attrs)
+            .create_identity_with_external_key(&cmd.key_id, key_attrs)
```
polvorin: this was a bit entangled and too much specific.     Previously it was doing *two* unrelated things:
1)  Attach an _existing_  kms key to a kms vault  to be able to use it from ockam (that was needed so far)
2) Create an identity using _that_  keyid 

These things were not mean to be in a single command, for example there was not way for creating an identity from an existing key on a file vault,  nor to use a kms key for other thing than to create an identity from it.     These should had been  two different commands  (I would swear there was an issue for that but coudln't find it). 

But If I'm understanding well the changes here,   the step 1) here is no longer needed?   you can refer to the existing kms key  directly without "attaching it"  first?    If that's the case I'd  suggest fixing the docs for the `key_id` parameter  (this command should work for any vault type now?). 

Then, the entire `attach_key`  command make no sense anymore :),  `attach_key`  is not attaching a key is creating an identity  (this can be done on a separate PR if you don't want to include further changes here)   so  this should be just an  optional  `key_id`  parameter to the `identity create` command , indicating to use that key instead of generating a random one.

etorreborre: Good catch, you're absolutely right. I am going to fix this

etorreborre: You are right, let's address this in the next PR.

implementations/rust/ockam/ockam_vault_aws/src/vault/aws_kms_client.rs
```diff
@@ -74,113 +81,144 @@ impl Kms {
...
+        log::error!(%key_id, "key type not supported to get a public key");
         Err(Error::UnsupportedKeyType.into())
     }
 
+    pub(crate) async fn compute_key_id(&self, public_key: &PublicKey) -> Result<KeyId> {
+        log::trace!(%public_key, "compute the key id for a public key");
+        let output = self
+            .client
+            .describe_key()
+            .key_id(public_key.to_string())
```
davide-baldo, line 151: I don't understand, isn't `pulic_key` already the key id?

implementations/rust/ockam/ockam_vault/src/traits/kms.rs
```diff
@@ -0,0 +1,39 @@
...
+
+/// This trait provides the main functions of a KMS
+///   - create and persist secrets
+///   - delete secrets
+///   - return the public key for a given key id
+///   - return the key id for a given public key
+///   - use a secret to sign a message
+///   - use a public key to verify a message signature
+#[async_trait]
+pub trait Kms: Sync + Send {
```
davide-baldo: really straightforward interface, love it :+1:

implementations/rust/ockam/ockam_vault/src/vault/vault_builder.rs
```diff
@@ -0,0 +1,119 @@
...
+            asymmetric_vault,
+            symmetric_vault,
+            signer,
+        }
+    }
+
+    /// Set a persistent storage as a file storage with a specific path
+    /// Note: this overrides all previously set implementations
+    #[cfg(feature = "std")]
+    pub async fn with_persistent_storage_path(&mut self, path: &Path) -> Result<&mut Self> {
```
davide-baldo, line 52: maybe these methods should be the starting point instead/along with `create`, so there is no risk of overwriting by mistake

implementations/rust/ockam/ockam_vault_aws/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+#![deny(unsafe_code)]
+#![warn(
+    missing_docs,
+    trivial_casts,
+    trivial_numeric_casts,
+    unused_import_braces,
+    unused_qualifications
+)]
+#![cfg_attr(not(feature = "std"), no_std)]
+
```
davide-baldo, line 12: extra newline might be confusing

implementations/rust/ockam/ockam_vault/src/storage/persistent_storage.rs
```diff
@@ -0,0 +1,288 @@
...
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::collections::BTreeMap;
+use std::path::Path;
+
+/// Storage for a Vault data backed by a file
+/// The `FileValueStorage` implementation takes care of locking / unlocking the underlying file
+/// in the presence of concurrent accesses
+pub struct PersistentStorage {
+    storage: Arc<FileValueStorage<StoredSecrets>>,
+    cache: Arc<dyn KeyValueStorage<KeyId, StoredSecret>>,
```
davide-baldo, line 15: can't we implement a neutral cache taking any `KeyValueStorage` as an argument?

davide-baldo (changes requested):

implementations/rust/ockam/ockam_vault_aws/src/vault/aws_kms_client.rs
etorreborre: The Kms can store secrets that don't have a public key, that's why everything has a key id

implementations/rust/ockam/ockam_vault_aws/src/lib.rs
etorreborre: I think that's ok, the attribute applies to the whole module, not to the next declaration. It is consistent with similar declarations in other `lib` files

implementations/rust/ockam/ockam_vault/src/storage/persistent_storage.rs
etorreborre: I don't see what that would look like with well-typed values? Could you sketch what you have in mind?

implementations/rust/ockam/ockam_vault/src/vault/vault_builder.rs
etorreborre: Right, I made top-level functions `Vault::create_with_xxx` to support the most frequent use cases so that the builder is only used if absolutely needed by other implementors to fine-tweak the setup.

implementations/rust/ockam/ockam_vault/src/storage/persistent_storage.rs
davide-baldo:
```
struct StorageCache<K,V>{
    cache: Arc<dyn KeyValueStorage<K, V>>,
    storage: Arc<dyn KeyValueStorage<K, V>>,
}

impl StorageCache {
  fn create(store_to_cache: Arc<dyn KeyValueStorage<K, V>>) -> Self {
   Self{
     cache: Default::default(),
     store: store_to_cache,
     }
   }
}

impl KeyValueStorage for StorageCache {
   async fn get(&self, key: &K) -> Result<Option<V>> {
     if let Some(x) = self.cache.get(key) {
       return x;
      }
      if let Some(x) = self.proxied.get(key) {
        self.cache.put(key,x);
        return x;
      }
      return None;
  }
  ...
}
```

etorreborre: I _think_ I see what you mean. Are you hinting at the fact that `PersistentStorage::storage` could already have its own cache, as part of its implementation? The problem is that this is a `FileValueStorage` not a `FileKeyValueStorage` for now, so it can not integrate a key / value cache since it has no notion of a key.

implementations/rust/ockam/ockam_vault/src/traits/asymmetric_vault.rs
```diff
@@ -0,0 +1,87 @@
...
+        &self,
+        salt: &KeyId,
+        info: &[u8],
+        ikm: Option<&KeyId>,
+        output_attributes: SmallBuffer<SecretAttributes>,
+    ) -> Result<SmallBuffer<KeyId>>;
+}
+
+/// Tests for implementations of the AsymmetricVault trait
+pub mod tests {
```
SanjoDeundiak, line 26: Can we avoid exposing it? Maybe hide by a feature flag, or move to a separate crate?

etorreborre: These traits were already exposed in the previous version. But this is a good idea, I have added a feature flag.

implementations/rust/ockam/ockam_vault/src/traits/types.rs
```diff
@@ -0,0 +1 @@
+
```
SanjoDeundiak: That's an empty file

implementations/rust/ockam/ockam_vault/src/types/key_pair.rs
```diff
@@ -0,0 +1,29 @@
+use crate::PublicKey;
+use ockam_core::KeyId;
+use zeroize::Zeroize;
+
+/// A public key
```
SanjoDeundiak: A KeyPair

implementations/rust/ockam/ockam_identity/Cargo.toml
```diff
@@ -21,7 +21,9 @@ noise_xx = ["ockam_key_exchange_xx"]
 software_vault = [
   "ockam_vault",
   "ockam_vault/rustcrypto",
+  "rustcrypto",
 ]
+rustcrypto = []
```
SanjoDeundiak: I don't think this feature should exist. It should be always enabled, otherwise we won't be able to connect to the orchestrator

implementations/rust/ockam/ockam_vault/src/types/stored_secret.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// Get the secret part of this stored secret
+    pub fn secret(&self) -> &Secret {
+        &self.secret
+    }
+
+    /// Check if the length of the secret is the same as the length prescribed by the attributes
+    fn check(secret: &Secret, attributes: &SecretAttributes) -> Result<()> {
+        // the secret must be at least the length mentioned in the attributes
+        // In the case of a NistP256 secret it might contain additional metadata (algorithm, version, etc...)
+        if secret.length() < attributes.length() as usize {
```
SanjoDeundiak, line 38: Interesting. Do you have an idea if NIST P256 private key is more than 32 bytes in our implementation? That's weird

implementations/rust/ockam/ockam_vault/src/traits/secrets_store.rs
```diff
@@ -0,0 +1,178 @@
...
+    async fn create_ephemeral_secret(&self, attributes: SecretAttributes) -> Result<KeyId>;
+    /// Import a secret and persist it to ephemeral memory
+    async fn import_ephemeral_secret(
+        &self,
+        secret: Secret,
+        attributes: SecretAttributes,
+    ) -> Result<KeyId>;
+    /// Export an ephemeral secret
+    /// Use the description in an error message if the secret cannot be found
+    async fn get_ephemeral_secret(&self, key_id: &KeyId, description: &str)
```
SanjoDeundiak, line 29: Adding description here just for better error message looks weird to me

implementations/rust/ockam/ockam_vault/src/traits/security_module.rs
```diff
@@ -0,0 +1,43 @@
...
+#[async_trait]
+pub trait SecurityModule: Sync + Send {
+    /// Create a new secret and return its key id
+    async fn create_secret(&self, attributes: SecretAttributes) -> Result<KeyId>;
+
+    /// Get the public key from a secret
+    async fn get_public_key(&self, key_id: &KeyId) -> Result<PublicKey>;
+
+    /// Return the key id corresponding to a given public key
+    async fn get_key_id(&self, public_key: &PublicKey) -> Result<KeyId>;
```
SanjoDeundiak, line 25: `SecretsStoreReader` also has it. Could you please explain why?

implementations/rust/ockam/ockam_vault/src/traits/types.rs
etorreborre: thanks, fixed

implementations/rust/ockam/ockam_vault/src/types/key_pair.rs
etorreborre: fixed

implementations/rust/ockam/ockam_identity/Cargo.toml
etorreborre: Indeed, I removed it then

implementations/rust/ockam/ockam_vault/src/traits/security_module.rs
etorreborre: That's because `SecretsStoreReader` is the high-level interface while the `SecurityModule` is part of the implementation, not exposed to the users of the `Vault`. And one delegates to the other.

implementations/rust/ockam/ockam_vault/src/traits/secrets_store.rs
etorreborre: I did this because I found it quite helpful to get a specific error message rather than "key not found". But if you feel it's too much, I can revert this.

implementations/rust/ockam/ockam_vault/src/types/stored_secret.rs
etorreborre: That's what I observed from testing: `Actual: 138, Expected: 32`

davide-baldo (approved):

PR #4897, opened 2023-05-10
Compact secure channel initialization with 3 packet exchange

davide-baldo:
In this PR the secure channel is further simplified and compacted by reducing the initial exchange to 3 packets.

Rust side:

- decryptor/encryptor were refactored into initiator and responder
- initiator callback is now using a dedicated object

Shortcomings:
- cannot replace initiator worker with decryptor worker, and must act as a relay
- ~~multiple credentials are accepted but only last is stored~~
- ~~`Finalizer` should be renamed and better integrated, suggestions are appreciated!~~
- ~~credentials are not exchanged using secure channel initialization by default (yet)~~

Rust follow-ups:
- Remove the present credential and relative service in a single revertible commit (also simplify credential presentation from yes|no|mutual to yes|no)
- Allow to store multiple credentials for a single identity


implementations/rust/ockam/ockam_identity/src/secure_channel/packets.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Message)]
+pub struct ThirdPacket {
+    //noise_xx: -> s, se
+    pub(super) key_exchange_with_payload: KeyExchangeWithPayload<IdentityAndCredential>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Message)]
+pub(super) struct IdentityAndCredential {
```
polvorin, line 29: @davide-baldo  I'm a bit lost on this one,   how is the on-the-wire representation of this struct?   it's using cbor?

etorreborre: @polvorin I think that this is using `bare` but since we are changing the protocols we might as well bite the bullet and use `cbor` now

polvorin:
@davide-baldo  I think we can move this to the ready for review state now

polvorin (approved):

polvorin (approved):

PR #4901, opened 2023-05-11
Support identity updates in services

etorreborre:
## Current behavior

Several services are referencing an `Identity` field which would need to be updated if there was an independent key rotation on that identity.

## Proposed changes

This PR replaces the reference to an `Identity` with an `IdentityIdentifier` so that the most up to date `Identity` for that identifier can be retrieved when needed. The retrieval is either done via the `IdentitiesRepository` interface or simply via the `IdentitiesReader` when a given module just requires retrieving identities.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/kafka/integration_test.rs
+++ b/implementations/rust/ockam/ockam_api/src/kafka/integration_test.rs
@@ -112,7 +112,7 @@ mod test {
                 .secure_channels
                 .create_secure_channel_listener(
                     context,
-                    &handle.identity.identifier(),
+                    &handle.identifier,
                     DefaultAddress::SECURE_CHANNEL_LISTENER,
                     SecureChannelListenerOptions::new().as_consumer_with_flow_control_id(
                         flow_controls,
--- a/implementations/rust/ockam/ockam_api/src/kafka/integration_test.rs
+++ b/implementations/rust/ockam/ockam_api/src/kafka/integration_test.rs
@@ -1,14 +1,8 @@
 #[cfg(test)]
 mod test {
-    use crate::hop::Hop;
-    use crate::kafka::protocol_aware::utils::{encode_request, encode_response};
-    use crate::kafka::secure_channel_map::ForwarderCreator;
-    use crate::kafka::{
-        KafkaInletController, KafkaPortalListener, KafkaSecureChannelControllerImpl,
-    };
-    use crate::nodes::registry::KafkaServiceKind;
-    use crate::test::NodeManagerHandle;
-    use crate::DefaultAddress;
+    use std::sync::atomic::{AtomicBool, Ordering};
+    use std::time::Duration;
+
     use bytes::{Buf, BufMut, BytesMut};
     use indexmap::IndexMap;
     use kafka_protocol::messages::produce_request::{PartitionProduceData, TopicProduceData};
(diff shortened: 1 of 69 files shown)


implementations/rust/ockam/ockam_api/src/kafka/integration_test.rs
```diff
@@ -1,14 +1,8 @@
...
-    use crate::hop::Hop;
-    use crate::kafka::protocol_aware::utils::{encode_request, encode_response};
-    use crate::kafka::secure_channel_map::ForwarderCreator;
-    use crate::kafka::{
-        KafkaInletController, KafkaPortalListener, KafkaSecureChannelControllerImpl,
-    };
-    use crate::nodes::registry::KafkaServiceKind;
-    use crate::test::NodeManagerHandle;
-    use crate::DefaultAddress;
+    use std::sync::atomic::{AtomicBool, Ordering};
```
SanjoDeundiak, line 3: Was there a reason to change the order of imports? For example, there is 1 line changed in this, but because of imports there are 39 lines changed.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -104,8 +106,14 @@ pub struct NodeManager {
 }
 
 impl NodeManager {
-    pub(super) fn identity(&self) -> Identity {
-        self.identity.clone()
+    pub(super) async fn identity(&self) -> Result<Identity> {
```
SanjoDeundiak: Should this function exist? That allows us to forget to fetch the latest Identity instance or e.g. rotate key using this cloned struct which will have no effect on the original Identity

SanjoDeundiak: This should be well documented at least

SanjoDeundiak (dismissed):
Thank you! Any changes to the website documentation required?

implementations/rust/ockam/ockam_api/src/kafka/integration_test.rs
etorreborre: I didn't see that. This is CLion deciding to sort imports alphabetically apparently but I can't find where this setting is coming from

implementations/rust/ockam/ockam_api/src/nodes/service.rs
etorreborre: Correct, it is not needed, I removed it.

implementations/rust/ockam/ockam_api/src/kafka/integration_test.rs
SanjoDeundiak: I think sorting imports is fine and is enforced by `cargo fmt`. The problem I think is that previously imports didn't have any blank lines between them, which resulted in a totally different order

SanjoDeundiak (approved):

implementations/rust/ockam/ockam_api/src/kafka/integration_test.rs
shanesveller: I'm not sure about Clion, but these are the settings I know for rustfmt/cargo fmt and rust-analyzer. IME the newlines between logical groupings are significant to both tools, like Sanjo said.

## Rustfmt

These were nightly-only when I started using them but may have stabilized:
https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#group_imports
https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#imports_granularity

```toml
group_imports = "StdExternalCrate" # other options: Preserve, One
imports_granularity = "Crate" # other options: Preserve, Module, Item, One 
```

## Rust Analyzer

https://rust-analyzer.github.io/manual.html#configuration
https://rust-analyzer.github.io/manual.html#auto-import

```json
"rust-analyzer.imports.granularity.group": "crate",
"rust-analyzer.imports.group.enable": true,
"rust-analyzer.imports.prefix": "self"
```
The second link describes other possible values.

etorreborre:
> Thank you! Any changes to the website documentation required?

Yes since I updated the examples. I'll prepare a PR

etorreborre:
@SanjoDeundiak documentation PR: https://github.com/build-trust/ockam-documentation/pull/20

PR #4906, opened 2023-05-12
chore(rust): adding extra logging in echo worker

davide-baldo:
Adding extra logging in echo worker to make the flow of the message clearer


etorreborre (approved):
🍏

PR #4908, opened 2023-05-12
feat(elixir): refactor secure channel

polvorin:
* Unify secure channel with identity secure channels  (-> all secure channels are "identity" secure channels)
* Use ockam worker (`AsymetricWorker` to get two different addresses) for implementation
* Use a single state machine,  prepare to adapt to changes on https://github.com/build-trust/ockam/pull/4897
* Add couple missing tests  (secure channels peers get terminated if peer gracefully terminate it,   tests around metric collection, etc)
* Move particularities of our implementation  (bare-encoding of binary messages,  routes, etc) out of the XX handshake module and into the secure channel itself
```
               ┌─────────────────┐                          ┌─────────────────┐
               │                 │                          │                 │
           ┌───┴──────┐    ┌─────┴────┐                 ┌───┴──────┐    ┌─────┴────┐
Plaintext  │ Address  │    │Inner     │                 │Inner     │    │ Address  │  Plaintext
  ◄──────► │          │    │Address   │ ◄────[...] ───► │Address   │    │          │ ◄──────►
           └───┬──────┘    └─────┬────┘                 └───┬──────┘    └─────┬────┘
               │  SecureChannel  │       Ciphertext         │   SecureChannel │
               └─────────────────┘                          └─────────────────┘
```


implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,611 @@
...
+          | {:identity_module, module()}
+          | {:encryption_options, encryption_options()}
+          | {:address, Ockam.Address.t()}
+          | {:trust_policies, list()}
+          | {:authorization, list() | map()}
+          | {:additional_metadata, map()}
+          | {:idle_timeout, non_neg_integer() | :infinity}
+
+  # Handshaking | IdentityExchange | Established are the different states in which
+  # the connection can be.   We don't use a proper fsm here as not directly supported
```
shanesveller: This is great context - might be good for inclusion in the moduledoc?

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,611 @@
...
 
-  @doc false
-  def peer(_channel), do: :ok
+  require Logger
 
-  def established?(channel) do
-    channel |> Node.whereis() |> GenStateMachine.call(:established?)
+  @type encryption_options :: [{:vault, Ockam.Vault.t()}, {:static_keypair, reference()}]
+  @type authorization :: list() | map()
+  @type trust_policies :: list()
```
shanesveller, line 50: Are we able to be more specific about the inner type, such as `list(binary())` or whatever would be correct? Or is this already the appropriate precision?

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,611 @@
...
+    field(:inner_address, Ockam.Address.t())
+    field(:vault_name, binary())
+    field(:peer_route, Ockam.Address.route())
+    field(:peer_identity_route, Ockam.Address.route())
+    field(:trust_policies, trust_policies())
+    field(:additional_metadata, map())
+    field(:channel_state, Handshaking.t() | IdentityExchange.t() | Established.t())
+  end
+
+  @handshake_timeout 20_000
```
shanesveller: Deliberate that this differs from the outer timeout on the `defdelegate`?

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,611 @@
...
+  @spec listener_child_spec([listener_opt()]) :: map()
+  def listener_child_spec(args) do
+    %{
+      id: __MODULE__,
+      start: {Spawner, :start_link, [spawner_options(args)]}
+    }
+  end
+
+  @spec start_link_channel([initiator_opt()], non_neg_integer) :: :ok | {:error, term()}
+  def start_link_channel(opts, handshake_timeout \\ 3000) do
```
shanesveller: 👀 Third unique value for a timeout I've seen, probably correct but just making sure.

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,611 @@
...
-
-    case Node.start_supervised(__MODULE__, options) do
-      {:ok, pid, address} ->
-        with {:ok, _pid, address} <- wait_for_channel(pid, address, timeout) do
-          {:ok, address}
-        end
-
-      error ->
-        error
+  # Deprecated, use start_link_channel/2
```
shanesveller: Optional, as it produces compilation warnings, but good for consumers' feedback IMO:
```suggestion
  @deprecated "use start_link_channel/2"
```

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,611 @@
...
+    Ockam.Worker.call(worker, :role)
   end
 
-  @doc false
-  def start_link(options) when is_list(options) do
-    address = Keyword.get(options, :address, Node.get_random_unregistered_address())
-    options = Keyword.put(options, :address, address)
+  def established?(worker) do
+    Ockam.Worker.call(worker, :established?)
+  end
```
shanesveller, line 204: Optional, if we're not going to supply docstrings for some of these, and they're just one-to-one wrappers of `call`, I'd consider metaprogramming them for brevity:

```
for msg <- ~w(disconnect established? get_remote_identity get_remote_identity_id role)a do # and so on
  def unquote(msg)(worker), do: Ockam.Worker.call(worker, unquote(msg))
end

File: implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex [hunk 7]
```diff
@@ -1,380 +1,611 @@
...
-      other -> other
-    end
-  end
+  defp worker_return({:stop, reason, channel_state}, worker_state),
+    do: {:stop, reason, Map.put(worker_state, :state, channel_state)}
+
+  defp new_xx(vault, opts) do
+    m1_payload = m2_payload = m3_payload = ""
+    {:ok, e_key_handle} = Ockam.Vault.secret_generate(vault, type: :curve25519)
+    {:ok, e_public_key} = Vault.secret_publickey_get(vault, e_key_handle)
```
shanesveller: Nit, but I'd like to see both calls either aliased or fully-qualified as appropriate.

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,611 @@
...
+    end
+  end
 
-    cond do
-      first_address === data.ciphertext_address ->
-        case decrypt_and_send_to_router(message, state, data) do
-          {:next_state, state, data} ->
-            {:next_state, state, data}
+  defp identity_from_opts(options) do
+    identity_module = Keyword.get(options, :identity_module, Identity.default_implementation())
```
shanesveller:
```suggestion
    identity_module = Keyword.get_lazy(options, :identity_module, &Identity.default_implementation/0)
```

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,611 @@
...
+            state
+            | channel_state: %{channel_state | decrypt_st: decrypt_st}
+          })
+
+        {:error, reason} ->
+          {:error, reason}
+      end
+    else
+      # The message couldn't be decrypted.  State remains unchanged
+      _error -> {:ok, state}
```
shanesveller: Should this be logged, metered with telemetry, or anything similar to give us some insight later?

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,611 @@
...
-  defp setup_key_establishment_protocol(options, data) do
-    case Keyword.get(options, :key_establishment_protocol, XXKeyEstablishmentProtocol) do
-      XXKeyEstablishmentProtocol ->
-        data = Map.put(data, :key_establishment_protocol, XXKeyEstablishmentProtocol)
-        data.key_establishment_protocol.setup(options, data)
+  defp spawner_options(opts) do
+    worker_opts =
+      opts
+      |> Keyword.drop([:address, :authorization])
+      |> Enum.map(fn
```
shanesveller: Nit: Almost zero practical difference in the outcome but slightly more intention-revealing IMO. The major distinction is it de-duplicates on purpose, which I'm not sure is benign/good/bad for us here.
```suggestion
      |> Keyword.new(fn
```

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,611 @@
...
+        {:responder_authorization, auth} -> {:authorization, auth}
+        other -> other
+      end)
+      |> Keyword.merge(address_prefix: "ISC_R_", role: :responder, restart_type: :temporary)
 
-      unexpected_protocol ->
-        {:error, {:unexpected_key_establishment_protocol, unexpected_protocol}}
-    end
+    Keyword.merge(Keyword.take(opts, [:address, :authorization]),
```
shanesveller: This is nicely clear as-written, but you could also begin the function with `{opts, worker_opts} = Keyword.split(opts, [:address, authorization])` and go from there.

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/key_establishment_protocol/xx/protocol.ex
```diff
@@ -119,55 +145,22 @@ defmodule Ockam.SecureChannel.KeyEstablishmentProtocol.XX.Protocol do
...
-  end
-
-  def decode(message_name, message_payload, %{xx_key_establishment_state: protocol_state} = data)
-      when message_name in [:message1, :message2, :message3] do
-    ## TODO: optimise double encoding of binaries
-    {:ok, message, ""} = :bare.decode(message_payload, :data)
-    decoder = String.to_existing_atom("decode_" <> Atom.to_string(message_name))
-    return_value = apply(__MODULE__, decoder, [message, protocol_state])
+  def encode(:message1, %{e: e, payloads: payloads} = state) do
+    payload = Map.get(payloads, :message1, "")
```
shanesveller, line 153: Will we eventually want to remove the default value here after the three-message work, or is always reasonable to have a fallback to empty string?

implementations/elixir/ockam/ockam/lib/ockam/worker/authorization.ex
```diff
@@ -119,7 +119,7 @@ defmodule Ockam.Worker.Authorization do
   to be handled by the worker.
   """
   def from_secure_channel(prev \\ :ok, message, _state) do
-    Logger.debug("check is secure #{inspect(message)}")
+    Logger.debug("check from secure channel #{inspect(message)}")
```
shanesveller, line 136: I realize this is no worse than the existing impl, but can we be any more conservative about limiting the inspect output? Do messages have a useful Inspect protocol definition?

implementations/elixir/ockam/ockam/mix.exs
```diff
@@ -14,7 +14,7 @@ defmodule Ockam.MixProject do
       version: @version,
       elixir: @elixir_requirement,
       consolidate_protocols: Mix.env() != :test,
-      elixirc_options: [warnings_as_errors: true],
+      elixirc_options: [warnings_as_errors: false],
```
shanesveller: Intentional to include? Do we have any tracking issues to be able to restore it in the future?

implementations/elixir/ockam/ockam/test/ockam/secure_channel_test.exs
```diff
@@ -265,18 +245,255 @@ defmodule Ockam.SecureChannel.Tests do
...
+    {:ok, vault} = SoftwareVault.init()
+    {:ok, alice, alice_id} = Identity.create(@identity_impl)
+
+    {:ok, listener} =
+      SecureChannel.create_listener(
+        identity: alice,
+        encryption_options: [vault: vault]
+      )
+
+    _listener_pid = Ockam.Node.whereis(listener)
```
shanesveller: Do we just want to establish this is a living process and not nil? Can this line be dropped?

implementations/elixir/ockam/ockam/test/ockam/secure_channel_test.exs
```diff
@@ -265,18 +245,255 @@ defmodule Ockam.SecureChannel.Tests do
...
+
+    assert id == alice_id
+
+    SecureChannel.disconnect(channel)
+    assert_receive {:DOWN, ^ref1, _, _, _}
+    assert_receive {:DOWN, ^ref2, _, _, _}
+  end
+
+  test "identity channel inner address is protected" do
+    ## Inner address is the one pointint to the other peer.
```
shanesveller:
```suggestion
    ## Inner address is the one pointing to the other peer.
```

implementations/elixir/ockam/ockam/test/ockam/secure_channel_test.exs
```diff
@@ -265,18 +245,255 @@ defmodule Ockam.SecureChannel.Tests do
                    10_000
   end
 
+  test "local secure channel" do
+    {:ok, vault} = SoftwareVault.init()
+    {:ok, alice, alice_id} = Identity.create(@identity_impl)
```
shanesveller: Is it worthwhile to hoist some of these up to setup functions if we're always scrutinizing a pair of Alice/Bob identities in each test body?

implementations/elixir/ockam/ockam_healthcheck/lib/healthcheck/healthcheck.ex
```diff
@@ -136,6 +138,7 @@ defmodule Ockam.Healthcheck do
          {:worker_init, _worker,
           {:handler_error, reason, _message,
            {Ockam.Transport.TCP, :handle_transport_message, [[]]}}}} ->
+          # How is this defined?
```
shanesveller, line 141: 👀

implementations/elixir/ockam/ockam_metrics/test/ockam_metrics/telemetry_poller_test.exs
```diff
@@ -0,0 +1,111 @@
...
+  doctest Ockam.Metrics.TelemetryPoller
+
+  alias Ockam.Identity
+  alias Ockam.Metrics.TelemetryPoller
+  alias Ockam.SecureChannel
+  alias Ockam.Vault.Software, as: SoftwareVault
+
+  setup do
+    Ockam.Node.register_address("test")
+    on_exit(fn -> Ockam.Node.unregister_address("test") end)
```
shanesveller: If we use a fixed node name that can collide, should this module be `async: false`?

implementations/elixir/ockam/ockam_metrics/test/test_helper.exs
```diff
@@ -1 +1,3 @@
-ExUnit.start()
+Application.ensure_all_started(:ockam)
+Application.ensure_all_started(:telemetry)
+ExUnit.start(capture_log: true, trace: true)
```
shanesveller, line 3: The scope of impact is limited, but do we want to permanently enable `trace` rather than using the opt-in CLI flag?

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
polvorin: :+1:   yes we should be able to be more specific,  will take a look at it.    I let at `list` and `map` when doing the refactor as as wasn't sure what exactly were the valid values for trust_policies or authorizations

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
polvorin: no.    will fix!

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
polvorin: :grimacing:   and this one is wrong

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/key_establishment_protocol/xx/protocol.ex
polvorin: good question.   Even after the upcoming changes,  the _first_ handshake message won't carry any payload in our protocol.  So this could had even be hardwired to be the an empty binary. 
_But_   there are existing test vectors we use for unit tests of the noise handshake that _do_ have payloads..  so that's right now the only value of keeping it open (that we can validate existing test vectors for noise,  not only ones explicitly written by us)

implementations/elixir/ockam/ockam/mix.exs
polvorin: not intentional,  was a leftover :+1:

implementations/elixir/ockam/ockam/test/ockam/secure_channel_test.exs
polvorin: it's completely useless :+1:     (was a leftover,   wanted to monitor the process and check that it terminated when the connecting side terminates,  but that is done on the spawned receiver,  not this process that is the listener itself)

implementations/elixir/ockam/ockam_metrics/test/ockam_metrics/telemetry_poller_test.exs
polvorin: good question :thinking:

implementations/elixir/ockam/ockam_metrics/test/test_helper.exs
polvorin: I don't know what it means :grimacing: ,  copied it to match how it was done on the other libs that already had unit tests defined

shanesveller: Here's the docs from the CLI flag equivalent:

```
  • --trace - runs tests with detailed reporting. Automatically sets
    --max-cases to 1. Note that in trace mode test timeouts will be ignored as
    timeout is set to :infinity
```

It disables any possible parallelism in the test suite as a whole. The order of modules and of test blocks within a given module is non-deterministic without `--seed`, but `mix test --trace` will only ever run a single test module at a time, in some shuffled order of test blocks.

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
polvorin: Yes letting it as a comment for now.  There is code using this and the behavior is slightly different (I copied the deprecated comment from the current version)

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
polvorin: added it to logs

implementations/elixir/ockam/ockam_metrics/test/test_helper.exs
polvorin: created https://github.com/build-trust/ockam/issues/4923  to tackle separately,  whatever we do should be the same on all the libs

implementations/elixir/ockam/ockam/test/ockam/secure_channel_test.exs
polvorin: done.   There is still two "styles" of tests on this file,   one that was here originally and the other for tests moved from the identity secure channel suite.
That can be addressed separately

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
polvorin: will pass on that :),   having type specs and docstrings will be helpful, we should start adding those at least on this kind of more user-facing places.

implementations/elixir/ockam/ockam/lib/ockam/worker/authorization.ex
polvorin: some explicit textual representation would be nice yes.  
So far we are logging the message this way in lots of places,  it's still "not-to-bad"  cause the things most likely to be of interest  (from/to  route,  and metadata)  are short enough and have okish text representation already.
Message' body is truncated

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,616 @@
...
+                |                 |                          |                 |
+            +---+------+    +-----+----+                 +---+------+    +-----+----+
+  Plaintext | Address  |    |Inner     |                 |Inner     |    | Address  |  Plaintext
+  <------>  |          |    |Address   | <----[...] ---> |Address   |    |          | <------>
+            +---+------+    +-----+----+    Ciphertext   +---+------+    +-----+----+
+                |  SecureChannel  |                          |   SecureChannel |
+                +-----------------+                          +-----------------+
+
+
+  The secure channel goes throw three stages:
```
hairyhum: throw -> through

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,616 @@
...
+  """
+  @spec listener_child_spec([listener_opt()]) :: map()
+  def listener_child_spec(args) do
+    %{
+      id: __MODULE__,
+      start: {Spawner, :start_link, [spawner_options(args)]}
+    }
+  end
+
+  @spec start_link_channel([initiator_opt()], non_neg_integer) :: :ok | {:error, term()}
```
hairyhum: Can we have `start_link_channel` function return `{:ok, pid}` or `{:ok, pid, info}` for consistency?

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,616 @@
...
+    do: {Encryptor.new(vault, k2, 0), Decryptor.new(vault, k1, 0)}
 
-      {:next_state, state,
-       %{data | encrypted_transport: %{encrypted_transport | encrypt: new_transport_state}}}
+  defp split(vault, k1, k2, :responder),
+    do: {Encryptor.new(vault, k1, 0), Decryptor.new(vault, k2, 0)}
+
+  # Check result of the handshake step, send handshake data to the peer if there is a message to exchange,
+  # and possible move to another state
+  defp continue_handshake({:complete, {k1, k2, h, _p}}, state) do
```
hairyhum: Code in this function clause is identical to `defp next_handshake_state({:complete, {k1, k2, h, _p}}, state) do` is there some reason not to reuse it?

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,616 @@
...
-    end
+  defp attach_metadata(msg, additional, %Established{peer_identity: i, peer_identity_id: id}) do
+    Message.with_local_metadata(msg, Map.merge(additional, %{identity: i, identity_id: id}))
   end
 
-  defp setup_callback_route(options, data) do
-    case Keyword.get(options, :callback_route) do
-      nil ->
-        {:ok, data}
+  defp handle_outer_message_impl(message, %Channel{channel_state: %Established{} = e} = state) do
```
hairyhum, line 568: Does that crash if it's not in the `Established` state?

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,616 @@
...
+    additional_metadata = Keyword.get(options, :additional_metadata, %{})
+    encryption_options = Keyword.get(options, :encryption_options, [])
+    key_exchange_timeout = Keyword.get(options, :key_exchange_timeout, @handshake_timeout)
+    vault_name = Keyword.get(options, :vault_name)
+
+    with {:ok, role} <- Keyword.fetch(options, :role),
+         {:ok, vault} <- vault_from_opts(encryption_options),
+         {:ok, identity} <- identity_from_opts(options),
+         {:ok, key_exchange_state} <-
+           new_xx(vault, Keyword.take(encryption_options, [:static_keypair])) do
```
hairyhum: I know the module is called `XX`, but can we use more descriptive function name here?

implementations/elixir/ockam/ockam_healthcheck/lib/healthcheck/healthcheck.ex
hairyhum: This pattern match was taken from the crash report basically.

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -1,380 +1,616 @@
...
+                |  SecureChannel  |                          |   SecureChannel |
+                +-----------------+                          +-----------------+
+
+
+  The secure channel goes throw three stages:
+    * Handshaking  (noise handshake)
+    * IdentityExchange (identity exchange and verification)
+    * Established (channel fully established and peer authenticated)
+
+  At this time, the implementation don't use a proper fsm as that's not directly supported
```
hairyhum, line 23: Was the session handshake too complex to accomodate the handshake stage?

hairyhum (commented):
Looks good. A few questions around initialization and error handling.

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
polvorin: yes.   I'll change it to discard with a loudly error,   if it do receive a message before handshake is complete is very likely an error.   (at that point,   neither   `create_channel`  nor `start_link_channel`  would had returned back to the caller,  nor responder would have contacted his address anywhere).

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
polvorin: yes,   I wanted it to be a drop-in replacement for `Ockam.Identity.SecureChannel` without changing its API,   but looking around there is no use of `start_link_channel`  yet :man_shrugging: ,  so ok to change

shanesveller (approved):
LGTM!

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
polvorin: yes found the session  code (pluggable/etc)  difficult to follow,   mainly what state was passed around,   and what were processes vs what are just callback.    Plus (right now)  handshake spand across several messages,  found it simpler to keep in a single fsm  (kind of like on the previous Channel implementation,  but not like in the Identity channel one)

hairyhum (approved):

shanesveller (approved):

PR #4909, opened 2023-05-12
Remove superfluous flow control

davide-baldo:
- hide the inlet address from the other side in kafka workers
- removed superfluous flow control

Draft because i want to do some extra validation on the `MultiAddr` instantiation

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -446,19 +446,31 @@ impl NodeManager {
         debug!("connecting to {}", &connection.addr);
         let context = Arc::new(connection.ctx.async_try_clone().await?);
 
-        let mut intermediary_services = vec![];
-        for protocol_value in connection.addr.iter() {
-            if protocol_value.code() == Service::CODE {
-                let local = protocol_value
-                    .cast::<Service>()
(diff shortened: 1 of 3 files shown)


SanjoDeundiak (commented):
Love the direction of this PR!

davide-baldo:
double-checked!

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -491,9 +503,9 @@ impl NodeManager {
 
         debug!("connected to {connection_instance:?}");
 
-        // Every piece of the chain must be part of the session to allow communication
-        for service in intermediary_services {
-            connection_instance.add_consumer(&service);
+        // make sure the service before flow control is a consumer
+        if let Some(address) = consumer_worker {
```
SanjoDeundiak, line 507: Are there any use cases for that? I had an impression that we can just get rid of it

davide-baldo: In practice, I think It wouldn't break anything (other than `hop`), but I reckon that if the user provides an extra address in the `MultiAddr` we should do anything possible to make sure it's supported out of the box.

etorreborre (approved):

PR #4913, opened 2023-05-12
bump nif version

metaclips:
This PR bumps NIF version to 0.87.0 and also reverts it to the build-trust repository


polvorin (approved):
:+1:

PR #4917, opened 2023-05-15
update readme during release

metaclips:
Update README.md file during release

### DIFF
--- a/tools/docker/builder/Dockerfile
+++ b/tools/docker/builder/Dockerfile
@@ -9,6 +9,7 @@ ARG COSIGN_VERSION=1.9.0
 ARG MOLD_VERSION=1.3.1
 ARG SHFMT_VERSION=3.5.1
 ARG COMMITLINT_VERSION=17.1.1
+ARG GH_CLI_VERSION=2.29.0
 
 ARG ELIXIR_SHA256=fb7a7f1ad171f87c2670e4759334daab18e4440a23f0d85802397b7e9d8b2b2f
 
@@ -16,6 +17,7 @@ ENV RUSTUP_HOME=/opt/rust/rustup \
     CARGO_HOME=/usr/rust/cargo \
     JAVA_HOME=/opt/java/openjdk \
     NODEJS_HOME=/opt/nodejs \
+    GH_CLI_HOME=/opt/gh \
     MOLD_HOME=/opt/mold
 
 RUN set -xe; \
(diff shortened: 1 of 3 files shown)


etorreborre:
@metaclips can you integrate the `gradle update_readmes` task as a release step instead?

metaclips:
> @metaclips can you integrate the `gradle update_readmes` task as a release step instead?

Ok. I'll do that

tools/docker/builder/Dockerfile
```diff
@@ -1,21 +1,23 @@
 FROM gcc:11.2.0@sha256:0eaeaee298c154ecd383cabe5301a5df9fde0cd3a729124caf574500b936389b
 
 ARG RUSTUP_INIT_VERSION=1.25.2
-ARG RUST_VERSION=1.68.0
+ARG RUST_VERSION=1.69.0
```
shanesveller: Can you please leave this out if possible, or else mention it more prominently in the PR title if we need to include it? This is kind of off-topic for the concern being tackled here.

From the project-level perspective, ideally we should be using a [Rustup toolchain file](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) in the repo root as a source of truth, and flowing to other areas from that. I believe this Dockerfile would already properly pick that up with minor changes.

tools/docker/builder/Dockerfile
```diff
@@ -135,6 +139,15 @@ RUN set -xe; \
     echo "${JQ_SHA256}  jq-${JQ_OS}" | sha256sum -c; \
     mv jq-${JQ_OS} /usr/bin/jq; \
     chmod +x /usr/bin/jq; \
+# Install for release
+    cargo install --locked git-cliff --version 0.7.0; \
+    cargo install --locked tomlq --version 0.1.0; \
+    cargo install --locked cargo-release --version 0.21.0; \
```
shanesveller, line 145: Thank you for pinning the versions 👍🏻 

 I'm familiar with git-cliff already. Is `tomlq` a runtime dependency of `cargo-readme` in some way? Or already included in other release steps?

metaclips: tomlq is used in other release steps, we use it to get version of crates during release.

tools/docker/builder/Dockerfile
metaclips: > This is kind of off-topic for the concern being tackled here.

Removed

etorreborre (approved):

PR #4925, opened 2023-05-16
chore(rust): only execute kafka stream test manually

davide-baldo:


etorreborre (approved):
Is it because it is flaky for now?

hairyhum:
> Is it because it is flaky for now?

@etorreborre mostly because it's deprecated in favour of kafka interceptor approach.

PR #4926, opened 2023-05-16
feat(node): poc of a node/typescript integration

glenngillen:
## Proposed changes

A proof-of-concept of an integration that works with NodeJS and Typescript. It currently downloads and installs the Ockam Command on-demand using an `install` script which (should) be executed as part of `npm install` once this is packaged and published to the npm registry. All methods are ultimately running a child process with Command rather than integrating with the Rust bindings directly in any way.


mrinalwadhwa:
This is awesome,

Backed up your branch here: https://github.com/build-trust/ockam/tree/backup/glenngillen/node-spike

I refactored this PR a little:

1. to be a package in `implementations/typescript`, so we can benefit from shared infra and publish this as `@ockam/command` on npm registry.
2. Remove the `child_process` package, which seems [unmaintained](https://www.npmjs.com/package/child_process) and held as a place holder by the npm security team. Replaced it with node's `child_process`.
3. Added some infra for testing with jest and formatting with prettier.

Instructions to try with pnpm:
https://github.com/build-trust/ockam/blob/975f0f3e6323d4681e5355b04100d8b354905c50/implementations/typescript/DEVELOP.md

I think we can land what's in this branch and then add the other commands in your branch in a follow up.

mrinalwadhwa (approved):

mrinalwadhwa:
Merged, will land the other commands from https://github.com/build-trust/ockam/tree/backup/glenngillen/node-spike in a follow up PR.

PR #4937, opened 2023-05-16
feat: support using modules to specify services, remove service config sources

hairyhum:
## Current behavior

Currently ockam services have to be configured via `Ockam.Services.Provider` module.

Often times it creates unnecessary complexity while adding nothing

## Proposed changes

Made it possible to start ockam services using `{module, args}` API if the module implements `child_spec/1` function.

Removed unused services config parsers as they're not providing enough flexibility to be useful.

## Checks


hairyhum:
Example impact could be seen in implementations/elixir/ockam/ockam_services/test/services/app_test.exs between `"start/stop service by name"` and `"start/stop service by module"`

implementations/elixir/ockam/ockam_services/lib/services/provider.ex
```diff
@@ -134,24 +170,12 @@ defmodule Ockam.Services.Provider do
 
   @spec get_service_providers_map(nil | list()) :: list()
```
mszpakowski: Not part of your changes, but this spec looks funky here

implementations/elixir/ockam/ockam_services/lib/services/provider.ex
```diff
@@ -60,6 +60,42 @@ defmodule Ockam.Services.Provider do
...
+          false ->
+            {:error, {:unknown_service, service_name}}
+        end
+    end
+  end
+
+  defp is_service_module?(service_name) do
+    case Code.ensure_loaded(service_name) do
+      {:module, _module} ->
+        function_exported?(service_name, :child_spec, 1)
```
mszpakowski, line 81: What would the consequences be if we provided a module that defines a `child_spec/1` but is not a "service"?

implementations/elixir/ockam/ockam_cloud_node/config/runtime.exs
```diff
@@ -110,13 +94,11 @@ config :ockam, identity_module: identity_module
...
-
-services_list = System.get_env("SERVICES_LIST")
+services_list = System.get_env("SERVICES_LIST", "")
 
-services_file = System.get_env("SERVICES_FILE")
-
-services_config_source = System.get_env("SERVICES_CONFIG_SOURCE")
+services =
+  String.split(services_list, ",", trim: true)
+  |> Enum.map(fn name -> String.trim(name) |> String.to_atom() end)
```
mszpakowski, line 101: This is somewhat of a change in behavior as the application will no longer start any services by default, and in order to start services on startup a specifically formatted services list has to be provided as an env variable.

We don't really have any readme where we could make note about this, but I just wanted to mention that this change in behavior and the way to set up services for `ockam_cloud_node` is currently not very visible.

mszpakowski (dismissed):

implementations/elixir/ockam/ockam_services/lib/services/provider.ex
hairyhum: Then it's a service. There is no clear definition of what it means to be a service, there was an assumption before that it would have some special properties, but as it turns out we would be better off with just supervised processes.
This PR is a first step towards this direction. Maybe we could go all the way and delete the services all together, but I would rather do that in stages.

implementations/elixir/ockam/ockam_cloud_node/config/runtime.exs
hairyhum: Yes. This is a breaking change on purpose. We don't currently document usage of `ockam_cloud_node` currently and it's only used in some internal components.

hairyhum:
This PR introduces a breaking change to configuration of `ockam_cloud_node` app.

It's no longer handles `SERVICES_CONFIG_SOURCE`, `SERVICES_JSON` or `SERVICES_FILE` env variables.
It's no longer starts a default list of services, all started services should be specified in the `SERVICES_LIST` env variable.

implementations/elixir/ockam/ockam_services/lib/services/provider.ex
```diff
@@ -60,6 +60,42 @@ defmodule Ockam.Services.Provider do
...
+        false
+    end
+  end
+
+  defp module_child_spec(service_name, service_args) do
+    id =
+      case Keyword.fetch(service_args, :id) do
+        {:ok, id} -> id
+        :error -> service_name
+      end
```
shanesveller:
```suggestion
      Keyword.get(service_args, :id, service_name)
```

implementations/elixir/ockam/ockam_services/lib/services/token_lease_manager.ex
```diff
@@ -1,5 +1,13 @@
 defmodule Ockam.Services.TokenLeaseManager do
-  @moduledoc false
+  @moduledoc """
+  Service to manage issuance and expiration for access tokens.
+
+  Options:
+
+  - cloud_service (tuple(2)) - tuple with module implementing Ockam.Services.TokenLeaseManager.CloudService and initialization options
+  - storage_service_module (atom) - module implementing Ockam.Services.TokenLeaseManager.StorageService which stores the tokens
```
shanesveller: If you surround the fully-qualified module names with backticks, it will get rendered as a clickable link when HTML docs are generated, such as on HexDocs once we start publishing these packages.

implementations/elixir/ockam/ockam_services/lib/services/token_lease_manager/cloud_service/influxdb.ex
```diff
@@ -1,5 +1,13 @@
 defmodule Ockam.Services.TokenLeaseManager.CloudService.Influxdb do
-  @moduledoc false
+  @moduledoc """
+  Implementation of Ockam.Services.TokenLeaseManager.CloudService for InfluxDB
+
+  Options:
+  - endpoint (string) - InfluxDB endpoint to connect to
```
shanesveller: Can we be more specific about whether this is a well-formed URI, hostname and optional port, or some other format?

implementations/elixir/ockam/ockam_services/lib/services/token_lease_manager/cloud_service/influxdb.ex
```diff
@@ -1,5 +1,13 @@
...
+  - org_id (string) - InfluxDB organization id to use
+  - token (string) - token with admin permissions authoriza API requests
```
shanesveller: This is slightly unclear to me, can you please comment here or touch up the text?

implementations/elixir/ockam/ockam_services/test/services/app_test.exs
```diff
@@ -6,13 +6,16 @@ defmodule Test.Services.AppTest do
...
     Application.put_env(:ockam_services, :services, [:my_service])
-    Application.put_env(:ockam_services, :services_config_source, nil)
 
     {:ok, _apps} = Application.ensure_all_started(:ockam_services)
 
-    sup = Ockam.Services
+    on_exit(fn ->
+      Application.delete_env(:ockam_services, :service_providers)
+      Application.delete_env(:ockam_services, :services)
+      Application.stop(:ockam_services)
```
shanesveller, line 15: I think this may behave oddly if this test module appears in the middle of the sequence for the test suite (or especially if the module was ever marked async). Is there a less extreme cleanup step we can take instead, or can we at least start this OTP app right back up as the next line? Maybe then we wouldn't need the `ensure_all_started`.

hairyhum: In the rest of this app tests the app is not supposed to be running and this test is testing the start sequence of the app. I think that stopping it as a cleanup step in that context makes sense.

shanesveller (approved):

mszpakowski (approved):

PR #4938, opened 2023-05-16
fix(elixir): fix ciphertext side authorization rule on secure channel

polvorin:
The authorization rule must be applied to the main address, but not to the inner address that is facing the peer.

_Maybe_  we could consider copying the authorization rule from the spawner to the inner address of the secure channel.

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
@@ -206,6 +206,13 @@ defmodule Ockam.SecureChannel.Channel do
   ## AsymmetricWorker callbacks
   @impl true
   def inner_setup(options, %{address: address, inner_address: inner_address} = state) do
+    # The authorization rules must apply to the outer address only.  Inner address is
+    # ciphertext and can come from tcp transport, from tunneled channel, etc.  The secure
+    # channel itself verifies the data comes from the right party.
+    # The call below explicitly set the inner_address authorization, that makes any existing
+    # authorization already setup on the worker, be applied to the main address instead of all.
+    # TODO AsyncWorker could provide out-of-the-box support for this.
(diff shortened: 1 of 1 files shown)


implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
```diff
@@ -206,6 +206,13 @@ defmodule Ockam.SecureChannel.Channel do
   ## AsymmetricWorker callbacks
   @impl true
   def inner_setup(options, %{address: address, inner_address: inner_address} = state) do
+    # The authorization rules must apply to the outer address only.  Inner address is
+    # ciphertext and can come from tcp transport, from tunneled channel, etc.  The secure
+    # channel itself verifies the data comes from the right party.
+    # The call below explicitly set the inner_address authorization, that makes any existing
+    # authorization already setup on the worker, be applied to the main address instead of all.
+    # TODO AsyncWorker could provide out-of-the-box support for this.
```
hairyhum, line 214: I think it should be possible to set address-specific authorization (using map instead of list in `:authorization` option) to set authorization for outer address only.
That is if it's known in advance.

hairyhum (approved):

implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex
polvorin: yes that's possible the problem is one must know the addresses,  or introduce some replacement logic for these.    This was hit on secure channels responders, spawned from a listener that had an authorization rule configured for responders.  At that point the addresses aren't know.    I feel this could be done at the AsyncWorker layer, but defered that for a future improvement

polvorin: (resolving to get it added to  merge queue)

Issue #44, opened 2019-04-09
We need a primitive that bundles key generation, signatures and encryption into one compatible package


malnick:

mrinalwadhwa:
subtask #42

mrinalwadhwa:
Added Cryptographer interface and a libsodium implementation of
https://github.com/ockam-network/ockam/commit/a5821593aa7a780446adf639669459b1b99838e5

mrinalwadhwa:
```
type Cryptographer interface {
	GenerateKeyPair() (publicKey []byte, secretKey []byte, err error)

	AnonymousEncrypt(unencrypted []byte, recipientPublicKey []byte) (encrypted []byte, err error)
	AnonymousDecrypt(encrypted []byte, recipientSecretKey []byte) (unencrypted []byte, err error)

	Sign(message []byte, secretKey []byte) (signed []byte, err error)
	VerifySignature(signed []byte, signerPublicKey []byte) (ok bool, message []byte, err error)

	AuthenticatedEncrypt(unencrypted []byte, senderSecretKey []byte, recipientPublicKey []byte, nonce []byte) (
		encrypted []byte, err error)
	AuthenticatedDecrypt(encrypted []byte, recipientSecretKey []byte, senderPublicKey []byte, nonce []byte) (
		unencrypted []byte, err error)
}
```

PR #4947, opened 2023-05-18
Identity data persistence

etorreborre:
This PR remove some data duplication for the storage of identities.

# Context

This PR is a follow-up to [a previous fix](https://github.com/build-trust/ockam/pull/4936) where an identity read from a configuration file was not found in the identities repository.

In general, if a components needs to access the most recent version of an `Identity` given its identifier, it will use the `IdentitiesRepository` which is implemented using the `LMDB` database storing all the data related to an identity (except for its keys which are stored in a `Vault`).

However at the moment the identity change history is also stored in independent configuration files, making it harder to ensure the consistency of data between the 2 storages.

This PR removes the identity history data from the individual configuration files (`alice.json`, `bob.json`, etc...). 

# Configuration file migration

In order to do this a migration step has been introduced to the loading of configuration data (which will help the migration of other configuration files, for `Vault` in particular).

# CliState initialization

However in order to do this it was important to control the exact moment where the `CliState` is loaded, so that the migrations only run once, at the beginning of the application. This needed some work to:

 1. remove the use of `CliState` by `IdentitiesState` for its implementation of the `delete` method (where `CliState` is used to check if a node is still using that identity))

 2. remove the initialization of a `CliState` by many command-line arguments parsers in order to get default values for node names and identities (note: this is why, in general, a value parser should never have any side effect)

(most of the files modified by this PR are the result of this change)

# Testing

This PR was tested by creating an identity with a previous `ockam` version then show this identity with the new version
```
ockam_old identity create i1
ockam_new identity show --name i1 --full
cat ~/.ockam/identities/i1.json
```
The identity `i1` is fully displayed and the `json` file has been migrated so that it contains only the identity identifier and the enrollement status.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state.rs
@@ -430,7 +430,7 @@ mod tests {
                         let entry_name = entry.file_name().into_string().unwrap();
                         found_entries.push(format!("{dir_name}/{entry_name}"));
                         if entry.path().is_dir() {
-                            assert_eq!(entry_name, "data");
+                            assert_eq!(entry_name, DATA_DIR_NAME);
                             entry.path().read_dir().unwrap().for_each(|entry| {
                                 let entry = entry.unwrap();
                                 let file_name = entry.file_name().into_string().unwrap();
@@ -452,7 +452,7 @@ mod tests {
                         let entry = entry.unwrap();
                         let entry_name = entry.file_name().into_string().unwrap();
                         if entry.path().is_dir() {
-                            assert_eq!(entry_name, "data");
+                            assert_eq!(entry_name, DATA_DIR_NAME);
                             entry.path().read_dir().unwrap().for_each(|entry| {
                                 let entry = entry.unwrap();
                                 let file_name = entry.file_name().into_string().unwrap();
--- a/implementations/rust/ockam/ockam_api/src/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state.rs
@@ -131,6 +131,20 @@ impl CliState {
         Ok(())
     }
 
+    pub fn delete_identity(&self, identity_state: IdentityState) -> Result<()> {
+        // Abort if identity is being used by some running node.
+        for node in self.nodes.list()? {
+            if node.config().identity_config()?.identifier() == identity_state.identifier() {
+                return Err(CliStateError::Invalid(format!(
+                    "Can't delete identity '{}' because is currently in use by node '{}'",
+                    &identity_state.name(),
+                    &node.name()
+                )));
+            }
+        }
+        identity_state.delete()
+    }
+
     /// Returns the default directory for the CLI state.
     pub fn default_dir() -> Result<PathBuf> {
         Ok(get_env_with_default::<PathBuf>(
--- a/implementations/rust/ockam/ockam_api/src/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state.rs
@@ -17,6 +17,7 @@ use ockam::identity::Identities;
 use ockam_core::compat::sync::Arc;
 use ockam_core::env::get_env_with_default;
 use ockam_identity::IdentityIdentifier;
+use ockam_node::Executor;
 use ockam_vault::Vault;
 use rand::random;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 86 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
```diff
@@ -290,6 +238,43 @@ mod traits {
             identity.delete()?;
             Ok(())
         }
+
+        async fn migrate(&self, path: &Path) -> Result<()> {
+            let contents = std::fs::read_to_string(path)?;
```
adrianbenavides, line 243: Sometimes I've had a `.DS_Store` file in some of the state directories, which would cause a failure here, so I think we should ignore files that can't be read.

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
```diff
@@ -85,15 +101,34 @@ pub trait StateDirTrait: Sized {
...
+                items.push(name);
+            }
+        }
+        Ok(items)
+    }
+
+    fn list_items_paths(&self) -> Result<Vec<PathBuf>> {
+        let mut items = Vec::default();
+        for name in self.list_items_names()? {
+            let path = self.path(&name);
```
adrianbenavides, line 133: Related to the other comment, maybe we could ignore pushing the items that don't exist after appending the `.json` prefix here.

adrianbenavides (commented):
It looks really good, thanks for taking the time to go through this. I think removing the command's default values/parsers is a big win, since they are ALL checked when running any command.

I left a small comment with a small bug 🐛

implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
etorreborre: Indeed I am going to fix that!

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
etorreborre: @adrianbenavides please see my last commit. I think that should work.

implementations/rust/ockam/ockam_api/src/cli_state.rs
```diff
@@ -68,33 +70,32 @@ pub struct CliState {
 }
 
 impl CliState {
-    pub fn try_default() -> Result<Self> {
+    /// Return an initialized CliState
+    /// There should only be one call to this function since it also performs a migration
+    /// of configuration files if necessary
+    pub fn initialize() -> Result<Self> {
```
adrianbenavides, line 76: Should we enforce that this function must be called only once?

etorreborre: This actually why I went on a hunt to all the places where we were instantiating the `CliState`. That being said, doing the initialization twice should be idempotent so we should be good.

adrianbenavides: That makes sense, thanks!

implementations/rust/ockam/ockam_command/src/relay/create.rs
```diff
@@ -37,8 +37,8 @@ pub struct CreateCommand {
     relay_name: String,
 
     /// Node for which to create the relay
-    #[arg(long, id = "NODE", display_order = 900, default_value_t = default_node_name(), value_parser = node_name_parser)]
-    to: String,
+    #[arg(long, id = "NODE", display_order = 900)]
+    to: Option<String>,
```
caoakleyii, line 41: I have a concern here.

And below with line #44.

From what I can tell, this will undo the changes that enable commands to create a required resource if needed.

Perhaps we can update the get_node_name/cli_state to allow for creating a resource on demand if one is required.

caoakleyii: To add to what Adrian said above. I do agree, I think it's great that the parsers and default values in this case are going away.

There's a few wins to be gained from moving it away from there; however it's still a functionality we should re-introduce.

etorreborre: @CAOakleyII the behaviour should be exactly the same as before if you have a look at the implementation for `get_node_name`. It uses `default_node_name` if there is no value provided and `node_name_parser` if a value is given. The only difference is that those 2 functions now accept a `CliState` as their argument instead of recreating it with `try_default`

etorreborre: BTW I have nothing at all against default values and parsers, on the contrary! I think it's better if we can parse user-input strings directly as proper types (`IdentityIdentifier`, `PublicKey` etc...) but not if that means triggering side-effects containing business logic or touching the file system.

caoakleyii: image

caoakleyii: The issue is before, it would call the "default_node_name" first, and if there's no "default node" it would return the name "default"


node_name_parser, then would be called after, and again if there's no "default node" it would create the node next.

etorreborre: Thanks @CAOakleyII I get it now. Please see my latest commits where:

 - I call `initialize_node_if_default` and `initialize_identity_if_default` before running an async command (because those 2 functions might call another command and we cannot nest commands)
 
 - I added unit tests to specify the expected behaviour for that piece of code

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
```diff
@@ -196,3 +228,69 @@ pub trait StateItemTrait: Sized {
...
+            &self.path
+        }
+
+        fn config(&self) -> &Self::Config {
+            &self.config
+        }
+    }
+}
+
+trait FileSystem {
```
adrianbenavides: Unused?

adrianbenavides (commented):
Looks good to me and I could run everything correctly now. Only a small comment about a piece of code that looks unused.

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
etorreborre: Good catch. This was the remains of an attempt to unit test without calling the file system. I removed it now

adrianbenavides (approved):
🚀

PR #4960, opened 2023-05-22
feat(rust): add hex flag to message send

### LINKED ISSUE #4918: Allow binary messages in ockam message send
### Current behavior

Only text messages can be sent via `ockam message send` command.
Occasionally, it's useful to be able to send binary messages, for example testing a decryptor.

### Desired behavior

Allow to send binary messages specifying the `--hex` flag.
When this flag is used, both message and response are hex-encoded.

```
ockam message send --from node1 --to /service/echo --hex 00aabbff
00aabbff
```

Neruelin:
Adding support for the hex binary flag to the message send CLI command.
Closes issue #4918 

## Current behavior

ockam message send CLI command doesn't have support for binary messages encoded as hex strings. 

## Proposed changes

ockam message send CLI command will support sending binary messages encoded as hex strings. Adds a '--hex' option telling ockam to interpret the message payload as a hex string and to decode it to a byte array before sending.

## Checks

- [✅] All commits in this Pull Request are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and Verified by Github.
- [✅] All commits in this Pull Request follow the Ockam [commit message convention](https://github.com/build-trust/.github/blob/main/CONTRIBUTING.md#commit-messages).
- [✅] I accept the Ockam Community [Code of Conduct](https://github.com/build-trust/.github/blob/main/CODE_OF_CONDUCT.md).
- [✅] I have accepted the Ockam [Contributor License Agreement](https://github.com/build-trust/ockam-contributors/blob/main/CLA.md) by adding my Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/blob/main/CONTRIBUTORS.csv) file in a separate pull request to the [build-trust/ockam-contributors](https://github.com/build-trust/ockam-contributors) repository. The easiest way to do this is to [edit the CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/edit/main/CONTRIBUTORS.csv) file in the github web UI and create a PR, this will mark the commit as verified.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/message/send.rs
+++ b/implementations/rust/ockam/ockam_command/src/message/send.rs
@@ -33,6 +33,10 @@ pub struct SendCommand {
     #[arg(short, long, value_name = "ROUTE")]
     pub to: MultiAddr,
 
+    /// Message is binary hex flag
+    #[arg(long)]
+    pub hex: bool,
+
     /// Override Default Timeout
     #[arg(long, value_name = "TIMEOUT", default_value = "10")]
     pub timeout: u64,
@@ -91,13 +95,28 @@ async fn rpc(mut ctx: Context, (opts, cmd): (CommandGlobalOpts, SendCommand)) ->
         let mut rpc = RpcBuilder::new(ctx, opts, &api_node)
             .tcp(tcp.as_ref())?
             .build();
-        rpc.request_with_timeout(req(&to, &cmd.message), Duration::from_secs(cmd.timeout))
+        if cmd.hex {
+            rpc.request_with_timeout(
+                req_with_bytes(
+                    &to,
+                    hex::decode(&cmd.message)
+                        .context("Provided message is not a valid hex string")?
+                        .as_slice(),
+                ),
+                Duration::from_secs(cmd.timeout),
+            )
             .await?;
-        let res = rpc.parse_response::<Vec<u8>>()?;
-        println!(
-            "{}",
-            String::from_utf8(res).context("Received content is not a valid utf8 string")?
-        );
+            let res = rpc.parse_response::<Vec<u8>>()?;
+            println!("{}", hex::encode(res));
+        } else {
+            rpc.request_with_timeout(req(&to, &cmd.message), Duration::from_secs(cmd.timeout))
+                .await?;
+            let res = rpc.parse_response::<Vec<u8>>()?;
+            println!(
+                "{}",
+                String::from_utf8(res).context("Received content is not a valid utf8 string")?
+            );
+        }
 
         // only delete node in case 'from' is empty and embedded node was started before
         if cmd.from.is_none() {
@@ -111,4 +130,11 @@ async fn rpc(mut ctx: Context, (opts, cmd): (CommandGlobalOpts, SendCommand)) ->
 
 pub(crate) fn req<'a>(to: &'a MultiAddr, message: &'a str) -> RequestBuilder<'a, SendMessage<'a>> {
     Request::post("v0/message").body(SendMessage::new(to, message.as_bytes()))
+}
+
+pub(crate) fn req_with_bytes<'a>(
+    to: &'a MultiAddr,
+    message: &'a [u8],
+) -> RequestBuilder<'a, SendMessage<'a>> {
+    Request::post("v0/message").body(SendMessage::new(to, message))
 }
--- a/implementations/rust/ockam/ockam_command/src/message/send.rs
+++ b/implementations/rust/ockam/ockam_command/src/message/send.rs
@@ -11,7 +11,7 @@ use ockam_multiaddr::MultiAddr;
 use crate::node::util::{delete_embedded_node, start_embedded_node_with_vault_and_identity};
 use crate::util::api::{CloudOpts, TrustContextOpts};
 use crate::util::{clean_nodes_multiaddr, extract_address_value, node_rpc, RpcBuilder};
-use crate::Result;
+use crate::{Result};
 use crate::{docs, CommandGlobalOpts};
 
 const LONG_ABOUT: &str = include_str!("./static/send/long_about.txt");
@@ -33,7 +33,7 @@ pub struct SendCommand {
     #[arg(short, long, value_name = "ROUTE")]
     pub to: MultiAddr,
 
-    /// Message is binary hex flag
+    /// Flag to indicate that the message is hex encoded
     #[arg(long)]
     pub hex: bool,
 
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_command/src/message/send.rs
```diff
@@ -112,3 +131,10 @@ async fn rpc(mut ctx: Context, (opts, cmd): (CommandGlobalOpts, SendCommand)) ->
 pub(crate) fn req<'a>(to: &'a MultiAddr, message: &'a str) -> RequestBuilder<'a, SendMessage<'a>> {
     Request::post("v0/message").body(SendMessage::new(to, message.as_bytes()))
 }
+
+pub(crate) fn req_with_bytes<'a>(
+    to: &'a MultiAddr,
+    message: &'a [u8],
+) -> RequestBuilder<'a, SendMessage<'a>> {
+    Request::post("v0/message").body(SendMessage::new(to, message))
+}
```
adrianbenavides: Since a `&str` can be converted to a bytes array, you could refactor the existing `req` function to do what `req_with_bytes` is doing so we only have one function here.

If you do this you'll be able to simplify the request building (the section between lines 98-119).

implementations/rust/ockam/ockam_command/src/message/send.rs
```diff
@@ -33,6 +33,10 @@ pub struct SendCommand {
     #[arg(short, long, value_name = "ROUTE")]
     pub to: MultiAddr,
 
+    /// Message is binary hex flag
```
adrianbenavides: I'd reword this to: `Flag to indicate that the message is hex encoded`

implementations/rust/ockam/ockam_command/src/message/send.rs
```diff
@@ -91,13 +95,28 @@ async fn rpc(mut ctx: Context, (opts, cmd): (CommandGlobalOpts, SendCommand)) ->
...
+            println!("{}", hex::encode(res));
+        } else {
+            rpc.request_with_timeout(req(&to, &cmd.message), Duration::from_secs(cmd.timeout))
+                .await?;
+            let res = rpc.parse_response::<Vec<u8>>()?;
+            println!(
+                "{}",
+                String::from_utf8(res).context("Received content is not a valid utf8 string")?
+            );
+        }
```
adrianbenavides: A comment on how to print the output. We are slowly refactoring all the commands to use the `opts.terminal` tool, and this would be a good chance to update it.

Here is [an example](https://github.com/build-trust/ockam/blob/ec7268aeef9b59bb22f3ae36a7515bd03ec09a4b/implementations/rust/ockam/ockam_command/src/vault/delete.rs#L52) on how to do it.

adrianbenavides (changes requested):
Hey @Neruelin, the PR looks great, thanks for taking the time to work on this one 🙏 

I've left some minor comments with things that could be improved. 

One last thing that I'd like to ask you for to make this PR perfect is to add a test within the [message bats tests file](https://github.com/build-trust/ockam/blob/ec7268aeef9b59bb22f3ae36a7515bd03ec09a4b/implementations/rust/ockam/ockam_command/tests/bats/message.bats). You can take a look at the first test of that file as a reference.

Let me know if you need any help!

implementations/rust/ockam/ockam_command/src/message/send.rs
Neruelin: I've updated the branch, let me know if the changes are what you are looking for. Ran into some issues understanding the lifetime of the data returned by the hex::decode call. Had to do some weird stuff to keep the data from being freed too early. If you have any idea how to circumvent this issue or improve it, I'd love to hear your thoughts.

implementations/rust/ockam/ockam_command/src/message/send.rs
```diff
@@ -91,13 +95,40 @@ async fn rpc(mut ctx: Context, (opts, cmd): (CommandGlobalOpts, SendCommand)) ->
...
+                // return empty vec that will never be used, to satisfy match
+                vec![]
+            }
+        };
+
+        let msg_req = if cmd.hex {
+            req(&to, hex_bytes.as_slice())
+        } else {
+            req(&to, cmd.message.as_bytes())
+        };
```
adrianbenavides: This is close to what I was suggesting in my previous comment but note how you have the `if cmd.hex` intertwined to return an error or build the request.

The trick here is to return the appropriate types at each step to avoid the problems you mentioned. This is how I'd do it so that we only use the `hex` flag to encode/decode the message:

```rust
// Prepare request
let mut rpc = RpcBuilder::new(ctx, opts, &api_node)
    .tcp(tcp.as_ref())?
    .build();
let to = crate::project::util::clean_projects_multiaddr(to, projects_sc)?;
let message = if cmd.hex {
    hex::decode(cmd.message)?
} else {
    cmd.message.as_bytes().to_vec()
};
let req = req(&to, message.as_slice());

// Send request
rpc.request_with_timeout(req, Duration::from_secs(cmd.timeout))
    .await?;
let res = {
    let res = rpc.parse_response::<Vec<u8>>()?;
    if cmd.hex {
        hex::encode(res)
    } else {
        String::from_utf8(res).context("Received content is not a valid utf8 string")?
    }
};
```

implementations/rust/ockam/ockam_command/src/message/send.rs
```diff
@@ -91,13 +95,40 @@ async fn rpc(mut ctx: Context, (opts, cmd): (CommandGlobalOpts, SendCommand)) ->
...
-        );
+        let res = {
+            let res = rpc.parse_response::<Vec<u8>>()?;
+            if cmd.hex {
+                hex::encode(res)
+            } else {
+                String::from_utf8(res).context("Received content is not a valid utf8 string")?
+            }
+        };
+        opts.terminal.write_line(&res)?;
```
adrianbenavides: This is still not right, as this is logging a message to stderr instead of writing the command output to stdout. To do that, you have to do the following: `opts.terminal.stdout().plain(res).write_line()?;`. Although you'll have to do a slight modification to the `go` function signature to receive the full `CommandGlobalOpts` instance (or flatten out the function code within the `rpc` function, up to you).

adrianbenavides (changes requested):

implementations/rust/ockam/ockam_command/src/message/send.rs
Neruelin: Modified the go signature and fixed the logging call.

implementations/rust/ockam/ockam_command/src/message/send.rs
Neruelin: Thanks for the feedback here, I had tried something similar early on but kept encountering an error where the temporary value returned from hex::decode was dropped before being assigned to the variable. I'm not sure where I went wrong before, but I think I understand why this does work.

adrianbenavides: Yeah, ownership rules can be a little bit tricky. The compiler is the best teacher on this 😆

adrianbenavides (approved):
Looks perfect now, thanks a lot @Neruelin for another great contribution!! 🚀

etorreborre (approved):

PR #4962, opened 2023-05-22
test(rust): use tokio test to simplify some unit tests

etorreborre:
This PR uses the `#[tokio::test]` macro on some unit tests. This removes some clippy warnings which I am getting locally though they don't appear on CI though (but I don't know exactly why):
```rust
  assert!(poll_once(async { ac.is_authorized(&msg).await })?);
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: you can reduce it to: `ac.is_authorized(&msg)`
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3401,6 +3401,7 @@ dependencies = [
  "spin 0.9.8",
  "subtle",
  "tinyvec",
+ "tokio",
  "tracing",
  "tracing-error",
  "zeroize",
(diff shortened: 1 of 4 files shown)


adrianbenavides (approved):
> This removes some clippy warnings which I am getting locally though they don't appear on CI though (but I don't know exactly why):

Probably because we use v1.68 on CI

etorreborre:
> Probably because we use v1.68 on CI

That's also what I'm using, there must be something else, I'll try to get to the bottom of it

PR #4978, opened 2023-05-23
fix(rust): retrieve the correct key id for a public key with aws kms

etorreborre:
This PR fixes the use of the AWS KMS as a backend for a `Vault`.

# Observed issue

When using the `AttachKey` command to create an identity a call to the AWS KMS client was performed to retrieve the key id corresponding to a given public key. 

This call is necessary because when we compute an `IdentityIdentifier` for an identity we derive it from the key id related to the public key used for the first change in the change history.
Before this PR the call to `kms_client.compute_key_id(&public_key)` was not implemented correctly and could not retrieve the key id.

# Diagnostic

The trouble is that there is no efficient implementation of this operation with the AWS KMS. That operation needs to list all the existing keys to retrieve the key with the right public key.

# Proposed fix

This PR fixes the issue by:

 - adding a local storage for public key <-> key id when the AWS security module is used
 - retrieving the key id directly from storage when the public key id is given and the mapping exists
 - retrieving the key id from the AWS KMS if it has never been accessed before (and then store the mapping locally)
 - doing a local verification of signatures instead of using the AWS KMS and incurring a remote call
 - removing the computation of an `IdentityIdentifier` from the key id associated to a public key. Now the identity identifier is directly a sha256 hash of the first public key of an identity

# Tests

Several tests have been added to:

 - check the low-level behaviour of the storage of the public key <-> key id mapping
 - check that a signature can be verified both locally and remotely with a given AWS KMS key
 - check that the identity creation done when running the `attach-key` command works when using an AWS KMS backend

Note: those tests cannot be part of the test suite for now because they require an access to an AWS account to create keys.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
@@ -348,7 +348,7 @@ mod tests {
     fn create_identity_config() -> IdentityConfig {
         let data = hex::decode("0144c7eb72dd1e633f38e0d0521e9d5eb5072f6418176529eb1b00189e4d69ad2e000547c93239ba3d818ec26c9cdadd2a35cbdf1fa3b6d1a731e06164b1079fb7b8084f434b414d5f524b03012000000020c6c52380125d42b0b4da922b1cff8503a258c3497ec8ac0b4a3baa0d9ca7b3780301014075064b902bda9d16db81ab5f38fbcf226a0e904e517a8c087d379ea139df1f2d7fee484ac7e1c2b7ab2da75f85adef6af7ddb05e7fa8faf180820cb9e86def02").unwrap();
         let identity = Identity::new(
-            IdentityIdentifier::from_string(
+            IdentityIdentifier::from_hex(
                 "fa804b7fca12a19eed206ae180b5b576860ae6512f196c189d90661bcc434b50",
             ),
             IdentityChangeHistory::import(data.to_vec().as_slice()).unwrap(),
--- a/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
@@ -319,11 +319,8 @@ mod traits {
 
 #[cfg(test)]
 mod tests {
-    use core::str::FromStr;
-
-    use ockam_identity::IdentityChangeHistory;
-
     use super::*;
+    use ockam_identity::IdentityChangeHistory;
 
     #[test]
     fn test_serialize() {
@@ -351,10 +348,9 @@ mod tests {
     fn create_identity_config() -> IdentityConfig {
         let data = hex::decode("0144c7eb72dd1e633f38e0d0521e9d5eb5072f6418176529eb1b00189e4d69ad2e000547c93239ba3d818ec26c9cdadd2a35cbdf1fa3b6d1a731e06164b1079fb7b8084f434b414d5f524b03012000000020c6c52380125d42b0b4da922b1cff8503a258c3497ec8ac0b4a3baa0d9ca7b3780301014075064b902bda9d16db81ab5f38fbcf226a0e904e517a8c087d379ea139df1f2d7fee484ac7e1c2b7ab2da75f85adef6af7ddb05e7fa8faf180820cb9e86def02").unwrap();
         let identity = Identity::new(
-            IdentityIdentifier::from_str(
-                "Pfa804b7fca12a19eed206ae180b5b576860ae6512f196c189d90661bcc434b50",
-            )
-            .unwrap(),
+            IdentityIdentifier::from_string(
+                "   fa804b7fca12a19eed206ae180b5b576860ae6512f196c189d90661bcc434b50",
+            ),
             IdentityChangeHistory::import(data.to_vec().as_slice()).unwrap(),
         );
         IdentityConfig {
--- a/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
@@ -349,7 +349,7 @@ mod tests {
         let data = hex::decode("0144c7eb72dd1e633f38e0d0521e9d5eb5072f6418176529eb1b00189e4d69ad2e000547c93239ba3d818ec26c9cdadd2a35cbdf1fa3b6d1a731e06164b1079fb7b8084f434b414d5f524b03012000000020c6c52380125d42b0b4da922b1cff8503a258c3497ec8ac0b4a3baa0d9ca7b3780301014075064b902bda9d16db81ab5f38fbcf226a0e904e517a8c087d379ea139df1f2d7fee484ac7e1c2b7ab2da75f85adef6af7ddb05e7fa8faf180820cb9e86def02").unwrap();
         let identity = Identity::new(
             IdentityIdentifier::from_string(
-                "   fa804b7fca12a19eed206ae180b5b576860ae6512f196c189d90661bcc434b50",
+                "fa804b7fca12a19eed206ae180b5b576860ae6512f196c189d90661bcc434b50",
             ),
             IdentityChangeHistory::import(data.to_vec().as_slice()).unwrap(),
         );
--- a/implementations/rust/ockam/ockam_api/src/cli_state/vaults.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/vaults.rs
@@ -40,9 +40,13 @@ pub struct VaultState {
 impl VaultState {
     pub async fn get(&self) -> Result<Arc<Vault>> {
         if self.config.aws_kms {
-            let config = AwsKmsConfig::new();
+            let config = AwsKmsConfig::default().await?;
             Ok(Vault::create_with_security_module(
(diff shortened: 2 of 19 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
```diff
@@ -351,10 +348,9 @@ mod tests {
     fn create_identity_config() -> IdentityConfig {
         let data = hex::decode("0144c7eb72dd1e633f38e0d0521e9d5eb5072f6418176529eb1b00189e4d69ad2e000547c93239ba3d818ec26c9cdadd2a35cbdf1fa3b6d1a731e06164b1079fb7b8084f434b414d5f524b03012000000020c6c52380125d42b0b4da922b1cff8503a258c3497ec8ac0b4a3baa0d9ca7b3780301014075064b902bda9d16db81ab5f38fbcf226a0e904e517a8c087d379ea139df1f2d7fee484ac7e1c2b7ab2da75f85adef6af7ddb05e7fa8faf180820cb9e86def02").unwrap();
         let identity = Identity::new(
-            IdentityIdentifier::from_str(
-                "Pfa804b7fca12a19eed206ae180b5b576860ae6512f196c189d90661bcc434b50",
-            )
-            .unwrap(),
+            IdentityIdentifier::from_string(
+                "   fa804b7fca12a19eed206ae180b5b576860ae6512f196c189d90661bcc434b50",
```
SanjoDeundiak: Are those spaces in the string intended?

SanjoDeundiak: And why we don't include the prefix `P` now?

implementations/rust/ockam/ockam_api/src/cli_state/vaults.rs
```diff
@@ -40,9 +40,13 @@ pub struct VaultState {
 impl VaultState {
     pub async fn get(&self) -> Result<Arc<Vault>> {
         if self.config.aws_kms {
-            let config = AwsKmsConfig::new();
+            let config = AwsKmsConfig::default().await?;
             Ok(Vault::create_with_security_module(
-                AwsSecurityModule::create(config).await?,
+                AwsSecurityModule::create_with_storage_path(
+                    config,
+                    self.vault_file_path().as_path(),
```
SanjoDeundiak, line 47: Are we reusing the same file path for the software vault storage and AWS public key <-> key id mapping?

implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
etorreborre: No, I fixed it now.

etorreborre: Because `from_string` includes it by default. I actually don't know why we need this prefix

implementations/rust/ockam/ockam_api/src/cli_state/vaults.rs
etorreborre: Yes because the 2 are exclusive

implementations/rust/ockam/ockam_vault_aws/src/vault/aws_kms_client.rs
```diff
@@ -143,22 +150,24 @@ impl AwsKmsClient {
...
+            log::error!(%public_key, %err, "failed to list all keys");
+            Error::MissingKeys
+        })?;
+
+        if let Some(keys) = output.keys() {
+            log::debug!(%public_key, "received keys");
+            for key in keys {
+                if let Some(key_id) = key.key_id() {
+                    let one_public_key = self.public_key(&key_id.to_string()).await?;
+                    if one_public_key == public_key.clone() {
```
SanjoDeundiak: Is Clone really necessary here?

implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
etorreborre: I renamed the function to `from_hex` to remove the ambiguity.

implementations/rust/ockam/ockam_vault_aws/src/vault/aws_kms_client.rs
etorreborre: Removed

implementations/rust/ockam/ockam_vault_aws/src/vault/aws_security_module.rs
```diff
@@ -52,13 +100,105 @@ impl SecurityModule for AwsSecurityModule {
...
         message: &[u8],
         signature: &Signature,
     ) -> Result<bool> {
-        let key_id = self.get_key_id(public_key).await?;
-        self.client.verify(&key_id, message, signature).await
+        use p256::ecdsa::{signature::Verifier as _, Signature, VerifyingKey};
+
+        let verifying_key =
+            VerifyingKey::from_public_key_der(public_key.data()).map_err(Self::from_pkcs8)?;
+        let ecdsa_signature = Signature::from_der(signature.as_ref()).map_err(Self::from_ecdsa)?;
```
SanjoDeundiak, line 118: I wonder if PublicKeys and Signature have the same format for both Software Vault and AWS Vault. Probably they are if things work, but I haven't checked

implementations/rust/ockam/ockam_vault_aws/src/vault/aws_security_module.rs
```diff
@@ -52,13 +100,105 @@ impl SecurityModule for AwsSecurityModule {
...
+    #[ignore]
+    async fn test_sign_verify() -> Result<()> {
+        let security_module = AwsSecurityModule::default().await?;
+        let key_id = security_module
+            .create_secret(SecretAttributes::NistP256)
+            .await?;
+        let message = b"hello world";
+        let signature = security_module.sign(&key_id, &message[..]).await?;
+
+        // Verify locally
```
SanjoDeundiak, line 187: Aren't these 2 calls identical?

implementations/rust/ockam/ockam_vault_aws/src/vault/aws_kms_client.rs
SanjoDeundiak: Could we do? :D
```
if &one_public_key == public_key {
```

implementations/rust/ockam/ockam_vault_aws/src/vault/aws_security_module.rs
etorreborre: That's the thing tested here:

 - the first one uses a local implementation `security_module.sign`
 - the second one uses a remote implementation via the client `security_module.client.sign`

This test is a sanity check that both methods return the same result.

implementations/rust/ockam/ockam_vault_aws/src/vault/aws_security_module.rs
etorreborre: That is what is tested with [the test below](https://github.com/build-trust/ockam/pull/4978/files/b901a7ee95e0313a7ec56934f905a286ae393c34#diff-755cfc699e456d340eaf60d41309f8be429b740193484b83b8cc0dfd253053a0R178).

implementations/rust/ockam/ockam_vault_aws/src/vault/aws_kms_client.rs
etorreborre: I'm doing it the other way around: https://github.com/build-trust/ockam/pull/4978/commits/d377f2a47b3681933dcaf569d8e56e77acf5c9b1. Is one more idiomatic than the other?

implementations/rust/ockam/ockam_vault_aws/src/vault/aws_security_module.rs
SanjoDeundiak: So why do both verify calls use `security_module.client`?

implementations/rust/ockam/ockam_vault_aws/src/vault/aws_kms_client.rs
SanjoDeundiak: Yeah, but isn't that causing an excess data copy? Which isn't much, but why don't we just use references?

etorreborre: `one_public_key == *public_key` is causing data copy?

SanjoDeundiak: I had an impression that `*public_key` creates a copy

etorreborre: It seems that both are the same but (according to at least one person on StackOverflow) comparing references is more frequent: https://stackoverflow.com/questions/74070545/comparing-a-value-and-a-reference. So I changed to that version.

implementations/rust/ockam/ockam_vault_aws/src/vault/aws_security_module.rs
etorreborre: Because I read too fast my own code :-| and confused `sign` for `verify` (even after your comment...). I have now implemented the test properly and executed it.

SanjoDeundiak (approved):

PR #4984, opened 2023-05-24
fix(rust): properle use `TrustContext` in both listener and client

davide-baldo:


implementations/rust/ockam/ockam_identity/src/credentials/authority_service.rs
```diff
@@ -13,6 +16,13 @@ pub struct AuthorityService {
     credentials: Arc<dyn Credentials>,
     identifier: IdentityIdentifier,
     own_credential: Option<Arc<dyn CredentialsRetriever>>,
+    inner_cache: Arc<RwLock<Option<CachedCredential>>>,
```
davide-baldo, line 19: this cache avoids asking a credential to the authority at each connection (previous behavior, but now it happens even if no credential is presented) and should make the connections more responsive.

implementations/rust/ockam/ockam_identity/src/secure_channel/listener.rs
```diff
@@ -94,14 +94,29 @@ impl Worker for IdentityChannelListener {
             .options
             .create_access_control(flow_control_id.clone())?;
 
+        let credentials = if self.options.credentials.is_empty() {
+            if let Some(trust_context) = &self.options.trust_context {
+                vec![
```
davide-baldo, line 99: currently we are not doing something like this on the client side, I wonder if we should

implementations/rust/ockam/ockam_identity/src/credentials/authority_service.rs
```diff
@@ -42,15 +53,38 @@ impl AuthorityService {
...
         for_identity: &IdentityIdentifier,
     ) -> Result<Credential> {
+        {
+            // check if we have a valid cached credential
+            let guard = self.inner_cache.read().unwrap();
+            let now = Timestamp::now().ok_or_else(|| {
+                Error::new(Origin::Application, Kind::Invalid, "invalid system time")
+            })?;
+            if let Some(cache) = guard.as_ref() {
+                // add an extra minute to have a bit of leeway for clock skew
```
polvorin, line 63: maybe just a minute is too low as we don't refresh credentials right now.   But we can see and adjust that latter on

implementations/rust/ockam/ockam_identity/src/secure_channel/listener.rs
polvorin: I think we shouldn't set the credential in neither the listener or connecting side.     Setting the trust_context should be the way to go.    And ask for a credential through it.   [There are different ways in which it can get the credential ](https://github.com/build-trust/ockam/blob/42b4d2794b37f2b7ba8a4e610257f4ddfe539885/implementations/rust/ockam/ockam_api/src/config/cli.rs#L279-L299),  as from an online issuer,  directly from memory or from credential stored on disk.   Thats was the idea at least.    I think it's fine to merge this as-is as not sure if that is how it ended up actually implemented.

polvorin (approved):

PR #4986, opened 2023-05-24
refactor(rust): extract a full state machine for the secure channel handshake

etorreborre:
The purpose of this PR is make the code involved in creating secure channels easier to review by:

 - having a clear separation of concerns between 3 aspects
    1. exchanging messages 
    2. keeping track of the current step in the protocol
    3. implementing the encoding and decoding of the protocol messages
  - putting all the code in the same crate and handshake operations close to one another

# Implementation

## Structure

The implementation consists in 3 layers:

image

 1. `HandshakeWorker`: a worker implementing the initiating and responding side of the protocol only dealing with receiving and sending Ockam messages. It delegates everything else to the state machine via `on_event`
 
 2. `InitiatorStateMachine` and `ResponderStateMachine`: explicit state machines keeping track of the current state depending on which message from [the Noise protocol](https://noiseprotocol.org/noise.html) has been sent or received

 3. `Handshake`: encoding and decoding of the Noise messages depending on the state variables: `s`, `e`, `h`,  `n`,... specified by the protocol

## Identity and credentials verification

The code checking the identity and credentials of the other party at the end of the handshake is performed by the state machine. That code is similar in both the `Initiator` and `Responder` case so it has been placed in a `CommonStateMachine` struct.

This means that the `Handshake` code is agnostic about the kind of payloads that are appended to message 2 and message 3.

## Suppression of the `ockam_key_exchange_xx` crate

We are only implementing one type of handshake for now (the Noise protocol defines several types of handshakes). In order to reduce indirections and the number of crates in the project all the handshake code has been inlined into the `ockam_identity` crate, close to the rest of the secure channels code.

### `KeyId`

The `KeyId` type doesn't have to be shared by several crates now so it has been moved to the `ockam_vault` where it is used for secrets management.

## Tests

The tests checking the exact outputs of the handshake procedure have been ported to the new code.


implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+impl HandshakeState {
+    /// Create a new HandshakeState with:
+    ///   - a static key
+    ///   - an ephemeral key
+    ///   - an encoded identity payload (identity + signature + credentials)
+    pub(super) fn new(s: KeyId, e: KeyId, message1_payload: Vec<u8>) -> HandshakeState {
+        HandshakeState {
+            s,
+            e,
+            k: "".to_string(),
```
SanjoDeundiak: Why don't use `Option` for `k`, `re`, `rs`, `ck`? I think `PublicKey` should always contain a valid Public Key. `KeyId` should point to an existent key in the Vault (though this one is hard to guarantee 100% of time)

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+            h: [0u8; SHA256_SIZE_USIZE],
+            ck: "".to_string(),
+            message1_payload,
+            status: Initial,
+        }
+    }
+
+    /// h = SHA256(h || data)
+    pub(super) fn mix_hash(&mut self, data: &[u8]) {
+        let mut input = self.h.to_vec();
```
SanjoDeundiak: Off topic for this PR, but can we make this operation more efficient? It's called pretty often

SanjoDeundiak: I'm talking about allocating the memory of the right size, which we know

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+        // 3. initialize the handshake
+        // We currently don't use any payload for message 1
+        Self::new_with_keys(vault, static_key, ephemeral_key, vec![]).await
+    }
+
+    /// Initialize the handshake
+    async fn new_with_keys(
+        vault: Arc<dyn XXVault>,
+        static_key: KeyId,
+        ephemeral_key: KeyId,
```
SanjoDeundiak: I don't think ability to provide pre-generated ephemeral key is useful

SanjoDeundiak: Or are there cases?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+    pub(super) fn get_handshake_keys(&self) -> Option<HandshakeKeys> {
+        match self.state.status.clone() {
+            Ready(keys) => Some(keys),
+            _ => None,
+        }
+    }
+}
+
+impl Handshake {
+    /// Create a new handshake initialized with
```
SanjoDeundiak: Please fix the description

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+    ///   - a static key
+    ///   - an ephemeral key
+    ///   - a payload containing the identity, a signature of the static key and the credential of
+    ///     the current party
+    pub(super) async fn new(vault: Arc<dyn XXVault>) -> Result<Handshake> {
```
SanjoDeundiak: Do you think it would be better to always pass static key?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+        // We currently don't use any payload for message 1
+        Self::new_with_keys(vault, static_key, ephemeral_key, vec![]).await
+    }
+
+    /// Initialize the handshake
+    async fn new_with_keys(
+        vault: Arc<dyn XXVault>,
+        static_key: KeyId,
+        ephemeral_key: KeyId,
+        message1_payload: Vec<u8>,
```
SanjoDeundiak: Why does it require payload here?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+    }
+
+    /// Compute a Diffie-Hellman key between a given key id and the other party public key
+    async fn dh(&self, key_id: &KeyId, public_key: &PublicKey) -> Result<KeyId> {
+        self.vault.ec_diffie_hellman(key_id, public_key).await
+    }
+
+    /// Compute two derived ck, and k keys based on existing ck and k keys + an optional
+    /// Diffie-Hellman key
+    async fn hkdf(&self, state: &mut HandshakeState, dh: Option<&KeyId>) -> Result<(KeyId, KeyId)> {
```
SanjoDeundiak: Does it make sense to split this function into 2 functions:
1. Derive new `ck` and `k`
2. Derive `k1` and `k2`
?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+
+/// The `HandshakeState` contains all the variables necessary to follow the Noise protocol
+#[derive(Debug, Clone)]
+pub(super) struct HandshakeState {
+    pub(super) s: KeyId,
+    e: KeyId,
+    k: KeyId,
+    re: PublicKey,
+    pub(super) rs: PublicKey,
+    n: u16,
```
SanjoDeundiak: According to the noise spec `n` should be u64. While during the handshake `n` value could be easily fitted into `u8`. Where does `u16` come from?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+
+    /// Encrypt a plaintext 'c' using the key 'k' and the additional data 'h'
+    async fn encrypt_and_hash(&self, state: &mut HandshakeState, c: &[u8]) -> Result<Vec<u8>> {
+        let mut nonce = [0u8; 12];
+        nonce[10..].copy_from_slice(&state.n.to_be_bytes());
+
+        let result = self
+            .vault
+            .aead_aes_gcm_encrypt(&state.k, c, nonce.as_ref(), &state.h)
+            .await
```
SanjoDeundiak: Could it be `?.to_vec();`?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+            .aead_aes_gcm_decrypt(&state.k, c, nonce.as_ref(), &state.h)
+            .await
+            .map(|b| b.to_vec())?;
+        state.mix_hash(c);
+        state.n += 1;
+        Ok(result)
+    }
+
+    /// Encrypt a plaintext 'c' using the key 'k' and the additional data 'h'
+    async fn encrypt_and_hash(&self, state: &mut HandshakeState, c: &[u8]) -> Result<Vec<u8>> {
```
SanjoDeundiak: Can we please have different names for encrypted and decrypted binaries? I guess `c` is cipher text, then maybe plain text could be `p`?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+    /// Encode the first message, sent from the initiator to the responder
+    pub(super) async fn encode_message1(&mut self) -> Result<Vec<u8>> {
+        let mut state = self.state.clone();
+        // output e.pubKey
+        let e_pub_key = self.get_public_key(&state.e).await?;
+        state.mix_hash(e_pub_key.data());
+        let mut message = e_pub_key.data().to_vec();
+
+        // output message 1 payload
+        message.extend(state.message1_payload.clone());
```
SanjoDeundiak: Can we get rid of message1 payload?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+        Ok(())
+    }
+
+    /// Encode the first message, sent from the initiator to the responder
+    pub(super) async fn encode_message1(&mut self) -> Result<Vec<u8>> {
+        let mut state = self.state.clone();
+        // output e.pubKey
+        let e_pub_key = self.get_public_key(&state.e).await?;
+        state.mix_hash(e_pub_key.data());
+        let mut message = e_pub_key.data().to_vec();
```
SanjoDeundiak, line 60: Can we at least have a check for endianness and throw an error if it's little-endian?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+                &state.ck,
+                b"",
+                dh,
+                vec![ck_attributes, Self::k_attributes()],
+            )
+            .await?;
+
+        match hkdf_output.as_slice() {
+            [new_ck, new_k] => {
+                self.vault.delete_ephemeral_secret(state.ck.clone()).await?;
```
SanjoDeundiak: Should we set state.ck = None, state.k = None?

SanjoDeundiak: Which is the point of delete_ephemeral_secret taking an owned argument

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_state_machine.rs
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    /// Deserialize a payload as D from a bare encoding
+    pub(super) fn deserialize<D: for<'a> Deserialize<'a>>(payload: Vec<u8>) -> Result<D> {
+        serde_bare::from_slice(payload.as_slice())
+            .map_err(|error| Error::new(Origin::Channel, Kind::Invalid, error))
+    }
+
+    /// Sign the static key used in the key exchange with the identity private key
+    async fn sign_static_key(&self, identity: Identity, key_id: &KeyId) -> Result<Signature> {
```
SanjoDeundiak, line 145: I think we should pass the signature to that state machine instead of signing here.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_state_machine.rs
```diff
@@ -0,0 +1,256 @@
...
+        let identity = self.decode_identity(peer.identity).await?;
+        self.verify_signature(&identity, &peer.signature, peer_public_key)
+            .await?;
+        self.verify_credentials(&identity, peer.credentials).await?;
+        self.their_identifier = Some(identity.identifier());
+        Ok(())
+    }
+
+    /// Deserialize a payload as D from a bare encoding
+    pub(super) fn deserialize<D: for<'a> Deserialize<'a>>(payload: Vec<u8>) -> Result<D> {
```
SanjoDeundiak: Could it be named `deserialize_payload`?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -0,0 +1,314 @@
...
+        &mut self,
+        context: &mut Self::Context,
+        message: Routed<Self::Message>,
+    ) -> Result<()> {
+        if let Some(decryptor) = self.decryptor.as_mut() {
+            // since we cannot replace this worker with the decryptor worker we act as a relay instead
+            return decryptor.handle_message(context, message).await;
+        };
+
+        // we only set it once to avoid redirects attack
```
SanjoDeundiak: How is it guaranteed that we only set it once? And what is the redirect attack in this case?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -0,0 +1,314 @@
...
+        self.remote_route = Some(message.return_route());
+        if let SendMessage(message) = self
+            .state_machine
+            .on_event(ReceivedMessage(Vec::<u8>::decode(
+                &message.into_transport_message().payload,
+            )?))
+            .await?
```
SanjoDeundiak, line 105: How should we handle error here? Should we shutdown the worker?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -0,0 +1,314 @@
...
+        );
+
+        if role.is_initiator() {
+            if let Some(timeout) = timeout {
+                callback_waiter.receive_timeout(timeout).await?;
+            }
+            Ok(addresses.encryptor)
+        } else {
+            // wait until the worker is ready to receive messages
+            context.wait_for(addresses.decryptor_remote.clone()).await?;
```
SanjoDeundiak: I don't understand why is it required. I had an impression that successfully starting a worker is already a guarantee that its mailbox was created and we can start send messages to it.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -0,0 +1,314 @@
...
+        addresses: Addresses,
+        identifier: IdentityIdentifier,
+        trust_policy: Arc<dyn TrustPolicy>,
+        decryptor_outgoing_access_control: Arc<dyn OutgoingAccessControl>,
+        credentials: Vec<Credential>,
+        trust_context: Option<TrustContext>,
+        remote_route: Option<Route>,
+        timeout: Option<Duration>,
+        role: Role,
+    ) -> Result<Address> {
```
SanjoDeundiak: What is the point of returning an Address here?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -0,0 +1,314 @@
...
+            Ok(addresses.decryptor_remote)
+        }
+    }
+
+    /// Return the route for the other party's handshake worker
+    /// That route is set once and for all once we receive a message coming from that worker
+    fn remote_route(&self) -> Route {
+        self.remote_route
+            .clone()
+            .expect("a remote route must have been already set for sending a message")
```
SanjoDeundiak: Can we replace that with an Error

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -0,0 +1,314 @@
...
+        );
+        trace!(
+            "SecureChannel {} started for addresses: {:?}",
+            role.clone(),
+            addresses.clone(),
+        );
+
+        if role.is_initiator() {
+            if let Some(timeout) = timeout {
+                callback_waiter.receive_timeout(timeout).await?;
```
SanjoDeundiak, line 217: This is unexpected for me that `HandshakeWorker::create()` would wait for the handshake to complete

SanjoDeundiak: Also that depending on a role this function exits on very different conditions.

SanjoDeundiak: Also that depending on a role this function exits on very different conditions.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    /// Finalize the handshake by creating a `DecryptorWorker` and an `EncryptorWorker`
+    /// Note that `EncryptorWorker` is actually started as an independent worker while
+    /// the `DecryptorWorker` is directly used by this worker to delegate the decryption of messages
+    async fn finalize(
+        &self,
+        context: &Context,
+        handshake_results: HandshakeResults,
+    ) -> Result<DecryptorWorker> {
```
SanjoDeundiak: I don't understand what is the purpose of `DecryptorWorker` now. It doesn't look like worker. Can we refactor it?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,641 @@
...
+    fn encrypted_key_size() -> usize {
+        Self::key_size() + AES_GCM_TAGSIZE_USIZE
+    }
+}
+
+/// The `HandshakeState` contains all the variables necessary to follow the Noise protocol
+#[derive(Debug, Clone)]
+pub(super) struct HandshakeState {
+    pub(super) s: KeyId,
+    e: KeyId,
```
SanjoDeundiak, line 448: When is that key gets deleted?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
SanjoDeundiak: That would also indicate that the caller is responsible for the life cycle of that key. Which is closer to RAII

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/responder_state_machine.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+}
+
+pub struct ResponderStateMachine {
+    common: CommonStateMachine,
+    handshake: Handshake,
+}
+
+impl ResponderStateMachine {
+    delegate! {
```
SanjoDeundiak, line 74: I have one concern regarding that macro: how is it going to look during debugging? Because I experienced few macros that make debugging very hard  and impossible to follow the run flow

SanjoDeundiak (commented):
Great improvement. I left many comments, I'm suspecting that some of them are not related to the actual changes you're doing here, but because a lot of code was moved it looks like it was added in this PR. Can we have the discussion here therefore?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: I did this on purpose because I had the intuition that making a stricter state machine where we truly prevent incorrect states would make the code more complex. I have now attempted to go in that direction with more precise state encoding the exact amount of knowledge that a state machine has: `initial -> k, ck -> re -> re, rs` (and variations of that). 

It indeed makes the code more complicated and IMO harder to review for people who follow the informal algorithm on the side where variables can just be uninitialized. So I propose that we leave the code like this for now.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: Which memory allocation are you referring to, the `to_vec` part? Can you please sketch out a different implementation?

etorreborre: Which memory allocation are you referring to, the `to_vec` part? Can you please sketch out a different implementation?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: This is a private function which has only been introduced for the tests where our test data mentions the ephemeral key

etorreborre: This is a private function which has only been introduced for the tests where our test data mentions the ephemeral key

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: Done

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: Good idea. I've move the generation one level up in the call-stack to make it easier to switch to retrieving a more permanent key in the future.

etorreborre: Good idea. I've move the generation one level up in the call-stack to make it easier to switch to retrieving a more permanent key in the future.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: It comes from the previous implementation. I added a commit to set it to `u64` to conform to the specification

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: Indeed. I simplified the expression now

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: Yes, I adjusted the naming

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: It is used in one of our tests

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: What is the fireproof way to test that?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: Yes that's better. Done.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: Since those 2 fields are not `Option` (because of the previous discussion) it is is not possible. Also we don't yet consume `state` so things have to stay like that. One thing I simplified though was to remove return values for `hkdf` since we have a `mut state` that we can mutate right away to assign the new values for `ck` and `k`.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_state_machine.rs
etorreborre: done.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_state_machine.rs
etorreborre: This code is inside the common code for both state machine. I changed the code so that the payload preparation (and signing) is triggered before we build the `Handshake` struct, right after we generate the static key.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
SanjoDeundiak: Could you then move it nearby the test that uses it inside its `#[cfg(test)]` section?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
SanjoDeundiak: IMHO, this stands against the idea of Rust. I agree that the code may need to become more cluttered in this case, but I see it from a different point of view: you were able to avoid all of that clutter because you relied on assumptions about your code flow. I see multiple problems with that approach:
1. For example, you think there is no need to check if `re` is empty because you know that it should be initialized with a valid value before it's used. But that is a very non-obvious assumption, and the compiler is not able to check. Also, that assumption relies on the assumption about the calling side - because `re` is initialized in `decode_message1` but used from a different function - `encode_message2`.
2. You could be wrong about your assumptions (people make mistakes), which can lead to the usage of uninitialized variables, which is potentially dangerous and unpredictable. Also, the code will be modified sooner or later, and who knows if the person modifying it will be able to easily figure out the assumptions you used and not break them while modifying the code.

I think we should avoid that. If `PublicKey` is a type that is meant to hold a valid public key, we should never break that assumption, in this case, the rust will also guarantee that it stays like that. For example, java will always allow you to put `null` as a value, people used to rely on assumptions in their code and were not checking values for null all the time, which led to `NullPointerException` becoming a meme. In Rust we have the luxury of explicitly distinguishing between variables that could be uninitialized and variables that are guaranteed to be initialized, and we rely on the compiler checking that for us. This is one of the reasons why rust is considered a safe language and one of the key reasons we use it.

On the other hand, I thought that switching to a State machine helps to avoid that problem in the first place because you have a set of states you can be in, and you know what the set of initialized variables in each state is (which also requires some clutter, so we need to evaluate what is better). But it looks like it was the case before that PR (I was not very familiar with the state we were in after my code was modifier but before this PR). So, is that the State machine we wanted to achieve?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
SanjoDeundiak: This is guaranteed to have 1 Heap allocation, while initial implementation could be 1 or 2
```rs
let mut input = Vec::with_capacity(SHA256_SIZE_USIZE + data.len());
input.extend_from_slice(&self.h);
input.extend_from_slice(data);
self.h = Self::sha256(&input);
```

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
SanjoDeundiak: https://doc.rust-lang.org/reference/conditional-compilation.html#target_endian

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_state_machine.rs
SanjoDeundiak: Have you pushed it? I don't get what you mean just from the description

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: Maybe I misunderstood your point. I thought you wanted to test if the sent data was sent by a big or little endian system. You actually want to make sure that the local system is little endian and say that we don't support big endian systems?

SanjoDeundiak: I meant that there is extra work we need to do to make this work correctly on big-endian. Since we don't want to spend time on it right now, a good enough solution would be to not compile on big-endian (which is better than compile but work incorrectly).

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: Done, I replaced uninitialized values with `None` and a proper error message if we were to access such a value.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: Thanks, done.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
etorreborre: Done.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: Moved.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_state_machine.rs
etorreborre: I have pushed the change now: https://github.com/build-trust/ockam/pull/4986/commits/d7bd2ea2b3a7f15f48f5ce1dbcb33f2812e44532

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: To encode it in `message1`. Maybe your question is: should we even have a field for something that is always `vec![]`?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: Done.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
etorreborre: I removed it, it's indeed not needed.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
etorreborre: > How is it guaranteed that we only set it once?

I changed the code to make it so

> And what is the redirect attack in this case?

I actually don't know. @davide-baldo can you please explain that?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
etorreborre: We should but we don't have the machinery to do it properly. This is not something that can be tackled properly in this PR unfortunately.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
etorreborre: Starting a worker starts its relay as a background thread and does not wait for the `initialize` function to be fully executed so if we want to make sure that it is ready to process messages we need to wait for the worker to be ready (which is what `wait_for` does)

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/responder_state_machine.rs
etorreborre: I didn't have any issue with that one while debugging. Let's monitor the situation and revert if it turns out to be problematic.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
etorreborre: It's not a worker anymore and it has been renamed to `DecryptorHandler` (there is still a `Decryptor` mirroring the `Encryptor`)

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
etorreborre: > This is unexpected for me that HandshakeWorker::create() would wait for the handshake to complete

If you don't wait you can't start sending messages right away because they won't reach the `EncryptorWorker`

> Also that depending on a role this function exits on very different conditions

I agree. I also think that the part about waiting for the decryptor worker is not even working. Because at the end of the Handshake there's no registered decryptor worker. 

I replaced that with a waiting condition only for the initiator, so that it knows when to send messages:

 - wait for the handshake to finish
 - wait for the encryptor worker to be ready

On the responder side the situation is less problematic since a message to be decrypted will always be queued after the last message in the handshake and will be processed by the decryptor handler which is not a worker (so there's nothing to synchronize)

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: I am not deleting ephemeral keys at the end of the handshake.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
SanjoDeundiak: I still don't get it. The comment says "wait until the worker is ready to receive messages".
If the purpose is "we can send messages to that worker", then the fact that `ctx.start_worker()` succeeded should be enough.
If the purpose is "the worker finished its initialization", then I don't get why we would need to wait for that here

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
SanjoDeundiak: But we should, right?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/mod.rs
```diff
@@ -0,0 +1,11 @@
+mod error;
+
+/// This directive makes sure that we only run the handshake protocol if it has been compiled
+/// on a little endian system since it is not supporting a big endian one at the moment
+#[cfg(target_endian = "little")]
```
SanjoDeundiak: I think it could be a descriptive compilation error instead of a "not found" error we'll get because this module just won't be included during the compilation. But that's fine if you don't want to spend time fixing it

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
etorreborre: I am *now* :-) (sorry for the very confusing typo in that kind of message)

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/mod.rs
etorreborre: Yes I agree this is not great. Is there an easy way to put a custom compilation error message in that case?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
etorreborre: I changed the code for just waiting on the initiator side to make sure that the encryptor is started: https://github.com/build-trust/ockam/pull/4986/files#diff-dc72819bff455b642637545345ddbeee20dd15ee61c15a899d411266bca85bfeR211

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/mod.rs
SanjoDeundiak: Sure,
```
#[cfg(not(target_endian = "little"))]
compile_error!("Key Exchange is only supported on little-endian machines");
```

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
```diff
@@ -1,13 +1,122 @@
...
+                );
+                Ok(())
+            }
+        }
+    }
+}
+
 pub(crate) struct Decryptor {
     current_key: KeyId,
     current_key_nonce: u64,
```
SanjoDeundiak, line 122: I think you're just moving the code, so it's okay if you want to ignore my comments here, but I'll write my feedback here, we can move it elsewhere if needed

SanjoDeundiak: The name `current_key_nonce` is misleading. It's not nonce at all, it can be called `number of rekys` or `epoch` or whatever

SanjoDeundiak: Maybe it can be renamed to something like `last_rekey_nonce`

SanjoDeundiak: Even better if we can could use `number_of_rekeys`, the math should be tweaked a little bit though

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
```diff
@@ -18,6 +127,16 @@ pub(crate) struct Decryptor {
 }
 
 impl Decryptor {
+    pub fn new(key: KeyId, vault: Arc<dyn XXInitializedVault>) -> Self {
+        Self {
+            current_key: key,
```
SanjoDeundiak: Is this key destroyed when the secure channel is shutdown?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/mod.rs
etorreborre: Done, thanks

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/initiator_state_machine.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn get_handshake_results(&self) -> Option<HandshakeResults> {
+        self.make_handshake_results(self.get_handshake_keys())
+    }
+}
+
+/// Implementation of the state machine actions, delegated to the Handshake module
+pub(super) struct InitiatorStateMachine {
+    pub(super) common: CommonStateMachine,
+    pub(super) handshake: Handshake,
+    pub(super) identity_payload: Vec<u8>,
```
SanjoDeundiak, line 70: It also includes credentials, right? Should we describe that in the name somehow? Or maybe in a doc-comment?

SanjoDeundiak: Also includes the static key signature I guess

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_state_machine.rs
SanjoDeundiak: I think it will be moved much further out in the future, but we should not bother with it right now

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -0,0 +1,344 @@
...
+                Err(IdentityError::UnknownChannelMsgDestination.into())
+            };
+            return result;
+        };
+
+        // set the remote route by taking the most up to date message return route
+        // In the case of the initiator the first return route mentions the secure channel listener
+        // address so we need to wait for the return route corresponding to the remote handshake worker
+        // when it has been spawned
+        self.remote_route = Some(message.return_route());
```
SanjoDeundiak: Can we update the remote_route only if the current message handling succeeds?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -0,0 +1,344 @@
...
+            .start(context)
+            .await?;
+
+        let decryptor_remote = addresses.decryptor_remote.clone();
+        debug!(
+            "Starting SecureChannel {} at remote: {}",
+            role.clone(),
+            &decryptor_remote
+        );
+        trace!(
```
SanjoDeundiak: I think this should be either removed or moved down to where the channel is already initialized

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -0,0 +1,344 @@
...
+                    identifier.clone(),
+                    credentials.clone(),
+                    trust_policy.clone(),
+                    trust_context.clone(),
+                )
+                .await?,
+            )
+        };
+
+        let (mut callback_waiter, callback_sender) = ockam_node::callback::new_callback();
```
SanjoDeundiak, line 184: I would like to attempt to change that to oneshot channel in my following PR

SanjoDeundiak: Also a little bit concerned with the fact that it's not used for the responder, but it should theoretically be fine. The receiver is just discarded.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -0,0 +1,344 @@
...
+        if role.is_initiator() {
+            // wait until the handshake is finished
+            if let Some(timeout) = timeout {
+                callback_waiter.receive_timeout(timeout).await?;
+            } else {
+                callback_waiter.receive().await?;
+            }
+
+            // make sure that the encryptor has finished initializing before sending messages
+            context.wait_for(addresses.encryptor.clone()).await?;
```
SanjoDeundiak: Sorry, but I still don't get why it's needed. What if we remove that line? What will happen?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,705 @@
...
+    k: Option<KeyId>,
+    re: Option<PublicKey>,
+    pub(super) rs: Option<PublicKey>,
+    n: u64,
+    h: [u8; SHA256_SIZE_USIZE],
+    ck: Option<KeyId>,
+    message1_payload: Vec<u8>,
```
SanjoDeundiak: What I didn't understand is why for m2 and m3, payload is passed to the `encode_message_n` function as an argument, while payload1 should be a field in the struct

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake.rs
```diff
@@ -0,0 +1,705 @@
...
+        self.k.clone().ok_or_else(|| {
+            Error::new(
+                Origin::KeyExchange,
+                Kind::Invalid,
+                "key id k should have been set",
+            )
+        })
+    }
+
+    pub(super) fn ck(&self) -> Result<KeyId> {
```
SanjoDeundiak, line 500: Did you try returning `Result<&KeyId>`? Was it tricky?

SanjoDeundiak (approved):

PR #4997, opened 2023-05-25
feat(rust): print warning when starting `Hop` service

SanjoDeundiak:


etorreborre (approved):

implementations/rust/ockam/ockam_command/src/service/start.rs
```diff
@@ -145,6 +147,12 @@ async fn run_impl(
         addr.color(OckamColor::PrimaryResource.color())
     ))?;
 
+    if is_hop_service {
+        opts.terminal.write_line(&fmt_warn!(
+            "SECURITY WARNING: Don't use Hop service in production nodes"
```
etorreborre, line 152: I think we should explain why.

SanjoDeundiak: I haven't found a way of explaining it in a very short sentence, and I didn't want to have a long text printed as an output of a successfully executed command :)

PR #4998, opened 2023-05-25
feat: extend direct authenticator to allow list and delete members, add Elixir client

hairyhum:
## Proposed changes

Add new APIs:
- list_members: `get /` will return members enrolled by the calling identity
  returns a map of IdentityIdentifier => AttributesEntry
- delete_member: `delete /<identity_id>` deletes member if it was enrolled
  by the calling identity

Add client for these APIs to Elixir implementation

NOTE: this is not full CRUD because we cannot overwrite attributes (not sure why it was done like that), but getting close.

NOTE: AttributesEntry returns attribute values as lists of numbers,
Elixir implementation will still convert them to binaries.

## Checks

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/credential/authenticator/direct/client.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/credential/authenticator/direct/client.ex
@@ -1,0 +1,121 @@
+defmodule Ockam.Credential.Authenticator.Direct.Client.AddMemberRequest do
+  @moduledoc """
+  Request data structure for Credential Authority authenticator API add_member
+  """
+  use TypedStruct
+
+  typedstruct do
+    plugin(Ockam.TypedCBOR.Plugin)
+    field(:identity_id, String.t(), minicbor: [key: 1])
+    field(:attributes, %{String.t() => String.t()}, minicbor: [key: 2])
+  end
+end
+
+defmodule Ockam.Credential.Authenticator.Direct.Client.AttributesEntry do
+  @moduledoc """
+  Data structure representing attribute entry for Credential Authority authenticator API list_members
+  """
+  use TypedStruct
+
+  typedstruct do
+    plugin(Ockam.TypedCBOR.Plugin)
+    ## Rust encodes attribute values as a list of bytes
+    field(:attributes, %{String.t() => {:list, :integer}}, minicbor: [key: 1])
+    field(:added_at, integer(), minicbor: [key: 2])
+    field(:expires, integer() | nil, minicbor: [key: 3])
+    field(:attested_by, String.t() | nil, minicbor: [key: 4])
+  end
+end
+
+defmodule Ockam.Credential.Authenticator.Direct.Client.ListMembersResponse do
+  @moduledoc """
+  Response data structure for Credential Authority authenticator API list_members
+
+  This API returns a map of identity identifiers to attribute entries
+  """
+
+  alias Ockam.Credential.Authenticator.Direct.Client.AttributesEntry
+
+  @schema {:map, :string, AttributesEntry.minicbor_schema()}
+
+  def decode_strict(data) do
+    with {:ok, decoded} <- Ockam.TypedCBOR.decode_strict(@schema, data) do
+      ## Convert attribute values to binaries
+
+      {:ok,
+       Map.new(decoded, fn {id, entry} ->
+         attributes =
+           Map.get(entry, :attributes, %{})
+           |> Map.new(fn {key, val} -> {key, :erlang.list_to_binary(val)} end)
+
+         {id, Map.put(entry, :attributes, attributes)}
+       end)}
+    end
+  end
+end
+
+defmodule Ockam.Credential.Authenticator.Direct.Client do
+  @moduledoc """
+  API client for Credential Authority authenticator service
+
+  Allows to add new members to credential authority
+  """
+
+  alias Ockam.API.Client, as: ApiClient
+  alias Ockam.API.Response, as: ApiResponse
+
+  alias Ockam.Credential.Authenticator.Direct.Client.AddMemberRequest
+  alias Ockam.Credential.Authenticator.Direct.Client.AttributesEntry
+  alias Ockam.Credential.Authenticator.Direct.Client.ListMembersResponse
+
+  @spec add_member(String.t(), %{String.t() => String.t()}, Ockam.Address.route()) ::
+          :ok | {:error, any()}
+  def add_member(identity_id, attributes, api_route) do
+    request =
+      AddMemberRequest.encode!(%AddMemberRequest{
+        identity_id: identity_id,
+        attributes: attributes
+      })
+
+    case ApiClient.sync_request(:post, "/", request, api_route) do
+      {:ok, %ApiResponse{status: 200}} ->
+        :ok
+
+      {:ok, %ApiResponse{status: status, body: body}} ->
+        {:error, {:api_error, status, body}}
+
+      {:error, _} = error ->
+        error
+    end
+  end
+
+  @spec list_members(Ockam.Address.route()) ::
+          {:ok, %{String.t() => AttributesEntry.t()}} | {:error, any()}
+  def list_members(api_route) do
+    with {:ok, %ApiResponse{status: 200, body: response}} <-
+           ApiClient.sync_request(:get, "/", "", api_route),
+         {:ok, members} <- ListMembersResponse.decode_strict(response) do
+      {:ok, members}
+    else
+      {:ok, %ApiResponse{status: status, body: body}} ->
+        {:error, {:api_error, status, body}}
+
+      {:error, _} = error ->
+        error
+    end
+  end
+
+  @spec delete_member(String.t(), Ockam.Address.route()) :: :ok | {:error, any()}
+  def delete_member(identity_id, api_route) do
+    case ApiClient.sync_request(:delete, "/#{identity_id}", "", api_route) do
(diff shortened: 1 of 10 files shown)


polvorin (dismissed):

implementations/elixir/ockam/ockam_typed_cbor/lib/typed_cbor.ex
```diff
@@ -150,8 +150,8 @@ defmodule Ockam.TypedCBOR do
 
   def from_cbor_term(:term, any), do: any
 
-  def from_cbor_term(schema, _) do
-    raise(Exception, "type mismatch, expected schema #{inspect(schema)}")
+  def from_cbor_term(schema, data) do
+    raise(Exception, "type mismatch, expected schema #{inspect(schema)}, data: #{inspect(data)}")
```
shanesveller, line 156: Mild caution that if `data` contains anything sensitive in a non-obfuscated form, this can leak that information to log output, exception tracking, etc. If we hit this function head that should be a truly exceptional case, though.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -190,6 +193,21 @@ impl DirectAuthenticator {
...
         self.attributes_writer.put_attributes(id, entry).await
     }
+
+    async fn list_members(
+        &self,
+        enroller: &IdentityIdentifier,
+    ) -> Result<HashMap<IdentityIdentifier, AttributesEntry>> {
+        // TODO: move filter to `list` function
+        let all_attributes = self.attributes_reader.list().await?;
+        let attested_by_me: HashMap<IdentityIdentifier, AttributesEntry> = all_attributes
```
shanesveller: Since the value must match the function's return type, you can generally elide the inner generic type hints with:
```suggestion
        let attested_by_me: HashMap<_, _> = all_attributes
```

In this instance you could even omit the type hint of HashMap altogether.
```suggestion
        let attested_by_me = all_attributes
```

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -211,13 +229,34 @@ impl Worker for DirectAuthenticator {
...
-                (Some(Method::Post), "/") | (Some(Method::Post), "/members") => {
+            let path_segments = req.path_segments::<5>();
+            let res = match (req.method(), path_segments.as_slice()) {
+                (Some(Method::Post), [""]) | (Some(Method::Post), ["members"]) => {
                     let add: AddMember = dec.decode()?;
                     self.add_member(&from, add.member(), add.attributes())
                         .await?;
                     Response::ok(req.id()).to_vec()?
                 }
+                // List members attested by out identity (enroller)
```
shanesveller:
```suggestion
                // List members attested by our identity (enroller)
```

implementations/elixir/ockam/ockam/lib/ockam/credential/authenticator/direct/client.ex
```diff
@@ -0,0 +1,121 @@
...
+  def add_member(identity_id, attributes, api_route) do
+    request =
+      AddMemberRequest.encode!(%AddMemberRequest{
+        identity_id: identity_id,
+        attributes: attributes
+      })
+
+    case ApiClient.sync_request(:post, "/", request, api_route) do
+      {:ok, %ApiResponse{status: 200}} ->
+        :ok
```
shanesveller, line 82: Does it make any sense for this branch to return the revised list of members as part of its response body? I know that's implemented elsewhere if we chose to pursue it.

implementations/elixir/ockam/ockam/lib/ockam/credential/authenticator/direct/client.ex
```diff
@@ -0,0 +1,121 @@
...
+
+  def decode_strict(data) do
+    with {:ok, decoded} <- Ockam.TypedCBOR.decode_strict(@schema, data) do
+      ## Convert attribute values to binaries
+
+      Enum.map(decoded, fn {id, entry} ->
+        attributes =
+          Map.get(entry, :attributes, %{})
+          |> Enum.map(fn {key, val} -> {key, :erlang.list_to_binary(val)} end)
+          |> Map.new()
```
shanesveller: This is directly equivalent:
```suggestion
          |> Map.new(fn {key, val} -> {key, :erlang.list_to_binary(val)} end)
```

implementations/elixir/ockam/ockam/lib/ockam/credential/authenticator/direct/client.ex
```diff
@@ -0,0 +1,121 @@
...
+
+        {id, Map.put(entry, :attributes, attributes)}
+      end)
+      |> Map.new()
```
shanesveller: This one likewise can be rolled up from `Enum.map |> Map.new/1` to just `Map.new/2`.

https://hexdocs.pm/elixir/Map.html#new/2

implementations/elixir/ockam/ockam/lib/ockam/credential/authenticator/direct/client.ex
```diff
@@ -0,0 +1,121 @@
...
+defmodule Ockam.Credential.Authenticator.Direct.Client.AttributesEntry do
+  @moduledoc """
+  Data structure representing attribute entry for Credential Authority authenticator API list_members
+  """
+  use TypedStruct
+
+  typedstruct do
+    plugin(Ockam.TypedCBOR.Plugin)
+    ## Rust encodes attribute values as a list of bytes
+    field(:attributes, %{String.t() => {:list, :integer}}, minicbor: [key: 1])
```
shanesveller, line 23: Is there anywhere here or outside of this changed code that needs to ensure we convert a map that has atom keys instead of string keys?

implementations/elixir/ockam/ockam_typed_cbor/lib/typed_cbor.ex
hairyhum: Yeah. I think I'll roll it back and add the logs instead.

implementations/elixir/ockam/ockam/lib/ockam/credential/authenticator/direct/client.ex
hairyhum: AddMember API doesn't return them and I didn't want to change the existing APIs.

implementations/elixir/ockam/ockam/lib/ockam/credential/authenticator/direct/client.ex
hairyhum: For a second I was hoping that there is a support for `:maps.map` finally, but it's not.

implementations/elixir/ockam/ockam/lib/ockam/credential/authenticator/direct/client.ex
hairyhum: Can you clarify the question? The keys are not converted anywhere, attributes should always assume string keys.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
hairyhum: Pretty sure compiler was complaining last time I did that. It doesn't anymore though.

implementations/elixir/ockam/ockam/lib/ockam/credential/authenticator/direct/client.ex
shanesveller: `decode_strict` below for example doesn't have any specific code, guards, or typespec about the keys definitely being strings. I had to read into the TypedCBOR implementation to learn that's what L39 `@schema {:map, :string, ...}` ensures.

https://github.com/build-trust/ockam/blob/94a8afb946dbb6c4f71dd47c112bcb89280270c8/implementations/elixir/ockam/ockam_typed_cbor/lib/typed_cbor.ex#L137-L141

I'm all set, I think!

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
shanesveller: Most likely once you added `Ok(attested_by_me)` it was able to successfully do inference without hinting, but couldn't tell how the value would be used up until that line was added.

shanesveller (approved):
LGTM, but you'll still need someone else from approvers to take a look.

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
```diff
@@ -211,13 +229,34 @@ impl Worker for DirectAuthenticator {
...
+                (Some(Method::Post), [""]) | (Some(Method::Post), ["members"]) => {
                     let add: AddMember = dec.decode()?;
                     self.add_member(&from, add.member(), add.attributes())
                         .await?;
                     Response::ok(req.id()).to_vec()?
                 }
+                // List members attested by our identity (enroller)
+                (Some(Method::Get), [""]) | (Some(Method::Get), ["members"]) => {
+                    let entries = self.list_members(&from).await?;
+                    Response::ok(req.id()).body(entries).to_vec()?
```
polvorin, line 243: at some point we will need to support either bigger packets or reassembly,  as it's now response sizes are limited by the max ockam packet size

polvorin (approved):

implementations/rust/ockam/ockam_api/src/authenticator/direct.rs
hairyhum: Yes. Makes sense.

PR #5018, opened 2023-05-29
refactor(rust): remove old config.json file and add migration

adrianbenavides:
The `config.json` file in the root $OCKAM_HOME directory contained a map with data about spaces and projects. This data was duplicated in the `projects` directory and the CLI code was not keeping these two different configurations synchronized correclty.

A new cli state has been added for spaces and a new migration is in place to correctly move the space and project lookups into the new `spaces` and `projects` directories within the CliState.


implementations/rust/ockam/ockam_command/src/node/util.rs
```diff
@@ -65,7 +65,17 @@ pub async fn start_embedded_node_with_vault_and_identity(
     let options = TcpListenerOptions::new();
     let listener = tcp.listen(&bind, options).await?;
 
-    let projects = cfg.inner().lookup().projects().collect();
+    let projects = {
```
etorreborre: This code seems to be duplicated with [this one](https://github.com/build-trust/ockam/pull/5018/files#diff-afac108b9e4f2377c7932bef242708078b72d74836268ccc0ff4a09a24f67a53R273). Can we refactor it?

etorreborre (dismissed):
Nice job! 🏆

implementations/rust/ockam/ockam_command/src/node/util.rs
adrianbenavides: Done!

etorreborre (approved):

PR #5025, opened 2023-05-30
docs(rust): document the layout of files for a node

etorreborre:
This PR provides an overview of all the data stored to disk when using the `ockam_api` crate with a `NodeManager`.

### DIFF
--- a/implementations/rust/ockam/ockam_api/README.md
+++ b/implementations/rust/ockam/ockam_api/README.md
@@ -11,6 +11,123 @@ and trustfully with cloud services and other devices.
 This crate supports the creation of a fully-featured Ockam Node
 (see [`NodeManager`](https://github.com/build-trust/ockam/blob/2fc6d7714a4e54f8734c172ad6480fedc6e3629c/implementations/rust/ockam/ockam_api/src/nodes/service.rs#L87) in [`src/nodes/service.rs`](https://github.com/build-trust/ockam/blob/2fc6d7714a4e54f8734c172ad6480fedc6e3629c/implementations/rust/ockam/ockam_api/src/nodes/service.rs)).
 
+## Configuration
+
+A `NodeManager` maintains its configuration as a list of directories and files stored under
+the `OCKAM_HOME` directory (`~/.ockam`) by default:
+```shell
+root
+├─ credentials
+│  ├─ c1.json
+│  ├─ c2.json
+│  └─ ...
+├─ defaults
+│  ├── credential -> ...
+│  ├── identity -> ...
+│  ├── node -> ...
+│  └── vault -> ...
+├─ identities
+│  ├─ data
+│  │  ├─ authenticated-storage.lmdb
+│  │  └─ authenticated-storage.lmdb-lock
+│  ├─ identity1.json
+│  ├─ identity2.json
+│  └─ ...
+├─ nodes
+│  ├─ node1
+│  │  ├─ default_identity -> ...
+│  │  ├─ default_vault -> ...
+│  │  ├─ policies-storage.lmdb
+│  │  ├─ policies-storage.lmdb-lock
+│  │  ├─ setup.json
+│  │  ├─ stderr.log
+│  │  ├─ stdout.log
+│  │  └─ version
+│  ├─ node2
+│  └─ ...
+├─ projects
+│  └─ default.json
+├─ trust_contexts
+│  └─ default.json
+└─ vaults
+   ├─ vault1.json
+   ├─ vault2.json
+   ├─ ...
+   └─ data
+      ├─ vault1.lmdb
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_api/README.md
```diff
@@ -11,6 +11,123 @@ and trustfully with cloud services and other devices.
...
+      ├─ vault1.lmdb
+      ├─ vault1.lmdb-lock
+      ├─ vault2.lmdb
+      ├─ vault2.lmdb-lock
+      └─ ...
+```
+## `credentials`
+
+Each file stored under the `credentials` directory contains the credential for a given identity.
+Those files are created with the `ockam credential store` command. They are then read during the creation of
```
hairyhum, line 66: In the use-case examples we retrieve credentials via `ockam project authenticate/enroll` using enrollment tokens. Are those credentials also saved here?

caoakleyii (dismissed):
Lgtm

implementations/rust/ockam/ockam_api/README.md
```diff
@@ -11,6 +11,123 @@ and trustfully with cloud services and other devices.
...
+## `defaults`
+
+This directory contains symlinks to other files or directories in order to specify which node,
+identity, credential or vault must be considered as a default when running a command expecting those
+inputs
+
+## `identities`
+
+This directory contains one file per identity and a data directory. An identity file is created
+with the `ockam identity create` command or created by default for some commands (in that case the
```
hairyhum, line 78: Would identities created by the identity service be also stored here? If not are they stored anywhere?

implementations/rust/ockam/ockam_api/README.md
```diff
@@ -11,6 +11,123 @@ and trustfully with cloud services and other devices.
...
+`defaults/identity` symlink points to that identity). The identity file contains:
+
+- the identity identifier
+- the enrollment status for that identity
+
+The `data` directory contains a LMDB database with other information about identities:
```
hairyhum, line 84: Is that information shared by all nodes? Which nodes read or write this LMDBs?

implementations/rust/ockam/ockam_api/README.md
```diff
@@ -11,6 +11,123 @@ and trustfully with cloud services and other devices.
...
+Note: for each `.lmdb` file there is a corresponding `lmdb-lock` file which is used to control
+the exclusive access to the LMDB database even if several OS processes are trying to modify it.
+
+## `nodes`
+
+This directory contains:
+
+ - symlinks to default values for the node: identity and vault
+ - a database for ABAC policies
+ - a setup file containing some configuration information for the node (is it an authority node?, what is the TCP listener address?,...)
```
hairyhum: Is this file generated by the node? Who writes and reads this file?

implementations/rust/ockam/ockam_api/README.md
```diff
@@ -11,6 +11,123 @@ and trustfully with cloud services and other devices.
...
+ - log files: for system errors and system outputs. The stdout.log file is where almost all the node logs are written
+ - a version number for the configuration
+
+## `projects`
+
+This directory contains a list of files, one per project that was created, either the default project
```
hairyhum, line 113: If we start a node in the environment which has project configurations, does the node have access to these configurations?

hairyhum (changes requested):
I think this also needs some information about which process, command or node creates and reads which files to understand the lifecycle of the files.

implementations/rust/ockam/ockam_api/README.md
etorreborre: No they are only stored when we do `ockam credential store`

implementations/rust/ockam/ockam_api/README.md
etorreborre: This database is shared by all the nodes which are started using the same `OCKAM_HOME`. All nodes can read and write this database, for example when they create a secure channel and store the other party identity.

implementations/rust/ockam/ockam_api/README.md
etorreborre: The identity service would store an identity directly in the shared database and no file would be created.

implementations/rust/ockam/ockam_api/README.md
etorreborre: This information is created when we create the node (regular or authority). It is read again if we restart the node.

implementations/rust/ockam/ockam_api/README.md
etorreborre: When you start a node you either specify nothing and that node uses the default project or you specify one of the project configurations.

hairyhum (approved):

PR #5037, opened 2023-05-31
feat(rust): display the source location of an error by default

etorreborre:
This PR improves the default display of an error to help diagnose issues faster.

# Diagnostic

For example [this discussion](https://github.com/build-trust/ockam/discussions/5016) shows an error which is raised within a `WorkerRelay`.  The error being displayed just mentioned that an operation failed for a given `Address` without telling us what went wrong exactly (a worker was already started at that address).

# Fix

The `Display` instance for an `Error` has been expanded (at least under the `std` flag) to show:

 - a message for the error cause
 - the origin and kind of the error
 - a source location

# Test

Two unit tests have been added for the display.

### DIFF
--- a/implementations/rust/ockam/ockam_core/src/error/code.rs
+++ b/implementations/rust/ockam/ockam_core/src/error/code.rs
@@ -408,13 +408,10 @@ impl From<u8> for Kind {
 
 impl core::fmt::Display for ErrorCode {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        // Kind of halfway between debug and display, TBH, but it's only used in
-        // the Error debug output.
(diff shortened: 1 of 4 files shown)


SanjoDeundiak (dismissed):

examples/rust/file_transfer/tests/100_32_test.bin
```diff
@@ -0,0 +1,2 @@
+�u���P��/f�
```
SanjoDeundiak: @etorreborre 😃

etorreborre: ach....

etorreborre: Fixed. Not a big fan of tests writing to the source directory. I'll come back to that at some stage

SanjoDeundiak (approved):

PR #5040, opened 2023-06-01
feat(rust): remove `force` flag from `node start` command

adrianbenavides:
We don't have a clear use-case for this flag, and it's making the command more complicated than necessary.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/start.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/start.rs
@@ -1,5 +1,6 @@
 use clap::Args;
 
+use colorful::Colorful;
 use ockam::TcpTransport;
 use ockam_api::cli_state::{StateDirTrait, StateItemTrait};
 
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_command/src/node/start.rs
```diff
@@ -45,12 +43,15 @@ async fn run_impl(
...
-        println!(
-            "Restart aborted, node: {} already running",
-            node_state.name()
-        );
+    // Abort if node is already running
+    if node_state.is_running() {
+        opts.terminal
+            .stdout()
+            .plain(fmt_err!(
+                "Restart aborted, node '{}' already running",
```
etorreborre: Maybe we could say that the node is already running and that the user should call `ockam node stop` then `ockam node start` if they want to restart the node?

etorreborre (dismissed):

implementations/rust/ockam/ockam_command/src/node/start.rs
adrianbenavides: Something like this?

> The node '{}' is already running. If you want to restart it you can call `ockam node stop {}` and then `ockam node start {}`

etorreborre: yes

etorreborre (approved):

PR #5043, opened 2023-06-01
fix(rust): `project create` when creating a new project

adrianbenavides:
This PR fixes two problems

#### 1. "project already exists" when running `ockam enroll`

The `check_project_readiness` function was initializing the project config file. Later, when persisting the fully populated config file, it was failing because it expected that file to not exist at that point, returning an `AlreadyExist` error (note the `create` vs `overwrite` calls in the changes).

#### 2. "space not found" when running `ockam project create` right after a fresh `ockam enroll`

After #5018, we stopped using the legacy config.json file, which contained data about spaces and projects. I forgot to update the code that should persist the space that is created within the `ockam enroll` command.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/enroll/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/enroll/mod.rs
@@ -197,7 +197,9 @@ async fn default_space<'a>(
         ))?;
         space
     };
-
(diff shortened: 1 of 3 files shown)


caoakleyii (approved):
LGTM; and fix worked locally

Issue #4952, opened 2023-05-19
Allow to store multiple credentials for a single identity


davide-baldo:
### Current behavior

When multiple credentials are present from a single identity, only the last one is kept stored.

### Desired behavior

Multiple credentials can be presented and are kept stored and used by other party.

PR #5044, opened 2023-06-01
fix(rust): persist credential after generating it in `project enroll` command

adrianbenavides:


implementations/rust/ockam/ockam_command/src/project/enroll.rs
```diff
@@ -177,6 +176,14 @@ async fn run_impl(
         .overwrite(&project.name, project.clone().try_into()?)?;
 
     let credential = client2.credential().await?;
+    let identity = identities()
+        .identities_creation()
+        .decode_identity(authority.identity())
+        .await?;
+    opts.state.credentials.overwrite(
```
polvorin, line 184: is this  per-identity?,    I mean, if the node has two identities,   credential for `i1`  would not override credential for `i2` right?

adrianbenavides: It's storing the credential using the project's authority identity as the issuer and the project's name as the file name (kind of the "unique id" in the file system). Should we use another name for the file to avoid false duplicates? (the identity name as you mentioned perhaps?)

davide-baldo: Maybe we should wait until this is implemented
https://github.com/build-trust/ockam/issues/4952

etorreborre:
@adrianbenavides I think that this PR can now be closed given all the recent changes + the new work around trust contexts removal.

PR #5048, opened 2023-06-02
feat(rust): add delete and list subcommands for kafka consumer/producer commands

adrianbenavides:
The `show` commands in this case doesn't make too much sense, as we would be returning the service address, which is the command main argument. This command would make sense if, in the future, we store more data about the kafka services in the NodeManager registry.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/lib.rs
+++ b/implementations/rust/ockam/ockam_api/src/lib.rs
@@ -171,6 +171,29 @@ impl DefaultAddress {
     pub const KAFKA_CONSUMER: &'static str = "kafka_consumer";
     pub const KAFKA_PRODUCER: &'static str = "kafka_producer";
     pub const RPC_PROXY: &'static str = "rpc_proxy_service";
+
+    pub fn is_valid(name: &str) -> bool {
+        matches!(
+            name,
+            Self::IDENTITY_SERVICE
+                | Self::AUTHENTICATED_SERVICE
+                | Self::FORWARDING_SERVICE
+                | Self::UPPERCASE_SERVICE
+                | Self::ECHO_SERVICE
+                | Self::HOP_SERVICE
+                | Self::CREDENTIALS_SERVICE
+                | Self::SECURE_CHANNEL_LISTENER
+                | Self::DIRECT_AUTHENTICATOR
+                | Self::CREDENTIAL_ISSUER
+                | Self::ENROLLMENT_TOKEN_ISSUER
+                | Self::ENROLLMENT_TOKEN_ACCEPTOR
+                | Self::VERIFIER
+                | Self::OKTA_IDENTITY_PROVIDER
+                | Self::KAFKA_CONSUMER
+                | Self::KAFKA_PRODUCER
+                | Self::RPC_PROXY
+        )
+    }
 }
 
 pub mod actions {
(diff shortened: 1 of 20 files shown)


implementations/rust/ockam/ockam_api/src/lib.rs
```diff
@@ -171,6 +171,29 @@ impl DefaultAddress {
     pub const KAFKA_CONSUMER: &'static str = "kafka_consumer";
     pub const KAFKA_PRODUCER: &'static str = "kafka_producer";
     pub const RPC_PROXY: &'static str = "rpc_proxy_service";
+
+    pub fn is_valid(name: &str) -> bool {
+        matches!(
```
etorreborre, line 176: I learned a new macro today :-)

implementations/rust/ockam/ockam_api/src/nodes/service/node_services.rs
```diff
@@ -768,6 +768,31 @@ impl NodeManagerWorker {
...
+        dec: &mut Decoder<'_>,
+        kind: KafkaServiceKind,
+    ) -> Result<ResponseBuilder> {
+        let body: DeleteServiceRequest = dec.decode()?;
+        let address = body.address();
+        let mut node_manager = self.node_manager.write().await;
+        let res = match node_manager.registry.kafka_services.remove(&address) {
+            None => Response::not_found(req.id()),
+            Some(e) => {
+                if !matches!(e.kind(), _kind) {
```
etorreborre: Shouldn't you test for equality here since matching against a pattern with a variable will always succeed?
My other comment would be that it is a bit weird to accept the operation, do the deletion, and then return an error. We should better do the check before deleting the worker.

implementations/rust/ockam/ockam_command/src/kafka/producer/create.rs
```diff
@@ -86,7 +86,7 @@ async fn rpc(ctx: Context, (opts, cmd): (CommandGlobalOpts, CreateCommand)) -> c
 
     let msgs = vec![
         format!(
-            "Buildling KafkaConsumer service {}",
+            "Buildling KafkaProducer service {}",
```
etorreborre: Good catch, and there is another one next to it: s/Buildling/Building :-)

implementations/rust/ockam/ockam_command/src/kafka/producer/delete.rs
```diff
@@ -0,0 +1,52 @@
+use crate::node::{get_node_name, initialize_node_if_default};
+use crate::util::{node_rpc, parse_node_name, Rpc};
+use crate::{docs, fmt_ok, node::NodeOpts, CommandGlobalOpts};
+use clap::Args;
+use colorful::Colorful;
+use ockam_api::nodes::models;
+use ockam_core::api::Request;
+
+const AFTER_LONG_HELP: &str = include_str!("./static/delete/after_long_help.txt");
```
etorreborre, line 9: I wonder if we could avoid mistakes by having a macro that is even smarter and:

 - takes the name of the current module to build the path
 - errors out (at compile time) if the path does not exist

implementations/rust/ockam/ockam_command/src/kafka/producer/create.rs
```diff
@@ -86,7 +86,7 @@ async fn rpc(ctx: Context, (opts, cmd): (CommandGlobalOpts, CreateCommand)) -> c
 
     let msgs = vec![
         format!(
-            "Buildling KafkaConsumer service {}",
```
etorreborre, line 89: I am wondering if we could refactor the whole consumer/producer commands to just be one module where everything is parametrized by the service kind. Because those too service are actually almost identical. Then any modification to the commands would always stay consistent

etorreborre (commented):
@adrianbenavides I think we need to revise the delete behavior. What do you think?

implementations/rust/ockam/ockam_command/src/kafka/producer/delete.rs
adrianbenavides: I like that idea! It's definitely worth it, I've lost the count of how many times I've imported the wrong file. I'll open an issue for this.

implementations/rust/ockam/ockam_api/src/nodes/service/node_services.rs
adrianbenavides: Not sure if I'm following your suggestion but here is my thought process. The `kafka_services` registry stores the addresses and their associated kafka services type (`KafkaServiceKind`), so I can't check if the services types match until after I retrieve an item from the registry. Once I have the item I was looking for, all I have to compare is the type.

Although, I saw a problem with the code: I'm not returning the address back to the registry in case of an error. I'll fix that too, but let me know if I got your point!

implementations/rust/ockam/ockam_command/src/kafka/producer/create.rs
adrianbenavides: I was checking this and I don't see why we couldn't do it. I'll create a `good-first` issue with the details!

implementations/rust/ockam/ockam_api/src/nodes/service/node_services.rs
etorreborre: Returning the address back is indeed one way to do it, but it feels a bit hacky to me. 
An extension method on the kafka_registry would be better IMO:

 - instead of calling `remove` you call `remove_consumer`
 - then the implementation of `remove_consumer` gets the info at the requested address and checks if this is really a consumer before removing it

That does not even have to be an extension method, just a get / check / delete (in case of a successful check) sequence, instead of delete / check / restore (in case of a failed check) sequence.

adrianbenavides: > That does not even have to be an extension method, just a get / check / delete (in case of a successful check) sequence, instead of delete / check / restore (in case of a failed check) sequence.

- `get / check / delete`: implies two lookups in the best case scenario (we remove the entry); and one in the worst case (the entry type doesn't match)
- `delete / check / restore`: the other way around (1 lookup in the best case, 2 in the worst)

In this case I opted for the second case because it favors the most common scenario (the user calls the right command to delete a given kafka service) at the expense of having a little weirder logic.

What are your thoughts on this? I know it's an unnecessary optimization, but doing it doesn't change the logic too much either...

etorreborre: I definitely prefer the first kind of logic because it prevents inconsistencies in general. I don't think this matters tremendously in that specific case but in general it does. If something has been removed when it shouldn't have been other processes might rely on that thing to exist.

The other alternative is to have one registry for producers and one registry for consumers then there's nothing to check.

adrianbenavides: Ok you convinced me 😆 Sooner that later we'll have to refactor the NodeManager registry, at least to move them out of the same lock. I think that will be also a good chance to revisit the individual registries.

etorreborre: BTW I think my initial comment still stands `!matches!(e.kind(), _kind)` is never going to be true, see [this StackOverflow question](https://stackoverflow.com/questions/68155018/rust-matches-macro-behaves-unexpectedly-in-regards-to-enum).

adrianbenavides: argh! let me check that, it's actually the first time I felt convinced to use that macro, but guess I'll have to get back to the classic match. Thanks for the heads up!

implementations/rust/ockam/ockam_api/src/lib.rs
adrianbenavides: I added a test for this function to verify that the macro works as expected in this case

implementations/rust/ockam/ockam_api/src/nodes/service/node_services.rs
adrianbenavides: should be ready now 🤞

etorreborre (approved):
Let's go 🐎

Issue #3434, opened 2022-09-07
Show command help in $PAGER or less


mrinalwadhwa:
# Currently:

We have many help commands that show a long help, for example `ockam secure-channel --help`. This hides the command syntax above the fold.

# Desired:

1. Detect if $PAGER is set, if so invoke that and send all help output to it.
2. If $PAGER is not set, default to `less`

---

_We love helping new contributors!_
_If you have questions or need help as you work on your first Ockam contribution, please leave a comment on [this discussion](https://github.com/ockam-network/ockam/discussions/1081)._
_If you're looking for other issues to contribute to, checkout [this discussion](https://github.com/build-trust/ockam/discussions/3206) and labels -  https://github.com/build-trust/ockam/labels/good%20first%20issue or https://github.com/build-trust/ockam/labels/help%20wanted_

phillyphil91:
Hey, i would like to take a look at this, as a first contribution. I am not entirely sure, I understand the details fully though. 
At what stage should the check for $PAGER happen? And once that's set to `less` instead of the normal display of the help commands, this should be "routed" as input to `less` ?

Edit: Found HELP_DETAIL or rather the place where `help::template` etc. is defined

phillyphil91:
In general it looks like only `man` uses the $MANPAGER or $PAGER env. So I guess I would have to find a way for the `help` output to use the tools specified by those ENVs as well. I can take a look if `clap` allows this or if there is any other way to do this. 
If I'm going off into the wrong direction, please let me know.

mrinalwadhwa:
@phillyphil91 thank you for spending time on this.
I think that is right direction.
We need to find a way in clap to override where the output goes.

mrinalwadhwa:
The clap community has a friendly discussions page here https://github.com/clap-rs/clap/discussions so asking there for some guidance might be helpful as well.

phillyphil91:
> The clap community has a friendly discussions page here https://github.com/clap-rs/clap/discussions so asking there for some guidance might be helpful as well.

Good idea. I will have a look there. 


> @phillyphil91 thank you for spending time on this. I think that is right direction. We need to find a way in clap to override where the output goes.

Clap allows you to write to stdout or to a io::Write object. 
![image](https://user-images.githubusercontent.com/57916112/189432866-88d046c7-f7ba-4e4f-9c29-1fcb2d84d998.png)
[...]

PR #5049, opened 2023-06-02
feat(rust): paginate help texts

i-b-o-t:
using $PAGER if set, defaulting to `less` and falling back to `more` or no pagination if no pager is available.

Addresses #3434

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/lib.rs
+++ b/implementations/rust/ockam/ockam_command/src/lib.rs
@@ -80,8 +80,7 @@ use secure_channel::{listener::SecureChannelListenerCommand, SecureChannelComman
 use service::ServiceCommand;
 use space::SpaceCommand;
 use status::StatusCommand;
-use std::path::PathBuf;
-use std::sync::Mutex;
+use std::{ffi::OsString, path::Path, path::PathBuf, process, process::Stdio, sync::Mutex};
 use tcp::{
     connection::TcpConnectionCommand, inlet::TcpInletCommand, listener::TcpListenerCommand,
     outlet::TcpOutletCommand,
@@ -112,7 +111,7 @@ after_long_help = docs::after_help(AFTER_LONG_HELP),
 version,
 long_version = Version::long(),
 next_help_heading = "Global Options",
-disable_help_flag = true
+disable_help_flag = true,
 )]
 pub struct OckamCommand {
     #[command(subcommand)]
@@ -315,13 +314,16 @@ pub fn run() {
         .map(replace_hyphen_with_stdin)
         .collect::<Vec<_>>();
 
-    let command = OckamCommand::parse_from(input);
+    match OckamCommand::try_parse_from(input) {
+        Ok(command) => {
+            if !command.global_args.test_argument_parser {
+                check_if_an_upgrade_is_available();
+            }
 
-    if !command.global_args.test_argument_parser {
-        check_if_an_upgrade_is_available();
-    }
-
-    command.run();
+            command.run();
+        }
+        Err(help) => show_help(help),
+    };
 }
 
 impl OckamCommand {
@@ -468,4 +470,79 @@ pub(crate) fn replace_hyphen_with_stdin(s: String) -> String {
     } else {
         s
     }
+}
+
+const ENV_FORCE_COLOR: &str = "ockam_force_color";
+
+fn show_help(help: clap::Error) {
+    use std::env;
+    let mut try_fallback = false;
+    let preferred_pager = env::var_os("PAGER").unwrap_or_else(|| {
+        try_fallback = true;
+        OsString::from("less")
+    });
+
+    if preferred_pager == "false" {
+        use clap::{ColorChoice::*, CommandFactory};
+        let possibly_forced = if env::var_os(ENV_FORCE_COLOR).is_some() {
+            Always
+        } else {
+            Auto
+        };
+
+        help.with_cmd(&OckamCommand::command().color(possibly_forced))
+            .exit();
+    }
+
+    if let Ok(()) = paginate_with(preferred_pager, &help) {
+        return;
+    }
+    if try_fallback {
+        if let Ok(()) = paginate_with(OsString::from("more"), &help) {
+            return;
+        }
+    }
+    paginate_with(OsString::from("false"), &help)
+        .expect("displaying help without pagination should always work");
+}
+
+fn paginate_with(pager: OsString, help: &clap::Error) -> Result<()> {
+    let mut pager_invocation = process::Command::new(&pager);
+    if Path::new(&pager)
+        .file_name()
+        .map_or("", |s| s.to_str().unwrap_or(""))
+        == "less"
+    {
+        pager_invocation.env("LESS", "FRX");
+        // - F: no pagination if the text fits entirely into the window
+        // - R: allow ANSI escapes output formatting
+        // - X: prevents clearing the screen on exit
+        // - using env var in case a lesser `less` poses as `less`
+    }
+    let mut pager_process = pager_invocation.stdin(Stdio::piped()).spawn()?;
+    let pipe = Stdio::from(pager_process.stdin.take().expect("stdin open?"));
+
+    let exit_status = {
+        let mut my_args = std::env::args_os();
+        let my_exe_name = my_args.next().unwrap_or("ockam".into());
+
+        let mut rerun = process::Command::new(my_exe_name);
+        rerun.args(my_args).env("PAGER", "false");
+        use atty::Stream::*;
+        let output_stream = if help.use_stderr() {
+            rerun.stderr(pipe);
+            Stderr
+        } else {
+            rerun.stdout(pipe);
+            Stdout
+        };
+        if atty::is(output_stream) {
+            rerun.env(ENV_FORCE_COLOR, "_");
+        }
+        rerun.status()?.code().unwrap_or(exitcode::SOFTWARE)
+        // dropping owned pipe hands over pager control to the user
+    };
+
+    pager_process.wait()?;
+    process::exit(exit_status);
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3388,6 +3388,7 @@ dependencies = [
  "assert_cmd",
  "async-recursion",
  "async-trait",
+ "atty",
  "clap 4.3.2",
  "clap_complete",
  "clap_mangen",
--- a/implementations/rust/ockam/ockam_command/Cargo.toml
+++ b/implementations/rust/ockam/ockam_command/Cargo.toml
@@ -55,6 +55,7 @@ path = "src/bin/ockam.rs"
 anyhow = "1"
 async-recursion = { version = "1.0.0" }
 async-trait = "0.1"
+atty = "0.2.14"
 clap = { version = "4.3.2", features = ["derive", "cargo", "wrap_help"] }
 clap_complete = "4.3.1"
 clap_mangen = "0.2.12"


mrinalwadhwa:
@i-b-o-t thank you for spending time on this.
Use of less worked like a charm, nice!

A few quick things:
1. Could you run `cargo fmt` and `cargo clippy` and fix any problems they report
2. Our commit messages are all lower case so please amend the commit message to `feat(rust): paginate help texts`

After that, I think we can merge this.

i-b-o-t:
Sure, I can do that.

Also, I noticed that the exit status of the process should be either 0 or 2 depending on whether the help text is requested by the user explicitly or shown as a consequence of a usage error. By the same logic, in case no pagination is available, the output should go to `stdout` or `stderr` streams respectively.

Also, checking whether the pager is `less` should better be done via `path::file_name` in case the environment variable contains a full path.

So give me a minute to take care of that.

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -470,7 +470,8 @@ fn paginate(text: &StyledStr) -> Result<()> {
 fn paginate_with(pager: &str, text: &StyledStr) -> Result<()> {
     let mut invocation = process::Command::new(pager);
 
-    if pager == "less" {
+    if Path::new(pager).file_name()
+            .map_or("", |s| s.to_str().unwrap_or("")) == "less" {
```
i-b-o-t: Could have gone `unwrap_or_default().to_str().unwrap_or_default()` but wasn't sure what the preference was. Though technically the slightly more complicated syntax, I found it read more obvious at first glance this way.

i-b-o-t:
In individual commits for easy reviewing. The first one is still the same, except for rewording its log entry to lowercase. Rebasing with `--autosquash` will turn them all into a single pretty one automagically, as you most probably know.

i-b-o-t:
```shell
$ PAGER=false ./target/debug/ockam -h 1>/dev/null && echo OK
OK
$ PAGER=false ./target/debug/ockam 2>/dev/null || echo OK
OK
$ cat mock/swallow 
#!/bin/sh

$ PAGER=./mock/swallow ./target/debug/ockam -h && echo OK
OK
$ PAGER=./mock/swallow ./target/debug/ockam || echo OK
OK
$ cat mock/less 
#!/bin/sh

if test "${LESS}" = "-F"; then
	echo OK
fi
$ PAGER=./mock/less ./target/debug/ockam 
OK
```

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -446,3 +448,62 @@ pub(crate) fn replace_hyphen_with_stdin(s: String) -> String {
         s
     }
 }
+
+const EXIT_OK: i32 = 0;
+const EXIT_USAGE: i32 = 2;
```
adrianbenavides: You should use the error codes provided at [util/exitcode.rs](https://github.com/build-trust/ockam/blob/d59ccf0c41bc49d733fcc3733daadc3f9af34c8d/implementations/rust/ockam/ockam_command/src/util/exitcode.rs#L3)

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -446,3 +448,62 @@ pub(crate) fn replace_hyphen_with_stdin(s: String) -> String {
...
+
+fn show_help(help: clap::Error) {
```
adrianbenavides, line 477: I would rename the `help` variable to `error`. It was a little confusing reading this variable until I got here.

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -446,3 +448,62 @@ pub(crate) fn replace_hyphen_with_stdin(s: String) -> String {
...
+    paginate(&help.render()).unwrap_or_else(|_| help.print().unwrap());
+
+    let _ = std::io::stdout().lock().flush();
+    let _ = std::io::stderr().lock().flush();
```
adrianbenavides: Is this really needed? I thought both streams were flushed automatically when exiting the program

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -446,3 +448,62 @@ pub(crate) fn replace_hyphen_with_stdin(s: String) -> String {
...
+    process::exit(if help.use_stderr() {
+        EXIT_USAGE
+    } else {
+        EXIT_OK
+    });
```
adrianbenavides: It would be very helpful to have the `if` block as a variable with a little explanation of what it's happening there (i.e. why an USAGE error if the error is written to stderr?)

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -297,13 +296,16 @@ pub fn run() {
...
+    match OckamCommand::try_parse_from(input) {
+        Ok(command) => {
+            if !command.global_args.test_argument_parser {
+                check_if_an_upgrade_is_available();
+            }
 
-    command.run();
+            command.run();
+        }
+        Err(help) => show_help(help),
```
adrianbenavides, line 325: Given that the `show_help` function is pretty short and it's only called here, I think it would be better to remove it and paste the code here.

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -446,3 +448,62 @@ pub(crate) fn replace_hyphen_with_stdin(s: String) -> String {
...
+
+    let _ = std::io::stdout().lock().flush();
+    let _ = std::io::stderr().lock().flush();
+    process::exit(if help.use_stderr() {
+        EXIT_USAGE
+    } else {
+        EXIT_OK
+    });
+}
+
```
adrianbenavides: The following two functions could be simplified. There is a library called [`which`](https://crates.io/crates/which) you could use to get the binaries paths.

```rust
fn paginate(text: &StyledStr) -> Result<()> {
    let default_pager = get_env_with_default("PAGER", "less".to_string())?;
    let mut pagers = vec![&default_pager, "more", "bat"];
    for pager in pagers.drain(..) {
        if let Ok(path) = which::which(pager) {
            let path = path
                .to_str()
                .expect(&format!("{pager}'s path is not valid UTF-8"));
            if let Ok(()) = paginate_with(path, text) {
                break;
            }
        }
    }
    Ok(())
}

fn paginate_with(pager_path: &str, text: &StyledStr) -> Result<()> {
    let mut cmd = process::Command::new(pager_path);
    // Add needed arguments to the pager command.
    match pager_path {
        "less" => {
            // No pagination if the text fits entirely into the window.
            // Using env var in case a lesser `less` poses as `less`.
            cmd.env("LESS", "-F");
        }
        _ => {}
    }
    let mut child = cmd.stdin(process::Stdio::piped()).spawn()?;
    // Write text to pager's stdin.
    {
        let mut stdin = child.stdin.take().unwrap();
        stdin.write_all(format!("{}", text).as_bytes())?;
    }
    child.wait()?;
    Ok(())
}
```

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -446,3 +448,62 @@ pub(crate) fn replace_hyphen_with_stdin(s: String) -> String {
...
+fn paginate_with(pager: &str, text: &StyledStr) -> Result<()> {
+    let mut invocation = process::Command::new(pager);
+
+    if Path::new(pager)
+        .file_name()
+        .map_or("", |s| s.to_str().unwrap_or(""))
+        == "less"
+    {
+        invocation.env("LESS", "-F");
```
adrianbenavides: Although this works as expected, the default behavior now strips the colors from the output. We still need that. I've tried passing the `-R` argument but it didn't work, so perhaps it's related to how the stdout/stderr streams are set up?

adrianbenavides (commented):
Nice job @i-b-o-t! It's pretty much there but I've left some comments with potential improvements. Thanks for taking the time with this 🙏

implementations/rust/ockam/ockam_command/src/lib.rs
i-b-o-t: It's taken from the `clap_builder::util::safe_exit` function and I just copied it over. I think you're right in case the compiler is instructed to perform a proper unwind (is that always the case for Ockam?). If not, I suspect it may depend on whether there's additional (non-OS) buffering in place.

implementations/rust/ockam/ockam_command/src/lib.rs
i-b-o-t: I basically like the suggestion. However, those are copied from `clap::Error::exit` which uses the same exit code for `USAGE` as prominent UNIX commands such as `ls` and the like. Using those constants would be a change in behavior, which I feel is outside the scope of this submission.

implementations/rust/ockam/ockam_command/src/lib.rs
i-b-o-t: I chose the name intentionally to indicate that, although the parameter's type is `Error`, it represents any help texts, including those that are issued upon the user's request - without an actual error!

implementations/rust/ockam/ockam_command/src/lib.rs
i-b-o-t: I'm not a big fan of comments TBH and I thought I wrote too many of them already :sweat_smile: . Again, the code is copied from `clap::Error::exit`. But I see what you mean, arguably `clap::Error::use_stderr()` would better be called something like `represents_usage_error()`. I will introduce a boolean variable to "say it with code".

implementations/rust/ockam/ockam_command/src/lib.rs
i-b-o-t: Humm... Sorry, no. I disagree again: `show_help` contains implementation details of an abnormal path of execution and the `run` function is the entry point. I'd prefer to give whoever is tracing into it a clean first impression by keeping it concise and on a single level of abstraction.

implementations/rust/ockam/ockam_command/src/lib.rs
i-b-o-t: How would that be a simplification? Also adding another dependency.. But more importantly, I think you are misreading my code, for this breaks when `PAGER=/opt/not-on-PATH/pager`. I added a fallback to `more` with devices in mind and don't think the overkill of probing any pager on the planet is necessary.

implementations/rust/ockam/ockam_command/src/lib.rs
i-b-o-t: I just tried `pgrep less` when running `git log` and it appears `less` is not running. Then again, pressing `h` shows a screen titled "SUMMARY OF LESS COMMANDS"...

Mysteriously
```shell
$ strings $( which git ) |grep LESS
LESS=FRX LV=-c
```

UPDATE: Ah, OK! Tthe `LV=-c` part is unrelated, a dash doesn't seem to be required, and the `-X` option seems a good idea too since it just keeps everything as-is on exit, so you can scroll back up to the help text (like to look up things while editing the command line).

i-b-o-t:
:heart_decoration: Thanks for your review! Please don't get too negative of an impression from me disagreeing in many places and let's not obscure the fact that it *did* raise valid points.

i-b-o-t:
Behaves just like Git now, AFAICT rebased and ready.

implementations/rust/ockam/ockam_command/src/lib.rs
adrianbenavides: Just to give you some more context, the error codes we use are the ones defined at https://man.freebsd.org/cgi/man.cgi?query=sysexits&apropos=0&sektion=0&manpath=FreeBSD+4.3-RELEASE&format=html

We'll review it again in case we should change them.

implementations/rust/ockam/ockam_command/src/lib.rs
adrianbenavides: Yeah, I get your point. The reason we try to make things as explicit and documented as possible is because this code (`ockam_command`) is read by people that are contributing to an opensource project for their first time (students, or people playing around with rust, for example), so we try to keep the code as less cryptic as possible, especially around non-obvious logic/assumptions. Of course, we don't want to document code that is self explanatory (as you said, sometimes a good variable name is good enough to explain what's going on).

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -447,3 +449,78 @@ pub(crate) fn replace_hyphen_with_stdin(s: String) -> String {
...
+        use clap::{ColorChoice::*, CommandFactory};
+        let possibly_forced = if env::var_os(ENV_FORCE_COLOR).is_some() {
+            Always
+        } else {
+            Auto
+        };
+
+        help.with_cmd(&OckamCommand::command().color(possibly_forced))
+            .exit();
+    }
```
adrianbenavides, line 495: The whole function feels obscure and with lots of hidden subtleties around the pager parsing. Here are some questions that I had while reading this particular block:

- This feels hacky, why not use an Option to represent "false" instead?
- If a user sets the "PAGER" env var to "false", why would they want to set it in the first place? I'd rather provide another env var to disable this feature.
- The user could have set the `NO_COLOR` env var, which is not taken into consideration here
- What's happening in this block? I had to read the documentation of the `with_cmd` and `exit` commands to learn that it's an early exit of the process. I also had to get into the details of the `paginate_with` to understand that they depend on each other. I wonder if this block should live in the `paginate_with` function.

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -447,3 +449,78 @@ pub(crate) fn replace_hyphen_with_stdin(s: String) -> String {
...
+    paginate_with(OsString::from("false"), &help)
+        .expect("displaying help without pagination should always work");
+}
+
+fn paginate_with(pager: OsString, help: &clap::Error) -> Result<()> {
+    let mut pager_invocation = process::Command::new(&pager);
+    if Path::new(&pager)
+        .file_name()
+        .map_or("", |s| s.to_str().unwrap_or(""))
+        == "less"
```
adrianbenavides, line 514: I'm not sure how portable would this be across different OS's. I still vote for using `which`. It will also look cleaner and it lets you check if the binary actually exists in the system.

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -447,3 +449,78 @@ pub(crate) fn replace_hyphen_with_stdin(s: String) -> String {
...
+    }
+    let mut pager_process = pager_invocation.stdin(Stdio::piped()).spawn()?;
+    let pipe = Stdio::from(pager_process.stdin.take().expect("stdin open?"));
+
+    let exit_status = {
+        let mut my_args = std::env::args_os();
+        let my_exe_name = my_args.next().unwrap_or("ockam".into());
+
+        let mut rerun = process::Command::new(my_exe_name);
+        rerun.args(my_args).env("PAGER", "false");
```
adrianbenavides, line 530: Ok, so this env var is only for internal use right?

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -447,3 +449,78 @@ pub(crate) fn replace_hyphen_with_stdin(s: String) -> String {
...
+        } else {
+            rerun.stdout(pipe);
+            Stdout
+        };
+        if atty::is(output_stream) {
+            rerun.env(ENV_FORCE_COLOR, "_");
+        }
+        rerun.status()?.code().unwrap_or(exitcode::SOFTWARE)
+        // dropping owned pipe hands over pager control to the user
+    };
```
adrianbenavides, line 544: I'm sure we can do this without having to rerun the same command. Did you need it to add the `ENV_FORCE_COLOR`?

adrianbenavides:
> 💟 Thanks for your review! Please don't get too negative of an impression from me disagreeing in many places and let's not obscure the fact that it did raise valid points.

No problem! That's the whole point of doing code reviews right? We are here to learn from each other 👍

adrianbenavides:
Hey @i-b-o-t , I was taking another look at the PR and I just wanted to clarify something. The PR looks great, it works as expected and we could merge it as is. All my last comments are just questions that would help me understand better your solution and small details that could be improved, but they won't block the PR from merging.

That being said, feel free to mark the comments as resolved. We can tackle whatever is left in a separate PR.

And thanks again for taking the time with this issue, it ended up being more tricky than I initially anticipated!

i-b-o-t:
Hey @adrianbenavides , thanks for giving it another chance - being able to take a step back and reconsider one's own position shows character! Defending it comes with wholehearted commitment, I know, nothing wrong with that as long as we realize when to hold on and take a break, a deep breath, another look or perspective into consideration.

**The simplest solutions is not always immediately obvious.**

I do think that Occam's Razor is a fine analogy to problem solving in software development and *given the scope and constants that constrain us here and now*, the implementation in a nutshell is as simple as it gets.

Once we widen the scope (e.g. by introducing further command line options, changing the public interface, as we both have suggested above) or find a way to redefine the constants (e.g. by requesting additional Clap API), there is always room for improvement piecing the puzzle together in a coherent fashion. However, in order for these efforts to be both manageable and motivating, I think it is a good idea to chop things up into individual working steps and then working that plan without too much deviation. Of course there are exceptions, like something being completely infeasible, but we don't have that case here. Also, there is almost always some nits to pick on; doesn't mean we should just because we could (invert priorities).

> The PR looks great, it works as expected and we could merge it as is

Just pushed the squashed commit.

EDIT: To complete my philosophical sermon and add another quote

> That's the whole point of doing code reviews right? We are here to learn from each other

I agree, plus hopefully merging better code than we would without.

From that angle, nitpicking certainly has its place in reviews, as it can mean teaching each other (often ideally generalizable) details to be aware of, but it also bears the risk of wasting time and inverting priorities, so there's a pair of anti-forces that need to balanced.

adrianbenavides:
Thanks for the message @i-b-o-t , I really appreciate it 🙏 . One last thing: we require the [commits to be signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits). Let me know if you want me to sign it or you want to do it yourself. It's a one time operation and it's pretty quick to set up.

i-b-o-t:
> One last thing: we require the [commits to be signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).

Are they not? I set up PGP the other day. I need to check on it...

```
$ git config commit.gpgsign
true
```

Shouldn't I see an error if they were not. Oh, or does it also need that "Signed off by" line in the commit message?

adrianbenavides:
> Shouldn't I see an error if they were not. Oh, or does it also need that "Signed off by" line in the commit message?

That's not needed. Looks like your commit is signed but can't be verified. I think you have to upload your public key to your githu account for the verification to work.

implementations/rust/ockam/ockam_command/src/lib.rs
i-b-o-t: I already had a previous state return those, but then just ended up using `clap::Error::exit` again for simplicity.

Since you mentioned elsewhere that you see a lot of novice contributors, probably also people who are new to Rust (I include myself in that group), and there's quite some setup to be done in total to get on board, the remapping the of these exit codes could be a really nice and easy follow-up task to file. Basically:

```
clap_error.print();

// from `clap_builder::util::safe_exit` (not needed if we always
// unwind the stack on exit):
let _ = std::io::stdout().lock().flush();
let _ = std::io::stderr().lock().flush();

let is_usage_error = clap_error.use_stderr();
// might as well replicate(private) `clap::Error::stream()`
// making it more obvious why that is so

process::exit(if is_usage_error { USAGE } else { OK });
```

implementations/rust/ockam/ockam_command/src/lib.rs
i-b-o-t: Where's my reply? Possibly hit the wrong button - gotta get used to GitHub again. Why the re-run is the way of the least resistance was posted elsewhere.

Introducing a command line flag like GCC's `-fdiagnostics-color=[auto|never|always]` setting `clap::ColorChoice` (which happens to be an `enum` over the same options) would be a possible way to get rid of that internal environment variable in a follow-up task.

i-b-o-t: That said, I personally don't really really like the above suggestion, since some settings already look crappy using a pager by default. Git doesn't need it. And there is means to control the automatic by connecting the output channels to (possibly pseudo-) terminals or not.

You mentioned a conventional `NO_COLOR` env var. Supporting that one (and keeping the internal one or piggy-backing on `NO_COLOR` with an internal, "impossible"-to-collide value that instead forces colors to ON - just for communicating with the child - both cases being mutually exclusive) would be another low-hanging fruit and better aligned with the rest of the existing interface.

adrianbenavides: We already support the `NO_COLOR` env var and we have a way to determine if the streams support color, although we should introduce the internal "force-color" setting as you mentioned. There's definitely room for improvement on this regard. We'll give it a thought after testing this for a while.

adrianbenavides (approved):
Let's get this merged! Thanks again @i-b-o-t, this was a great contribution! 🚀

PR #5056, opened 2023-06-05
Fix Typescript Dependabot Path

metaclips:
We were using a wrong path when updating dependency version with dependabot


mrinalwadhwa:
good catch but since we've shifted to pnpm, we have to wait for support to land, should be any week now https://github.com/dependabot/dependabot-core/issues/1736#issuecomment-1567300493

mrinalwadhwa (approved):

PR #5057, opened 2023-06-05
Remove AccessControl from mandatory start worker arguments

SanjoDeundiak:
1. Remove Access Control arguments from `ctx.start_worker` as with FlowControls existence we don't need to worry about it 
2. Remove Access Control arguments from `ctx.start_processor` as with FlowControls existence we don't need to worry about it 
3. Remove `LocalSourceOnly` and `LocalOnwardOnly` as with FlowControls existence they are not needed and they don't function as intended with the current TCP transport design
4. Improve `WorkerBuilder` and `ProcessorBuilder`


implementations/rust/ockam/ockam_node/src/processor_builder.rs
```diff
@@ -4,98 +4,172 @@ use crate::{relay::ProcessorRelay, Context, NodeMessage};
...
+    pub async fn start(self, context: &Context) -> Result<()> {
+        start(context, self.mailboxes, self.processor).await
     }
 }
+
+pub struct ProcessorBuilderOneAddress<P>
+where
+    P: Processor<Context = Context>,
+{
+    incoming_ac: Arc<dyn IncomingAccessControl>,
```
davide-baldo, line 77: I wonder if we can just default to DenyAll for incoming access control in processors

davide-baldo (approved):
LGTM

implementations/rust/ockam/ockam_node/src/processor_builder.rs
SanjoDeundiak: We do default to DenyAll, if you need other AC you should call `with_incoming_access_control` or `with_outgoing_access_control`

PR #5058, opened 2023-06-05
chore(elixir): remove duplicate tests

polvorin:
we have [simpler tests](https://github.com/build-trust/ockam/blob/develop/implementations/elixir/ockam/ockam/test/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm_test.exs#L29) exercising the arrival of out-of-order packets, duplicates, etc.    This was a more involved way that is not needed now.


hairyhum:
To me is seems like the tests operate on a different layer and those are more important as they test more high level scenarios with messaging. 
Maybe we could simplify the secure channel tests instead of removing them?

polvorin:
> To me is seems like the tests operate on a different layer and those are more important as they test more high level scenarios with messaging. Maybe we could simplify the secure channel tests instead of removing them?

agree this is a upper layer  so while the other test checks that the encryption state behaves correctly when out-of-order or trash packet are received,   here  it test that the _process_ holding that state do not crash or do something bad.   Will be enough to remove all but 1 test from here just to check that,   and let the specific ones on the encryption/decryption unit tests?.     

Will try to simplify the man in the middle,  I'm not very fond of that implementation

PR #5064, opened 2023-06-05
feat(rust): full local kafka implementation

davide-baldo:
Implemented a local Kafka usage scenario to pass data locally.
I achieved this by replicating the same logic of the elixir side, keeping full compatibility with the elixir implementation.
What was missing was the `outlet` node, the node which connects to the Kafka cluster.
These components were missing:
 - `OutletManager`: intercepts Kafka traffic and dynamically creates outlets (the existing implementation was generalized to allow any arbitrary interceptor)
 - `PrefixForwarder`: adds a prefix to the static forwarder
 - the `KafkaOutlet` service, which starts the bootstrap outlet as well as above services.

In this implementation, the more general Kafka use case is hardened, applying credential validation, between producer and consumer, and flow control to every interaction. To simplify a bit, the existing implementation of the random unique ID associated to each topic/partition was dropped in favor of directly using the secure channel decryptor address of the consumer instead, removing the secure channel controller listener on the consumer side.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/kafka/outlet_service/prefix_forwarder.rs
+++ b/implementations/rust/ockam/ockam_api/src/kafka/outlet_service/prefix_forwarder.rs
@@ -1,0 +1,107 @@
+use crate::kafka::KAFKA_OUTLET_CONSUMERS;
+
+use crate::DefaultAddress;
+use core::str::from_utf8;
+use ockam::{Context, Result, Routed, Worker};
+use ockam_core::errcode::{Kind, Origin};
+use ockam_core::flow_control::{FlowControlId, FlowControlPolicy};
+use ockam_core::{Address, AllowAll, AllowOnwardAddress};
+
+/// This service applies a prefix to the provided static forwarding address.
+/// This service was created mainly to keep full compatibility with the existing
+/// erlang implementation.
+pub struct PrefixForwarderService {
+    prefix: String,
+    secure_channel_listener_flow_control_id: FlowControlId,
+}
+impl PrefixForwarderService {
+    pub async fn create(
+        context: &Context,
+        secure_channel_listener_flow_control_id: FlowControlId,
+    ) -> Result<()> {
+        // add the this worker as consumer for the secure channel listener
+        let worker_address = Address::from_string(KAFKA_OUTLET_CONSUMERS);
+        context.flow_controls().add_consumer(
+            worker_address.clone(),
+            &secure_channel_listener_flow_control_id,
+            FlowControlPolicy::SpawnerAllowMultipleMessages,
+        );
+
+        let worker = Self {
+            prefix: "consumer_".to_string(),
+            secure_channel_listener_flow_control_id,
+        };
+
+        context
+            .start_worker_with_access_control(
+                worker_address,
+                worker,
+                AllowAll,
+                AllowOnwardAddress(DefaultAddress::FORWARDING_SERVICE.into()),
+            )
+            .await
+    }
+}
+
+#[ockam::worker]
+impl Worker for PrefixForwarderService {
+    type Message = Vec<u8>;
+    type Context = Context;
+
+    async fn handle_message(
+        &mut self,
+        ctx: &mut Context,
+        msg: Routed<Self::Message>,
+    ) -> Result<()> {
+        // the payload is just a string
+        let address = match msg.payload().get(1..) {
+            Some(address) => match from_utf8(address) {
+                Ok(v) => v.to_string(),
+                Err(_e) => {
+                    return Err(ockam_core::Error::new(
+                        Origin::Application,
+                        Kind::Invalid,
+                        "invalid address",
+                    ));
+                }
+            },
+            None => {
+                return Err(ockam_core::Error::new(
+                    Origin::Application,
+                    Kind::Invalid,
+                    "invalid address",
+                ));
+            }
+        };
+
+        let new_address = format!("{}_{}", &self.prefix, address);
+
+        debug!(
+            "prefix forwarder, renamed from {} to {}",
+            address, new_address
+        );
+
+        let mut bytes = new_address.clone().into_bytes();
+        let mut new_payload: Vec<u8> = vec![bytes.len() as u8];
+        new_payload.append(&mut bytes);
+
+        let mut message = msg.into_local_message();
+        let transport_message = message.transport_mut();
+
+        transport_message.onward_route.step()?;
+
+        // prefix consumer_ to the address
+        transport_message.payload = new_payload;
+
+        ctx.forward(message).await?;
+
+        // The new forwarder needs to be reachable by the default secure channel listener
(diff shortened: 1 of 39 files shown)


davide-baldo:
rebasing

polvorin (dismissed):
LGTM but I'm not sure familiar with the kafka scenario,  maybe should get some other :eye:  as well

davide-baldo:
something is broken after the rebase, WIP

davide-baldo:
Fixed

caoakleyii (approved):

implementations/rust/ockam/ockam_api/src/util.rs
```diff
@@ -449,4 +478,22 @@ pub mod test_utils {
             identifier: identity.identifier(),
         })
     }
+
+    async fn create_identity_zero(secure_channels: &Arc<SecureChannels>) -> Result<Identity> {
```
SanjoDeundiak, line 482: What is the reason for creating such Identity?

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
```diff
@@ -136,20 +136,25 @@ pub struct CreateOutlet<'a> {
     #[b(2)] pub worker_addr: Cow<'a, str>,
     /// A human-friendly alias for this portal endpoint
     #[b(3)] pub alias: Option<CowStr<'a>>,
+    /// Allow the outlet to be reachable from the default secure channel, useful when we want to
+    /// tighten the flow control
+    #[b(4)] pub reachable_from_default_secure_channel: bool,
```
SanjoDeundiak, line 141: Should be `#[n(4)]`, but at this point I'm not sure if it's even handled somehow differently under the hood

implementations/rust/ockam/ockam_api/src/kafka/outlet_service/prefix_forwarder.rs
```diff
@@ -0,0 +1,107 @@
...
+
+        let mut message = msg.into_local_message();
+        let transport_message = message.transport_mut();
+
+        transport_message.onward_route.step()?;
+
+        // prefix consumer_ to the address
+        transport_message.payload = new_payload;
+
+        ctx.forward(message).await?;
```
SanjoDeundiak, line 96: That's questionable

implementations/rust/ockam/ockam_api/src/kafka/outlet_service/prefix_forwarder.rs
```diff
@@ -0,0 +1,107 @@
...
+use core::str::from_utf8;
+use ockam::{Context, Result, Routed, Worker};
+use ockam_core::errcode::{Kind, Origin};
+use ockam_core::flow_control::{FlowControlId, FlowControlPolicy};
+use ockam_core::{Address, AllowAll, AllowOnwardAddress};
+
+/// This service applies a prefix to the provided static forwarding address.
+/// This service was created mainly to keep full compatibility with the existing
+/// erlang implementation.
+pub struct PrefixForwarderService {
```
SanjoDeundiak, line 13: Could you please elaborate on what this service is doing and what is the use-case? Do you think it could be worth updating the original ForwardingService instead?

SanjoDeundiak (commented):
It is pretty hard to follow honestly. Left some comments

implementations/rust/ockam/ockam_api/src/util.rs
davide-baldo: During node manager initiation inside a test, I want to be able to add an authority with a credential.
Since the node manager currently accepts only configuration-based authority parameters, I needed to create the credential **before** the node manager initialization. Hence, I needed to create the same identity twice, the easiest (and most reproducible) way was to create it using a private key made of zeroes.

I'm creating a follow-up to refactor the node creation to remove this hack

implementations/rust/ockam/ockam_api/src/kafka/outlet_service/prefix_forwarder.rs
davide-baldo: The use case in the Erlang implementation is to avoid having a client registering arbitrary addresses inside the node, reducing the possibility of vulnerabilities.
It could be useful in this security sense only if we blocked the forwarding service by default.
By having this standalone service, we can have the same semantic of the orchestrator without having to adjust for special cases.

implementations/rust/ockam/ockam_api/src/kafka/outlet_service/prefix_forwarder.rs
davide-baldo: What do you mean? the worker is spawned with `AllowOnwardAddress(DefaultAddress::FORWARDING_SERVICE.into()),`

implementations/rust/ockam/ockam_api/src/util.rs
davide-baldo: https://github.com/build-trust/ockam/issues/5118

implementations/rust/ockam/ockam_api/src/kafka/outlet_service/prefix_forwarder.rs
davide-baldo: It could be useful to generalize for keep compatibility reasons, but i would wait until we actually have a 2nd usage

SanjoDeundiak: But what if the original ForwardingService had a prefix (could be empty) argument in the constructor? Would it then cover both cases?

implementations/rust/ockam/ockam_api/src/kafka/outlet_service/prefix_forwarder.rs
SanjoDeundiak: Yeah, I see that, but it feels weird to me when we forward a message, relying on someone who sent us it to put the right address in the `onward_route`, when we ourselves know where exactly that message should go.

SanjoDeundiak: It's a redundancy at best, possible vulnerability at worst (if you forgot to setup access controls)

implementations/rust/ockam/ockam_api/src/util.rs
SanjoDeundiak: Ok, I think having a pre-generated key as a constant is the better way. I don't want us to have the all-zero private key, even in tests.

implementations/rust/ockam/ockam_api/src/kafka/outlet_service/prefix_forwarder.rs
SanjoDeundiak: My main point is that writing a proper ForwardingService could be tricky, given AccessControls challenges. Having 2 implementations reduces our changes that both are correct :D

PR #5072, opened 2023-06-06
Clean `FlowControls` resources

SanjoDeundiak:
`FlowControls` struct contains much information about `Spawner`s, `Producer`s, and `Consumer`s, which was never freed. This is an imperfect attempt to free those resources when an `Address` is stopped on the node. The problem with that approach is that it's possible to add an `Address` to the `FlowControls` and not start any `Worker` on that `Address`. In this case, `Address` is never stopped by the node, and cleanup for that `Address` is never called. In real life, such a situation should be very rare.

### DIFF
--- a/implementations/rust/ockam/ockam_core/src/flow_control/flow_controls/tests.rs
+++ b/implementations/rust/ockam/ockam_core/src/flow_control/flow_controls/tests.rs
@@ -1,0 +1,209 @@
+use crate::flow_control::{FlowControlPolicy, FlowControls};
+use crate::Address;
+use rand::distributions::Distribution;
+use rand::distributions::Uniform;
+use rand::prelude::{IteratorRandom, SliceRandom, ThreadRng};
+use rand::Rng;
+
+fn add_random_consumer(flow_controls: &FlowControls, mut rng: &mut ThreadRng) -> Vec<Address> {
+    let address = Address::random_local();
+
+    let choice = Uniform::from(0..3).sample(&mut rng);
+
+    match choice {
+        0 => {
+            // Add a Consumer to a non-existent FlowControlId
+            let flow_control_id = FlowControls::generate_id();
+
+            flow_controls.add_consumer(
+                address.clone(),
+                &flow_control_id,
+                FlowControlPolicy::ProducerAllowMultiple,
+            );
+
+            vec![address]
+        }
+        1 => {
+            // Add a Consumer to a random Spawner
+            let spawner_flow_control_id = flow_controls
+                .spawners
+                .read()
+                .unwrap()
+                .values()
+                .choose(&mut rng)
+                .cloned();
+            match spawner_flow_control_id {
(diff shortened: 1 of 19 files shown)


implementations/rust/ockam/ockam_core/src/flow_control/flow_controls/tests.rs
```diff
@@ -0,0 +1,209 @@
...
+                }
+            }
+            None => vec![],
+        }
+    }
+}
+
+#[test]
+fn test_cleanup() {
+    let mut rng = rand::thread_rng();
```
davide-baldo, line 160: When testing with a random factor, it's simpler to debug failure when either we have a fixed seed or the seed is printed out

implementations/rust/ockam/ockam_node/src/router/stop_worker.rs
```diff
@@ -41,14 +41,6 @@ pub(super) async fn exec(
         }
     };
 
-    // Remove all secondary addresses
-    for addr in record.address_set().iter() {
```
davide-baldo, line 45: I don't get why removing the aliases is no longer necessary

davide-baldo (approved):
LGTM - (I have a gut feeling we'll need to optimize the data access in a few months)

implementations/rust/ockam/ockam_node/src/router/stop_worker.rs
SanjoDeundiak: It's still needed, but it was removed twice. One time here and one time in `InternalMap::free_address` after shutdown acknowledgment was received from a Worker. I think there are a lot of similar surprises in `ockam_node` crate...

implementations/rust/ockam/ockam_core/src/flow_control/flow_controls/tests.rs
SanjoDeundiak: I think there is a benefit of having a random seed each time because there is a possible situation where the test succeeds for the fixed seed and doesn't succeed for some other seed. As for seeding - AFAIK, we don't want tests to print anything.

SanjoDeundiak: Even though the seed idea is good, but we probably would use it manually if that test ever fails for us

etorreborre: In my previous company we had the following strategy:

 - always run the regular CI with a fixed seed because we value non-flakiness more than finding an exceptional use case. Especially because in many cases a failing random tests indicated an incorrect assumption of the test assertion but the production code was fine
 
 - run property tests in a weekly CI with a random seed and more tests per property (BTW @SanjoDeundiak why didn't you use `quickcheck` there instead of hand-rolling arbitrary data generation?)

I think that it is a good balance between stability of the dev process and bug finding

PR #5074, opened 2023-06-06
test(rust): in bats tests, check if a generated random port number is already in use

adrianbenavides:
We have some flaky tests that fail because the port being used to create a tcp-listener is already in use (generally by another OS process outside of ockam's realm). Failing tests: [1](https://github.com/build-trust/ockam/actions/runs/5166030533/jobs/9306704471?pr=5055), [2](https://github.com/build-trust/ockam/actions/runs/5187990020/jobs/9351034259?pr=5057)

I've modified the `random_port` function to use `netstat` to check if the port is already picked up. Note that collisions might still happen if two tests run in parallel and generate the same random number.

### DIFF
--- a/implementations/rust/ockam/ockam_command/tests/bats/load/base.bash
+++ b/implementations/rust/ockam/ockam_command/tests/bats/load/base.bash
@@ -98,7 +98,23 @@ random_str() {
 
 # Returns a random port in the range 49152-65535
 random_port() {
-  shuf -i 49152-65535 -n 1
+  port=0
+  max_retries=10
+  i=0
+  while [[ $i -lt $max_retries ]]; do
+    port=$(shuf -i 10000-65535 -n 1)
+    netstat -latn -p tcp | grep $port >/dev/null
+    if [[ $? == 1 ]]; then
+      break
+    fi
+    ((i++))
+    continue
+  done
+  if [ $i -eq $max_retries ]; then
+    echo "Failed to find an open port" >&3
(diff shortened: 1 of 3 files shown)


implementations/rust/ockam/ockam_command/tests/bats/load/base.bash
```diff
@@ -98,7 +98,24 @@ random_str() {
...
+  port=0
+  max_retries=10
+  i=0
+  while [[ $i -lt $max_retries ]]; do
+    port=$(shuf -i 10000-65535 -n 1)
+    netstat -latn -p tcp | grep $port >/dev/null
+    if [[ $? == 1 ]]; then
+      break
+    fi
+    sleep 1
```
davide-baldo: Sleeping shouldn't give any benefit since we'll choose another random port

davide-baldo (dismissed):
We could still run into race condition if multiple tests choose the same random at the same time, but I don't think we will actually hit it, ever.

implementations/rust/ockam/ockam_command/tests/bats/load/base.bash
adrianbenavides: I put that there while thinking on this:

> Note that collisions might still happen if two tests run in parallel and generate the same random number.

But still, it makes no sense right? 😆 I'll remove it

SanjoDeundiak (approved):

PR #5101, opened 2023-06-09
WIP: improve type safety for `FlowControls`

SanjoDeundiak:
1. Add `ProducerFlowControlId` and `SpawnerFlowControlId`. The implementation is the same, but now we have type-safety guarantees, so the user can't confuse using one in place of the other


implementations/rust/ockam/ockam_core/src/flow_control/flow_controls/flow_controls_debug.rs
```diff
@@ -1,33 +0,0 @@
-use crate::flow_control::FlowControls;
-use crate::Address;
-
-impl FlowControls {
-    /// Prints debug information regarding Flow Control for the provided address
-    #[allow(dead_code)]
-    pub fn debug_address(&self, address: &Address) {
```
davide-baldo, line 7: It's still used by `access_control.rs`, `debugger` feature should fail

davide-baldo: I would really appreciate it if you could keep some form of extra logging, so far, it really helped me track the flow in complex situations

davide-baldo (commented):
I think it's a worthwhile tradeoff between usage simplification and implementation complexity, and should avoid some pitfalls :+1:

implementations/rust/ockam/ockam_identity/tests/common/mod.rs
```diff
@@ -108,24 +110,29 @@ impl SecureChannelListenerInfo {
     }
 }
 
+#[allow(dead_code)]
+pub enum FId {
+    Spawner(SpawnerFlowControlId),
```
davide-baldo: It feels ""wrong"" to have something that logically addresses two different levels of abstraction, instances and spawner, the issue was present even before but was hidden.
Do we really need the possibility to specify a producer?

implementations/rust/ockam/ockam_core/src/flow_control/flow_controls/flow_controls_api.rs
```diff
@@ -88,13 +113,25 @@ impl FlowControls {
...
-        let consumers = self.consumers.read().unwrap();
+    pub fn get_consumers_info_for_producer(
+        &self,
+        flow_control_id: &ProducerFlowControlId,
+    ) -> ProducerConsumersInfo {
+        let consumers = self.consumers_for_producers.read().unwrap();
+        consumers.get(flow_control_id).cloned().unwrap_or_default()
+    }
+
+    /// Get known Consumers for the given [`FlowControlId`]
```
etorreborre, line 124: For a given `SpawnerFlowControlId`

implementations/rust/ockam/ockam_core/src/flow_control/flow_controls/flow_controls_api.rs
```diff
@@ -88,13 +113,25 @@ impl FlowControls {
     }
 
     /// Get known Consumers for the given [`FlowControlId`]
```
etorreborre, line 115: For a given `ProducerFlowControlId`

implementations/rust/ockam/ockam_core/src/flow_control/flow_controls/flow_controls_api.rs
```diff
@@ -79,7 +104,7 @@ impl FlowControls {
     }
 
     /// Mark that given [`Address`] is a Spawner for to the given [`FlowControlId`]
```
etorreborre, line 106: For the given `SpawnerFlowControlId`

implementations/rust/ockam/ockam_core/src/flow_control/flow_controls/flow_controls_api.rs
```diff
@@ -88,13 +113,25 @@ impl FlowControls {
...
+    /// Get known Consumers for the given [`FlowControlId`]
+    pub fn get_consumers_info_for_spawner(
+        &self,
+        flow_control_id: &SpawnerFlowControlId,
+    ) -> SpawnerConsumersInfo {
+        let consumers = self.consumers_for_spawners.read().unwrap();
         consumers.get(flow_control_id).cloned().unwrap_or_default()
     }
 
     /// Get [`FlowControlId`] for which given [`Address`] is a Spawner
```
etorreborre, line 133: Gen `SpawnerFlowControlId`

implementations/rust/ockam/ockam_core/src/flow_control/flow_controls/flow_controls_debug.rs
etorreborre: I agree this was quite useful to me too :-)

implementations/rust/ockam/ockam_core/src/flow_control/policy.rs
```diff
@@ -4,11 +4,9 @@ use minicbor::{Decode, Encode};
...
 #[rustfmt::skip]
 #[cbor(index_only)]
-pub enum FlowControlPolicy {
-    /// Producer is allowed to send any number of messages to the Consumer
-    #[n(0)] ProducerAllowMultiple,
+pub enum SpawnerFlowControlPolicy {
     /// Spawner is allowed to send only one message to the Consumer
-    #[n(1)] SpawnerAllowOnlyOneMessage,
+    #[n(0)] AllowOnlyOneMessage,
     /// Spawner is allowed to send any number of messages to the Consumer
```
etorreborre, line 10: BTW it would be nice to have a bit of documentation giving some examples on why and when we should use one or the other.

implementations/rust/ockam/ockam_identity/tests/common/mod.rs
```diff
@@ -108,24 +110,29 @@ impl SecureChannelListenerInfo {
     }
 }
 
+#[allow(dead_code)]
+pub enum FId {
```
etorreborre: Could we reuse the name `FlowControlId` here?

implementations/rust/ockam/ockam/tests/forwarder.rs
```diff
@@ -36,13 +36,13 @@ async fn test1(ctx: &mut Context) -> Result<()> {
 async fn test2(ctx: &mut Context) -> Result<()> {
     let tcp_listener_options = TcpListenerOptions::new();
     let options = ForwardingServiceOptions::new()
-        .service_as_consumer(
+        .service_as_consumer_for_spawner(
             &tcp_listener_options.spawner_flow_control_id(),
-            FlowControlPolicy::SpawnerAllowMultipleMessages,
+            SpawnerFlowControlPolicy::AllowMultipleMessages,
```
etorreborre, line 41: Why not `AllowOnlyOneMessage` here?

etorreborre (commented):
I think this is an improvement. I still have an issue wrapping my head around the flow control policy. At first it would seems that any spawner like a `SecureChannelListener` should be a consumer for another spawner like `TcpConnectionListener` with `AllowOnlyOneMessage`. But this is contradicted by `ForwardingService` which is defined as a consumer with `AllowMultipleMessages`. If it turns out that this should be `AllowOnlyOneMessage` in that case that suggests another simplification:

 - if we can mark addresses as with their different roles in `FlowControls` then we can automatically infer if the policy is `AllowOnlyOne` or `AllowMultiple` by checking if the flow is a `Spawner -> Spawner` flow
 - then maybe we don't need to have `SpawnerFlowControld` vs `ProducerFlowControlId`?

implementations/rust/ockam/ockam_core/src/flow_control/flow_controls/flow_controls_debug.rs
SanjoDeundiak: @davide-baldo I added some logs when you add producers/spawners/consumers. I hoped it would be enough to debug if something didn't pass the access control. E.g., if you see the "did not pass outgoing access control" message, you should already see the Addresses. And then, you can use those Addresses to search the log and see how it was set up?

implementations/rust/ockam/ockam_identity/tests/common/mod.rs
SanjoDeundiak: I removed `FId`. Tnx

implementations/rust/ockam/ockam_core/src/flow_control/policy.rs
SanjoDeundiak: Ok, so the decision is to remove the policy for now. I'll add some explanation to the PR description

SanjoDeundiak:
Replaced with #5122

PR #5110, opened 2023-06-10
refactor(rust): refactoring common `create` logic around Kafka producer and consumer

### LINKED ISSUE #5087: Extract common logic from `kafka create` commands
### Current behavior

The [`kafka-consumer create`](https://github.com/build-trust/ockam/blob/4842414602f5694193b9d62df45f2bf49d3c70fa/implementations/rust/ockam/ockam_command/src/kafka/consumer/create.rs) and [`kafka-producer create`](https://github.com/build-trust/ockam/blob/4842414602f5694193b9d62df45f2bf49d3c70fa/implementations/rust/ockam/ockam_command/src/kafka/producer/create.rs) commands share the same internal logic with small variations in the output texts and target endpoint.

### Desired behavior

Refactor that logic out to remove the code duplication from those commands.

loghijiaha:
Refactor(rust): refactor common create logic around Kafka producer and consumer

## Current behavior

Kafka producer and consumer have a common logic to resolve a node with slight variations
https://github.com/build-trust/ockam/issues/5087

## Proposed changes

Refactored to avoid the code duplication

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/kafka/util.rs
+++ b/implementations/rust/ockam/ockam_command/src/kafka/util.rs
@@ -1,0 +1,107 @@
+use std::net::SocketAddr;
+
+use colorful::Colorful;
+use tokio::{sync::Mutex, try_join};
+
+use ockam::{Context, TcpTransport};
+use ockam_api::nodes::models::services::{StartKafkaProducerRequest, StartServiceRequest};
+use ockam_api::port_range::PortRange;
+use ockam_core::api::Request;
+use ockam_multiaddr::MultiAddr;
+
+use crate::node::{get_node_name, NodeOpts};
+use crate::service::start::start_service_impl;
+use crate::terminal::OckamColor;
+use crate::util::process_nodes_multiaddr;
+use crate::{display_parse_logs, fmt_log, fmt_ok, CommandGlobalOpts};
+
+pub struct ArgOpts {
+    pub endpoint: String,
+    pub kafka_entity: String,
+    pub node_opts: NodeOpts,
+    pub addr: String,
+    pub bootstrap_server: SocketAddr,
+    pub brokers_port_range: PortRange,
+    pub project_route: MultiAddr,
+}
+
+pub async fn rpc(ctx: Context, (opts, args): (CommandGlobalOpts, ArgOpts)) -> crate::Result<()> {
+    let ArgOpts {
+        endpoint,
+        kafka_entity,
+        node_opts,
+        addr,
+        bootstrap_server,
+        brokers_port_range,
+        project_route,
+    } = args;
+
+    opts.terminal
+        .write_line(&fmt_log!("Creating {} service...\n", kafka_entity))?;
+
+    display_parse_logs(&opts);
+
+    let project_route = process_nodes_multiaddr(&project_route, &opts.state)?;
+
+    let is_finished = Mutex::new(false);
+    let send_req = async {
+        let tcp = TcpTransport::create(&ctx).await?;
+        let node_name = get_node_name(&opts.state, &node_opts.at_node);
+
+        let payload = StartKafkaProducerRequest::new(
+            bootstrap_server.to_owned(),
+            brokers_port_range,
+            project_route,
+        );
+        let payload = StartServiceRequest::new(payload, &addr);
+        let req = Request::post(endpoint).body(payload);
+        start_service_impl(&ctx, &opts, &node_name, &kafka_entity, req, Some(&tcp)).await?;
+
+        *is_finished.lock().await = true;
+
+        Ok::<_, crate::Error>(())
+    };
+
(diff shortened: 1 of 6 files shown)


implementations/rust/ockam/ockam_command/src/kafka/create.rs
```diff
@@ -23,10 +23,16 @@ use crate::{
...
+
+#[derive(Clone, Debug, Args)]
+pub struct ArgOpts {
+    pub endpoint: String,
+    pub kafka_entity: String,
+}
 
-/// Create a new Kafka Producer
 #[derive(Clone, Debug, Args)]
 pub struct CreateCommand {
```
adrianbenavides: We need a `CreateCommand` struct for both the producer and consumer commands so we can write their specific documentation. It's OK if we have to duplicate the commands arguments. The point here is to reuse the `rpc` function you have below.

That way, we'll have two separate commands with their own descriptions BUT with the same _internal_ logic.

adrianbenavides (changes requested):
Thanks @lper1582 for taking your time to work on this! The changes look great, but there is a small "code organization" issue to solve before we can merge it.

Also, don't forget to run `cargo clippy` and `cargo fmt` before pushing your changes to make sure the rust compiler is happy!

adrianbenavides:
In the meantime, if you could also submit a PR to the ockam-contributors crate as explained here:

> I have accepted the Ockam [Contributor License Agreement](https://github.com/build-trust/ockam-contributors/blob/main/CLA.md) by adding my Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/blob/main/CONTRIBUTORS.csv) file in a separate pull request to the [build-trust/ockam-contributors](https://github.com/build-trust/ockam-contributors) repository. The easiest way to do this is to [edit the CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/edit/main/CONTRIBUTORS.csv) file in the github web UI and create a PR, this will mark the commit as verified

Once you do that, we'll be able to approve your PR's 👍

implementations/rust/ockam/ockam_command/src/kafka/create.rs
loghijiaha: Thanks for the comments. Will do refactoring.

implementations/rust/ockam/ockam_command/src/kafka/util.rs
```diff
@@ -0,0 +1,117 @@
...
+use ockam_core::api::Request;
+use ockam_multiaddr::MultiAddr;
+
+use crate::{CommandGlobalOpts, display_parse_logs, fmt_log, fmt_ok};
+use crate::node::{get_node_name, NodeOpts};
+use crate::service::start::start_service_impl;
+use crate::terminal::OckamColor;
+use crate::util::process_nodes_multiaddr;
+
+pub trait KafkaCreateCommand {
```
adrianbenavides: This is better but I think it's a little overkill. Traits are used to define shared behavior, where each type using that trait will have different implementation details. In this particular case, the trait is only used to map the struct fields into functions, so why not pass a struct with the needed fields to the `rpc` function from line 33? 

The main benefit you get from passing a struct instead of implementing a trait is that, if you want to modifiy (add/remove) the needed fields, you only have to do it in one place, instead of doing it in the trait AND every implementation. Another side-effect of traits, although not important in this scenario, is that they generate more code in the compiled binary (check [this out](https://blog.rust-lang.org/2015/05/11/traits.html#static-dispatch) if you want more details).

I hope this make sense!

loghijiaha: Good catch! I am already passing a struct `ArgOpts` which can be used to pass the params.

adrianbenavides (dismissed):
Looks perfect now, LGTM! Thanks @lper1582 for an awesome first contribution! 🚀

adrianbenavides:
@lper1582 looks like the email you used to sign the [CLA is different](https://github.com/build-trust/ockam-contributors/pull/231/files) from the one you used to sign your commits. 

I've squashed your commits into a single one, so you should `git pull --rebase` before pushing your changes again. If you have trouble resigning the commit, we can always close this PR and open a new one with a fresh commit. Let me know if you need any help!

loghijiaha:
@adrianbenavides Created a pr for my email changes [#234](https://github.com/build-trust/ockam-contributors/pull/234). Will that work with the current PR if we re-run the workflow?

adrianbenavides:
> @adrianbenavides Created a pr for my email changes [#234](https://github.com/build-trust/ockam-contributors/pull/234). Will that work with the current PR if we re-run the workflow?

Looks like that did the trick, thanks @lper1582 !

adrianbenavides (approved):

SanjoDeundiak (approved):

PR #5046, opened 2023-06-01
test(rust): add unit test for the medic session

Alef-gabriel:
## Description
introduces unit test for the functionality responsible for restarting the topology, including TCP connections, secure channels, etc., for Ockam Relays and Inlets.

## Changes Made

- Added integration tests for the restart mechanism.
- Simulated the monitoring of Sessions and periodic ping messages.
- Verified the detection of outages and the subsequent recreation of routes, Relays, and Inlets.

PR #5143, opened 2023-06-17
test(rust): add session monitoring unit test

Alef-gabriel:
## Current behavior

introduces unit test for the functionality responsible for restarting the topology, including TCP connections, secure channels, etc., for Ockam Relays and Inlets.

## Proposed changes
Added integration tests for the restart mechanism.
Simulated the monitoring of Sessions and periodic ping messages.
Verified the detection of outages and the subsequent recreation of routes, Relays, and Inlets.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/session/mod.rs
+++ b/implementations/rust/ockam/ockam_api/src/session/mod.rs
@@ -228,3 +228,97 @@ impl Worker for Collector {
         Ok(())
     }
 }
+
+use crate::echoer::Echoer;
+use core::sync::atomic::{AtomicBool, Ordering};
+use crate::hop::Hop;
+use ockam_core::{AsyncTryClone, Result};
+use crate::session::sessions::Session;
+
+#[ockam::test]
+async fn test_session_monitoring(ctx: &mut Context) -> Result<()> {
+    // Create a new Medic instance
+    let medic = Medic::new();
+
+    let sessions: Arc<Mutex<Sessions>> = medic.sessions();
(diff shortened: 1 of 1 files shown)


implementations/rust/ockam/ockam_api/src/session/mod.rs
```diff
@@ -228,3 +228,97 @@ impl Worker for Collector {
         Ok(())
     }
 }
+
```
adrianbenavides, line 231: All the code below this line should be behind the test flag:

```rust
#[cfg(test)]
mod tests {
    use crate::echoer::Echoer;
    ...
}
```

implementations/rust/ockam/ockam_api/src/session/mod.rs
```diff
@@ -228,3 +228,97 @@ impl Worker for Collector {
...
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+        continue;
+    }
+
+    // Shut down the test
+    medic_task.abort();
+    ctx.stop().await?;
+    Ok(())
+}
```
adrianbenavides: Make sure to run `cargo fmt`. I think there are a couple of formatting issues that should be fixed.

adrianbenavides (commented):
This is a retake of this one right? https://github.com/build-trust/ockam/pull/5046 I'm pinging @davide-baldo to dig into the test details!

I've left a couple of comments with minor details to fix.

etorreborre (approved):
Nice job, thanks!

PR #5149, opened 2023-06-20
fix(rust): improve error message listing a stopped node

### LINKED ISSUE #5146: Display a user-friendly error message when listing resources on a stopped node
### Observed behavior

The `list` commands should show a more descriptive error message when the node that is being queried is not running.

### Steps to reproduce

```sh
ockam node create n
ockam service list --at n # works as expected
ockam node stop n
ockam service list --at n # throws the error below

× connection peer was not found (origin: Transport, kind: Misuse, source location: /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/convert/mod.rs:727:9)
```

### Expected behavior

The error message should be user-friendly. Something like: `The node n is not running`.

### Ockam Version

0.89

### Proposed solution

Right after parsing the node name, check whether the node is running or not:

```rust
let node_name = parse_node_name(&node_name)?;
if !opts.state.nodes.get(&node_name)?.is_running() {
    return Err(miette!("The node {node_name} is not running"));
}
```

### Commands to fix

If you want to help with this issue, feel free to pick one of the commands below and apply the fix!

- [ ] `ockam node list`
- [ ] `ockam service list`
- [ ] `ockam tcp listener list`
- [ ] `ockam credential list`
- [ ] `ockam identity list`
- [ ] `ockam lease list`
- [ ] `ockam policy list`
- [ ] `ockam project list`
- [ ] `ockam project addon list`
- [ ] `ockam relay list`
- [ ] `ockam secure-channel list`
- [ ] `ockam space list`
- [ ] `ockam trust-context list`
- [ ] `ockam vault list`
- [ ] `ockam worker list`
- [ ] `ockam kafka-consumer list`
- [ ] `ockam kafka-producer list`

BradLewis:
## Current behavior

Currently when you try to list a stopped node, you get the obscure error
```shell
$ ockam service list --at my_stopped_node
OCK500 connection peer was not found (origin: Transport, kind: Misuse, source location: /rustc/90c541806f23a127002de5b4038be731ba1458ca/library/core/src/convert/mod.rs:727:9)

       Version 0.89.0, hash: 11a2724331ecd51a1640a543930dd61a940a9a24
```

## Proposed changes

Will now check the error to see if it is due to the node being stopped. If it is, will instead return the message.

```shell
$ ockam service list --at my_stopped_node
  × The node 'my_stopped_node' is not running

         Version 0.89.0, hash: 1ba8e7b57506c9e2853c14187fd23e123e2e62a2

```

Issue is here https://github.com/build-trust/ockam/issues/5146

Note this will also slightly change the error for when a user requests a node that doesn't exist from 

```shell
$ ockam service list --at my_nonexistent_node
OCK500 not found

       Version 0.89.0, hash: 11a2724331ecd51a1640a543930dd61a940a9a24

``` 

to

```shell
$ ockam service list --at my_nonexistent_node
  × not found

         Version 0.89.0, hash: 4c5009ac04d11da27d9286e2930a91ce1f2f1fe7
``` 

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/service/list.rs
+++ b/implementations/rust/ockam/ockam_command/src/service/list.rs
@@ -2,9 +2,11 @@ use std::fmt::Write;
 
 use clap::Args;
 use colorful::Colorful;
+use miette::miette;
 use miette::IntoDiagnostic;
 use ockam::{Context, TcpTransport};
 
+use ockam_api::cli_state::StateDirTrait;
 use ockam_api::nodes::models::services::{ServiceList, ServiceStatus};
 use tokio::sync::Mutex;
 use tokio::try_join;
@@ -43,6 +45,10 @@ async fn run_impl(
 ) -> miette::Result<()> {
     let node_name = get_node_name(&opts.state, &cmd.node_opts.at_node);
     let node_name = parse_node_name(&node_name)?;
+
+    if !opts.state.nodes.get(&node_name)?.is_running() {
+        return Err(miette!("The node '{}' is not running", node_name));
+    }
 
     let tcp = TcpTransport::create(ctx).await.into_diagnostic()?;
     let mut rpc = RpcBuilder::new(ctx, &opts, &node_name).tcp(&tcp)?.build();
--- a/implementations/rust/ockam/ockam_command/src/kafka/consumer/list.rs
+++ b/implementations/rust/ockam/ockam_command/src/kafka/consumer/list.rs
@@ -1,10 +1,12 @@
 use crate::node::{get_node_name, initialize_node_if_default, NodeOpts};
 use crate::util::{node_rpc, parse_node_name, Rpc};
 use crate::{docs, fmt_err, CommandGlobalOpts};
+use miette::miette;
 
 use clap::Args;
 use colorful::Colorful;
 
+use ockam_api::cli_state::StateDirTrait;
 use ockam_api::nodes::models;
 
 use ockam_api::DefaultAddress;
@@ -33,6 +35,11 @@ async fn run_impl(
 ) -> miette::Result<()> {
     let node_name = get_node_name(&opts.state, &cmd.node_opts.at_node);
     let node_name = parse_node_name(&node_name)?;
+
+    if !opts.state.nodes.get(&node_name)?.is_running() {
+        return Err(miette!("The node '{}' is not running", node_name));
+    }
+
     let mut rpc = Rpc::background(&ctx, &opts, &node_name)?;
     rpc.request(Request::get(format!(
         "/node/services/{}",
--- a/implementations/rust/ockam/ockam_command/src/kafka/producer/list.rs
+++ b/implementations/rust/ockam/ockam_command/src/kafka/producer/list.rs
@@ -1,10 +1,12 @@
 use crate::node::{get_node_name, initialize_node_if_default, NodeOpts};
 use crate::util::{node_rpc, parse_node_name, Rpc};
 use crate::{docs, fmt_err, CommandGlobalOpts};
+use miette::miette;
 
 use clap::Args;
 use colorful::Colorful;
 
+use ockam_api::cli_state::StateDirTrait;
 use ockam_api::nodes::models;
 
 use ockam_api::DefaultAddress;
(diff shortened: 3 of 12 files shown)


implementations/rust/ockam/ockam_command/src/service/list.rs
```diff
@@ -67,21 +68,31 @@ async fn run_impl(
...
-    )?;
-    let json = serde_json::to_string_pretty(&services.list).into_diagnostic()?;
-    opts.terminal
-        .stdout()
-        .plain(plain)
-        .json(json)
-        .write_line()?;
-
-    Ok(())
+    let result = try_join!(send_req, progress_output);
```
adrianbenavides: We have a small utilty function to detect whether a node is running or not. Right after line 45 you can use the following snippet:

```rust
let node_name = parse_node_name(&node_name)?;
if !opts.state.nodes.get(&node_name)?.is_running() {
    return Err(miette!("The node {node_name} is not running"));
}
```

That way you don't have to check for specific errors/error messages.

adrianbenavides (commented):
Hey @BradLewis, thanks for submitting this PR! I've left a comment about how to detect if a node is not running querying the node's state. Hope it helps, and let me know if you have any question!

implementations/rust/ockam/ockam_command/src/service/list.rs
BradLewis: Ah that's much nicer! Thanks I'll make those changes.

BradLewis:
I believe I've updated all of the `list` commands that I could find now. 
I also squashed all my commits into one this time given the changes were pretty small, but going forward, do you guys prefer  a single commit or just many smaller commits?

BradLewis:
Hi Adrian, I have a couple of quick questions regarding the changes you made on the linked issue.
1. Is there anything I can do regarding the build issue? Seems to just be a library that needs updating.
2. I've updated all the commands that support the `--at` flag, however some of the commands listed, such as `ockam node list` don't support it, so I'm not sure there's anything needed to be done there.
3. Currently I don't parse the node name through the `parse_node_name` function due to it being fallible, and I didn't want to change existing functionality where possible, and some functions use the non-parsed node name. Would it be better to run the name through `parse_node_name` before checking if a node is running?

Thanks!

adrianbenavides:
> Hi Adrian, I have a couple of quick questions regarding the changes you made on the linked issue.
> 
> 1. Is there anything I can do regarding the build issue? Seems to just be a library that needs updating.

Nope, don't worry about that, we are looking into that already and it's external to this PR.

> 2. I've updated all the commands that support the `--at` flag, however some of the commands listed, such as `ockam node list` don't support it, so I'm not sure there's anything needed to be done there.

You're correct, I added all the list commands, but only those with the `--at` flag are affected by the bug described at the issue.

> 3. Currently I don't parse the node name through the `parse_node_name` function due to it being fallible, and I didn't want to change existing functionality where possible, and some functions use the non-parsed node name. Would it be better to run the name through `parse_node_name` before checking if a node is running?

Ideally, we'd like to use the `parse_node_name` in all cases so that the user can write `/node/n` or `n` as the argument. So, if you have spotted any command where we are not doing it already, it would awesome if you apply that fix as well.

BradLewis:
> Ideally, we'd like to use the `parse_node_name` in all cases so that the user can write `/node/n` or `n` as the argument. So, if you have spotted any command where we are not doing it already, it would awesome if you apply that fix as well.

Perfect, thanks for the clarifications. I'll make those changes after work this evening.

adrianbenavides:
> > Ideally, we'd like to use the `parse_node_name` in all cases so that the user can write `/node/n` or `n` as the argument. So, if you have spotted any command where we are not doing it already, it would awesome if you apply that fix as well.
> 
> Perfect, thanks for the clarifications. I'll make those changes after work this evening.

Whenever you can, thanks! 🙏

BradLewis:
I added `parse_node_name` for the `tcp-listener list --at` command as it was missing, however for the rest of them that aren't using `parse_node_name`, they're using the function `extract_address_value` which as far as I can tell does basically the same thing but supports more than just nodes. Not sure if we want to update them to just use `parse_node_name` instead since `--at` only supports nodes, but if not then all should be good here.

adrianbenavides:
> they're using the function `extract_address_value` which as far as I can tell does basically the same thing but supports more than just nodes.

That's exactly right, but don't worry about that. I think that's a little out of the scope for this PR and we can tackle that anytime.

adrianbenavides:
One last thing before we can merge your PR, we require all commits to be signed as described here:

> All commits in this Pull Request are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and Verified by Github.

Once you have your keys setup, you will have to (probably squash and) force-push to sign your commits. If you need any help let me know!

BradLewis:
> One last thing before we can merge your PR, we require all commits to be signed as described here:

Thanks, I'm not sure why these previous commits weren't signed but that's fixed now. I've also squashed the 2 commits into 1.

Edit: Ah actually it might be due to me rebasing the branch onto develop through the PR itself rather than the CLI. Either way it's fixed now.

adrianbenavides (approved):
Ready to merge! Thanks again @BradLewis for helping out with this and congrats on your first contribution to Ockam!

PR #5151, opened 2023-06-20
rename release workflows

metaclips:
Rename release workflows


tools/scripts/release/release-rollback.sh
```diff
@@ -5,11 +5,27 @@ set -ex
...
-USER_TYPE="orgs"
+export BASH_XTRACEFD=3
+
+if [[ -z $OWNER ]]; then
+  OWNER="build-trust"
+fi
+
+USER_TYPE="users"
+
+user_type=$(gh api \
```
snandam: small nitpick. For the same of readability. Instead of using `user_type` lowercase can you name it `type` as it matches the output you are looking at anyways? Since you are already using  uppercase `USER_TYPE` variable just above this line.

metaclips: Fixed, thanks

snandam (approved):

nazmulidris (approved):

PR #5154, opened 2023-06-20
refactor(rust): use better names for symmetric keys involved in secure channels

etorreborre:
The current names are confusing since they make as if asymmetric encryption was used.


etorreborre:
Closing because those names were not necessarily helping either.

PR #5155, opened 2023-06-20
validate credentials of receiving identity for outgoing messages in inlets and outlets

davide-baldo:
Currently, we don't validate outgoing messages from inlets or outlets. The impact is vastly reduced by the fact that inlets/outlet communication always requires a ping/pong handshake, and this will force validation of the incoming message once during initialization. But after that, the communication will keep going regardless of the remote credential validity, this PR is aimed to fix specifically this case.

The approach used is to update the outgoing access control when the first message is received, capturing the source local_info.

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/context/context.rs
+++ b/implementations/rust/ockam/ockam_node/src/context/context.rs
@@ -77,6 +77,11 @@ impl Context {
         &self.mailboxes
     }
 
+    /// Returns a mutable reference to the mailboxes of this context
+    pub fn mailboxes_mut(&mut self) -> &mut Mailboxes {
+        &mut self.mailboxes
+    }
+
     /// Shared [`FlowControls`] instance
     pub fn flow_controls(&self) -> &FlowControls {
         &self.flow_controls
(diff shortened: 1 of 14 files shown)


implementations/rust/ockam/ockam_node/src/context/context.rs
```diff
@@ -77,6 +77,11 @@ impl Context {
         &self.mailboxes
     }
 
+    /// Returns a mutable reference to the mailboxes of this context
+    pub fn mailboxes_mut(&mut self) -> &mut Mailboxes {
```
SanjoDeundiak, line 81: I'm not sure modifying mailboxes while the worker is running (especially when the worker itself should do it) is a good idea, it's very easy to shoot yourself in the foot by, for example, creating a race condition. Is there a way to set up mailboxes from the beginning?

implementations/rust/ockam/ockam_transport_tcp/src/portal/outlet_listener.rs
```diff
@@ -92,6 +93,8 @@ impl Worker for TcpOutletListenWorker {
             return_route.clone(),
             addresses.clone(),
             self.options.incoming_access_control.clone(),
+            self.options.outgoing_access_control_factory.clone(),
```
SanjoDeundiak, line 96: Can't we instantiate the outgoing access control at this point already?

SanjoDeundiak (changes requested):

davide-baldo:
Needs heavy changes for a new design

PR #5160, opened 2023-06-21
feat(rust): update ockam api services error responses to using a struct

caoakleyii:
## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/flow_controls.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/flow_controls.rs
@@ -14,14 +14,7 @@ impl NodeManagerWorker {
         req: &Request<'_>,
         dec: &mut Decoder<'_>,
     ) -> Result<ResponseBuilder, ResponseBuilder<Error>> {
-        let request: AddConsumer = match dec.decode() {
-            Ok(r) => r,
-            Err(e) => {
-                let err_body =
-                    Error::new(req.path()).with_message(format!("Unable to decode request: {}", e));
-                return Err(Response::bad_request(req.id()).body(err_body));
-            }
-        };
+        let request: AddConsumer = dec.decode()?;
 
         let mut route = match local_multiaddr_to_route(request.address()) {
             None => {
@@ -35,11 +28,12 @@ impl NodeManagerWorker {
         let addr = match route.step() {
             Ok(a) => a,
             Err(e) => {
-                let err_body = Error::new(req.path()).with_message(format!(
-                    "Unable to retrieve address from {}. {}",
-                    request.address(),
-                    e
-                ));
+                let err_body = Error::new(req.path())
+                    .with_message(format!(
+                        "Unable to retrieve address from {}.",
+                        request.address(),
+                    ))
+                    .with_cause(Error::new(req.path()).with_message(e.to_string()));
                 return Err(Response::bad_request(req.id()).body(err_body));
             }
         };
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/flow_controls.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/flow_controls.rs
@@ -1,7 +1,7 @@
 use crate::local_multiaddr_to_route;
 use crate::nodes::models::flow_controls::AddConsumer;
 use minicbor::Decoder;
-use ockam_core::api::{Request, Response, ResponseBuilder};
+use ockam_core::api::{Error, Request, Response, ResponseBuilder};
 use ockam_core::Result;
 use ockam_node::Context;
 
@@ -13,16 +13,40 @@ impl NodeManagerWorker {
         ctx: &Context,
         req: &Request<'_>,
         dec: &mut Decoder<'_>,
-    ) -> Result<ResponseBuilder> {
-        let request: AddConsumer = dec.decode()?;
+    ) -> Result<ResponseBuilder, ResponseBuilder<Error>> {
+        let request: AddConsumer = match dec.decode() {
+            Ok(r) => r,
+            Err(e) => {
+                let err_body = Error::new(req.path())
+                    .with_message(format!("Unable to decode request: {}", e));
+                return Err(Response::bad_request(req.id()).body(err_body));
+            }
+        };
 
         let mut route = match local_multiaddr_to_route(request.address()) {
-            None => return Ok(Response::bad_request(req.id())),
+            None => {
+                let err_body = Error::new(req.path())
+                    .with_message(format!("Invalid address: {}", request.address()));
+                return Err(Response::bad_request(req.id()).body(err_body));
+            }
             Some(r) => r,
         };
-        let addr = route.step()?;
+
+        let addr = match route.step() {
+            Ok(a) => a,
+            Err(e) => {
+                let err_body = Error::new(req.path()).with_message(format!(
+                    "Unable to retrieve address from {}. {}",
+                    request.address(),
+                    e
+                ));
+                return Err(Response::bad_request(req.id()).body(err_body));
+            }
+        };
         if !route.is_empty() {
-            return Ok(Response::bad_request(req.id()));
+            let err_body = Error::new(req.path())
+                .with_message(format!("Invalid address: {}", request.address()));
+            return Err(Response::bad_request(req.id()).body(err_body));
         }
 
         ctx.flow_controls()
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/flow_controls.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/flow_controls.rs
@@ -17,8 +17,8 @@ impl NodeManagerWorker {
         let request: AddConsumer = match dec.decode() {
             Ok(r) => r,
             Err(e) => {
-                let err_body = Error::new(req.path())
-                    .with_message(format!("Unable to decode request: {}", e));
+                let err_body =
+                    Error::new(req.path()).with_message(format!("Unable to decode request: {}", e));
                 return Err(Response::bad_request(req.id()).body(err_body));
             }
         };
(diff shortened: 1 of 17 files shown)


implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -150,6 +150,12 @@ impl<'a> Rpc<'a> {
                 // Overwrite error to swallow inner cause and hide it from end-user
                 miette!("The request timed out, please make sure the command's arguments are correct or try again")
             })?;
+
+        if self.is_ok().is_err() {
+            let err: Error = self.parse_response()?;
+            let err_msg = err.message().unwrap_or_default().to_string();
```
adrianbenavides, line 156: Should we add a default error string in case the error message is empty?

implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -171,6 +177,12 @@ impl<'a> Rpc<'a> {
                 // Overwrite error to swallow inner cause and hide it from end-user
                 miette!("The request timed out, please make sure the command's arguments are correct or try again")
             })?.body();
+
+        if self.is_ok().is_err() {
+            let err: Error = self.parse_response()?;
+            let err_msg = err.message().unwrap_or_default().to_string();
```
adrianbenavides, line 183: Same here:

> Should we add a default error string in case the error message is empty?

implementations/rust/ockam/ockam_core/src/api.rs
```diff
@@ -387,20 +389,30 @@ pub struct Error<'a> {
...
             #[cfg(feature = "tag")]
             tag: TypeTag,
             method: None,
-            path: Some(path.into()),
+            path: Some(path.to_string()),
+            message: None,
+        }
+    }
+
+    pub fn new_without_path() -> Self {
```
adrianbenavides, line 415: This feels a little counter-intuitive. I would expect to have a `fn new()` and then a `fn with_path(path)`.

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
```diff
@@ -342,7 +368,13 @@ impl NodeManagerWorker {
             .await
         {
             Ok(_) => Ok(Response::ok(req.id())),
-            Err(_err) => Ok(Response::bad_request(req.id())),
+            Err(err) => {
+                let err_body = Error::new(req.path()).with_message(format!(
+                    "Unable to stop listener {}. {}",
+                    listener_address, err
```
adrianbenavides: A general comment regarding how to handle the internal `err`'s. Since we don't have an entire control over the content of the errors, perhaps it would be a better idea to store the `err` as the `cause` as a separate variable so we can handle it separately from the main error message in the command layer.

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
```diff
@@ -99,17 +100,13 @@ impl NodeManagerWorker {
         req: &Request,
         tcp: &TcpTransport,
         address: String,
-    ) -> ResponseBuilder<TransportStatus> {
-        let sender = match Self::find_connection(tcp, address) {
+    ) -> Result<ResponseBuilder<TransportStatus>, ResponseBuilder<Error>> {
+        let sender = match Self::find_connection(tcp, address.to_string()) {
```
adrianbenavides, line 104: This function is only used here and returns an Option. Perhaps it would be better if we handle the `not found` error within that function, so callers don't have to.

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
```diff
@@ -176,17 +168,24 @@ impl NodeManagerWorker {
...
         let node_manager = self.node_manager.read().await;
-        let CreateTcpConnection { addr, .. } = dec.decode()?;
+        let CreateTcpConnection { addr, .. } = match dec.decode() {
+            Ok(it) => it,
+            Err(err) => {
+                let err_body = Error::new(req.path())
+                    .with_message(format!("Unable to decode request: {}", err));
+                return Err(Response::bad_request(req.id()).body(err_body));
+            }
+        };
```
adrianbenavides: I wonder if it would be better to encapsulate the `dec.decode()` call in a function that wraps the result into a Result<> to avoid all this boilerplate.

implementations/rust/ockam/ockam_api/src/nodes/service/node_services.rs
```diff
@@ -647,17 +655,21 @@ impl NodeManagerWorker {
...
+                listener_address,
+                body_req.bootstrap_server_addr.ip(),
+                body_req.bootstrap_server_addr.port(),
+                body_req.brokers_port_range(),
+                outlet_node_multiaddr,
+                KafkaServiceKind::Producer,
+            )
+            .await
+        {
+            return Ok(e.to_vec()?);
```
adrianbenavides, line 671: Why are we returning an Ok in cases like this? What is that error converted to? I don't understand how are we handling this from the command layer.

implementations/rust/ockam/ockam_api/src/nodes/service/forwarder.rs
```diff
@@ -157,8 +163,9 @@ impl NodeManagerWorker {
             )
         } else {
             error!(%remote_address, "Forwarder not found in the node registry");
-
-            Ok(Response::not_found(req.id()).body(None))
+            let err_body = Error::new(req.path())
+                .with_message(format!("Forwarder, {}, not found.", remote_address));
```
adrianbenavides: Perhaps could be renamed to `Forwarder with address {} not found.`? I find it a little confusing with those commas there.

implementations/rust/ockam/ockam_api/src/nodes/service/flow_controls.rs
```diff
@@ -13,16 +13,40 @@ impl NodeManagerWorker {
...
+                return Err(Response::bad_request(req.id()).body(err_body));
+            }
+        };
 
         let mut route = match local_multiaddr_to_route(request.address()) {
-            None => return Ok(Response::bad_request(req.id())),
+            None => {
+                let err_body = Error::new(req.path())
+                    .with_message(format!("Invalid address: {}", request.address()));
+                return Err(Response::bad_request(req.id()).body(err_body));
```
adrianbenavides, line 23: IIUC, it doesn't matter now whether we return an Ok(Response) or an Err(Reponse) because they are both unwrapped when calling `encode_request_result` right?

Would it be the same if we just refactor this unwrapping? https://github.com/build-trust/ockam/blob/9dd38a7c399ba28664bec13b0340d44eb0891b0a/implementations/rust/ockam/ockam_api/src/nodes/service.rs#L909

adrianbenavides (commented):
Looks great! I've left some comments with some potential improvements. We could tackle them in a separate PR if that makes more sense to you.

implementations/rust/ockam/ockam_core/src/api.rs
caoakleyii: I can understand that, however, most error cases the path is known and used.

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
caoakleyii: I should update these, I implemented the From trait for the decode error; these can just use the `?` operator

implementations/rust/ockam/ockam_api/src/nodes/service/node_services.rs
caoakleyii: This is an Ok() because the result is a `Vec<u8>` which will in this line of code be the Response with an error body serialized.

The only propagated error would be if we couldn't serialize the response to a vec<u8>, which means we won't be able to send it back to the command.

implementations/rust/ockam/ockam_api/src/nodes/service/flow_controls.rs
caoakleyii: That would be ideal if everything returned a `ResponseBuilder<T>` within the services, however there's a large handful of services being called here that only return a `Vec<u8>`

I don't think it would be the right step to change those services either, though this was something that I debated.

implementations/rust/ockam/ockam_command/src/util/mod.rs
caoakleyii: I'm not entirely sure what a default error message would look like here.

It could be anything from an internal error, to a resource not found, or authorization issue.

The calling code, at this point could provide more context as to what failed, but not why.

implementations/rust/ockam/ockam_command/src/util/mod.rs
caoakleyii: See comment above.

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
caoakleyii: Perhaps we bundle this as part of a future refactor? I do think the way we our api handlers & essentially our repository pattern interact, needs to be improved.

The `find_x` functions should def throw an error in this case `NotFound`, that the API handlers should be able to  unwrap safely with `?`

But I don't think the `find_x` functions should be returning a ResponseBuilder, if they're primary function is to be a DAL

implementations/rust/ockam/ockam_api/src/nodes/service/flow_controls.rs
adrianbenavides: Ok, got it, good to know for the next iterations!

implementations/rust/ockam/ockam_command/src/util/mod.rs
adrianbenavides: > The calling code, at this point could provide more context as to what failed, but not why.

Yeah, that makes sense.

implementations/rust/ockam/ockam_core/src/api.rs
adrianbenavides: My concern here is having a 50/50 builder-like API... maybe we can revisit it in a future iteration.

implementations/rust/ockam/ockam_api/src/nodes/service/node_services.rs
adrianbenavides: Got it!

adrianbenavides (approved):
🚀

PR #5162, opened 2023-06-21
stop encryptor and decryptor workers together

Alef-gabriel:
## Current behavior
Currently, when the Encryptor is stopped, only the Encryptor address is returned to the client, leaving the Decryptor Worker running. This causes an inconsistency where one component is stopped while the other remains active.

## Proposed changes
In this pull request, we propose modifying the Encryptor's shutdown function to also stop the Decryptor Worker, ensuring both components are stopped together and write a test to prove it.

### DIFF
--- a/implementations/rust/ockam/ockam_identity/tests/channel.rs
+++ b/implementations/rust/ockam/ockam_identity/tests/channel.rs
@@ -1049,3 +1049,78 @@ async fn test_channel_delete_ephemeral_keys(ctx: &mut Context) -> Result<()> {
 
     Ok(())
 }
+
+#[allow(non_snake_case)]
+#[ockam_macros::test]
+async fn should_stop_encryptor__and__decryptor__in__secure_channel(
+    ctx: &mut Context,
+) -> Result<()> {
+    let secure_channels = secure_channels();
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_identity/tests/channel.rs
```diff
@@ -1049,3 +1049,78 @@ async fn test_channel_delete_ephemeral_keys(ctx: &mut Context) -> Result<()> {
 
     Ok(())
 }
+
+#[allow(non_snake_case)]
+#[ockam_macros::test]
+async fn should_stop_encryptor__and__decryptor__in__secure_channel(
```
etorreborre, line 1055: We don't have a strong enforcement of names in our unit tests at the moment but my preference would be to

 - start a test function with `test` (we have this in many places)
 - don't try to use `__` in the function names so that: `#[allow(non_snake_case)]` is not necessary. I think that in most cases the name reads just as fine

This is just my preference :-)

etorreborre (approved):
Nice job! That makes me realize that we are actually never shutting down workers on the responder side. I have created an issue for this: https://github.com/build-trust/ockam/issues/5164, you might want to tackle it next!

PR #5163, opened 2023-06-22
chore(rust): remove redundancies in markdown required and optional arguments

### LINKED ISSUE #5030: Remove hardcoded `required` and `optional` text from arguments' description
This is implicit from `<ARG>` (required) and `[ARG]` (optional), so we don't need to manually write that information as part of the arguments descriptions. An example of this can be found [here](https://github.com/build-trust/ockam/blob/663e686b2ffe60778292dcfdd4f17716bbc0fa21/implementations/rust/ockam/ockam_command/src/tcp/connection/create.rs#L31).

Here is the full list of commands that need to be fixed:

Remove the `required` string from:
- [ ] `ockam tcp-connection create`
- [ ] `ockam secure-channel create`
- [ ] `ockam secure-channel-listener delete`
- [ ] `ockam worker list`

Remove the `optional` string from:
- [ ] `ockam node create`
- [ ] `ockam authority create`
- [ ] `ockam relate create`

### LINKED ISSUE #5158: Add an "optional" tag to commands arguments' description when generating markdown docs
### Current behavior

Although there is an implicit differentiation between required/optional args (`<ARG>` for required args and `[ARG]` for optional), but we can't assume every user knows that. This is done automatically when showing the `--help` in the terminal, but not in the markdown docs (all arguments are shown as `<ARG>`), which creates an inconsistency.

We need to improve the markdown docs generation to include that information more explicitly. 

### Desired behavior

Related to https://github.com/build-trust/ockam/issues/5030 (they can be completed in parallel).

We should add a piece of text to optional arguments so the user can easily differentiate between required/optional arguments.

To do so, the `markdown` command must be updated as follows. In the [`generate_arg_markdown`](https://github.com/build-trust/ockam/blob/663e686b2ffe60778292dcfdd4f17716bbc0fa21/implementations/rust/ockam/ockam_command/src/markdown/mod.rs#L251) function, the `value_name` is always written as "\`<{value_name}>\`". This should be changed so that:
- if `arg.is_required_set()` is true, print it as "\`<{value_name}>\`"
- else, print it as "\`[{value_name}]\`, (optional)"

BradLewis:
## Current behavior

In the markdown documentation, it can be unclear to the user that `[ARG]` means optional and `<ARG>` means required. This PR ensures all optional args will be marked as `optional` automatically.
There are also some arguments that explicitly use optional or required, but after the first change this is now redundant and so they have been removed.

## Proposed changes

The proposed changes will change the generated markdown from this
image
to this
image

This closes the issues #5030 and #5158 

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/markdown/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/markdown/mod.rs
@@ -256,33 +256,38 @@ fn generate_arg_markdown(buffer: &mut Vec<u8>, arg: &clap::Arg) -> io::Result<()
         Some([]) => unreachable!(),
         None => arg.get_id().to_string().to_ascii_uppercase(),
     };
+    let (formatted_value_name, optional) = match arg.is_required_set() {
+        true => (format!("<{value_name}>"), ""),
+        false => (format!("[{value_name}]"), " (optional)"),
+    };
 
     match (arg.get_short(), arg.get_long()) {
         (Some(short), Some(long)) => {
             if arg.get_action().takes_values() {
-                write!(buffer, "`-{short}`, `--{long} <{value_name}>`")?
+                write!(buffer, "`-{short}`, `--{long} {formatted_value_name}`")?
             } else {
                 write!(buffer, "`-{short}`, `--{long}`")?
             }
         }
         (Some(short), None) => {
             if arg.get_action().takes_values() {
-                write!(buffer, "`-{short} <{value_name}>`")?
+                write!(buffer, "`-{short} {formatted_value_name}`")?
             } else {
                 write!(buffer, "`-{short}`")?
             }
         }
         (None, Some(long)) => {
             if arg.get_action().takes_values() {
-                write!(buffer, "`--{} <{value_name}>`", long)?
+                write!(buffer, "`--{} {formatted_value_name}`", long)?
             } else {
                 write!(buffer, "`--{}`", long)?
             }
         }
         (None, None) => {
-            write!(buffer, "`<{value_name}>`",)?;
+            write!(buffer, "`{formatted_value_name}`")?;
         }
     }
+    write!(buffer, "{optional}")?;
 
     if let Some(help) = arg.get_help() {
         writeln!(buffer, "<br/>")?;
--- a/implementations/rust/ockam/ockam_command/src/authority/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/authority/create.rs
@@ -70,15 +70,15 @@ pub struct CreateCommand {
     #[arg(group = "trusted", long, value_name = "PATH")]
     reload_from_trusted_identities_file: Option<PathBuf>,
 
-    /// Okta: URL used for accessing the Okta API (optional)
+    /// Okta: URL used for accessing the Okta API
     #[arg(long, value_name = "URL", default_value = None)]
     tenant_base_url: Option<String>,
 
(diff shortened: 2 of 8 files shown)


implementations/rust/ockam/ockam_command/src/markdown/mod.rs
```diff
@@ -256,31 +256,35 @@ fn generate_arg_markdown(buffer: &mut Vec<u8>, arg: &clap::Arg) -> io::Result<()
...
+    let formatted_value_name = match arg.is_required_set() {
+        true => format!("<{value_name}>"),
+        false => format!("[{value_name}], (optional)"),
+    };
 
     match (arg.get_short(), arg.get_long()) {
         (Some(short), Some(long)) => {
             if arg.get_action().takes_values() {
-                write!(buffer, "`-{short}`, `--{long} <{value_name}>`")?
+                write!(buffer, "`-{short}`, `--{long} {formatted_value_name}`")?
```
adrianbenavides, line 267: A small detail here: note that the `, (optional)` piece should be left out of the code block. Also, now that I take a second look at this, I think it feels better without the `,` before the `(optional)`. 

To avoid hacky stuff, maybe the cleanest approach would be to return two values so you can easily handle the code blocks:

```rust
let (value_name, optional) = match arg.is_required_set() {
    true => (format!("<{value_name}>"), ""),
    false => (format!("[{value_name}]"), "(optional)"),
};
```

```rust
write!(buffer, "`-{short}`, `--{long} {value_name}` {optional}")?
```

adrianbenavides (commented):
Looks great, thanks @BradLewis ! Just a small comment about the `(optional)` formatting.

implementations/rust/ockam/ockam_command/src/markdown/mod.rs
BradLewis: Fixed!

implementations/rust/ockam/ockam_command/src/markdown/mod.rs
```diff
@@ -256,31 +256,38 @@ fn generate_arg_markdown(buffer: &mut Vec<u8>, arg: &clap::Arg) -> io::Result<()
...
+                write!(buffer, "`--{} {formatted_value_name}`", long)?
             } else {
                 write!(buffer, "`--{}`", long)?
             }
         }
         (None, None) => {
-            write!(buffer, "`<{value_name}>`",)?;
+            write!(buffer, "`{formatted_value_name}` {optional}")?;
         }
     }
```
adrianbenavides: I was testing this and I noticed a pattern. The `(optional)` text should be in all `write!()` calls, regardless of whether we are showing the `formatted_value_name` or not. It would seem that the `(optional)` text is actually a constant in every branch, so we could actually have the following:

```rust
match (arg.get_short(), arg.get_long()) {
    // You just include the `<{value_name}>` by 
    // `{formatted_value_name}` replacement where needed
    ...
}
// Note that the `optional` variable should now include the initial whitespace
write!(buffer, "{optional}")?;
```

Let me know if that makes sense to you!

BradLewis: Makes sense to me! I've made those changes and it's gone from 
image
to
image

davide-baldo (approved):
Looks good to me! thanks!

davide-baldo:
I'll merge it as soon as the code freeze is lifted!

PR #5166, opened 2023-06-23
feat(rust): add flag to control whether a node redirects the logs to a file

adrianbenavides:
A couple of problems were found on the `node create` command: the first is a bug about how we handle node logs files; the second is an inconsistency about the expected behaviour when passing the `foreground` flag or not.

## Writing to logs

When creating a fresh foreground node with logging enabled (i.e. passing the `-v` flag), it won't persist the logs into files because the logger setup is done before the node state is created:

https://github.com/build-trust/ockam/blob/d4d921fc5a503a7e38c9a4869662da2abf005717/implementations/rust/ockam/ockam_command/src/lib.rs#L428-L440

If the node is stopped (e.g. via ctrl+c, or EOF), and the node is recreated again using `ockam node create ...`, it will start writing the logs into the files beacuse the node state is now accessible when setting up the logger. So, the first time a foreground node is created it won't write into files; subsequent `node create` calls will do.

There are two problems here:
1. We don't have a way to create a fresh foreground node that write the logs into files (you have to recreate it to do so)
2. We don't have a way to create a foreground node that doesn't write the logs into files

Proposed solutions:
1. A little tricky but doable: we just need a way to partially initialize the state (create the main node directory at least), so we can query the logs file path before actually creating the node state
2. Add a new argument to control whether we want to write logs into files or not (`--disable-file-logging`)

## Inconsistency of foreground/background nodes when recreating nodes

There was an inconsistency when creating a background node, where it would return an error when running the command for a stopped node. As explained before, this can be done with foreground nodes and it's a nice feature, since it doesn't force the user to learn/use the `ockam node start` command.

```sh
# The following is expected to work, recreating the node after it was stopped
ockam node create foreground -vv -f
ockam node stop foreground
ockam node create foreground -vv -f

# The following was returning an "Already exists" error
ockam node create background
ockam node stop background
ockam node create background
```

The second case has now the same behaviour as the first one.

### DIFF
--- a/implementations/rust/ockam/ockam_command/tests/bats/nodes_unit.bats
+++ b/implementations/rust/ockam/ockam_command/tests/bats/nodes_unit.bats
@@ -1,0 +1,93 @@
+#!/bin/bash
+
+# ===== SETUP
+
+setup() {
+  load load/base.bash
+  load_bats_ext
+  setup_home_dir
+}
+
+teardown() {
+  teardown_home_dir
+}
+
+# ===== TESTS
+
+@test "node - can recreate a background node after it was stopped" {
+  n="$(random_str)"
+  run "$OCKAM" node create "$n"
+  assert_success
+
+  # Fail to create a node with the same name
+  run "$OCKAM" node create "$n"
+  assert_failure
+
+  run "$OCKAM" node stop "$n"
+  assert_success
+
+  # Recreate node
+  run "$OCKAM" node create "$n"
+  assert_success
+}
+
+@test "node - can recreate a foreground node after it was stopped" {
+  n="$(random_str)"
+  $OCKAM node create $n -f &
+  sleep 0.1
+  run "$OCKAM" node show "$n"
+  assert_success
+
+  # Fail to create a node with the same name
+  run "$OCKAM" node create "$n" -f
+  assert_failure
+
+  run "$OCKAM" node stop "$n"
+  assert_success
+
+  # Recreate node
+  $OCKAM node create $n -f &
+  sleep 0.1
+  run "$OCKAM" node show "$n"
+  assert_success
+}
+
+@test "node - logs to file" {
+  n="$(random_str)"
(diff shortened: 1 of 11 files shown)


implementations/rust/ockam/ockam_command/tests/bats/nodes_unit.bats
```diff
@@ -0,0 +1,93 @@
...
+  $OCKAM node create $n -f &
+  sleep 0.1
+  run "$OCKAM" node show "$n"
+  assert_success
+
+  # Fail to create a node with the same name
+  run "$OCKAM" node create "$n" -f
+  assert_failure
+
+  run "$OCKAM" node stop "$n"
```
polvorin, line 45: good to add these tests cases :+1:  

for a future improvement,  would be good to kill -9  the running node instead/in addition of calling  `node stop`.  That will rule out the possibility of recreate working because of some cleanup done by a "clean" stop of the node instead of a machine restart, etc.

polvorin (approved):
:+1:

adrianbenavides:
⚠️  I'm leaving it as a draft after today's release.

implementations/rust/ockam/ockam_command/tests/bats/nodes_unit.bats
adrianbenavides: So, you mean manually killing the process and then try to recreate it right? If that process doesn't exist anymore then it should proceed as usual and recreate the node. I think that code is already there, I'll try to add the test!

polvorin: yes exactly but I'd say totally fine to not include that on this PR

adrianbenavides: I'll leave it for another PR. I have the tests in place but it looks like there is a problem on foreground nodes and they don't handle the situation correctly.

PR #5171, opened 2023-06-26
feat(rust): Implement upgrade and uninstall commands

### LINKED ISSUE #4916: Implement `ockam uninstall` and `ockam upgrade` commands
# Current

We have an install script that:
https://github.com/build-trust/ockam/blob/develop/install.sh

1. Downloads an architecture specific ockam release binary, by default for the latest released version on Github.
2. Installs the binary to `${install_path}/bin/ockam` an install path, which is by default `~/.ockam`
3. Adds an env file to `${install_path}/env` that sets up `$OCKAM_HOME`
4. Adds `${install_path}/env` to various shell specific rc files like bashrc etc. 

We also have a homebrew formula that installs ockam as part of homebrew:
https://github.com/build-trust/homebrew-ockam

We also have some logic in Ockam Command code that tells us there is a new release available:
https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/upgrade.rs

# Desired

Implement `ockam uninstall` command which would:
1. do the opposite steps of the install script if the install was done using the install script
2. run `brew uninstall ockam` if the install was done as part of homebrew

Implement `ockam upgrade` command which would:
1. Do some pre-upgrade steps
  i. for now this can list all running nodes and stop them.
2. Upgrade the `ockam` command binary with the version suggested in upgrade.json
3. Do some post-upgrade steps
  i. for now this can restart all the nodes that were stopped.

https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/lib.rs#L373

---

BradLewis:
I've created this as a draft PR for now to get your feedback on this approach. It probably needs more work with things like the error handling and the messages, but I figured it'd be worthwhile getting feedback on the more "happy" paths before working on making that better. Let me know what you think! 

## Current behavior 

Currently there is no way for the user to upgrade or uninstall `ockam` directly through the binary.

## Proposed changes

Resolves https://github.com/build-trust/ockam/issues/4916

This PR adds both an `upgrade` and `uninstall` command. 

### Upgrade

For the upgrade command it works as follows:
1. Checks to see if a new update is available by checking the github releases.
2. Stops all the running nodes.
3. Checks to see if it was installed via `homebrew`. If it was it runs `brew upgrade ockam`. It will then restart all the nodes it stopped in step 2.
4. If it was installed via the `install.sh` script it will check to see if it can find location of the script by looking at `$OCKAM_HOME` or `$HOME/.ockam`. If it can't find it then it stops the upgrade.
5. If all is okay to upgrade, it will delete the existing `ockam` binary.
6. It will then upgrade `ockam` using the `install.sh` script, specifying the specific version and the install path.
7. Once the upgrade is completed, it'll restart all the nodes that were stopped in step 2.

In order to do this I made some slight changes, namely writing the `install.sh` script to the $OCKAM_HOME folder, and add the `$OCKAM_HOME` variable to the environment variable files.

I also implemented some logic to fetch the latest versions from the releases on github rather than the `upgrade.json` file as I found that the `upgrade.json` file seems to be out of date? (`0.87.0` and `0.88.0` are empty, and `0.86.0` points to `0.87.0` etc). Not sure if this is intentional however.

### Uninstall

For the `uninstall` command, 
1. it will first check if it was uninstalled through the `install.sh` script or with `homebrew`. 
2. If it was installed with `homebrew`, it'll run `brew uninstall ockam`, otherwise it'll delete the `ockam` binary.
3. It will then clear all references to `$OCKAM_HOME` and then delete the `$OCKAM_HOME` directory. 

I'm not sure about this final step however as it'll delete all the metadata on nodes and whatnot the user has, which could be troublesome if they just wanted to do a simple uninstall/reinstall to fix an issue or something. Let me know what you think of that.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/upgrade.rs
+++ b/implementations/rust/ockam/ockam_command/src/upgrade.rs
@@ -125,22 +125,21 @@ fn start_nodes(stopped_nodes_names: &[String], opts: &CommandGlobalOpts) -> miet
         let node_state = opts.state.nodes.get(node_name)?;
         node_state.kill_process(false)?;
         let node_setup = node_state.config().setup();
-        let x = &node_setup.default_tcp_listener()?.addr.to_string();
         spawn_node(
             opts,
-            node_setup.verbose,
-            node_state.name(),
-            x,
-            None,
-            None,
-            None,
-            None,
-            None,
-            None,
-            None,
-            None,
-            None,
-            false,
+            node_setup.verbose, // Previously user-chosen verbosity level
+            node_name,          // The selected node name
+            &node_setup.api_transport()?.addr.to_string(), // The selected node api address
+            None,               // No project information available
+            None,               // No trusted identities
+            None,               // "
+            None,               // "
+            None,               // Launch config
+            None,               // Authority Identity
+            None,               // Credential
+            None,               // Trust Context
+            None,               // Project Name
+            node_setup.disable_file_logging,
         )?;
         opts.terminal
             .write_line(fmt_ok!("Restarted node {}", node_state.name()))?;
@@ -150,7 +149,7 @@ fn start_nodes(stopped_nodes_names: &[String], opts: &CommandGlobalOpts) -> miet
 
 fn upgrade_ockam(latest_version: &str, opts: &CommandGlobalOpts) -> miette::Result<()> {
     let stopped_nodes_names = stop_all_running_nodes(opts)?;
-    let installer = get_installer();
+    let installer = get_installer()?;
     let result = installer.upgrade(latest_version);
     // Try to restart nodes even if upgrade failed
     start_nodes(&stopped_nodes_names, opts)?;
--- a/implementations/rust/ockam/ockam_command/src/upgrade.rs
+++ b/implementations/rust/ockam/ockam_command/src/upgrade.rs
@@ -14,9 +14,9 @@ use crate::{
     fmt_info, fmt_ok, node::util::spawn_node, terminal::ConfirmResult, CommandGlobalOpts, Result,
 };
 
-pub fn check_if_an_upgrade_is_available() {
+pub fn check_if_an_upgrade_is_available(opts: &CommandGlobalOpts) {
     if upgrade_check_is_enabled() {
-        check_upgrade_sync(); // check if a new version has been released
+        check_upgrade_sync(opts); // check if a new version has been released
     }
 }
 
@@ -117,7 +117,6 @@ fn restart_nodes(stopped_nodes_names: &[String], opts: &CommandGlobalOpts) -> mi
         .write_line(fmt_info!("Restarting all stopped nodes"))?;
     for node_name in stopped_nodes_names.iter() {
         let node_state = opts.state.nodes.get(node_name)?;
-        node_state.kill_process(false)?;
         let node_setup = node_state.config().setup();
         spawn_node(
             &opts,
@@ -142,7 +141,7 @@ fn restart_nodes(stopped_nodes_names: &[String], opts: &CommandGlobalOpts) -> mi
 
 fn upgrade_ockam(opts: &CommandGlobalOpts) -> miette::Result<()> {
     let stopped_nodes_names = stop_all_running_nodes(opts)?;
-    let installer = get_installer()?;
+    let installer = get_installer(opts)?;
     let result = installer.upgrade();
     // Try to restart nodes even if upgrade failed
     restart_nodes(&stopped_nodes_names, opts)?;
--- a/implementations/rust/ockam/ockam_command/src/upgrade.rs
+++ b/implementations/rust/ockam/ockam_command/src/upgrade.rs
@@ -1,95 +1,27 @@
-use crate::util::installer::upgrade;
+use crate::util::{
+    github::{check_upgrade_sync, get_latest_release_version_sync},
+    installer::get_installer,
+    local_cmd,
+};
 use clap::{crate_version, Args};
 use colorful::Colorful;
 use miette::miette;
-use ockam::Context;
 use ockam_api::cli_state::{StateDirTrait, StateItemTrait};
 use ockam_core::env::get_env_with_default;
-use serde::Deserialize;
 use std::env;
-use tokio::runtime::Builder;
 
 use crate::{
-    fmt_info, fmt_ok, node::util::spawn_node, terminal::ConfirmResult, util::node_rpc,
-    CommandGlobalOpts, Result,
+    fmt_info, fmt_ok, node::util::spawn_node, terminal::ConfirmResult, CommandGlobalOpts, Result,
 };
-
-#[derive(Deserialize, Debug)]
-struct UpgradeFile {
-    upgrade_message: Option<String>,
-    upgrade_message_macos: Option<String>,
-}
 
 pub fn check_if_an_upgrade_is_available() {
     if !upgrade_check_is_disabled() {
-        // check if a new version has been released
-        Builder::new_current_thread()
-            .enable_all()
-            .build()
-            .unwrap()
-            .block_on(check());
-    }
-}
-
-async fn check() {
-    let url = format!(
-        "https://github.com/build-trust/ockam/releases/download/ockam_v{}/upgrade.json",
-        crate_version!()
-    );
-    let resp = reqwest::get(url).await;
-
-    if let Ok(r) = resp {
-        if let Ok(upgrade) = r.json::<UpgradeFile>().await {
-            if let Some(message) = upgrade.upgrade_message {
-                eprintln!("\n{}", message.yellow());
-
-                if cfg!(target_os = "macos") {
-                    if let Some(message) = upgrade.upgrade_message_macos {
-                        eprintln!("\n{}", message.yellow());
-                    }
-                }
-
-                eprintln!();
-            }
-        }
+        check_upgrade_sync(); // check if a new version has been released
     }
 }
 
 fn upgrade_check_is_disabled() -> bool {
     get_env_with_default("OCKAM_DISABLE_UPGRADE_CHECK", false).unwrap_or(false)
-}
-
-#[derive(Deserialize, Debug)]
-struct LatestRelease {
-    name: String,
-}
-
-impl LatestRelease {
-    fn version(&self) -> Result<&str> {
-        let result = self.name.split_once('v');
-        match result {
-            Some((_, version)) => Ok(version),
-            None => Err(miette!("Failed to get latest release version").into()),
-        }
-    }
-}
-
-async fn get_latest_release_version() -> Result<LatestRelease> {
-    let url = "https://api.github.com/repos/build-trust/ockam/releases/latest";
-    let client = reqwest::Client::new();
-    let resp = client
-        .get(url)
-        .header("Accept", "application/vnd.github+json")
-        .header("X-GitHub-Api-Version", "2022-11-28")
-        .header("User-Agent", "Ockam Command")
-        .send()
-        .await;
-    if let Ok(r) = resp {
-        if let Ok(release) = r.json::<LatestRelease>().await {
-            return Ok(release);
-        }
-    }
-    Err(miette!("Failed to get latest release").into())
 }
 
 #[derive(Clone, Debug, Args)]
@@ -107,15 +39,12 @@ pub struct UpgradeCommand {
 
 impl UpgradeCommand {
     pub fn run(self, opts: CommandGlobalOpts) {
-        node_rpc(run_impl, (opts, self));
+        local_cmd(run_impl(opts, self));
     }
 }
 
-async fn run_impl(
-    _ctx: Context,
-    (opts, cmd): (CommandGlobalOpts, UpgradeCommand),
-) -> miette::Result<()> {
-    let latest_release = get_latest_release_version().await?;
+fn run_impl(opts: CommandGlobalOpts, cmd: UpgradeCommand) -> miette::Result<()> {
+    let latest_release = get_latest_release_version_sync()?;
     let latest_version = latest_release.version()?;
     let current_version = crate_version!();
 
@@ -165,7 +94,7 @@ async fn run_impl(
         latest_version
     ))?;
 
-    upgrade_ockam(latest_version, &opts).await?;
+    upgrade_ockam(latest_version, &opts)?;
     opts.terminal
         .stdout()
         .plain(fmt_ok!("Ockam upgraded to version {}", latest_version))
@@ -211,6 +140,7 @@ fn start_nodes(stopped_nodes_names: &[String], opts: &CommandGlobalOpts) -> miet
             None,
             None,
             None,
+            false,
         )?;
         opts.terminal
             .write_line(fmt_ok!("Restarted node {}", node_state.name()))?;
@@ -218,9 +148,10 @@ fn start_nodes(stopped_nodes_names: &[String], opts: &CommandGlobalOpts) -> miet
     Ok(())
 }
 
-async fn upgrade_ockam(latest_version: &str, opts: &CommandGlobalOpts) -> miette::Result<()> {
+fn upgrade_ockam(latest_version: &str, opts: &CommandGlobalOpts) -> miette::Result<()> {
     let stopped_nodes_names = stop_all_running_nodes(opts)?;
-    let result = upgrade(latest_version).await;
+    let installer = get_installer();
+    let result = installer.upgrade(latest_version);
     // Try to restart nodes even if upgrade failed
     start_nodes(&stopped_nodes_names, opts)?;
     result
--- a/implementations/rust/ockam/ockam_command/src/upgrade.rs
+++ b/implementations/rust/ockam/ockam_command/src/upgrade.rs
@@ -15,13 +15,13 @@ use crate::{
 };
 
 pub fn check_if_an_upgrade_is_available() {
-    if !upgrade_check_is_disabled() {
+    if upgrade_check_is_enabled() {
         check_upgrade_sync(); // check if a new version has been released
     }
 }
 
-fn upgrade_check_is_disabled() -> bool {
-    get_env_with_default("OCKAM_DISABLE_UPGRADE_CHECK", false).unwrap_or(false)
+fn upgrade_check_is_enabled() -> bool {
+    !get_env_with_default("OCKAM_DISABLE_UPGRADE_CHECK", false).unwrap_or(false)
 }
 
 #[derive(Clone, Debug, Args)]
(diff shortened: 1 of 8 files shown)


mrinalwadhwa:
@BradLewis this is awesome, thank you for spending time on this. I'll try the commands and leave some thoughts soon

install.sh
```diff
@@ -190,8 +191,12 @@ download() {
   curl --proto '=https' --tlsv1.2 --location --silent --fail --show-error --output "$install_path/bin/ockam" "$_url"
   info "Downloaded ockam binary at the specified directory: $install_path/bin/ockam"
 
-  info "Granting permission to execute: chmod u+x $install_path/bin/ockam"
+  info "Downloading scripts"
+  curl --proto '=https' --tlsv1.2 --location --silent --show-error --output "$install_path/install.sh" "$_script_base_url/install.sh"
+
+  info "Granting permission to execute: chmod u+x $install_path/bin/ockam, $install_path/install.sh"
```
mrinalwadhwa: Instead of saving the old installer, I think it would be safer to always install a new version using the current install.sh in our git repo. This way we can tackle any breaking changes in the installers assumptions.

implementations/rust/ockam/ockam_command/src/util/installer.rs
```diff
@@ -0,0 +1,181 @@
...
+    let binary_path = std::env::current_exe().into_diagnostic()?;
+    std::fs::remove_file(binary_path)
+        .into_diagnostic()
+        .context("Failed to delete binary")
+}
+
+fn upgrade_binary(version: &str) -> miette::Result<()> {
+    let ockam_home = get_ockam_home()?;
+    let ockam_home = ockam_home.to_str().expect("Invalid ockam home");
+    delete_binary()?;
```
tigercat2000: Perhaps this should rename the old binary until the install script has successfully run, and then delete it? That way a failed upgrade doesn't leave the user with no ockam binary, which it could with this approach.

BradLewis: Yeah I was thinking about solving it like that as well, I'll make those changes.

install.sh
BradLewis: Yeah I like that idea! That'll save needing to have the installer local on the users computer too. I'll make those changes.

BradLewis:
Okay I've refactored the implementation quite a bit and improved the error handling as well. This split the upgrade and uninstall process for `homebrew` and binary installations into 2 separate logical flows. For `homebrew` it's pretty simple, just run the respective commands. For the binary installation it's a little more involved.

For upgrade:
1. Backup binary and environment files
2. Remove `ockam` lines in the environment files
3. Download installation script
4. Run the installation script to install the latest version and rewrite all the environment files.
If any point in the process fails, everything gets restored from backup

For installation it follows a similar pattern, with everything being backed up and restored if anything goes wrong.

Let me know what you think, and if you like this approach I'll open this as a non-draft PR 🙂

implementations/rust/ockam/ockam_command/src/upgrade.rs
```diff
@@ -1,51 +1,157 @@
...
-struct UpgradeFile {
-    upgrade_message: Option<String>,
-    upgrade_message_macos: Option<String>,
-}
+use crate::{
+    fmt_info, fmt_ok, node::util::spawn_node, terminal::ConfirmResult, CommandGlobalOpts, Result,
+};
 
 pub fn check_if_an_upgrade_is_available() {
     if !upgrade_check_is_disabled() {
```
etorreborre: Not your code but I would go with `if upgrade_check_is_enabled` rather than a double-negation :-)

implementations/rust/ockam/ockam_command/src/upgrade.rs
```diff
@@ -1,51 +1,157 @@
...
+            .plain(fmt_info!(
+                "You are already running the latest version of ockam: {}",
+                current_version
+            ))
+            .write_line()?;
+        return Ok(());
+    }
+
+    if cmd.check {
+        opts.terminal
```
etorreborre: We can refactor L.63 with L.73 since they are the same.

implementations/rust/ockam/ockam_command/src/upgrade.rs
```diff
@@ -1,51 +1,157 @@
...
+            ))
+            .write_line()?;
+        return Ok(());
+    }
+    opts.terminal.write_line(fmt_info!(
+        "A new version of ockam is available: {}. Current ockam version: {}",
+        latest_version,
+        current_version
+    ))?;
+    if !cmd.yes {
```
etorreborre, line 72: This might not be the right PR to do it but we could probably refactor all the `--yes` option behavior across commands. For example this is the same behavior for the `install` and the `upgrade` command

implementations/rust/ockam/ockam_command/src/upgrade.rs
```diff
@@ -1,51 +1,157 @@
...
+            node_state.kill_process(false)?;
+            opts.terminal
+                .write_line(fmt_ok!("Stopped node {}", node_state.name()))?;
+            stopped_nodes.push(node_state.name().to_string());
+        }
+    }
+    Ok(stopped_nodes)
+}
+
+fn start_nodes(stopped_nodes_names: &[String], opts: &CommandGlobalOpts) -> miette::Result<()> {
```
etorreborre: Since this function takes some `stopped_nodes` as parameters I think that `restart_nodes` would be a better name.

implementations/rust/ockam/ockam_command/src/upgrade.rs
```diff
@@ -1,51 +1,157 @@
...
+    opts.terminal
+        .write_line(fmt_info!("Restarting all stopped nodes"))?;
+    for node_name in stopped_nodes_names.iter() {
+        let node_state = opts.state.nodes.get(node_name)?;
+        node_state.kill_process(false)?;
```
etorreborre: Why stop the nodes again?

implementations/rust/ockam/ockam_command/src/util/github.rs
```diff
@@ -0,0 +1,130 @@
...
+        .header("X-GitHub-Api-Version", "2022-11-28")
+        .header("User-Agent", "Ockam Command")
+        .send()
+        .await;
+    if let Ok(r) = resp {
+        if let Ok(release) = r.json::<LatestRelease>().await {
+            return Ok(release);
+        }
+    }
+    Err(miette!("Failed to get latest release").into())
```
etorreborre: It would be good to at least log the reason why the latest release could not be accessed by logging the error response.

implementations/rust/ockam/ockam_command/src/util/github.rs
```diff
@@ -0,0 +1,130 @@
...
+    let url = format!(
+        "https://github.com/build-trust/ockam/releases/download/ockam_v{}/upgrade.json",
+        crate_version!()
+    );
+    let resp = reqwest::get(url).await;
+
+    if let Ok(r) = resp {
+        if let Ok(upgrade) = r.json::<UpgradeFile>().await {
+            if let Some(message) = upgrade.upgrade_message {
+                eprintln!("\n{}", message.yellow());
```
etorreborre: We need to use the `opts.terminal` instead of `eprintln` here.

implementations/rust/ockam/ockam_command/src/util/installer.rs
```diff
@@ -0,0 +1,411 @@
...
+                    fs::canonicalize(brew_ockam_path).into_diagnostic()?;
+                let current_exe = std::env::current_exe().into_diagnostic()?;
+                let canonical_current_exe = fs::canonicalize(current_exe).into_diagnostic()?;
+                if canonical_brew_ockam_path == canonical_current_exe {
+                    return Ok(true);
+                }
+            }
+            Ok(false)
+        }
+        Err(_) => Ok(false),
```
etorreborre: It would be good to print the return message of the `brew` command in case of an error.

implementations/rust/ockam/ockam_command/src/util/installer.rs
```diff
@@ -0,0 +1,411 @@
...
+fn remove_ockam_lines_from_files(files: &[String]) -> miette::Result<()> {
+    let home = home();
+    for file in files.iter() {
+        let path = home.join(file);
+        remove_okcam_lines_from_file(path)?;
+    }
+    Ok(())
+}
+
+fn remove_okcam_lines_from_file(file: PathBuf) -> miette::Result<()> {
```
etorreborre:
```suggestion
fn remove_ockam_lines_from_file(file: PathBuf) -> miette::Result<()> {
```

etorreborre (commented):
Hi @BradLewis I left a bunch of comments and suggestions for you to consider. Next I would like to test your PR a bit on my machine to double-check that it is also working for me. How have you tested it so far?

implementations/rust/ockam/ockam_command/src/upgrade.rs
BradLewis: This was part of the original requirements mentioned on the issue here https://github.com/build-trust/ockam/issues/4916.

implementations/rust/ockam/ockam_command/src/util/github.rs
BradLewis: This was also part of the original code that checked for the latest version from the `version.json` file on GitHub, I just moved it to this file. I can refactor this as part of this PR as well if you would like. 

I could also rework this to not use the `version.json` file and instead just pull the latest version from the GitHub releases?

implementations/rust/ockam/ockam_command/src/util/installer.rs
BradLewis: I've added an `eprintln!` statement for this

implementations/rust/ockam/ockam_command/src/upgrade.rs
etorreborre: That's not how I read it. Are referring to this line?
>  for now this can restart all the nodes that were stopped.

We have already killed the process of each node, it doesn't seem necessary to do it again.

implementations/rust/ockam/ockam_command/src/util/github.rs
etorreborre: > I can refactor this as part of this PR as well if you would like.

I see. Yes that would be nice. We are trying to move away from bare `printlns` in order to completely control the rendering in the terminal.

etorreborre: > I could also rework this to not use the version.json file and instead just pull the latest version from the GitHub releases?

I don't understand that part. Are you talking about the `upgrade.json` file? I actually tried to download it from `https://github.com/build-trust/ockam/releases/download/ockam_v0.90.0/upgrade.json` and it comes back empty. Is that the same for you too? Because then this means that we also need to fix that part somewhere in our release process (FYI @metaclips)

implementations/rust/ockam/ockam_command/src/util/installer.rs
etorreborre: Can you please use `opts.terminal` in that case too?

implementations/rust/ockam/ockam_command/src/util/github.rs
BradLewis: Yeah exactly. I mentioned it in the description for this PR, but I found that the `upgrade.json` were never really up-to-date. You can play around with it by changing the `version` in the `Cargo.toml` file. Currently I only no suggestions for `0.89.0`, `0.88.0` and `0.87.0`. If I set it to `0.86.0` then it informs me that there is an update available, but that the latest version is `0.87.0`.
image
Currently it pulls this information from the release of the version that the user is using, meaning these files would need to be updated every release to be effective. 
That's why I ended up just pulling the latest version from the release instead of just using the `version.json` file.

implementations/rust/ockam/ockam_command/src/upgrade.rs
BradLewis: Ah sorry I misunderstood what you meant.
I think that's a relic from earlier attempts where I was having a problem with the nodes not restarting correctly. I'll remove it

implementations/rust/ockam/ockam_command/src/util/installer.rs
BradLewis: Done! Let me know what you think. Will look like this for the upgrade check.
image

BradLewis:
> Hi @BradLewis I left a bunch of comments and suggestions for you to consider. Next I would like to test your PR a bit on my machine to double-check that it is also working for me. How have you tested it so far?

So for testing I've been setting the version to something lower, eg `0.88.0`, building the project and then moving the binary to the location I would like it to be. For testing with `brew`, I replace the binary that brew installs, for the script installation, I replace the binary there and for the standalone installation I just run it from the cargo build directory.

From there I just make sure all the nodes have restarted and if it was installed via the script, that all the environment files were updated correctly.

I will say this is definitely one that needs to be tested thoroughly given all the changes it can make on the end users machine. I split the logic into 3 flows (installation via `brew`, installation via the install script, and someone just downloading the binary from the GitHub releases), and then tried to infer which of the 3 were used, but there might be missed edge cases there.

One thing I haven't been able to test is the upgrade via `brew` as I'm not sure how to install an old version. But the uninstall command works as expected.

implementations/rust/ockam/ockam_command/src/upgrade.rs
```diff
@@ -1,51 +1,149 @@
...
+        return Ok(());
+    }
+
+    if !cmd.yes {
+        match opts.terminal.confirm(&fmt_info!(
+            "This will upgrade ockam to the latest version. Are you sure?"
+        ))? {
+            ConfirmResult::Yes => {}
+            ConfirmResult::No => {
+                return Ok(());
```
adrianbenavides, line 78: What happens if the user doesn't want to upgrade? Will they see the warning every time they run a command?

implementations/rust/ockam/ockam_command/src/upgrade.rs
```diff
@@ -1,51 +1,149 @@
...
+        opts.terminal
+            .write_line(fmt_ok!("Restarted node {}", node_state.name()))?;
+    }
+    Ok(())
 }
 
-fn upgrade_check_is_disabled() -> bool {
-    get_env_with_default("OCKAM_DISABLE_UPGRADE_CHECK", false).unwrap_or(false)
+fn upgrade_ockam(opts: &CommandGlobalOpts) -> miette::Result<()> {
+    let stopped_nodes_names = stop_all_running_nodes(opts)?;
```
adrianbenavides, line 143: Wouldn't it be safer to move this at least after `let installer = get_installer(opts)?;`? Or, even better, right before restarting the nodes. I'm not sure what will happen to running processes after doing the upgrade (are they killed after removing the current binary version?).

implementations/rust/ockam/ockam_command/src/upgrade.rs
BradLewis: This will only run whenever they run `ockam upgrade`. Currently there is a check that is run on every command that looks that will warn you if you're out of date.

adrianbenavides: Ok got it, thanks! 🙏

implementations/rust/ockam/ockam_command/src/upgrade.rs
BradLewis: So, if you run the upgrade without stopping the nodes, everything will still be running, so we could only do the restart cycle after the upgrade has run successfully just to ensure the nodes get upgraded to the latest version.

As an aside to this, with all the changes made with the nodes logic since the `0.90.0` release, the nodes will no longer be restarted correctly on this branch. I'm pretty sure this is just due to the nodes trying to be restarted with the `0.90.0` version after you run the upgrade? I created a new branch https://github.com/BradLewis/ockam/tree/bradlewis/upgrade-test that is just sitting on the latest release just to test this and everything is restarted as expected.

adrianbenavides: That's interesting... can you share more details on how to reproduce the restart problem?

BradLewis: Sure.
1. Create a node using `ockam node create n` using any version of `ockam`.
2. Change the version in the `Cargo.toml` file for `ockam_command` to something below `0.90.0`.
3. Run `cargo run upgrade --yes` to "upgrade" the ockam binary to the latest version. This should just replace the one in `target/debug/ockam` with the latest one from github.
4. Everything appears to run smoothly, however if you run `cargo run node list`, you'll see that the node is stopped.

I'm pretty sure this is because it's trying to restart the nodes using the `0.90.0` release binary, but the "new" nodes are not compatible with that version.

If you run `ockam node list` you'll see that it thinks there are no nodes on the system 
```shell
❯ ockam node list
       ┌───────────────────┐
       │       Nodes       │
       └───────────────────┘

     ! No nodes found on this system.

❯ cargo run node list
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
     Running `/Users/bradlewis/repos/ockam/target/debug/ockam node list`
       ┌───────────────────┐
       │       Nodes       │
       └───────────────────┘

     │ Node n (default) UP
     │ Process id 21821

     │ Node m  UP
     │ Process id 22503

```

You can also easily reproduce this behaviour by first creating a node with release `0.90.0`, then a new one with `cargo run` and then listing the nodes

```shell
❯ ockam node create n
       Creating Node n...

     ✔︎ Node n created successfully

       To see more details on this node, run:
       ockam node show

❯ ockam node list
       ┌───────────────────┐
       │       Nodes       │
       └───────────────────┘

     │ Node n (default) UP
     │ Process id 26852


❯ cargo run node create m
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
     Running `/Users/bradlewis/repos/ockam/target/debug/ockam node create m`
       Creating Node m...

     ✔︎ Node m created successfully

       To see more details on this node, run:
       ockam node show

❯ cargo run node list
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
     Running `/Users/bradlewis/repos/ockam/target/debug/ockam node list`
       ┌───────────────────┐
       │       Nodes       │
       └───────────────────┘

     │ Node n (default) UP
     │ Process id 26852

     │ Node m  UP
     │ Process id 27086

❯ ockam node list
       ┌───────────────────┐
       │       Nodes       │
       └───────────────────┘

     ! No nodes found on this system.

```

Given all this I imagine what's happening is that when the nodes are trying to be restarted, they are trying to be restarted with the `0.90.0` binary, but the `0.90.0` binary doesn't see the nodes so nothing happens. Not sure why there isn't any error though.

If I run this all instead on my other branch I created, everything works as you would expect.

PR #5172, opened 2023-06-26
add ockam command orchestrator ci

metaclips:
This PR allows us run bats test over orchestrator. It run bats test on [merge queue](https://github.blog/changelog/2022-08-18-merge-group-webhook-event-and-github-actions-workflow-trigger/) so that we can workflows with a mutex lock.

- Create a PR and request for review, on approval, PR is added to merge queue
- On merge queue orchestrator bats test is run, if successful PR is merge, if orchestrator bats test fails, PR exits merge queue and PR author must check cause of failure and fix

More
- [x] To run test with a lock, we will need to set maximum number of build in the merge queue to 1
Screenshot 2023-06-29 at 5 09 41 PM

- [x] We need to allow the `actions/upload-artifact` action so that we can upload the logs
- [x] Set `ORCHESTRATOR_DEVELOPMENT_CONTROLLER_ID` and `ORCHESTRATOR_DEVELOPMENT_CONTROLLER_ADDRESS` secrets
- [x] Allow read access to the [artifacts-helper](https://github.com/orgs/build-trust/packages/container/package/artifacts-helper) docker image
- [ ] `Let’s skip this for now` 👉We also need to add the `Rust - test_orchestrator_ockam_command` test to list of list of [status check](https://betterprogramming.pub/github-status-checks-and-branch-protection-made-easy-b70d6d9ffc76)

### DIFF
--- a/.github/workflows/rust-ignored.yml
+++ b/.github/workflows/rust-ignored.yml
@@ -113,3 +113,9 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - run: 'echo "Rust - lint_cargo_lock - Ignored"'
+
+  test_orchestrator_ockam_command:
+    name: Rust - test_orchestrator_ockam_command
+    runs-on: ubuntu-20.04
+    steps:
+      - run: 'echo "Rust - test_orchestrator_ockam_command - Ignored"'
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -392,3 +392,26 @@ jobs:
           toolchain: ${{ matrix.toolchain }}
           target: ${{ matrix.target }}
           platform_operating_system: ${{ matrix.os }}
+
+  test_orchestrator_ockam_command:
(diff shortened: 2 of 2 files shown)


.github/workflows/rust-ignored.yml
```diff
@@ -113,3 +113,10 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - run: 'echo "Rust - lint_cargo_lock - Ignored"'
+
+  test_orchestrator_ockam_command:
+    name: Rust - test_orchestrator_ockam_command
+    runs-on: ubuntu-20.04
+    container: ghcr.io/build-trust/artifacts-helper:latest
```
shanesveller: Can we please omit the `container` if the only steps are a no-op `echo`, so that it doesn't spend time and bandwidth to pull a custom image until one is functionally needed?

metaclips: Done

hairyhum (approved):

PR #5173, opened 2023-06-26
build ockam on cross platforms `on: pull_request`

metaclips:
Build ockam on cross platform `on: pull_request` to ensure we do not break our releases while merging PRs. Also updates cache version

### DIFF
--- a/.github/actions/build_binaries/action.yml
+++ b/.github/actions/build_binaries/action.yml
@@ -1,0 +1,55 @@
+name: Build Ockam Binaries For Different Architechtures
+description: Build Ockam Binaries For Different Architechtures
+
+inputs:
+  use_cross_build:
+    description: Indicate If Cross Should Be Used To Build Binary Instead Of Cargo
+    required: true
+  toolchain:
+    description: Indicate Toolchain To Use, Nightly or Stable
+    default: stable
+  target:
+    description: Indicate Binary Target, e.g. x86_64-apple-darwin
+    required: true
+  platform_operating_system:
+    description: Operating System Which We Are Building Binaries On
+    required: true
+
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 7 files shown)


.github/actions/build_binaries/action.yml
```diff
@@ -0,0 +1,55 @@
+name: Build Ockam Binaries For Different Architechtures
+description: Build Ockam Binaries For Different Architechtures
+
+inputs:
+  use_cross_build:
+    description: Indicate If Cross Should Be Used To Build Binary Instead Of Cargo
```
etorreborre, line 6: I don't know what the accepted convention but Do We Need To Capitalize Every Word :-)?

.github/actions/build_binaries/action.yml
```diff
@@ -0,0 +1,55 @@
...
+      run: |
+        set -x
+        use_cross_build="${{ inputs.use_cross_build }}"
+
+        if [[ "${{ inputs.target }}" =~ .+-musl(.+)?  ]]; then
+          export RUSTFLAGS='-C target-feature=+crt-static'
+        fi
+
+        if [[ $use_cross_build == "true" ]]; then
+          cross build --bin ockam --target ${{ inputs.target }} --verbose --release
```
etorreborre, line 52: I don't see any difference between the `if` and the `else` part

etorreborre (approved):
Except the if/else thing I don't see anything wrong (but I'm not an expert in Github actions :-))

.github/actions/build_binaries/action.yml
metaclips: One uses [cross](https://github.com/cross-rs/cross) to build ockam command and the else part uses conventional cargo command

.github/actions/build_binaries/action.yml
metaclips: I don't think there's a convention, its more of preference of mine.

PR #5185, opened 2023-06-27
fix(rust): forced-killed foreground nodes can be recreated with `ockam node create`

adrianbenavides:
Follow up of https://github.com/build-trust/ockam/pull/5166

These changes fix a problem where a foreground node forcibly killed (e.g. due to an OS restart) was unable to get recreated, as it wasn't removing the previous tcp-listener and we were trying to use it as the "default-tcp-listener", resulting in a "peer not found" error.

To avoid this kind of misuse, the list of transports stored as part of the node config has been replaced by a single transport, which will be overwritten upon restarting a node.

To provide more context:
- We were storing a list of transports so we could recreate them after a node restart, but this feature was never developed
- We were using the first tcp transport of that list to enable communication between the command and the api, which felt unnecessary and can lead to problems if it's not keep properly updated

The following snippet now works as expected:

```sh
ockam node create n -f &
kill -9 $(cat $OCKAM_HOME/nodes/n/pid)
ockam node create n -f &
```

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
@@ -4,13 +4,14 @@ use crate::cli_state::{
     VaultState,
 };
 use crate::config::lookup::ProjectLookup;
-use crate::nodes::models::transport::{CreateTransportJson, TransportMode, TransportType};
+use crate::nodes::models::transport::CreateTransportJson;
+use backwards_compatibility::*;
 use nix::errno::Errno;
+use ockam_core::compat::collections::HashSet;
 use ockam_core::compat::sync::Arc;
 use ockam_identity::{IdentityIdentifier, LmdbStorage};
 use ockam_vault::Vault;
 use serde::{Deserialize, Serialize};
-use std::collections::HashSet;
 use std::fmt::{Display, Formatter};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
@@ -167,8 +168,6 @@ pub struct NodeConfig {
     default_vault: PathBuf,
     #[serde(skip)]
     default_identity: PathBuf,
-    // TODO
-    // authorities: AuthoritiesConfig,
 }
 
 impl NodeConfig {
@@ -303,12 +302,7 @@ pub struct NodeSetupConfig {
     /// The field might be missing in previous configuration files, hence it is an Option
     pub authority_node: Option<bool>,
     pub project: Option<ProjectLookup>,
-    transports: HashSet<CreateTransportJson>,
-    // TODO
-    // secure_channels: ?,
-    // inlets: ?,
-    // outlets: ?,
-    // services: ?,
+    pub api_transport: Option<CreateTransportJson>,
 }
 
 impl NodeSetupConfig {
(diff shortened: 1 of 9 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
```diff
@@ -41,7 +41,7 @@ pub trait StateDirTrait: Sized + Send + Sync {
     }
 
     /// Do not run any migration by default
-    async fn migrate(&self, _item_path: &Path) -> Result<()> {
+    async fn migrate(&self, _path: &Path) -> Result<()> {
         Ok(())
```
davide-baldo, line 45: I was expecting a migrate call here

implementations/rust/ockam/ockam_command/src/node/show.rs
```diff
@@ -275,26 +275,26 @@ pub async fn is_node_up(rpc: &mut Rpc<'_>, wait_until_ready: bool) -> Result<boo
...
 
-    let timeout =
+    let retries =
         FixedInterval::from_millis(IS_NODE_UP_TIME_BETWEEN_CHECKS_MS as u64).take(attempts);
 
     let cli_state = rpc.opts.state.clone();
     let node_name = rpc.node_name().to_owned();
     let now = std::time::Instant::now();
-    for t in timeout {
+    for timeout_duration in retries {
```
davide-baldo, line 284: :pray:

implementations/rust/ockam/ockam_command/tests/bats/nodes_unit.bats
```diff
@@ -31,10 +31,50 @@ teardown() {
...
+  n="$(random_str)"
+  run "$OCKAM" node create "$n"
+  assert_success
+
+  # Fail to create a node with the same name
+  run "$OCKAM" node create "$n"
+  assert_failure
+
+  kill -9 $(cat $OCKAM_HOME/nodes/$n/pid)
+  sleep 0.2
```
davide-baldo: 200ms may be a bit agressive

davide-baldo: you could try something like this:
```while kill -9 $(cat $OCKAM_HOME/nodes/$n/pid) 2>/dev/null; do sleep 0.1; done;```

kill will fail if the process doesn't exist anymore and will quit the loop

File: implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs [hunk 4]
```diff
@@ -393,10 +385,63 @@ impl NodePaths {
...
+        /// This flag is used to determine how the node status should be
+        /// displayed in print_query_status.
+        /// The field might be missing in previous configuration files, hence it is an Option
+        pub authority_node: Option<bool>,
+        pub project: Option<ProjectLookup>,
+        pub transports: HashSet<CreateTransportJson>,
+    }
+
+    #[cfg(test)]
+    impl NodeSetupConfigV1 {
```
davide-baldo, line 432: this method should be unused

davide-baldo (commented):
Looks good, but i don't get why the migration is not automatically called

implementations/rust/ockam/ockam_api/src/cli_state/traits.rs
adrianbenavides: This is the default implementation, which does nothing. You have to add an specific impementation for the config structs you want to migrate.

implementations/rust/ockam/ockam_command/tests/bats/nodes_unit.bats
adrianbenavides: I like that, thanks!

implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
adrianbenavides: It's kept only for testing purposes, so we don't have to remove old tests:
- https://github.com/build-trust/ockam/blob/895630c07c9a24aaa8d43a01e84042a4bf7786a1/implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs#L581
- https://github.com/build-trust/ockam/blob/895630c07c9a24aaa8d43a01e84042a4bf7786a1/implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs#L603

implementations/rust/ockam/ockam_command/tests/bats/nodes_unit.bats
adrianbenavides: done

davide-baldo (dismissed):

adrianbenavides:
⚠️ UPDATE: I've disabled the tests that were using `kill` to stop the node, as it was having an unexpected behavior on CI. I tried several different approaches but none worked. I'll move those tests to rust code as soon as I can.

davide-baldo (approved):

PR #5186, opened 2023-06-27
feat(rust): improve tcp disconnect api

SanjoDeundiak:


davide-baldo (approved):

implementations/rust/ockam/ockam_transport_tcp/src/transport/connection.rs
```diff
@@ -66,8 +66,8 @@ impl TcpTransport {
         ))
     }
 
-    /// Interrupt an active TCP connection given its `Address`
-    pub async fn disconnect(&self, address: &Address) -> Result<()> {
-        self.ctx.stop_worker(address.clone()).await
+    /// Interrupt an active TCP connection given its Sender `Address`
+    pub async fn disconnect(&self, address: impl Into<Address>) -> Result<()> {
```
etorreborre, line 70: @SanjoDeundiak I still think that this is not the right level of abstraction because it is not meaningful to say that `TcpTransport::disconnect` can disconnect anything that has an `Address`. I could actually pass a `SecureChannel` and the compiler would be happy to stop some worker with the secure channel address.

I think we should really pass a `TcpConnection` here so that we don't make this kind of mistake.

SanjoDeundiak: I don't think this function should accept `TcpConnection` as an argument because the calling side is not obliged to store the whole `TcpConnection` object from the time the connection was created. Sender `Address` is enough to both use and stop the connection. If we follow your suggestion, a caller who only has the sender `Address` will need either:
1. Use `ctx.stop_worker, which I think is a worse trade-off, since it will require him to actually investigate the internals to figure out if that's the correct way of disconnecting
2. Run a query to `TcpRegistry` to recreate the `TcpConnection` object, from which only Sender `Address` is going to be used in that call, which is what we had from the very beginning

SanjoDeundiak: I agree that type-safety is not perfect in this case

SanjoDeundiak: The other way would be to agree that stopping a Worker is the correct approach to stop any Worker-owned resource (secure channel, tcp connection, portal, etc.) in our system. Functions like `Tcp::disconnect` and `stop_secure_channel` can be deleted. But we're not there yet to achieve guarantee such behavior

SanjoDeundiak: I will merge this PR since it still makes the code better than it was. But we can continue the discussion and align with a further plan

etorreborre: > a caller who only has the sender Address will need either

That's the situation I am questioning. In which situation would a caller have a bare `Address` and still know that this corresponds to a `TcpConnection` because it is calling `TcpTransport`?

We would be much better-off respecting the levels of abstractions where transports are built on top of workers and workers/addresses stays an implementation detail.

Besides what I'm proposing I see 2 other alternatives:

#### Entity ids

Another way to properly model this, which I still think is not necessary, would be to introduce a notion of `TcpConnectionId` and have an API where:

 - there is a function `TcpConnection::tcp_connection_id(&self) -> TcpConnectionId`
 - and then `TcpTransport::disconnect(&self, tcp_connection_id: TcpConnectionId) -> Result<()>`
 
Then internally we implement `TcpConnectionId` as a wrapper of `Address`.
Maybe this addresses the objection that a caller does not need to have a full `TcpConnection` when all they need is an identifier for it?

 #### An ambient access to the notion of `Worker/Router/Address`
 
The other approach is to expect that many entities we create, `TcpConnection`, `SecureChannel`, `TcpOutlet` etc... are processes that have an `Address`, and that they can be stopped with `ctx.stop(address)`.
This makes the concept of `Worker` and `Address` concepts at the same level as a `SecureChannel`.

The current situation is not satisfactory to me because it is not in 1. (complete abstraction) nor in 2. (complete exposure). 

#### My point of view

My position is that we should stop manipulating `Addresses` all over the place when we can have better typed constructs. I don't see why we wouldn't be able to get there.

PR #5200, opened 2023-06-28
feat(rust): add timer to close secure channel

Alef-gabriel:
## Current behavior

Currently, when to establish a secure channel, two workers are created: decrypt and encrypt. However, when the initiator shuts down its connection without explicitly removing the workers, they remain active on the responder side. This can lead to issues such as performance degradation or memory consumption over time.

## Proposed changes
A timer to shut down the workers when a message is not received

### DIFF
--- a/implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
+++ b/implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
@@ -41,7 +41,9 @@ pub(crate) struct HandshakeWorker {
     role: Role,
     remote_route: Option<Route>,
     decryptor_handler: Option<DecryptorHandler>,
-    idle_timeout: Option<Arc<AtomicBool>>,
+    /// This boolean tracks the activity on the initiator side of the connection
+    /// It is set to false if the connection is inactive after a while
+    is_idle: Option<Arc<AtomicBool>>,
 }
 
 #[ockam_core::worker]
@@ -81,14 +83,16 @@ impl Worker for HandshakeWorker {
         context: &mut Self::Context,
         message: Routed<Self::Message>,
     ) -> Result<()> {
+        // We just received a message. This means that the connection
+        // is still active, so we mark it as such
+        if let Some(is_idle) = self.is_idle.as_ref() {
+            is_idle.store(false, Ordering::Relaxed);
+        };
+
         // Once the decryptor has been initialized, let it handle messages
         // Some messages can come from other systems using the remote address
         // and some messages can come from the current node when the decryptor
         // used to support the decryption of Kafka messages for example
-        if let Some(idle_timeout) = self.idle_timeout.as_ref() {
-            idle_timeout.store(false, Ordering::Relaxed);
-        };
-
         if let Some(decryptor_handler) = self.decryptor_handler.as_mut() {
             let msg_addr = message.msg_addr();
 
@@ -159,7 +163,7 @@ impl HandshakeWorker {
         trust_context: Option<TrustContext>,
         remote_route: Option<Route>,
         timeout: Option<Duration>,
-        idle_timeout: Option<Arc<AtomicBool>>,
+        is_idle: Option<Arc<AtomicBool>>,
         role: Role,
     ) -> Result<()> {
         let vault = to_xx_vault(secure_channels.vault());
@@ -206,7 +210,7 @@ impl HandshakeWorker {
             remote_route: remote_route.clone(),
             addresses: addresses.clone(),
             decryptor_handler: None,
-            idle_timeout,
+            is_idle,
         };
 
         WorkerBuilder::new(worker)
--- a/implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
+++ b/implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
@@ -41,7 +41,7 @@ pub(crate) struct HandshakeWorker {
     role: Role,
     remote_route: Option<Route>,
     decryptor_handler: Option<DecryptorHandler>,
-    flag: Arc<AtomicBool>,
+    idle_timeout: Option<Arc<AtomicBool>>,
 }
 
 #[ockam_core::worker]
@@ -85,6 +85,10 @@ impl Worker for HandshakeWorker {
         // Some messages can come from other systems using the remote address
         // and some messages can come from the current node when the decryptor
         // used to support the decryption of Kafka messages for example
+        if let Some(idle_timeout) = self.idle_timeout.as_ref() {
+            idle_timeout.store(false, Ordering::Relaxed);
+        };
+
         if let Some(decryptor_handler) = self.decryptor_handler.as_mut() {
             let msg_addr = message.msg_addr();
 
@@ -95,7 +99,6 @@ impl Worker for HandshakeWorker {
             } else {
                 Err(IdentityError::UnknownChannelMsgDestination.into())
             };
-            self.flag.store(true, Ordering::Relaxed);
             return result;
         };
 
@@ -156,7 +159,7 @@ impl HandshakeWorker {
         trust_context: Option<TrustContext>,
         remote_route: Option<Route>,
         timeout: Option<Duration>,
-        flag: Arc<AtomicBool>,
+        idle_timeout: Option<Arc<AtomicBool>>,
         role: Role,
     ) -> Result<()> {
         let vault = to_xx_vault(secure_channels.vault());
@@ -203,7 +206,7 @@ impl HandshakeWorker {
             remote_route: remote_route.clone(),
             addresses: addresses.clone(),
             decryptor_handler: None,
-            flag,
+            idle_timeout,
         };
 
         WorkerBuilder::new(worker)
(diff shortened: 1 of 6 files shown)


implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -93,6 +95,7 @@ impl Worker for HandshakeWorker {
             } else {
                 Err(IdentityError::UnknownChannelMsgDestination.into())
             };
+            self.flag.store(true, Ordering::Relaxed);
```
Alef-gabriel: In this case why is this not working? if I put this line outside of this code block It's work normally.

Alef-gabriel:
@davide-baldo

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
davide-baldo: I can speculate that the spawned task is checking the flag **before** the initialization completes, whereas if you move the flag.store in the upper block it would also be called during initilization.

implementations/rust/ockam/ockam_identity/src/secure_channels/secure_channels.rs
```diff
@@ -108,10 +112,27 @@ impl SecureChannels {
...
             Role::Initiator,
         )
         .await?;
 
+        let self_clone = self.clone();
+        let ctx_clone = ctx.async_try_clone().await?;
+        let addr = addresses.encryptor.clone();
+        tokio::spawn(async move {
+            let flag_value = Arc::clone(&flag);
+            while !flag_value.load(Ordering::Relaxed) {
```
davide-baldo: You are already checking below:
```
if !flag_value.load(Ordering::Relaxed) {
    self_clone.stop_secure_channel(&ctx_clone, &addr).await.unwrap();
    break;
}
```
If the flag changes state just after you set it to false (highly unlikely by possible), the loop would exit in the main condition and `self_clone.stop_secure_channel(&ctx_clone, &addr).await.unwrap();` would not be called.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -40,6 +41,7 @@ pub(crate) struct HandshakeWorker {
     role: Role,
     remote_route: Option<Route>,
     decryptor_handler: Option<DecryptorHandler>,
+    flag: Arc<AtomicBool>,
```
davide-baldo: Every variable name is an opportunity you have to give hints to the reader about your intention, try to give a more meaningful name that reflects what you had in mind, even if imperfect.

davide-baldo:
I like it! The direction is excellent!

The only note I have is that using a boolean is somewhat fragile because you don't have any guarantee about when a task is actually run. For example, if some piece of the code misbehaves and blocks an asynchronous thread, every task that is using that thread will be suspended until the piece of code returns. In this case, the secure channel may be idle for the duration of the suspension (which could be minutes) and not detect it.

The most common solution would be to store the current timestamp (less performant) instead of storing the boolean. But being a simple timeout, even if we are not super precise, I think it's ok, the goal is to clean up resources, and it's doing that.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
davide-baldo: probably starting from a `true` value would do the trick

Alef-gabriel:
Hi @davide-baldo !
I'm not sure if it's suitable, but the security channel takes twice as long to close with timeout.

implementations/rust/ockam/ockam_identity/tests/channel.rs
```diff
@@ -1106,7 +1106,7 @@ async fn should_stop_encryptor__and__decryptor__in__secure_channel(
     let local_info = IdentitySecureChannelLocalInfo::find_info(msg.local_message())?;
     assert_eq!(local_info.their_identity_id(), alice.identifier());
     assert_eq!("Hello, Bob!", msg.body());
-
+    dbg!(ctx.list_workers().await?);
```
SanjoDeundiak, line 1109: Do you think it makes sense to remove `dbg!` after the test is passing?

implementations/rust/ockam/ockam_identity/tests/channel.rs
```diff
@@ -1118,9 +1118,161 @@ async fn should_stop_encryptor__and__decryptor__in__secure_channel(
...
+#[allow(non_snake_case)]
+#[ockam_macros::test(timeout = 21000)]
+async fn should_stop_encryptor__and__decryptor__per__timeout(ctx: &mut Context) -> Result<()> {
+    let secure_channels = secure_channels();
+    let identities_creation = secure_channels.identities().identities_creation();
+
+    let bob = identities_creation.create_identity().await?;
+    let alice = identities_creation.create_identity().await?;
+
+    let bob_trust_policy = TrustIdentifierPolicy::new(alice.identifier());
```
SanjoDeundiak, line 1138: This is not a needed detail for that test. Might just use `SecureChannelListenerOptions::new()`. But it's also OK as it is

implementations/rust/ockam/ockam_identity/tests/channel.rs
```diff
@@ -1118,9 +1118,161 @@ async fn should_stop_encryptor__and__decryptor__in__secure_channel(
...
+    child_ctx
+        .send(
+            route![sc.clone(), child_ctx.address()],
+            "Hello, Bob!".to_string(),
+        )
+        .await?;
+
+    let msg = child_ctx.receive::<String>().await?;
+
+    let local_info = IdentitySecureChannelLocalInfo::find_info(msg.local_message())?;
```
SanjoDeundiak, line 1183: I think these 2 lines are not needed here

implementations/rust/ockam/ockam_identity/tests/channel.rs
```diff
@@ -1118,9 +1118,161 @@ async fn should_stop_encryptor__and__decryptor__in__secure_channel(
...
+    assert_eq!(local_info.their_identity_id(), alice.identifier());
+    assert_eq!("Hello, Bob!", msg.body());
+
+    let mut additional_workers = ctx.list_workers().await?;
+    additional_workers.retain(|w| !initial_workers.contains(w));
+    let works_count = additional_workers.len();
```
SanjoDeundiak, line 1189: Could we please, instead of relying on the number of workers, instead check if the specific encryptors and decryptors are on the list of workers? I think there are potentially a lot of workers on the system that can start/stop even though we don't do anything, which can make this test incorrect in the future.

implementations/rust/ockam/ockam_identity/src/secure_channels/secure_channels.rs
```diff
@@ -108,10 +112,32 @@ impl SecureChannels {
...
         .await?;
 
+        let self_clone = self.clone();
+        let ctx_clone = ctx.async_try_clone().await?;
+        let addr = addresses.encryptor.clone();
+
+        // start a background thread to observe the connection activity
+        // If a message is received before maximum_idle_time has passed then the connection is
+        // active. Otherwise, we close the secure channel
+        spawn(async move {
```
SanjoDeundiak, line 127: That will shut down the channel in the range of [maximum_idle_time, 2 * maximum_idle_time) from the last message. Not quite precise, but it should be fine. Could you please mention that detail in the doc-comment?
Also, mentioning that Secure Channels are shut down due to inactivity is worth mentioning somewhere in the doc-comments of the public API. Could you please add that?

implementations/rust/ockam/ockam_identity/tests/channel.rs
SanjoDeundiak: That is also potentially the reason why `should_stop_encryptor__and__decryptor__per__timeout` is failing on the CI

implementations/rust/ockam/ockam_identity/src/secure_channels/secure_channels.rs
```diff
@@ -108,10 +112,32 @@ impl SecureChannels {
...
+            loop {
+                sleep(maximum_idle_time).await;
+                if is_idle.load(Ordering::Relaxed) {
+                    self_clone
+                        .stop_secure_channel(&ctx_clone, &addr)
+                        .await
+                        .unwrap();
```
SanjoDeundiak, line 134: Let's please handle the error gracefully, we generally want to avoid panics

implementations/rust/ockam/ockam_identity/src/secure_channel/listener.rs
```diff
@@ -99,6 +98,7 @@ impl Worker for IdentityChannelListener {
             self.options.trust_context.clone(),
             None,
             None,
+            None,
```
SanjoDeundiak, line 101: What is the reason that we don't use idle flag here?

SanjoDeundiak (changes requested):
Nice approach. Left few comments. Also, could you also please clone the `idle` flag to the Encryptor worker? Because currently using the encryption side of the Secure Channel doesn't reset the flag, therefore if we only use one-side communtication, the Secure Channel will be shut down soon :)

SanjoDeundiak:
Hey @Alef-gabriel , do you plan to get back to this PR?

PR #5201, opened 2023-06-29
feat(rust): add support for json format on ockam_command

adrianbenavides:
Using the `OCKAM_LOG_FORMAT` env variable, we can now override the default logs format with `json` and `pretty` formats.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/logs/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/logs/mod.rs
@@ -1,15 +1,14 @@
 use crate::logs::rolling::{RollingConditionBasic, RollingFileAppender};
 
-use ockam_core::env::{get_env, get_env_with_default};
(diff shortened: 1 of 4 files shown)


implementations/rust/ockam/ockam_command/src/logs/mod.rs
```diff
@@ -24,20 +23,16 @@ fn log_max_files() -> usize {
     get_env_with_default("OCKAM_LOG_MAX_FILES", default).unwrap_or(default) as usize
 }
 
+fn log_format() -> String {
+    let default = "full".to_string();
```
etorreborre: So it can actually be `json`, `pretty` or `full`? Why not use an enum to model those 3 cases exactly?

adrianbenavides: I'm actually going to use "default" instead of "full" for that case. On it!

etorreborre (approved):
🚀

PR #5202, opened 2023-06-29
test(rust): fix force kill node bats tests

adrianbenavides:


implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
```diff
@@ -135,7 +135,11 @@ impl NodeState {
         if let Ok(Some(pid)) = self.pid() {
             let mut sys = System::new();
             sys.refresh_processes();
-            sys.process(Pid::from(pid as usize)).is_some()
+            if let Some(p) = sys.process(Pid::from(pid as usize)) {
+                !matches!(p.status(), ProcessStatus::Dead | ProcessStatus::Zombie)
+            } else {
+                false
+            }
```
adrianbenavides: @davide-baldo all the tests are working now finally! I had to include this change to better detect whether a process is actually running or not. Let me know if this feels off or something.

davide-baldo: :partying_face:
I would add a small comment, someone may wonder why the processes become zombies in the first place!

mrinalwadhwa (approved):

PR #5204, opened 2023-06-29
feat(elixir): simplify semantic of out-of-order  window

polvorin:
Both elixir and rust implementations tie the out-of-order window size to the rekey interval, but the elixir one was less strict allowing a somewhat bigger window in practice, as long as the required decryption key was available.

This PR change it, enforcing a window of fixed REKEY-EACH size around the expected nonce to handle out-of-order messages,  matching the rust behavior.

### DIFF
--- a/implementations/elixir/ockam/ockam/test/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm_test.exs
+++ b/implementations/elixir/ockam/ockam/test/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm_test.exs
@@ -88,4 +88,79 @@ defmodule Ockam.SecureChannel.EncryptedTransportProtocol.AeadAesGcmTests do
     {:ok, ciphertext, _encryptor} = Encryptor.encrypt(<<>>, plain, encryptor)
     {:ok, ^plain, _decryptor} = Decryptor.decrypt(<<>>, ciphertext, decryptor)
   end
+
+  test "out of order, exact sliding window" do
+    # Test values taken from nonce_tracker.rs test case
+    {:ok, encryptor_vault} = SoftwareVault.init()
+    {:ok, decryptor_vault} = SoftwareVault.init()
+    shared_k = :crypto.strong_rand_bytes(32)
+    {:ok, ke} = Vault.secret_import(encryptor_vault, [type: :aes], shared_k)
+    {:ok, kd} = Vault.secret_import(decryptor_vault, [type: :aes], shared_k)
+    key_renewal_interval = 32
+    encryptor = Encryptor.new(encryptor_vault, ke, 0, key_renewal_interval)
+    decryptor = Decryptor.new(decryptor_vault, kd, 0, key_renewal_interval)
+
+    {msgs, _encryptor} =
+      Enum.reduce(0..(key_renewal_interval * 5), {[], encryptor}, fn _i, {acc, encryptor} ->
+        plain = :crypto.strong_rand_bytes(64)
+        {:ok, ciphertext, encryptor} = Encryptor.encrypt(<<>>, plain, encryptor)
+        {[{plain, ciphertext} | acc], encryptor}
+      end)
+
+    msgs = msgs |> Enum.reverse()
+
+    {plaintext, ciphertext} = Enum.at(msgs, 0)
+    {:ok, ^plaintext, decryptor} = Decryptor.decrypt(<<>>, ciphertext, decryptor)
+
+    {plaintext, ciphertext} = Enum.at(msgs, 1)
+    {:ok, ^plaintext, decryptor} = Decryptor.decrypt(<<>>, ciphertext, decryptor)
+
+    {_, ciphertext} = Enum.at(msgs, 0)
+    {:error, _} = Decryptor.decrypt(<<>>, ciphertext, decryptor)
+
+    {plaintext, ciphertext} = Enum.at(msgs, key_renewal_interval + 2)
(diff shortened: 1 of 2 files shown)


implementations/elixir/ockam/ockam/test/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm_test.exs
```diff
@@ -88,4 +88,50 @@ defmodule Ockam.SecureChannel.EncryptedTransportProtocol.AeadAesGcmTests do
     {:ok, ciphertext, _encryptor} = Encryptor.encrypt(<<>>, plain, encryptor)
     {:ok, ^plain, _decryptor} = Decryptor.decrypt(<<>>, ciphertext, decryptor)
   end
+
+  test "out of order, exact sliding window" do
+    # We can't share the _same_ k between encryptor and decryptor on the same vault, as when the encryptor
```
etorreborre: Maybe it would be clearer to describe that for this test:

 - we represent the initiator vault and the responder vault
 - we run the behavior of the `Encryptor` on the initiator node and `Decryptor` on the responder node

implementations/elixir/ockam/ockam/test/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm_test.exs
```diff
@@ -88,4 +88,50 @@ defmodule Ockam.SecureChannel.EncryptedTransportProtocol.AeadAesGcmTests do
...
+        {[{plain, ciphertext} | acc], encryptor}
+      end)
+
+    msgs = msgs |> Enum.reverse()
+
+    # msg is out of the receive window
+    {_, ciphertext} = Enum.at(msgs, 33)
+    {:error, :out_of_window} = Decryptor.decrypt(<<>>, ciphertext, decryptor)
+
+    # msg is on the receive window
```
etorreborre:
```suggestion
    # msg is in the receive window
```

implementations/elixir/ockam/ockam/test/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm_test.exs
```diff
@@ -88,4 +88,50 @@ defmodule Ockam.SecureChannel.EncryptedTransportProtocol.AeadAesGcmTests do
...
+    {plaintext, ciphertext} = Enum.at(msgs, 10)
+    {:ok, ^plaintext, decryptor} = Decryptor.decrypt(<<>>, ciphertext, decryptor)
+
+    # msg is on the receive window
```
etorreborre:
```suggestion
    # msg is in the receive window
```

implementations/elixir/ockam/ockam/test/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm_test.exs
```diff
@@ -88,4 +88,50 @@ defmodule Ockam.SecureChannel.EncryptedTransportProtocol.AeadAesGcmTests do
...
+    {plaintext, ciphertext} = Enum.at(msgs, 2)
+    {:ok, ^plaintext, decryptor} = Decryptor.decrypt(<<>>, ciphertext, decryptor)
+
+    # msg is on the receive window now
```
etorreborre:
```suggestion
    # msg is it the receive window now
```

implementations/elixir/ockam/ockam/test/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm_test.exs
```diff
@@ -88,4 +88,50 @@ defmodule Ockam.SecureChannel.EncryptedTransportProtocol.AeadAesGcmTests do
...
+    {plaintext, ciphertext} = Enum.at(msgs, 33)
```
etorreborre: It would be good to add more boundary tests:

 - 1 before the lower part of the window
 - exactly on the lower part of the window
 - 1 after the lower part of the window
 - 1 before the upper part of the window
 - exactly on the upper part of the window
 - 1 after the upper part of the window

etorreborre (dismissed):
Thanks Pablo. I left some comments and suggestions

implementations/elixir/ockam/ockam/test/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm_test.exs
polvorin: would be misleading,  initiator and responder both have an encryptor and decryptor.       The encryptor on initiator side has the same key than the decryptor on responder side,    and same way around (decryptor on initiator side has the same key than encryptor on responder side).   Each flow direction  rekeys independently.

implementations/elixir/ockam/ockam/test/ockam/secure_channel/encrypted_transport_protocol/aead_aes_gcm_test.exs
polvorin: thanks,  I now adapted  the test in [rust side](https://github.com/build-trust/ockam/blob/344c4736e31c3b104bdc1a31f133603b4377138e/implementations/rust/ockam/ockam_identity/src/secure_channel/nonce_tracker.rs#L61-L83) so we use the same case on both

etorreborre (approved):
🔦

PR #5213, opened 2023-07-03
feat(rust): foreground nodes will always write logs to stdout

adrianbenavides:
- Remove the `--disable-file-logging` flag
- Background nodes (`node create`) will always write the logs to the nodes' configuration log files
- Foreground ndoes (`node create -f`) will always write the logs to stdout, so the user can decide where to redirect the output (if needed)

```sh
# Backgrdound nodes log to the node's log files
$ ockam node create

# Foregrond nodes log to stdout
$ ockam node create -f -vv
```

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/start.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/start.rs
@@ -74,7 +74,7 @@ async fn run_impl(
         None,                                          // Credential
         None,                                          // Trust Context
         None,                                          // Project Name
-        node_setup.disable_file_logging,
+        true,                                          // Restarted nodes will log to files
     )?;
 
     // Print node status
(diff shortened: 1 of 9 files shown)


davide-baldo:
I don't understand why  `--child_process` flag is required, shouldn't `--foreground` do the same?

adrianbenavides:
> I don't understand why `--child_process` flag is required, shouldn't `--foreground` do the same?

The `--child-process` flag helps us figure out wether the node is being created in a subprocess or in the main process, and we use that to make some assumptions (e.g. initialize the node state, disable output colors). It's confusing at first sight, that's for sure.

More details of the two cases:

- When creating a background node (default use-case, `ockam node create`), it runs the same command as a subprocess but adding both `--child-process` and `--foreground` flags.

- When creating a foreground node (`ockam node create -f`), `--foreground` is passed but NOT `--child-process`.

implementations/rust/ockam/ockam_command/src/node/start.rs
```diff
@@ -74,7 +74,7 @@ async fn run_impl(
         None,                                          // Credential
         None,                                          // Trust Context
         None,                                          // Project Name
-        node_setup.disable_file_logging,
+        true,
```
shanesveller: Can we please add a trailing comment like the above lines had, since the bare boolean is a little less clear about intent or content?

adrianbenavides: Done!

etorreborre (approved):

PR #5214, opened 2023-07-03
allow running bats test with a different ockam command cli version

metaclips:
This PR allows us run bats test with a different version of ockam command CLI
Screenshot 2023-07-04 at 1 17 24 PM

### DIFF
--- a/.github/actions/cargo_target_dir_cache/action.yml
+++ b/.github/actions/cargo_target_dir_cache/action.yml
@@ -1,5 +1,8 @@
 name: CARGO_TARGET_DIR Cache
 description: CARGO_TARGET_DIR Cache
+inputs:
+  ockam_dir:
+    description: Ockam Directory
 runs:
   using: composite
   steps:
--- a/.github/actions/cargo_target_dir_cache/action.yml
+++ b/.github/actions/cargo_target_dir_cache/action.yml
@@ -2,7 +2,7 @@ name: CARGO_TARGET_DIR Cache
 description: CARGO_TARGET_DIR Cache
 inputs:
   ockam_dir:
-    description: Ockam Directory
+    description: Ockam root directory to cache (this should be the directory where the target folder is found)
 runs:
   using: composite
   steps:
@@ -10,7 +10,7 @@ runs:
       shell: bash
     - uses: actions/cache@67b839edb68371cc5014f6cea11c9aa77238de78
       with:
-        path: target
+        path: ${{ inputs.ockam_dir != '' && format('{0}/target', inputs.ockam_dir) || 'target' }}
         key: cache-cargo-target-${{ github.workflow }}-${{ github.job }}-ghcr.io/build-trust/ockam-builder@sha256:5ab42598e35509cad3ea9c1e1bd0ed135ed1340c6ae44b762b1c8bbec31d5c68-${{ hashFiles('**/Cargo.lock') }}-${{ hashFiles('./rustc_version.txt') }}
         restore-keys: cache-cargo-target-${{ github.workflow }}-${{ github.job }}-ghcr.io/build-trust/ockam-builder@sha256:5ab42598e35509cad3ea9c1e1bd0ed135ed1340c6ae44b762b1c8bbec31d5c68-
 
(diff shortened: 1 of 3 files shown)


SnowBiz (dismissed):
Looks good!

.github/actions/cargo_target_dir_cache/action.yml
```diff
@@ -1,13 +1,16 @@
 name: CARGO_TARGET_DIR Cache
 description: CARGO_TARGET_DIR Cache
+inputs:
+  ockam_dir:
+    description: Ockam Directory
```
shanesveller: I find this variable name and description to be a little vague. Can you think of any phrasing that would be more explicit about its purpose, the content that we expect to read from there or write to there, etc.? In the typical case, is this the root of the git clone? Is there a sensible default that could be included?

.github/workflows/rust.yml
```diff
@@ -333,34 +335,47 @@ jobs:
...
+          ockam_dir: ockam_cli
 
       - name: Build Binary
         shell: bash
+        working-directory: ockam_cli
         run: |
           rustc --version
           set -x
           cargo build --bin ockam
+          echo "PATH=$(pwd)/target/debug:$PATH" >> $GITHUB_ENV;
```
shanesveller, line 401: ℹ️ I don't necessarily think it's the wrong choice but just want to highlight that we'd be testing with a debug build of the binary which in some cases on the Rust side can have materially different behavior than a `--release` build.

.github/actions/cargo_target_dir_cache/action.yml
metaclips: It's the root directory to be cached, this is where the target directory is.

.github/workflows/rust.yml
metaclips: I'd prefer we run with debug so that we can execute [debug_assert!](https://doc.rust-lang.org/std/macro.debug_assert.html#uses) calls

.github/actions/cargo_target_dir_cache/action.yml
shanesveller: Answering my question in-line here is helpful but to make my feedback more explicit: prior to merging, please find a more descriptive variable name and/or a clearer `description` value that expresses that answer **in the code**, because having to run git-blame and find the PR introducing this code in order to understand why it's written that way is a bad experience.

metaclips: @shanesveller I already pushed a commit updating the description https://github.com/build-trust/ockam/blob/d1034e4bb5c027c7db48fd4c95abe99365d60be5/.github/actions/cargo_target_dir_cache/action.yml#L5. I'll be resolving the conversation, please unresolve if you think we can use a better description.

.github/workflows/rust.yml
```diff
@@ -377,34 +379,47 @@ jobs:
...
 
       - name: Run Script On Ubuntu
-        if: matrix.build == 'linux_86'
+        working-directory: ockam_bats
         shell: bash
         run: |
-          export PATH=$(pwd)/target/debug:$PATH;
+          ockam --version
```
shanesveller, line 407:
```suggestion
          echo $(which ockam)
          ockam --version
```

Optional, but it may be nice to leave ourselves some breadcrumbs to quickly confirm that PATH is resolving the way you've designed it just above.

shanesveller (dismissed):

.github/workflows/rust.yml
metaclips: Done

.github/workflows/rust.yml
```diff
@@ -370,34 +372,48 @@ jobs:
...
+        working-directory: ockam_cli
         run: |
           rustc --version
           set -x
           cargo build --bin ockam
+          echo "PATH=$(pwd)/target/debug:$PATH" >> $GITHUB_ENV;
 
       - name: Run Script On Ubuntu
-        if: matrix.build == 'linux_86'
+        working-directory: ockam_bats
```
shanesveller, line 404: It's deliberate that we don't limit this one on the matrix key anymore, right?

shanesveller (approved):

.github/workflows/rust.yml
metaclips: Yeah, we only run against a single matrix build

PR #4895, opened 2023-05-10
Scaffolding for desktop app

glenngillen:
I started playing with the idea of having a desktop/GUI to interface with Ockam. My initial thinking on what would make a useful set of MVP feature were:

- [ ] System/app tray icon that keeps Ockam mostly hidden from view
- [ ] Click the Ockam icon, menu lists all `relays` I have access to
- [ ] I click a `relay` and it asks if I want to make that service available locally
- [ ] Click "ok" and it binds to a local port (by default the same port that the outlet is routing too, fallback to prompting for a port)

What I've got so far on this branch:

- [x] Scaffolding for a [Tauri](https://tauri.app) app using Rust + Svelte + Typescript
- [x] App displaying Ockam icon in system tray 
- [x] Hooks to make sure I could get click events from the system tray menu (all that is implemented is quit + some debugging statements)
- [x] PoC of a bridge between the Typescript/Svelte code and Rust functions 

I've also completely overlooked all the complexities of ensuring `ockam` command is available, that the local machine is enrolled, what to do if it's not, etc.

PR #5222, opened 2023-07-03
desktop app prototype - 0

mrinalwadhwa:
To try the app:

```
cd implementations/typescript
pnpm install

cd implementations/rust/ockam/ockam_app
cargo tauri dev
```

---

We have to resolve with changes in #4895 
[4c5a232](https://github.com/build-trust/ockam/pull/5222/commits/4c5a232beb38958f5371a270dfdcf4be717be13c) is maybe the only changeset here that is not in #4895


tools/nix/parts/all.nix
```diff
@@ -22,8 +22,11 @@
 
       shellHook = ''
         ${config.devShells.rust.shellHook or ""}
-        export HOME=$PWD/.home
         export DYLD_FALLBACK_LIBRARY_PATH=$(rustc --print sysroot)/lib
+
+        export HOME=$PWD/.home
+        export PATH=$PWD/.home/.cargo/bin:$PATH
+        cargo install tauri-cli --version "^2.0.0-alpha"
```
shanesveller, line 29: Later today, I'll contribute an actual Nix derivation for this CLI, so please omit this line and the one prepending PATH prior to merging. 

Generally `shellHook` should not be used to make any permanent modifications to an end user's system, as it violates the principle of least surprise. In a Nix environment there's typically a more idiomatic way to accomplish what's needed.

tools/nix/parts/all.nix
```diff
@@ -22,8 +22,11 @@
 
       shellHook = ''
         ${config.devShells.rust.shellHook or ""}
-        export HOME=$PWD/.home
         export DYLD_FALLBACK_LIBRARY_PATH=$(rustc --print sysroot)/lib
+
+        export HOME=$PWD/.home
```
shanesveller, line 27: Please drop this line altogether as well - anyone who uses direnv and nix-direnv (as I do) will find that nearly all of the software on their system that uses XDG config dirs is now broken while inside this devShell. Most values of `$EDITOR`, for example.

@mrinalwadhwa do you recall what this was being set for? Typically we can relocate this env var to a wrapper script or other more localized definition to control for whatever it helps with.

tools/nix/parts/all.nix
etorreborre: I'm resolving the conversation. Please open a PR when this one is merged

tools/nix/parts/all.nix
etorreborre: I'm resolving the conversation. Please open a PR when this one is merged

davide-baldo (approved):

PR #5223, opened 2023-07-04
Self-contained confluent demo

glenngillen:
Trying to add a non-trivial demo that combines a number of our existing solutions: Confluent + Telegraf + InfluxDB

The proposed implementation is:

* A node.js kafka producer (`settopbox`) that is approximating what a home IoT/smart device might send. It sends events to an ockam kafka-producer on localhost
* The kafka-producer is configured to use a cluster on confluent cloud
* A telegraf instance that is using kafka as an input source, receiving events via a kafka-consumer on localhost
* Telegraf is also configured to use InfluxDB as an output, sending events via a secure channel


glenngillen:
@mrinalwadhwa here's that demo I was playing with. I _think_ the issue is something to do with the auth or config in `index.js`.

PR #5227, opened 2023-07-04
run ockam test with multiple threads on `cargo nextest`

metaclips:
This PR reduces time it takes to run test with `cargo nextest`, it increase the number of threads taken to run tests that require multithread environments https://nexte.st/book/test-groups.html?highlight=max-thread#comparison-with-threads-required
This PR also disables the `ockam_kafka` test from running so as to reduce kakfa billing


etorreborre (approved):
Thanks @metaclips!

PR #5240, opened 2023-07-05
feat: Improve Nix support for Typescript, Tauri

shanesveller:
Extracted from [my PR comments](https://github.com/build-trust/ockam/pull/5222#pullrequestreview-1512699702) on #5222.

## Current behavior

The current implementation of our `shellHook` performs certain operations imperatively using shell logic, and also (re)installs the Tauri CLI unconditionally when entering our `default` devShell - which will silently pick up new versions any time the Tauri project publishes a new crate version in their 2.0 series.

## Proposed changes

I've moved more of `shellHook` into repeatable Nix content, and packaged the latest 2.0 alpha of `tauri-cli` using the [standard Rust packaging machinery from Nixpkgs](https://nixos.org/manual/nixpkgs/stable/#compiling-rust-applications-with-cargo).

I've also corrected some Typescript-related setup code that was actually inoperable and not accomplishing what it intended to. The reason we can use `someVersion = "1.2.3"` for our Rust/Elixir definitions is because our flake [uses specific community Nix library code](https://github.com/build-trust/ockam/blob/fc6200df0fc517e8de49596f7e33b267694387b7/tools/nix/flake.nix#L5-L10) to make precisely versioned Nix derivations available for consumption. That isn't a property of vanilla nix/nixpkgs, which generally only packages one version or in some cases only distinguishes between major versions. 

We don't have such a Nix library for Typescript/Node/NPM packages, so I used the [versioned Node package trees](https://nixos.org/manual/nixpkgs/stable/#javascript-packages-nixpkgs) in Nixpkgs and a [standard override](https://nixos.org/manual/nixpkgs/stable/#sec-pkg-override). Crucially, overriding only the `version` attribute of a derivation **does not** meaningfully change the content of the result, except to alter its metadata. The existing code on `develop` was still producing whatever version of `pnpm` is in nixpkgs according to our `flake.lock`, and superficially labeling it as an older revision.

Currently unaddressed as of creating this PR is the override of `HOME` inside the `shellHook`. This is fairly problematic for anyone who runs software that looks into the XDG directories for configuration/state, such as anyone who uses `nix-direnv` or who runs their `$EDITOR`, `ssh`, `git`, etc. from inside `nix develop` (as I personally do). I need additional information about the original motive to understand what that line is solving for, so I can replace it with a less disruptive approach. That can be added to this PR or addressed separately.

## Checks


etorreborre (approved):
🚀

etorreborre:
@shanesveller I also don't see why the `HOME` change is necessary. Let's remove it. @mrinalwadhwa does that work for you? Were you relying on this?

mrinalwadhwa:
I was playing around with what using nix in our CI would look like:

In CI, I think:
1. it should completely ignore the host machine
2. we need to identify things that should be cached.

Setting `HOME` and running `nix develop ./tools/nix --ignore-environment` gave me a home directory that seems mostly cacheable.

Screenshot 2023-07-05 at 11 46 54 AM

mrinalwadhwa (approved):

PR #5256, opened 2023-07-06
Refactor the enroll feature + reset

etorreborre:
This PR refactors the enroll feature so that we reuse code from the `ockam_command` crate.
Additionally it introduces a `reset` action to enable faster manual testing when repeatedly trying to enroll:
image

Reset is only available once the enrollment has been done (and then `Enroll...` is disabled):
image

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/util/api.rs
+++ b/implementations/rust/ockam/ockam_command/src/util/api.rs
@@ -238,18 +238,16 @@ fn node_service(service_name: &str) -> String {
 pub mod enroll {
     use ockam_api::cloud::enroll::auth0::{Auth0Token, AuthenticateAuth0Token};
 
-    use crate::enroll::*;
-
     use super::*;
 
-    pub(crate) fn auth0<'a>(
-        cmd: EnrollCommand,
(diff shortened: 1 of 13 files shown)


implementations/rust/ockam/ockam_command/src/util/api.rs
```diff
@@ -238,18 +238,16 @@ fn node_service(service_name: &str) -> String {
...
     use ockam_api::cloud::enroll::auth0::{Auth0Token, AuthenticateAuth0Token};
 
-    use crate::enroll::*;
-
     use super::*;
 
-    pub(crate) fn auth0<'a>(
-        cmd: EnrollCommand,
+    pub fn auth0<'a>(
+        route: &MultiAddr,
```
mrinalwadhwa, line 244: ❤️ we should make this type of change across the board - so UI layer objects `*Command` aren't injected into underlying logic.

adrianbenavides (approved):
Looks great! I've rebased my upcoming changes (tcp outlet creation) on top of this, with some additions regarding error handling and the system tray building process.

LGTM!

PR #5267, opened 2023-07-07
add docs.ockam.io command reference bats test

metaclips:
Add [documentation command reference](https://docs.ockam.io/reference/command) bats test. This test is grouped to a single file `command-reference.bats` for easy code navigation.

### DIFF
--- a/implementations/rust/ockam/ockam_command/tests/bats/command-reference.bats
+++ b/implementations/rust/ockam/ockam_command/tests/bats/command-reference.bats
@@ -1,0 +1,248 @@
+#!/bin/bash
+
+# ===== SETUP
+
+setup() {
+  load load/base.bash
+  load load/orchestrator.bash
+  load_bats_ext
+  setup_home_dir
+  skip_if_orchestrator_tests_not_enabled
+  copy_local_orchestrator_data
+}
+
+teardown() {
+  teardown_home_dir
+}
+
+# ===== TESTS https://docs.ockam.io/reference/command/nodes
+@test "nodes" {
+  run_success "$OCKAM" node create
+
+  run_success "$OCKAM" node create n1
+  run_success "$OCKAM" node create n2 --verbose
+
+  run_success "$OCKAM" node list
+  assert_output --partial "Node n1  UP"
+
+  run_success "$OCKAM" node stop n1
+  assert_output --partial "Stopped node 'n1'"
+
+  run_success "$OCKAM" node start n1
+
+  run_success "$OCKAM" node delete n1 --yes
+  run_success "$OCKAM" node delete --all --yes
+}
+
+@test "workers and services" {
+  run_success "$OCKAM" node create n1
+  run_success "$OCKAM" worker list --at n1
+  run_success "$OCKAM" message send hello --to /node/n1/service/uppercase
+  assert_output "HELLO"
+}
+
+@test "projects - list" {
+  run_success "$OCKAM" project list
+}
+
+@test "space - list" {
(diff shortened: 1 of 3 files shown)


implementations/rust/ockam/ockam_command/tests/bats/command-reference.bats
```diff
@@ -0,0 +1,344 @@
...
+
+  run "$OCKAM" node create n1
+  assert_success
+
+  run "$OCKAM" node create n2 --verbose
+  assert_success
+
+  run "$OCKAM" node list
+  assert_success
+  assert_output --partial "Node n1  UP"
```
etorreborre, line 26: Interesting because we see some small inconsistencies here. It's `n1` in one case and `'n1'` in the other case.

implementations/rust/ockam/ockam_command/tests/bats/command-reference.bats
```diff
@@ -0,0 +1,344 @@
...
+  assert_success
+
+  worker_id=$(ockam tcp-connection create --from a --to 127.0.0.1:7000 | grep -o "[0-9a-f]\{32\}" | head -1)
+
+  output=$(ockam secure-channel create --from a --to /worker/${worker_id}/service/forward_to_b/service/api \
+    | ockam message send hello --from a --to -/service/uppercase)
+  assert [ "$output" == "HELLO" ]
+}
+
+@test "elastic encrypted relays" {
```
etorreborre, line 147: Why "elastic" here? I don't find it in the documentation

implementations/rust/ockam/ockam_command/tests/bats/command-reference.bats
```diff
@@ -0,0 +1,344 @@
...
+  copy_local_orchestrator_data
+}
+
+teardown() {
+  teardown_home_dir
+}
+
+# ===== TESTS https://docs.ockam.io/reference/command/nodes
+@test "nodes" {
+  run "$OCKAM" node create
```
etorreborre: I was wondering if we could write a `run_success` function which does `run xxx; assert_success` in one line. What do you think?

implementations/rust/ockam/ockam_command/tests/bats/command-reference.bats
```diff
@@ -0,0 +1,344 @@
...
+
+  output=$("$OCKAM" secure-channel create \
+    --from n1 --to /node/n2/service/l \
+    --identity i1 --authorized $(cat /tmp/i1.identifier) \
+      | "$OCKAM" message send hello --from n1 --to -/service/uppercase)
+
+  assert [ "$output" == "HELLO" ]
+}
+
+@test "anchoring trust in a credential ussuer" {
```
etorreborre:
```suggestion
@test "anchoring trust in a credential issuer" {
```

etorreborre (dismissed):
Great 🐰

implementations/rust/ockam/ockam_command/tests/bats/command-reference.bats
adrianbenavides: AFAIK, we could replace the `run xxx; assert_success` pattern by just `xxx` because bats scripts are configured to exit on error by default. I read this recently, I'll try to post the link to the source if I remember where.

adrianbenavides: Although, having the `run_success` you mentioned will make things more explicit and easier to understand.

etorreborre (approved):
🐎

PR #5290, opened 2023-07-09
feat(rust): add support to create and list tcp-outlets on the desktop app

adrianbenavides:
- Initializes a default node if there is none in the system
- Adds a new menu item to create tcp-outlet on the default node using a random port
- Adds an event listener to update the tray menu, adding the list of tcp-outlets when starting the application and after creating a new one

image


implementations/rust/ockam/ockam_app/src/app/system_tray.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// Refresh the system tray menu with the latest state, including all list items.
+    pub fn refresh(ctx: &TauriCtx) -> Result<()> {
+        let menu = Self::get_full_menu(ctx).unwrap_or(Self::default());
+        ctx.app_handle().tray_handle().set_menu(menu)?;
+        Ok(())
+    }
+
+    fn get_full_menu(ctx: &TauriCtx) -> Result<SystemTrayMenu> {
+        let enroll = EnrollActions::new();
+        let tcp = TcpOutletActions::full(ctx)?;
```
etorreborre, line 48: One improvement in the future would be to just gray out the items that can't be accessed in case of an error. Can you please create an issue for it?

etorreborre (approved):
Nice 💮

implementations/rust/ockam/ockam_app/src/app/system_tray.rs
adrianbenavides: This call will fail if we fail to retrieve that information from the node. In that case, the TCP outlet list will be empty.

etorreborre: My concern is that if anything changes in the implementation then `?` will short-circuit `get_full_menu` and nothing will be displayed.

adrianbenavides: Note that if `get_full_menu` fails then the default menu (with basic menu items) is displayed:

https://github.com/build-trust/ockam/pull/5290/files#diff-344bd0bf09709196cb556acc10f998293791e3ac3a795b0c1dcbeccb43a58c31R41

```rust
let menu = Self::get_full_menu(ctx).unwrap_or(Self::default());
```

adrianbenavides: @etorreborre let me know if you want me to dig into this any further.

etorreborre: @adrianbenavides that's indeed at least a good backup. I was aiming for a more fine-grained behavior but there's no need to tackle this now.

PR #5336, opened 2023-07-13
Enroll a user with a PKCE flow

### LINKED ISSUE #5265: Use a PKCE flow to enroll users
See [this issue](https://github.com/build-trust/ockam/issues/5228) for more context.

The OAuth 2.0 Authorization Code flow with PKCE is described in this [document](https://datatracker.ietf.org/doc/html/rfc7636). 
This eventually allows users to enroll without having to copy and paste a token in a window.

The purpose of this feature is to implement such a flow for both the desktop application and the command line application.

etorreborre:
This PR does several things:

 1. refactor the Auth0 providers as implementations of a common trait
 2. implement the PKCE flow for retrieving an `Auth0Token` (tested only for the Ockam implementation, not the Okta one)
 3. add logging to the desktop application
 4. retrieve the space + project and mark the user as enrolled after enrollment
 5. clean-up the unused `CloudOpts` options in a few commands

Fixes #5265


implementations/rust/ockam/ockam_api/src/cloud/enroll.rs
```diff
@@ -148,6 +159,18 @@ pub mod auth0 {
         pub access_token: Token,
     }
 
+    #[derive(serde::Deserialize, Debug)]
+    #[cfg_attr(test, derive(PartialEq, Eq, Clone))]
+    pub struct UserInfo {
+        pub sub: String,
+        pub nickname: String,
+        pub name: String,
+        pub picture: String,
```
shanesveller, line 168: Is this one expected to be a fully-qualified URL, relative URL, something else?

implementations/rust/ockam/ockam_command/src/enroll/auth0_service.rs
```diff
@@ -0,0 +1,526 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// This test can only run with an open browser in order to authenticate the user
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    #[ignore]
```
shanesveller: `ignore` can optionally take a string description - we could use that to summarize why a given test is opt-in, like we currently used comments for:
https://doc.rust-lang.org/reference/attributes/testing.html#the-ignore-attribute

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -0,0 +1,371 @@
...
+) -> Result<Space<'a>> {
+    // Get available spaces for node's identity
+    opts.terminal
+        .write_line(&fmt_log!("Getting available spaces in your account..."))?;
+
+    let mut rpc = RpcBuilder::new(ctx, opts, node_name).build();
+    let is_finished = Mutex::new(false);
+    let send_req = async {
+        rpc.request(api::space::list(&CloudOpts::route())).await?;
+        *is_finished.lock().await = true;
```
shanesveller, line 169: Would a `std::sync::atomic::AtomicBool` also work, and would it have any advantages?

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -0,0 +1,371 @@
...
+            "Creating space {}...",
+            name.to_string().color(OckamColor::PrimaryResource.color())
+        )];
+        let progress_output = opts.terminal.progress_output(&message, &is_finished);
+        let (space, _) = try_join!(send_req, progress_output)?;
+        space.to_owned()
+    }
+    // If it has, return the first one on the list
+    else {
+        for space in &available_spaces {
```
shanesveller, line 217: This will be sensitive to whatever order is returned by the backend, correct?

implementations/rust/ockam/ockam_api/src/cloud/enroll.rs
etorreborre: This looks like [a fully qualified URL](https://github.com/build-trust/ockam/issues/5341).

implementations/rust/ockam/ockam_command/src/enroll/auth0_service.rs
etorreborre: Thanks, done.

implementations/rust/ockam/ockam_command/src/enroll/command.rs
etorreborre: Good question. I think this would more to the point, rather than having a `Mutex` on a `bool`. (I just moved that code around in that PR though, I did not touch it)

implementations/rust/ockam/ockam_command/src/enroll/command.rs
etorreborre: Yes, there seems to be an implicit assumption that the order of spaces, hence the default one, will always be the same.

adrianbenavides: Yes I'm afraid. We could derive the default space from the default project. But if don't have either, we decided to just get the first. The user can later set the default with the `space default` command

implementations/rust/ockam/ockam_command/src/enroll/command.rs
adrianbenavides: Definitely! The obvious advantage is the performance, but atomics also has a nicer API in my opinion.

adrianbenavides (approved):
Looks good to me! 🚀

shanesveller (approved):
🎉

PR #5337, opened 2023-07-13
refactor(rust|ts): use compatible types on tauri `commands`  

adrianbenavides:
- make tauri commands compliant (i.e. compatible arguments and return types). I realized this after trying to generate the handlers for those functions with the `generate_handler!` macro

And a couple of useless things for now that could be useful in the next iterations  
- move svelte components to `lib` directory, which is the expected [default for $lib](https://kit.svelte.dev/docs/modules#$lib)
- add the `TcpOutlets` component to serve as an example on how to use async on svelte


implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
```diff
@@ -12,13 +11,14 @@ use ockam_command::{CommandGlobalOpts, GlobalArgs};
...
-pub fn enroll_user() -> Result<()> {
+pub fn enroll_user() {
     let options = CommandGlobalOpts::new(GlobalArgs::default());
     if options.state.identities.default().is_err() {
         create_default_identity(&options);
     }
-    embedded_node(rpc, options)?;
-    Ok(())
+    if let Err(e) = embedded_node(rpc, options) {
+        println!("Error while enrolling user: {e:?}");
```
shanesveller, line 21:
```suggestion
        // use tracing::error
        error!("Error while enrolling user: {e:?}");
```
or
```suggestion
        eprintln!("Error while enrolling user: {e:?}");
```

implementations/rust/ockam/ockam_app/src/tcp/outlet/create.rs
```diff
@@ -10,13 +8,13 @@ use tauri::Manager;
...
-pub fn create(ctx: TauriCtx) -> Result<()> {
+pub fn tcp_outlet_create(app_handle: tauri::AppHandle) {
     let opts = CommandGlobalOpts::new(GlobalArgs::default());
     initialize_node_if_default(&opts, &None);
-    embedded_node(rpc, opts)?;
-    ctx.app_handle()
-        .trigger_global(crate::app::events::SYSTEM_TRAY_ON_UPDATE, None);
-    Ok(())
+    if let Err(e) = embedded_node(rpc, opts) {
+        println!("Error while creating TCP outlet: {e:?}");
```
shanesveller, line 15:
```suggestion
        error!("Error while creating TCP outlet: {e:?}");
```

implementations/rust/ockam/ockam_app/src/tcp/outlet/list.rs
```diff
@@ -1,23 +1,25 @@
...
-pub fn list(_ctx: &TauriCtx) -> Result<OutletList> {
+pub fn tcp_outlet_list() -> TauriCommandResult<Vec<OutletStatus>> {
     let opts = CommandGlobalOpts::new(GlobalArgs::default());
     initialize_node_if_default(&opts, &None);
-    let res = embedded_node(rpc, opts)?;
-    Ok(res)
+    match embedded_node(rpc, opts) {
+        Ok(res) => Ok(res),
+        Err(e) => Err(format!("Error while listing TCP outlets: {e:?}")),
+    }
```
shanesveller, line 18: Is `Result::map_err` helpful here? Are we expecting the body of the success arm to also evolve?

implementations/rust/ockam/ockam_app/src/tcp/outlet/menu_items.rs
```diff
@@ -46,7 +43,7 @@ impl TcpOutletActions {
 }
 
 /// Event listener for the "Create..." menu item
-pub fn on_create(ctx: TauriCtx) -> tauri::Result<()> {
-    let _ = create(ctx);
+pub fn on_create(app_handle: AppHandle) -> tauri::Result<()> {
+    tcp_outlet_create(app_handle);
```
shanesveller, line 47: Is this/should this be a fallible operation?

implementations/typescript/ockam/ockam_app/src/lib/TcpOutlets.svelte
```diff
@@ -0,0 +1,38 @@
+<script>
+    import { invoke } from '@tauri-apps/api/tauri'
+    import { listen, emit } from '@tauri-apps/api/event';
+
+    let tcp_outlets = invoke('tcp_outlet_list');
+
+    async function create() {
+        await invoke('tcp_outlet_create');
+    }
```
shanesveller, line 9: I just started studying Svelte syntax last night, but is this a case where [reactive declaration](https://learn.svelte.dev/tutorial/reactive-declarations) or [reactive statements](https://learn.svelte.dev/tutorial/reactive-statements) might come in handy?

implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
adrianbenavides: At the time I wrote this I wasn't sure about how we were going to handle logs. Now that Eric has enabled the tauri logging plugin, makes total sense to replace all the prints to tracing logs.

Thanks for the heads up!

implementations/rust/ockam/ockam_app/src/tcp/outlet/list.rs
adrianbenavides: > Are we expecting the body of the success arm to also evolve?

I don't think so. I've refactored that to use the `map_err` function.

implementations/rust/ockam/ockam_app/src/tcp/outlet/menu_items.rs
adrianbenavides: Not sure yet if it should. Since we are only using the tray menu, right now we can't do much when something fails, other than logging the error. Perhaps we should wait until we start displaying errors to the user in popup windows.

implementations/typescript/ockam/ockam_app/src/lib/TcpOutlets.svelte
adrianbenavides: How would you do it? To be honest, I struggled a little to make this part work and tried different approaches, so what I wrote might not be the optimal solution.

They way I had this working was: `create` button clicked -> call backend -> backend handles the call and emits an event -> frontend reacts to the event -> updates the `tcp_outlets` variable -> svelte reacts to the updated variable

The `tcp_outlets` variable is a promise that can be handled with an [svelte's await block](https://svelte.dev/examples/await-blocks).

Issue #5323, opened 2023-07-11
Improve Secure Channel shutdown (make Encryptor or Decryptor remove itself from the Registry)


SanjoDeundiak:
### Current behavior

Secure Channel consists of a [Encryptor Worker](https://github.com/build-trust/ockam/blob/2af0cf2ed87d0d1f438fadde8cdc10d619c65efb/implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs) and a [Handshake/Decryptor Worker](https://github.com/build-trust/ockam/blob/2af0cf2ed87d0d1f438fadde8cdc10d619c65efb/implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs). Currently, the proper way to stop a secure channel is implemented [here](https://github.com/build-trust/ockam/blob/2af0cf2ed87d0d1f438fadde8cdc10d619c65efb/implementations/rust/ockam/ockam_identity/src/secure_channels/secure_channels.rs#L123), so both of the Workers are shutdown in that function. Also, the corresponding entry is removed from the Registry. However, that introduces room for inconsistency since a library user may not know about this function or not have access to it (if they don't have access to the `SecureChannels` instance). In this case, it's very tempting to use `ctx.stop_worker()` API and shut down only one of the Workers (probably, Handshake/Decryptor Worker, because we primarily use the Encryptor Address after establishing a Secure Channel) or forget to remove the Registry entry.

### Desired behavior

The possibility for that mistake could be avoided if calling `ctx.stop_worker()` with either Encryptor or Decryptor Address would shut down the Secure Channel properly. To achieve that 2 things are needed:
1. Make Encryptor and Decryptor shut down each other in their shutdown functions [here](https://github.com/build-trust/ockam/blob/2af0cf2ed87d0d1f438fadde8cdc10d619c65efb/implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs#L134)  and [here](https://github.com/build-trust/ockam/blob/2af0cf2ed87d0d1f438fadde8cdc10d619c65efb/implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs#L124). Which is #5322 
2. Make one of these Workers remove its entry from the Registry in its shutdown function. Which is this issue.

---

_We love helping new contributors!_
[...]

PR #5340, opened 2023-07-13
feat(rust): make encryptor and decryptor shut down each other

aleksandarskrbic:
## Proposed changes

Make Encryptor and Decryptor shut down each other in their shutdown functions.
Unregister Encryptor in the `shutdown` function of `HandshakeWorker`.

https://github.com/build-trust/ockam/issues/5322
https://github.com/build-trust/ockam/issues/5323

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
+++ b/implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
@@ -121,7 +121,10 @@ impl Worker for EncryptorWorker {
         Ok(())
     }
 
-    async fn shutdown(&mut self, _context: &mut Self::Context) -> Result<()> {
+    async fn shutdown(&mut self, context: &mut Self::Context) -> Result<()> {
+        let _ = context
+            .stop_worker(self.addresses.decryptor_internal.clone())
+            .await;
         self.encryptor.shutdown().await
     }
 }
--- a/implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
+++ b/implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
@@ -131,10 +131,16 @@ impl Worker for HandshakeWorker {
         Ok(())
     }
 
-    async fn shutdown(&mut self, _context: &mut Self::Context) -> Result<()> {
+    async fn shutdown(&mut self, context: &mut Self::Context) -> Result<()> {
+        let _ = context.stop_worker(self.addresses.encryptor.clone()).await;
+        self.secure_channels
+            .secure_channel_registry
+            .unregister_channel(&self.addresses.encryptor);
+
         if let Some(handler) = &self.decryptor_handler {
             handler.shutdown().await?
         }
+
         Ok(())
     }
 }
--- a/implementations/rust/ockam/ockam_identity/src/secure_channels/secure_channels.rs
+++ b/implementations/rust/ockam/ockam_identity/src/secure_channels/secure_channels.rs
@@ -1,6 +1,6 @@
 use crate::identities::Identities;
 use crate::identities::IdentitiesVault;
-use crate::identity::{IdentityError, IdentityIdentifier};
+use crate::identity::IdentityIdentifier;
 use crate::secure_channel::handshake_worker::HandshakeWorker;
 use crate::secure_channel::{
     Addresses, IdentityChannelListener, Role, SecureChannelListenerOptions, SecureChannelOptions,
@@ -121,26 +121,6 @@ impl SecureChannels {
 
     /// Stop a SecureChannel given an encryptor address
     pub async fn stop_secure_channel(&self, ctx: &Context, channel: &Address) -> Result<()> {
-        if let Some(entry) = self.secure_channel_registry.unregister_channel(channel) {
-            let err1 = ctx
-                .stop_worker(entry.encryptor_messaging_address().clone())
-                .await
-                .err();
-            let err2 = ctx
-                .stop_worker(entry.decryptor_messaging_address().clone())
-                .await
-                .err();
-
-            if let Some(err1) = err1 {
-                return Err(err1);
-            }
-            if let Some(err2) = err2 {
-                return Err(err2);
-            }
-        } else {
-            return Err(IdentityError::SecureChannelNotFound.into());
-        }
-
-        Ok(())
+        ctx.stop_worker(channel.clone()).await
     }
 }


implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -122,6 +122,7 @@ impl Worker for EncryptorWorker {
     }
 
     async fn shutdown(&mut self, _context: &mut Self::Context) -> Result<()> {
+        _context.stop_worker(self.addresses.decryptor_internal.clone()).await?;
```
SanjoDeundiak: Let's rename `_context` to `context`. In Rust `_` prefix indicates that a variable is not used.

SanjoDeundiak: Also, could you please ignore the error from this call here? There is the possibility that this call will fail. For example, if Decryptor is shutting down the Encryptor, then by the time we reach that statement, Decryptor Address no longer exists.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -133,7 +133,18 @@ impl Worker for HandshakeWorker {
 
     async fn shutdown(&mut self, _context: &mut Self::Context) -> Result<()> {
         if let Some(handler) = &self.decryptor_handler {
-            handler.shutdown().await?
+            let encryptor_address = &handler.addresses.encryptor;
+            if let Some(entry) = self.secure_channels.secure_channel_registry.unregister_channel(encryptor_address) {
+                let err = _context.stop_worker(entry.decryptor_messaging_address().clone()).await.err();
```
SanjoDeundiak: It's probably very unclear from the public API, but the messaging address corresponds to the same Decryptor worker, so we don't need to shut it down here. Shutting down any of the Worker's Addresses leads to shutting down the whole Worker.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -133,7 +133,18 @@ impl Worker for HandshakeWorker {
...
+                if let Some(err) = err {
+                    return Err(err);
+                }
+            } else {
+                return Err(IdentityError::SecureChannelNotFound.into());
+            }
+
+            _context.stop_worker(handler.addresses.encryptor.clone()).await?;
```
SanjoDeundiak: Let's just attempt to stop that Address either `self.decryptor_handler` is Some or None. And let's ignore the error.

SanjoDeundiak (changes requested):
Nice work! Left a few suggestions. Could you also please change the commit message start with `feat(rust):`?

aleksandarskrbic:
> Nice work! Left a few suggestions. Could you also please change the commit message start with `feat(rust):`?

@SanjoDeundiak I've implemented the suggested changes(hope that I understood the comments correctly).

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -131,10 +131,13 @@ impl Worker for HandshakeWorker {
         Ok(())
     }
 
-    async fn shutdown(&mut self, _context: &mut Self::Context) -> Result<()> {
+    async fn shutdown(&mut self, context: &mut Self::Context) -> Result<()> {
         if let Some(handler) = &self.decryptor_handler {
-            handler.shutdown().await?
+            handler.shutdown().await?;
+            self.secure_channels.secure_channel_registry.unregister_channel(&handler.addresses.encryptor);
```
SanjoDeundiak: Let's move this and the next call to the beginning of the function. We want to call those functions even if handler.shutdown returned an error

aleksandarskrbic: Cool, just one more question.
If we want to ignore errors what is the preferred way?
1. `context.stop_worker(self.addresses.encryptor.clone()).await.err();`
2. `let _ = context.stop_worker(self.addresses.encryptor.clone()).await;`
@SanjoDeundiak

SanjoDeundiak:
```
let _ = context.stop_worker(self.addresses.encryptor.clone()).await;
```
This one, tnx for asking.
Just to clarify: my comment was not about ignoring the error from `handler.shutdown().await?;`, but to making sure that even if an error occurred in that call - we still run the other 2 actions

aleksandarskrbic: @SanjoDeundiak I understood that I am ignoring the error from `stop_worker`. Thanks!

SanjoDeundiak:
@aleksandarskrbic I just realized one more thing, now the [original function](https://github.com/build-trust/ockam/blob/2af0cf2ed87d0d1f438fadde8cdc10d619c65efb/implementations/rust/ockam/ockam_identity/src/secure_channels/secure_channels.rs#L123) doesn't need to do anything other than stopping the encryptor. Would you like to do that as well? Or we can create an issue for someone else

SanjoDeundiak (dismissed):

SanjoDeundiak:
Could you also please run `cargo fmt`?

implementations/rust/ockam/ockam_identity/src/secure_channels/secure_channels.rs
```diff
@@ -122,21 +122,8 @@ impl SecureChannels {
     /// Stop a SecureChannel given an encryptor address
     pub async fn stop_secure_channel(&self, ctx: &Context, channel: &Address) -> Result<()> {
         if let Some(entry) = self.secure_channel_registry.unregister_channel(channel) {
```
SanjoDeundiak: We now do it inside the Decryptor shutdown, right? So we don't need to do it here. `stop_secure_channel` function should be one line

aleksandarskrbic: So it should look like this ?
```rust
    /// Stop a SecureChannel given an encryptor address
    pub async fn stop_secure_channel(&self, ctx: &Context, channel: &Address) -> Result<()> {
        ctx.stop_worker(channel.clone()).await
    }
```

@SanjoDeundiak

aleksandarskrbic: or it should just return `Ok(()))` ?

SanjoDeundiak: Yes, look correct

SanjoDeundiak (dismissed):
Very nice contribution! Let's confirm CI passes and we should be good to merge

aleksandarskrbic:
> Very nice contribution! Let's confirm CI passes and we should be good to merge

Some checks fails due to unused import, but I fixed it now! @SanjoDeundiak

SanjoDeundiak (approved):

SanjoDeundiak (approved):

PR #5343, opened 2023-07-13
refactor(rust): tray menu handling to add/remove items and keep all the items' state in sync

adrianbenavides:
## System Tray Menu limitations

When initializing the tray menu, you do it through the `SystemTray` struct, which accepts a `SystemTrayMenu` instance. To populate the initial state of the tray system, you can add a list of `CustomMenuItem` instances with their own initial state (title, enabled/disabled, icon). With all those objects in place, a `SystemTray` is initialized.

When you want to modify the `SystemTrayMenu` at runtime, you have a limited access to it through the `SystemTrayMenuHandle`, which only lets you retrieve items by ID (`SystemTrayMenuItemHandle`) and modify their properties (title, icon, enabled/disabled). 

There is no way to add new items to it, the only way I could find is to rebuild the whole menu. This has some challenges:
- When recreating the `CustomMenuItem` items, you have to keep their previous properties (in our current design, we enable/disable the `enroll` and `reset` buttons on some situations).
- There is no direct conversion between `CustomMenuItem` and `SystemTrayMenuItemHandle`.

## Proposed solution

This PR adds an abstraction layer (`TrayMenu`) to easily manage the sections of the tray menu and keep the underlying `CustomMenuItem` in sync with its paired `SystemTrayMenuItemHandle`, so we can rebuild the menu without losing data.


implementations/rust/ockam/ockam_app/src/app/tray_menu.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn add_menu_items(self, items: &[CustomMenuItem]) -> Self;
+}
+
+impl SystemTrayMenuItems for SystemTrayMenu {
+    fn add_menu_items(self, items: &[CustomMenuItem]) -> Self {
+        let mut tm = self;
+        for item in items.iter() {
+            tm = tm.add_item(item.clone());
+        }
+        tm
```
shanesveller: Probably my FP background biasing me, but can this also be written like so?

```suggestion
        items.iter().fold(self, |menu, &item| menu.add_item(item.clone())
```

adrianbenavides: Ohh love it!

adrianbenavides:
Rewritten at https://github.com/build-trust/ockam/pull/5402

Issue #5321, opened 2023-07-11
Add `sqlite` support as an alternative to `lmdb`


SanjoDeundiak:
### Current behavior

Currently, we have an issue with `lmdb`: When an Ockam node lives inside a K8s pod with external storage, the locking mechanism of `lmdb` doesn't work properly because we access the database from two processes with the same pid at the same time. That happens during the pod restart when the new pod is started before the old one is shut down to achieve zero downtime.

### Desired behavior

We would like to try SQLite, as its concurrency capabilities are meant to be much better than in `lmdb`. However, there is no 100% guarantee that it will work (at least, I wasn't able to find evidence of someone succeeding with exactly the same use case). However, using SQLite will make much easier potential further migration to a dedicated pod with a SQL implementation (like PostgreSQL).
For this task, it should be enough to implement a `SQLiteStorage` struct, which should implement traits [Storage](https://github.com/build-trust/ockam/blob/91b4db1c9233f3201b2a7cda27741b8e573a8482/implementations/rust/ockam/ockam_identity/src/identities/storage/storage.rs) and [PolicyStorage](https://github.com/build-trust/ockam/blob/91b4db1c9233f3201b2a7cda27741b8e573a8482/implementations/rust/ockam/ockam_abac/src/traits.rs).
Implementation could go [here](https://github.com/build-trust/ockam/tree/91b4db1c9233f3201b2a7cda27741b8e573a8482/implementations/rust/ockam/ockam_identity/src/identities/storage) in a subfolder

---

_We love helping new contributors!_
_If you have questions or need help as you work on your first Ockam contribution, please leave a comment on [this discussion](https://github.com/ockam-network/ockam/discussions/1081)._
_If you're looking for other issues to contribute to, checkout [this discussion](https://github.com/build-trust/ockam/discussions/3206) and labels -  https://github.com/build-trust/ockam/labels/good%20first%20issue or https://github.com/build-trust/ockam/labels/help%20wanted_

alcolmenar:
hey! first time contributor here and relatively new to Rust. I was hoping to take this issue on if possible. I might need some guidance here so I have a few questions as I'm not super familiar with the code just yet:

1. Is there a specific sqlite crate that you'd like to use? It seems like there are 2 popular ones: sqlite and rusqlite. Looks like there's another crate tokio-rusqlite providing an async connection for rusqlite. I'm thinking rusqlite.
[...]

PR #5356, opened 2023-07-15
feat(rust): enable Sqlite storage as an alternative

alcolmenar:
## Current behavior

Currently there's a bug with the locking mechanism in lmdb when running in k8s during pod restart/redeployment

## Proposed changes

#5321 Proposes to use sqlite as an alternative external storage interface. 

* Created sqlite_storage.rs in ockam_identity and ockam_abac
* Implemented the Storage trait and PolicyStorage trait in the above locations, respectively
* Moved PolicyEntry to storage module in ockam_abac so it can be used by both lmdb_storage and sqlite_storage
* Create `sqlite` feature in both crates
* Test and added `tempfile` as a dev dependency for both crates

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_abac/src/storage/mod.rs
+++ b/implementations/rust/ockam/ockam_abac/src/storage/mod.rs
@@ -3,3 +3,24 @@ pub mod lmdb_storage;
 
 #[cfg(feature = "std")]
 pub use lmdb_storage::*;
+
+#[cfg(feature = "sqlite")]
+pub mod sqlite_storage;
+
+#[cfg(feature = "sqlite")]
+pub use sqlite_storage::*;
+
+use minicbor::{Decode, Encode};
+use ockam_core::compat::borrow::Cow;
+
+use crate::Expr;
+
+/// Policy storage entry.
+///
+/// Used instead of storing plain `Expr` values to allow for additional
+/// metadata, versioning, etc.
+#[derive(Debug, Encode, Decode)]
+#[rustfmt::skip]
+struct PolicyEntry<'a> {
+    #[b(0)] expr: Cow<'a, Expr>,
+}
--- a/implementations/rust/ockam/ockam_identity/src/identities/storage/sqlite_storage.rs
+++ b/implementations/rust/ockam/ockam_identity/src/identities/storage/sqlite_storage.rs
@@ -1,0 +1,191 @@
+use crate::Storage;
+use core::str;
+use ockam_core::async_trait;
+use ockam_core::compat::sync::{Arc, Mutex};
+use ockam_core::compat::vec::Vec;
+use ockam_core::errcode::{Kind, Origin};
+use ockam_core::{Error, Result};
+use ockam_node::tokio::task::{self, JoinError};
+use rusqlite::{params, Connection};
+use std::fmt;
+use std::path::Path;
+use tokio_retry::strategy::{jitter, FixedInterval};
+use tokio_retry::Retry;
+use tracing::debug;
+
+/// Storage using the Sqlite database
+#[derive(Clone)]
+pub struct SqliteStorage {
+    /// Sqlite Connection
+    conn: Arc<Mutex<Connection>>,
+}
+
+impl fmt::Debug for SqliteStorage {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("SqliteStore")
+    }
+}
+
+impl SqliteStorage {
+    const CREATE_IDENTITY_TABLE_SQL: &str = "CREATE TABLE IF NOT EXISTS identity (
+        id INTEGER PRIMARY KEY,
+        identity_id TEXT NOT NULL,
+        key TEXT NOT NULL,
+        value BLOB
+    );";
+    const CREATE_IDENTITY_INDEX_SQL: &str =
+        "CREATE UNIQUE INDEX IF NOT EXISTS idx_identity_id_key ON identity (identity_id, key);";
+
+    const CREATE_POLICY_TABLE_SQL: &str = "CREATE TABLE IF NOT EXISTS policy (
+        id INTEGER PRIMARY KEY,
+        resource TEXT NOT NULL,
+        action TEXT NOT NULL,
+        value BLOB
+    );";
+    const CREATE_POLICY_INDEX_SQL: &str = "CREATE UNIQUE INDEX IF NOT EXISTS idx_policy_resource_action ON policy (resource, action);";
+
+    /// Constructor
+    pub async fn new<P: AsRef<Path>>(p: P) -> Result<Self> {
+        // Not sure we need this
+        // creating a new database might be failing a few times
+        // if the files are currently being held by another pod which is shutting down.
+        // In that case we retry a few times, between 1 and 10 seconds.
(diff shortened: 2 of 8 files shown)


alcolmenar:
@SanjoDeundiak what do you think of the PR?

SanjoDeundiak:
Hey @alcolmenar, reviewing this PR is on our team's TODO list. We'll get back to you shortly!

implementations/rust/ockam/ockam_abac/src/storage/mod.rs
```diff
@@ -3,3 +3,28 @@ pub mod lmdb_storage;
 
 #[cfg(feature = "std")]
 pub use lmdb_storage::*;
+
+#[cfg(feature = "sqlite")]
```
SanjoDeundiak, line 7: You can do it using `cfg_if`, since you added that dependency. It's also fine as it is now though

implementations/rust/ockam/ockam_abac/src/storage/mod.rs
```diff
@@ -3,3 +3,28 @@ pub mod lmdb_storage;
...
+pub mod sqlite_storage;
+
+#[cfg(feature = "sqlite")]
+pub use sqlite_storage::*;
+
+cfg_if::cfg_if! {
+    if #[cfg(feature = "std")] {
```
SanjoDeundiak: Did you add `std` requirement because it was inside the `lmdb_storage.rs`, which was behind the `std` feature? Because now when it's moved, I think all of this can be fine without `std`

implementations/rust/ockam/ockam_identity/src/identities/storage/sqlite_storage.rs
```diff
@@ -0,0 +1,185 @@
...
+impl Storage for SqliteStorage {
+    async fn get(&self, id: &str, key: &str) -> Result<Option<Vec<u8>>> {
+        let db = self.clone();
+        let id = String::from(id);
+        let key = String::from(key);
+
+        let t = move || {
+            let conn = db.conn.lock().unwrap();
+            let result = conn
+                .query_row::<Vec<u8>, _, _>(
```
SanjoDeundiak, line 96: Do you think it makes sense to move all of the queries and make them prepared statements?

implementations/rust/ockam/ockam_identity/src/identities/storage/sqlite_storage.rs
```diff
@@ -0,0 +1,185 @@
...
+        Ok(SqliteStorage {
+            conn: Arc::new(Mutex::new(conn)),
+        })
+    }
+}
+
+#[async_trait]
+impl Storage for SqliteStorage {
+    async fn get(&self, id: &str, key: &str) -> Result<Option<Vec<u8>>> {
+        let db = self.clone();
```
SanjoDeundiak: Does it make sense to clone only the connection object?

implementations/rust/ockam/ockam_identity/src/identities/storage/sqlite_storage.rs
```diff
@@ -0,0 +1,185 @@
...
+use std::path::Path;
+use tokio_retry::strategy::{jitter, FixedInterval};
+use tokio_retry::Retry;
+use tracing::debug;
+
+/// Storage using the Sqlite database
+#[derive(Clone)]
+pub struct SqliteStorage {
+    /// Sqlite Connection
+    pub conn: Arc<Mutex<Connection>>,
```
SanjoDeundiak: Could you please make it private and create a getter that would return a clone of the `Arc`?

SanjoDeundiak (changes requested):
Looks really good!
For the future we may explore possibilties for better separation, e.g., it would make sense to me to have `SqliteStorage` somewhere in a common crate, where everyone could access it. While individual users of that storage could define their own statements on how to create tables for them. Not sure if the data should be stored in one file in this case. But that's definitely out of scope of this PR. Could you please make minor fixes and let's merge it.

SanjoDeundiak:
@alcolmenar  could you please also run `cargo fmt` and squash everything into 1 commit?

implementations/rust/ockam/ockam_abac/src/storage/mod.rs
alcolmenar: Initially, I didn't have the cfg_if but it failed in CI during the no_std testing. I can't seem to find the failing check anymore though but it was most likely due to the `Cow` import and usage. Is there a better way to gate this?

implementations/rust/ockam/ockam_identity/src/identities/storage/sqlite_storage.rs
alcolmenar: I spent some time trying to figure out how to do this actually, because I would've preferred it, but it seems pretty awkward to do. We'd have to create a hash map stored in `SqliteStorage` for the statements that were previously prepared. 

We could try to use the `StatementCache` in the `Connection` struct, but they are hashed by the actual sql string so it doesn't really help with code clarity. It might help with speed though so it doesn't have to re-prepare the statement. Looks like all that needs to be done is replacing `prepare` with `prepare_cached`. I'm not really sure how performant this needs to be though?

implementations/rust/ockam/ockam_identity/src/identities/storage/sqlite_storage.rs
alcolmenar: I'm pretty fresh on Rust still, but I'm not sure what the benefit would be of doing that. From my understanding, the `Clone` derivation will implement a clone method where a new struct will be created with the individual fields would be cloned. Also since the struct has a single field, it wouldn't be subject to alignment padding.

I suppose one benefit would be to be able to use `Arc::clone` for clarity?

implementations/rust/ockam/ockam_identity/src/identities/storage/sqlite_storage.rs
alcolmenar: sure thing

implementations/rust/ockam/ockam_identity/src/identities/storage/sqlite_storage.rs
SanjoDeundiak: Ok, let's not spend time on this now

implementations/rust/ockam/ockam_identity/src/identities/storage/sqlite_storage.rs
SanjoDeundiak: I don't think there will be any difference in how it actually works under the hood. I noticed that we use only the connection struct itself, so I think it makes sense to pass only the required object and nothing more. There is only one field in that struct, though, so there is not much difference.

implementations/rust/ockam/ockam_abac/src/storage/mod.rs
SanjoDeundiak: The only std import I see here is `Cow`, which can be replaced with `use ockam_core::compat::borrow::Cow;`. Would that allow us to remove `std` cfg here?

alcolmenar: ah yes it does, thank you for pointing that out! I'll update the PR with cfg_if removed

SanjoDeundiak (approved):
Very good job! Let's wait for CI success and merge it

PR #5367, opened 2023-07-17
Remove `Backend` and use `tauri` `State`

etorreborre:
`Backend` was introduced in an attempt to be able to test the application without necessarily enrolling or resetting anything.
However, for now, it is a bit premature as we are still trying to find out what is the best application structure.

The other commit is using `tauri`'s managed `State` to instantiate the global options once and for all (as `quiet` so the terminal output is not mangled with regular logs) and pass it around.


shanesveller (approved):
This LGTM, no offense taken if you also wait for additional review from Davide/Adrian.

etorreborre:
Thanks @shanesveller

PR #5385, opened 2023-07-18
feat(rust): added metadata and terminal concepts

davide-baldo:
This is the first PR toward outgoing message validation.

Currently, there is no way to tell apart local addresses from remote ones, and it's impossible to determine the remote identity when going outside the node just by reading the route.

Here I introduce the concept of terminal address, every address that talks to a remote endpoint is deemed terminal, such as TCP transport, secure channel, etc…
Common workers and producers should not be affected in any way.
I also added a more broad concept of address metadata: since now, we can understand which address is local, it's nice to be able to associate arbitrary data to a specific address (for example the remote identity of a terminal address).

see `test/router.rs` to see the builder syntax.

### DIFF
--- a/implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
+++ b/implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
@@ -212,6 +212,7 @@ impl HandshakeWorker {
                 &addresses,
                 decryptor_outgoing_access_control,
             ))
+            .terminal(addresses.decryptor_remote.clone())
             .start(context)
             .await?;
 
@@ -320,6 +321,7 @@ impl HandshakeWorker {
 
             WorkerBuilder::new(encryptor)
                 .with_mailboxes(Mailboxes::new(main_mailbox, vec![api_mailbox]))
+                .terminal(self.addresses.encryptor.clone())
                 .start(context)
                 .await?;
         }
--- a/implementations/rust/ockam/ockam_node/src/context/context.rs
+++ b/implementations/rust/ockam/ockam_node/src/context/context.rs
@@ -6,7 +6,7 @@ use ockam_core::compat::collections::HashMap;
 use ockam_core::compat::sync::{Arc, RwLock};
 use ockam_core::compat::{string::String, vec::Vec};
 use ockam_core::flow_control::FlowControls;
-use ockam_core::{async_trait, Address, Mailboxes, RelayMessage, Result, TransportType};
+use ockam_core::{async_trait, Address, Mailboxes, RelayMessage, Result, Route, TransportType};
 
 #[cfg(feature = "std")]
 use core::fmt::{Debug, Formatter};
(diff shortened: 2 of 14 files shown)


implementations/rust/ockam/ockam_node/src/worker_builder.rs
```diff
@@ -40,6 +41,7 @@ where
             outgoing_ac: Arc::new(AllowAll),
             worker: self.worker,
             address: address.into(),
+            metadata: None,
```
SanjoDeundiak, line 44: Wouldn't everything become much simpler if `metadata` was part of the mailbox?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -320,6 +321,7 @@ impl HandshakeWorker {
 
             WorkerBuilder::new(encryptor)
                 .with_mailboxes(Mailboxes::new(main_mailbox, vec![api_mailbox]))
+                .terminal(self.addresses.encryptor.clone())
```
SanjoDeundiak, line 324: Should we also add `IdentityId` to the attributes?

implementations/rust/ockam/ockam_node/src/context/context.rs
```diff
@@ -161,4 +161,45 @@ impl Context {
             .ok_or_else(|| NodeError::NodeState(NodeReason::Unknown).internal())??;
         Ok(())
     }
+
+    /// Finds the terminal address of a route, if present
+    pub async fn find_terminal_address<A: Into<Route>>(&self, route: A) -> Result<Option<Address>> {
+        let route = route.into();
+        let addresses: Vec<Address> = route.iter().cloned().collect();
```
SanjoDeundiak, line 168: Would be cool to have such function on the `Route` itself

implementations/rust/ockam/ockam_node/src/messages.rs
```diff
@@ -7,6 +7,17 @@ use core::{fmt, sync::atomic::AtomicUsize};
 use ockam_core::compat::{string::String, sync::Arc, vec::Vec};
 use ockam_core::{Address, Error, RelayMessage, Result, TransportType};
 
+/// A set of metadata for a particular address
+#[derive(Debug)]
+pub struct AddressMetadata {
+    /// Address for this metadata
+    pub address: Address,
```
SanjoDeundiak, line 14: Is it required here?

implementations/rust/ockam/ockam_node/src/messages.rs
```diff
@@ -30,7 +43,16 @@ pub enum NodeMessage {
     /// Stop an existing worker
     StopWorker(Address, bool, SmallSender<NodeReplyResult>),
     /// Start a new processor
-    StartProcessor(Address, SenderPair, SmallSender<NodeReplyResult>),
+    StartProcessor {
+        /// The set of addresses in use by this processor
+        addrs: Vec<Address>,
```
SanjoDeundiak, line 48: Let's put metadata along with the Address, so you don't need to find the correspondence later

implementations/rust/ockam/ockam_transport_tcp/src/workers/receiver.rs
```diff
@@ -83,6 +83,7 @@ impl TcpRecvProcessor {
         );
         ProcessorBuilder::new(receiver)
             .with_mailboxes(Mailboxes::new(mailbox, vec![internal]))
+            .terminal(addresses.receiver_address().clone())
```
SanjoDeundiak, line 86: Might also put TCP attribute

SanjoDeundiak (changes requested):
Overall looks good, but I think there is a possibility for simplification. If you agree, I'll do another review pass after that

implementations/rust/ockam/ockam_node/src/worker_builder.rs
davide-baldo: Yes, but it would be a redundant and possibly outdated information since it's stored within a shared structure in the router.
I'm more leaned toward keeping the builder a bit more complex and avoiding duplicates that could stray the casual contributor

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
davide-baldo: It would be nice, my reasoning so far was to keep `ockam_node` and `ockam_identity` decoupled (and in general avoiding identity-specific concepts within node), what's your take?

implementations/rust/ockam/ockam_node/src/messages.rs
davide-baldo: ~~Yes, it's a side effect of avoiding maps in data-passing structures, in the “persistent” data structure maps are used and address is no longer present [1]~~

[1] https://github.com/build-trust/ockam/pull/5385/files#diff-9f8e13d0fa1ee3b7ab2917969b047c923c02dc6f768215d1636fe4de9370d290R267

implementations/rust/ockam/ockam_transport_tcp/src/workers/receiver.rs
davide-baldo: The same reasoning of identity applies here as well, decoupling transport protocol

implementations/rust/ockam/ockam_node/src/messages.rs
davide-baldo: yeah, sounds cleaner

implementations/rust/ockam/ockam_node/src/context/context.rs
davide-baldo: done!

SanjoDeundiak:
Replaced with #7824

PR #5394, opened 2023-07-19
feat(rust): fix output formatting of ockam node list

### LINKED ISSUE #5384: Improve spacing in output when calling `ockam node list`
### Observed behavior

Calling the below
```bash
ockam node create n1
ockam node create n2
```
and then calling `ockam node list` give inconsistent output
```
       ┌───────────────────┐
       │       Nodes       │
       └───────────────────┘
     │ Node n1 (default) UP
     │ Process id 57293

     │ Node n2  UP
     │ Process id 57405
```
`Node n2  UP` adds an extra unneeded space

### Expected behavior

On calling `ockam node list` we should not get irregular spacing in output

### Ockam Version

ockam v0.90.0

qxtaiba:
## Current behavior

In the original implementation of NodeListOutput, when the `self.is_default` field is `false`, an extra whitespace appears in the output when using `format!` macro to construct the final output string.

## Proposed changes

The issue that this PR fixes is described in #5384.

In this fix, the default variable is defined appropriately based on the value of `self.is_default`. In the case that `self.is_default` is `true`, the default variable holds the value " (default)", which includes a space before the "(default)" string. In the case that `self.is_default` is `false`, the default variable holds an empty string, which contains no visible characters.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/list.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/list.rs
@@ -4,6 +4,7 @@ use crate::util::{api, node_rpc, Rpc};
 use crate::{docs, CommandGlobalOpts, Result};
 use clap::Args;
 use colorful::Colorful;
+use indoc::formatdoc;
 use miette::Context as _;
 use ockam::Context;
 use ockam_api::cli_state::StateDirTrait;
@@ -141,18 +142,19 @@ impl Output for NodeListOutput {
             ),
         };
         let default = match self.is_default {
-            true => "(default)".to_string(),
+            true => " (default)".to_string(),
             false => "".to_string(),
         };
 
-        let output = format!(
-            r#"Node {node_name} {default} {status}
-{pid}"#,
-            node_name = self
-                .node_name
-                .to_string()
-                .color(OckamColor::PrimaryResource.color()),
-        );
+        let output = formatdoc! {"
+        Node {node_name}{default} {status}
+        {pid}",
+        node_name = self
+            .node_name
+            .to_string()
+            .color(OckamColor::PrimaryResource.color()),
+        };
+
         Ok(output)
     }
 }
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_command/src/node/list.rs
```diff
@@ -141,18 +142,20 @@ impl Output for NodeListOutput {
...
-        let output = format!(
-            r#"Node {node_name} {default} {status}
-{pid}"#,
-            node_name = self
-                .node_name
-                .to_string()
-                .color(OckamColor::PrimaryResource.color()),
-        );
+        let mut output = String::new();
+        write!(
```
shanesveller: Was there some aspect that did not behave correctly when using `format`? If it works fine, I think we'd probably prefer to continue using that style and syntax here and only adjust the whitespace.

adrianbenavides: In case you did this to avoid the weird new line formatting that we had before, you can use `formatdoc`:

```rust
let output = formatdoc! {"
    Node {node_name}{default} {status}
    {pid}",
    node_name = self
        .node_name
        .to_string()
};
```

qxtaiba: Thanks. My latest commit uses `formatdoc` instead to get around new line formatting. 

@shanesveller If you prefer I leave this implemented with `format` then also happy to do that. Let me know!

shanesveller: No need if `formatdoc` also helped, but thanks for checking!

implementations/rust/ockam/ockam_command/src/node/list.rs
```diff
@@ -141,18 +142,19 @@ impl Output for NodeListOutput {
...
 
-        let output = format!(
-            r#"Node {node_name} {default} {status}
-{pid}"#,
-            node_name = self
-                .node_name
-                .to_string()
-                .color(OckamColor::PrimaryResource.color()),
-        );
+        let output = formatdoc! {"
```
davide-baldo, line 149: nice!

davide-baldo (approved):
LGTM

Issue #5400, opened 2023-07-19
CI fails with `No space left on device` when running cargo test on nightly


metaclips:
CI fails with `No space left on device` when restoring from cache and running `cargo nextest on nightly`, recent failure can be found here https://github.com/build-trust/ockam/actions/runs/5603390367/jobs/10252019705. We should investigate if this issue is due to incorrect caching or on the cargo rust end.

PR #5401, opened 2023-07-19
remove node modules build so that we have enough ci space

metaclips:
This takes a stab at fixing https://github.com/build-trust/ockam/issues/5400 by removing node module builds so as to give more space for cargo build


adrianbenavides (approved):
Great!

PR #5402, opened 2023-07-20
refactor(rust): tray menu handling to add/remove items and keep all the items' state in sync

### LINKED ISSUE #5362: Remove all menu items before enroll, add them after enroll succeeds.
# Current

We're building a Desktop System Tray app for Ockam, see https://github.com/build-trust/ockam/issues/5228

When we first start the app the menu currently looks like this:

Screenshot 2023-07-16 at 7 51 58 AM

# Desired

Before enroll, the menu should look like this:

```
Enroll...
-------------
Please Enroll.
-------------
Quit
```

All other items should appear after enroll.

# Steps to reproduce

You can run the app as follows:

```
cd implementations/typescript
pnpm install
pnpm build

cd ../../implementations/rust/ockam/ockam_app
cargo tauri dev
```

# Source Code

The code for this is here:

- https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_app
- https://github.com/build-trust/ockam/blob/develop/implementations/typescript/ockam/ockam_app

---

_If you have questions or need help as you work on your first Ockam contribution, please leave a comment on [this discussion](https://github.com/ockam-network/ockam/discussions/1081) or join us on [discord](https://discord.gg/bewvnm6zqS). If you're looking for other issues to contribute to, checkout labels -  https://github.com/build-trust/ockam/labels/good%20first%20issue or https://github.com/build-trust/ockam/labels/help%20wanted_

adrianbenavides:
Remake of https://github.com/build-trust/ockam/pull/5343

## System Tray Menu limitations

When initializing the tray menu, you do it through the `SystemTray` struct, which accepts a `SystemTrayMenu` instance. To populate the initial state of the tray system, you can add a list of `CustomMenuItem` instances with their own initial state (title, enabled/disabled, icon). With all those objects in place, a `SystemTray` is initialized.

When you want to modify the `SystemTrayMenu` at runtime, you have a limited access to it through the `SystemTrayMenuHandle`, which only lets you retrieve items by ID (`SystemTrayMenuItemHandle`) and modify their properties (title, icon, enabled/disabled). 

There is no way to add new items to it, the only way I could find is to rebuild the whole menu. This has some challenges:
- When recreating the `CustomMenuItem` items, you have to keep their previous properties (in our current design, we enable/disable the `enroll` and `reset` buttons on some situations).
- There is no direct conversion between `CustomMenuItem` and `SystemTrayMenuItemHandle`.

## Proposed solution

This PR adds an abstraction layer (`TrayMenu`) to easily manage the sections of the tray menu and keep the underlying `CustomMenuItem` in sync with its paired `SystemTrayMenuItemHandle`, so we can rebuild the menu without losing data.

It also keeps the current pre/post enroll menu items.

Pre enroll:

image

Post enroll:

image


implementations/rust/ockam/ockam_app/src/app/tray_menu.rs
```diff
@@ -0,0 +1,118 @@
...
+use tauri::{
+    AppHandle, CustomMenuItem, Manager, SystemTrayEvent, SystemTrayHandle, SystemTrayMenu,
+    SystemTrayMenuItem, Wry,
+};
+use tracing::error;
+
+#[derive(Default)]
+pub struct TrayMenu {
+    pub enroll: EnrollTrayMenuSection,
+    pub tcp: TcpOutletTrayMenuSection,
```
etorreborre, line 18: The naming should probably be something like `SharedServicesTrayMenuSection` here

implementations/rust/ockam/ockam_app/src/app/tray_menu.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub options: OptionsTrayMenuSection,
+}
+
+impl TrayMenu {
+    pub fn init(&self) -> Self {
```
etorreborre: I have the impression that:

 - `init` is not necessary since there is `default`
 - `build` could be private
 - `refresh` would be enough to both create the initial version and any refreshed ones

implementations/rust/ockam/ockam_app/src/app/mod.rs
```diff
@@ -1,13 +1,32 @@
...
 pub use app_state::*;
 pub use logging::*;
 pub use process::*;
-pub use system_tray::*;
+use std::error::Error;
+use tauri::{App, Manager, SystemTray, Wry};
+
+pub use tray_menu::*;
+
+/// Set up the Tauri application so that it listens to tray update events
```
etorreborre: The comment needs to be adjusted now

implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
```diff
@@ -35,7 +34,12 @@ pub async fn enroll_user(app: &AppHandle<Wry>) -> Result<()> {
     enroll_with_token(&context, app_state.options())
         .await
         .unwrap_or_else(|e| error!("{:?}", e));
-    app.trigger_global(app::events::SYSTEM_TRAY_ON_UPDATE, None);
+    // Update tray menu. Use block to release the write lock.
+    {
```
etorreborre: Is it possible to make this block a one-liner since it is repeated for each backend action?

etorreborre: Also I am not sure what is the difference between calling the refresh directly here vs sending a `SYSTEM_TRAY_ON_UPDATE` event and having a listener doing this every time the event is received.

etorreborre (dismissed):
@adrianbenavides I left some comments. I have the feeling that this is a bit too complicated but I can't propose a full alternative yet:

 - having the whole menu in the `AppState` feels wrong to me.
 - and we still don't have all the flexibility we need since we are rebuilding everything on each background action

On the other hand, given the system tray API, we have to pass a brand new menu when we want to change just one menu item. There's a fundamental issue with how we want to display things (as a very dynamic menu) and how the system tray works.

One thing I would prefer possibly would be to have only some "model" state in the `AppState`, like `is_enrolled`, `shared_services` etc... and then have a function rebuilding the menu on top of that. 

This could also be the start of the persistence work where we load the previous `AppState` model from disk and we sync it whenever something changes (when a new service is shared for example).

Let's go with this for now and maybe revisit later.

implementations/rust/ockam/ockam_app/src/app/tray_menu.rs
adrianbenavides: > init is not necessary since there is default

👍 

> build could be private

I need to call that function from the `setup_app` function

> refresh would be enough to both create the initial version and any refreshed ones

I don't get this one. You mean building the menu out of a new menu instance instead of mutating the existing one?

implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
adrianbenavides: Yeah you are right. In the last minute I did some changes to have the menu manipulation code inside the TrayMenu struct, instead of scattered around in the handlers. I didn't realize they all look the same now 😅  I'll reintroduce the event to handle this!

implementations/rust/ockam/ockam_app/src/app/tray_menu.rs
etorreborre: > I need to call that function from the setup_app function

Isn't `lock.init().build(state.is_enrolled())` the same as`lock.refresh(app_handle)` on start-up?

> I don't get this one. You mean building the menu out of a new menu instance instead of mutating the existing one?

Don't sweat it, I'll try to make a pass on this next week to see if I can propose anything different.

adrianbenavides:
> On the other hand, given the system tray API, we have to pass a brand new menu when we want to change just one menu item. There's a fundamental issue with how we want to display things (as a very dynamic menu) and how the system tray works.

I think we could avoid refreshing the whole menu every time we process the `refresh` event. Perhaps if we send some metadata in the event payload we can make smarter decisions on what to update.

> One thing I would prefer possibly would be to have only some "model" state in the AppState, like is_enrolled, shared_services etc... and then have a function rebuilding the menu on top of that.

This would be ideal, right now it feels obvious that we are only relying on the `is_enrolled` status to build the menu. I really hope we can simplify the tray menu handling. I'll keep a close eye on it.

implementations/rust/ockam/ockam_app/src/app/tray_menu.rs
adrianbenavides: > Isn't lock.init().build(state.is_enrolled()) the same aslock.refresh(app_handle) on start-up?

Not exactly. Note that after replacing `init` by default, we no longer need an async runtime to initialize it as it just creates an instance without any list item (it only includes the static items that do not depend on any node state).

The `refresh` function does need the async runtime to query data from the node.

etorreborre (dismissed):

etorreborre (approved):

implementations/rust/ockam/ockam_app/src/app/tray_menu.rs
adrianbenavides: I have to include this "shared service" concept when we create the relay+outlet, so I'll include this rename in my next PR 👍

PR #5411, opened 2023-07-21
Leaner enroll

etorreborre:
This PR refactors the enroll functionality from the desktop app by calling directly the `NodeManager` for:

 - getting the user space and creating one if there is none
 - getting the user project and creating one if there is none

### `NodeManager` API

Some functions of the `NodeManager` are now more directly accessible. For example the `create_space` function can now be called without having to start decoding a `Response`:

```rust
pub async fn create_space(
    &self,
    ctx: &Context,
    req: CreateSpace,
    route: &MultiAddr,
    identity_name: Option<String>,
) -> Result<Space> {
    Response::parse_response(
        self.create_space_response(ctx, CloudRequestWrapper::new(req, route, identity_name)).await?
    )
}

pub(crate) async fn create_space_response(
    &self,
    ctx: &Context,
    req_wrapper: CloudRequestWrapper<CreateSpace>,
) -> Result<Vec<u8>> {
....
```

That function delegates its implementation to `create_space_response` which creates a space by making a request to the controller and returning the response as a list of bytes. 

The `create_space_response` function is convenient for implementing the routing API of a `NodeManager` since we don't need to decode the response, we can just forward it as `Vec<u8>`:
```rust
(Post, ["v0", "spaces"]) => self.create_space_response(ctx, dec.decode()?).await?,
```

#### Generalization

The approach above has been implemented for the "space" and "project" parts of the `NodeManager` API but should be generalized so that:

 - all the routes are implementing this approach: to list workers, create policies etc...
 - we can define a `NodeManagerApi` trait defining all the `NodeManager` functions (and maybe different traits for different domains)
 - we can create a `NodeManagerClient` struct implementing the `NodeManagerApi` trait and implement it by creating requests, then decoding responses. This way we will be able to interact with a remote node using a well-typed and consistent API
 - all functions of the desktop app can access the `NodeManagerApi` directly without having to create any request
 - possibly the use for an "embedded node" in the `ockam_command` crate disappears since the `NodeManager` can be used directly

### Decoding responses

In order to introduce reusable `NodeManager` functions I had to:

 1. move some utility functions to the [`Response` struct](https://github.com/build-trust/ockam/compare/etorreborre/refactoring/leaner-enroll?expand=1#diff-1956a8258c1f4f6fe6f14fe58aace63355c293b7a09192ea689f90fdf888c087R131) (to parse the response status and body)
 
The `Response` API is not really great (for example `is_ok()` does not return a `bool`) but I haven't tried to improve it much for now.

 2. change the constraints on [`parse_response`](https://github.com/build-trust/ockam/compare/etorreborre/refactoring/leaner-enroll?expand=1#diff-1956a8258c1f4f6fe6f14fe58aace63355c293b7a09192ea689f90fdf888c087R133). It now requires a `T: for<'a> Decode<'a, ()>` (instead of just `Decode<'a, ()>`)

 3. fix the derived `Decode` instances for all the data types used in the API for the `NodeManager`. Because of the change in 2. the  lifetime polymorphic (I removed their `<'a>` parameter) structs cannot have a general enough `Decode` instance. 
 
So I simplified those data types to avoid them being lifetime polymorphic. The impact is more allocations but I couldn't find another way to make this work. In the end, I think this is a reasonable trade-off if we get more flexibility and modularity. That being said I'm all ears if someone can find a way to get the best of both worlds

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cloud/project.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/project.rs
@@ -101,11 +101,11 @@ impl Project {
 pub struct OktaConfig {
     #[cfg(feature = "tag")]
     #[serde(skip)]
-    #[cbor(b(0))] pub tag: TypeTag<6434814>,
-    #[cbor(b(1))] pub tenant_base_url: Url,
-    #[cbor(b(2))] pub certificate: String,
-    #[cbor(b(3))] pub client_id: String,
-    #[cbor(b(4))] pub attributes: Vec<String>,
+    #[cbor(n(0))] pub tag: TypeTag<6434814>,
+    #[cbor(n(1))] pub tenant_base_url: Url,
+    #[cbor(n(2))] pub certificate: String,
+    #[cbor(n(3))] pub client_id: String,
+    #[cbor(n(4))] pub attributes: Vec<String>,
 }
 
 impl OktaConfig {
@@ -171,13 +171,13 @@ pub struct InfluxDBTokenLeaseManagerConfig {
     #[cfg(feature = "tag")]
     #[serde(skip)]
     #[cbor(n(0))] pub tag: TypeTag<4166488>,
-    #[cbor(b(1))] pub endpoint: String,
-    #[cbor(b(2))] pub token: String,
-    #[cbor(b(3))] pub org_id: String,
-    #[cbor(b(4))] pub permissions: String,
-    #[cbor(b(5))] pub max_ttl_secs: i32,
-    #[cbor(b(6))] pub user_access_rule: Option<String>,
-    #[cbor(b(7))] pub admin_access_rule: Option<String>,
+    #[cbor(n(1))] pub endpoint: String,
+    #[cbor(n(2))] pub token: String,
+    #[cbor(n(3))] pub org_id: String,
+    #[cbor(n(4))] pub permissions: String,
+    #[cbor(n(5))] pub max_ttl_secs: i32,
+    #[cbor(n(6))] pub user_access_rule: Option<String>,
+    #[cbor(n(7))] pub admin_access_rule: Option<String>,
 }
 
 impl InfluxDBTokenLeaseManagerConfig {
@@ -215,8 +215,8 @@ impl InfluxDBTokenLeaseManagerConfig {
 pub struct CreateProject {
     #[cfg(feature = "tag")]
     #[n(0)] pub tag: TypeTag<8669570>,
-    #[b(1)] pub name: String,
-    #[b(3)] pub users: Vec<String>,
+    #[n(1)] pub name: String,
+    #[n(3)] pub users: Vec<String>,
 }
 
 impl CreateProject {
--- a/implementations/rust/ockam/ockam_api/src/cloud/project.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/project.rs
@@ -231,10 +231,9 @@ impl CreateProject {
 }
 
 mod node {
-    use minicbor::Decoder;
     use tracing::trace;
 
-    use ockam_core::api::Request;
+    use ockam_core::api::{Request, Response};
     use ockam_core::{self, Result};
     use ockam_node::Context;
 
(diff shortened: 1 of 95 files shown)


implementations/rust/ockam/ockam_api/src/cloud/space.rs
```diff
@@ -1,83 +1,80 @@
...
 #[cfg_attr(test, derive(Clone))]
 #[rustfmt::skip]
 #[cbor(map)]
-pub struct CreateSpace<'a> {
+pub struct CreateSpace {
     #[cfg(feature = "tag")]
     #[n(0)] pub tag: TypeTag<2321503>,
-    #[b(1)] pub name: CowStr<'a>,
-    #[b(2)] pub users: Vec<CowStr<'a>>,
+    #[b(1)] pub name: String,
```
adrianbenavides: A heads up about CBOR:
- `#[n{x}]` is used on types that can be cloned
- `#[b{x}]` is used on types that can be borrowed

It probably doesn't have any effect in this case though. I think it's mostly valuable when working with borrowed types.

implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
```diff
@@ -72,3 +90,60 @@ async fn create_default_identity(state: CliState) -> Result<()> {
     info!("created a new identity state");
     Ok(())
 }
+
+async fn retrieve_space(app_state: &AppState) -> Result<Space> {
+    let default_space_name = "default".to_string();
```
adrianbenavides: Should we extract the default names we use in the ockam_app for spaces/projects/nodes?

implementations/rust/ockam/ockam_api/src/cloud/project.rs
```diff
@@ -296,17 +290,25 @@ mod node {
...
+            ctx: &Context,
+            route: &MultiAddr,
+        ) -> Result<Vec<Project>> {
+            let bytes = self
+                .list_projects_response(ctx, CloudRequestWrapper::bare(route))
+                .await?;
+            Response::parse_response_body(bytes.as_slice())
+        }
+
+        pub(crate) async fn list_projects_response(
```
adrianbenavides, line 304: I think this will come in very handy. I'm not sure I like the `_response` suffix though. Perhaps, in another PR, we could introduce a new type that can handle both the typed and encoded response so we can write something like `self.list_projects().encode()` and avoid having to write two methods.

adrianbenavides (dismissed):
Looks great! I left some minor non-blocking comments.

implementations/rust/ockam/ockam_api/src/cloud/space.rs
etorreborre: Thanks, I did not realize this. I made the necessary changes.

implementations/rust/ockam/ockam_api/src/cloud/project.rs
etorreborre: I am also not a big fan of that name but I don't see how we can avoid 2 methods. 

One is public and returns a decoded value and the other one is crate local and keeps the undecoded response.
For the first one it is better to have a direct call: `list_projects()`. For the second one it is better to avoid the unnecessary decode + encode that would happen with `list_projects().encode()`. What kind of data type are you thinking of to fix this?

implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
etorreborre: Yes, I fixed that

implementations/rust/ockam/ockam_api/src/cloud/project.rs
adrianbenavides: I was playing around with this idea and I realized we would have do it the other way around: `list_projects()` returns the encoded body, and then `list_projects().decode()` would return the typed struct. 

As you pointed in the description, we need to return well-typed structs from the API so we can stop moving the responsibility of decoding the body to its expected type up to the API callers (cli, app), which is very uncomfortable.

I was thinking something along these lines:

```rust
struct ResponseBody<T>
where
    T: for<'b> Decode<'b, ()>,
{
    inner: Vec<u8>,
    t: PhantomData<T>,
}

impl<T> ResponseBody<T>
where
    T: for<'b> Decode<'b, ()>,
{
    pub fn new(encoded: Vec<u8>) -> Self {
        Self {
            inner: encoded,
            t: PhantomData::default(),
        }
    }

    pub fn inner(&self) -> &[u8] {
        &self.inner
    }

    pub fn decode(&self) -> Result<T> {
        Response::parse_response_body(&self.inner)
    }
}

#[derive(Debug, Clone, Encode, Decode)]
#[cbor(map)]
struct Space {
    #[n(0)]
    name: String,
}

fn create_space() -> Result<ResponseBody<Space>> {
    let raw_body: Vec<u8> = {
        // handle request
        vec![]
    };
    Ok(ResponseBody::new(raw_body))
}

fn handler() -> Result<()> {
    let body = create_space()?;
    let space = body.decode()?;
    Ok(())
}
```

adrianbenavides (approved):
🚂

PR #5417, opened 2023-07-24
lint rust docs in docs.ockam.io

metaclips:
This PR allows us check Rust code documentation in the docs.ockam.io website ensuring they are same with Rust example in this repository


etorreborre (dismissed):
Excellent!

etorreborre (approved):

PR #5426, opened 2023-07-24
Refactor the system tray building / refresh

etorreborre:
This PR simplifies the creation / refresh of the system tray by:

 - taking all the necessary state (`is_enrolled`, `list_outlets`) from the `NodeManager` stored within the `AppState`
 - making the `build` function `async`
 - removing the additional structs used to keep track of the system tray state

The division of labor between the `NodeManager` and the `NodeManagerWorker` is not yet fully reworked. I have just done the minimum necessary to be able to access command/query functions on the `NodeManager` directly and I will come back at some stage to make everything consistent.


adrianbenavides (approved):
Nice! Love the simplification of the tray menu 🙏

PR #5428, opened 2023-07-25
feat(rust): add popup window to tcp-outlet/service creation

### LINKED ISSUE #5347: Add popup windows to customize the creation of tcp-outlets
### Current behavior

The TCP outlets are created using the default parameters (outlet address, random port, random alias).

### Desired behavior

When clicking the `Create...` button it should open a [temporary window](https://github.com/tauri-apps/tao/blob/dev/examples/window.rs) so the user can introduce the TCP outlet details, pre-populating the inputs with the default values.

adrianbenavides:
When clicking the `Create...` button it will open a popup window to input the tcp outlet details (port and service name). The popup window will be closed after clicking the `Create` button.

### DIFF
--- a/implementations/typescript/ockam/ockam_app/src/routes/service/ServiceCreate.svelte
+++ b/implementations/typescript/ockam/ockam_app/src/routes/service/ServiceCreate.svelte
@@ -1,0 +1,30 @@
+<script>
+  import { invoke } from '@tauri-apps/api/tauri';
+
+  let service = "/service/outlet";
+  let port = "10000";
+
+  async function submit() {
+    await invoke('tcp_outlet_create', {service: service, port: port});
+  }
+</script>
+
+<div>
+  <h1>Service details</h1>
+  <div>
+    <label for="name">Service that you want to share:</label>
+    <div>
+      <input id="name" placeholder="{service}" bind:value="{service}" />
+    </div>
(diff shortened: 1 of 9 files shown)


implementations/typescript/ockam/ockam_app/src/routes/service/ServiceCreate.svelte
```diff
@@ -0,0 +1,30 @@
...
+  <div>
+    <label for="name">Service that you want to share:</label>
+    <div>
+      <input id="name" placeholder="{service}" bind:value="{service}" />
+    </div>
+  </div>
+  <div>
+    <label for="name">Service IP address:</label>
+    <div>
+      <input id="ip_address" placeholder="{port}" bind:value="{port}" />
```
shanesveller: Should the element `id` be renamed to port as well?

shanesveller (commented):
LGTM, just one small nit/question.

implementations/typescript/ockam/ockam_app/src/routes/service/ServiceCreate.svelte
adrianbenavides: yes! I forgot to rename it

shanesveller (approved):

PR #5430, opened 2023-07-25
Display user information

etorreborre:
This PR displays the user information after enrollment as "Full name (github handle)"
image

Unfortunately we are reaching the limits of the system tray in terms of graphical display. For example it would have been nice to be able to show the user avatar picture but this is not possible.

### DIFF
--- a/implementations/rust/ockam/ockam_app/src/app/app_state.rs
+++ b/implementations/rust/ockam/ockam_app/src/app/app_state.rs
@@ -1,10 +1,11 @@
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
 
 use miette::IntoDiagnostic;
 
 use ockam::Context;
 use ockam::{NodeBuilder, TcpListenerOptions, TcpTransport};
 use ockam_api::cli_state::{CliState, StateDirTrait};
+use ockam_api::cloud::enroll::auth0::UserInfo;
 use ockam_api::config::lookup::ProjectLookup;
 use ockam_api::nodes::models::portal::OutletStatus;
 use ockam_api::nodes::service::{
@@ -15,6 +16,8 @@ use ockam_api::nodes::{NodeManager, NodeManagerWorker};
 use ockam_command::node::util::init_node_state;
 use ockam_command::util::api::{TrustContextConfigBuilder, TrustContextOpts};
 use ockam_command::{CommandGlobalOpts, GlobalArgs, Terminal};
+
+use crate::app::model_state::ModelState;
 
 pub const NODE_NAME: &str = "default";
 pub const SPACE_NAME: &str = "default";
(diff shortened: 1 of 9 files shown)


adrianbenavides (approved):

implementations/rust/ockam/ockam_app/src/app/app_state.rs
```diff
@@ -1,10 +1,11 @@
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
```
shanesveller, line 1: We'll need to be mindful about not holding locks across an `await`, but I think using a tokio type instead would include a different set of footguns.

implementations/rust/ockam/ockam_app/src/tcp/outlet/tray_menu.rs
```diff
@@ -9,6 +9,10 @@ pub(crate) async fn build_outlets_section(
     app_state: &AppState,
     tray_menu: SystemTrayMenu,
 ) -> SystemTrayMenu {
+    if !app_state.is_enrolled() {
```
shanesveller, line 12: Very tidy to just short circuit here, thanks.

implementations/rust/ockam/ockam_app/src/enroll/tray_menu.rs
```diff
@@ -1,14 +1,31 @@
...
     if app_state.is_enrolled() {
-        tray_menu
+        match app_state.get_user_info().await {
+            Some(user_info) => {
+                let item = CustomMenuItem::new(
+                    ENROLL_MENU_USER_NAME,
+                    format!("{} ({})", user_info.name, user_info.nickname),
+                );
+                #[cfg(target_os = "macos")]
+                let item = item.native_image(NativeImage::User);
```
shanesveller, line 24: Doesn't this unconditionally replace the formatted string, which hadn't been added to the menu yet? Can you caption a NativeImage? If not, wouldn't we probably want both items included on MacOS, in whichever order is appropriate?

etorreborre: This just adds an image on the left of the menu title, on MacOs, which is what you can see in the PR description. I think this should also do the right thing on Linux but maybe you can confirm?

implementations/rust/ockam/ockam_app/src/app/app_state.rs
etorreborre: Yes the whole topic is a can of worms (I read [this interesting post](https://cliffle.com/blog/lilos-cancel-safety/) about it btw).

implementations/rust/ockam/ockam_app/src/enroll/tray_menu.rs
shanesveller: Gotcha! My primary OS for Ockam work is MacOS, and I can't currently build the Tauri app on my NixOS machine because of the webkit dependencies, which are too new for my configuration. I don't currently have access to another Linux distribution to test with something that is closer to Tauri's documented requirements.

PR #5432, opened 2023-07-25
Persist model data

etorreborre:
This PR introduces a way to persist data in-between application runs and uses this feature to persist the user information.

### DIFF
--- a/implementations/rust/ockam/ockam_app/src/app/app_state.rs
+++ b/implementations/rust/ockam/ockam_app/src/app/app_state.rs
@@ -1,6 +1,7 @@
 use std::sync::{Arc, RwLock};
 
 use miette::IntoDiagnostic;
+use tauri::async_runtime::{block_on, spawn};
 
 use ockam::Context;
 use ockam::{NodeBuilder, TcpListenerOptions, TcpTransport};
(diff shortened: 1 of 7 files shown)


implementations/rust/ockam/ockam_app/src/app/app_state.rs
```diff
@@ -92,20 +99,37 @@ impl AppState {
...
+    pub async fn set_user_info(&self, user_info: UserInfo) -> Result<()> {
+        {
+            let mut model_state = self.model_state.write().unwrap();
+            model_state.set_user_info(user_info);
+        }
+
+        let model_state_clone = {
+            let model_state = self.model_state.read().unwrap();
+            model_state.clone()
+        };
```
adrianbenavides: Couldn't this be rewritten as follows?

```rust
let model_state = {
    let mut model_state = self.model_state.write().unwrap();
    model_state.set_user_info(user_info);
    model_state.clone()
};
```

etorreborre: Indeed. Done

adrianbenavides (approved):
LGTM!

PR #5435, opened 2023-07-26
refactor(rust): create relay from desktop app

adrianbenavides:
## Summary

- extract relay creation as a public function in the `NodeManagerWorker`
- remove the `projects` field from `NodeManager` to load them from the `CliState`: there's no need to set the projects before hand, as we can load when we need them through the CliState

## Work to do

This PR adds the call to create the relay after the user is enrolled. There is a major block that is a little out of scope for this PR: the relay creation fails because the NodeManagerWorker is initialized before the enrollment, hence, it's not initialized with the project/trust-context info. We need to figure out how to update the NodeManagerWorker after the enrollment is done. This problem can be reproduced with:

```rust
ockam reset -y
ockam node create p
ockam enroll
ockam relay create --to p
```

### DIFF
--- a/implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
+++ b/implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
@@ -1,11 +1,12 @@
 use miette::{miette, IntoDiagnostic};
+
 use tauri::{AppHandle, Manager, State, Wry};
 use tracing::log::{error, info};
 
 use ockam::identity::IdentityIdentifier;
 use ockam_api::cli_state;
 use ockam_api::cli_state::traits::StateDirTrait;
-use ockam_api::cli_state::{CliState, SpaceConfig};
+use ockam_api::cli_state::SpaceConfig;
 use ockam_api::cloud::project::Project;
 use ockam_api::cloud::space::{CreateSpace, Space};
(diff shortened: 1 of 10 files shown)


implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
```diff
@@ -21,21 +23,11 @@ use crate::Result;
...
     let app_state: State<AppState> = app.state::<AppState>();
-    if app_state.state().identities.default().is_err() {
-        info!("creating a default identity");
-        create_default_identity(app_state.state())
-            .await
-            .map_err(|e| {
-                error!("{:?}", e);
-                e
-            })?;
-    };
```
adrianbenavides: This part is not needed, as the vault/identity are initialized when creating the AppState instance in the application setup. Let me know if I'm missing something here @etorreborre

etorreborre: I think I was the one missing something :-). Please remove it.

etorreborre (dismissed):
🔢 I really want to revisit the `NodeManagerWorker/NodeManager`. We don't get enough reuse from all that code and I think that this is partially due to the fact that we are locking the whole struct all at once.

etorreborre (approved):

PR #5437, opened 2023-07-26
Fix the re-enrolling after a reset

etorreborre:
This PR allows a user to enroll, then reset and enroll again.

The fix consists in:

 - recreating / initializing a full `CliState` after wiping out the Ockam home directory (and associating that state to the `AppState)
    
 - resetting the `NodeManager` that is associated to the `AppState`
   - stopping the `Medic` instance (I changed a bit the API to do this)
   - recreating a new `NodeManager`

 - resetting the `ModelStateRepository` since the `CliState` has been reinitialized with different paths

There is one remaining issue. When re-enrolling:

 - the browser window stays open with the redirect link
 - the app never receives a token and times-out after 30s.
 - then another enroll eventually works

I will investigate this issue separately.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -334,7 +334,7 @@ impl NodeManager {
         transport_options: NodeManagerTransportOptions,
         trust_options: NodeManagerTrustOptions,
     ) -> Result<Self> {
-        info!("NodeManager::create: create transports");
+        info!("create transports");
         let api_transport_id = random_alias();
         let mut transports = BTreeMap::new();
         transports.insert(
@@ -342,7 +342,7 @@ impl NodeManager {
             transport_options.api_transport_flow_control_id.clone(),
         );
 
-        info!("NodeManager::create: create the identity repository");
+        info!("create the identity repository");
         let cli_state = general_options.cli_state;
         let node_state = cli_state.nodes.get(&general_options.node_name)?;
 
@@ -361,7 +361,7 @@ impl NodeManager {
                 Some(f) => BootstrapedIdentityStore::new(Arc::new(f), repository.clone()),
             });
 
-        info!("NodeManager::create: create the secure channels service");
+        info!("create the secure channels service");
         let secure_channels = SecureChannels::builder()
             .with_identities_vault(vault)
             .with_identities_repository(identities_repository.clone())
@@ -395,14 +395,14 @@ impl NodeManager {
             policies,
         };
 
-        info!("NodeManager::create: {}", s.node_name);
         if !general_options.skip_defaults {
-            info!("NodeManager::create: starting default services");
+            info!("starting default services");
             if let Some(tc) = trust_options.trust_context_config {
-                info!("NodeManager::create: configuring trust context");
+                info!("configuring trust context");
                 s.configure_trust_context(&tc).await?;
             }
         }
+        info!("created a node manager for the node: {}", s.node_name);
 
         Ok(s)
     }
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -334,7 +334,7 @@ impl NodeManager {
         transport_options: NodeManagerTransportOptions,
         trust_options: NodeManagerTrustOptions,
     ) -> Result<Self> {
-        info!("create transports");
+        debug!("create transports");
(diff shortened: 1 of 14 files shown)


implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -331,13 +334,15 @@ impl NodeManager {
         transport_options: NodeManagerTransportOptions,
         trust_options: NodeManagerTrustOptions,
     ) -> Result<Self> {
+        info!("NodeManager::create: create transports");
```
shanesveller: Should this use `target` or another syntax to define "NodeManager::create" segment of the string without manually doing things `tracing` can handle for us more directly?

implementations/rust/ockam/ockam_app/src/app/app_state.rs
```diff
@@ -1,7 +1,8 @@
-use std::sync::{Arc, RwLock};
+use std::sync::Arc;
 
-use miette::IntoDiagnostic;
-use tauri::async_runtime::{block_on, spawn};
+use miette::{miette, IntoDiagnostic};
+use tauri::async_runtime::{block_on, spawn, RwLock};
+use tracing::log::{error, info};
```
shanesveller:
```suggestion
use tracing::{error, info};
```
Rust Analyzer gets me on this one accidentally somewhat often.

implementations/rust/ockam/ockam_app/src/app/app_state.rs
```diff
@@ -155,7 +194,14 @@ async fn make_node_manager(
         .listen(&"127.0.0.1:0", options)
         .await
         .into_diagnostic()?;
-    let projects = ProjectLookup::from_state(opts.state.projects.list()?).await?;
+
+    let projects = ProjectLookup::from_state(
+        opts.state
+            .projects
+            .list()
+            .unwrap_or(std::default::Default::default()),
```
shanesveller: Can we directly use `unwrap_or_default` here? That exists on both Option/Result.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
etorreborre: Absolutely. I went with a similar style of log messages in that method but that's redundant. I removed that now.

implementations/rust/ockam/ockam_app/src/app/app_state.rs
etorreborre: Done (I think I'm actually responsible for a fair amount of those. I'll be careful from now on)

implementations/rust/ockam/ockam_app/src/app/app_state.rs
etorreborre: Much better, thanks

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -356,15 +361,16 @@ impl NodeManager {
                 Some(f) => BootstrapedIdentityStore::new(Arc::new(f), repository.clone()),
             });
 
+        info!("create the secure channels service");
```
adrianbenavides: Regarding the log level, I wrote a small document some months ago while reviewing the ockam_api and ockam_command crates. I also added some general guidelines used in popular repos.

The main problem I found with defaulting to `info` is that logs tend to be very verbose.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -356,15 +361,16 @@ impl NodeManager {
...
         let secure_channels = SecureChannels::builder()
             .with_identities_vault(vault)
             .with_identities_repository(identities_repository.clone())
             .build();
 
         let policies: Arc<dyn PolicyStorage> = Arc::new(node_state.policies_storage().await?);
 
-        let medic = Medic::new();
-        let sessions = medic.sessions();
+        info!("NodeManager::create: start the Medic");
```
adrianbenavides: This one should also have the `NodeManager::create:` removed right?

etorreborre: I missed that one, thanks

implementations/rust/ockam/ockam_api/src/nodes/service.rs
etorreborre: You're right, I changed that to `debug`

adrianbenavides (dismissed):

shanesveller (approved):

PR #5445, opened 2023-07-27
feat(rust): persist tcp-outlets and loads them when initializing the app

adrianbenavides:


implementations/rust/ockam/ockam_app/src/app/app_state.rs
```diff
@@ -227,12 +225,38 @@ fn create_model_state_repository(state: CliState) -> Arc<dyn ModelStateRepositor
...
+fn load_model_state(
+    model_state_repository: Arc<dyn ModelStateRepository>,
+    node_manager: &mut NodeManager,
+    context: Arc<Context>,
+) -> ModelState {
+    block_on(async {
+        match model_state_repository.load().await {
+            Ok(model_state) => {
+                let model_state = model_state.unwrap_or(ModelState::default());
+                for tcp_outlet in model_state.get_tcp_outlets() {
```
etorreborre: I suggest that we start moving this code to a separate function to show that this is a separate concern: updating the `NodeManager` with the model state data. This will make it easier to extend, and also easier to unit test.

implementations/rust/ockam/ockam_app/src/enroll/tray_menu.rs
```diff
@@ -14,7 +14,7 @@ pub(crate) async fn build_enroll_section(
     tray_menu: SystemTrayMenu,
 ) -> SystemTrayMenu {
     if app_state.is_enrolled().await {
-        match app_state.get_user_info().await {
+        match app_state.model(|m| m.get_user_info()).await {
```
etorreborre, line 17: :100:

implementations/rust/ockam/ockam_app/src/tcp/outlet/create.rs
```diff
@@ -11,22 +13,40 @@ pub async fn tcp_outlet_create(
...
+        .into_diagnostic()
+        .wrap_err("Invalid IP address")?;
+    let worker_addr = extract_address_value(&service).wrap_err("Invalid service address")?;
+    let mut node_manager = app_state.node_manager.get().write().await;
+    match node_manager
+        .create_outlet(&app_state.context(), tcp_addr, worker_addr, None, true)
+        .await
+    {
+        Ok(status) => {
+            info!(tcp_addr = status.tcp_addr, "Outlet created");
```
etorreborre: We need to start moving the names from `outlet` to `shared_service`. The whole module should be `shared_service` rather than `tcp/outlet` in order to align with the app conceptual model

implementations/rust/ockam/ockam_app/src/app/app_state.rs
adrianbenavides: 100% agree. What do you think about moving it to a `tcp/outlet/model_state` module to avoid cluttering the main `app/model_state` module?

implementations/rust/ockam/ockam_app/src/tcp/outlet/create.rs
adrianbenavides: I didn't do that on purpose for the following reason: 
- From the user perspective, they only see services
- Internally, a shared service is a combination of a tcp-outlet running on the local node + a relay running on the project node 

If we start calling "shared services" the code that is handling tcp-outlets specifically, we are goigng to blur the separation between those concepts, which could be confusing. Perhaps we should group tcp-outlet and relay code under a "shared_service" module, so we can have that concept represented properly.

What do you think?

etorreborre: > Perhaps we should group tcp-outlet and relay code under a "shared_service" module, so we can have that concept represented properly

Yes, that's what I meant. We should have some top-level modules reflecting the app structure and names

adrianbenavides: Ready!

etorreborre (dismissed):
🦝 Thanks for the notes on the relay creation

etorreborre (approved):

PR #5451, opened 2023-07-27
feat: Create and accept sharing invites from CLI

shanesveller:
## Current behavior

New Orchestrator sharing endpoints are not supported in current CLI subcommands.

## Proposed changes

Add Rust API clients for creating, listing, and redeeming sharing invites for Spaces, Projects, and Services, where the last one is specific in purpose to the Ockam Desktop App.

## Checks


etorreborre:
@shanesveller is it still in draft?

implementations/rust/ockam/ockam_command/src/share/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+mod list;
+mod output;
+mod service;
+
+pub use accept::AcceptCommand;
+pub use create::CreateCommand;
+pub use list::ListCommand;
+pub use service::ServiceCreateCommand;
+
+/// Manage spaces in Ockam Orchestrator
```
shanesveller: Bad copy-paste:
```suggestion
/// Manage sharing invitations in Ockam Orchestrator
```

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
```diff
@@ -0,0 +1,163 @@
...
+#[cbor(map)]
+#[rustfmt::skip]
+pub struct CreateServiceInvite {
+    #[n(1)] pub expires_at: Option<String>,
+    #[n(2)] pub project_id: String,
+    #[n(3)] pub recipient_email: String,
+
+    // TODO: Is there an identity type I should use that may be CBOR-encoded?
+    // TODO: Should route be a MultiAddr?
+    #[n(4)] pub project_identity: String,
```
etorreborre: There is an `IdentityIdentifier` type that can be used for this.

implementations/rust/ockam/ockam_api/src/cloud/share/accept.rs
```diff
@@ -0,0 +1,85 @@
+use minicbor::{Decode, Encode};
+use serde::Serialize;
+
+use super::*;
+
+#[derive(Clone, Debug, Decode, Encode, Serialize)]
+#[cbor(map)]
+#[rustfmt::skip]
+pub struct AcceptInvite {
```
etorreborre: It would be great to have a class diagram + sequence diagram showing how those new data types relate, and how they are created/exchanged

implementations/rust/ockam/ockam_api/src/cloud/share/accept.rs
```diff
@@ -0,0 +1,85 @@
...
+
+        pub(crate) async fn accept_invite_response(
+            &self,
+            ctx: &Context,
+            req_wrapper: CloudRequestWrapper<AcceptInvite>,
+        ) -> Result<Vec<u8>> {
+            let cloud_multiaddr = req_wrapper.multiaddr()?;
+            let req_body = req_wrapper.req;
+            let label = "accept_share";
+            let req_builder = Request::post("/v0/redeem_invite").body(req_body);
```
etorreborre, line 61: Why call it `redeem` on one side and `accept` on the other?

implementations/rust/ockam/ockam_api/src/cloud/share/accept.rs
```diff
@@ -0,0 +1,85 @@
...
+
+            self.request_controller(
+                ctx,
+                label,
+                None,
+                &cloud_multiaddr,
+                API_SERVICE,
+                req_builder,
+                // TODO: Should this include an identifier?
+                None,
```
etorreborre, line 70: If you don't pass an identifier the default identity that is associated to the `NodeManager` will be used to start a secure channel with the `Orchestrator`

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
```diff
@@ -0,0 +1,163 @@
+use minicbor::{Decode, Encode};
+use serde::Serialize;
+
+use super::*;
+
+#[derive(Clone, Debug, Decode, Encode, Serialize)]
+#[cbor(map)]
+#[rustfmt::skip]
+pub struct CreateInvite {
+    #[n(1)] pub expires_at: Option<String>,
```
etorreborre, line 12: Can we use a `Timestamp` datatype instead? There is one used for credential data (see `credential_data.rs`) that we could reuse.

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[n(2)] pub grant_role: RoleInShare,
+    #[n(3)] pub recipient_email: Option<String>,
+    #[n(4)] pub remaining_uses: Option<usize>,
+    #[n(5)] pub scope: ShareScope,
+    #[n(6)] pub target_id: String,
```
etorreborre, line 17: What is this notion of a `target`. The unique identifier for a shared service? Maybe `shared_service_id` would be more explicit then

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    impl NodeManager {
+        // TODO: Should identity_name not be an Option for this use-case?
+        //       This is an authenticated-only request
+        pub async fn create_invite(
+            &self,
+            ctx: &Context,
+            req: CreateInvite,
+            route: &MultiAddr,
+            identity_name: Option<String>,
```
etorreborre, line 57: See my comment above. `identity_name` is a way to select the identity used to create a secure channel. It has no meaning outside of a local installation

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[n(3)] pub recipient_email: Option<String>,
+    #[n(4)] pub remaining_uses: Option<usize>,
+    #[n(5)] pub scope: ShareScope,
+    #[n(6)] pub target_id: String,
+}
+
+#[derive(Clone, Debug, Decode, Encode, Serialize)]
+#[cbor(map)]
+#[rustfmt::skip]
+pub struct CreateServiceInvite {
```
etorreborre: In terms of vocabulary I understand that `invite` is a more informal than `invitation`. Would `invitation` be better?

implementations/rust/ockam/ockam_api/src/cloud/share/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+            other => Err(format!("unknown role: {other}")),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Decode, Deserialize, Encode, Serialize)]
+#[cbor(index_only)]
+#[rustfmt::skip]
+pub enum ShareScope {
+    #[n(0)] Project,
```
etorreborre, line 53: Are we supposed to share a full project / space? What does that mean? Does it mean that any relay created on a project will be accessible then? If you could collect the data model in a small diagram we could discuss that during the next pod meeting (either to get agreement or to present it to others)

etorreborre: Also I don't understand the interactions between `ShareScope` and `target_id`. What is the difference between a `CreateInvite` with a `Service` scope and a `CreateServiceInvite`?

implementations/rust/ockam/ockam_command/src/share/accept.rs
```diff
@@ -0,0 +1,85 @@
...
+use crate::util::{node_rpc, Rpc};
+use crate::{docs, CommandGlobalOpts};
+
+const PREVIEW_TAG: &str = include_str!("../static/preview_tag.txt");
+
+#[derive(Clone, Debug, Args)]
+#[command(
+    before_help = docs::before_help(PREVIEW_TAG),
+)]
+pub struct AcceptCommand {
```
etorreborre, line 21: I am questioning the very title of this PR here :-). Do we need to implement the command right away? Can we just hook the invitation flow to the desktop app for now? I think that we can create separate github issues if people want to extend the ockam command. 

If this is for testing I'd rather write some integration tests. For example, for the enroll functionality I added [those tests](https://github.com/build-trust/ockam/blob/9171d74d4ad4ab66658dce7c9123acfb2e0632ff/implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs#L469). Unfortunately a few of them open the browser so I had to mark them as ignored. But your tests should be able to run without having to run bats tests.

implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -261,6 +265,8 @@ pub enum OckamSubcommand {
     Space(SpaceCommand),
     Project(ProjectCommand),
     Admin(AdminCommand),
+    #[cfg(feature = "orchestrator")]
```
etorreborre, line 268: I am not if it is really correct to add the `Orchestrator` feature. We haven't been using anything like this for the `Space` and `Project` command and yet they also talk to the `Orchestrator`.

etorreborre (commented):
🍏

implementations/rust/ockam/ockam_api/src/cloud/share/accept.rs
shanesveller: No principled reason - I had used `redeem` on all of the backend code, and I think I'd probably favor using that wording everywhere in a vacuum. However, `accept` feels a little more approachable in user-facing copy. WDYT?

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
shanesveller: The Invite mechanism on the backend is lightly polymorphic - you can also invite someone to join your Space or Project, as a feature that's independent of the Desktop app features but with a ton of conceptual overlap to a Service invite. The Service invite needs additional mandatory information to resolve, so it's special cased in a few spots. 

All that to say `target_id` is used here to be a single named argument that on the backend may be a project ID, or a space ID, and it uses `scope` argument to determine which type to attempt to resolve as.

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
shanesveller: That's fine with me - the backend will fairly consistently be using `invite` until/unless I do another refactor pass there, though I don't see it as harmful for that particular disconnect to exist.

implementations/rust/ockam/ockam_api/src/cloud/share/mod.rs
shanesveller: Repeating from another comment since it's contextual here too:
> The Service invite needs additional mandatory information to resolve, so it's special cased in a few spots.

A `CreateInvite` with Service scope doesn't contain all the needed details and would always fail on the request to the backend, so I should probably make a constructor function that provides that feedback more immediately without needing to reach a backend. It's valid to be in any of the three scopes in the read responses, which is why I've currently got it in the enum. I'll ponder whether any other redesigns would be better for safety/correctness or otherwise help prevent misuse.

implementations/rust/ockam/ockam_api/src/cloud/share/accept.rs
etorreborre: My personal inclination would be `accept` everywhere

implementations/rust/ockam/ockam_command/src/lib.rs
shanesveller: From my POV that's fine, as this is a tacit proposal that we should start labeling functionality that are quite specific to using our SAAS backend from things that can work without it - not least of which because that's useful information to convey in our documentation and help output. Even if the default binary includes the whole kitchen sink, this practice would provide us better indications when reading or reviewing code and also force us to provide better seams in the modules and implementation code. At least, it would as long as we periodically use `cargo check --no-default-features`, which currently fails in a few spots even prior to my proposal here.

If you want me to pull that proposal out to a separate conversation that's fine, but it's far easier for me to design net-new code with that practice in mind and strip it out if we reach a team decision to not pursue it, than to retrofit it later.

implementations/rust/ockam/ockam_command/src/share/accept.rs
shanesveller: I don't see harm in doing this work now as it's not a significant amount of added code and little of it is superfluous from the desktop POV. I still need an acceptance flow in both cases and my feedback loop for vetting the end-to-end implementation can be considerably easier with CLI vs Tauri.

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
etorreborre: Ah, great I understand now. I wonder then if being a bit more specific about "ids" would be better in terms of safety:
```rust
enum ShareScope {
  SpaceScope(SpaceId),
  ProjectScope(ProjectId),
  ServiceScope(ServiceId),
}
```

implementations/rust/ockam/ockam_api/src/cloud/share/mod.rs
etorreborre: Yes, if it's not complicated I think it'd be better to rule out cases that would fail.

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
shanesveller: Currently the backend expects and returns string-encoded ISO8601 timestamps, can `Timestamp` de/serialize to and from that? At a glance it looks like its inner representation is Unix epoch time instead.

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[n(1)] pub expires_at: Option<String>,
+    #[n(2)] pub project_id: String,
+    #[n(3)] pub recipient_email: String,
+
+    // TODO: Is there an identity type I should use that may be CBOR-encoded?
+    // TODO: Should route be a MultiAddr?
+    #[n(4)] pub project_identity: IdentityIdentifier,
+    #[n(5)] pub project_route: String,
```
adrianbenavides, line 30: We can. MultiAddr has a `try_from_cbor` method, but I'm sure if we should implement the Decode/Encode traits manually for it to work though. We haven't used this anywhere yet, so we can investigate this later and refactor more requests types.

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
```diff
@@ -0,0 +1,163 @@
...
+                &cloud_multiaddr,
+                API_SERVICE,
+                req_builder,
+                None,
+            )
+            .await
+        }
+
+        // TODO: Should identity_name not be an Option for this use-case?
+        //       This is an authenticated-only request
```
adrianbenavides: This is related to https://github.com/build-trust/ockam/pull/5451/files#r1277582610. I think it still makes sense to have the option, in case we want to use different project/identities pairs from the same machine.

implementations/rust/ockam/ockam_api/src/cloud/share/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+impl FromStr for RoleInShare {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "admin" => Ok(Self::Admin),
+            "guest" => Ok(Self::Guest),
+            "service" => Ok(Self::Service),
```
adrianbenavides: The from/to conversions of the `Service` differ, should it be `service` in both cases?

shanesveller: Thank you for catching this! I wrote those across two separate working sessions, but they should be in agreement, so I'll switch the From to also be `service_user`.

implementations/rust/ockam/ockam_api/src/cloud/share/accept.rs
shanesveller: This has been changed everywhere in this repo, but not on the backend.

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
shanesveller: I can circle back to make a [time::OffsetDateTime](https://docs.rs/time/0.3.23/time/struct.OffsetDateTime.html#method.format) directly CBOR-representable in a future PR. For now I'd like to defer that work and keep the String representation here rather than switch both here and the backend to Unix-epoch numeric timestamps. 

The latter are fine to work with programmatically but IMO are user-unfriendly when they need to be supplied as form or CLI input.

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
shanesveller: I've attempted to moved all wording, module names, commit messages, etc. to use invitation rather than invite. I'll only continue to use the latter when needed as a verb.

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
shanesveller: I would like to pursue that in a future PR for correctness' sake, but I don't think it's mechanically needed for this behavior for now. The backend is currently going to store and retrieve the route fields as an opaque string value and pass it back to clients unaltered, without validating its syntax or semantics.

implementations/rust/ockam/ockam_api/src/cloud/share/mod.rs
shanesveller: This should be corrected in the last push.

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
adrianbenavides: What's the status of this? Should be create an issue to keep track of it?

adrianbenavides (approved):
Looks good to me! I propose creating issues for the open comments and resolve them so we can merge it as soon as possible.

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
shanesveller: I think Eric's suggestion has good merit, so I'd like to consider it a nice-to-have and moved it to a tracking issue https://github.com/build-trust/ockam/issues/5517 to pursue later on. Most versions of this idea further decouples the client's internal data model from the backend's representation of the same ideas, so I'd like to make sure whomever tackles it does approach doing so with appropriate care.

implementations/rust/ockam/ockam_api/src/cloud/share/mod.rs
shanesveller: I believe this is also captured in spirit as part of #5517, please let me know if that's not the case.

implementations/rust/ockam/ockam_api/src/cloud/share/create.rs
shanesveller: I captured this desire as well as part of #5517.

PR #5458, opened 2023-07-29
feat(rust): create a relay at the default project after enrolling and when starting the app

adrianbenavides:
Creates a relay at the default project:
1. After the user is enrolled successfully
2. When the application starts, if the user is already enrolled

The most important change is how we store and handle the AppState's NodeManagerWorker, which is now fully replaced, instead of only replacing the inner NodeManager. This is needed because, when we want to reset the node, we can't just recreate the inner's NodeManage, as we need to trigger the [Worker' initialization method](https://github.com/build-trust/ockam/blob/0f1ffe70de1851a4b241282e2932ed269beb0efa/implementations/rust/ockam/ockam_api/src/nodes/service.rs#L843) to initialize the services with the trust-context/project data.


implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
```diff
@@ -28,8 +28,12 @@ pub async fn enroll_user(app: &AppHandle<Wry>) -> Result<()> {
         .await
         .map(|i| info!("Enrolled a new user with identifier {}", i))
         .unwrap_or_else(|e| error!("{:?}", e));
-    create_relay(&app_state).await?;
     app.trigger_global(crate::app::events::SYSTEM_TRAY_ON_UPDATE, None);
+    // Reset the node manager to include the project's setup, needed to create the relay.
```
etorreborre, line 32: 🆗

implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
```diff
@@ -60,7 +66,8 @@ async fn enroll_with_token(app_state: &AppState) -> Result<IdentityIdentifier> {
 
 async fn retrieve_space(app_state: &AppState) -> Result<Space> {
     info!("retrieving the user space");
-    let node_manager = app_state.node_manager.get().read().await;
+    let node_manager_worker = app_state.node_manager().await;
```
etorreborre: Maybe we should rename `app_state.node_manager()` to `app_state.node_manager_worker()`. 
I was going the ambiguity at first but now that both are more exposed it's probably better to be more explicit.

implementations/rust/ockam/ockam_app/src/shared_service/relay/create.rs
```diff
@@ -1,23 +1,45 @@
...
     Ok(())
 }
+
+/// Create a relay at the default project
+pub async fn create_relay_impl(
+    context: &Context,
+    cli_state: &CliState,
+    node_manager: &NodeManagerWorker,
+) -> Result<Option<ForwarderInfo>> {
+    match cli_state.projects.default() {
```
etorreborre, line 30: Neat

etorreborre (dismissed):
🍔 Great!

implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
adrianbenavides: Done!

etorreborre (approved):

PR #5461, opened 2023-07-31
test(rust): fix the bats test verifying credentials

etorreborre:
This test does not pass because an incorrect issuer was specified for checking a credential.


metaclips (approved):
Thanks

PR #5462, opened 2023-07-31
pin dprint to version 0.39.1

metaclips:
This PR pins dprint to an old version `0.39.1` which still supports the `toml-0.5.4.wasm` format

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -63,11 +63,10 @@ jobs:
         with:
           ref: ${{ github.event.inputs.commit_sha }}
       - name: Install Dprint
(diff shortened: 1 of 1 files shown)


shanesveller (approved):
Can we please also create a separate tracking issue for:
- Confirm whether we want to keep using `dprint` for this concern going forward, which I think is up for discussion
- If so, determine what changes need to be made in-repo in order to be compatible with dprint 0.40+ or what external blockers prevent this
- If not, outline what qualities we need / what risks we're solving for with this tool, identity viable alternatives, make a selection, and create a PR to adopt that one.

PR #5470, opened 2023-08-01
fix(rust): on enroll, persist project data before updating node's project data

adrianbenavides:
At https://github.com/build-trust/ockam/pull/5458 I overlooked the fact that the `add_project_info_to_node_state` function doesn't persist the project data if it's not there (it overwrites the existing data, but doesn't create it if it doesn't exist).


etorreborre (approved):
Thanks @adrianbenavides!

PR #5471, opened 2023-08-01
test(rust): parametrize command-reference.bats tests

adrianbenavides:
- parametrize command-reference.bats tests so they can be run in parallel
- don't check the result of `node delete` or `reset` commands in bats tests, as those commands are run in the `teardown` function after a test is completed


etorreborre (approved):
Thanks @adrianbenavides ✅

PR #5476, opened 2023-08-01
feat(rust): added a direct local kafka for simple deployments

davide-baldo:
Currently it's possible to use a local kafka but it requires a centralized relay node (outlet node), with this direct kafka mode, the sidecar will connect directly to the kafka broker, but only one consumer is allowed.

(needs to be enrolled in a project in order to work)

In this setup the client acts both as consumer and producer for itself (uses kafka as a local queue):
```
ockam  kafka-direct create  --consumer-route ""
```

In this one we spawn 2 sidecar, one referencing the other as a consumer:
```
ockam  kafka-direct create
ockam node create second
ockam  kafka-direct create  --bind-address 127.0.0.1:6092 --brokers-port-range 6001-6100 --consumer-route /node/default --at second
```

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
@@ -59,21 +59,6 @@ impl NodeManager {
             ));
         }
 
-        // Check that there is no entry in the registry with the same tcp address
-        if self
-            .registry
-            .outlets
-            .values()
-            .any(|outlet| outlet.tcp_addr == tcp_addr)
-        {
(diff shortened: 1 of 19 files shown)


implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
```diff
@@ -59,21 +59,6 @@ impl NodeManager {
             ));
         }
 
-        // Check that there is no entry in the registry with the same tcp address
```
davide-baldo, line 62: There is no real issue with having multiple outlets with the same destination

polvorin (approved):

PR #5489, opened 2023-08-02
docs: add a data model for the desktop application

etorreborre:
This PR adds the start of a data model for the desktop application in order to:

 - specify the main concepts exposed by the application, in particular the new notions of "shared service", "invitation", "connected service", etc...
 - write out a few rules around those concepts: 
   - what is the lifecycle of an invitation, 
   - how is the route of a `Service` formed, 
   - what should happen on reset exactly

The purpose is both to:

 - document the application as is it being built
 - serve as a support to think about required features / data (which helps creating stories for the development of the application)

_Note_: the model and its description is really incomplete for now!


documentation/architecture/desktop_application/data-model.md
```diff
@@ -0,0 +1,80 @@
...
+
+The user information (name, email address, ...) is retrieved during the enrollment process and
+displayed in the application.
+
+## Reset
+
+On reset we remove all local state: user information, shared services, connected services etc...
+The default space and project are however not deleted and will be accessed if the user enrolls again.
+
+On reset we remove all remote state:
```
shanesveller: I find these two phrasings slightly contradictory when viewed with my own biases:

> The default space and project are however not deleted

> On reset we remove all remote state

Although you do enumerate just below what actually does get removed. Perhaps we can switch out or remove "all" in the second passage?

documentation/architecture/desktop_application/data-model.md
```diff
@@ -0,0 +1,80 @@
...
+
+There are 3 main activities available to a user when using the application:
+
+1. Enroll / Reset / Quit
+2. Share a local service / Invite a user to a shared service
+3. Accept an invitation / Connect to a remote service
+
+The following sections show the main data types supporting the scenarios above
+
+# Enrollment
```
shanesveller, line 13: Rather than earmark a bunch of spots, I would probably capitalize `Space`, `Project`, `Project Node` and perhaps "invitation" - let me know if you think the slight informality is preferable in this context.

documentation/architecture/desktop_application/data-model.md
```diff
@@ -0,0 +1,80 @@
...
+
+![enroll](./diagrams/enroll.png)
+
+When a user enrolls with the application:
+
+ 1. a space is either created or retrieved for them (if they previously enrolled with the same email address)
+ 2. a project is either created or retrieved for them (if they previously enrolled with the same email address)
+ 3. a relay is created in order to support the routing of all the shared services
```
shanesveller: Can we link to the dedicated docs page on the word `relay`, the first time it appears in this document? Same with secure channel, probably?

documentation/architecture/desktop_application/data-model.md
```diff
@@ -0,0 +1,80 @@
...
+     3. the route of the service is the concatenation of the route of the relay + the TCP outlet route
+
+## Inviting
+
+![invite](./diagrams/invite.png)
+
+In order to create an invitation to a service we need to:
+
+1. select a shared service
+2. choose an expiration date (default: 5 days)
```
shanesveller: The backend, if not supplied, will choose 14 days' expiration. If I set a different default in my CLI PR that was a mistake on my part.

documentation/architecture/desktop_application/data-model.md
```diff
@@ -0,0 +1,80 @@
...
+3. select an email address for the invitee
```
shanesveller, line 67: This point doesn't require change to the doc, just mentioning it for transparency about the backend implications. In the Desktop use-case, we will consider a recipient email mandatory, but in the more general case (like what the CLI implements), an invitation to join a Project or Space can be generated without limiting it to a specific recipient by email address, so the field is optional for those types of invitations. Again, out of scope to document this for material specifically about the Desktop app.

documentation/architecture/desktop_application/data-model.md
```diff
@@ -0,0 +1,80 @@
...
+4. create a token containing all the above information
+
+*TODO*: specify how the token must be signed so that the recipient can verify that
+the sender possesses the correct email address.
```
shanesveller: Can we talk about where this goal/stipulation comes from, what risk it addresses, etc.? It's not something that we currently account for in the existing solution.

documentation/architecture/desktop_application/data-model.md
```diff
@@ -0,0 +1,80 @@
...
+
+All invitations must be persisted on the Orchestrator so that, for each service owner,
+it is possible to tell which invitations have been accepted (expired invitations must be removed).
+
+The invitations are also persisted in the desktop app. A refresh action can be added to query the
```
shanesveller, line 75: I haven't PRed the branch yet but this is accounted for in my code that provides the menu-bar UI for invitations.

documentation/architecture/desktop_application/diagrams/invite.puml
```diff
@@ -0,0 +1,48 @@
...
+skinparam monochrome true
+skinparam defaultFontName Courier
+skinparam roundCorner 0
+skinparam wrapWidth 400
+
+class Invitation {
+  token: InvitationToken
+  expires_at: Timestamp
+  status: InvitationStatus
+  to: Email
```
shanesveller, line 11: As I mentioned in another comment, at the data-modeling level `recipient_email` is optional but will always be populated for Invitations with a `service` scope, which are the only kinds of invitations the Desktop app will be especially interested in at this stage.

On the backend, we do also allow invitations to grant access to an entire Project or Space using the same mechanisms and workflows, while Service invitations are a specialized flavor with additional metadata needed for network-level traffic routing.

I don't know how in the weeds you want this diagram to be/become. If you like, I can drop the Elixir syntax of the CBOR representations involved as a reply on this comment thread.

That said, I am hesitant for the team to invest a lot of energy to document this sub-topic until I feel that the backend design has stabilized sufficiently. Right now I anticipate generating a bunch of churn or "well actually this is really more like..." discussions. Can we perhaps defer this diagram and related prose for a later PR?

documentation/architecture/desktop_application/data-model.md
etorreborre: Changed.

documentation/architecture/desktop_application/data-model.md
etorreborre: I tend to use the informal capitalization unless I refer to something more formal, in a diagram, or in some code.

documentation/architecture/desktop_application/data-model.md
etorreborre: Thanks. I actually did not know if a value had been selected already.

documentation/architecture/desktop_application/data-model.md
etorreborre: I wrote this to raise the discussion :-). I am going to replace the TODO with something like "specify the content and properties of the token"

documentation/architecture/desktop_application/data-model.md
etorreborre: I have added two links.

documentation/architecture/desktop_application/data-model.md
etorreborre: Yes, I kept the 2 topics separated on purpose

documentation/architecture/desktop_application/diagrams/invite.puml
etorreborre: The purpose of this document is to give a conceptual model of how things work for the desktop app at a high-level, in order to discuss rules and features.

The exact data structures and how they are shared across other purposes (like a general mechanism for adding members to a space or a project) should not be part of this document. It is meant to be more of a specification rather than a design document. For example this is where we should specify the lifecycle of an invitation and the purpose of a field like `remaining uses` (what's the initial value, how is decremented, what happens when it reaches 0?).

documentation/architecture/desktop_application/data-model.md
adrianbenavides: The `remove all remote state` is still there right?

etorreborre: Sorry I hadn't pushed yet

documentation/architecture/desktop_application/data-model.md
```diff
@@ -30,12 +32,15 @@ displayed in the application.
...
-On reset we remove the following remote state:
-  - invitations
-  - relays
+We also remove the following remote state: invitations and relays. The default space and project are however not
+deleted and will be accessible if the user enrolls again.
+
+## Quit
+
+On quit, the application's node is killed but the local state is kept. When the application is restarted,
+a new node is created restoring its previous state using the local and remote state.
```
etorreborre: We should be a bit more precise here because we don't really create a node, the app _acts as a node_ with its own `NodeManager`. Maybe just adding `in memory node` instead of just `node`

adrianbenavides: Good point, updating it!

adrianbenavides (approved):

davide-baldo (approved):

PR #5352, opened 2023-07-14
feat: Add Tailwind-based styling to desktop scaffold

shanesveller:
## Current behavior

The desktop app is currently un-styled and straightforward.

## Proposed changes

Install Tailwind CSS, configure SvelteKit's top-level layout to enable it, and populate a shared navbar.

Configure Tailwind to honor our [public style guide](https://www.ockam.io/style-guide), including minimal adherence to font choices.

Move the enrollment UI to a sub-route, primarily for educational reasons rather than notable technical benefit.

## Checks

- [X] All commits in this Pull Request are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and Verified by Github.
- [X] All commits in this Pull Request follow the Ockam [commit message convention](https://github.com/build-trust/.github/blob/main/CONTRIBUTING.md#commit-messages).
- [X] I accept the Ockam Community [Code of Conduct](https://github.com/build-trust/.github/blob/main/CODE_OF_CONDUCT.md).
- [X] I have accepted the Ockam [Contributor License Agreement](https://github.com/build-trust/ockam-contributors/blob/main/CLA.md) by adding my Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/blob/main/CONTRIBUTORS.csv) file in a separate pull request to the [build-trust/ockam-contributors](https://github.com/build-trust/ockam-contributors) repository. The easiest way to do this is to [edit the CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/edit/main/CONTRIBUTORS.csv) file in the github web UI and create a PR, this will mark the commit as verified.

PR #5490, opened 2023-08-02
feat(typescript): style the tcp-outlet create popup window

adrianbenavides:
- Built on top of @shanesveller's PR https://github.com/build-trust/ockam/pull/5352
- Use a fixed-size popup window, that will be displayed in the top-right corner
- Show error message received from backend
- Use tailwindcss to style the window using a similar style to that from MacOS settings


implementations/typescript/ockam/ockam_app/src/routes/service/ServiceCreate.svelte
```diff
@@ -9,22 +9,28 @@
   }
 </script>
 
-<div>
-  <h1>Service details</h1>
-  <div>
-    <label for="name">Service that you want to share:</label>
-    <div>
-      <input id="name" placeholder="{service}" bind:value="{service}" />
+<div class="border-b mb-4 pb-2 font-bold text-xl">Service details</div>
```
etorreborre, line 22: Doesn't this go against the idea of having a separate section for styling in Svelte?
If that's possible I see the following advantages:

 - the structure of the UI is more obvious (now it's obfuscated by the styling)
 - a lot less attributes have to be repeated. For example `class="w-full px-4 bg-transparent border-none focus:outline-none text-right"` for the input fields

etorreborre (dismissed):
The styling of that window is nice but I have some doubts about the maintainability in the long term

implementations/typescript/ockam/ockam_app/src/routes/service/ServiceCreate.svelte
adrianbenavides: I was looking into it, and it seems we can compose custom classes out of tailwind's components. I'll refactor this to extract the components structure into single classes.

shanesveller: > Doesn't this go against the idea of having a separate section for styling in Svelte?

Also a yes from my POV, but Tailwind's utility classes are themselves a productivity advantage, and leaning into that was a design goal I was pursuing during #5352. Having worked in backend and server-side-rendered HTML for years and full-stack development with React for a lesser number of years, I've found "semantic" HTML classes to be one of the biggest and most wasteful bikeshed topics of my entire career.

shanesveller: I'll readily admit this is an appeal-to-authority, but the creator of Tailwind himself believes the `apply` feature to be a mistake/antipattern and I agree based on my own experiences.

https://twitter.com/adamwathan/status/1226511611592085504

> Confession: The `apply` feature in Tailwind basically only exists to trick people who are put off by long lists of classes into trying the framework.

> You should almost never use it 😬

> Reuse your utility-littered HTML instead.

shanesveller: > the structure of the UI is more obvious

Generally the idiomatic way to preserve this property when embracing Tailwind means encapsulating the "noisy" utility classes into in-house reusable components with props, like `<ServiceTable items={...} />` or `<ActionButton>`/`<CancelButton>`, and so on. You build up a UI component library pretty fast and as a virtuous byproduct you get more consistent styling/flavor throughout the project rather than 8 subtly different `<button>` elements.

implementations/rust/ockam/ockam_app/src/shared_service/tray_menu.rs
```diff
@@ -36,11 +37,19 @@ pub(crate) async fn build_shared_services_section(
 
 /// Event listener for the "Create..." menu item
 pub fn on_create(app: &AppHandle<Wry>) -> tauri::Result<()> {
-    tauri::WindowBuilder::new(
+    let w = tauri::WindowBuilder::new(
```
shanesveller: Have you by any chance identified the syntax needed to either get an existing window by ID, then fall back to create it on the fly as you do here? I've only done minimal research because "show existing window" and "create net-new window" are both straightforward, but for this and my invitations use-case I think not erroring if the window is already displayed, just bringing it into the foreground or marking it as always-on-top, would be an improvement.

shanesveller (dismissed):
New UI looks great to me! My question about reusing existing windows is absolutely not a blocker, just an opportunity for polish.

implementations/typescript/ockam/ockam_app/src/routes/service/ServiceCreate.svelte
etorreborre: > the biggest and most wasteful bikeshed topics of my entire career

That's a strong statement given the number of possible bikesheds :-).

> Generally the idiomatic way to preserve this property when embracing Tailwind means encapsulating the "noisy" utility classes into in-house reusable components with props, like <ServiceTable items={...} /> or <ActionButton>/<CancelButton>, and so on

That's what I was thinking of. Isn't that what Svelte encourages you to do as well?

shanesveller: Svelte at least partially encourages you to write traditional CSS in-line within your component files. My recommendation is that we should either do that and skip adopting Tailwind, or lean into Tailwind's utility classes and skip the `<style>` element on our Svelte components in 99% of cases, but not use both concurrently.

adrianbenavides: I'm not seeing myself writing the css code to support flex layouts, to be honest 😆 

At the end of the day, we want to use basic components that are already tested and available in any modern css framework. I would only use svelte inline styles to customize something very particular to that view.

What I can try though is writing a component for that kind of view, but I didn't want to go into that just yet, as we don't know if we are going to reuse it.

implementations/rust/ockam/ockam_app/src/shared_service/tray_menu.rs
adrianbenavides: I've done some changes and will look like this:
```rust
match app.get_window(SHARED_SERVICE_WINDOW_ID) {
    None => .. // Create from scratch
    Some(w) => w.show()?,
}
```

Is that what you mean?

shanesveller: Yeap, exactly, thank you.

adrianbenavides:
I've added some tweaks to the view to show the error message returned from the backend, and I've added a "Cancel" button to close the window.

etorreborre (approved):

Issue #5255, opened 2023-07-06
Parse `Duration` arguments with custom units (ms, s, m)


adrianbenavides:
### Current behavior

Most of the duration-related arguments used in the Ockam Command are integers that we later use in code to build a `Duration` instance. We explicit in the arguments' description the expected units, using that assumption to build the `Duration::from_secs()`.

### Desired behavior

We would like to give the user the flexibility to express the units as part of the argument. For example:

```
$ ockam message send --to n --timeout 10s
```

Some formats that we would like to support:
- `10`: seconds (the default unit in case none is provided)
- `10s`: seconds
- `10ms`: milliseconds
- `10m`: minutes

### How to do it

There are various libraries implementing this already, but we would like to avoid them because they have too many features that we are not interested in. We can do this with a simple regex to support the few cases we are targeting.

We need to refactor the duration-related arguments so they are parsed with a custom function that does the transformation from a human-duration value to a `Duration` instance. For example, in the [`message send`](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/message/send.rs#L43) command:

```rust
/// Override default timeout (in seconds)
#[arg(long, value_name = "TIMEOUT", default_value = "10")]
pub timeout: u64,
```

should be 

```rust
/// Override default timeout
#[arg(long, value_name = "TIMEOUT", default_value = "10", value_parser = human_duration_parser)]
pub timeout: Duration,
```

MFAlexandru:
Hello, I am a first time contributor, I would like to look into this, is it still open ?

etorreborre:
Hi @MFAlexandru. Absolutely, please create [a PR in the `ockam-contributors` repository](https://github.com/build-trust/ockam-contributors/pulls?q=is%3Apr+is%3Aclosed) if you are contributing for the first time and it's all yours!

hlieberman:
Hi @MFAlexandru!  Are you still interested in working on this? I was taking a look at it myself, but I don't want to step on work you've done if it's something you're working on.

MFAlexandru:
No, it is all yours

etorreborre:
Hi @hlieberman, please go ahead!

etorreborre:
This has now been done with the latest work on it in https://github.com/build-trust/ockam/pull/5677

PR #5498, opened 2023-08-02
Add human-writable duration support (Closes: #5255)

hlieberman:
## Current behavior

Currently, durations are supplied to the ockam command with a fixed unit attached (e.g., timeout in ms).  This PR adds support for specifying "human-writable" durations, such as "10s", "1m", or "150ms". 

## Proposed changes

This PR only supports milliseconds, seconds, and minutes, as was discussed in #5255.  Note, I'm not 100% sure that the test here provides juice that's worth the squeeze; it's a pretty close reimplementation that I can't see catching a lot of errors, with an additional library on top for testing to boot.  That being said, it's at least simple enough to read through.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/util/duration.rs
+++ b/implementations/rust/ockam/ockam_command/src/util/duration.rs
@@ -1,0 +1,59 @@
+use clap::error::{Error, ErrorKind};
+use once_cell::sync::OnceCell;
+use regex::Regex;
+use std::time::Duration;
+
+static DURATION_REGEX: OnceCell<Regex> = OnceCell::new();
+
+pub(crate) fn duration_parser(arg: &str) -> Result<Duration, clap::Error> {
+    let needles = DURATION_REGEX
+        .get_or_init(|| {
+            Regex::new(r"(?P<numeric_duration>[0-9]+)(?P<length_sigil>m|s|ms)?$").unwrap()
+        })
+        .captures(arg)
+        .ok_or(Error::raw(ErrorKind::InvalidValue, "Invalid duration."))?;
+    let time = needles["numeric_duration"]
+        .parse::<u64>()
+        .map_err(|_| Error::raw(ErrorKind::InvalidValue, "Invalid duration."))?;
+
+    match needles.name("length_sigil") {
+        Some(n) => match n.as_str() {
+            "ms" => Ok(Duration::from_millis(time)),
+            "m" => Ok(Duration::from_secs(60 * time)),
+            "s" => Ok(Duration::from_secs(time)),
+            _ => unreachable!("Alternatives excluded by regex."),
+        },
+        None => Ok(Duration::from_secs(time)),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use proptest::prelude::*;
+    use std::time::Duration;
+
+    use crate::util::duration::duration_parser;
+
+    const ONE_YEAR_MILLIS: u64 = 31_536_000_000;
+
+    proptest! {
+        #[test]
+        fn test_reverse_duration(arg in 0..ONE_YEAR_MILLIS) {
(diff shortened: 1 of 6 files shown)


implementations/rust/ockam/ockam_command/src/util/duration.rs
```diff
@@ -0,0 +1,59 @@
+use clap::error::{Error, ErrorKind};
+use lazy_static::lazy_static;
+use regex::Regex;
+use std::time::Duration;
+
+lazy_static! {
+    static ref DURATION_REGEX: Regex =
+        Regex::new(r"(?P<numeric_duration>[0-9]+)(?P<length_sigil>m|s|ms)?$").unwrap();
+}
```
adrianbenavides: We use `once_cell`, which is less verbose. Check [this out as an example](https://github.com/build-trust/ockam/blob/13f093392ea1e2a8a44dad71870996b5fdc746b3/implementations/rust/ockam/ockam_command/src/run/parser.rs#L293).

adrianbenavides (commented):
Awesome, thank you so much @hlieberman for helping us with this one! 

I've left a minor comment on `lazy_static vs once_cell`, but other than that, looks pretty nice 👏

implementations/rust/ockam/ockam_command/src/util/duration.rs
adrianbenavides: Now that I think about it, [`OnceCell` is already part of the std library](https://doc.rust-lang.org/stable/std/cell/struct.OnceCell.html) since 1.70, which we support on our CI pipeline, so feel free to use that version instead.

hlieberman: Sounds good.  I've switched it to use `std::sync::OnceLock`, since statics need to be thread-safe.

hlieberman:
A meta-question about how y'all prefer things.  When you have a PR like this, do you prefer the patches force-pushed on top to keep up with develop, or do you like them as a flow once it's been cut off of main develop and then rebased/squashed at the end?

adrianbenavides:
> A meta-question about how y'all prefer things. When you have a PR like this, do you prefer the patches force-pushed on top to keep up with develop, or do you like them as a flow once it's been cut off of main develop and then rebased/squashed at the end?

We usually squash all the PR's commits into one unless the individual commits are significant by themselves. For example, we try to avoid merging into develop commits like "fix clippy warnings". In this case, I'd squash them into one using the first commit description.

Regarding rebasing on top of develop, you don't need to do that unless there are conflicts, so don't worry about it. 

One last thing: make sure you run `cargo fmt` and `cargo clippy` to fix any potential formatting issue. Also, the commit linter doesn't allow raw uppercase, so you should write it like ``feat(rust): switch `util::Duration` to use `OnceLock` ``.

implementations/rust/ockam/ockam_command/src/util/duration.rs
adrianbenavides: We can't use the `std` types yet, as CI is using rust's 1.68 version. There is a [PR in progress](https://github.com/build-trust/ockam/pull/5514) to upgrade it to the latest version, but I'd recommend you to use once_cell in the meantime. We'll do the switch to std after that PR gets merged.

hlieberman:
>  Also, the commit linter doesn't allow raw uppercase, so you should write it like `` feat(rust): switch `util::Duration` to use `OnceLock` ``.

Aha, that's what it was yelling at me about.  I kept checking that I hadn't somehow capitalized `feat(rust)` and just misread it.

Updated to use `once_cell`.

adrianbenavides (approved):
Great, LGTM! Thanks @hlieberman for an awesome first contribution 🥳

SanjoDeundiak (approved):

PR #5514, opened 2023-08-03
update rust version

metaclips:
Update builder image to use Rust version 1.71.1, and also version of healthcheck base image to debian 12.1


shanesveller:
Can you please bump the Rust version specified in `tools/nix/flake.Nix` as well? No need to change the `default` value in the adjacent file, though.

https://github.com/build-trust/ockam/blob/0642d5eb39eba02daabe25f3f383525eb1c43970/tools/nix/flake.nix#L26

This potentially also needs a `flake.lock` bump for the rust-overlay input, if the new cron workflow in #5473 was not successful, but that aspect does not need to block the merge of this PR.

As we've discussed before, adopting a [rust-toolchain.toml](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) file will be a preferred longer-term solution, because we can have our Docker builder image, contributors that do use Nix, and contributors that do not use Nix all honoring the same specification with only one place to change to cover them all.

adrianbenavides:
@metaclips I've removed an assert on the bats `force_kill_node` helper method, so we can let the test fail on its own when doing the operations after calling that function (e.g. [recreating the node here](https://github.com/build-trust/ockam/blob/ef037e3a19f29fc96df0d862db37326d5747d4d8/implementations/rust/ockam/ockam_command/tests/bats/nodes_unit.bats#L67)). Tests are passing now, so the fix should be ok.

.github/workflows/release-binaries.yml
```diff
@@ -319,22 +319,12 @@ jobs:
...
-        with:
-          args: |
-            bash -c "export PRIVATE_KEY='${{ secrets.COSIGN_PRIVATE_KEY }}' && export COSIGN_PASSWORD='${{ secrets.COSIGN_PRIVATE_KEY_PASSWORD }}';
-            rustup target add ${{ matrix.target }};
-            set -xe && pwd && uname -a && ls && cargo clean && cargo build --release --package ockam-ffi --target ${{ matrix.target }};
-            rm -rf target/release && mv target/${{ matrix.target }}/release target
+        run: |
+          cargo install --version 0.1.16 cross
+          cross build --release --package ockam-ffi --target ${{ matrix.target }}
+          rm -rf target/release && mv target/${{ matrix.target }}/release target
```
shanesveller:
```suggestion
          rm -rf target/release && mv target/${{ matrix.target }}/release target/
```

rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "1.71.1"
+components = [ "rustfmt", "clippy" ]
```
shanesveller: Let's keep lists alphabetized pretty much anywhere it doesn't break the intended behavior.

adrianbenavides (approved):

PR #5520, opened 2023-08-04
feat(rust): email verification

polvorin:
on enroll, ensure email verification has been completed before continuing with the enrollment process.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
+++ b/implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
@@ -119,7 +119,7 @@ impl OidcService {
     }
 
     /// Return the information about a user once authenticated
-    pub async fn get_user_info(&self, token: OidcToken) -> Result<UserInfo> {
+    pub async fn get_user_info(&self, token: &OidcToken) -> Result<UserInfo> {
         let client = self.provider().build_http_client()?;
         let access_token = token.access_token.0.clone();
         let req = || {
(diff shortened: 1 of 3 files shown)


implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
```diff
@@ -134,6 +134,33 @@ impl OidcService {
...
     }
 
+    pub async fn wait_for_email_verification(
+        &self,
+        token: &OidcToken,
+        opts: &CommandGlobalOpts,
+    ) -> Result<UserInfo> {
+        let spinner_option = opts.terminal.progress_spinner();
+        loop {
+            let user_info = self.get_user_info(token).await?;
```
davide-baldo, line 144: There is a decent chance we hit a 429 `Too Many Requests`, current retry strategy go up to maximum 1 second: 
```
let retry_strategy = ExponentialBackoff::from_millis(10).take(3);
```
I think we should either relax it a little or handle 429 in a slightly different way

davide-baldo (approved):
Looks good

implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
polvorin: right, I'll create a ticket to handle this error more gracefully.    Here we are using a fixed 5 seconds delay between attempts,   we can tweak that too (although  I didn't want to use a large max backoff  time due to not being user frienly:  we are blocking the ui until the user verify his email,   if we take too long _after_ it has verified it to respond, it will be confusing)

polvorin: https://github.com/build-trust/ockam/issues/5630

PR #5521, opened 2023-08-04
refactor(rust): using parser to obtain identity identifier

### LINKED ISSUE #5225: Change argument for `ockam credential issue` command from Identity to IdentityIdentifier
### Current behavior

The `ockam credential issue` [command](https://github.com/build-trust/ockam/tree/344c4736e31c3b104bdc1a31f133603b4377138e/implementations/rust/ockam/ockam_command/src/credential/issue.rs#L23) expects the full identity of the subject as the `for` argument.

### Desired behavior

We can use the `IdentityIdentifier` instead of the full `Identity` as the `for` argument. 

Instead of decoding the `Identity` [here](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/credential/issue.rs#L54C18-L54C26) we can simply read the `for` argument as an `IdentityIdentifier` (using a `value_parser`, [like this is done for a `SocketAddr`](https://github.com/build-trust/ockam/blob/344c4736e31c3b104bdc1a31f133603b4377138e/implementations/rust/ockam/ockam_command/src/tcp/outlet/create.rs#L38) for example).

Then as we can see on [this line](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/credential/issue.rs#L95) we just use the identifier so that line can be simplified.

documentation needs to be updated as well.

GabrielBarbosaGV:
## Current behavior
As of now, to obtain the Identity Identifier from the Credential Issue command, an Identity is generated first, and then the Identifier is derived from it.

## Proposed changes
Use ockam_identity::IdentityIdentifier::decode_hex() directly, avoiding the overhead of generating an Identity from which to obtain the Identifier. To transform the received input into actionable data, a parser is applied to the input.

resolves #5225 

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/util/parsers.rs
+++ b/implementations/rust/ockam/ockam_command/src/util/parsers.rs
@@ -1,7 +1,9 @@
 use crate::Result;
 use miette::miette;
+use ockam_identity::IdentityIdentifier;
 use ockam_transport_tcp::resolve_peer;
 use std::net::SocketAddr;
+use std::str::FromStr;
 
 /// Helper function for parsing a socket from user input
 /// It is possible to just input a `port`. In that case the address will be assumed to be
@@ -17,6 +19,13 @@ pub(crate) fn socket_addr_parser(input: &str) -> Result<SocketAddr> {
     };
     Ok(resolve_peer(address.to_string())
         .map_err(|e| miette!("cannot parse the address {address} as a socket address: {e}"))?)
+}
+
+/// Helper fn for parsing an identity from user input by using
+/// [`ockam_identity::IdentityIdentifier::from_str()`]
+pub(crate) fn identity_identifier_parser(input: &str) -> Result<IdentityIdentifier> {
+    IdentityIdentifier::from_str(input)
+        .map_err(|_| miette!("Invalid identity identifier: {input}").into())
 }
 
 #[cfg(test)]
--- a/implementations/rust/ockam/ockam_command/src/credential/issue.rs
+++ b/implementations/rust/ockam/ockam_command/src/credential/issue.rs
@@ -2,7 +2,7 @@ use ockam_core::compat::collections::HashMap;
 
 use crate::identity::{get_identity_name, initialize_identity_if_default};
 use crate::{
-    util::{node_rpc, print_encodable},
+    util::{node_rpc, parsers::identity_identifier_parser, print_encodable},
     vault::default_vault_name,
     CommandGlobalOpts, EncodeFormat, Result,
 };
@@ -12,15 +12,15 @@ use miette::{miette, IntoDiagnostic};
 use ockam::identity::CredentialData;
 use ockam::Context;
 use ockam_api::cli_state::traits::{StateDirTrait, StateItemTrait};
-use ockam_identity::{identities, Identity};
+use ockam_identity::IdentityIdentifier;
 
 #[derive(Clone, Debug, Args)]
 pub struct IssueCommand {
     #[arg(long = "as")]
     pub as_identity: Option<String>,
 
-    #[arg(long = "for", value_name = "IDENTITY_ID")]
-    pub for_identity: String,
+    #[arg(long = "for", value_name = "IDENTIFIER", value_parser = identity_identifier_parser)]
+    pub identity_identifier: IdentityIdentifier,
 
     /// Attributes in `key=value` format to be attached to the member
     #[arg(short, long = "attribute", value_name = "ATTRIBUTE")]
@@ -51,17 +51,8 @@ impl IssueCommand {
         Ok(attributes)
     }
 
-    pub async fn identity(&self) -> Result<Identity> {
-        let identity_as_bytes = match hex::decode(&self.for_identity) {
-            Ok(b) => b,
-            Err(e) => return Err(miette!(e).into()),
-        };
-
-        let identity = identities()
-            .identities_creation()
-            .decode_identity(&identity_as_bytes)
-            .await?;
-        Ok(identity)
+    pub fn identity_identifier(&self) -> IdentityIdentifier {
+        self.identity_identifier.clone()
     }
 }
 
@@ -92,7 +83,7 @@ async fn run_impl(
     let issuer = ident_state.identifier();
 
     let credential_data =
-        CredentialData::from_attributes(cmd.identity().await?.identifier(), issuer.clone(), attrs)
+        CredentialData::from_attributes(cmd.identity_identifier(), issuer.clone(), attrs)
             .into_diagnostic()?;
     let credential = identities
         .credentials()
--- a/implementations/rust/ockam/ockam_command/src/credential/verify.rs
+++ b/implementations/rust/ockam/ockam_command/src/credential/verify.rs
@@ -1,22 +1,24 @@
 use std::path::PathBuf;
 
 use crate::{
-    fmt_err, fmt_log, fmt_ok, util::node_rpc, vault::default_vault_name, CommandGlobalOpts, Result,
+    fmt_err, fmt_log, fmt_ok, util::node_rpc, vault::default_vault_name, CommandGlobalOpts,
 };
 use miette::miette;
 
 use clap::Args;
 use colorful::Colorful;
 use ockam::Context;
-use ockam_identity::{identities, Identity};
+use ockam_identity::IdentityIdentifier;
 use tokio::{sync::Mutex, try_join};
+
+use crate::util::parsers::identity_identifier_parser;
 
 use super::validate_encoded_cred;
 
 #[derive(Clone, Debug, Args)]
 pub struct VerifyCommand {
-    #[arg(long = "issuer")]
-    pub issuer: String,
+    #[arg(long = "issuer", value_name = "IDENTIFIER", value_parser = identity_identifier_parser)]
+    pub issuer: IdentityIdentifier,
 
     #[arg(group = "credential_value", value_name = "CREDENTIAL_STRING", long)]
     pub credential: Option<String>,
@@ -33,16 +35,8 @@ impl VerifyCommand {
         node_rpc(run_impl, (opts, self));
     }
 
-    pub async fn issuer(&self) -> Result<Identity> {
-        let identity_as_bytes = match hex::decode(&self.issuer) {
-            Ok(b) => b,
-            Err(e) => return Err(miette!(e).into()),
-        };
-        let identity = identities()
-            .identities_creation()
-            .decode_identity(&identity_as_bytes)
-            .await?;
-        Ok(identity)
+    pub fn issuer(&self) -> IdentityIdentifier {
+        self.issuer.clone()
     }
 }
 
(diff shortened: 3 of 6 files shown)


mrinalwadhwa:
@GabrielBarbosaGV Thank you for spending time on this, while we review please

- [x] Accept the Ockam [Contributor License Agreement](https://github.com/build-trust/ockam-contributors/blob/main/CLA.md) by adding your Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/blob/main/CONTRIBUTORS.csv) file in a separate pull request to the [build-trust/ockam-contributors](https://github.com/build-trust/ockam-contributors) repository.

The easiest way to do this is to [edit the CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/edit/main/CONTRIBUTORS.csv) file in the github web UI and create a PR, this will make the commit as verified.

GabrielBarbosaGV:
@mrinalwadhwa I mistakenly had opened a PR to my own repository, but have now corrected this by creating a PR directly to build-trust's, from the branch in which I had made modifications. Thank you!

SanjoDeundiak:
@GabrielBarbosaGV could you please update the corresponding bats test that is now failing? It's located in `/implementations/rust/ockam/ockam_command/tests/bats/trust_context.bats`. The test name is `trust context - trust context with an offline authority; Credential Exchange is performed`

GabrielBarbosaGV:
Hello @SanjoDeundiak! Unfortunately, I am getting this error when trying to run the bats tests:

image

Since I am on Linux, I also had a brew related error, but there was a commented line right after the troublesome one, and swapping them fixed that. However, I am not sure why the screenshot errors are happening. What could I do to fix this?

As for the failing test, I must confess I am not aware of why it does not pass - to me, it seems that the generated identities are properly hex encoded, and obtained by using

```bash
ockam identity show
```

This leads me to believe that the error would be in the new implementation rather than the test. Is this correct?

Many thanks!

SanjoDeundiak:
@GabrielBarbosaGV the problem is that this bats test passes the whole Identity as an `--for` argument to `credential issue` command. But in this PR you changed that command so that now it expects the Identifier in place of this argument. So the failing test should be updated to align with the new command behaviour. To get the Identifier instead of the full Identity you need to use `ockam identity show alice` instead of `ockam identity show alice --full --encoding hex`.

Could you please provide more information about how to reproduce the other error that you're having?

GabrielBarbosaGV:
@SanjoDeundiak Ah, I see - thank you! I will implement this as soon as I am able to get the bats tests running. As for the setup to get the error, I did not do much, only having cd'd into

```bash
$OCKAM_DIR/implementations/rust/ockam/ockam_command/tests/bats
```

and then run

```bash
bats ./trust_context.bats
```

What information could I provide that would be more of use?

Many thanks!

adrianbenavides:
Hey @GabrielBarbosaGV ! Let's see if I can help you debugging the bats tests problem. I'd say there is an issue with the binary that is being used by the bats tests. Could you try running the following?

First, let's build the binary:

```bash
cargo build -p ockam_command
```

Let's make sure the following command exists:

```bash
./target/debug/ockam node delete --all --force --yes
```

And finally, try running the bats tests specifying the path to the ockam binary:

```bash
OCKAM=./target/debug/ockam bats ./implementations/rust/ockam/ockam_command/tests/bats/trust_context.bats
```

SanjoDeundiak:
@GabrielBarbosaGV bats tests, by default, use the ockam binary that your command line points to. You can run `which ockam` to see what is the path to it. The latest binary that you build is in `./target/debug/ockam`. It appears that the `--yes` flag doesn't exist in the latest released version but exists in the current develop branch, so you're probably using the wrong version. Please use @adrianbenavides instruction above to redirect bats tests to the right binary.

GabrielBarbosaGV:
@adrianbenavides @SanjoDeundiak Good day! It is indeed as you pointed out, and tests now run. Thank you!

I have not, however, been able to make the tests pass by changing

```bash
ockam identity show alice --full --encoding hex
```

to

```bash
ockam identity show alice
```

as I now get

image

I have tried sudo, but the output is unchanged.

implementations/rust/ockam/ockam_command/src/util/parsers.rs
```diff
@@ -30,6 +31,12 @@ pub(crate) fn socket_addr_parser(input: &str) -> Result<SocketAddr> {
     }
 }
 
+/// Helper fn for parsing an identity from user input by using
+/// [`ockam_identity::IdentityIdentifier::decode_hex()`]
+pub(crate) fn identity_identifier_parser(input: &str) -> Result<IdentityIdentifier> {
+    Ok(IdentityIdentifier::from_hex(input))
```
adrianbenavides: If we now use identifiers when calling `ockam identity show`, then this function should expect the plain identifier, not the `hex` encoded version. So this should be:

```rust
IdentityIdentifier::from_str(input)
    .map_err(|_| miette!("Invalid identity identifier: {input}").into())
```

adrianbenavides (commented):
Regarding the failing tests, looks like you will also need to modify the `credential store` and `credential verify` commands. 

Note that the `issuer` argument is being parsed as an encoded full identity, where it should be an IdentityIdentifier. 

I guess it should be as simple as using the `identity_identifier_parser` you already have in the `credential issue` command. The attribute names should be kept though.

implementations/rust/ockam/ockam_command/src/util/parsers.rs
GabrielBarbosaGV: I see! Once this change is made, and the parser is utilized, how may I obtain the Identity from the IdentityIdentifier? The credential store command, for instance, needs the IdentityIdentifier, and obtains it by hex decoding the string, but later also needs the Identity, when it passes it to

```rust
CredentialConfig::new()
```

Also, after the modification, should further changes be made to the bats tests?

Many thanks for your help!

SanjoDeundiak: You can get Identifier from an Identity, but not vice versa. Does that make sense?

GabrielBarbosaGV: I see - I'd thought there was a one-to-one mapping between the two, and, from the Identifier, one might recover the Identity itself. Although, as the command is receiving the value for the first time, there would be nowhere to go to retrieve it, only create it - although I am still to get a reasonable grasp of the concepts. If it is not possible to obtain the Identity from the Identifier, then can the parser not be used for `credential store` and `credential verify`?

Two other things I only now realize are that I neglected to document the changes to the command, as I did not see that the format of the `--for` argument would change, and that perhaps, from that, calling this a refactor might be inappropriate. I apologize for the back-and-forth, but hope I can learn from the experience and better contribute for it.

Given this Identifier-Identity situation, what would be best, and where is the place for editing the corresponding documentation?

Many thanks!

GabrielBarbosaGV: @adrianbenavides @SanjoDeundiak Good day! Just bumping as I am stuck - apologies.

adrianbenavides: Hey @GabrielBarbosaGV ! I wanted to respond earlier but we had a busy week, sorry about that 🙏  

> If it is not possible to obtain the Identity from the Identifier, then can the parser not be used for credential store and credential verify?

I'm not sure I'm understanding the issue here, but this might help: in the `verify` command, the `issuer` argument is expected to be a hex encoded full identity. We then [parse it](https://github.com/build-trust/ockam/blob/18ac04b4f9d5257bdc5b76a0826066abc5c12626/implementations/rust/ockam/ockam_command/src/credential/verify.rs#L36) as an `Identity`, only to retrieve its [identifier](https://github.com/build-trust/ockam/blob/18ac04b4f9d5257bdc5b76a0826066abc5c12626/implementations/rust/ockam/ockam_command/src/credential/verify.rs#L82). So, we don't really need the identity in the first place, just the identifier. the same happens with the `issue` command.

On the other side, the `store` command DOES need the identity to [store the credential](https://github.com/build-trust/ockam/blob/18ac04b4f9d5257bdc5b76a0826066abc5c12626/implementations/rust/ockam/ockam_command/src/credential/store.rs#L99). In this case, we need to pass the full identity.

A good improvement here would be adding documentation to the `issuer` arguments of each of these commands to specify the expected input (hex-encoded identity, identifier). 

> and where is the place for editing the corresponding documentation?

The [`value_name`](https://github.com/build-trust/ockam/blob/18ac04b4f9d5257bdc5b76a0826066abc5c12626/implementations/rust/ockam/ockam_command/src/credential/issue.rs#L22) should be renamed. Other than this, since the argument was not renamed (it's still `--for`) I don't think your changes will affect any code example, docs, tests.

Let me know if I can help with anything else 👍

GabrielBarbosaGV: Thanks for the help, @adrianbenavides! I have made some modifications which I believe will make the bats tests pass as soon as one other thing is dealt with, which is an error when trying to use IdentityIdentifier::from_str. It is failing at this point, in

ockam/implementations/rust/ockam/ockam_identity/src/identity/identity_identifier.rs:92

```rust
impl TryFrom<&str> for IdentityIdentifier {
    type Error = Error;

    fn try_from(value: &str) -> Result<Self> {
        let value = value.trim();
        if value.starts_with(Self::PREFIX) {
            Ok(Self(value.to_string()))
        } else {
            Err(IdentityError::InvalidIdentityId.into())
        }
    }
}
```

The PREFIX is just the uppercase letter "P". What would the argument to `ockam credential issue --for <ARGUMENT>` have to look like to be accepted, but still be a plain identity identifier?

adrianbenavides: This is where it gets tricky.
- You want to use `identifier=$($OCKAM identity show i1 --encoding hex)` when you want to use that variable in `credential issue/verify` commands
- You want to use `idt1=$($OCKAM identity show i1 --full --encoding hex)` when you want to use that variable in `credential store` commands

adrianbenavides: Make sure to pull the latest changes. I've rebased the latest develop changes to fix a conflict.

GabrielBarbosaGV: Good day, @adrianbenavides! I thank you once again for the patience and help.

I've just made changes that pass the bats tests. However, there is a failing Rust test (ckam_command enroll::oidc_service::tests::test_wait_for_authorization_code) whose cause for failure I do not know. How does the failing Rust test relate to my changes?

adrianbenavides: We recently modified those tests because they were flaky on CI. I've squashed your commits and rebased on top of latest develop. I'll do a final pass, but looks like we are ready to merge!

adrianbenavides (dismissed):
All bats tests green, LGTM 🚀  

It was a little tricker than we initially anticipated, but you made it! Thanks again @GabrielBarbosaGV for this great first contribution 🙏

etorreborre (approved):

PR #5574, opened 2023-08-09
feat(rust): support decoding role-in-project from api list response

shanesveller:
## Current behavior

`ockam project list --output json` and similar-in-purpose Rust code does currently not allow the caller to distinguish the access role they hold in the context of each Project they have access to.

## Proposed changes

The Orchestrator backend is also being revised to include `user_roles` details on the Projects show/list endpoints, which in the current implementation will contain 0 or 1 entries. If present, the single record within will include an identifier, email address, and access role for the **calling user** within that project's scope.

The current implementation as of this branch does not account for users which are attached to the Space instead, and `ockam spaces list` will eventually need a corresponding PR like this one to distinguish that level of access.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/schema.cddl
+++ b/implementations/rust/ockam/ockam_api/src/schema.cddl
@@ -48,7 +48,8 @@ project = {
     7: space_id,
     ?8: project_node_identity, ; optional, it can be missing if the cloud node hasn't started yet
     ?9: authority_access_route, ; optional, it can be missing if the authority hasn't started yet
-    ?10: authority_identity ; optional, hex encoded authority identity
+    ?10: authority_identity, ; optional, hex encoded authority identity
+    16: [* project_user_role]
 }
 
 project_node_identity = identity_id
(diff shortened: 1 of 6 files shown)


implementations/rust/ockam/ockam_api/src/schema.cddl
```diff
@@ -48,7 +48,8 @@ project = {
     7: space_id,
     ?8: project_node_identity, ; optional, it can be missing if the cloud node hasn't started yet
     ?9: authority_access_route, ; optional, it can be missing if the authority hasn't started yet
-    ?10: authority_identity ; optional, hex encoded authority identity
+    ?10: authority_identity, ; optional, hex encoded authority identity
```
shanesveller, line 51: Some currently omitted fields, all optional today:

https://github.com/build-trust/ockam/blob/dac74a08269fbf85eb835ff51a13bb3953487b41/implementations/rust/ockam/ockam_api/src/cloud/project.rs#L54-L69

adrianbenavides: Should we still add them? I feel like we forgot to update the cddl file when they were included. I wonder if the schema tests would fail if they were initialized [here](https://github.com/build-trust/ockam/blob/700528dbd7269224f025541c41fe171b94d7f1b0/implementations/rust/ockam/ockam_api/src/cloud/project.rs#L539).

shanesveller: I've updated the schema and property testing in 7441784cc36ac72ebbf7ac82ef7d8fbfd4bbbdf7...f72fc18d43603a8f7380aeb04d5ecdbb9167b7db, let me know what you think!

adrianbenavides: Awesome, thank you for taking care of that 🙏

adrianbenavides (approved):
🚀

PR #5585, opened 2023-08-11
refactor(rust): use project-id instead of project-name as the key to store projects state

adrianbenavides:
When using the Desktop App, a user will have to access the data from their own projects as well as from other users. This will cause a collition where multiple projects will be named "default".

This PR refactors the project's CliState section so that the project files are stored using the project id as the file name, instead of the project name. 

## WIP status

This change would break the UX of passing project names in the CLI arguments. Should we explore the usage of random-human-readable names? (e.g. https://github.com/allenap/rust-petname or https://github.com/fnichol/names)

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/projects.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/projects.rs
@@ -10,14 +10,17 @@ pub struct ProjectsState {
 
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct ProjectState {
-    name: String,
     path: PathBuf,
     config: ProjectConfig,
 }
 
 impl ProjectState {
+    pub fn id(&self) -> &str {
+        &self.config.id
+    }
+
     pub fn name(&self) -> &str {
-        &self.name
+        &self.config.name
     }
 }
 
@@ -54,6 +57,7 @@ mod traits {
     use crate::cli_state::file_stem;
     use crate::cli_state::traits::*;
     use ockam_core::async_trait;
+    use std::path::Path;
 
     #[async_trait]
     impl StateDirTrait for ProjectsState {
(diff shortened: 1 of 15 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/projects.rs
```diff
@@ -69,6 +73,27 @@ mod traits {
         fn dir(&self) -> &PathBuf {
             &self.dir
         }
+
+        async fn migrate(&self, path: &Path) -> Result<()> {
+            // Rename the project file using the project id
+            let project = match ProjectState::load(path.to_path_buf()) {
```
shanesveller, line 79: If we can use let-else, it may be a very minor improvement here?

implementations/rust/ockam/ockam_api/src/cli_state/projects.rs
```diff
@@ -98,3 +121,34 @@ mod traits {
...
+            name: project_name,
+            id: random_string(),
+            ..Default::default()
+        };
+        let project = ProjectState::new(project_path.clone(), project_config).unwrap();
+        assert!(project_path.exists());
+
+        // Run migration
+        let projects_state = ProjectsState::new(tmp_dir.path().to_path_buf());
+        projects_state.migrate(&project_path).await.unwrap();
```
shanesveller, line 148: Should this have one more assertion that the original path no longer exists?

shanesveller (commented):
I think I appreciate and agree with the motive for the change and the technical implementation of the switch here makes sense to me in that context. Let me know if you'd like me to take another look.

adrianbenavides:
From our latest discussions, it seems that we can avoid doing this change for now, so I'm going to close it until we are sure we want to pursue this. I still think this is doable, and we can workaround the UX issue I mentioned in the description using symbolic links for example, but we could introduce bugs and break examples/docs if we rush it.

PR #5586, opened 2023-08-11
feat(rust): optionally share service when creating the tcp-outlet

adrianbenavides:
Add a new field to the "Share a service" form to optionally send the invitation to the given email. I've also included some modifications to make the font smaller to match the system sizes.

image

### DIFF
--- a/implementations/rust/ockam/ockam_app/src/shared_service/tcp_outlet/create.rs
+++ b/implementations/rust/ockam/ockam_app/src/shared_service/tcp_outlet/create.rs
@@ -6,6 +6,8 @@ use tracing::{debug, error, info};
 
 use crate::app::AppState;
 use crate::error::Error;
+use crate::invitations::build_args_for_create_service_invitation;
+use crate::invitations::commands::create_service_invitation;
 
 /// Create a TCP outlet within the default node.
 #[tauri::command]
@@ -13,8 +15,10 @@ pub async fn tcp_outlet_create(
     app: AppHandle<Wry>,
     service: String,
     port: String,
+    email: String,
 ) -> Result<(), String> {
-    tcp_outlet_create_impl(app, service, port)
+    let email = if email.is_empty() { None } else { Some(email) };
+    tcp_outlet_create_impl(app, service, port, email)
         .await
         .map_err(|e| {
             error!("{:?}", e);
@@ -27,6 +31,7 @@ async fn tcp_outlet_create_impl(
     app: AppHandle<Wry>,
     service: String,
     port: String,
+    email: Option<String>,
 ) -> crate::Result<()> {
     debug!(%service, %port, "Creating an outlet");
     let app_state = app.state::<AppState>();
@@ -54,5 +59,13 @@ async fn tcp_outlet_create_impl(
             Ok(())
         }
         Err(_) => Err(Error::Generic("Failed to create outlet".to_string())),
+    }?;
+    if let Some(email) = email {
+        let invite_args =
+            build_args_for_create_service_invitation(&app, &tcp_addr.to_string(), &email).await?;
+        create_service_invitation(invite_args, app)
+            .await
+            .map_err(Error::Generic)?;
     }
+    Ok(())
 }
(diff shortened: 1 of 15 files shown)


implementations/rust/ockam/ockam_app/src/shared_service/tcp_outlet/create.rs
```diff
@@ -6,15 +6,19 @@ use tracing::{debug, error, info};
...
 /// Create a TCP outlet within the default node.
 #[tauri::command]
 pub async fn tcp_outlet_create(
     app: AppHandle<Wry>,
     service: String,
     port: String,
+    email: String,
 ) -> Result<(), String> {
-    tcp_outlet_create_impl(app, service, port)
+    let email = if email.is_empty() { None } else { Some(email) };
```
shanesveller, line 20: 🤔 Given the desired use case, I think we eventually will want to be able to supply 0, 1, or several emails. Comma-separated or maybe newline-separated, for the form input, which might need to change to a textedit.

implementations/rust/ockam/ockam_app/src/shared_service/tray_menu.rs
```diff
@@ -44,7 +44,7 @@ fn shared_service_submenu(outlet: &OutletStatus) -> SystemTraySubmenu {
         // NOTE: Event handler for dynamic ID is defined in crate::invitations::tray_menu module,
         // and reached via crate::app::tray_menu::fallback_for_id
         submenu = submenu.add_item(CustomMenuItem::new(
-            format!("invitation-create-for-{}", outlet.alias),
+            format!("invitation-create-for-{}", outlet.tcp_addr),
```
shanesveller, line 47: Wouldn't a TCP address be not necessarily unique? Can I expose the same TCP address and port as two separate relays with different non-overlapping route paths?

implementations/typescript/ockam/ockam_app/tailwind.config.cjs
```diff
@@ -2,5 +2,13 @@
 export default {
   content: ['./src/**/*.{html,js,svelte,ts}'],
   plugins: [],
+  theme: {
+    fontSize: {
+      sm: ['11px'],
```
shanesveller: I would recommend to generally use `ems`/`rems` units rather than pixels for font sizing.

https://www.w3docs.com/snippets/css/which-is-better-to-use-in-css-px-em-or-rem.html

The upstream font utility classes use `rems`:
https://tailwindcss.com/docs/font-size

I believe most or all of this change would be undone if we decide to stick to `rems`?

shanesveller (commented):
Thanks for this! Had some smaller questions.

implementations/rust/ockam/ockam_app/src/shared_service/tcp_outlet/create.rs
adrianbenavides: Does the API endpoint support multiple recipient emails already?

implementations/rust/ockam/ockam_app/src/shared_service/tray_menu.rs
adrianbenavides: True... perhaps we should add a new ID ourselves for the desktop app specifically. The problem with the `alias` is that is None by default, so we can't use that either.

implementations/typescript/ockam/ockam_app/tailwind.config.cjs
adrianbenavides: I've done the conversion to `rem`. 

> I believe most or all of this change would be undone if we decide to stick to rems?

What do you mean by this? Changing the root font size directly?

implementations/rust/ockam/ockam_app/src/shared_service/tcp_outlet/create.rs
shanesveller: Today, not as a single request - that would have to be multiple backend requests for each, though they could fire in parallel. If we did this often a bulk endpoint would probably make sense.

implementations/typescript/ockam/ockam_app/tailwind.config.cjs
shanesveller: The content here in `tailwind.config.cjs`, if rewritten in terms of `rems`, might be duplicating the original definitions, unless the specific numbers differ. For the ones that are identical to the originals, we could omit them here.

adrianbenavides: I've updated the `tailwind.config.cjs` file with rem values.

adrianbenavides:
@shanesveller I'm going to create two separate issues to tackle the last two comments, as they are not blocking the basic functionality. If that's OK with you, feel free to mark them as resolved.

shanesveller (approved):

PR #5587, opened 2023-08-11
feat(rust): added support for ockam:// url scheme in linux desktop app

davide-baldo:
Adding the OPEN_URL event with the Desktop App for Linux.
On Linux, the convention is to simply call the application with the custom URL as the first argument of the process.
It's up to the implementer to “pass” the event to the main process if necessary.
In this PR, rather than rely on ockam protocol, I opted for a Unix socket to decouple the event from the initialization of the node and to avoid exposing an extra attack surface (in the form of a Worker).

The mechanism is Linux-specific, but the event itself should be supported by other platforms as well.


implementations/rust/ockam/ockam_app/src/app/events.rs
```diff
@@ -1 +1,2 @@
 pub const SYSTEM_TRAY_ON_UPDATE: &str = "app/system_tray/on_update";
+pub const OPEN_URL: &str = "app/invite/clicked";
```
shanesveller: We can also consider moving this into `src/invite/events.rs`, where it might want a comment to distinguish that it's not exactly a Tauri event in the same sense as the others. 

Otherwise I think I'd like to see the string content a little more agnostic to what clickable action the app should trigger for the user, is that possible or does it complicate things in an unproductive way?

implementations/rust/ockam/ockam_app/linux-desktop-template
```diff
@@ -0,0 +1,12 @@
...
+{{#if comment}}
+Comment={{comment}}
+{{/if}}
+Exec={{exec}} %U
+Icon={{icon}}
+Name={{name}}
+Terminal=false
+Type=Application
+MimeType=x-scheme-handler/ockam;
+X-KDE-Protocols=ockam;
```
shanesveller: For my own curiosity - are these two sufficient to also cover Gnome/XFCE?

implementations/rust/ockam/ockam_app/linux-desktop-template
```diff
@@ -0,0 +1,12 @@
+[Desktop Entry]
```
shanesveller: Which kind of templating syntax does this mechanism use? My understanding is it's a feature provided by Tauri, so I'm not challenging it, just wanted to identify it.

As a minor nit, I also would personally find this combination of filename and folder location a little bit unclear about its purpose without having read this PR.

implementations/rust/ockam/ockam_app/src/linux_url_plugin.rs
```diff
@@ -0,0 +1,63 @@
...
+                if let Some(url) = std::env::args().nth(1) {
+                    if url.starts_with("ockam:") {
+                        info!("received url: {}", url);
+                        handle.trigger_global(crate::app::events::URL_OPEN, Some(url));
+                    } else {
+                        warn!("ignored argument: {}", url);
+                    }
+                }
+
+                for stream in listener.incoming().flatten() {
```
shanesveller: Should this do any of the following, either eventually or during this pass?
- Spawn a tokio task to handle these asynchronously using `tauri::async_runtime`
- Reuse the buffer for the lifetime of a single connection
- Allow the connection to send multiple sequential commands
- Defend against input that is some valid URL followed by some arbitrary content that may be hostile? (I'm imagining something slightly analogous to SQL injection, if we had some vulnerable code within the event handler(s) that trigger by `...URL_OPEN`.)

If the multiple-commands is an intentional no, then can we deliberately close out the client socket after the `trigger_global`?

implementations/rust/ockam/ockam_app/src/linux_url_plugin.rs
```diff
@@ -0,0 +1,63 @@
...
+use std::time::Duration;
+
+use tauri::{
+    async_runtime::spawn,
+    plugin::{Builder, TauriPlugin},
+    Manager, Runtime,
+};
+use tokio::time::sleep;
+
+pub(crate) const OCKAM_OPEN_URL_SOCK: &str = "/tmp/.ockam-open-url-sock";
```
shanesveller: I wonder if this should use [$XDG_RUNTIME_DIR or $XDG_CONFIG_HOME](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#basics) rather than assuming these:
- The system `TMPDIR` is `/tmp` (that's a symlink on MacOS, so it usually still works in practice)
- The running user definitely has write access to the root of `$TMPDIR`
- No one malicious also has simultaneous write access to `OCKAM_OPEN_URL_SOCK` path while it exists

You've already defended against the third concern though 👍🏻

implementations/rust/ockam/ockam_app/src/app/events.rs
shanesveller: You adjusted this already 👍🏻

implementations/rust/ockam/ockam_app/linux-desktop-template
davide-baldo: Yes, `MimeType=x-scheme-handler/ockam;` is a Freedesktop convention, and Debian by default should have an automatic hook that calls `update-desktop-database`

implementations/rust/ockam/ockam_app/linux-desktop-template
davide-baldo: > Which kind of templating syntax does this mechanism use? 

Actually, I didn't investigate it, I just added upon the existing template


> As a minor nit, I also would personally find this combination of filename and folder location a little bit unclear about its purpose without having read this PR.

What about `packaging/linux/application-description-template`? Do you have some suggestions?

shanesveller: I think your suggestion here is a good improvement 👍🏻

implementations/rust/ockam/ockam_app/src/linux_url_plugin.rs
davide-baldo: > Spawn a tokio task to handle these asynchronously using tauri::async_runtime

Absolutely, this is a blunder on my side! fixing it right away

> Reuse the buffer for the lifetime of a single connection
> Allow the connection to send multiple sequential commands

A connection should be used only once (when the second process starts) to pass the URL and that's it. I don't think there is another reason to use the connection for now.

> If the multiple-commands is an intentional no, then can we deliberately close out the client socket after the trigger_global?

Yes, I can manually call `close` rather than waiting for the object to be dropped to better convey the intention.

implementations/rust/ockam/ockam_app/src/linux_url_plugin.rs
davide-baldo: Agree, seems like `XDG_RUNTIME_DIR` is a better choice overall :+1:

implementations/rust/ockam/ockam_app/src/linux_url_plugin.rs
davide-baldo: ops, `close` doesn't exist, I've added an explicit drop

davide-baldo:
Seems like the tauri has its own implementation of the same mechanism on alpha11, i'll close this and adopt their solution

PR #5588, opened 2023-08-11
feat(elixir): add customizable wrapper around tcp calls

mszpakowski:
## Current behavior

Currently our TCP calls are made directly using transport modules like `:gen_tcp`, `:ranch_tcp` and `:ssl`.

## Proposed changes

We'd like to wrap TCP calls in a customizable wrapper, that would allow for performing actions around TCP calls, for example emitting metrics.

Default wrapper wouldn't perform any actions, and would simply execute a transport module call.

## Checks

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/transport/portal/inlet_worker.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/transport/portal/inlet_worker.ex
@@ -21,8 +21,11 @@ defmodule Ockam.Transport.Portal.InletWorker do
   @impl true
   def setup(options, state) do
     peer_route = options[:peer_route]
+    tcp_wrapper = Keyword.get(options, :tcp_wrapper, Ockam.Transport.TCP.DefaultWrapper)
     Logger.info("Starting inlet worker.  Outlet peer: #{inspect(peer_route)}")
-    {:ok, Map.merge(state, %{peer_route: peer_route, stage: :wait_for_socket})}
+
+    {:ok,
+     Map.merge(state, %{peer_route: peer_route, stage: :wait_for_socket, tcp_wrapper: tcp_wrapper})}
   end
 
   @impl true
@@ -84,11 +87,11 @@ defmodule Ockam.Transport.Portal.InletWorker do
   end
 
   defp handle_protocol_msg(
-         %{stage: :connected, socket: socket} = state,
+         %{socket: socket, stage: :connected, tcp_wrapper: tcp_wrapper} = state,
          {:payload, data},
          _return_route
        ) do
-    :ok = :gen_tcp.send(socket, data)
+    :ok = tcp_wrapper.wrap_tcp_call(:gen_tcp, :send, [socket, data])
     {:ok, state}
   end
 
--- a/implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/default_wrapper.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/default_wrapper.ex
@@ -1,0 +1,11 @@
+defmodule Ockam.Transport.TCP.DefaultWrapper do
+  @moduledoc """
+  Default implementation of a TCP wrapper behaviour.
+  """
+  @behaviour Ockam.Transport.TCP.Wrapper
+
+  @impl true
+  def wrap_tcp_call(transport_module, function, args \\ []) when is_list(args) do
+    apply(transport_module, function, args)
+  end
+end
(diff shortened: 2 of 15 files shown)


implementations/elixir/ockam/ockam/lib/ockam/transport/portal/inlet_worker.ex
```diff
@@ -97,4 +96,10 @@ defmodule Ockam.Transport.Portal.InletWorker do
     :ok = :gen_tcp.close(socket)
     {:stop, :normal, state}
   end
+
+  defp send_over_tcp(data, %{socket: socket, tcp_wrapper: tcp_wrapper}) do
+    tcp_wrapper.wrap_tcp_call(fn ->
+      :gen_tcp.send(socket, data)
+    end)
```
shanesveller: What do you think about receiving this argument in as MFA syntax, as in:
 ```suggestion
    tcp_wrapper.wrap_tcp_call([:gen_tcp, :send, [socket, data])
```
similar to what you'd use with `Kernel.apply/3`? I think this change might skip us from allocating an anonymous-function closure with ever-changing captures, that would need to get garbage collected later.

This has some other tradeoffs as well:
- Either preventing use of an anonymous function, or requiring two function heads in `Ockam.Transport.TCP.DefaultWrapper` to permit both styles
- In cases where consumers of the library use hot-code-reloading, anonymous functions will use whatever was in place when the call-site occurred, but apply-calls would always use the then-latest definition of that module + function when the execution actually happens, which is kind of a weird nuance sometimes.

File: implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/default_wrapper.ex [hunk 2]
```diff
@@ -0,0 +1,11 @@
+defmodule Ockam.Transport.TCP.DefaultWrapper do
+  @moduledoc """
+  Default implementation of a TCP wrapper behaviour.
+  """
+  @behaviour Ockam.Transport.TCP.Wrapper
+
+  @impl true
+  def wrap_tcp_call(fun) do
+    fun.()
+  end
```
shanesveller: If we decided to use the MFA syntax, I think this could literally be a `Kernel.apply` call instead:
```suggestion
  def wrap_tcp_call(module, func, args \\ []) when is_list(args) do
    apply(module, func, args)
  end
```

implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/wrapper.ex
```diff
@@ -0,0 +1,8 @@
+defmodule Ockam.Transport.TCP.Wrapper do
+  @moduledoc """
+  Behaviour definition for a TCP call wrapper.
+  Allows performing additional work around TCP calls, like
+  emitting telemetry events.
+  """
+  @callback wrap_tcp_call(fun :: function()) :: any()
```
shanesveller: 🤔 I wonder if the typespec should express that the return value is always equivalent to that of the inner gen_tcp send, and that the wrapper logic should only impart side effects?

implementations/elixir/ockam/ockam_kafka/lib/services/provider.ex
```diff
@@ -203,11 +206,18 @@ defmodule Ockam.Services.Kafka.Provider do
...
          authorization: authorization
        ]},
-      {Ockam.Kafka.Interceptor.OutletManager, [outlet_prefix, ssl, ssl_options]}
+      {Ockam.Kafka.Interceptor.OutletManager,
+       [
+         outlet_prefix: outlet_prefix,
+         ssl: ssl,
+         ssl_options: ssl_options,
+         tcp_wrapper: tcp_wrapper
+       ]}
```
shanesveller, line 220: Keyword arguments are a welcome improvement 👍🏻

implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/wrapper.ex
mszpakowski: At least for now we want to make it relaxed and allow returning custom values.

implementations/elixir/ockam/ockam/lib/ockam/transport/portal/inlet_worker.ex
mszpakowski: Yeah, that's definitely necessary. I changed it to use `Kernel.apply/3`

shanesveller (approved):

PR #5655, opened 2023-08-17
Upgrade tauri alpha11

davide-baldo:
In this PR i've upgraded `ockam_app` to the latest alpha11, however in both Linux and MacOS I've noticed major instability:
- menu doesn't appear/refresh on linux: https://github.com/tauri-apps/tauri/issues/7648
- double tray icon on MacOs with a segmentation fault on first interaction

It's very possible I messed up something, but the fact that the `api` example has the same behavior makes me thing there is a major issue within `muda`

### DIFF
--- a/implementations/rust/ockam/ockam_app/tauri.conf.json
+++ b/implementations/rust/ockam/ockam_app/tauri.conf.json
@@ -16,7 +16,7 @@
     "version": "0.1.0"
   },
   "tauri": {
-    "systemTray": {
+    "tray-icon": {
       "iconPath": "icons/icon.png",
       "iconAsTemplate": true
     },
@@ -25,7 +25,7 @@
       "category": "DeveloperTool",
       "copyright": "",
       "deb": {
-        "depends": []
+        "desktopTemplate": "packaging/linux/application-description-template"
       },
       "externalBin": [],
       "icon": [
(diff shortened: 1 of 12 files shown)


implementations/rust/ockam/ockam_app/tauri.conf.json
```diff
@@ -25,7 +25,7 @@
       "category": "DeveloperTool",
       "copyright": "",
       "deb": {
-        "depends": []
+        "desktopTemplate": "packaging/linux/application-description-template"
```
shanesveller, line 28: It's almost certainly harmless to include now, but any chance this one snuck in from your other working branch?

shanesveller (commented):
Thank you very much for taking this on 🍻

implementations/rust/ockam/ockam_app/tauri.conf.json
davide-baldo: ops, yep, it snuck in, and it shouldn't be necessary anymore since I'm patching tauri directly

davide-baldo:
On Mac the `api` example is working correctly, so I've probably made a major error

davide-baldo:
Superseded by https://github.com/build-trust/ockam/pull/5913

PR #5656, opened 2023-08-17
feat(rust): create tcp-inlets using accepted invitations

adrianbenavides:


implementations/rust/ockam/ockam_app/src/invitations/commands.rs
```diff
@@ -118,6 +130,165 @@ pub async fn refresh_invitations<R: Runtime>(app: AppHandle<R>) -> Result<(), St
         let mut writer = invitation_state.write().await;
         *writer = invitations.into();
     }
+    refresh_inlets(&app).await.map_err(|e| e.to_string())?;
     app.trigger_global(REFRESHED_INVITATIONS, None);
     Ok(())
 }
+
+async fn refresh_inlets<R: Runtime>(app: &AppHandle<R>) -> crate::Result<()> {
```
shanesveller, line 135: Not intentionally bike-shedding, but does this one make more sense here or in crate::shared_service somewhere? I see that it kind of straddles both concerns in the current iteration.

implementations/rust/ockam/ockam_command/src/run/parser.rs
```diff
@@ -174,8 +174,8 @@ pub struct Config {
 pub struct NodeConfig {
     #[serde(rename(deserialize = "depends-on"))]
     pub depends_on: Option<String>,
-    #[serde(rename(deserialize = "enroll-ticket"))]
-    pub enroll_ticket: Option<String>,
+    #[serde(rename(deserialize = "enrollment-ticket"))]
```
shanesveller, line 178: Does this have ramifications for the BATS suite?

implementations/rust/ockam/ockam_app/src/invitations/commands.rs
```diff
@@ -118,6 +130,165 @@ pub async fn refresh_invitations<R: Runtime>(app: AppHandle<R>) -> Result<(), St
...
+        };
+        // Iterate the invitations and create the inlets if they don't exist
+        for invitation_inlet_data in invitations {
+            if let Some(inlet) = node_inlets
+                .iter()
+                // The inlets we create are named after the remote service name (i.e. the remote tcp-outlet name)
+                .find(|inlet| inlet.alias == invitation_inlet_data.service_name)
+            {
+                // If the inlet already exists, mark it to remove it from the node
+                node_inlets_to_delete.push(inlet);
```
shanesveller: Does this need one more step or check to ensure that inlets that do already exist and that should still exist afterwards don't get removed by mistake? This seems like it would remove existing ones unconditionally.

adrianbenavides: I've simplified that part. With the naming strategy for the nodes, there is a 1:1 relationship between nodes and tcp-inlets, so the HashMap disapears. I've simplified that part [here](https://github.com/build-trust/ockam/pull/5656/files#diff-fc44c8a5e4b38ca36b52578860084fa796e4931a7b0efdbeb3396c6d36e5ada5R146).

implementations/rust/ockam/ockam_app/src/invitations/commands.rs
adrianbenavides: Not sure. If we only look at privacy/visibility rules, these new functions are only called from this module, so if I had to choose between the two modules, I guess the `invitations` module makes more sense.

implementations/rust/ockam/ockam_command/src/run/parser.rs
adrianbenavides: I don't think so, but I'll double check it.

shanesveller: If it looks clear to you, I'll hit resolve. CI should catch it if not 🤞🏻

implementations/rust/ockam/ockam_app/src/invitations/commands.rs
shanesveller: That looks clearer to me, thank you.

shanesveller (approved):

Issue #5580, opened 2023-08-10
Support custom URI handler for clickable links in Desktop app in MacOS


davide-baldo:
Blocking: https://github.com/build-trust/ockam/issues/5577

In MacOs it's pretty easy to declare the support for custom URLs, just add this snipped to the `pInfo` file of the application. However, there is non trivial to intercept the event, as there is a [multi-year ongoing discussion](https://github.com/rust-windowing/winit/issues/2120) about these events, maybe by mapping the specific event rather than allowing any arbitrary event would be easier.

```
<key>CFBundleURLTypes</key>
<array>
        <dict>
                <key>CFBundleTypeRole</key>
                <string>Editor</string>
                <key>CFBundleURLName</key>
                <string>io.ockam</string>
                <key>CFBundleURLSchemes</key>
                <array>
                        <string>ockam</string>
                        <string>ockam-invite</string>
                </array>
        </dict>
</array>
```
And register the App as handler with:
`/System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/LaunchServices.framework/Versions/A/Support/lsregister -f /Applications/Ockam\ Desktop.app/`

And confirm the registration with:
`/System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/LaunchServices.framework/Versions/A/Support/lsregister -dump URLSchemeBinding  | grep ockam`

Issue #5579, opened 2023-08-10
Support custom URI handler for clickable links in Desktop app in Linux


davide-baldo:
Blocking: https://github.com/build-trust/ockam/issues/5577

PR #5663, opened 2023-08-18
added ockam:// scheme association in app for linux and mac

davide-baldo:
This pr is based on alpha11 upgrade branch.

Added support for both linux and mac for the `ockam://` custom scheme.
Since tauri doesn't handle linux url passing I've recovered work [from previous pr](https://github.com/build-trust/ockam/pull/5587).
Both `.desktop` file and `pList` respectively for linux and mac, are now directly generated by tauri.

Note that to apply a patch, tauri **has been overridden to https://github.com/build-trust/tauri.git**
In order to start `cargo tauri build` the tauri-cli must be compiled and installed from the fork, for this reason the cli has been overridden in the build system. The docker build image wasn't changed because gradle build is deprecated and will be removed soon.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,3 +49,9 @@ inherits = "test"
 opt-level = 2
 [profile.dev.package.adler]
 opt-level = 1
+
+[patch.crates-io]
+# temporary fork for url scheme - https://github.com/build-trust/ockam/issues/5667
+tauri = { git = "https://github.com/build-trust/tauri.git", branch = "add-url-scheme" }
+tauri-build = { git = "https://github.com/build-trust/tauri.git", branch = "add-url-scheme" }
+tauri-utils = { git = "https://github.com/build-trust/tauri.git", branch = "add-url-scheme" }
--- a/.github/actions/build_binaries/action.yml
+++ b/.github/actions/build_binaries/action.yml
@@ -88,7 +88,7 @@ runs:
       if: inputs.build_app == 'true'
       run: |
         set -x
-        cargo install --locked tauri-cli --version 2.0.0-alpha.10
+        cargo install tauri-cli --git https://github.com/build-trust/tauri.git --branch add-url-scheme
 
         cd implementations/rust/ockam/ockam_app/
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -49,3 +49,7 @@ inherits = "test"
 opt-level = 2
 [profile.dev.package.adler]
 opt-level = 1
+
+[patch.crates-io]
+tauri = { git = "https://github.com/davide-baldo/tauri.git", branch="add-url-scheme" }
+tauri-build = { git = "https://github.com/davide-baldo/tauri.git", branch="add-url-scheme" }
```
shanesveller: Could we please fork this into the build-trust GitHub org if possible? And mark a tracking issue that we can comment above this, so we'll know in the future when it's safe to remove again?

davide-baldo: makes perfect sense, done: 
https://github.com/build-trust/ockam/issues/5667
https://github.com/build-trust/tauri

davide-baldo:
solves https://github.com/build-trust/ockam/issues/5579 and https://github.com/build-trust/ockam/issues/5580

implementations/rust/build.gradle
```diff
@@ -141,8 +141,17 @@ task build_docs {
   }
 }
 
+task tauri_cli {
+  doLast {
+    exec {
+      environment environmentVars()
+      commandLine cargo('install', 'tauri-cli', '--git', 'https://github.com/build-trust/tauri.git', '--branch', 'add-url-scheme', '--force')
```
metaclips: Can we move these commands here https://github.com/build-trust/ockam/blob/develop/implementations/rust/Makefile, we plan on deprecating gradle

metaclips (changes requested):

implementations/rust/build.gradle
davide-baldo: The command is also inside nix environments. Is makefile always invoked from a nix environment during CI?

metaclips (approved):

PR #5672, opened 2023-08-18
feat(rust): specify enrollment ticket ttl on desktop app

polvorin:
set to to expire in 30 days.   Mostly as a workaround,  a better approach is likely needed to either
* don't use enrollment ticket and rely on email verification
* use enrollment ticket,  but with unlimited use (instead of one-time),  and likely unlimited ttl as well?,  with it managed by the inviter.

This need a bit more tough,  the extension to 30 days allows for using the feature in the general case meantime


mrinalwadhwa:
This will need an update if https://github.com/build-trust/ockam/pull/5677 lands before it.

etorreborre:
@polvorin https://github.com/build-trust/ockam/pull/5677 is now ready so you can base your PR on the new `expires-in` argument

PR #5676, opened 2023-08-19
wait_for_email_verification testing and http mock

talir0drigues:
## Current behavior
We dont have a unit test for `wait_for_email_verification`, and for some other functions on OidcService. The reason for that are the http calls,  to test it either we would use a mock library or change the code for implementing each function separately. 

## Proposed changes

We create the unit test for `wait_for_email_verification`. However, to achieve that we currently modify the code where we extract the function that performs the http call to a standalone implementation for OidcService (line 50). 
This is not the ideal scenario, since to test any other function that depends on http call, we would need to restructure the code. 
As a next step I'm trying to isolate the http calls to another implementation, so we can use something like [mockall](https://lib.rs/crates/mockall) in the future. 

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
+++ b/implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
@@ -1,8 +1,8 @@
+use async_trait::async_trait;
 use std::borrow::Borrow;
 use std::io::stdin;
 use std::str::FromStr;
 use std::sync::Arc;
-use async_trait::async_trait;
 
 use colorful::Colorful;
 use miette::{miette, IntoDiagnostic};
@@ -42,12 +42,12 @@ impl Default for OidcService {
 }
 
 #[async_trait]
-pub trait GetUserInfo{
+pub trait GetUserInfo {
     async fn get_user_info(&self, token: &OidcToken) -> Result<UserInfo>;
 }
 
 #[async_trait]
-impl GetUserInfo for OidcService{
+impl GetUserInfo for OidcService {
     /// Return the information about a user once authenticated
     async fn get_user_info(&self, token: &OidcToken) -> Result<UserInfo> {
         let client = self.provider().build_http_client()?;
@@ -65,7 +65,7 @@ impl GetUserInfo for OidcService{
     }
 }
 
-pub async fn wait_for_email_verification <T:GetUserInfo> (
+pub async fn wait_for_email_verification<T: GetUserInfo>(
     client: T,
     token: &OidcToken,
     opts: &CommandGlobalOpts,
@@ -185,7 +185,6 @@ impl OidcService {
             .into_diagnostic()?;
         res.json().await.map_err(|e| miette!(e).into())
     }
-
 
     pub(crate) async fn validate_provider_config(&self) -> miette::Result<()> {
         if let Err(e) = self.device_code().await {
@@ -503,16 +502,13 @@ impl OidcService {
     }
 }
 
-
-
-
 #[cfg(test)]
 mod tests {
-    use std::sync::Mutex;
+    use super::*;
+    use crate::GlobalArgs;
     use ockam_api::cloud::enroll::Token;
     use ockam_node::Executor;
-    use crate::GlobalArgs;
-    use super::*;
+    use std::sync::Mutex;
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     #[ignore = "this test can only run with an open browser in order to authenticate the user"]
@@ -584,23 +580,22 @@ mod tests {
     enum ClientState {
         FirstCall,
         SecondCall,
-        Finished
+        Finished,
     }
 
     struct Client {
-        state: Arc<Mutex<ClientState>>
+        state: Arc<Mutex<ClientState>>,
     }
 
     #[async_trait]
-    impl GetUserInfo for Client{
+    impl GetUserInfo for Client {
         async fn get_user_info(&self, _token: &OidcToken) -> Result<UserInfo> {
-
-                let mut guard = self.state.lock().unwrap();
+            let mut guard = self.state.lock().unwrap();
 
             match *guard {
                 ClientState::FirstCall => {
-                     *guard = ClientState::SecondCall;
-                    return Ok(UserInfo{
+                    *guard = ClientState::SecondCall;
+                    return Ok(UserInfo {
                         sub: "".to_string(),
                         nickname: "bad_nickname".to_string(),
                         name: "".to_string(),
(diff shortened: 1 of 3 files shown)


implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
```diff
@@ -520,4 +580,66 @@ mod tests {
...
+
+    struct Client {
+        state: Arc<Mutex<ClientState>>
+    }
+
+    #[async_trait]
+    impl GetUserInfo for Client{
+        async fn get_user_info(&self, _token: &OidcToken) -> Result<UserInfo> {
+
+                let mut guard = self.state.lock().unwrap();
```
etorreborre: @talir0drigues you're going to need to format your code with `cargo fmt` to pass the CI checks.

implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
```diff
@@ -520,4 +580,66 @@ mod tests {
         assert!(code.is_ok());
         assert_eq!(code.unwrap(), "12345".to_string())
     }
+
+    enum ClientState {
+        FirstCall,
```
etorreborre, line 605: In terms of code layout I would rather see all the tests, `test_user_info`, `test_wait_for_email_verification`,... then the support code

implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
```diff
@@ -520,4 +580,66 @@ mod tests {
...
+            }
+        }
+    }
+
+    #[test] fn test_wait_for_email_verification() -> Result<()>{
+        let opts = CommandGlobalOpts::new(GlobalArgs::default()).set_quiet();
+        let authorization_code = OidcToken{ token_type: TokenType::Bearer, access_token: Token("".to_string())};
+
+        let result = Executor::execute_future( async move {
+            wait_for_email_verification(Client { state: Arc::new(Mutex::new(ClientState::FirstCall ))}, &authorization_code, &opts).await
```
etorreborre: You could implement a separate `Client::new()` method to reduce that code here.

implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
```diff
@@ -520,4 +580,66 @@ mod tests {
...
+                        email_verified: true,
+                    })
+                }
+
+                ClientState::Finished => panic!("an extra call!")
+            }
+        }
+    }
+
+    #[test] fn test_wait_for_email_verification() -> Result<()>{
```
etorreborre: Why not use `#[tokio::test]` here and make this function an `async` function? Then you don't need to use the `Executor`

implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
```diff
@@ -520,4 +580,66 @@ mod tests {
...
+    #[async_trait]
+    impl GetUserInfo for Client{
+        async fn get_user_info(&self, _token: &OidcToken) -> Result<UserInfo> {
+
+                let mut guard = self.state.lock().unwrap();
+
+            match *guard {
+                ClientState::FirstCall => {
+                     *guard = ClientState::SecondCall;
+                    return Ok(UserInfo{
```
etorreborre: `UserInfo` could be factored out to save a bit of vertical space and just show which field is relevant: 
```rust
let mut user_info = UserInfo {.....};
...
Ok(user_info),
...
user_info.email_verified = true;
Ok(user_info)
```

etorreborre (commented):
I left a few comments for refactoring opportunities but I think that before writing those tests we should rethink the design instead: 

*  instead of building a `reqwest` client and building / sending queries wrapped with exponential backoffs and retries we could create our own `HttpClient` supporting a standard retry strategy (integrating `reqwest` and `tokio_retry`). This way we will always benefit from retries when doing http calls which should always be considered as flaky

* we should revisit the interface for `OidcProvider`. At the moment it is actually only a list of configuration parameters. We could instead equip it with methods showing its real responsibilities
    * `get_user_info`
    * `get_device_code`
    * `get_authorization_token`
    * `poll_oidc_token`
    * `get_redirect_url` (necessary to start the local server)
    * the redirect timeout goes within the `OidcService`

 * this would encapsulate most of the low-level details of an `OidcProvider` like `client_id`, `device_code_url` etc... and would let us stub this interface to the fully test the `OidcService`
    
 * then the responsibility of the `OidcService` is merely to orchestrate the calls to the `OidcProvider` and provide user information for progress and errors

What do you think?

implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
talir0drigues: makes sense! In rust is it common to create a aux file for support code?

implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
talir0drigues: much better!

implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
talir0drigues: so, in these cases a constructur is better even if we will use it for one test instead of longer lines of code?

implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
etorreborre: I don't know if there's a general best practice. I generally push this code down with a comment like `// TEST HELPERS` or similar

implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
etorreborre: My preference is to make the body of the test as obvious as possible and move implementation details elsewhere. So in that case I would create a single-use constructor, yes.

implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
talir0drigues: that's basically happens because when we create the ops `let opts = CommandGlobalOpts::new(GlobalArgs::default()).set_quiet();` it calls a [executor](https://github.com/build-trust/ockam/blob/772eb8cb6aafbb46e3be0403a3cf22e2ff015993/implementations/rust/ockam/ockam_api/src/cli_state.rs#L109). So if we make the function async, it throws a error saying we cannot create a runtime within another runtime.

talir0drigues:
> 
I agree! While we were creating this test the question on why the responsibilities weren't separated arouse. Having a httpclient would help. 
How do you think we can proceed? I close this PR and open a task to refactor `OidcProvider` and `OidcService` ?

implementations/rust/ockam/ockam_command/src/enroll/oidc_service.rs
etorreborre: Right, I've had this kind of issue in the past :-). One solution is to just pass a `Terminal` to `wait_for_email_verification` since that function does not need a full set of global options.

etorreborre:
> How do you think we can proceed? I close this PR and open a task to refactor OidcProvider and OidcService ?

Yes I think that would be the right thing to do now.

PR #5681, opened 2023-08-21
feat(rust): disable `backtrace` for errors by default

SanjoDeundiak:
Disable backtraces by default

### DIFF
--- a/implementations/rust/ockam/ockam_core/src/error/inner/mod.rs
+++ b/implementations/rust/ockam/ockam_core/src/error/inner/mod.rs
@@ -79,12 +79,17 @@ impl ErrorData {
         #[allow(unused_mut)]
         let mut local: Vec<LocalPayloadEntry> = vec![];
 
-        #[cfg(all(feature = "std", feature = "error-traces"))]
+        #[cfg(all(feature = "std", feature = "backtrace"))]
         if trace_config::BACKTRACE_ENABLED.get() {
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_core/src/error/inner/mod.rs
```diff
@@ -79,12 +79,17 @@ impl ErrorData {
         #[allow(unused_mut)]
         let mut local: Vec<LocalPayloadEntry> = vec![];
 
-        #[cfg(all(feature = "std", feature = "error-traces"))]
+        #[cfg(all(feature = "std", feature = "backtrace"))]
         if trace_config::BACKTRACE_ENABLED.get() {
+            // This is called everytime an Error instance is created and takes up to 300ms
+            // on a fast CPU, which is way slower than expected from such type of operation.
+            // Therefore, this feature is disabled by default, but can be used.
+            // It's also possible to replace `Backtrace::new()` with `Backtrace::new_unresolved()`
```
etorreborre, line 87: That's a good point. Why not do `new_unresolved` by default and resolve when we want to log the error?

SanjoDeundiak: I already spent a lot of unexpected time on this issue and didn't have more time to investigate how much faster it is, what the output difference is, what else should be done, and so on.

etorreborre (approved):

PR #5682, opened 2023-08-21
fix(rust): on `project enroll` do not exit early if a trust context with same name exists

adrianbenavides:
- on `project enroll` do not exit early if a trust context with same name exists
- remove local redundant trust-context test
- unset OCKAM_LOG in bats tests


SanjoDeundiak (approved):
Tnx for fixing that

PR #5709, opened 2023-08-22
feat(rust): on desktop enroll, wait until project is ready

adrianbenavides:
## Current

When the user runs the application with no projects created previously, the app crashes.

## Proposed

Wait until the project is fully created, as we do in the CLI.


polvorin (approved):
that  wait_until_project_is_ready()  function looks like something we have already in some other place?  (like the command app when enrolling?)
would it be possible to reuse?

etorreborre:
@polvorin this is could be done if we had a way to replace the terminal messages by log messages. That's definitely something to investigate if we want to reduce the maintenance between the command line and the desktop application

etorreborre (approved):
👍 I tested it locally, it worked fine for me

PR #5712, opened 2023-08-22
chore(rust): remove an unnecessary security advisory

etorreborre:
The rustsec website says that the advisory has been withdrawn now: https://rustsec.org/advisories/RUSTSEC-2022-0048.html. In any case we don't have a dependency on `xml-rs`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5828,9 +5828,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-webpki"
-version = "0.101.2"
+version = "0.101.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


shanesveller:
⚠️  `xml-rs` is still present as of b370b26efac5696a4e1547415cc594baef274d30:
```shell
cargo tree -i -p xml-rs --target all
xml-rs v0.8.16
└── serde-xml-rs v0.6.0
    └── bluez-async v0.7.2
        └── btleplug v0.11.0
            └── ockam_transport_ble v0.45.0 (.../ockam/ockam/implementations/rust/ockam/ockam_transport_ble)
```

etorreborre:
Correct I missed the `--target all` :-). That advisory has been removed so we should be good to go anyway.
That being said there is another issue with the `rustls-webpki` dependency, I'm going to upgrade it to fix it.

shanesveller (approved):

PR #5713, opened 2023-08-22
feat(rust): ignore expired accepted invitations

adrianbenavides:
Accepted invitations are filtered out on the backend, but the user could try to use an invitation right before the poll event that filters that invitation.

In order to give the user a specific error message, this PR's checks the `expires_at` field before contacting the backend.

### DIFF
--- a/implementations/rust/ockam/ockam_app/src/invitations/commands.rs
+++ b/implementations/rust/ockam/ockam_app/src/invitations/commands.rs
@@ -1,4 +1,6 @@
 use tauri::{AppHandle, Manager, Runtime, State};
+use time::format_description::well_known::iso8601::Iso8601;
+use time::OffsetDateTime;
 use tracing::{debug, error, info, warn};
 
 use ockam_api::cli_state::{CliState, StateDirTrait};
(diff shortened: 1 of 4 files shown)


implementations/rust/ockam/ockam_app/src/invitations/commands.rs
```diff
@@ -223,6 +229,13 @@ impl InletDataFromInvitation {
         cli_state: &CliState,
         invitation: &InvitationWithAccess,
     ) -> crate::Result<Option<Self>> {
+        let expires_at =
+            OffsetDateTime::parse(&invitation.invitation.expires_at, &Iso8601::DEFAULT).unwrap();
+        if expires_at > OffsetDateTime::now_utc() {
```
shanesveller:
```suggestion
        if expires_at < OffsetDateTime::now_utc() {
```

adrianbenavides: Wow, I don't know how to think in dates anymore 😆

shanesveller (approved):

PR #5718, opened 2023-08-22
fix: Filter projects for admin role during app interactions

### LINKED ISSUE #5704: The correct default project must be retrieved in the desktop application
When the desktop application starts up it retrieves the first project that the current user, alice, can access, based on the name `default`. This project might have been created by another user, bob who invited alice in order to share a service (with a role = Service).

Alice will then not be able to create enrollment tickets with that project since she is not an administrator of bob's project.

The fix consists in [filtering the list of retrieved projects](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs#L127) based on the `UserInfo` retrieved when the user authenticates:
```rust
// get the user email from the application state
let user_email = app_state
        .model(|m| m.get_user_info())
        .await
        .map(|u| u.email)
        .unwrap_or_default();

// only keep the projects where the user is an Admin
let project = match projects
        .iter()
        .filter(|p| {
            p.user_roles
                .iter()
                .any(|r| r.role == RoleInShare::Admin && r.email == user_email)
        })
        .find(|p| p.name == *PROJECT_NAME)
    {
....
```

Note: the code above can be refactored with  a `Project::is_admin(user_info: UserInfo) -> bool` function.

shanesveller:
## Current behavior

Orchestrator users may be members of multiple Spaces, multiple Projects, and hold varying roles within each. In order to share an invitation for another user to connect through a project, the inviter must hold an Admin role in either the parent Space or in the Project itself.

## Proposed changes

Pre-filter the list of projects for Admin roles when choosing a Project for an app interaction, such as creating an Enrollment Ticket or service invitation. This will prevent most classes of backend 401 under the existing logic, where the inference may select a Project you have no rights to.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cloud/mod.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/mod.rs
@@ -9,7 +9,7 @@ use ockam_multiaddr::MultiAddr;
 
 use crate::error::ApiError;
 
-use self::share::RoleInShare;
+use self::share::{RoleInShare, ShareScope};
 
 pub mod addon;
 pub mod enroll;
@@ -42,6 +42,7 @@ pub struct ProjectUserRole {
     #[n(1)] pub email: String,
     #[n(2)] pub id: usize,
     #[n(3)] pub role: RoleInShare,
+    #[n(4)] pub scope: ShareScope,
 }
 
 /// A wrapper around a cloud request with extra fields.
--- a/implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
+++ b/implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
@@ -124,7 +124,14 @@ async fn retrieve_project(app_state: &AppState, space: &Space) -> Result<Project
             .await
             .map_err(|e| miette!(e))?
     };
-    let project = match projects.iter().find(|p| p.name == *PROJECT_NAME) {
+
+    let email = app_state.user_email().await.unwrap_or_default();
+
+    let project = match projects
+        .iter()
+        .filter(|p| p.has_admin_with_email(&email))
+        .find(|p| p.name == *PROJECT_NAME)
+    {
         Some(project) => project.clone(),
         None => {
             let ctx = &app_state.context();
--- a/implementations/rust/ockam/ockam_api/src/cloud/project.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/project.rs
@@ -14,6 +14,7 @@ use crate::cloud::addon::ConfluentConfigResponse;
 use crate::error::ApiError;
 use crate::minicbor_url::Url;
 
+use super::share::RoleInShare;
 use super::ProjectUserRole;
 
 #[derive(Encode, Decode, Serialize, Deserialize, Debug, Default, Clone, Eq, PartialEq)]
@@ -73,11 +74,14 @@ pub struct Project {
 }
 
 impl Project {
-    pub fn is_ready(&self) -> bool {
-        !(self.access_route.is_empty()
-            || self.authority_access_route.is_none()
-            || self.identity.is_none()
-            || self.authority_identity.is_none())
+    pub fn access_route(&self) -> Result<MultiAddr> {
+        MultiAddr::from_str(&self.access_route).map_err(|e| ApiError::generic(&e.to_string()))
+    }
+
+    pub fn has_admin_with_email(&self, email: &str) -> bool {
+        self.user_roles
+            .iter()
+            .any(|ur| ur.role == RoleInShare::Admin && ur.email == email)
     }
 
     pub async fn is_reachable(&self) -> Result<bool> {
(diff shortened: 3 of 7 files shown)


implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
```diff
@@ -124,7 +125,14 @@ async fn retrieve_project(app_state: &AppState, space: &Space) -> Result<Project
             .await
             .map_err(|e| miette!(e))?
     };
-    let project = match projects.iter().find(|p| p.name == *PROJECT_NAME) {
+
+    let email = app_state.user_email().await.unwrap_or_default();
+
+    let project = match projects
```
etorreborre, line 130: Can this be refactored with [this part](https://github.com/build-trust/ockam/pull/5718/files#diff-fc44c8a5e4b38ca36b52578860084fa796e4931a7b0efdbeb3396c6d36e5ada5R65) as one `get_user_project` command?

implementations/rust/ockam/ockam_app/src/projects/mod.rs
```diff
@@ -1,8 +1,24 @@
...
 
 pub(crate) mod commands;
 mod error;
 pub(crate) mod events;
 pub(super) mod plugin;
 
 pub(crate) type State = Vec<Project>;
+pub(crate) type SyncState = Arc<RwLock<State>>;
+
+pub(crate) trait ProjectExt {
```
etorreborre: I would put this code alongside the `Project` code since it is entirely the same domain of project / users / roles.

shanesveller: Makes sense, I moved it, currently visible in 2bdd06efa13cfe8fbb1271527b9ab2b5a5d62338.

implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
shanesveller: They're not trivially identical, since one of them wants to find by ID and one wants to find by name. I took a shot at consolidating the shared concerns with 08c4d617acebdf096b0f57209c53194f20516deb...17177e61eb8680bf5a07c7b6fbfebdcc211b7efc.

etorreborre: Thanks

implementations/rust/ockam/ockam_api/src/cloud/mod.rs
```diff
@@ -42,6 +42,7 @@ pub struct ProjectUserRole {
     #[n(1)] pub email: String,
     #[n(2)] pub id: usize,
     #[n(3)] pub role: RoleInShare,
+    #[n(4)] pub scope: ShareScope,
```
etorreborre, line 45: I don't see this field being used anywhere in the Rust code. Is it supposed to be here?

etorreborre (approved):
Thanks for this fix. The only remaining question I have is about the `ShareScope` in `ProjectUserRole`. What is it used for? Is it set by the Orchestrator?

implementations/rust/ockam/ockam_api/src/cloud/mod.rs
shanesveller: > What is it used for? Is it set by the Orchestrator?

That's correct, and the general purpose is the same as when it was introduced during #5451. Users may hold non-overlapping roles in a Project and the Space that Project belongs to, so this can be used for client-side logic to distinguish when needed. For example, a user can be a Guest in a Space and an Admin in a Project within that Space. In that kind of combination, without this disambiguating field, the `user_roles` list would appear like:
```json
[
  {"email": "user@example.com", "id": 12345, "role": "Guest" },
  {"email": "user@example.com", "id": 12345, "role": "Admin" }
]
```
After:
```json
[
  {"email": "user@example.com", "id": 12345, "role": "Guest", "scope": "Space" },
  {"email": "user@example.com", "id": 12345, "role": "Admin", "scope": "Project" }
]
```

etorreborre: Ok. Given the name `ProjectUserRole` I was expecting that the `scope` was already `Project` and that the query for getting the `user_roles` was already specifying for which project. Isn't there a bit of mismatch in names then?

shanesveller: > Given the name ProjectUserRole I was expecting that the scope was already Project and that the query for getting the user_roles was already specifying for which project.

Perhaps renaming `scope` to something reflecting "origin/source of this granted role" would've reduced or avoided your partially-incorrect assumption here. The `user_roles` list is indeed specific to the individual project that appears in. Read on if you want more color.

---

Today, most users with access to a Project, especially in the relatively common case where they are newly enrolled are accessing the Project as an Admin of the Space the project belongs to rather than having direct rights attached to the Project itself. When more specific information is not present, a user's effective Project rights are inherited from their rights on the parent Space.

The backend previously did not report the access levels that are conveyed from the parent Space in this response payload, only the ones that are directly attached to the Project. Taken together with the two points above, this meant that the `user_roles` list was nearly always empty (and thus unhelpful) in the majority of scenarios. This was recently changed on the backend for more complete information.

The reason any of this matters is because the existing client behavior often relies far too much on underspecified inference, which was often developed without input or feedback from team members who are intimately familiar with the backend implementation details.

TLDR - When creating invitations, or anywhere else we'll need to display candidates for a user to take similar elevated actions on, we actually only care about identifying projects where the user is an acting Admin, regardless of where those rights are specifically granted.

PR #5723, opened 2023-08-23
chore: use a different name for the desktop application executable

### LINKED ISSUE #5703: There must be no executable names conflicts between the command line application and the desktop application
### Observed behavior

When we currently build the command line application with `cargo build` we end up with an executable named `ockam` in `target/release/debug`.

Similarly when we build the desktop application with `cd implementations/rust/ockam/ockam_app; cargo tauri dev` we also get an executable named `ockam` in `target/release/debug`.

Since the desktop application expects a command-line executable named `ockam` to be on the `PATH` if we run `cargo build` then `cargo tauri dev` then we end up with the desktop application calling itself.

### Steps to reproduce

1. run `cargo build`
2. run `cd implementations/rust/ockam/ockam_app; cargo tauri dev`
3. try to create a shared service. The logs should show that the `ticket` option does not exist for the `ockam` executable

### Expected behavior

One immediate solution would be to rename `productName: "Ockam"` to `productName: "OckamDesktop"` in the `tauri.conf` file. The drawback is that we lose the short and sweet "Ockam" name as a bundled application.

Other alternatives involve:

 - using `cargo install` to a specific path and set the `OCKAM` env. variable accordingly
 - finding a `tauri` option, if it exists, to keep the `Ockam` bundle name but be able to specify another executable name inside the bundle

### Ockam Version

0.90.0

etorreborre:
This will avoid conflicts with the ockam command line executable.

Fixes #5703.

### DIFF
--- a/implementations/rust/ockam/ockam_app/tauri.conf.json
+++ b/implementations/rust/ockam/ockam_app/tauri.conf.json
@@ -12,7 +12,7 @@
     "distDir": "../../../typescript/ockam/ockam_app/build"
   },
   "package": {
-    "productName": "Ockam",
+    "productName": "OckamDesktop",
     "version": "0.1.0"
   },
   "tauri": {


adrianbenavides (approved):
🚀

PR #5725, opened 2023-08-23
#5707 make sure that we can execute the ockam command from the desktop application

### LINKED ISSUE #5707: The `OCKAM` environment variable needs to be a fully-qualified path when using the desktop app
# Observed behavior

If the `OCKAM` environment variable is not a full path pointing to the command line executable then the desktop application will not find the executable to issue commands such as `ockam project ticket`.

# Steps to reproduce

Set `OCKAM=target/debug/ockam` and start the desktop application with `cd implementations/rust/ockam/ockam_app; cargo tauri dev`.

# Expected behavior

There should be a check at the start of the desktop app to verify if the `ockam` executable can be found. If this not the case, return an error message to the user and stop the application

etorreborre:
## Description

This PR checks the availability of the `ockam` command (used to start inlets) and exits the desktop application if the command can not be executed.

## Tests

It can be tested by:

1. running `cargo build`
2. running `cd implementations/rust/ockam/ockam_app; cargo tauri dev; cd -`
3. => the application should start and show the `ockam` version in the logs

Then, stop the desktop application and:

1. run `export OCKAM=target/debug/ockam`
2. run `cd implementations/rust/ockam/ockam_app; cargo tauri dev; cd -`
3. the application should exit

Then, stop the desktop application and:

1. run `export OCKAM=/<your ockam root directory>/target/debug/ockam`
2. run `cd implementations/rust/ockam/ockam_app; cargo tauri dev; cd -`
3. the application should start and show the `ockam` version in the logs

----
Fixes #5707

### DIFF
--- a/implementations/rust/ockam/ockam_app/src/cli/mod.rs
+++ b/implementations/rust/ockam/ockam_app/src/cli/mod.rs
@@ -15,7 +15,7 @@ pub(crate) fn cli_bin() -> Result<String> {
 /// its version
 pub(crate) fn check_ockam_executable() -> Result<()> {
     let ockam_path = cli_bin()?;
-    if ockam_path != "ockam".to_string() && !ockam_path.starts_with("/") {
+    if ockam_path != *"ockam" && !ockam_path.starts_with('/') {
         let message = format!("The OCKAM environment variable must be defined with an absolute path. The current value is: {ockam_path}");
         error!(message);
         return Err(Generic(message));
@@ -31,7 +31,7 @@ pub(crate) fn check_ockam_executable() -> Result<()> {
             "The ockam command is available {:?}",
             std::str::from_utf8(&v.stdout)
                 .unwrap_or("can't decode the ockam version")
-                .split("\n")
+                .split('\n')
                 .collect::<Vec<&str>>()
                 .join(" ")
         ),
--- a/implementations/rust/ockam/ockam_app/src/cli/mod.rs
+++ b/implementations/rust/ockam/ockam_app/src/cli/mod.rs
@@ -1,5 +1,40 @@
+use crate::error::Error::Generic;
+use crate::Result;
 use ockam_core::env::get_env_with_default;
+use tracing::{error, info};
 
-pub(crate) fn cli_bin() -> crate::Result<String> {
-    Ok(get_env_with_default("OCKAM", "ockam".to_string())?)
+/// Return the ockam executable path either from the OCKAM env. variable or as `ockam`
+pub(crate) fn cli_bin() -> Result<String> {
+    let ockam_path = get_env_with_default("OCKAM", "ockam".to_string())?;
+    Ok(ockam_path)
+}
+
+/// Check that the OCKAM environment variable defines an absolute path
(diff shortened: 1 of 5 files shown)


implementations/rust/ockam/ockam_app/src/cli/mod.rs
```diff
@@ -1,5 +1,40 @@
...
+/// its version
+pub(crate) fn check_ockam_executable() -> Result<()> {
+    let ockam_path = cli_bin()?;
+    if ockam_path != *"ockam" && !ockam_path.starts_with('/') {
+        let message = format!("The OCKAM environment variable must be defined with an absolute path. The current value is: {ockam_path}");
+        error!(message);
+        return Err(Generic(message));
+    };
+
+    match duct::cmd!(ockam_path, "--version").stdout_capture().run() {
```
adrianbenavides, line 24: I was trying to hack this a little bit and I found two things:
- the tauri app binary accepts and ignores the `--version` argument. It didn't return an error and continued with its normal execution
- when running `cargo tauri dev` with the hacked version (conflicting binaries), the app hangs and the `check_ockam_executable` function never returns.

For the second point, perhaps we should be more conservative and exit early if the process doesn't return in a given time.

```rust
let handle = cmd.start()?;
for ... {
    let is_finished = handle.try_wait()?;
    ...
}
```

shanesveller: If we're not married to `duct`, `tokio::process` has [Command::kill_on_drop](https://docs.rs/tokio/latest/tokio/process/index.html#droppingcancellation) which we could layer with `tokio::time::timeout`.

I'm almost certain this can be accomplished and successful with `duct` without much trouble, just to be clear, so I'm just offering the option.

adrianbenavides: Something to keep in mind is that at that point we don't have the tauri async runtime ready, and I'm not sure we can create an ad-hoc tokio runtime for this (haven't tried it though, perhaps it's also a little overkill).

etorreborre: Thanks for testing @adrianbenavides!

> For the second point, perhaps we should be more conservative and exit early if the process doesn't return in a given time

I think that we don't have to do this anymore since [there shouldn't be anymore conflicts with application names now](https://github.com/build-trust/ockam/pull/5723).

adrianbenavides: Perhaps then this PR doesn't add the value it originally intended?

etorreborre: Yes it does because during our testing I had some issue with the fact that the `OCKAM` env variable was not defined as an absolute path. Then:

- it brings an additional check it is pointing to at least an executable responding to `--version` (which will not be the desktop app because of the renaming)
- it displays the `ockam` version in the logs which can be useful to diagnose a mismatch of versions between the command line application and the desktop application

That being said we should quickly work towards _not_ having to use the `ockam` executable at all :-).

adrianbenavides (approved):

PR #5734, opened 2023-08-24
fix(rust): do not create tcp-outlet and relay if user is not enrolled

adrianbenavides:
Under some circumstances, the app won't be able to reach the project to create the relay, even if it's enrolled (e.g. the user deletes the default project and then tries to run the app).

This PR adds a check so that it doesn't try to create the tcp-outlets and realy if the user is not enrolled and has the default space/project properly set.

### DIFF
--- a/implementations/rust/ockam/ockam_app/src/app/app_state.rs
+++ b/implementations/rust/ockam/ockam_app/src/app/app_state.rs
@@ -163,32 +163,8 @@ impl AppState {
         }
     }
 
-    /// Return true if the user is enrolled
-    /// At the moment this check only verifies that there is a default project.
-    /// This project should be the project that is created at the end of the enrollment procedure
     pub async fn is_enrolled(&self) -> Result<bool> {
-        let identity_state = self.state().await.identities.get_or_default(None)?;
-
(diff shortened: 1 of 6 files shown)


implementations/rust/ockam/ockam_app/src/app/app_state.rs
```diff
@@ -326,7 +303,7 @@ fn load_model_state(
                 model_state
             }
             Err(e) => {
-                println!("cannot load the model state: {e:?}");
+                error!(?e, "cannot load the model state");
```
etorreborre, line 306: 🏅

implementations/rust/ockam/ockam_app/src/options/tray_menu.rs
```diff
@@ -33,7 +33,7 @@ pub(crate) async fn build_options_section(
             let error_item = CustomMenuItem::new(
                 ERROR_MENU_ID,
                 "The application state is corrupted, please re-enroll, reset or quit the application",
-            );
+            ).disabled();
```
etorreborre, line 36: 👏

etorreborre (approved):

PR #5735, opened 2023-08-24
Fix the building of the desktop application

etorreborre:
This PR fixes the execution of `cargo tauri build`

 - there was a compilation error with a feature flag for develop
 - there was a pre-rendering error with a Svelte route for the `/invite/[outlet_addr]` dynamic route

In addition this PR brings another solution to [the overlap of executable names when testing the desktop application during development](https://github.com/build-trust/ockam/pull/5723).

Our objective is to:

 1. be able to create a `.dmg` bundle with the name `Ockam`. This name is defined by the `package.productName` property in `tauri.conf.json`
 
 2. be able to run both `cargo build` and `cd implementations/rust/ockam/ockam_app; cargo tauri dev;` in development to produce both the desktop application executable and the command-line executable in the `target/debug` directory. Since the desktop executable uses the command-line executable we need different names. Unfortunately the `cargo tauri dev` command reuses `package.productName` to define the desktop executable name.

The solution adopted in #5723 contradicts objective 1. I tried to use the possibility to [override the `Info.plist` file](https://jonaskruckenberg.github.io/tauri-docs-wip/building/macos-bundle.html) when bundling the Mac application but that didn't work (at least not for the `.dmg` file).

This PR proposes another solution: override the `tauri.conf.json` file at development time:
```
cd implementations/rust/ockam/ockam_app; cargo tauri dev -c tauri.conf.dev.json; cd -
```
With the command above we can run the desktop application in development and not get a name conflict for our executables.

### DIFF
--- a/implementations/rust/ockam/ockam_app/src/lib.rs
+++ b/implementations/rust/ockam/ockam_app/src/lib.rs
@@ -1,3 +1,19 @@
+//! This crate contains the implementation of the Ockam desktop application.
+//!
+//! In order to run the application in development you need to execute:
+//! ```sh
+//! # to build the `ockam` executable in the target/debug directory
+//! cargo build
+//!
+//! # to build the `ockam_desktop` executable in the target/debug directory and start it
+//! # the overridden tauri configuration renames the package.productName value from "Ockam" to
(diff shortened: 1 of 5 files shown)


adrianbenavides (approved):
Nice!!

PR #5741, opened 2023-08-24
fix(rust): do not filter out accepted invitations, as we can keep them unless revoked

adrianbenavides:
Once an invitation is accepted, they can be used at any time to create the TCP inlet. They only situation where we'll want to discard an accepted invitation is if the inviter revokes it, and we don't have that in place yet.


etorreborre (dismissed):
👍

shanesveller (dismissed):

etorreborre (approved):

PR #5748, opened 2023-08-24
move rust workflow to use nix environment

metaclips:
This PR moves our Rust workflows to use the Nix environment

### DIFF
--- a/implementations/rust/Makefile
+++ b/implementations/rust/Makefile
@@ -1,4 +1,8 @@
-build:
+build_docs:
+	cargo doc --no-deps
+build_examples:
+	cargo --locked build --examples
+build_packages:
 	cargo build
 build_release:
 	cargo --locked build --release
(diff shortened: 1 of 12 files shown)


implementations/rust/Makefile
```diff
@@ -1,37 +1,82 @@
...
-	cargo build --package $*
+	cargo build --locked --package $*
+
+check_no_std:
+	cargo --version
+	cargo check --locked --target thumbv7em-none-eabihf -p ockam --no-default-features --features 'no_std alloc software_vault'
+	# no_std example project
+	cd examples/rust/example_projects/no_std
+	cargo check --example hello
+	# RUSTFLAGS='-Dwarnings' cargo check --example hello --no-default-features --features 'alloc no_std'
```
etorreborre: Can you please remove the commented out code here?

etorreborre (dismissed):
This looks good! I wanted to try some commands locally and couldn't find a way to execute them. Could you please provide an example in the README?

etorreborre (dismissed):
Great stuff and thanks for the README!

As suggested before I think it would eventually be a great idea to switch to `just` instead of `make`: https://github.com/casey/just

metaclips:
> Great stuff and thanks for the README!
> 
> As suggested before I think it would eventually be a great idea to switch to `just` instead of `make`: https://github.com/casey/just

Thanks @etorreborre we can treat adding Just tooling in subsequent PRs

etorreborre (dismissed):

metaclips:
@etorreborre there's an inconsistency in the merge queue caused by this PR, I am pushing a fix

etorreborre (dismissed):

etorreborre (approved):

PR #5753, opened 2023-08-25
feat(rust): show push notifications when enrolling through the app

adrianbenavides:
Note: the notification icon is picked up automatically by tauri:
- when using `tauri dev` it uses the terminal icon
- when using `tauri dev` it uses the bundled icon from `tauri.conf.json`

image

- It shows a notification if we have to create a new project from scratch, to let the user know that it will take some time.
- It also shows a notification (unconditionally) at the end of the enrollment process.

### DIFF
--- a/implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
+++ b/implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
@@ -1,9 +1,9 @@
 use miette::{miette, IntoDiagnostic, WrapErr};
 
 use tauri::{AppHandle, Manager, State, Wry};
-use tracing::{error, info};
+use tauri_plugin_notification::NotificationExt;
+use tracing::{debug, error, info};
 
-use ockam::identity::IdentityIdentifier;
 use ockam_api::cli_state;
 use ockam_api::cli_state::traits::StateDirTrait;
 use ockam_api::cli_state::SpaceConfig;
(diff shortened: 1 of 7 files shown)


implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
```diff
@@ -27,8 +27,7 @@ pub async fn enroll_user(app: &AppHandle<Wry>) -> Result<()> {
     let app_state: State<AppState> = app.state::<AppState>();
     enroll_with_token(app, &app_state)
         .await
-        .map(|i| info!("Enrolled a new user with identifier {}", i))
```
etorreborre, line 30: I would still leave the logs on if we ever need to diagnose what happened for a specific user.

etorreborre (dismissed):
I did not realize that tauri supported notifications, this is great! 
Maybe my only reservation would be that when you enroll, the menu gets hidden and you still don't know what's going on. So you're very tempted to click on the menu. And then this might instantly dismiss the notification. That's what happened to me.

We could show a first notification to tell the user to wait: e226a7727

image

implementations/rust/ockam/ockam_app/src/enroll/enroll_user.rs
adrianbenavides: I just moved that into the `enroll_with_token` function: https://github.com/build-trust/ockam/pull/5753/files#diff-24be761feb7cea4a0282d3086c1eeeaa89a3655cbbad8f7cc41abec52a785cf9R82

etorreborre: Ah, great I did not notice this, thanks.

adrianbenavides:
@etorreborre I've added the extra notification you proposed! Thanks for testing it out 🙏

etorreborre (approved):

mrinalwadhwa:
Nice!

PR #5805, opened 2023-08-28
refactor(rust): move common code to `api` so we can remove `command` from `app`

adrianbenavides:
Move all the `ockam_command` code that was used from `ockam_app` to `ockam_api`. All the moved code was related to either the `CliState` or the `Enroll` service, so it felt like the right move to take all that code to `ockam_api`, which acts as the common interface for both the Cli and the App.

After this refactor, the `ockam_app` doesn't depend anymore on the `ockam_command` crate, which gives us a big compilation time improvement.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4627,6 +4627,7 @@ version = "0.33.0"
 dependencies = [
  "anyhow",
  "aws-config",
+ "base64-url",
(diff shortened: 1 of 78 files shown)


etorreborre (approved):
Great move! 🏎️

PR #5595, opened 2023-08-14
feat(rust): support hostname and/or port on `tcp-outlet create` `--to` argument

alcolmenar:
## Current behavior

Some inputs accept ip addresses and ports but should be able to support hostname and ports. This PR enables hostname and ports for the tcp outlet command. Note: some command access inputs that don't support hostnames and ports, i.e. when specifying a bind address and port.

## Proposed changes

- Change tcp outlet to parameter to be String type
- Update usage of those parameters to enable String type
- Created the HostPort struct that uses the `Host` struct within the url crate
- Created a host_parser util function that validates input for port, ip:port, and hostname:port generating a HostPort struct
- Implemented some conversions to reuse the `host_parser` function with the `socket_addr_parser`

Fixes #5563 

## Checks

Tests pass

- [x] All commits in this Pull Request are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and Verified by Github.
- [x] All commits in this Pull Request follow the Ockam [commit message convention](https://github.com/build-trust/.github/blob/main/CONTRIBUTING.md#commit-messages).
- [x] I accept the Ockam Community [Code of Conduct](https://github.com/build-trust/.github/blob/main/CODE_OF_CONDUCT.md).
- [x] I have accepted the Ockam [Contributor License Agreement](https://github.com/build-trust/ockam-contributors/blob/main/CLA.md) by adding my Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/blob/main/CONTRIBUTORS.csv) file in a separate pull request to the [build-trust/ockam-contributors](https://github.com/build-trust/ockam-contributors) repository. The easiest way to do this is to [edit the CONTRIBUTORS.csv](https://github.com/build-trust/ockam-contributors/edit/main/CONTRIBUTORS.csv) file in the github web UI and create a PR, this will mark the commit as verified.

Issue #5563, opened 2023-08-08
`ockam tcp-outlet create` should support `HOSTNAME:PORT` format


SanjoDeundiak:
Currently, `ockam tcp-outlet create` command located [here](https://github.com/build-trust/ockam/blob/c0ebe05dd644c271ba394b61d8b58edf8e0a31ee/implementations/rust/ockam/ockam_command/src/tcp/outlet/create.rs) only supports `--to` argument in the form of `SocketAddr` (`IP:PORT`). However, supporting the `HOSTNAME:PORT` format is sometimes required (e.g., in containerized environments).  Therefore, `ockam tcp-outlet create` should support `HOSTNAME:PORT` arguments as well.

As an example, `ockam tcp-connection create` command [here](https://github.com/build-trust/ockam/blob/c0ebe05dd644c271ba394b61d8b58edf8e0a31ee/implementations/rust/ockam/ockam_command/src/tcp/connection/create.rs) supports such format.

---

_We love helping new contributors!_
_If you have questions or need help as you work on your first Ockam contribution, please leave a comment on [this discussion](https://github.com/ockam-network/ockam/discussions/1081)._
_If you're looking for other issues to contribute to, checkout [this discussion](https://github.com/build-trust/ockam/discussions/3206) and labels -  https://github.com/build-trust/ockam/labels/good%20first%20issue or https://github.com/build-trust/ockam/labels/help%20wanted_

alcolmenar:
hey @SanjoDeundiak, can I take this one on?

SanjoDeundiak:
@alcolmenar sure

mrinalwadhwa:
@alcolmenar thank you for spending time on this .. everywhere we accept a socket address we want to support all three of these:

1. IP:PORT
2. HOSTNAME:PORT
3. PORT (this assumes IP is 127.0.0.1)

alcolmenar:
@mrinalwadhwa sure thing. ok I'll see what I can do

mrinalwadhwa:
Awesome, thank you.

alcolmenar:
@mrinalwadhwa @SanjoDeundiak Question, in regards to the enabling `HOSTNAME:PORT` for all places SocketAddr is used, there are some cases a bind is being created on the machine running the ockam command, e.g. tcp-inlet where a socket address would be needed usually 127.0.0.1 or 0.0.0.0. I'm thinking to only change areas of the code where a hostname can be used?

PR #5818, opened 2023-08-28
Fix the creation of an outlet when sharing a service

etorreborre:
This PR fixes a bug where we are not able to create an outlet:

image.

The fix consists in using proper types to propagate:

 - the socket address for an outlet
 - the worker address for an outlet worker

The propagation of these types leads to a change in the parsing of a socket address for the command line as well.
This partially reverts some of the code done in https://github.com/build-trust/ockam/pull/5595 so please @adrianbenavides have a look at this part in particular. From my testing it looks that we can use any of these input arguments to create an outlet now:

 - `--to 8080`
 - `--to localhost:8080`
 - `--to 127.0.0.1:8080` (IP V4)
 - `--to [::1]:8080` (IP V6)

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/util/parsers.rs
+++ b/implementations/rust/ockam/ockam_command/src/util/parsers.rs
@@ -1,109 +1,77 @@
-use crate::{exitcode::SOFTWARE, Result};
+use crate::Result;
 use miette::miette;
+use ockam_transport_tcp::resolve_peer;
 use std::net::SocketAddr;
-use url::Host;
 
-/// Helper fn for parsing ip and port from user input
-/// It can parse a string containing either an `ip:port` pair or just a `port`
-/// into a valid SocketAddr instance. Uses the host_parser fn to initially parse
-/// the input as a HostPort and further validates non-Domain type hosts.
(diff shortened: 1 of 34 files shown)


implementations/rust/ockam/ockam_command/src/util/parsers.rs
```diff
@@ -1,109 +1,77 @@
...
+    fn test_localhost() {
+        let input = "localhost:9999";
+        let result = socket_addr_parser(input);
+        assert!(result.is_ok());
+        assert!(result.is_ok());
+        assert_eq!(
+            SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 9999),
+            result.unwrap()
+        );
+    }
```
adrianbenavides, line 71: Nice! This is the test I wasn't sure it would work with `SocketAddr`.

adrianbenavides (approved):
Love the renames and the fact that we are propagating the types correctly... thanks for that! I've done some manual testing using the CLI to make sure the problem from https://github.com/build-trust/ockam/issues/5563 is not reproducible and it all looks good 👍

PR #5826, opened 2023-08-29
refactor(rust): create relay in the background and periodically checks if it needs to be recreated

### LINKED ISSUE #5607: Create relay in the background to avoid blocking the UI

adrianbenavides:
Fixes https://github.com/build-trust/ockam/issues/5607

The project relay is now managed in the background and checked periodically to make sure it exists. It also refactors the `shared_service` module, rewriting the setup function as a plugin, following the line of `projects` and `invitations`.


etorreborre (approved):
That looks good to me except for one piece of logic which I don't understand.
When we load the model state we should populate the list of relays in the `NodeManager`.

Then the background check verifies periodically if the app project forwarder is still referenced by the `NodeManager` (via its `registry`). In which situation can a forwarder be removed from the `registry` associated to the `NodeManager`?

adrianbenavides:
> That looks good to me except for one piece of logic which I don't understand. When we load the model state we should populate the list of relays in the `NodeManager`.
> 
> Then the background check verifies periodically if the app project forwarder is still referenced by the `NodeManager` (via its `registry`). In which situation can a forwarder be removed from the `registry` associated to the `NodeManager`?

The problem is not that the relay gets removed, but rather that it fails to be created. It happened to me that for some reason the controller timed out when creating the relay. In that situation, if we block the process until it's created, the tray menu won't even boot up.

Perhaps we could implement it with a simpler retry strategy... Does that sound better? It would definitely simplify some things around this logic.

etorreborre:
Now I get it thanks.

> Perhaps we could implement it with a simpler retry strategy... Does that sound better? It would definitely simplify some things around this logic.

Yes, maybe a retry strategy with increasing timeouts would be simpler.
In the meantime let's merge this one to at least unblock the current situation.

Issue #5600, opened 2023-08-14
Use TCP Outlets' "worker address" as identifiers in the Ockam App's tray menu items


adrianbenavides:
### Current behavior

Context [here](https://github.com/build-trust/ockam/pull/5586#discussion_r1291238197).

We are not using the right ID to create TCP Outlets menu items in the desktop app.

### Desired behavior

- Use the [TCP Outlet worker address on the system tray menu items](https://github.com/build-trust/ockam/blob/11f533e37247d6afdebc710ff22275b3e1ad08dd/implementations/rust/ockam/ockam_app/src/shared_service/tray_menu.rs#L47)
- The [`on_click` event handler should be refactored](https://github.com/build-trust/ockam/blob/11f533e37247d6afdebc710ff22275b3e1ad08dd/implementations/rust/ockam/ockam_app/src/invitations/tray_menu.rs#L187), changing the variable name (the actual event handler is not yet implemented and it's out of the scope for this issue)

SanjoDeundiak:
I think Ockam Address is a unique identifier for an Outlet. Maybe pair Address&Node_name if we want to manage multiple nodes from the Desktop App. What do you think? @adrianbenavides

adrianbenavides:
We only care about TCP outlets we create for a single local node, so I guess the TCP outlet worker address should be enough then? If that's the case I'll rewrite the issue to use that field.

SanjoDeundiak:
Yeah, the Address is unique inside a node. I would like to avoid introducing another ID, since I just deleted it few month ago :)

adrianbenavides:
Issue updated, thanks @SanjoDeundiak

PR #5830, opened 2023-08-29
fix(rust): fix the sharing of a service by using the socket address of the outlet

etorreborre:
We now designate the service to share with its socket address instead of its worker address, then sending the invitation works: 
image

Part of the issue was due to a refactoring where we changed a function argument name (in #5818) but that argument was also used in a Svelte page function call.


adrianbenavides (approved):
Great, I think this solves https://github.com/build-trust/ockam/issues/5600 once and for all! 💪

PR #5832, opened 2023-08-29
fix(rust): tcp-inlet handling on desktop app

adrianbenavides:
- refactor(rust): invitation state to store the inlet data, so we can show it in the menu
- fix(rust): create tcp-inlet for invitation using separate commands instead of `run`
- fix(rust): don't use expired enrollment tokens when creating the tcp-inlet
- refactor(rust): relay is created using a retry strategy instead of using an event listener

There is a pending TODO that I'll address in a separate PR, as it needs changes on the `tcp-inlet create` command and it would complicate this PR too much.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4698,6 +4698,7 @@ dependencies = [
  "tauri-runtime",
  "thiserror",
  "tokio",
+ "tokio-retry",
  "tracing",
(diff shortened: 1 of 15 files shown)


etorreborre (approved):
Can't wait to try it 🚀

PR #5849, opened 2023-08-31
feat(rust): update the accepted invitation menu with the tcp-inlet socket address

adrianbenavides:
It also contains multiple changes mainly related to improving error management and logging, as well as changes that enhance error propagation rather than silent ignoring, for enroll, reset.

Most of the complexity of the tcp-inlet creation/management will be reduced once we are able to refactor that to stop using the CLI directly.


etorreborre (dismissed):
🌹

etorreborre (approved):

Issue #5684, opened 2023-08-21
In the `Share a service` window, the `Name` attribute should not have the `/service/` prefix


adrianbenavides:
## Current

In the Desktop app, when creating a new service (i.e. the "Share a service" window), the `Name` value contains a prefix that is constant, so it shouldn't be visible to the user.

![image](https://github.com/build-trust/ockam/assets/12375782/078b0d8c-d8cc-43d5-a59b-0878792f32da)

## Desired

The `Name` placeholder should just contain the service name without the `/service/` prefix.

The app backend doesn't need to be updated, as it already sanitizes the service name value and accepts both `/service/name` and `name` formats.

---

_We love helping new contributors!_
_If you have questions or need help as you work on your first Ockam contribution, please leave a comment on [this discussion](https://github.com/ockam-network/ockam/discussions/1081)._
_If you're looking for other issues to contribute to, checkout [this discussion](https://github.com/build-trust/ockam/discussions/3206) and labels -  https://github.com/build-trust/ockam/labels/good%20first%20issue or https://github.com/build-trust/ockam/labels/help%20wanted_

PR #5851, opened 2023-08-31
Remove service prefix

etorreborre:
This PR removes the `/service` prefix when creating a service
image

And also removes the `Worker address` field for a shared service, since this has no meaning for an application user
image

### DIFF
--- a/implementations/typescript/ockam/ockam_app/src/routes/service/ServiceCreate.svelte
+++ b/implementations/typescript/ockam/ockam_app/src/routes/service/ServiceCreate.svelte
@@ -2,7 +2,7 @@
   import { getCurrent } from "@tauri-apps/plugin-window";
   import { invoke } from "@tauri-apps/api/tauri";
 
-  let service = "/service/outlet";
+  let service = "";
   let port = "10000";
   let email = "";
   let error_message = "";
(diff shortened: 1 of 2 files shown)


etorreborre:
Fixes #5684

implementations/typescript/ockam/ockam_app/src/routes/service/ServiceCreate.svelte
```diff
@@ -2,7 +2,7 @@
   import { getCurrent } from "@tauri-apps/plugin-window";
   import { invoke } from "@tauri-apps/api/tauri";
 
-  let service = "/service/outlet";
+  let service = "";
```
adrianbenavides, line 5: I originally added a default value following the "default" principles we've been implementing in the CLI, and I think it's comfortable when you are just working with a single service because I don't have to type anything. Why do you think we should remove the default value? Just curious.

etorreborre: I just couldn't find a good default name. I am also not sure if this is a good idea to encourage users to share a service called "service" since this will show up in the invitee email. Their experience will be better if they receive a meaningful name

adrianbenavides: Yeah that makes sense. Let's do that then! Thanks 🙏

adrianbenavides (approved):

PR #5866, opened 2023-09-01
update gradle workflow to use make

metaclips:
Update [gradle.yml](https://github.com/build-trust/ockam/blob/develop/.github/workflows/gradle.yml) workflow to instead run make commands and use Nix


etorreborre (approved):
🦅

etorreborre (approved):

PR #5868, opened 2023-09-01
feat(rust): restructure the desktop ui

### LINKED ISSUE #5867: The system tray menu must be more explicit
There might be some quick UI wins if we:

 - stop using disabled test to represent section titles (because they look like an action could be possible but is currently not)
 - change the vocabulary: an accepted invitation is actually a remote service we can connect to
 - display names instead of UUIDs for invitations (this might require more work cf #5844)

One thing to attempt is use submenus as section names instead of disabled text.

etorreborre:
This PR is a proposal to structure the desktop application UI differently:

### Main screen
image

### Local services

image

### Invitations

image

### Sent invitations

image

Fixes #5867


etorreborre:
This was done as a prototype and can be closed now.

PR #5878, opened 2023-09-04
tests(rust): restructure bats tests to deprecate the `unit.bats` file

adrianbenavides:
I've also added asserts to almost every possible command we run, to know exactly what command is failing and fail fast.

- CI orchestrator tests: https://github.com/build-trust/ockam-artifacts/actions/runs/6073971677


implementations/rust/ockam/ockam_command/tests/bats/message_orchestrator.bats
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+# ===== SETUP
```
etorreborre, line 3: Nice split 💯

etorreborre (dismissed):
Excellent 🏇

etorreborre (approved):

PR #5900, opened 2023-09-06
fix(rust): capture cmds' output, so it's not written in the logs

### LINKED ISSUE #5710: There must not be any terminal message in the desktop app logs when accepting an invitation
# Observed behavior

When an invitation has been accepted we can see this message in the logs:
image

It comes from the `refresh_inlets` function in [`invitations/command.rs`](https://github.com/build-trust/ockam/blob/b370b26efac5696a4e1547415cc594baef274d30/implementations/rust/ockam/ockam_app/src/invitations/commands.rs#L135).

# Expected behavior

We should make sure that the `--quiet` argument is really working or find another way to delete the node (possibly restart it instead?)

adrianbenavides:
Fixes https://github.com/build-trust/ockam/issues/5710


etorreborre (approved):
🙏 thank you!

PR #5901, opened 2023-09-06
refactor(rust): extract the output of request results from the rpc code

etorreborre:
This PR extracts the printing of responses from an RPC code itself and refactors the `Rpc` API.

## Printing responses

The ability to print responses directly from the `Rpc` struct has been removed and pushed to the `CommandOpts` struct (via the `OutputFormat`). This opens the possibility to later refactor the code so that the terminal can be used instead of a straight `println` to the console.

## `Rpc`

The API of the `Rpc` struct is modified to avoid 2 part calls where we first send the request, then parse the response:

 - `tell` sends a request and don't try to decode the response
 - `ask` sends a request and decodes the response

There are 2 additional cases which are only used once each:

 - `tell_and_get_status` sends a request and returns the response status
 - `ask_and_get_reply` is for the case when we are also interested to surface a failure if the request returns a response with a non-OK code. In that case a special `Reply` data type has been defined (only one case where this is used)

Note that the support of a timing out the call has been made orthogonal to those 3 methods, this is now a property of the `Rpc` client itself.

## Further directions

There will be further PRs to refactor the `RpcBuilder/Rpc` code with the intent of:

 - removing the "`Embedded`" mode
 - being able to create a `Rpc` client directly from the `ockam_api` crate


mrinalwadhwa:
Love what the ask and tell functions are doing!
But wish they had more descriptive name 🤔

etorreborre:
> But wish they had more descriptive name 🤔

I thought they would speak to your Elixir / Erlang heart where `tell` means "send a message to an actor and don't care about the response", whereas `ask` is "send a message and wait for a response" ([example](https://www.bartoszsypytkowski.com/dont-ask-tell-2/)).

I'm happy to use better names if you have any suggestion :-)

implementations/rust/ockam/ockam_command/tests/bats/portals.bats
```diff
@@ -40,7 +40,7 @@ teardown() {
 
   # Test deletion of a previously deleted TCP inlet
   run_failure $OCKAM tcp-inlet delete "test-inlet" --at /node/n2 --yes
-  assert_output --partial "NotFound"
+  assert_output --partial "not found"
```
etorreborre, line 43: The full message is now more explicit but the `OCKAM` error code is `500` (before and after the PR) where it should rather be `404`

etorreborre:
@mrinalwadhwa I attempted something which unfortunately doesn't work. I thought I could replace both `ask` and `tell` with `call` (Remote Procedure "Call") and have compiler happily infer a Unit type for the decoded value when we don't care.

The program compiles but unfortunately the decoding fails at runtime because when no response body exists, the minicbor decoder cannot decode this to `()`. 

I tried to work around this limitation but I don't know if the type system would allow to special case some behaviour just for the Unit type. And even if that was the case, I'm not sure if it is worth the additional complexity.

implementations/rust/ockam/ockam_core/src/api.rs
```diff
@@ -209,6 +238,34 @@ impl Response {
...
+        }
+    }
+}
+
+/// The Reply enum separates two possible cases when interpreting a Response
+///  1. there is a successfuly decodable value of type T
+///  2. the request failed and there is an API error (the optional status is also provided)
+pub enum Reply<T> {
+    Successful(T),
+    Failed(Error, Option<Status>),
```
adrianbenavides, line 268: Nice, love this addition 👍

implementations/rust/ockam/ockam_command/src/output/encode_format.rs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        let o = match self {
+            EncodeFormat::Plain => e.output().wrap_err("Failed serialize output")?,
+            EncodeFormat::Hex => {
+                let bytes = minicbor::to_vec(e).expect("Unable to encode response");
+                hex::encode(bytes)
+            }
+        };
+
+        print!("{o}");
```
adrianbenavides, line 31: Is this a `print` instead of `println` for a particular reason?

adrianbenavides (approved):
Love it, feels much better now!

implementations/rust/ockam/ockam_command/src/output/encode_format.rs
etorreborre: Good catch, that should probably be a `println` but I just moved the code as is. I left a note on [this follow-up issue](https://github.com/build-trust/ockam/issues/5904) to deal with the question.

adrianbenavides: OK, sounds good!

PR #5911, opened 2023-09-07
feat(elixir): adapt to new identity design

polvorin:
* Adapts to new identity design
* Get rid of the separate sidecar and vault ffi
* Use rustler for rust>elixir bindings on Identity/Credentials (implementations/elixir/ockam/ockly/native/ockly)
* Remove identity service and verifier service from ockam (those were only used by the "sidecar")

@SanjoDeundiak 
I think it's ok to merge into your branch.   There are some couple of missing things that I can iterate after,  mainly:
* There is no history change comparison on elixir (need to write the binding for that),  right now the function just  happily says histories match
* Examples using secure channels on examples/elixir/* need to be adapted
* There are some error conditions that can be better handled on the lib
* The interface with aws vault is there,  but I haven't tested it

[This](https://github.com/build-trust/ockam/blob/34a6f1ee2a960f213b6b53e38753f8eece02e892/implementations/elixir/ockam/ockly/native/ockly/src/lib.rs) is what replace the sidecar ,  some light :eyes: welcomed,  there is likely some generalization to be done there but we can iterate as well.   The interface is limited what was needed on elixir' side (so for example,  it only attest to securechannel keys,  the secret type it's not given as argument, etc).


implementations/rust/ockam/ockam_api/src/nodes/service/node_identities.rs
```diff
@@ -27,47 +24,11 @@ impl NodeIdentities {
         self.identities.vault()
     }
 
-    pub(super) fn identities_repository(&self) -> Arc<dyn IdentitiesRepository> {
```
SanjoDeundiak, line 29: I'm not sure removing it is useful and related to that PR, but if these functions just become unused, should be OK

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+                .build()
+                .await
+        } else {
+            identities_ref.identities_creation().create_identity().await
+        }
+    })
+    .map_err(|_| Error::BadArg)?;
+
+    let exported = identity.export().map_err(|_| Error::BadArg)?;
+    let id = identity.identifier().to_string();
```
SanjoDeundiak, line 90: Let's keep it in binary as 20 bytes

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+    let mut binary = NewBinary::new(env, id.len());
```
SanjoDeundiak, line 91: It's surprising how inconvenient it is. `&[u8]` -> `NewBinary` could have been a one simple call instead of 3

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+    binary.copy_from_slice(id.as_bytes());
+    let mut exp_binary = NewBinary::new(env, exported.len());
+    exp_binary.copy_from_slice(&exported);
+    Ok((binary.into(), exp_binary.into()))
+}
+
+#[rustler::nif]
+fn attest_purpose_key<'a>(
+    env: Env<'a>,
+    identifier: String,
```
SanjoDeundiak, line 101: Let's use binaries where we can

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+    let identifier = Identifier::from_str(&identifier)
+        .map_err(|_| Error::Term(Box::new(atoms::invalid_identifier())))?;
+    let purpose_key = block_future(async move {
+        let key_id = identities_ref
+            .purpose_keys()
+            .purpose_keys_creation()
+            .vault()
+            .secure_channel_vault
+            .import_static_secret(
+                Secret::new(public_key.to_vec()),
```
SanjoDeundiak: That doesn't look correct. I will add a possibility to attest a PurposeKey having only a public key available.

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+    attestation: Binary,
+) -> NifResult<bool> {
+    let identities_ref = identities_ref()?;
+    let attestation: PurposeKeyAttestation = minicbor::decode(&attestation)
+        .map_err(|_| Error::Term(Box::new(atoms::attestation_decode_error())))?;
+    let k = PublicKey::new(public_key.as_slice().to_vec(), SecretType::X25519);
+    block_future(async move {
+        let identity = identities_ref
+            .identities_creation()
+            .import(None, &identity)
```
SanjoDeundiak, line 147: Are the cases where we should check the Identifier match here or on the calling side?

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+    .map_err(|_| Error::Term(Box::new(atoms::attest_error())))?;
+    let encoded = minicbor::to_vec(purpose_key.attestation())
+        .map_err(|_| Error::Term(Box::new(atoms::attestation_encode_error())))?;
+    let mut exp_binary = NewBinary::new(env, encoded.len());
+    exp_binary.copy_from_slice(&encoded);
+    Ok(exp_binary.into())
+}
+
+#[rustler::nif]
+fn verify_purpose_key_attestation(
```
SanjoDeundiak: Let's rename it since it only works with Secure Channel key. Or let's add support for Credentials keys

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+    let mut binary = NewBinary::new(env, id.len());
+    binary.copy_from_slice(id.as_bytes());
+    Ok(binary.into())
+}
+
+#[rustler::nif]
+fn issue_credential<'a>(
+    env: Env<'a>,
+    issuer_identity: Binary,
+    subject_identifier: String,
```
SanjoDeundiak, line 192: Do we import subject's Identity before making that call?

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+    let identities_ref = identities_ref()?;
+    let subject_identifier = Identifier::from_str(&subject_identifier)
+        .map_err(|_| Error::Term(Box::new(atoms::invalid_identifier())))?;
+    let credential = block_future(async move {
+        let issuer = identities_ref
+            .identities_creation()
+            .import(None, &issuer_identity)
+            .await
+            .map_err(|_| atoms::identity_import_error())?;
+        let mut attr_builder = AttributesBuilder::with_schema(SchemaId(0));
```
SanjoDeundiak, line 205: Schema should be customizable I guess

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+    let id = imported_identity.identifier().to_string();
+    let mut binary = NewBinary::new(env, id.len());
+    binary.copy_from_slice(id.as_bytes());
+    Ok(binary.into())
+}
+
+#[rustler::nif]
+fn issue_credential<'a>(
+    env: Env<'a>,
+    issuer_identity: Binary,
```
SanjoDeundiak, line 191: Issuer should have valid Credential PurposeKey to perform that, or the default one will be created under the hood

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+    subject_identifier: String,
+    attrs: HashMap<String, String>,
+    duration: u64,
+) -> NifResult<Binary<'a>> {
+    let identities_ref = identities_ref()?;
+    let subject_identifier = Identifier::from_str(&subject_identifier)
+        .map_err(|_| Error::Term(Box::new(atoms::invalid_identifier())))?;
+    let credential = block_future(async move {
```
SanjoDeundiak: That's also `CredentialAndPurposeKey` FYI

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+        .map_err(|_| Error::Term(Box::new(atoms::credential_encode_error())))?;
+    let mut binary = NewBinary::new(env, encoded.len());
+    binary.copy_from_slice(&encoded);
+    Ok(binary.into())
+}
+
+#[rustler::nif]
+fn verify_credential(
+    expected_subject: String,
+    authorities: Vec<Binary>,
```
SanjoDeundiak, line 232: I wonder if it's easier to first import Identities, and then pass only Identifiers. Not only for this call, for others as well

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+    true
+}
+
+fn load_aws_vault() -> bool {
+    block_future(async move {
+        if let Ok(vault) = AwsSigningVault::create().await {
+            let builder = ockam_identity::Identities::builder().with_vault(Vault::new(
+                Arc::new(vault),
+                Vault::create_secure_channel_vault(),
+                Vault::create_credential_vault(),
```
SanjoDeundiak: This one should be AWS as well

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -0,0 +1,352 @@
...
+                .expires_at
+                .deref(),
+            attr_map,
+        ))
+    });
+    attributes.map_err(|reason: Atom| Error::Term(Box::new(reason)))
+}
+
+#[rustler::nif]
+fn import_signing_secret(secret: Binary) -> NifResult<String> {
```
SanjoDeundiak, line 284: Is it required?

SanjoDeundiak (changes requested):
I didn't review Elixir code. Also, I haven't had experience with rustler, so may be missing some details on correct use of it. I reviewed rust code, looks like a good start. We can merge it and iterate from there

SanjoDeundiak:
Could you please look at why CI fails? I think only fails that are already present in my PR are expected here, others can be fixed.

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
polvorin: oh,  good catch, the *name* is wrong!.    The argument is not the `public_key`,  it's actually the `secret` .     It is hacky as we import the secret just to be able to attest to the corresponding public key.  

Originally I added what you said (a way to attest to public key directly) and so the public key was passed,  but removed that to not introduce changes on that part of the ockam_identity crate that was giving me conflicts.   But forgot to rename the argument according.     Will fix it (and yes not requiring the secret will simplify this part, will be a good addition)

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
polvorin: right,   this and the above one are only for secure channels keys, will rename

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
polvorin: yes,  I assumed a default one would be created (and a new default after a restart),  but it's ok this is for testing only right now  (we don't have an authority implementation on elixir,   so only need to verify credentials but not really generate them other than for testing/examples)

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
polvorin: ah,  i see what you mean.   We don't enforce that,  it happens to be that way because this is used for tests so the identity is generated by us.    But will fix this so it's evident what's going on

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
polvorin: thanks!, will fix

polvorin:
Closing in favor of https://github.com/build-trust/ockam/pull/5976.   Created issues for the pending TODOs

PR #5913, opened 2023-09-07
feat(rust): upgraded ockam app to tauri alpha.11

davide-baldo:
Tauri upgraded to 2.0.0.alpha.11 had to migrate the menu creation and adjust dependencies.
Currently, [separator menu item leads to crash](https://github.com/tauri-apps/tauri/issues/7760) and has been temporary removed.

### DIFF
--- a/implementations/rust/ockam/ockam_app/src/app/mod.rs
+++ b/implementations/rust/ockam/ockam_app/src/app/mod.rs
@@ -7,6 +7,8 @@ pub use setup::*;
 pub use state::*;
 pub use tray_menu::*;
 
+#[cfg(debug_assertions)]
+mod dev_tools;
 pub(crate) mod events;
 #[cfg(feature = "log")]
 mod logging;
--- a/implementations/rust/ockam/ockam_app/src/app/process.rs
+++ b/implementations/rust/ockam/ockam_app/src/app/process.rs
@@ -1,8 +1,8 @@
 use crate::app::events::system_tray_on_update;
-use tauri::{AppHandle, RunEvent, Wry};
+use tauri::{AppHandle, RunEvent, Runtime};
 
 /// This is the function dispatching application events
-pub fn process_application_event(app: &AppHandle<Wry>, event: RunEvent) {
+pub fn process_application_event<R: Runtime>(app: &AppHandle<R>, event: RunEvent) {
     match event {
         RunEvent::ExitRequested { api, .. } => {
             api.prevent_exit();
--- a/implementations/rust/ockam/ockam_app/src/app/setup.rs
+++ b/implementations/rust/ockam/ockam_app/src/app/setup.rs
@@ -1,40 +1,64 @@
 use crate::app::events::{SystemTrayOnUpdatePayload, SYSTEM_TRAY_ON_UPDATE};
-use crate::app::{build_tray_menu, process_system_tray_event};
+use crate::app::{build_tray_menu, process_system_tray_menu_event};
 use std::error::Error;
-use tauri::{async_runtime::spawn, App, Manager, SystemTray, Wry};
+use std::mem::forget;
+use std::sync::{Arc, Mutex};
+use tauri::async_runtime::spawn;
+use tauri::menu::Menu;
+use tauri::tray::TrayIconBuilder;
+use tauri::{App, Manager, Runtime};
 use tracing::{debug, error, info};
 
 /// Create the initial version of the system tray menu and the event listeners to update it.
-pub fn setup(app: &mut App<Wry>) -> Result<(), Box<dyn Error>> {
+pub fn setup<R: Runtime>(app: &mut App<R>) -> Result<(), Box<dyn Error>> {
     debug!("Setting up app");
 
-    let handle = app.handle();
-    let tray_menu = tauri::async_runtime::block_on(build_tray_menu(&handle, None));
-    SystemTray::new()
-        .with_menu(tray_menu)
-        .on_event(move |event| process_system_tray_event(&handle, event))
+    // Building the tray here rather than using the `tauri.conf.json` mechanism
+    // to sidestep breaking changes the file format and consequent dependency
+    // on a specific version of tauri-cli.
+    TrayIconBuilder::with_id("tray")
+        .tooltip("Ockam")
+        .icon_as_template(true)
+        .icon(app.default_window_icon().unwrap().clone())
+        .on_menu_event(process_system_tray_menu_event)
         .build(app)
         .expect("Couldn't initialize the system tray menu");
 
-    let handle = app.handle();
-    app.listen_global(SYSTEM_TRAY_ON_UPDATE, move |event| {
-        let payload = match event.payload() {
-            Some(p) => match SystemTrayOnUpdatePayload::try_from(p) {
-                Ok(p) => Some(p),
-                Err(e) => {
-                    error!(?e, "Couldn't deserialize payload");
-                    None
-                }
-            },
-            None => None,
-        };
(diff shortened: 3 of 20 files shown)


implementations/rust/ockam/ockam_app/src/invitations/plugin.rs
```diff
@@ -12,7 +12,7 @@ use super::commands::*;
 use super::events::{REFRESHED_INVITATIONS, REFRESH_INVITATIONS};
 use super::state::InvitationState;
 
-const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(60);
+const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(15);
```
adrianbenavides: I think this should be at least 30s (the default api timeout) so we don't chain multiple poll events.

implementations/rust/ockam/ockam_app/src/app/mod.rs
```diff
@@ -7,6 +7,7 @@ pub use setup::*;
 pub use state::*;
 pub use tray_menu::*;
 
+mod dev_tools;
```
adrianbenavides, line 11: This whole module should be behind the `debug_assertions` flag right? Or at least behind `not(release)` if we don't want to use tauri's flags.

implementations/rust/ockam/ockam_app/src/app/process.rs
```diff
@@ -1,8 +1,8 @@
 use crate::app::events::system_tray_on_update;
-use tauri::{AppHandle, RunEvent, Wry};
+use tauri::{AppHandle, RunEvent, Runtime};
 
 /// This is the function dispatching application events
-pub fn process_application_event(app: &AppHandle<Wry>, event: RunEvent) {
+pub fn process_application_event<R: Runtime>(app: &AppHandle<R>, event: RunEvent) {
```
adrianbenavides, line 5: Out of the scope of this PR but, what's the benefit of using this trait bound for this argument? I think we have mix uses when passing the `AppHandle` argument.

implementations/rust/ockam/ockam_app/src/app/setup.rs
```diff
@@ -1,40 +1,65 @@
...
-    let tray_menu = tauri::async_runtime::block_on(build_tray_menu(&handle, None));
-    SystemTray::new()
-        .with_menu(tray_menu)
-        .on_event(move |event| process_system_tray_event(&handle, event))
+    // Building the tray here rather than using the `tauri.conf.json` mechanism
+    // to sidestep breaking changes the file format and consequent dependency
+    // on a specific version of tauri-cli.
+    TrayIconBuilder::with_id("tray")
+        .tooltip("Ockam")
+        .title("Ockam")
```
adrianbenavides: I think either tooltip or title should not be used. This is how it looks on MacOS at least:

image

implementations/rust/ockam/ockam_app/src/shared_service/tray_menu.rs
```diff
@@ -10,56 +10,61 @@ pub const SHARED_SERVICE_HEADER_MENU_ID: &str = "shared_service_header";
...
-    let tm = tray_menu
-        .add_native_item(SystemTrayMenuItem::Separator)
-        .add_item(CustomMenuItem::new(SHARED_SERVICE_HEADER_MENU_ID, "Shared").disabled())
-        .add_item(CustomMenuItem::new(
-            SHARED_SERVICE_CREATE_MENU_ID,
-            "Create...",
-        ));
+    for outlet in app_state.tcp_outlet_list().await {
+        builder = builder.item(&shared_service_submenu(app_handle, &outlet))
+    }
```
adrianbenavides: I'm mostly positive that this loop is duplicating the shared services items?

image

implementations/rust/ockam/ockam_app/src/invitations/tray_menu.rs
```diff
@@ -32,206 +28,237 @@ const PATH_ENCODING_SET: AsciiSet = CONTROLS
...
+                .enabled(false)
+                .build(app_handle),
+            &MenuItemBuilder::with_id(
                 format!("invitation-sent-cancel-{}", invitation.id),
                 "Cancel",
             )
-            .disabled(),
-        );
-    SystemTraySubmenu::new(id, submenu)
+            .enabled(false)
```
adrianbenavides, line 107: Looks like this is not being honored. Do you know if it's a known problem of the current alpha version?

image

implementations/rust/ockam/ockam_app/src/shared_service/tray_menu.rs
```diff
@@ -77,6 +82,11 @@ pub fn on_create(app: &AppHandle<Wry>) -> tauri::Result<()> {
             // TODO: ideally we should use Position::TrayCenter, but it's broken on the latest alpha
             let _ = w.move_window(Position::TopRight);
             w.show()?;
+
+            #[cfg(debug_assertions)]
+            {
+                w.open_devtools();
+            }
```
adrianbenavides: I found this pretty annoying/intrusive while testing the app.

adrianbenavides (commented):
Looks great! I expected this upgrade to be definitely more destructive. I hope we don't suffer this kind of breaking changes in the comming versions 😓

implementations/rust/ockam/ockam_app/src/app/process.rs
davide-baldo: Since tauri has multiple runtimes (I think it's desktop and mobile) by using the trait bound we can be neutral, and I just think it should be the cleanest way to do it to avoid surprises on upgrade.

implementations/rust/ockam/ockam_app/src/app/setup.rs
davide-baldo: Yeah, it's the title. I loved the way it looked, I'll remove it and we can refine the view later.

adrianbenavides: OK, let's keep @nazmulidris updated on this so he can keep it in mind while developing the new design 👍

implementations/rust/ockam/ockam_app/src/app/process.rs
adrianbenavides: >  (I think it's desktop and mobile)

That's the piece I was missing, thanks!

implementations/rust/ockam/ockam_app/src/shared_service/tray_menu.rs
davide-baldo: Maybe we can add a tickbox in the develop tools meun, disabled by default

implementations/rust/ockam/ockam_app/src/app/setup.rs
nazmulidris: @adrianbenavides @davide-baldo Currently in the designs we have the icon looking like this 
desktop app icon

I can easily change it to accommodate icon + text if that is what we decide to go with.

implementations/rust/ockam/ockam_app/src/invitations/tray_menu.rs
davide-baldo: I didn't notice it, but yeah, seems to be the case

implementations/rust/ockam/ockam_app/src/app/setup.rs
davide-baldo: let's focus on merge, so we can adjust it later

etorreborre (approved):

Issue #5902, opened 2023-09-06
Refactor the `Rpc` support


etorreborre:
The `ockam_command` crate contains some support to make RPC calls to a node. However this RPC support is also used to make local calls to a `NodeManager`. 

The purpose of this story is to remove the `Embedded` mode from the `Rpc` struct so that local calls are made directly.
At the minimum this supposes to:

 - [x] remove the integrating printing of response from the `Rpc` client
 - [x] simplify the `Rpc` API and directly return typed values when possible (at the moment, the `Rpc` API exposes too many details about the processing of a response)
 - [x] possibly revisit the necessity of a `RpcBuilder` (Maybe merge it within the `Rpc` struct)
 - [x] simplify the creation of a `Rpc` into 2 distinct cases: "embedded" and "background"
 - [x] make specific RPC clients for accessing: the controller, the project node, the authority node
 - [x] create typed interfaces for each of those clients depending on their capabilities (for example the `Controller` can manage projects)
     - [x] Projects
     - [x] Spaces
     - [x] Subscriptions
     - [x] Invitations
     - [x] Enrollment 
 - [x] use the typed interfaces everywhere possible
 - [x] remove unused code using the old `RpcClient`
 - [x] fix the examples involving credentials
 - [x] fix the [support for creating tickets in one case](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/project/ticket.rs#L142)
 - [x] simplify the relationship between `NodeManagerWorker` and `NodeManager` (all the logic should be in the `NodeManager`)
 - [x] put back the `Connection` support for creating secure channels
 - [x] pass the unit tests
 - [x] pass the local bats tests
 - [x] document and comment
 - [x] pass the Orchestrator bats tests

PR #5916, opened 2023-09-07
Separate the Rcp data type into embedded and background

etorreborre:
This PR is preparing further work for simplifying `Rpc` calls:

 - all the creation of `Rpc` have been reduced to either `Rpc::embedded` or `Rpc::background`
 - a `&mut Rpc` is passed around instead of `Context + CommandGlobalOpts` and have `Rpc` recreated in different places
 - this allows a simplification of the `Rpc` type which doesn't need a lifetime parameter anymore, nor to to embark an optional `TcpTransport`

This goes towards implementing [this story](https://github.com/build-trust/ockam/issues/5902).

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/admin/subscription.rs
+++ b/implementations/rust/ockam/ockam_command/src/admin/subscription.rs
@@ -173,9 +173,7 @@ async fn run_impl(
             space_id,
         } => {
             let subscription_id = utils::subscription_id_from_cmd_args(
-                &ctx,
-                &opts,
-                rpc.node_name(),
+                &mut rpc,
                 controller_route,
                 subscription_id,
                 space_id,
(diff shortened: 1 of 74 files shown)


davide-baldo:
@etorreborre 
Won't re-creating the RPC instance re-initialize the embedded node multiple times? It could even spawn sub-process multiple times, does it have known side effects?

etorreborre:
It shouldn't. The `Rpc` instance is instantiated once at the top-level then passed down as many times as necessary. Eventually anyway this will be replaced with a `NodeManager` instance.

adrianbenavides (approved):
Great! I did some manual testing to try some bad usage cases that caused a failure in the previous implementation (e.g. trying to create two RPC's would fail if they were not sharing the same TCP transport instance). I also run all the bats tests on CI and it's all green!

PR #5932, opened 2023-09-08
Make direct controller calls for the "embedded" RPC case

etorreborre:
This PR refactors the support for making RCP calls:

 1. it removes RPC calls when they are only an indirection for calling a method on an in-memory `NodeManager`
 
 2. it standardizes the various existing RPC clients into 4 clients aimed at 4 different types of nodes:
      a. the Controller
      b. the project node
      c. the authority node
      d. a background node running on the same machine

The `Controller`, `ProjectNode` and `AuthorityNode` are all using a `SecureClient`. This means that they can create a secure channel to a remote node and send requests. However they have specific types to be able to distinguish what they can do in reality:

 - for example a `Controller` client can manage spaces and projects
 - and an `AuthorityNode` can issue credentials

 3. all other calls, when destined to create an in-process entity: tcp-inlet, secure-channel, identity, etc... should use the `InMemoryNode` directly

This is a PR touching many files. My recommendation is to 

  - have a look at the top-level structs introduced by this PR to understand the use cases
  - review the PR crate by crate, from low-level dependencies to high-level ones. I have also added comments on the PR directly to help the review.

## Main structs

If you just want an overview of the main structs you can read the following files:

 - [`in_memory_node`](https://github.com/build-trust/ockam/blob/27747c309c4484165544191502c02fd66da649a8/implementations/rust/ockam/ockam_api/src/nodes/service/in_memory_node.rs) starts a node in the current process and can create clients
 - [`background_node`](https://github.com/build-trust/ockam/blob/27747c309c4484165544191502c02fd66da649a8/implementations/rust/ockam/ockam_api/src/nodes/service/background_node.rs) client for sending request to a node on the same machine
 - [`secure_client`](https://github.com/build-trust/ockam/blob/dd5af7ef020ff5ed6f571675ea214c5679c2f6ab/implementations/rust/ockam/ockam_identity/src/secure_channels/secure_client.rs) and [`secure_clients`](https://github.com/build-trust/ockam/blob/dd5af7ef020ff5ed6f571675ea214c5679c2f6ab/implementations/rust/ockam/ockam_api/src/cloud/secure_clients.rs) various clients using secure channels
 - [`Client`](https://github.com/build-trust/ockam/blob/dd5af7ef020ff5ed6f571675ea214c5679c2f6ab/implementations/rust/ockam/ockam_node/src/api.rs) low level client for making requests / responses to a given route
 
### `ockam_core`

 - the `tag` feature (used to add `TypeTags` to CBOR structures) was unused and has been removed. In its place I added a quickcheck test to verify that a given struct serializes according to the CBOR schema

 - the `TypeTag` field has been removed from all data types (but CBOR indices are left starting from `1` for backwards compatibility for now)
 
 - the `Request` and `Response` structs have been renamed to `RequestHeader` and `ResponseHeader` (which is what they are)
 
 - `Response` creation functions (to build `ok`, `bad_request` etc...) have been centralized to the `Response` declaration

### `ockam_identity`

 - `credentials` most of the changes here are adaptations due to the changes in `ockam_core`
 - `secure_client` contains a client which can create a secure channel to a node, send typed requests, and get typed responses back

### `ockam_multiaddr` 

Just added a convenient `MultiAddr::starts_with` method

### `ockam_node`

 - `api` introduce a `Client` which can send typed requests and receive typed responses on a given route. It is used by the `SecureClient` but can be used as a standalone. This allows to centralize in one place all the uses of the `request` function
 
 - the `rpc_client` has been removed

### `ockam_api`

I am listing here the most importants changes besides those resulting from the `ockam_core` refactorings.

- `auth` the ad-hoc client for authorization has been replaced with a trait and a `BackgroundNode` implementation
- `authenticator` the ad-hoc client has been removed
- `enrollment_tokens` the ad-hoc client has been removed
- `issuer` some traits and implementations for an `AuthorityNode` have been introduced
- `cloud` interfaces have been introduced to access spaces, projects, addons etc... with a corresponding `Controller` implementation
- `enroll` the `Enrollment` trait has been added to support the various enrollment methods
- `influxdb_lease` the ad-hoc `OrchestratorApi` client has been removed in favour of an interface + a `ProjectNode` implementation

- `nodes/connection` the `Connection` code has been simplified so that a `ConnectionBuilder` eventually creates a `Connection` without calling it an "instance". The interface has been changed in order to be able to pass a `NodeManager`. This was necessary to get a better separation between the `NodeManagerWorker` and the `NodeManager`
- `nodes/registry` the registries of a `NodeManager` have now interior mutability. This allows to avoid locking the whole `NodeManager` just to modify one registry and having to pass a mutable reference to the `NodeManager`
- `nodes/service/background_node` contains a `BackgroundNode` client to send requests to a local node
- `nodes/service/in_memory_node` contains an `InMemoryNode` which is a wrapper around a `NodeManager` + the code to manage sessions. This is the entry point for the `ockam_command` crate. With an `InMemoryNode` you can either run some in process node services (inlets, outlets, etc...) but also create clients to communicate with other nodes
- `nodes/service/*` files have a bunch of changes trying to put more code from `NodeManagerWorker` into `NodeManager` but there would be more clean-ups to do

### `ockam_app`

Mostly the `ockam_app` doesn't have to maintain a lock on a `NodeManagerWorker` anymore but it can use an `InMemoryNode` directly, and in many places just a `Controller` client.

### `ockam_command`

The modifications in `ockam_command` are mostly the effect of using either:

 - a `BackgroundNode` to communicate with a node on the same machine
 - a `Controller` to manage spaces, projects, invitations etc...
 - a `Project` client to call the project node
 - an `Authority` client to call the authority node
 - an `InMemoryNode` to start a background node

There are also a few edge case uses of an `InMemoryNode`, to send a message for example, but this can be refactored later


implementations/rust/ockam/ockam_identity/src/secure_channels/secure_client.rs
```diff
@@ -0,0 +1,179 @@
...
+        req: Request<T>,
+        timeout: Duration,
+    ) -> Result<Vec<u8>>
+    where
+        T: Encode<()>,
+    {
+        let sc = self.create_secure_channel(ctx).await?;
+        let route = route![sc.clone(), api_service];
+        let client = Client::new(&route, Some(timeout));
+        let response = client.request(ctx, req).await?;
```
adrianbenavides: To guarantee the secure channel is stopped, shouldn't we delay unwrapping the result here  after calling `stop_secure_channel`?

adrianbenavides (dismissed):
Looks awesome! I followed the walkthrough while reviewing the changes, but I didn't go through every single file in the cases were the same kind of change was being applied to several related modules.

That being said, I couldn't find anything that should hold back this PR from merging. The API feels super easy to use and I really like the different clients you extracted from the tangled mess we had. Tests are also green locally, so I'll leave it as approved 🚀

adrianbenavides (approved):

PR #5941, opened 2023-09-11
refactor(rust): implement new tray menu design for general items and shared services

adrianbenavides:
Note: separators are not visible because they are currently not working on tauri's 2.0.0-alpha.11 version.

In this PR I've modified the top-level items. The invitations details within shared services as well as the full invitations section will be done in different PR's.

image
Screenshot 2023-09-11 at 15 09 42

## Bugs fixed

There were a couple of bugs where local `default` project was being overwritten by remote's `default` project:
- fix(rust): race condition bug where the `refresh_projects` function was overwritting the user's owned project data
- fix(rust): modify enrollment ticket's project name to avoid rewritting the user's owned project data 

There was also a dead lock when refreshing the inlets, which made the application unable to finish a background task + it never updated the invitation's menu:
- fix(rust): dead lock when refreshing inlets (reader and a writer simultaneously over the same data)


davide-baldo:
I think reference icons directly with a `File` could break unless the `icons` directory is copied when building the package on Linux and on Mac or the icon needs to be embedded like `.icon(Icon::Raw(include_bytes!("my-icon.png").to_vec()))`

adrianbenavides:
> I think reference icons directly with a `File` could break unless the `icons` directory is copied when building the package on Linux and on Mac or the icon needs to be embedded like `.icon(Icon::Raw(include_bytes!("my-icon.png").to_vec()))`

That makes a lot of sense, thanks for the heads up!

davide-baldo (approved):

PR #5963, opened 2023-09-13
Move ockam command and docs test to use nix env

metaclips:
Move ockam command and docs test to use nix env


davide-baldo (approved):
nice!

PR #5965, opened 2023-09-13
refactor(rust): new menu design for invitations

adrianbenavides:
## Current limitations

- I wanted to show the pending invites as a submenu of the shared service. Currently we can't do that because the pending invitation data we receive from the backend doesn't include the shared service name, only the invitation id, so we don't have a way to match them
- The backend doesn't return the accepted invitations to the sharer user, only the accepted invitations from the invited user perspective
- Related to the first point, the pending invitations can't show the service name, only the invitation id

## Shared services
image

## Pending invitations
image

## Connected services
image

### DIFF
--- a/implementations/rust/ockam/ockam_app/src/invitations/tray_menu.rs
+++ b/implementations/rust/ockam/ockam_app/src/invitations/tray_menu.rs
@@ -1,19 +1,22 @@
+use arboard::Clipboard;
 use percent_encoding::{percent_encode, AsciiSet, CONTROLS};
+use std::collections::HashMap;
 use std::net::SocketAddr;
-use tauri::menu::{MenuBuilder, MenuEvent, MenuItemBuilder, Submenu, SubmenuBuilder};
-use tauri::{AppHandle, Manager, Runtime, State};
+use tauri::async_runtime::spawn;
+use tauri::menu::{
+    IconMenuItemBuilder, MenuBuilder, MenuEvent, MenuItemBuilder, NativeIcon, Submenu,
+    SubmenuBuilder,
+};
+use tauri::{AppHandle, Icon, Manager, Runtime, State};
 use tauri_plugin_positioner::{Position, WindowExt};
-use tracing::{debug, trace, warn};
+use tracing::{debug, error, trace, warn};
 
-use ockam_api::cloud::share::{InvitationWithAccess, ReceivedInvitation, SentInvitation};
+use ockam_api::cloud::share::{ReceivedInvitation, SentInvitation, ServiceAccessDetails};
 
 use super::state::SyncInvitationsState;
 use crate::app::AppState;
 use crate::invitations::state::AcceptedInvitations;
 
-pub const INVITATIONS_PENDING_HEADER_MENU_ID: &str = "sent_invitations_header";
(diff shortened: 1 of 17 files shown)


implementations/rust/ockam/ockam_app/src/invitations/tray_menu.rs
```diff
@@ -100,12 +86,13 @@ fn pending_invitation_menu<R: Runtime>(
             &MenuItemBuilder::with_id(id.clone(), &invitation.recipient_email)
                 .enabled(false)
                 .build(app_handle),
-            &MenuItemBuilder::with_id(
-                format!("invitation-sent-cancel-{}", invitation.id),
-                "Cancel",
-            )
-            .enabled(false)
-            .build(app_handle),
+            // TODO: not supported yet
```
etorreborre: I would just remove commented out code. By the time we implement it the variables, data types might have changed and it won't be hard to re-create (same thing for other places in the PR).
But otherwise good move to only display what's relevant!

implementations/rust/ockam/ockam_app/src/invitations/tray_menu.rs
```diff
@@ -115,24 +102,24 @@ fn add_received_menu<R: Runtime>(
...
-
-    let mut submenu_builder =
-        SubmenuBuilder::with_id(app_handle, INVITATIONS_RECEIVED_HEADER_MENU_ID, header_text);
-
+    debug!(
+        count = received.len(),
+        "Building menu for received invitations"
+    );
+
+    let header_text = format!(
```
etorreborre, line 103: I suspect that this kind of code should be put in a `text` directory / file somewhere for text manipulation with an appropriate `pluralize` function. I've had to do that quite a few times to display quantities in various libraries.

implementations/rust/ockam/ockam_app/src/invitations/tray_menu.rs
```diff
@@ -348,12 +323,10 @@ fn on_create<R: Runtime>(app: &AppHandle<R>, outlet_socket_addr: &str) -> tauri:
...
 
-fn on_connect<R: Runtime>(_app: &AppHandle<R>, invite_id: &str) -> tauri::Result<()> {
-    trace!(?invite_id, "connecting to service via spawn");
-    todo!()
-}
-
-fn on_decline<R: Runtime>(_app: &AppHandle<R>, invite_id: &str) -> tauri::Result<()> {
-    trace!(?invite_id, "declining invite via spawn");
-    todo!()
+fn on_copy<R: Runtime>(_app: &AppHandle<R>, socket_address: &str) -> tauri::Result<()> {
```
etorreborre, line 303: Great UX improvement!

etorreborre (dismissed):
Nice improvements!

implementations/rust/ockam/ockam_app/src/invitations/tray_menu.rs
adrianbenavides: Noted! I'll tackle this in a separate PR

implementations/rust/ockam/ockam_app/src/invitations/tray_menu.rs
adrianbenavides: I've removed the TODO's blocks 👍

etorreborre (approved):

PR #5973, opened 2023-09-14
update nix lockfile and remove nix update workflow

metaclips:
This PR updates the Nix lock file and also remove the workflow that updates Nix lock file as it uses multiple external GitHub actions we do not currently support https://github.com/build-trust/ockam/actions/runs/6187486431


mrinalwadhwa (approved):

mrinalwadhwa:
Looks good, let's create an issue to track adding some other way of updating flake.lock

PR #5979, opened 2023-09-15
feat(rust): app event debouncer

adrianbenavides:
If I understand [Tauri's docs](https://tauri.app/v1/references/architecture/process-model/#the-core-process) correctly, the backend (rust part) runs as a single process (core process). Even though we `spawn` tasks to process background operations, some times the core process gets stuck if one of those background tasks takes longer than expected (couldn't reach the orchestrator, for example). 

Also, since we process most of the events as spawned tasks, multiple events can get processed concurrently, and some times, the same event gets fired up multiple times in a short period of time. 

This PR implements an event debouncer to prevent the same event from being processed multiple times concurrently and free up the core process from useless background tasks.

Here's a non sintetic example of debounced events while going through the enrolling process:

Screenshot 2023-09-19 at 15 00 05

### DIFF
--- a/implementations/rust/ockam/ockam_app/src/app/state/mod.rs
+++ b/implementations/rust/ockam/ockam_app/src/app/state/mod.rs
@@ -3,12 +3,15 @@ mod repository;
 
 #[cfg(debug_assertions)]
 use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering;
 use std::sync::Arc;
+use std::sync::RwLock as StdRwLock;
 
 use miette::IntoDiagnostic;
 use ockam_multiaddr::MultiAddr;
 use tauri::async_runtime::{block_on, spawn, RwLock};
-use tracing::{error, info, warn};
+use tauri::{AppHandle, Manager, Runtime};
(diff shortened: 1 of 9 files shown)


implementations/rust/ockam/ockam_app/src/app/state/mod.rs
```diff
@@ -331,3 +367,69 @@ fn load_model_state(
         }
     })
 }
+
+pub type EventName = String;
+type IsProcessing = AtomicBool;
+
+struct EventManager {
+    events: HashMap<EventName, IsProcessing>,
```
davide-baldo: We can simplify using a Vec<>

davide-baldo (dismissed):

implementations/rust/ockam/ockam_app/src/app/state/mod.rs
adrianbenavides: You mean storing a tuple in the vector for each item? `Vec<(EventName, IsProcessing)>`

davide-baldo (approved):

PR #5994, opened 2023-09-18
feat(rust): add menu items to connect to/disconnect from an invited service

### LINKED ISSUE #5988: TCP inlet port should be persisted and attached to the service so that we can reestablish the connection using the same socket address

adrianbenavides:
Fixes https://github.com/build-trust/codebase/issues/576 and fixes https://github.com/build-trust/ockam/issues/5988

- persist tcp inlets data to reuse the same address when recreating them
- add menu items to manually connect/disconnect to/from an invited service

### DIFF
--- a/implementations/rust/ockam/ockam_app/src/invitations/commands.rs
+++ b/implementations/rust/ockam/ockam_app/src/invitations/commands.rs
@@ -15,7 +15,7 @@ use ockam_api::nodes::models::portal::InletStatus;
 
 use crate::app::{AppState, PROJECT_NAME};
 use crate::cli::cli_bin;
-use crate::invitations::state::{InvitationState, TcpInlet};
+use crate::invitations::state::{Inlet, InvitationState};
 use crate::projects::commands::{create_enrollment_ticket, SyncAdminProjectsState};
 use crate::shared_service::relay::RELAY_NAME;
 
@@ -153,7 +153,7 @@ async fn refresh_inlets(
 
     let cli_state = app_state.state().await;
     let cli_bin = cli_bin()?;
-    let mut inlets_socket_addrs = vec![];
+    let mut running_inlets = vec![];
     for invitation in &invitations_state.accepted.invitations {
         match InletDataFromInvitation::new(
             &cli_state,
@@ -161,13 +161,12 @@ async fn refresh_inlets(
             &invitations_state.accepted.inlets,
         ) {
             Ok(i) => match i {
-                Some(i) => {
+                Some(mut i) => {
                     if !i.enabled {
                         debug!(node = %i.local_node_name, "TCP inlet is disabled by the user, skipping");
                         continue;
                     }
 
-                    let mut inlet_is_running = false;
                     debug!(node = %i.local_node_name, "Checking node status");
                     if let Ok(node) = cli_state.nodes.get(&i.local_node_name) {
                         if node.is_running() {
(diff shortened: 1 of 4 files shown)


implementations/rust/ockam/ockam_app/src/invitations/commands.rs
```diff
@@ -130,35 +133,40 @@ pub async fn refresh_invitations<R: Runtime>(app: AppHandle<R>) -> Result<(), St
...
-        .await
-        .map_err(|e| e.to_string())?;
     app.trigger_global(REFRESHED_INVITATIONS, None);
     Ok(())
 }
 
-async fn refresh_inlets<R: Runtime>(
-    app: &AppHandle<R>,
-    accepted_invitations: Option<&Vec<InvitationWithAccess>>,
+async fn refresh_inlets(
```
etorreborre, line 144: The functional programmer in me is not a big fan of passing guards to functions. I'd rather have a `refresh_inlets` function which takes `AppState`, `InvitationState` and returns `InvitationState`
```rust
...
        let invitation_state: State<'_, SyncInvitationsState> = app.state();
        let mut invitation_state = invitation_state.write().await;
        invitation_state.replace_by(invitations.clone());
        *invitation_state = refresh_inlets(state.inner(), invitation_state.clone())
            .await
            .map_err(|e| e.to_string())?;
    }
    app.trigger_global(REFRESHED_INVITATIONS, None);
    Ok(())
}

async fn refresh_inlets(
    app_state: &AppState,
    invitations_state: InvitationState,
) -> crate::Result<InvitationState> {
....
```
That is at the expense of a `.clone` but that opens up the door to unit testing something like `refresh_inlets` (if we can get rid of the external call to `ockam` one day).

implementations/rust/ockam/ockam_app/src/invitations/commands.rs
```diff
@@ -182,21 +190,16 @@ async fn refresh_inlets<R: Runtime>(
...
-                                    .push((invitation.invitation.id.clone(), inlet_socket_addr));
+                                running_inlets.push((invitation.invitation.id.clone(), i));
+                                continue;
                             }
                         }
                     }
-                    if inlet_is_running {
-                        continue;
-                    }
                     debug!(node = %i.local_node_name, "Deleting node");
```
etorreborre, line 203: I would extract this code to a `delete_inlet` function since there is already a `create_inlet`. Even better would be to have a trait encapsulating all the command line operations so that we can just provide another implementation with direct calls to the `NodeManager` (or an equivalent) one day.

(same thing for the code doing the `show inlet`)

implementations/rust/ockam/ockam_app/src/invitations/commands.rs
```diff
@@ -210,9 +213,9 @@ async fn refresh_inlets<R: Runtime>(
                     .stdout_capture()
                     .run();
                     match create_inlet(&i).await {
-                        Ok(inlet_socket_addr) => {
-                            inlets_socket_addrs
-                                .push((invitation.invitation.id.clone(), inlet_socket_addr));
+                        Ok(socket_addr) => {
```
etorreborre, line 216: I think that there are plenty of opportunities to refactor that code and the code on line 234 by having better entities. Maybe having a direct representation called `TheirService` which would contain invitation information (who? what?) and inlet data (socket address). 

And having `Services` instead of `InvitationState` where there's an interface with a method like `service_invitation_accepted(invitation_id: String,  created_inlet: InletStatus)` which would update the internal data: remove the received invitation and create a new service inside a `their_services` field.

What I would be aiming for would be a better separation of interaction with the external world and updates to the internal state in a more encapsulated way (right now we expose bare maps).

In terms of vocabulary we could have:

 - `SentInvitation { recipient_email: String, service_outlet_name: String....}`
 - `ReceivedInvitation { sender_email: String, ....}`
 - `ServiceOutlet { name: String,  , socket_addr: SocketAddr, node_name: String, connected: bool }` 
 - `ServiceInlet { name: String, sender_email: String, socket_addr: SocketAddr, node_name: String, connected: bool }`

implementations/rust/ockam/ockam_app/src/invitations/commands.rs
```diff
@@ -305,23 +312,83 @@ async fn create_inlet(inlet_data: &InletDataFromInvitation) -> crate::Result<Soc
...
     pub service_route: String,
     pub enrollment_ticket_hex: Option<String>,
+    pub socket_addr: Option<SocketAddr>,
 }
 
 impl InletDataFromInvitation {
     pub fn new(
         cli_state: &CliState,
         invitation: &InvitationWithAccess,
+        inlets: &HashMap<String, Inlet>,
```
etorreborre, line 391: This constructor feels wrong to me. That function should probably somewhere else, and, as a result, return a new `ServiceInlet`, ready to be added to the list of services.

etorreborre (approved):
While this is all working I think there are opportunities for refactoring along the lines of:

 - separating external effects (command line, cli state) from internal updates
 - clearing up the vocabulary
 - encapsulating state changes for the `InvitationsState`

I'll try to give it a go if that can be put on my roadmap.

PR #6002, opened 2023-09-18
retry flaky bats test

metaclips:
Retry flakey ockam command bats test using [BATS_TEST_RETRIES special variable](https://bats-core.readthedocs.io/en/stable/writing-tests.html#special-variables)


adrianbenavides (approved):
Nice!

PR #6013, opened 2023-09-20
chore(elixir): simplify CI job for elixir

polvorin:
We need to run the test suite and linter, a separate "build" step doesn't add anything. It takes way longer to setup the nix environment than to run the actual tests/linting, it doesn't make sense to have then split in parallel.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -52,83 +52,16 @@ defaults:
     shell: nix develop ../../tools/nix#elixir --command bash {0}
 
 jobs:
-  lint:
-    name: Elixir - lint_${{ matrix.mix_project }}
-    runs-on: ubuntu-22.04
-    strategy:
-      fail-fast: false
-      matrix:
-        mix_project:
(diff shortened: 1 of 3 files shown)


.github/workflows/elixir.yml
```diff
@@ -52,83 +52,18 @@ defaults:
...
-      - uses: ./.github/actions/nix_installer
-      - uses: ./.github/actions/elixir_cache
-        with:
-          job_name: build_${{ matrix.mix_project }}
-      - run: make build_${{ matrix.mix_project }}
-        working-directory: implementations/elixir
-
   test:
-    name: Elixir - test_${{ matrix.mix_project }}
+    name: Elixir - Test
```
mrinalwadhwa: Let's rename this to `Elixir - test` so its matched the small case naming convention of all our other checks.

metaclips: Done

mrinalwadhwa (approved):

PR #6023, opened 2023-09-21
fix(rust): app's window size on monitors with a scale factor > 1

### LINKED ISSUE #6016: The window height of the service sharing screen is too short when display is scaled
## Current behavior

The desktop app height is short. And the buttons on the bottom aren't visible.
image

On macOS when display scaling is enabled so that it isn't 1-1 virtual pixel to actual pixel.
image

## Expected behavior

The buttons should be visible. Perhaps increase the height of the window.

adrianbenavides:
### Scale factor 1

Note the size differences in the placeholders and buttons.

image

### Scale factor 2

Note the scrolling bar on the "Share a service" window.

image

## Solution

This is how it looks now in both scale factors.

Screenshot 2023-09-21 at 10 06 37


etorreborre (approved):
🥇 Nice!

PR #6024, opened 2023-09-21
feat(rust): move "create service" menu item to the "your services" section

adrianbenavides:
## Before

Screenshot 2023-09-20 at 20 27 01

## Now

Screenshot 2023-09-21 at 10 26 10


etorreborre (approved):
Just reading the title below "Shared services with you". It seems to me that "Services shared with you" would read better.

davide-baldo (approved):

nazmulidris:
@etorreborre Good catch. I will update the UX diagram & docs w/ this 👍🏽

PR #6031, opened 2023-09-21
refactor(rust): add trait to interact with background nodes from the app

adrianbenavides:
Apply the refactors proposed by @etorreborre in the following comments:
1. https://github.com/build-trust/ockam/pull/5994#discussion_r1329728611
2. https://github.com/build-trust/ockam/pull/5994#discussion_r1329732055

It also fixes a bug where existing TCP inlets where failing to be created after restarting the app.


etorreborre (dismissed):
That's great. I think this will mesh well with my current work so that we can eventually get rid of command line calls

etorreborre (approved):

PR #6036, opened 2023-09-22
Authority API

SanjoDeundiak:
1. Previously, all Authority members were in isolated scopes, given who was their enrollers. Therefore, an enroller could list/delete only members that they added. This stops us from implementing certain Ockam App functionality. Now scopes are eliminated and each enroller can see list all members and delete any member.
2. Remove legacy code
3. Small refactoring


SanjoDeundiak:
Blocked by minor changes @polvorin needs to make on the Cloud side

implementations/rust/ockam/ockam_api/src/authority_node/authority.rs
```diff
@@ -114,7 +114,7 @@ impl Authority {
         }
 
         let direct = crate::authenticator::direct::DirectAuthenticator::new(
-            configuration.clone().trust_context_identifier(),
+            configuration.clone().project_identifier(),
```
etorreborre: I thought that this was the other way around and that `project_identifier` was actually subsumed by `trust_context_identifier` now? @polvorin?

implementations/rust/ockam/ockam_api/src/authenticator/direct/authenticator.rs
```diff
@@ -0,0 +1,132 @@
...
+            attributes_writer,
+            attributes_reader,
+        })
+    }
+
+    async fn add_member<'a>(
+        &self,
+        enroller: &Identifier,
+        id: &Identifier,
+        attrs: &HashMap<CowStr<'a>, CowStr<'a>>,
```
etorreborre, line 38: Do we really benefit from `CowStr` instead of just consuming `Strings` here?

implementations/rust/ockam/ockam_api/src/authenticator/enrollment_tokens/authenticator.rs
```diff
@@ -12,6 +12,7 @@ pub(super) const MAX_TOKEN_DURATION: Duration = Duration::from_secs(600);
 #[derive(Clone)]
 pub struct EnrollmentTokenAuthenticator {
     pub(super) trust_context: String,
+    // TODO: Replace with something sane and standard + implement expiration
```
etorreborre, line 15: @nazmulidris that looks that a good issue to create

implementations/rust/ockam/ockam_api/tests/authority.rs
```diff
@@ -0,0 +1,390 @@
...
+
+    let sc = secure_channels
+        .create_secure_channel(ctx, &member, route!["api"], SecureChannelOptions::new())
+        .await?;
+
+    let member_client = DirectAuthenticatorClient::new(
+        RpcClient::new(route![sc, DefaultAddress::DIRECT_AUTHENTICATOR], ctx).await?,
+    );
+
+    let timeout = Arc::new(AtomicBool::new(true));
```
etorreborre, line 65: The name "timeout" is weird here. It's more like `list_is_accessible_for_non_admin` which should not switch from `false` to `true`

implementations/rust/ockam/ockam_api/tests/authority.rs
```diff
@@ -0,0 +1,390 @@
...
+#[ockam_macros::test]
+async fn test_one_admin(ctx: &mut Context) -> Result<()> {
+    let secure_channels = secure_channels();
+
+    let admins = setup(ctx, secure_channels.clone(), 1).await?;
+    let admin = &admins[0];
+
+    let now = now()?;
+
+    let members = admin.client.list_member_ids().await?;
```
etorreborre, line 90: I know this is obvious to you but it would be good to have comments indicating what is the intention here. Like "An admin is also a member of a given project", "the attributes of a admin do not expire", etc...

implementations/rust/ockam/ockam_api/tests/authority.rs
```diff
@@ -0,0 +1,390 @@
...
+        .client
+        .add_member(member1.clone(), attributes1)
+        .await?;
+
+    admin2
+        .client
+        .add_member(member2.clone(), attributes2)
+        .await?;
+
+    let mut members1 = admin1.client.list_member_ids().await?;
```
etorreborre, line 225: The important point that should underlined here is that each admin can see all the members.

implementations/rust/ockam/ockam_api/tests/authority.rs
```diff
@@ -0,0 +1,390 @@
...
+use ockam_node::{Context, RpcClient};
+use rand::{thread_rng, Rng};
+use std::collections::BTreeMap;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+use tempfile::NamedTempFile;
+use tracing::log::warn;
+
+async fn default_configuration() -> Result<Configuration> {
```
etorreborre: As a general note, as a reader of tests, I like to see:

 - the tests first where the name reveals the intention or the context
 - the implementation details further down the file (otherwise I need read / analyze structs and functions before even know how / why they will be used)
 - some comments telling me what we are trying to test exactly, what's the rule that we are checking, and what's our strategy for testing it if it is a bit involved

Otherwise our tests serve a non-regression protection but not really as a documentation of the expected behaviour.

etorreborre (dismissed):
I left a few comments but my main concern is about the use of `project_id` instead of `trust_context_id` because I thought that TrustContexts were generalizing the authority data that was previously available through projects only.

implementations/rust/ockam/ockam_api/src/authenticator/direct/authenticator.rs
SanjoDeundiak: We don't, but I just moved that code

implementations/rust/ockam/ockam_api/src/authority_node/authority.rs
SanjoDeundiak: The credential will have TrustContextId field, which, in this case, equals to the ProjectId.

implementations/rust/ockam/ockam_api/src/authenticator/enrollment_tokens/authenticator.rs
SanjoDeundiak: Do you think we should first move to SQLite? Or maybe not?

implementations/rust/ockam/ockam_api/tests/authority.rs
SanjoDeundiak: The problem is that we don't have a proper error handling for such case (request is just dropped at Incoming ABAC), therefore this call if unsuccessful - just times out. Hence the name.

implementations/rust/ockam/ockam_api/src/authority_node/authority.rs
etorreborre: I see

implementations/rust/ockam/ockam_api/tests/authority.rs
etorreborre: I understand the implementation :-). I think it would read better with a different name. Not a big deal if you want to leave it like that.

implementations/rust/ockam/ockam_api/tests/authority.rs
SanjoDeundiak: Fixed

implementations/rust/ockam/ockam_api/tests/authority.rs
SanjoDeundiak: I hope fixed

etorreborre (approved):

PR #6071, opened 2023-09-25
fix(rust): reset cli state if it can't be parsed

### LINKED ISSUE #6047: Investigate why command manual generation is failing.
This script to generate our manual is failing
https://github.com/build-trust/ockam-documentation/blob/command/manual/template/mdbook/build.sh

```
thread 'main' panicked at 'Failed to load the local Ockam configuration: Serde(Error("data did not match any variant of untagged enum IdentityConfigs", line: 0, column: 0))', implementations/rust/ockam/ockam_command/src/lib.rs:241:44
```

### LINKED ISSUE #6069: Can't run `reset` with latest version (0.94.0)
### Observed behavior

I've just upgraded to the latest release, and I'm unable to run any commands. I assume it's something to do with schema differences between versions. The release notes call out that you'll need to `ockam reset` for the new version, but it's not possible to run even that command after upgrading: 

```shell
❯ RUST_BACKTRACE=1 ockam reset
thread 'main' panicked at 'Failed to load the local Ockam configuration: Serde(Error("data did not match any variant of untagged enum IdentityConfigs", line: 0, column: 0))', implementations/rust/ockam/ockam_command/src/lib.rs:241:44
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: ockam_command::CommandGlobalOpts::new
   4: ockam_command::OckamCommand::run
   5: ockam_command::run
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Running `rm -rf ~/.ockam/` first resolves the issue.

### Steps to reproduce

1. Have an existing configuration from a previous version
2. Upgrade to v0.94.0
3. Run `ockam reset`

### Expected behavior

`reset` to always work to allow someone to get back to a working state

### Ockam Version

0.94.0

adrianbenavides:
## Problem description

In out last release (0.93.0, 0.94.0) introduced some breaking changes related to identities with no possible migration strategy. This leads to the following UX problems:

1. The user is forced to run `ockam reset` to remove any previous configuration, so that it can be regenerated with the new format
2. The `ockam reset` will try to parse the previous configuration, but, because there is no migration, it will fail, forcing the user to manually delete the ockam directory (`$OCKAM_HOME`, `~/.ockam`)

## Proposed solution

1. The `CliState`'s `delete` method has been refactored to be static, so that we don't need to initialize it first before removing it.

2. In both the CLI and the App, when it fails to initialize the CliState (this will happen if the configuration has been tampered or if we introduced breaking changes without migrations), it will backup the current configuration and do an automatic reset. In the CLI, a message will be written to stderr to inform them of such operation.

## How to test

From `develop` run:

```
ockam identity create
```

Now, tamper the identity config json file. For example, change the `identifier` field to `identifier2`. Now run:

```
ockam status
```

It should fail with the error described at https://github.com/build-trust/ockam/issues/6069. Switch to this PR's branch and run the same command again. You should see a message like the following and everything should work as expected:

```
The ".ockam" directory has been reset and has been backed up to ".ockam.bak"
```

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/credentials.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/credentials.rs
@@ -58,6 +58,7 @@ mod traits {
     use crate::cli_state::file_stem;
     use crate::cli_state::traits::*;
     use ockam_core::async_trait;
+    use std::path::Path;
 
     #[async_trait]
     impl StateDirTrait for CredentialsState {
@@ -66,8 +67,10 @@ mod traits {
         const DIR_NAME: &'static str = "credentials";
         const HAS_DATA_DIR: bool = false;
 
-        fn new(dir: PathBuf) -> Self {
-            Self { dir }
+        fn new(root_path: &Path) -> Self {
+            Self {
+                dir: Self::build_dir(root_path),
+            }
         }
 
         fn dir(&self) -> &PathBuf {
--- a/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
@@ -223,8 +223,10 @@ mod traits {
         const DIR_NAME: &'static str = "identities";
         const HAS_DATA_DIR: bool = true;
 
-        fn new(dir: PathBuf) -> Self {
-            Self { dir }
+        fn new(root_path: &Path) -> Self {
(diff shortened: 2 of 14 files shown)


etorreborre (approved):
Excellent idea

PR #6093, opened 2023-09-26
feat(elixir): share elixir build dir to speedup CI testing

polvorin:
keeps the CI test task time in line while not depending on precompiled ffi binaries. It's a middle ground of using elixir' umbrella project that I was afraid of switching at this point


mrinalwadhwa (dismissed):

mrinalwadhwa:
Hmm, something may have been broken by this PR https://github.com/build-trust/ockam/pull/6086
Even though all the checks on that PR passed.

Might be because now have 2 different Cargo.locks and dependabot is only maintaining one of them.

https://github.com/build-trust/ockam/actions/runs/6306571341/job/17121911999?pr=6093

polvorin:
> Hmm, something may have been broken by this PR #6086 Even though all the checks on that PR passed.
> 
> Might be because now have 2 different Cargo.locks and dependabot is only maintaining one of them.
> 
> https://github.com/build-trust/ockam/actions/runs/6306571341/job/17121911999?pr=6093

yes exactly, and it wasn't detected because the nif was not compiled.   This won't update the cargo dep automatically,  but at least will break and we will notice (for example if some rust api the nif depends on changes).
The rust part of the nif could be moved to the implementations/rust  folder but that require some investigation on how to handle properly

mrinalwadhwa:
> The rust part of the nif could be moved to the implementations/rust folder but that require some investigation on how to handle properly

Let's file a separate issue for this.

mrinalwadhwa (approved):

PR #6094, opened 2023-09-26
Vault type-safety

SanjoDeundiak:
1. Introduce more granular types for Vault
2. Propagate these types to everything that was dependent on Vault


implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -113,27 +190,28 @@ fn create_identity(env: Env, existing_key: Option<String>) -> NifResult<(Binary,
 fn attest_secure_channel_key<'a>(
     env: Env<'a>,
     identifier: String,
-    secret: Binary,
+    secret: Binary, // TODO: PublicKey is enough here
```
SanjoDeundiak, line 193: @polvorin

implementations/rust/ockam/ockam_command/src/output/output.rs
```diff
@@ -434,15 +436,15 @@ impl fmt::Display for X25519PublicKeyDisplay {
     }
 }
 
-pub struct Ed25519PublicKeyDisplay(pub Ed25519PublicKey);
+pub struct Ed25519PublicKeyDisplay(pub EdDSACurve25519PublicKey);
```
etorreborre, line 439: Why do we need newtypes to display public keys? Do we already have several ways to display the same type of public key? If not, maybe we can introduce a newtype only when this happens?

implementations/rust/ockam/ockam_identity/src/purpose_keys/builder/common.rs
```diff
@@ -0,0 +1,28 @@
+use crate::utils::now;
+use crate::TimestampInSeconds;
+use ockam_core::Result;
+
+pub(crate) enum Ttl {
+    CreatedNowWithTtl(TimestampInSeconds),
```
etorreborre, line 6: Isn't the value inside `CreatedNowWihTtl` a `Duration` and not a timestamp? This is a reason where I'd rather use `DateTimes` and not timestamps in our code

implementations/rust/ockam/ockam_identity/src/purpose_keys/builder/credential.rs
```diff
@@ -0,0 +1,121 @@
+use ockam_core::compat::sync::Arc;
+use ockam_core::Result;
+use ockam_vault::{SigningKeyType, SigningSecretKeyHandle};
+
+use crate::models::{PurposePublicKey, TimestampInSeconds};
+use crate::purpose_keys::Ttl;
+use crate::{CredentialPurposeKey, Identifier, Purpose, PurposeKeyCreation};
+
+/// Default TTL for an Identity key
+pub const DEFAULT_CREDENTIAL_PURPOSE_KEY_TTL: TimestampInSeconds =
```
etorreborre, line 10: This would be better represented as a `Duration`

implementations/rust/ockam/ockam_identity/src/purpose_keys/builder/credential.rs
```diff
@@ -0,0 +1,121 @@
...
+/// Builder for [`CredentialPurposeKey`]
+pub struct CredentialPurposeKeyBuilder {
+    purpose_keys_creation: Arc<PurposeKeyCreation>,
+
+    identifier: Identifier,
+    key: Key,
+    ttl: Ttl,
+}
+
+impl CredentialPurposeKeyBuilder {
```
etorreborre, line 27: I am annoyed that the code in `CredentialPurposeKeyBuilder` and `SecureChannelPurposeKeyBuilder` expose much of the internals of `PurposeKeyCreation`.

An alternative would be to define:

 - `CredentialPurposeKeyOptions`
 - `SecureChannelPurposeKeyOptions`

and have 2 methods on `PurposeKeysCreation`:

 - `create_credential_purpose_key(&self, options: CredentialPurposeKeyOptions)`
 - `create_secure_channel_purpose_key(&self, options: SecureChannelPurposeKeyOptions)`

This way we keep the type safety, still hide the internals of `PurposeKeysCreation`, and we can even factor out a few lines of code between the 2 implementations.

implementations/rust/ockam/ockam_identity/src/purpose_keys/purpose_key_creation.rs
```diff
@@ -0,0 +1,313 @@
...
+        Self {
+            vault,
+            identities_reader,
+            identity_keys,
+            repository,
+        }
+    }
+
+    /// Return [`PurposeKeysRepository`] instance
+    pub fn repository(&self) -> Arc<dyn PurposeKeysRepository> {
```
etorreborre, line 40: If we move the implementation in `CredentialsPurposeKeyBuilder` and `SecureChannelsPurposeKeyBuilder` internal to `PurposeKeysCreation` I think we can remove:

 - `repository`
 - `purpose_keys_verification`
 - `secure_channel_purpose_key_builder`
 - `credential_purpose_key_builder`
 - `vault`

This limits the possibility that these elements, in particular `repository` and `vault` are accessed via too many different code paths.

implementations/rust/ockam/ockam_vault/src/lib.rs
```diff
@@ -46,8 +45,43 @@ mod software;
...
+
+/// Feature set compatibility checks
+
+#[cfg(all(
+    feature = "disable_default_noise_protocol",
+    not(feature = "OCKAM_XX_25519_AES256_GCM_SHA256"),
+    not(feature = "OCKAM_XX_25519_ChaChaPolyBLAKE2s"),
+    not(feature = "OCKAM_XX_25519_AES128_GCM_SHA256")
+))]
+compile_error! {"NOISE protocol name not selected, please enable on of the following features: \"OCKAM_XX_25519_ChaChaPolyBLAKE2s\", \"OCKAM_XX_25519_AES128_GCM_SHA256\", \"OCKAM_XX_25519_AES256_GCM_SHA256\""}
```
etorreborre: s/please enable on/please enable one

implementations/rust/ockam/ockam_vault/src/software/legacy/secret.rs
```diff
@@ -23,24 +25,14 @@ impl<'de> Deserialize<'de> for Secret {
         D: Deserializer<'de>,
     {
         #[derive(Deserialize)]
-        pub enum SecretV1 {
-            Key(#[serde(with = "hex_encoding")] SecretKeyVec),
-            Aws(KeyId),
-        }
-        #[derive(Deserialize)]
-        struct SecretV2(#[serde(with = "hex_encoding")] SecretKeyVec);
+        struct SecretV2(#[serde(with = "hex_encoding")] Vec<u8>);
```
etorreborre, line 28: Do we even need `Secrets` if there's only V2 now?

implementations/rust/ockam/ockam_vault/src/lib.rs
etorreborre: I think it would be more explicit/direct to have 3 exclusive features and declare `OCKAM_XX_25519_AES256_GCM_SHA256` as the default one). Otherwise we have to think in terms of double negation later in the code: 
```
if #[cfg(any(not(feature = "disable_default_noise_protocol")
```

But maybe I missed something that makes this approach problematic?

One thing I find unfortunate is that there's no `oneof` exclusive choice so we still have to write
```
#[cfg(not(any(
    feature = "OCKAM_XX_25519_AES256_GCM_SHA256",
    feature = "OCKAM_XX_25519_AES128_GCM_SHA256"
    feature = "OCKAM_XX_25519_ChaChaPolyBLAKE2s",
    )))]
compile_error! {"please select exactly one noise protocol"}

#[cfg(all(
    feature = "OCKAM_XX_25519_AES256_GCM_SHA256",
    any(
        feature = "OCKAM_XX_25519_ChaChaPolyBLAKE2s",
        feature = "OCKAM_XX_25519_AES128_GCM_SHA256"
    ))]
compile_error! {"only one protocol can be selected"}

#[cfg(all(
    feature = "OCKAM_XX_25519_AES128_GCM_SHA256"
    any(
        feature = "OCKAM_XX_25519_ChaChaPolyBLAKE2s",
        feature = "OCKAM_XX_25519_AES256_GCM_SHA256",
    ))]
compile_error! {"only one protocol can be selected"}

#[cfg(all(
    feature = "OCKAM_XX_25519_ChaChaPolyBLAKE2s"
    any(
        feature = "OCKAM_XX_25519_AES128_GCM_SHA256",
        feature = "OCKAM_XX_25519_AES256_GCM_SHA256",
    ))]
compile_error! {"only one protocol can be selected"}
```

implementations/rust/ockam/ockam_vault/src/lib.rs
```diff
@@ -46,8 +45,43 @@ mod software;
...
+    not(feature = "disable_default_noise_protocol"),
+    any(
+        feature = "OCKAM_XX_25519_AES256_GCM_SHA256",
+        feature = "OCKAM_XX_25519_ChaChaPolyBLAKE2s",
+        feature = "OCKAM_XX_25519_AES128_GCM_SHA256"
+    )
+))]
+compile_error! {"please enable disable_default_noise_protocol feature to customize Noise protocol"}
+
+#[cfg(all(
```
etorreborre, line 73: The conditions here are the same than the ones on L63. I think that the fact that exactly one protocol is enabled has to be encoded using my comment above.

implementations/rust/ockam/ockam_vault/src/types/secrets.rs
```diff
@@ -0,0 +1,58 @@
+use ockam_core::compat::vec::Vec;
+
+/// A handle to a secret inside a vault.
```
etorreborre, line 3: I could be a good idea to give examples for what kind of bytes will end up being a secret handle

etorreborre (dismissed):
The additional types are great!

The 2 parts that I am bit more uncertain about are the `PurposeKeysCreation` and the use of a `disable` feature flag

implementations/rust/ockam/ockam_command/src/output/output.rs
SanjoDeundiak: I would like to avoid pulling Display related things to `ockam_identity` crate, I think it's `ockam_command` specific how we would like to print it. Therefore, introducing new type is the way to imply this behaviour on the type from another crate

implementations/rust/ockam/ockam_identity/src/purpose_keys/builder/common.rs
SanjoDeundiak: The motivation behind this is that on-the-wire it eventually becomes the `TimestampInSeconds`. We could introduce a separate type for duration or use `Duration` type from `core`. It's an additional layer of abstraction (therefore, requiring conversions between the two) without clear benefit to me. I would like to keep it simple

implementations/rust/ockam/ockam_identity/src/purpose_keys/builder/credential.rs
SanjoDeundiak: See above

implementations/rust/ockam/ockam_identity/src/purpose_keys/builder/credential.rs
SanjoDeundiak: What you suggest is exactly what I attempted initially (and succeeded). But because type-safety increased the separation between two types of PurposeKeys, it caused a lot of code duplications between the types which were hard to reuse (this version was bigger by around 300-400 lines of code). I decided that the way it is now is better, with its cons.

implementations/rust/ockam/ockam_vault/src/software/legacy/secret.rs
SanjoDeundiak: I didn't want to mess with it, so just removed the old type so that it compiles. All of it will go away since we don't want to store secrets in the file at all

implementations/rust/ockam/ockam_vault/src/types/secrets.rs
SanjoDeundiak: That is implementation-specific. It's already 2 different things for AWS and Software implementations

SanjoDeundiak: Basically it's an arbitrary vector of bytes that would help a Vault to identify the secret, we don't have any restrictions on what it is

implementations/rust/ockam/ockam_vault/src/lib.rs
SanjoDeundiak: The problem is that we use `default-features = false` everywhere, and your approach would require explicitly setting the default protocol each time we have a dependency on `ockam_vault` (therefore, that decision is spread into multiple places). And when we have a transitive dependency, like `X -> ockam_identity -> ockam_vault`, that would imply now having the same features existing on `ockam_identity` crate. With my approach it's much easier to depent on `ockam_vault` from other crates in case of default Noise protocol. I haven't tried custom protocols, since they're not yet implemented anyways, but I suspect that my approach is simpler in that case as well. If that won't be true, we can change that when we get there

implementations/rust/ockam/ockam_vault/src/lib.rs
SanjoDeundiak: Nice catch, it looks like we need to go with your long check

implementations/rust/ockam/ockam_identity/src/purpose_keys/purpose_key_creation.rs
SanjoDeundiak: Could you please elaborate?

implementations/rust/ockam/ockam_command/src/output/output.rs
etorreborre: I see. I also think that displaying data for `command` should have even its own type class.

implementations/rust/ockam/ockam_identity/src/purpose_keys/purpose_key_creation.rs
etorreborre: I'd rather enforce a separation between interfaces and implementations. If we create a struct with a certain purpose (no pun intended) like `PurposeKeysCreation` it should be able to do things via its public methods without have to expose its internals like `repository` or `vault`. Especially in the case of `vault` where we reach even further to use a more specific vault.

implementations/rust/ockam/ockam_vault/src/types/secrets.rs
etorreborre: It's ok that you explain that to me here, but I think that would be better in a comment :-). In particular because a reader might wonder why this is not a simple `String`

implementations/rust/ockam/ockam_identity/src/purpose_keys/builder/common.rs
etorreborre: I still think that the right level of abstraction in all the code is `Duration` (and `DateTime` in other places). The fact that both can be serialized as a number of seconds should only be a serialization concern.

etorreborre (approved):

implementations/rust/ockam/ockam_identity/src/purpose_keys/purpose_key_creation.rs
SanjoDeundiak: Do you mean you would prefer to make those dependencies inaccessible outside of the structure?

implementations/rust/ockam/ockam_vault/src/types/secrets.rs
SanjoDeundiak: I'll add that comment in the next PR.
P.S. Phrase `simple String` triggered me :D, `Vec<u8>` is what simpler than `String`)

PR #6152, opened 2023-09-28
fix: confirm terminating ockam enroll

### LINKED ISSUE #6061: Make `ockam enroll` project configuration step more robust
## Current behavior
When a user runs `ockam enroll` and they don't have a project already created in their account on Ockam Orchestrator, then it might take ~ 2 minutes for this process to complete. During this time `ockam enroll` displays a [progress spinner](https://docs.rs/indicatif/latest/indicatif/struct.ProgressBar.html#method.new_spinner) and "Configuring project..." text in the terminal. Here's a [video at timestamp 1:04](https://asciinema.org/a/608023?t=1:04) that display this. 

This file creates a new spinner and manages display it: [rust/ockam/ockam_command/src/project/util.rs](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/project/util.rs#L157)

Image

During this time, if the user presses <kbd>Ctrl+C</kbd> then it terminates the `ockam` process. This is not desirable since it leaves the local state in an inconsistent state. And they have to recover from this by running `ockam reset -y` and then run `ockam enroll` again.

> To install Ockam command from a prebuilt binary or from source, please take a look at this [README.md](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/README.md#instructions-on-how-to-install-ockam-command).

## Desired behavior
1. If the user presses <kbd>Ctrl+C</kbd>, detect this signal and prompt them that enrollment is currently in progress, but don't stop enrollment. In the prompt mention that if they would like to exit anyway they should press <kbd>Ctrl+C</kbd> again. Here's an example of the prompt copy: "ockam enroll is still in progress. If you would like to stop the enrollment process, press Ctrl+C again".

1. If they press <kbd>Ctrl+C</kbd> again, then we should exit, but before exiting the current process, remove the local state (the equivalent of running `ockam reset -y`).

## Implementation details
- In the <kbd>Ctrl+C</kbd> handler, we could check if the enrollment process is running and prevent the process from shutting down, showing the user a message like "Press Ctrl+C again if you really want to stop the enrollment process". 
- The next time we intercept the <kbd>Ctrl+C</kbd> signal, we actually try and stop the enrollment process, do cleanup, and exit. This prevents the need for the user to run `ockam reset -y` to clean up invalid local state from an interrupted enrollment.

## Open design discussion
Here are some notes on the codebase that will be useful to figure out how to implement this:
- This is the [CTRL+C handler](https://github.com/build-trust/ockam/blob/d74d6e8ad6e4da75d3dfbb199a5da47e85dc5cbb/implementations/rust/ockam/ockam_command/src/shutdown.rs#L26).
- In the `Implementation` details above, we mention that we need to introduce new functionality to `check if the enrollment process is running`. This is not currently implemented. One approach is to storing some information, when the enrollment process is underway, as a temp file in the `~/.ockam` directory. It is then possible to check whether this temp file exists, and if so, it means that enrollment is probably in progress.
- If you choose to work on this, you can collaborate w/ the Ockam team members to figure what the best approach is to implement this.

---

adiSuper94:
Fixes #6061 

## Current behavior
Okam enroll takes ~ 2min. Exiting does not give any warning

## Proposed changes
Warn user about existing. And run `ockam reset` if user confirms exit.

## Checks

- [x] There are no Merge commits in this Pull Request. The Ockam repo maintains a linear commit history so we merge PRs by rebasing them to the develop branch. Rebasing to the latest develop branch and force pushing to your PR branch is okay.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/enroll/command.rs
+++ b/implementations/rust/ockam/ockam_command/src/enroll/command.rs
@@ -1,3 +1,7 @@
+use std::process;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+
 use clap::Args;
 use colorful::Colorful;
 use miette::{IntoDiagnostic, WrapErr};
@@ -56,6 +60,32 @@ async fn rpc(ctx: Context, (opts, cmd): (CommandGlobalOpts, EnrollCommand)) -> m
     run_impl(&ctx, opts, cmd).await
 }
 
+fn ctrlc_handler(opts: CommandGlobalOpts) {
+    let is_confirmation = Arc::new(AtomicBool::new(false));
+    ctrlc::set_handler(move || {
+        if is_confirmation.load(Ordering::Relaxed) {
+            let _ = opts.terminal.write_line(
+                format!(
+                    "\n{} Received Ctrl+C again. Cancelling {}. Please try again.",
+                    "!".red(), "ockam enroll".bold().light_yellow()
+                )
+                .as_str(),
+            );
+            process::exit(2);
+        } else {
+            let _ = opts.terminal.write_line(
+                format!(
+                    "\n{} {} is still in progress. If you would like to stop the enrollment process, press Ctrl+C again.",
+                    "!".red(), "ockam enroll".bold().light_yellow()
+                )
+                .as_str(),
+            );
+            is_confirmation.store(true, Ordering::Relaxed);
+        }
+    })
+    .expect("Error setting Ctrl-C handler");
+}
+
 async fn run_impl(
     ctx: &Context,
     opts: CommandGlobalOpts,
@@ -65,6 +95,7 @@ async fn run_impl(
         "Enrolling your default Ockam identity with Ockam Orchestrator...\n"
     ))?;
 
+    ctrlc_handler(opts.clone());
     display_parse_logs(&opts);
 
     let oidc_service = OidcService::default();


implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -50,17 +60,59 @@ impl EnrollCommand {
...
     ctx: &Context,
     opts: CommandGlobalOpts,
     _cmd: EnrollCommand,
 ) -> miette::Result<()> {
+    let touch_file_path = CliState::default_dir()?.join("enroll_touch_file");
+    set_signal_handler(opts.terminal.clone());
+    File::create(&touch_file_path)
+        .into_diagnostic()
+        .wrap_err("Could not create enroll touch file")?;
+    // Create a channel for communicating back to the main thread
```
adiSuper94: ignore this comment, copy pasta 🍝 mistake.

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -50,21 +60,62 @@ impl EnrollCommand {
...
+        }
+        false
+    }
 }
 
 async fn rpc(ctx: Context, (opts, cmd): (CommandGlobalOpts, EnrollCommand)) -> miette::Result<()> {
     run_impl(&ctx, opts, cmd).await
 }
 
+fn set_signal_handler(terminal: Terminal<TerminalStream<Term>>) {
```
adrianbenavides: You could pass here the `CommandGlobalOpts` which gives you access to both the `Terminal` and the `CliState`.

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -50,21 +60,62 @@ impl EnrollCommand {
...
+            }
+            if EnrollCommand::is_running() {
+                if let Ok(path) = CliState::default_dir() {
+                    let touch_file_path = path.join("enroll_touch_file");
+                    let _ = std::fs::remove_file(touch_file_path);
+                }
+            }
+            process::exit(2);
+        } else if !confirmation.load(std::sync::atomic::Ordering::Relaxed) {
+            let _ = terminal.write_line(
```
adrianbenavides: This could be a simple `else` right?

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -50,21 +60,62 @@ impl EnrollCommand {
...
+fn set_signal_handler(terminal: Terminal<TerminalStream<Term>>) {
+    let confirmation = Arc::new(AtomicBool::new(false));
+    ctrlc::set_handler(move || {
+        if confirmation.load(std::sync::atomic::Ordering::Relaxed) {
+            if CliState::delete().is_ok() {
+                let _ = terminal.write_line("Local Ockam configuration deleted");
+            }
+            if EnrollCommand::is_running() {
+                if let Ok(path) = CliState::default_dir() {
+                    let touch_file_path = path.join("enroll_touch_file");
```
adrianbenavides: The flag file name should be passed as a parameter of this function.

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -97,6 +148,9 @@ async fn run_impl(
             .color(OckamColor::PrimaryResource.color()),
         user_info.email
     ))?;
+    std::fs::remove_file(&touch_file_path)
+        .into_diagnostic()
+        .wrap_err("Could not delete enroll touch file")?;
```
adrianbenavides: The problem here about how the temp file is managed is that, if something fails before reaching this point, the file will not be deleted.

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -50,21 +60,62 @@ impl EnrollCommand {
...
+    })
+    .expect("Error setting Ctrl-C handler");
+}
+
 async fn run_impl(
     ctx: &Context,
     opts: CommandGlobalOpts,
     _cmd: EnrollCommand,
 ) -> miette::Result<()> {
+    let touch_file_path = CliState::default_dir()?.join("enroll_touch_file");
```
adrianbenavides: Let's call the temp file `.enroll`

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -50,21 +60,62 @@ impl EnrollCommand {
...
+    pub fn is_running() -> bool {
+        if let Ok(path) = CliState::default_dir() {
+            return Path::new(&path.join("enroll_touch_file")).exists();
+        }
+        false
+    }
 }
 
 async fn rpc(ctx: Context, (opts, cmd): (CommandGlobalOpts, EnrollCommand)) -> miette::Result<()> {
     run_impl(&ctx, opts, cmd).await
```
adrianbenavides, line 60: You could move the file create/remove logic + ctrlc handler here:

```
File::create(&touch_file_path)..
let res = run_impl(&ctx, opts, cmd).await;
std::fs::remove_file...
res
```

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -50,21 +60,62 @@ impl EnrollCommand {
...
 
+fn set_signal_handler(terminal: Terminal<TerminalStream<Term>>) {
+    let confirmation = Arc::new(AtomicBool::new(false));
+    ctrlc::set_handler(move || {
+        if confirmation.load(std::sync::atomic::Ordering::Relaxed) {
+            if CliState::delete().is_ok() {
+                let _ = terminal.write_line("Local Ockam configuration deleted");
+            }
+            if EnrollCommand::is_running() {
+                if let Ok(path) = CliState::default_dir() {
```
adrianbenavides: You can access the state dir using `opts.state.dir`.

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -50,21 +60,62 @@ impl EnrollCommand {
...
     run_impl(&ctx, opts, cmd).await
 }
 
+fn set_signal_handler(terminal: Terminal<TerminalStream<Term>>) {
+    let confirmation = Arc::new(AtomicBool::new(false));
+    ctrlc::set_handler(move || {
+        if confirmation.load(std::sync::atomic::Ordering::Relaxed) {
+            if CliState::delete().is_ok() {
+                let _ = terminal.write_line("Local Ockam configuration deleted");
+            }
```
adrianbenavides: Let's not remove anything here, as this operation will delete all the state, including the stuff that was created before the user run the `enroll` command.

implementations/rust/ockam/ockam_command/src/enroll/command.rs
adrianbenavides: Also, let's rename this to `ctrlc_handler`.

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -50,21 +60,62 @@ impl EnrollCommand {
         initialize_identity_if_default(&opts, &self.identity);
         node_rpc(rpc, (opts, self));
     }
+
+    pub fn is_running() -> bool {
```
adrianbenavides: Similarly, this method should receive the `CliState` instance as an argument, so you can get its directory as `state.dir`

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -50,21 +60,62 @@ impl EnrollCommand {
...
 }
 
 async fn rpc(ctx: Context, (opts, cmd): (CommandGlobalOpts, EnrollCommand)) -> miette::Result<()> {
     run_impl(&ctx, opts, cmd).await
 }
 
+fn set_signal_handler(terminal: Terminal<TerminalStream<Term>>) {
+    let confirmation = Arc::new(AtomicBool::new(false));
+    ctrlc::set_handler(move || {
+        if confirmation.load(std::sync::atomic::Ordering::Relaxed) {
```
adrianbenavides: Consider importing `std::sync::atomic::Ordering::Relaxed` to reduce visual noise.

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -50,21 +60,62 @@ impl EnrollCommand {
...
+        false
+    }
 }
 
 async fn rpc(ctx: Context, (opts, cmd): (CommandGlobalOpts, EnrollCommand)) -> miette::Result<()> {
     run_impl(&ctx, opts, cmd).await
 }
 
+fn set_signal_handler(terminal: Terminal<TerminalStream<Term>>) {
+    let confirmation = Arc::new(AtomicBool::new(false));
```
adrianbenavides: By convention, booleans should be named as actions so they read more naturally: `is_confirmed`

implementations/rust/ockam/ockam_command/src/enroll/command.rs
adrianbenavides: Also, the variable should have a more meaningful name: `exit_confirmation_file_path`

adrianbenavides (commented):
Hey @adiSuper94, thanks for submitting this PR. I left some comments with improvements to make the code easier to read and more consistent with the project's standards.

implementations/rust/ockam/ockam_command/src/enroll/command.rs
adiSuper94: #6061 suggests doing `ockam reset -y` when enroll is killed.
That is why I added this.

adiSuper94:
@adrianbenavides @nazmulidris 
Making these changes prompted me to go back to the drawing board to see if I am over complicating a simple fix.

The purpose of creating a touch file is to check if  `ockam enroll` is running.
The Ctrl-C handler defined in [shutdown.rs](https://github.com/build-trust/ockam/blob/d74d6e8ad6e4da75d3dfbb199a5da47e85dc5cbb/implementations/rust/ockam/ockam_command/src/shutdown.rs#L26) which is referred in the ticket #6061 is not been used here. Instead we are setting a different signal handler. So we trivially know that the `ockam enroll` is running.

So maybe we can do away with the whole touch file thing ? The `atomic bool` will tell us if ctrlc has been pressed twice.

Do we want to consider the case the user calls enroll simultaneously, from multiple terminal sessions ? 

Note: I did not make changes to the existing ctrlc handler in shutdown.rs, because it not registered for for the `ockam enroll`
 process.

adiSuper94:
Regarding the `CliState::delete()`,  I though that if `enroll` process is aborted midway, the user will have to an ockam reset anyway before retrying. And I assume doing that will erase all state changes made before `ockam enroll` ran. 
 
 If the purpose of not calling delete here is to make the user explicitly aware of delete operation, then not adding the delete call makes sense.
 
 But if that is not the case, can we just do the delete along with printing a clear message ?

adrianbenavides:
> So maybe we can do away with the whole touch file thing? The atomic bool will tell us if ctrlc has been pressed twice.

Yeah that's right, makes total sense. When we wrote the issue, we were thinking that the [shutdown.rs](https://github.com/build-trust/ockam/blob/d74d6e8ad6e4da75d3dfbb199a5da47e85dc5cbb/implementations/rust/ockam/ockam_command/src/shutdown.rs#L26) ctrl_c handler was being used in this command as well. Good catch!

I think the implementation is perfect now 👍 

> Do we want to consider the case the user calls enroll simultaneously, from multiple terminal sessions ?

> Regarding the CliState::delete(), I though that if enroll process is aborted midway, the user will have to an ockam reset anyway before retrying. And I assume doing that will erase all state changes made before ockam enroll ran.

Answering both questions, there are no operations done during the enrollment process that would break the local state. No matter where you cancel the `enroll` command, you can always run it again, or even after you've been enrolled successfully.

Perhaps, if the user cancels the `enroll` command, we should print a message to let them know that the enrollment process was not finished and that they need to run the command again. What do you think?

adiSuper94:
@adrianbenavides 
> Perhaps, if the user cancels the enroll command, we should print a message to let them know that the enrollment process was not finished and that they need to run the command again. What do you think?

Sounds right.

adrianbenavides (approved):
LGTM! Awesome PR @adiSuper94, thanks a lot 💪

PR #6169, opened 2023-09-29
fix(rust): add json output to `node list` command

### LINKED ISSUE #6128: Add `--output json` support to `ockam node list`
# Current behavior

Currently running `ockam node list --output json` produces an error like the following:

Image

It also shows a spinner before it displays the error message.

# Desired behavior

~~It should not display a spinner while it is working. (updated: sep 29 2023)~~ 

It should provide a list of node information in JSON format.

Please take a look at this [file](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/node/list.rs#L28) to get started.

---

solidiquis:
## Current behavior

This PR closes https://github.com/build-trust/ockam/issues/6128

Issue summary:

`$ ockam node list --output json` emits an error rather than a JSON-formatted list of nodes.

## Proposed changes

This PR has `NodeListOutput` derive `Serialize` from `serde` so that it can be transformed into JSON via `serde_json::to_string_pretty`.

Screen Shot 2023-09-29 at 1 46 28 AM

## Checks

- [x] There are no Merge commits in this Pull Request. The Ockam repo maintains a linear commit history so we merge PRs by rebasing them to the develop branch. Rebasing to the latest develop branch and force pushing to your PR branch is okay.
- [x] I have accepted the Ockam [Contributor License Agreement](https://github.com/build-trust/ockam-contributors/blob/main/CLA.md) by adding my Git/Github details in a row at the end of the [CONTRIBUTORS.csv](https://github.com/build-trust/ockam/blob/develop/.github/CONTRIBUTORS.csv) file in a separate pull request to the [build-trust/ockam](https://github.com/build-trust/ockam) repository. The easiest way to do this is to [edit the CONTRIBUTORS.csv](https://github.com/build-trust/ockam/edit/develop/.github/CONTRIBUTORS.csv) file in the github web UI and create a PR, this will mark the commit as verified.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/list.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/list.rs
@@ -2,6 +2,8 @@ use clap::Args;
 use colorful::Colorful;
 use indoc::formatdoc;
 use miette::Context as _;
+use miette::IntoDiagnostic;
+use serde::Serialize;
 use tokio::sync::Mutex;
 use tokio::try_join;
 
@@ -100,13 +102,22 @@ async fn run_impl(
         ));
     }
 
-    let list = opts
+    let plain = opts
         .terminal
         .build_list(&nodes, "Nodes", "No nodes found on this system.")?;
-    opts.terminal.stdout().plain(list).write_line()?;
+
+    let json = serde_json::to_string_pretty(&nodes).into_diagnostic()?;
+
+    opts.terminal
+        .stdout()
+        .plain(plain)
+        .json(json)
+        .write_line()?;
+
     Ok(())
 }
 
+#[derive(Serialize)]
 pub struct NodeListOutput {
     pub node_name: String,
     pub status: String,
--- a/implementations/rust/ockam/ockam_command/tests/bats/command-reference.bats
+++ b/implementations/rust/ockam/ockam_command/tests/bats/command-reference.bats
@@ -28,7 +28,8 @@ teardown() {
   run_success "$OCKAM" node create "$n2" --verbose
 
   run_success "$OCKAM" node list
-  assert_output --partial "Node $n1 UP"
+  assert_output --partial "\"node_name\": \"$n1\""
+  assert_output --partial "\"status\": \"Running\""
 
   run_success "$OCKAM" node stop "$n1"
   assert_output --partial "Stopped node '$n1'"


implementations/rust/ockam/ockam_command/src/terminal/mod.rs
```diff
@@ -347,6 +349,29 @@ impl<W: TerminalWriter> Terminal<W, ToStdErr> {
...
+            let json_node = serde_json::to_string_pretty(node).map_err(|e| {
+                Error::InternalError {
+                    error_message: e.to_string(),
+                    exit_code: exitcode::SOFTWARE
+                }
+            })?;
+            serialized_nodes.push(json_node);
+        }
+
+        let output = serialized_nodes.join(",\n");
```
adrianbenavides: The serialization should be done in one step to avoid any manual intervention with the serialization process. Since the input objects implement the `Serialize` trait, you could just serialize the whole input array as is.

```
pub fn build_json_list(&self, items: &[impl Serialize], empty_message: &str) -> Result<String> {
    if items.is_empty() {
        let output = format!("{}", &fmt_warn!("{empty_message}"));
        return Ok(output);
    }
    Ok(serde_json::to_string_pretty(items)?)
}
```


Furthermore, note that the `Terminal` instance knows about the output format already, so you don't need to do the check externally. Rather the `build_list` method should check the format and print the items in the expected format. You might need to update the bounds requirements of the `items` argument to include the `Serialize` trait:

```
pub fn build_list(
    &self,
    items: &Vec<impl crate::output::Output>,
    header: &str,
    empty_message: &str,
) -> Result<String> {
    match self.output_format {
        OutputFormat::Plain => {
            ..
        }
        OutputFormat::Json => {
            ..
        }
    }
)
```

adrianbenavides (commented):
Thanks for submitting this PR @solidiquis! It's in the right direction, but it could be simplified a little bit to make the external API easier to use and keep the commands cleaner.

implementations/rust/ockam/ockam_command/src/terminal/mod.rs
solidiquis: Thanks for the great input @adrianbenavides. The two I things we should consider before putting everything in `build_list` is:

1. There are a lot of types that are passed into `build_list` that do not implement `Serialize`. `CredentialOutput`, `IdentityListOutput`, and `SecureChannelListOutput`, just to name a few. So adding `Serialize` as an additional trait bound would require that I derive `Serialize` for all of those types. I'm happy to do that if you think that's appropriate.

2. The `header` argument isn't necessary it seems for the JSON output, so passing`header` into `build_list` when the output format is JSON seems a bit awkward, but it feels like a minor issue.

So all-in-all I agree with your suggestions — just want your approval before I start deriving `Serialize` for everything that is passed into `build_list`.

adrianbenavides: I was reviewing the code around building the output and I think we have a better way to do this. Since generating the json format is as easy as calling `serde_json::to_string_pretty(items)?`, we could leave the `build_list` as is and do the following:

```rust
let plain = opts
    .terminal
    .build_list(&nodes, "Nodes", "No nodes found on this system.")?;
let json = serde_json::to_string_pretty(&nodes).into_diagnostic()?;
opts.terminal
    .stdout()
    .plain(plain)
    .json(json)
    .write_line()?;
```

That way you don't have to worry about other commands. We can take care of the rest in separate PR's.

implementations/rust/ockam/ockam_command/src/node/list.rs
```diff
@@ -100,13 +102,22 @@ async fn run_impl(
...
         .terminal
         .build_list(&nodes, "Nodes", "No nodes found on this system.")?;
-    opts.terminal.stdout().plain(list).write_line()?;
+
+    let json = serde_json::to_string_pretty(&nodes).into_diagnostic()?;
+
+    opts.terminal
+        .stdout()
+        .plain(plain)
+        .json(json)
```
solidiquis, line 114: @adrianbenavides thanks for the really valuable input. It's going to really help me with my future contributions.

Lastly, do you think it's worth it for me to check the `OutputFormat` to allocate only either the plain or JSON string? It would prevent us from having to allocate both the plain and json string when only one is needed.

To do that I could either get the `OutputFormat` from `opts` -> `CommandGlobalOpts` -> `GlobalArgs` or add `pub fn output_format(&self) -> &OutputFormat` to `Terminal`.

Let me know if you think that's necessary, otherwise we can just roll with this. Thanks!

adrianbenavides: Nah, don't worry about that. In this layer (`ockam_command`) it's more important to have a clean and simple code rather than an optimized and complex solution. It's true that this part could be improved to avoid serializing the output in different formats, but that should be hidden from the commands' code, but it's out of the scope of this PR.

adrianbenavides (dismissed):
Nice, thank you so much @solidiquis, LTGM! 🚀 

As you might already noticed, there are a bunch of commands were the json output is lacking. We'll try to create the issues for those, but you can go ahead and try to fix them if you are open to that. If not, I hope you can find other issues that might be of interest to you.

See you around! 👍

mrinalwadhwa:
@adrianbenavides @solidiquis we need a rebase to develop to fix `lint_commits` failing check, since @solidiquis only recently accepted the CLA.

solidiquis:
Hey @mrinalwadhwa — just did the rebase!

@adrianbenavides thanks for the help! I'll be tackling some more JSON issues today. If there's anything else you'd like me to work on as well feel free to tag me in an issue :)

solidiquis:
I will investigate https://github.com/build-trust/ockam/actions/runs/6353623243/job/17258703955?pr=6169

adrianbenavides:
> I will investigate https://github.com/build-trust/ockam/actions/runs/6353623243/job/17258703955?pr=6169

When the stdout is non-interactive, the CLI defaults to either json or machine output, so that's why this test is failing. I think you just need to [change this line](https://github.com/build-trust/ockam/blob/dd163852c7a8cd7c906f1cec0ad3284646c7aa0b/implementations/rust/ockam/ockam_command/tests/bats/command-reference.bats#L31) to:

```
assert_output --partial "\"node_name\": \"$n1\""
assert_output --partial "\"status\": \"Running\""
```

solidiquis:
> When the stdout is non-interactive

Ahhh.. classic. Thanks for saving me a bunch of time @adrianbenavides . Just pushed your suggested changes. If you could please run the test suite again that'd be awesome 🙏

adrianbenavides:
> Just pushed your suggested changes. If you could please run the test suite again that'd be awesome 🙏

Btw, there is a small guide [here](https://github.com/build-trust/ockam/blob/b3ebd2f99b2077212a0c7eddb0bbad419e5ffbe8/implementations/rust/ockam/ockam_command/tests/bats/README.md) on how to run the bats tests.

You basically need to:
- Install the bats library as described
- Compile the CLI (`cargo build -p ockam_command`)
- Set the OCKAM env variable (I usually point it to `YOUR_PROJECT_PATH/target/debug/ockam`
- Run `bats implementations/rust/ockam/ockam_command/tests/bats` to execute the local tests

adrianbenavides (approved):
OK, now it's ready! Thanks again @solidiquis 😄

PR #6172, opened 2023-09-29
feat(rust): make `Timestamp` arithmetic operations usage safer

### LINKED ISSUE #6115: Investigate and fix - rust error panicked at 'attempt to subtract with overflow'
https://github.com/build-trust/ockam/actions/runs/6329983244/job/17191393169?pr=6110#step:6:711

Screenshot 2023-09-27 at 12 42 25 PM

SanjoDeundiak:


adrianbenavides (approved):
🚀

implementations/rust/ockam/ockam_api/tests/authority.rs
```diff
@@ -95,7 +95,7 @@ async fn one_admin_test_api(ctx: &mut Context) -> Result<()> {
     assert_eq!(members.len(), 1);
     let attrs = members.get(&admin.identifier).unwrap();
 
-    assert!(attrs.added() - now < 5.into());
+    assert!(attrs.added().abs_diff(now) < 5.into());
```
etorreborre, line 98: Shouldn't we still try to test that `attrs.added() >= now`?

SanjoDeundiak: The problem is that now is calculated after the setup, and should be actually >= added. Which was exactly my mistake here and the reason why it failed

etorreborre: I'm still not sure I understand. Why not move `now` instead?
```rust
// take now right at the start of the test
let now = now()?;

// some set-up 
let admins = setup(ctx, secure_channels.clone(), 1).await?;
...
// add some attributes
admin.client.add_member(ctx, member.clone(), attributes)

// retrieve some attributes
let members = admin.client.list_members(ctx).await.unwrap();
let attrs = members.get(&member).unwrap();

// check that the attributes have been added after the storage has been created
assert!(attrs.added() >= now);
```
Isn't the last assertion what we are really after, rather than an absolute difference which says nothing about _when_ the attributes were created except that it was somewhat fast?

Maybe I'm missing something but I don't get the logic here.

SanjoDeundiak: We want to test that the value of `added` field is correct. There are several ways of defining "correct" here. I think the definition "lies inside some small interval around `now`" is absolutely sane here. I don't think testing `added >= now` is useful. I was also bitten recently by my wrong assumptions, which caused a bug, and I needed to add some drift allowance [here](https://github.com/build-trust/ockam/blob/75727da16d6f45b6ef81f3b4cc966db900365bba/implementations/rust/ockam/ockam_identity/src/purpose_keys/purpose_keys_verification.rs#L95). It should not be a problem when everything is on the same machine, but makes me think that it's an unnecessary check.

etorreborre: It can indeed make sense to add an assertion to check if it doesn't take too long to add an attribute, as a sanity check that we don't inadvertently create a timestamp in 100 years for example. 

What annoys me with using an absolute diff is that it kind of puts causality under the carpet. I expect the attribute to be `added` _after_ the user called the function to add a member. If we end up with `added()` being consistently 3 seconds before the call to `add_member` this could make up some very confusing logs.

SanjoDeundiak:
```
What annoys me with using an absolute diff is that it kind of puts causality under the carpet. I expect the attribute to be added after the user called the function to add a member. If we end up with added() being consistently 3 seconds before the call to add_member this could make up some very confusing logs.
```
1. Casualty is not expected in this test but is totally fine.
2. While it's possible to set an incorrect Timestamp or mangle it during some conversions (e.g. confuse seconds and minutes), it's very hard to imagine an implementation that would be consistently off by -3 seconds.
3. We're spending too much time discussing an insignificant problem IMHO.

etorreborre: > We're spending too much time discussing an insignificant problem IMHO

This is clearly not a major problem :-D. I am having some compassion for the poor fellow, like me, who will find this code peculiar in the future and will wonder why things were written that way. But I can see that our brains are wired differently :-)

PR #6174, opened 2023-09-29
cache nix store

metaclips:
Cache Nix dependencies and binary for faster CI runtime

### DIFF
--- a/.github/actions/nix_installer/action.yml
+++ b/.github/actions/nix_installer/action.yml
@@ -1,8 +1,0 @@
-name: Nix Installer
-description: Nix Installer
-
-runs:
-  using: composite
-  steps:
-    - name: Install Nix
-      uses: DeterminateSystems/nix-installer-action@3ebd1aebb47f95493b62de6eec0cac3cd74e50a9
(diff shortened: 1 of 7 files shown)


etorreborre (approved):
I was a bit surprised that there was not a single command to copy all nix-related directories at once but otherwise this looks good to me.

metaclips:
> I was a bit surprised that there was not a single command to copy all nix-related directories at once but otherwise this looks good to me.

We have the Nix copy command https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-copy.html which copies the Nix store and allows us to restore for every consecutive workflow run, I tried using `nix copy` but fail to cache due to some `unsigned nix dependencies` and also fails to copy dependencies that we build, e.g. [rust-overlay](https://github.com/oxalica/rust-overlay) which builds Rust and its tools, so I decided to just copy the Nix state and cache as-is.

metaclips:
> I was a bit surprised that there was not a single command to copy all nix-related directories at once but otherwise this looks good to me.

We have the Nix copy command https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-copy.html which copies the Nix store and allows us to restore for every consecutive workflow run, I tried using `nix copy` but fail to cache due to some `unsigned nix denpendencies` and also fails to copy dependencies that we build, e.g. [rust-overlay](https://github.com/oxalica/rust-overlay) which builds Rust and its tools, so I decided to just copy the Nix state and cache as-is.

PR #6214, opened 2023-10-02
Remove the `skip_defaults` flag

etorreborre:
The `skip_defaults` flag used when creating a `NodeManager` does not need to be kept on the struct for its whole lifetime.
That flag is set to `true` when a launch config has been passed on the command line and is otherwise set to `false`.


adrianbenavides (dismissed):
Wow, the NodeManager constructor is finally losing some weight!

adrianbenavides:
@etorreborre looks like the authority node initialization is not being done properly (two bats tests failing), but I can't seem to see what's wrong.

etorreborre:
@adrianbenavides I fixed the issue, the authority tests are passing now.

adrianbenavides (dismissed):

adrianbenavides (approved):

PR #6244, opened 2023-10-03
feat(rust): Improve JSON output of ockam node show

### LINKED ISSUE #6142: Improve `ockam node show --output json` by including more information
# Current behavior

If you run `ockam node show --output json` it currently does not show much information. Here's a screenshot.

Image

# Desired behavior

It should display information that is available when running the command with plain text output `ockam node show`.

Please take a look at this [file](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/node/show.rs#L78) to get started.

---

kriogenia:
## Current behavior

Running `ockam node show --output json` currently only shows the name of the node instead of all the information shown with the plain option

## Proposed changes

Closes #6142

The PR creates a new struct in the `ockam_api` with the information to show in the `ockam node show` command. This new struct implements the `Output` trait with the plain printing logic. It implements Serialize to build the JSON response. Both response are now printed using the `OutputFormat`.

Added `Serialize` to all components that are being printed in the command, mimicking the same information as format as what was currently being printed via plain.  

Example of plain output and new JSON output:

```
Node:
  Name: start (default)
  Status: UP
  Route To Node:
    Short: /node/start
    Verbose: /dnsaddr/localhost/tcp/44041
  Identity: I1b3787e02c02453fe5cd1be07e1fd62b5ca466ff
  Transports:
    Transport:
      Type: TCP
      Mode: Listening
      Socket: 127.0.0.1:44041
      Worker: <none>
      FlowControlId: c0b280ee5872661f857639dd3b5bded7
  Secure Channel Listeners:
    Listener:
      Address: /service/api
      FlowControlId: 66ee5fe42fab11fb1b7121ab086e30d0
  Inlets:
  Outlets:
  Services:
    Service:
      Type: uppercase
      Address: /service/uppercase
    Service:
      Type: echo
      Address: /service/echo
    Service:
      Type: credentials
      Address: /service/credentials
 ```

```json
{
  "is_default": true,
  "name": "start",
  "is_up": true,
  "route": {
    "short": "/node/start",
    "verbose": "/dnsaddr/localhost/tcp/44041"
  },
  "identity": "I1b3787e02c02453fe5cd1be07e1fd62b5ca466ff",
  "transports": [
    {
      "type": "Tcp",
      "mode": "Listen",
      "socket": "127.0.0.1:44041",
      "worker": "<none>",
      "flow_control": {
        "id": "c0b280ee5872661f857639dd3b5bded7"
      }
    }
  ],
  "secure_channel_listeners": [
    {
      "address": "/service/api",
      "flow_control": {
        "id": "66ee5fe42fab11fb1b7121ab086e30d0"
      }
    }
  ],
  "inlets": [],
  "outlets": [],
  "services": [
    {
      "address": "/service/uppercase",
      "type": "uppercase"
    },
    {
      "address": "/service/echo",
      "type": "echo"
    },
    {
      "address": "/service/credentials",
      "type": "credentials"
    }
  ]
}
```

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/models/show.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/models/show.rs
@@ -1,0 +1,163 @@
+use std::fmt::Display;
+
+use colorful::Colorful;
+
+use ockam_multiaddr::{
+    proto::{DnsAddr, Node, Tcp},
+    MultiAddr,
+};
+use serde::Serialize;
+
+use crate::output::Output;
+
+use super::{
+    portal::{ShowInletStatus, ShowOutletStatus},
+    secure_channel::ShowSecureChannelListener,
+    services::ShowServiceStatus,
+    transport::ShowTransportStatus,
+};
+
+/// Information to display in the `ockam node show` command
+#[derive(Debug, Serialize)]
+pub struct ShowNodeResponse {
+    pub is_default: bool,
+    pub name: String,
+    pub is_up: bool,
+    pub route: RouteToNode,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub identity: Option<String>,
+    pub transports: Vec<ShowTransportStatus>,
+    pub secure_channel_listeners: Vec<ShowSecureChannelListener>,
+    pub inlets: Vec<ShowInletStatus>,
+    pub outlets: Vec<ShowOutletStatus>,
+    pub services: Vec<ShowServiceStatus>,
+}
+#[derive(Debug, Serialize)]
+pub struct RouteToNode {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub short: Option<MultiAddr>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub verbose: Option<MultiAddr>,
+}
+
+impl ShowNodeResponse {
+    pub fn new(
+        is_default: bool,
+        name: &str,
+        is_up: bool,
+        node_port: Option<u16>,
+    ) -> ShowNodeResponse {
+        let mut m = MultiAddr::default();
+        let short = m.push_back(Node::new(name)).ok().map(|_| m);
+
+        let verbose = node_port.and_then(|port| {
+            let mut m = MultiAddr::default();
+            if m.push_back(DnsAddr::new("localhost")).is_ok() && m.push_back(Tcp::new(port)).is_ok()
+            {
+                Some(m)
+            } else {
+                None
+            }
+        });
+
+        ShowNodeResponse {
+            is_default,
+            name: name.to_owned(),
+            is_up,
+            route: RouteToNode { short, verbose },
+            identity: None,
+            transports: Default::default(),
+            secure_channel_listeners: Default::default(),
+            inlets: Default::default(),
+            outlets: Default::default(),
+            services: Default::default(),
+        }
+    }
+}
+
+impl Display for ShowNodeResponse {
+    fn fmt(&self, buffer: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(buffer, "Node:")?;
+
+        if self.is_default {
+            writeln!(buffer, "  Name: {} (default)", self.name)?;
+        } else {
+            writeln!(buffer, "  Name: {}", self.name)?;
+        }
+
+        writeln!(
+            buffer,
+            "  Status: {}",
+            match self.is_up {
+                true => "UP".light_green(),
+                false => "DOWN".light_red(),
+            }
+        )?;
+
+        writeln!(buffer, "  Route To Node:")?;
+        if let Some(short) = &self.route.short {
+            writeln!(buffer, "    Short: {short}")?;
+        }
+        if let Some(verbose) = &self.route.verbose {
+            writeln!(buffer, "    Verbose: {verbose}")?;
+        }
+
+        if let Some(identity) = &self.identity {
+            writeln!(buffer, "  Identity: {}", identity)?;
+        }
+
+        writeln!(buffer, "  Transports:")?;
+        for e in &self.transports {
+            writeln!(buffer, "    Transport:")?;
+            writeln!(buffer, "      Type: {}", &e.tt)?;
+            writeln!(buffer, "      Mode: {}", &e.mode)?;
+            writeln!(buffer, "      Socket: {}", &e.socket)?;
+            writeln!(buffer, "      Worker: {}", &e.worker)?;
+            writeln!(buffer, "      FlowControlId: {}", &e.flow_control)?;
+        }
+
+        writeln!(buffer, "  Secure Channel Listeners:")?;
+        for e in &self.secure_channel_listeners {
+            writeln!(buffer, "    Listener:")?;
+            if let Some(ma) = &e.address {
+                writeln!(buffer, "      Address: {ma}")?;
+            }
+            writeln!(buffer, "      FlowControlId: {}", &e.flow_control)?;
+        }
+
+        writeln!(buffer, "  Inlets:")?;
+        for e in &self.inlets {
+            writeln!(buffer, "    Inlet:")?;
+            writeln!(buffer, "      Listen Address: {}", e.listen_address)?;
+            if let Some(r) = &e.route_to_outlet {
+                writeln!(buffer, "      Route To Outlet: {r}")?;
+            }
+        }
+
+        writeln!(buffer, "  Outlets:")?;
+        for e in &self.outlets {
+            writeln!(buffer, "    Outlet:")?;
+            writeln!(buffer, "      Forward Address: {}", e.forward_address)?;
+            if let Some(ma) = &e.address {
+                writeln!(buffer, "      Address: {ma}")?;
+            }
+        }
+
+        writeln!(buffer, "  Services:")?;
+        for e in &self.services {
+            writeln!(buffer, "    Service:")?;
+            writeln!(buffer, "      Type: {}", e.service_type)?;
+            if let Some(ma) = &e.address {
+                writeln!(buffer, "      Address: {ma}")?;
+            }
+        }
+
+        Ok(())
+    }
+}
+
+impl Output for ShowNodeResponse {
+    fn output(&self) -> crate::error::Result<String> {
+        Ok(self.to_string())
+    }
+}
--- a/implementations/rust/ockam/ockam_command/src/node/show.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/show.rs
@@ -1,24 +1,26 @@
 use clap::Args;
-use colorful::Colorful;
+use miette::IntoDiagnostic;
+use ockam_api::nodes::models::secure_channel::SecureChannelListenersList;
+use ockam_api::nodes::models::services::ServiceList;
+use ockam_api::nodes::models::transport::TransportList;
+use ockam_api::nodes::BackgroundNode;
+use ockam_node::Context;
 use tokio_retry::strategy::FixedInterval;
 use tracing::{info, trace, warn};
 
 use ockam_api::cli_state::{CliState, StateDirTrait, StateItemTrait};
 use ockam_api::nodes::models::portal::{InletList, OutletList};
-use ockam_api::nodes::models::secure_channel::SecureChannelListenersList;
-use ockam_api::nodes::models::services::ServiceList;
-use ockam_api::nodes::models::transport::TransportList;
-use ockam_api::nodes::BackgroundNode;
-use ockam_api::{addr_to_multiaddr, route_to_multiaddr};
-use ockam_core::Route;
-use ockam_multiaddr::proto::{DnsAddr, Node, Tcp};
-use ockam_multiaddr::MultiAddr;
-use ockam_node::Context;
 
 use crate::node::get_node_name;
 use crate::node::util::check_default;
 use crate::util::{api, node_rpc};
-use crate::{docs, CommandGlobalOpts, OutputFormat, Result};
+use crate::{docs, CommandGlobalOpts, Result};
+
+use super::models::portal::{ShowInletStatus, ShowOutletStatus};
+use super::models::secure_channel::ShowSecureChannelListener;
+use super::models::services::ShowServiceStatus;
+use super::models::show::ShowNodeResponse;
+use super::models::transport::ShowTransportStatus;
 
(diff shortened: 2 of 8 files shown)


implementations/rust/ockam/ockam_command/tests/bats/nodes.bats
```diff
@@ -41,9 +41,14 @@ force_kill_node() {
   run_success "$OCKAM" node create "$n"
 
   run_success "$OCKAM" node show "$n"
-  assert_output --partial "/dnsaddr/localhost/tcp/"
+  assert_output --partial "/dnsaddr/localhost/tcp/44041"
```
adrianbenavides: The assigned port is random. We just want to make sure it has an address, so it's OK to revert it back.

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
```diff
@@ -150,11 +150,16 @@ impl CreateOutlet {
...
     #[n(1)] pub bind_addr: String,
+    #[serde(skip)]
     #[n(2)] pub worker_addr: String,
+    #[serde(skip)]
     #[n(3)] pub alias: String,
     /// An optional status payload
+    #[serde(skip)]
     #[n(4)] pub payload: Option<String>,
+	#[serde(rename = "route_to_outlet", serialize_with = "super::address::serialize_route_as_multiaddr")]
     #[n(5)] pub outlet_route: String,
```
adrianbenavides: There is a major problem with this approach here: the skip/rename rules you are adding here are tailored to match one specific use case: printing this struct in the CLI. So, instead of doing that, I'd propose to create a new struct in the `ockam_command` crate with just the fields that you need.

The same goes for the `OutletStatus` struct.

implementations/rust/ockam/ockam_api/src/nodes/models/address.rs
```diff
@@ -0,0 +1,19 @@
...
+{
+    s.serialize_some(&addr_to_multiaddr(t.clone()))
+}
+
+pub fn serialize_route_as_multiaddr<S>(route: &String, s: S) -> Result<S::Ok, S::Error>
+where
+S: Serializer,
+{
+	s.serialize_some(&Route::parse(route).and_then(|r| route_to_multiaddr(&r)))
+}
```
adrianbenavides: Not ideal, but I'd move this to `utils.rs` where we have the rest of multiaddr utility functions. We'll probably move all those functions to its own file, but we can do that in a separate PR.

implementations/rust/ockam/ockam_api/src/nodes/models/show.rs
```diff
@@ -0,0 +1,76 @@
+use ockam_multiaddr::{
```
adrianbenavides: This whole file should be moved to `ockam_command`.

implementations/rust/ockam/ockam_api/src/nodes/models/show.rs
```diff
@@ -0,0 +1,76 @@
...
+#[derive(Debug, Serialize)]
+pub struct ShowNodeResponse<'a> {
+    pub is_default: bool,
+    pub name: &'a str,
+    pub is_up: bool,
+    pub route: RouteToNode,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub identity: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none", flatten)]
+    pub tcp_listeners: Option<TransportList>,
```
adrianbenavides: Would it work if you add the `rename` attribute here insteaf of in the `TransportList`? If you can do that, you would be able to leave the `ockam_api` structs unmodified, which would be desirable.

implementations/rust/ockam/ockam_command/src/node/show.rs
```diff
@@ -180,77 +63,51 @@ pub async fn print_query_status(
...
-        );
-    } else {
-        let node_state = cli_state.nodes.get(node_name)?;
-        // Get short id for the node
-        let default_id = match node_state.config().identity_config() {
-            Ok(resp) => resp.identifier().to_string(),
-            Err(_) => String::from("None"),
-        };
+    let node_state = cli_state.nodes.get(node_name)?;
+    let node_port = node_state
```
adrianbenavides: Note that the port won't be available until the node is up and running. Also note that the `is_node_up` function is looping a few times to wait unti the node is up, that's why this function must be the first to run before querying any node information. Same problem with the `node_state`: if the node is not up when you call it, you won't be getting all the node information.

implementations/rust/ockam/ockam_command/src/node/show.rs
```diff
@@ -180,77 +63,51 @@ pub async fn print_query_status(
...
 
-    Ok(())
+            // Get list of outlets
+            let outlets: OutletList = node.ask(ctx, api::list_outlets()).await?;
+            node_info.outlets = Some(outlets);
+
+            node_info
+        };
+
+    Ok(opts.global_args.output_format.println_value(&node_info)?)
```
adrianbenavides: Since you now have an instance that implements the `Output` trait, the correct way to print the command output is throught the `terminal` instance. It would look something like:

```rust
opts.terminal
    .stdout()
    .plain(node_info)
    .json(serde_json::to_string_pretty(&node_info)?)
    .write_line()?;
```


[Here is an exampe](https://github.com/build-trust/ockam/blob/1a49eba84f17ea3b368aa22a1ffc6e53a474b474/implementations/rust/ockam/ockam_command/src/relay/delete.rs#L58) of another command using it.

adrianbenavides (commented):
Hey @kriogenia , the PR looks really great, I was looking forward to this refactor! I've left some comments with some suggestions.

Also, don't forget to run `cargo fmt` and `cargo clippy`, it looks like there are some files without the empty line at the end.

implementations/rust/ockam/ockam_api/src/nodes/models/show.rs
kriogenia: That was my first attempt but flatten doesn't work with rename. It makes as the struct being flattened could hold multiple attributes instead of just one like we have. A possible workaround to leave the API structs unmodified would be using a custom serializer or taking out the `Vec` with the list instead of passing the whole `TransportList`, but I would say that in this particular cases specifying the name of the lists to be more meaningful like "transports" or "services" vs. just "list" is probably good.

implementations/rust/ockam/ockam_api/src/nodes/models/show.rs
kriogenia: Should I dump all its content (and the new serialization structs) to the `show.rs` file or do I create a new `model` module in `node` like is in `ockam_api` to store all this serialization logic?

implementations/rust/ockam/ockam_command/tests/bats/nodes.bats
kriogenia: Reverted

implementations/rust/ockam/ockam_api/src/nodes/models/address.rs
kriogenia: Moved

implementations/rust/ockam/ockam_api/src/nodes/models/show.rs
kriogenia: The current version has the models/show.rs version

implementations/rust/ockam/ockam_command/src/node/show.rs
kriogenia: Reverted to previous way of doing it

implementations/rust/ockam/ockam_command/src/node/show.rs
kriogenia: Fixed

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
```diff
@@ -150,11 +150,16 @@ impl CreateOutlet {
...
     #[n(1)] pub bind_addr: String,
+    #[serde(skip)]
     #[n(2)] pub worker_addr: String,
+    #[serde(skip)]
     #[n(3)] pub alias: String,
     /// An optional status payload
+    #[serde(skip)]
     #[n(4)] pub payload: Option<String>,
+	#[serde(rename = "route_to_outlet", serialize_with = "crate::util::serialize_route_as_multiaddr")]
     #[n(5)] pub outlet_route: String,
```
adrianbenavides: We should revert the serde customizations for this struct and the OutletStatus. We use them in other crates "as is", so they rely on the current unmodified structure, and these changes would break that code.

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
```diff
@@ -237,10 +246,11 @@ impl OutletStatus {
...
 
 /// Response body when returning a list of Inlets
-#[derive(Debug, Clone, Decode, Encode)]
+#[derive(Debug, Clone, Decode, Encode, Serialize)]
 #[rustfmt::skip]
 #[cbor(map)]
 pub struct InletList {
+	#[serde(rename = "inlets")]
     #[n(1)] pub list: Vec<InletStatus>
 }
```
adrianbenavides: This still looks off to me... I think the proper way of doing this would be to delete all these `XList` structs that contain a single field with a vector, and refactor the endpoints that use them to return the plain vector. That way, you don't need to mess with serde anymore and you can customize the `ockam node show` structure as you want.

Since this is a little out of the scope for this PR, let me know if you want to tackle that as part of this PR. Otherwise I'll create a new issue for it.

implementations/rust/ockam/ockam_api/src/nodes/models/services.rs
```diff
@@ -365,7 +365,9 @@ impl StartOktaIdentityProviderRequest {
 #[rustfmt::skip]
 #[cbor(map)]
 pub struct ServiceStatus {
+	#[serde(rename = "address", serialize_with = "crate::util::serialize_as_multiaddr")]
```
adrianbenavides: If we do this, then the `deserialize_with` attribute should be defined as well. Try to write a unit test to check that the serde methods work as expected.

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
kriogenia: Yes, I completely agree. In fact, in my change to the inlets and outlets list I got rid of the list wrapper to reduce complexity and I think that it's way better. I agree that the change can be out of scope of the PR but I would gladly take on it after (or before) this one. You can create the issue and assign it to me.

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
kriogenia: Fixed

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
kriogenia: Fixed

implementations/rust/ockam/ockam_api/src/nodes/models/services.rs
kriogenia: Would you prefer making the same approach as we did with inlets and outlets? Creating a version of the struct to handle the custom serialization without affecting api? I can also do the deserialization and tests. Whatever you think is better.

kriogenia:
@adrianbenavides I updated the whole PR to use custom structs instead of modifying those at `ockam_api`. Now the PR features only changes in `ockam_command`. If you want to get rid of the list wrappers as we discussed I can take into the issue if you want.

implementations/rust/ockam/ockam_command/src/node/models/show.rs
```diff
@@ -0,0 +1,163 @@
...
+use super::{
+    portal::{ShowInletStatus, ShowOutletStatus},
+    secure_channel::ShowSecureChannelListener,
+    services::ShowServiceStatus,
+    transport::ShowTransportStatus,
+};
+
+/// Information to display in the `ockam node show` command
+#[derive(Debug, Serialize)]
+pub struct ShowNodeResponse<'a> {
```
adrianbenavides: I think you could get rid of that lifetime. It's only used by one of the fields, so it's not doing much. Removing it will also simplify the definition of the traits it implements.

implementations/rust/ockam/ockam_command/src/node/models/show.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub identity: Option<String>,
+    pub transports: Vec<ShowTransportStatus>,
+    pub secure_channel_listeners: Vec<ShowSecureChannelListener>,
+    pub inlets: Vec<ShowInletStatus>,
+    pub outlets: Vec<ShowOutletStatus>,
+    pub services: Vec<ShowServiceStatus>,
+}
+#[derive(Debug, Serialize)]
+struct RouteToNode {
```
adrianbenavides: It fails to compile because this struct is not public. Make sure to run `cargo fmt` and `cargo clippy` to make the compiler happy 🤖

adrianbenavides (commented):
Awesome! I think it feels way simpler and predictable now 👍  I left a couple of minor comments, but we can merge the PR as soon as those are fixed.

>  If you want to get rid of the list wrappers as we discussed I can take into the issue if you want.

Don't worry about that, the PR is perfect as it is now!

implementations/rust/ockam/ockam_api/src/nodes/models/services.rs
adrianbenavides: I think you already solved this, so no custom deserialization or tests are needed.

implementations/rust/ockam/ockam_command/src/node/models/show.rs
kriogenia: Oh mb, I forgot that the type was being exposed directly from the properties. Fixed.

implementations/rust/ockam/ockam_command/src/node/models/show.rs
kriogenia: Removed

adrianbenavides (approved):
Good to merge! Thank you so much @kriogenia for this awesome PR, we've been looking forward to this for quite a while now 💪

PR #6316, opened 2023-10-05
Refactor attributes types

etorreborre:
[NOTE]: the `ockam_documentation` update [PR is ready to be merged](https://github.com/build-trust/ockam-documentation/pull/74).

Attributes are used to create credentials for a given identity. When a credential is received and validated, its attributes are stored by a node. Then resources can have associated policies declaring that an identity can access it only if the attributes associated to that identity satisfies a logic expression.

The `ockam_abac` crate defines the exact type of values which can be used in expressions: string, bool, int, float, and a sequence of such values. Moreover, throughout the code attributes names are always expected to be strings (which is reasonable).

Yet we model attributes in our code as `name: Vec<u8>` and `value: Vec<u8>`. This is unnecessarily general. By having proper types for attribute names and values we signal to the library what can be really used and we avoid making mistakes by passing incompatible values.

This PR changes the type of `AttributesEntry` so that it contains a map of `String` -> `AttributeValue` pairs.

### Notes

 - the actual encoding / decoding of data exchanged with the Controller is still left as a pair of `Vec<u8>` where effectively those bytes only ever represent UTF-8 encoded strings at the moment (even for values)

 - the `AttributeValue` data type has been introduced to separate the declarations / persistence of attributes in the `ockam_identity` repository from their potential use as `Expr`essions in the `ockam_abac` crate (since there is a dependency from `ockam_abac` to `ockam_identity`)

### DIFF
--- a/implementations/rust/ockam/ockam_abac/src/attribute_access_control.rs
+++ b/implementations/rust/ockam/ockam_abac/src/attribute_access_control.rs
@@ -1,4 +1,3 @@
-use core::str::from_utf8;
 use ockam_core::async_trait;
 use ockam_core::compat::boxed::Box;
 use ockam_core::compat::fmt;
@@ -10,7 +9,7 @@ use ockam_core::Result;
 use ockam_core::{IncomingAccessControl, RelayMessage};
 use tracing as log;
 
-use crate::expr::str;
+use crate::expr::{attribute_value, str};
 use crate::Expr::*;
 use crate::{eval, Env, Expr};
 use ockam_core::compat::format;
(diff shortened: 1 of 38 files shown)


implementations/rust/ockam/ockam_abac/src/attribute_access_control.rs
```diff
@@ -82,28 +82,16 @@ impl AbacAccessControl {
                         "attribute key with whitespace ignored"
                     }
                 }
-                match str::from_utf8(value) {
```
etorreborre, line 97: No need for this last minute conversion now.

implementations/rust/ockam/ockam_abac/src/attribute_access_control.rs
```diff
@@ -82,28 +82,16 @@ impl AbacAccessControl {
...
+
+                if environment.contains(key) {
+                    log::debug! {
+                        policy = %self.expression,
+                        id     = %id,
+                        key    = %key,
+                        "attribute already present"
                     }
+                } else {
+                    environment.put(format!("subject.{key}"), attribute_value(value.clone()));
```
etorreborre: `attribute_value` is the function that transforms an `AttributeValue` into an `Expr` to add to the environment.

etorreborre:
Closed in favor of https://github.com/build-trust/ockam/pull/6979

PR #6373, opened 2023-10-09
ci: Created a composite actions and used it in .github/workflows/all.yml

### LINKED ISSUE #6238: Make the commit lint workflow a reusable workflow across all build-trust repositories
We have a workflow that checks commits when a pull request is created, it entails
- Checking if users sign their commits
- Checking if users commit messages follows a naming convention
- Checking if users PR have a linear commit history

The workflow file can be found [here](https://github.com/build-trust/ockam/blob/develop/.github/workflows/all.yml) in the ockam repository and [here](https://github.com/build-trust/ockam-website/blob/develop/.github/workflows/commits.yml) in the ockam-website repository, we should create a parent workflow which we can reuse across the two repository, so that we do not repeat code

milinddethe15:
#6238 
https://github.com/build-trust/ockam-website/blob/develop/.github/workflows/commits.yml needs changes to import this composite action. Environment variables must be declared accurately to ensure proper functionality.


metaclips:
Would need your input here @mrinalwadhwa

metaclips:
Would need your input here @mrinalwadhwa 🙏

metaclips (approved):
Approved 🚀. Thanks.

etorreborre (approved):

PR #6449, opened 2023-10-10
feat(rust): improve `ockam vault list` to mark the default vault

### LINKED ISSUE #6137: Improve `ockam vault list` to mark the default vault
# Current behavior

If you have multiple vaults and run `ockam vault list`, it currently does not show which one is the default vault. The screenshot below shows this.

Image

# Desired behavior

The default vault (which you can get by running `ockam vault show`) should be marked with a suffix `(default)` so that it is easy to see which one is the default.

---

melissa-kun-li:
Closes #6137

This commit will:
- create the VaultListOutput struct in ockam_command/src/vault/list.rs to represent the output for `vault list`
- implement `output()` for VaultListOutput to be called in build_list() for rendering
- suffix the default vault with '(default)' on the `ockam vault list` command

## Current behavior
`ockam vault list` doesn't mark the default vault

## Proposed changes
mark the default vault in `ockam vault list`.

Tests:

1) First vault created set to default
image

image
2) Additional vaults created are not default
image

image
3) Changing default vault
image

image

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/vaults.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/vaults.rs
@@ -80,6 +80,10 @@ impl VaultState {
     pub fn name(&self) -> &str {
         &self.name
     }
+
+    pub fn is_aws(&self) -> bool {
+        self.config.is_aws()
+    }
 }
 
 impl Display for VaultState {
--- a/implementations/rust/ockam/ockam_command/src/output/output.rs
+++ b/implementations/rust/ockam/ockam_command/src/output/output.rs
@@ -375,26 +375,6 @@ impl Output for VaultState {
         )?;
         Ok(output)
     }
-
-    fn list_output(&self) -> Result<String> {
-        let mut output = String::new();
-        writeln!(
-            output,
-            "Vault {}",
-            self.name().color(OckamColor::PrimaryResource.color())
-        )?;
-        write!(
-            output,
-            "Type {}",
-            match self.config().is_aws() {
-                true => "AWS KMS",
-                false => "OCKAM",
-            }
-            .to_string()
-            .color(OckamColor::PrimaryResource.color())
-        )?;
-        Ok(output)
-    }
 }
 
 fn human_readable_time(time: TimestampInSeconds) -> String {
--- a/implementations/rust/ockam/ockam_command/src/vault/list.rs
+++ b/implementations/rust/ockam/ockam_command/src/vault/list.rs
@@ -1,8 +1,14 @@
 use clap::Args;
+use colorful::Colorful;
 use miette::IntoDiagnostic;
+use std::fmt::Write;
 
 use ockam_api::cli_state::traits::StateDirTrait;
+use ockam_api::cli_state::StateItemTrait;
+use ockam_api::cli_state::VaultConfig;
 
+use crate::output::Output;
+use crate::terminal::OckamColor;
 use crate::util::local_cmd;
 use crate::{docs, CommandGlobalOpts};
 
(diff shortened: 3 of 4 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/vaults.rs
```diff
@@ -80,6 +80,10 @@ impl VaultState {
     pub fn name(&self) -> &str {
         &self.name
     }
+
+    pub fn is_aws(&self) -> bool {
+        self.config.is_aws()
+    }
```
adrianbenavides, line 86: I would remove this method and access the `config` directly. The different implementors of the `StateItemTrait` have different `config` structures with lots of fields. We don't want to start wrapping all those fields in higher level functions and keep it as lean as possible.

implementations/rust/ockam/ockam_command/src/output/output.rs
```diff
@@ -375,26 +375,6 @@ impl Output for VaultState {
...
-            "Type {}",
-            match self.config().is_aws() {
-                true => "AWS KMS",
-                false => "OCKAM",
-            }
-            .to_string()
-            .color(OckamColor::PrimaryResource.color())
-        )?;
-        Ok(output)
-    }
```
adrianbenavides, line 397: Great!

implementations/rust/ockam/ockam_command/src/vault/list.rs
```diff
@@ -25,11 +29,67 @@ impl ListCommand {
...
+            vault.name().to_string(),
+            vault.is_aws(),
+            default_vault_name == *vault.name(),
+        );
+        vault_output.push(i);
+    }
+
+    let plain =
+        opts.terminal
+            .build_list(&vault_output, "Vaults", "No vaults found on this system.")?;
```
adrianbenavides: You can also simplify a little how you build the output:

```rust
let output = vaults
    .iter()
    .map(|v| {
        VaultListOutput::new(
            v.name().to_string(),
            v.config().is_aws(),
            opts.state.vaults.is_default(v.name()).unwrap_or(false),
        )
    })
    .collect::<Vec<_>>();
```

Now that we have a proper type to handle the vault ouptut, you could use it for all the formats, so that the output has the same structure:

```rust
let plain = opts
    .terminal
    .build_list(&output, "Vaults", "No vaults found on this system.")?;
let json = serde_json::to_string_pretty(&output).into_diagnostic()?;
```

For this to work though, you will need to add `#[derive(serde::Serialize)]` to the `VaultListOutput`.

adrianbenavides (commented):
Hey @melissa-kun-li , thanks for submitting this PR, it looks pretty great! I've added some comments to simplify a little how you build the output.

If you want to run the bats tests, you'll need to follow the instructions from [here](https://github.com/build-trust/ockam/blob/b3ebd2f99b2077212a0c7eddb0bbad419e5ffbe8/implementations/rust/ockam/ockam_command/tests/bats/README.md).

Once you have that, you can run:

```bash
# cd to the `ockam` repo directory
OCKAM=./target/debug/ockam
cargo build -p ockam_command
bats ./implementations/rust/ockam/ockam_command/tests/bats/vault.bats
```

Let me know if you have any questions 👍

implementations/rust/ockam/ockam_command/src/vault/list.rs
melissa-kun-li: Ahh that is simpler, thanks! 
The current json output looks like this:
```
    Running `/Users/melissali/Documents/ockam/ockam/target/debug/ockam vault list --output json`
[
  {
    "name": "v1",
    "path": "/Users/melissali/.ockam/vaults/v1.json",
    "data_path": "/Users/melissali/.ockam/vaults/data/v1-storage.json",
    "config": {
      "aws_kms": false
    }
  },
  {
    "name": "v2",
    "path": "/Users/melissali/.ockam/vaults/v2.json",
    "data_path": "/Users/melissali/.ockam/vaults/data/v2-storage.json",
    "config": {
      "aws_kms": false
    }
  }
]
```

If I use the vault output for the json output (after I also add more attributes to the VaultListOutput struct), I can get it to look like this:
```
     Running `/Users/melissali/Documents/ockam/ockam/target/debug/ockam vault list --output json`
[
  {
    "name": "v1",
    "config": {
      "aws_kms": false
    },
    "is_default": false,
    "path": "/Users/melissali/.ockam/vaults/v1.json",
    "data_path": "/Users/melissali/.ockam/vaults/data/v1-storage.json"
  },
  {
    "name": "v2",
    "config": {
      "aws_kms": false
    },
    "is_default": true,
    "path": "/Users/melissali/.ockam/vaults/v2.json",
    "data_path": "/Users/melissali/.ockam/vaults/data/v2-storage.json"
  }
]
```
Is this the same idea you had? I tried to preserve all the original keys, just adding the "is_default" field.

melissa-kun-li:
@adrianbenavides Thanks for the review and letting me know how to run the bats test!  My local changes didn't appear in the tests until I did `OCKAM=./target/debug/ockam` like you said

I made some of the changes you recommended, plus some additional changes so that it's usable in the json output too.  I'm a bit unsure if it's the right way to go about it though, because I did `v.config().clone()` to get the config into `VaultListOutput` so it can be used in the json. Is this the right approach + ok to use clone here?

implementations/rust/ockam/ockam_command/src/vault/list.rs
adrianbenavides: You can drop the `path` and `data_path` fields, as they are not very useful to the CLI user. Also, you could flatten the `config` field by adding the `#[serde(flatten)]` attribute to that field in the `VaultListOutput` struct.

adrianbenavides:
> I made some of the changes you recommended, plus some additional changes so that it's usable in the json output too. I'm a bit unsure if it's the right way to go about it though, because I did `v.config().clone()` to get the config into `VaultListOutput` so it can be used in the json. Is this the right approach + ok to use clone here?

No problem with cloning in this case, as the `VaultConfig` only contains a bool 👍

adrianbenavides (dismissed):
Thank you so much for your contribution @melissa-kun-li 🙏 

Ready to merge! 🚀

adrianbenavides (approved):

mrinalwadhwa (approved):

etorreborre (approved):

PR #6480, opened 2023-10-13
delete node interactively

### LINKED ISSUE #6450: Make `ockam node delete` (no args) interactive by asking the user to choose from a list of nodes to delete (tuify)
# Current behavior

When a user runs `ockam node delete` without any arguments, currently they are prompted to delete the default node. Here's a screenshot of this.

Image

# Desired behavior

Instead of deleting the default node, change the behavior of this command so that it becomes interactive with the user. Note that this can only be allowed to happen if:
- the terminal is interactive, and,
- the [`--no-input`](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/terminal/mod.rs#L265) or `--quiet` flags have *not* been passed, and,
- no arguments have been passed in to the command.

This what the user flow should look like in interactive mode.
1) Use the [tuify](https://crates.io/crates/r3bl_tuify) crate's `select_from_list()` function to ask the user to select from a list of node names. They should be able to make multiple selections. If there are none, then simply display a message saying that none exist (not an error) and exit.

2) Once the user has made their selection(s), display a confirmation prompt letting the user know that that a destructive operation is about to take place with the node names they have selected. Use `select_from_list()` to display this prompt. They should be able to make a single selection.
   - If they have made no selections then exit by displaying a message to the user letting them know they did not make any selection.

3) If they choose "YES", then delete each selected node, and display a success / failure message for each node that got deleted. 
   - If an operation succeeds then use the ✅ emoji as a prefix for the message. 
   - And if it fails, then use the ⚠️ emoji as a prefix for the message.

> The following is a video showing a generic user flow that shows the user experience when a user is asked to choose a from a list of "things". Then an action is performed on each "thing". Finally status information is reported for each action that is performed on each "thing". Please note that this flow is *not* specific to this issue. 
> Image
>  https://asciinema.org/a/612547

# Implementation details

- If it hasn't been added already, please add the [tuify](https://crates.io/crates/r3bl_tuify) crate the `Cargo.toml` file [here](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/Cargo.toml#L55) by adding this line `r3bl_tuify = "0.1.16"` to the `[dependencies]` section.
- You can learn more about how to use the `select_from_list()` in the tuify crate [here](https://docs.rs/r3bl_tuify/latest/r3bl_tuify/). Here is an [example](https://github.com/r3bl-org/r3bl_rs_utils/blob/ockam-demo/tuify/examples/main_interactive.rs) that is very similar to the steps you might have to take for this issue.
- Here's a [file](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/terminal/mod.rs#L255) you can look at to detect when a terminal is interactive or not.
- Here's a [file](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/node/delete.rs#L43) you can look at to see where the `ockam node delete` command is implemented.

---
<i>

We love helping new contributors! ❤️
If you have questions or need help as you explore, please join us on [Discord](https://discord.gg/bewvnm6zqS). If you're looking for other issues to contribute to, please checkout our [good first issues](https://github.com/build-trust/ockam/labels/good%20first%20issue).
</i>

mickm3n:
For implementing https://github.com/build-trust/ockam/issues/6450

## Current behavior

When a user runs `ockam node delete` without any arguments, currently they are prompted to delete the default node. Here's a screenshot of this.

Image

## Proposed changes

Instead of deleting the default node, change the behavior of this command so that it becomes interactive with the user. Note that this can only be allowed to happen if:

* the terminal is interactive, and,
* the `--no-output` or `--quiet` flags have _not_ been passed, and,
* no arguments have been passed in to the command.

When there one or more nodes could be deleted, after a user runs `ockam node delete`, the user will see
* image
* If the user doesn't select one and press Enter, the user will see
  * image
* But if the user does select and press Enter, the user will see
  * image
  * After selecting `YES`
    * image

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/delete.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/delete.rs
@@ -1,11 +1,12 @@
 use clap::Args;
 use colorful::Colorful;
+use ockam_api::cli_state::StateDirTrait;
 
-use crate::node::get_node_name;
+use crate::node::get_default_node_name;
 use crate::node::util::{delete_all_nodes, delete_node};
-
+use crate::terminal::tui::DeleteMode;
 use crate::util::local_cmd;
-use crate::{docs, fmt_ok, CommandGlobalOpts};
+use crate::{docs, fmt_ok, fmt_warn, CommandGlobalOpts};
 
 const LONG_ABOUT: &str = include_str!("./static/delete/long_about.txt");
 const AFTER_LONG_HELP: &str = include_str!("./static/delete/after_long_help.txt");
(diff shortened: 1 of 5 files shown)


mickm3n:
@nazmulidris I don't know how to make node deletion fail, so I haven't tested the failed scenario. If you have any advice, I could try it.

Any advice is welcome!

nazmulidris:
> I don't know how to make node deletion fail, so I haven't tested the failed scenario. If you have any advice, I could try it. Any advice is welcome!

@adrianbenavides Any thoughts on this?

implementations/rust/ockam/ockam_command/src/node/util.rs
```diff
@@ -34,6 +34,15 @@ pub fn delete_node(opts: &CommandGlobalOpts, name: &str, force: bool) -> miette:
...
 }
 
+pub fn get_all_node_names(opts: &CommandGlobalOpts) -> miette::Result<Vec<String>> {
+    let nodes_states = opts.state.nodes.list()?;
+    let mut nodes = Vec::new();
+    for s in nodes_states {
+        nodes.push(s.name().to_string());
+    }
+    Ok(nodes)
+}
```
adrianbenavides: There is a function for this already: `opts.state.nodes.list_items_names()?`

implementations/rust/ockam/ockam_command/src/node/delete.rs
```diff
@@ -40,32 +40,90 @@ impl DeleteCommand {
...
+            if opts
+                .terminal
+                .confirm_interactively(format!(
+                    "Would you like to delete these items : {:?}?",
+                    selected_list
+                ))
+                .unwrap_or(false)
+            {
+                let (deleted_nodes, delete_fail_nodes_with_error) =
+                    delete_selected_nodes(&opts, selected_list, cmd.force)?;
```
adrianbenavides: This could be replaced by:

```rust
let res = node_names
    .iter()
    .map(|name| (name, opts.state.nodes.delete_sigkill(name, cmd.force)))
    .collect::<Vec<_>>();
```

implementations/rust/ockam/ockam_command/src/node/delete.rs
```diff
@@ -40,32 +40,90 @@ impl DeleteCommand {
...
+
+                for deleted_node in deleted_nodes {
+                    println!("✅ Deleted Node: '{}'", &deleted_node);
```
adrianbenavides: All the `println` calls should be replaced by calls to `opts.terminal.write_line`. Also, you should use `fmt_ok` or the other [fmt helpers](https://github.com/build-trust/ockam/blob/dcae9ee9a2a343a13beaf78d29b0efd66e04555b/implementations/rust/ockam/ockam_command/src/terminal/fmt.rs) instead of embedding emojis directly.

adrianbenavides (commented):
> @nazmulidris I don't know how to make node deletion fail, so I haven't tested the failed scenario. If you have any advice, I could try it.

The easiest way would be to manually delete the `pid` file in the node's state directory before deleting it through the command.

mickm3n:
Thanks for your review. I'll check.

mickm3n:
Hi @adrianbenavides, I fix all the problems in your feedback. Please help review again 🙏

I successfully triggered the failed condition by modifying the pid in the node's pid file.
image

However, if I remove the PID file, the result will be succeeded, but the node process will remain alive. I have no idea how to trace the problem; maybe we should create an issue.

adrianbenavides:
> Hi @adrianbenavides, I fix all the problems in your feedback. Please help review again 🙏
> 
> I successfully triggered the failed condition by modifying the pid in the node's pid file. image

Yeah, that's right, you had to modify it, not remove it. Removing the PID file is expected to have no effect when removing the node, my mistake!

implementations/rust/ockam/ockam_command/src/node/delete.rs
```diff
@@ -40,32 +41,94 @@ impl DeleteCommand {
     }
 }
 
+enum DeleteMode {
+    All,
+    Selected(Option<Vec<String>>),
+    Single(Option<String>),
```
adrianbenavides: Perhaps, instead of using `Option` in these two variants, you could add a new variant called `None`, so that you don't have to check the `Option` when deleting the nodes. It would also help removing the `unwrap` from line 96: `let selected_node_names = option_selected_node_names.unwrap();`

To make things even easier to understand, I would add a `Default` variant that should be used when `DeleteCommand`'s `node_name` is `None`.

implementations/rust/ockam/ockam_command/src/terminal/mod.rs
```diff
@@ -252,8 +259,54 @@ impl<W: TerminalWriter> Terminal<W> {
...
+        match &user_input {
+            Some(it) => {
+                if it.contains(&"YES".to_string()) {
+                    return Some(true);
+                } else {
+                    return Some(false);
+                }
+            }
+            None => Some(false),
+        }
```
adrianbenavides: This never returns `None` right? Could we return a plain `bool`?

implementations/rust/ockam/ockam_command/src/node/delete.rs
```diff
@@ -40,32 +41,94 @@ impl DeleteCommand {
...
+                            fmt_ok!("Deleted Node: '{}'\n", name)
+                        } else {
+                            fmt_warn!(
+                                "Failed to delete Node: '{}', Error: '{}'\n",
+                                name,
+                                res.as_ref().unwrap_err()
+                            )
+                        }
+                    })
+                    .collect::<String>();
```
adrianbenavides: You can simplify this as follows (no need to show the error to the user):

```rust
let output = selected_node_names
    .iter()
    .map(|name| {
        if opts.state.nodes.delete_sigkill(name, cmd.force).is_ok() {
            fmt_ok!("Node {name} deleted\n")
        } else {
            fmt_warn!("Failed to delete Node '{name}'\n")
        }
    })
    .collect::<String>();
```

adrianbenavides (commented):
I've left some more comments. Looking pretty great so far!

mickm3n:
@adrianbenavides, I appreciate your feedback. You saved me from the Option and iterator hell. I learned a lot. I fixed all of them; please help review them again. Many thanks!

adrianbenavides (dismissed):
Looks fantastic and ready to merge! Thank you so much for this contribution @mickm3n, will get merged soon 🥳

nazmulidris (approved):
Looks great to me! 🎉

PR #6515, opened 2023-10-13
add rust cache composite action

metaclips:
Add a composite action that only save rust dependency if it's run as a cron job so that we do not exceed github cache storage limit


mrinalwadhwa:
nice! good idea.

etorreborre (approved):

PR #6648, opened 2023-10-18
increase cron job time for better caching

metaclips:
We now only cache data when workflows run as a cron job, this ensures that we only cache from the default branch and that we are not over-caching (which causes github to delete our relevant caches), this PR ensures that we run the cron-job (which uploads caches) every two hours so that we are using latest cache data across all workflows

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -36,8 +36,12 @@ on:
     branches:
       - develop
   schedule:
-    # At 1 am.
-    - cron: "0 1 * * *"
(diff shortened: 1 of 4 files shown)


davide-baldo (approved):

mrinalwadhwa:
I suspect, once a day when most people on our team are not working might be sufficient here

metaclips:
> I suspect, once a day when most people on our team are not working might be sufficient here

The issue with running once a day is that, when the team deletes the cache manually using, `gh cache delete --all`, we won't have any cache stored till the next day.

PR #6649, opened 2023-10-18
Improves macOS Swift Application responsiveness

davide-baldo:
Improved responsiveness by having a parallel inlet refresh and status reporting for each
Added a scheduler for refresh operations instead of a timed loop
Trigger refresh only when states change (for both `ApplicationState` and internal refresh of inlets re-triggered only when invites changes)
`NodeManager` is now reachable and `ockam_app`can be viewed from CLI (ockam node show ockam_app)
On exit, every app-related node is deleted.
A connection bug was also fixed: duct cmd was executing in blocking mode inside an async context, this could trigger several issues such as timeouts or stalls.

On the UI side, the sub-menu of groups was rewritten to match the current U. Also minor stylistic touch ups on the create/share service.

### DIFF
--- a/implementations/rust/ockam/ockam_app_lib/src/background_node.rs
+++ b/implementations/rust/ockam/ockam_app_lib/src/background_node.rs
@@ -1,6 +1,7 @@
 use crate::cli::cli_bin;
 use crate::error::Error;
 use crate::Result;
+use ockam::compat::tokio::task::spawn_blocking;
 use ockam_api::nodes::models::portal::InletStatus;
 use ockam_core::async_trait;
 use std::net::SocketAddr;
@@ -85,37 +86,42 @@ impl BackgroundNodeClient for Cli {
 #[async_trait]
 impl Nodes for Cli {
     async fn create(&mut self, node_name: &str) -> Result<()> {
-        duct::cmd!(
-            &self.bin,
-            "--no-input",
-            "node",
-            "create",
-            node_name,
-            "--trust-context",
-            node_name
-        )
-        .before_spawn(log_command)
-        .stderr_null()
-        .stdout_capture()
-        .run()?;
-        debug!(node = %node_name, "Node created");
+        let bin = self.bin.clone();
+        let node_name = node_name.to_string();
+        spawn_blocking(move || {
+            let _ = duct::cmd!(
+                &bin,
+                "--no-input",
+                "node",
+                "create",
+                &node_name,
+                "--trust-context",
+                &node_name
+            )
+            .before_spawn(log_command)
+            .stderr_null()
+            .stdout_capture()
+            .run()
+            .map(|_| debug!(node = %node_name, "Node created"));
+        })
+        .await?;
+
         Ok(())
     }
 
     async fn delete(&mut self, node_name: &str) -> Result<()> {
         debug!(node = %node_name, "Deleting node");
-        let _ = duct::cmd!(
-            &self.bin,
-            "--no-input",
-            "node",
-            "delete",
-            "--yes",
-            node_name
-        )
-        .before_spawn(log_command)
-        .stderr_null()
-        .stdout_capture()
-        .run();
+        let bin = self.bin.clone();
+        let node_name = node_name.to_string();
+        spawn_blocking(move || {
+            let _ = duct::cmd!(&bin, "--no-input", "node", "delete", "--yes", &node_name)
+                .before_spawn(log_command)
+                .stderr_null()
+                .stdout_capture()
+                .run();
+        })
+        .await?;
+
         Ok(())
     }
 }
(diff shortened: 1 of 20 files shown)


implementations/rust/ockam/ockam_app_lib/src/background_node.rs
```diff
@@ -85,37 +86,45 @@ impl BackgroundNodeClient for Cli {
...
-            "--trust-context",
-            node_name
-        )
-        .before_spawn(log_command)
-        .stderr_null()
-        .stdout_capture()
-        .run()?;
+        {
+            let bin = self.bin.clone();
+            let node_name = node_name.to_string();
```
adrianbenavides: Perhaps we could rename this to `node_name_moved` and remove the extra indentation?

implementations/rust/ockam/ockam_app_lib/src/background_node.rs
```diff
@@ -85,37 +86,45 @@ impl BackgroundNodeClient for Cli {
...
+                    "--trust-context",
+                    &node_name
+                )
+                .before_spawn(log_command)
+                .stderr_null()
+                .stdout_capture()
+                .run()
+            })
+            .await
+            .unwrap()?;
```
adrianbenavides: Perhaps, instead of returning the result from duct just to unwrap it, and since we can't recover from that failure, we could do the following:

```rust
spawn_blocking(move || {
    let _ = duct::cmd!(
        &bin,
        "--no-input",
        "node",
        "create",
        &node_name_moved,
        "--trust-context",
        &node_name_moved
    )
    .before_spawn(log_command)
    .stderr_null()
    .stdout_capture()
    .run()
    .map(|_| debug!(node = %node_name, "Node created"));
})
.await?;
```

implementations/rust/ockam/ockam_app_lib/src/background_node.rs
```diff
@@ -85,37 +86,45 @@ impl BackgroundNodeClient for Cli {
...
+        let node_name = node_name.to_string();
+        spawn_blocking(move || {
+            let _ = duct::cmd!(&bin, "--no-input", "node", "delete", "--yes", &node_name)
+                .before_spawn(log_command)
+                .stderr_null()
+                .stdout_capture()
+                .run();
+        })
+        .await
+        .unwrap();
```
adrianbenavides: I'd say we have the conversions in place to transform the error returned from `spawn_blocking` into our own error.

implementations/rust/ockam/ockam_app_lib/src/invitations/commands.rs
```diff
@@ -196,93 +199,151 @@ impl AppState {
...
+            self.publish_state().await;
+            self.schedule_inlets_refresh_now();
+        }
+
         Ok(())
     }
 
     pub(crate) async fn refresh_inlets(&self) -> crate::Result<()> {
-        debug!("Refreshing inlets");
+        info!("Refreshing inlets");
```
adrianbenavides, line 217: We have an `info` log at the end of the function. Wouldn't it be too spammy to have both logs in the info level?

implementations/rust/ockam/ockam_app_lib/src/state/mod.rs
```diff
@@ -60,6 +64,10 @@ pub struct AppState {
     application_state_callback: ApplicationStateCallback,
     notification_callback: NotificationCallback,
     node_manager: Arc<RwLock<Arc<InMemoryNode>>>,
+    refresh_project_scheduler: Arc<Mutex<Option<Scheduler>>>,
+    refresh_invitations_scheduler: Arc<Mutex<Option<Scheduler>>>,
+    refresh_inlets_scheduler: Arc<Mutex<Option<Scheduler>>>,
```
adrianbenavides: Have you considered using [cells](https://docs.rs/once_cell/latest/once_cell/#safe-initialization-of-global-data) here? Perhaps you could replace `Mutex<Option>` by `OnceCell<Scheduler>`.

The instantiation should be `refresh_project_scheduler: Arc::new(OnceCell<Scheduler>),`.


The initialization would look like `self.refresh_project_scheduler.set(Arc::new(Scheduler::create(...)))`.


And finally, using it would be as simple as `self.refresh_project_scheduler.get()` if you add a method like the `global` method described in the docs.

I haven't checked this in detail, but in case in can be helpful.

implementations/rust/ockam/ockam_app_lib/src/invitations/commands.rs
davide-baldo: i think having one log entry every 10 seconds is acceptable

implementations/rust/ockam/ockam_app_lib/src/state/mod.rs
davide-baldo: this seems MUCH better!

davide-baldo: I ended up using `OnceLock` 👍

implementations/rust/ockam/ockam_app_lib/src/background_node.rs
davide-baldo: done! 👍

implementations/rust/ockam/ockam_app_lib/src/background_node.rs
davide-baldo: done! 👍

implementations/rust/ockam/ockam_app_lib/src/background_node.rs
davide-baldo: i didn't need either in the end!

implementations/rust/ockam/ockam_app_lib/src/state/mod.rs
adrianbenavides: Nice, never had the chance to see `OnceLock` in action yet, looks great!

adrianbenavides (approved):
LGTM! 🚀

PR #6665, opened 2023-10-19
feat(rust): make node show interactive

### LINKED ISSUE #6471: Make `ockam node show` (no args) interactive by asking the user to choose from a list of node names to show (tuify)
# Current behavior

When a user runs `ockam node show` without any arguments, currently they are presented with information about the default node (if one exists). Here's a screenshot of this.

Image

If no default node exists, then this error is displayed:

Image

# Desired behavior

Change the behavior of this command so that it becomes interactive with the user when no arguments are passed. Note that this can only be allowed to happen if:

- the terminal is interactive, and,
- the [`--no-input`](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/terminal/mod.rs#L265) or `--quiet` flags have *not* been passed, and,
- no arguments have been passed in to the command.

This what the user flow should look like in interactive mode.

If there are no nodes, then simply print a message letting the user know that there are no nodes to display. This is not an error.

If one or more nodes exist, then:

1) Use the [tuify](https://crates.io/crates/r3bl_tuify) crate's `select_from_list()` function to ask the user to select from a list of node names. They should be able to make multiple selections.

2) Once the user has made their selection(s), then run `show` for each selected node, and display the results of the `show` command for each node that got selected. 
   - They should be able to make a single selection.
   - If they have made no selections then exit by displaying a message to the user letting them know they did not make any selection.

> The following is a video showing a generic user flow that shows the user experience when a user is asked to choose a from a list of "things". Then an action is performed on each "thing". Finally status information is reported for each action that is performed on each "thing". Please note that this flow is *not* specific to this issue. 
> Image
>  https://asciinema.org/a/612547

# Implementation details

- Please use the functions `select_multiple`, and `confirm_interactively` in this [file](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/terminal/mod.rs#L281).

- Here's a similar PR to get some ideas from: https://github.com/build-trust/ockam/pull/6480

- You can learn more about how to use the `select_from_list()` in the tuify crate [here](https://docs.rs/r3bl_tuify/latest/r3bl_tuify/). Here is an [example](https://github.com/r3bl-org/r3bl_rs_utils/blob/ockam-demo/tuify/examples/main_interactive.rs) that is very similar to the steps you might have to take for this issue.
- Here's a [file](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/terminal/mod.rs#L255) you can look at to detect when a terminal is interactive or not.

- Here's a [file](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/node/show.rs#L51) you can look at to see where the `ockam node show` command is implemented.

---
<i>

We love helping new contributors! ❤️
If you have questions or need help as you explore, please join us on [Discord](https://discord.gg/bewvnm6zqS). If you're looking for other issues to contribute to, please checkout our [good first issues](https://github.com/build-trust/ockam/labels/good%20first%20issue).
</i>

mickm3n:
Close https://github.com/build-trust/ockam/issues/6471

## Current behavior

When a user runs ockam node show without any arguments, currently they are presented with information about the default node (if one exists). Here's a screenshot of this.

Image

If no default node exists, then this error is displayed:
 
Image

## Proposed changes

Change the behavior of this command so that it becomes interactive with the user when no arguments are passed.

User flow
- if an argument is provided
  - show specific node (align with current behavior)
- if the user terminal is not interactive
  - show the default node (align with current behavior)
- if the user terminal is interactive
  - if there are no nodes
image
  - if there are one or more nodes, let the user select 
image
    - if the user selects nothing
image
    - if the user selects some, let the user confirm
image
      - if the user confirms
image

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/show.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/show.rs
@@ -12,6 +12,7 @@ use ockam_api::cli_state::{CliState, StateDirTrait, StateItemTrait};
 use ockam_api::nodes::models::portal::{InletList, OutletList};
 
 use crate::node::get_node_name;
+use crate::node::list;
 use crate::node::util::check_default;
 use crate::util::{api, node_rpc};
 use crate::{docs, CommandGlobalOpts, Result};
@@ -52,10 +53,51 @@ async fn run_impl(
     ctx: Context,
     (opts, cmd): (CommandGlobalOpts, ShowCommand),
 ) -> miette::Result<()> {
-    let node_name = get_node_name(&opts.state, &cmd.node_name);
-    let mut node = BackgroundNode::create(&ctx, &opts.state, &node_name).await?;
-    let is_default = check_default(&opts, &node_name);
-    print_query_status(&opts, &ctx, &node_name, &mut node, false, is_default).await?;
+    if cmd.node_name.is_some() || !opts.terminal.can_ask_for_user_input() {
+        show_node(&opts, &ctx, &get_node_name(&opts.state, &cmd.node_name)).await?;
+        return Ok(());
+    }
+
+    let node_names = opts.state.nodes.list_items_names()?;
+    match node_names.len() {
+        0 => {
+            opts.terminal
+                .stdout()
+                .plain("There are no nodes to show")
+                .write_line()?;
+        }
+        1 => {
+            show_node(&opts, &ctx, &node_names[0]).await?;
+        }
+        _ => {
+            let selected_node_names = opts.terminal.select_multiple(
+                "Select one or more nodes that you want to show".to_string(),
+                node_names,
+            );
+            match selected_node_names.len() {
+                0 => {
+                    opts.terminal
+                        .stdout()
+                        .plain("No nodes selected to show")
+                        .write_line()?;
+                }
+                1 => {
+                    show_node(&opts, &ctx, &selected_node_names[0]).await?;
+                }
+                _ => {
+                    let nodes = list::get_nodes_info(ctx, &opts, selected_node_names).await?;
+                    list::print_nodes_info(&opts, nodes)?;
+                }
+            }
+        }
+    }
+    Ok(())
+}
+
+async fn show_node(opts: &CommandGlobalOpts, ctx: &Context, node_name: &str) -> miette::Result<()> {
+    let mut node = BackgroundNode::create(ctx, &opts.state, node_name).await?;
+    let is_default = check_default(opts, node_name);
+    print_query_status(opts, ctx, node_name, &mut node, false, is_default).await?;
     Ok(())
 }
 
--- a/implementations/rust/ockam/ockam_command/src/node/list.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/list.rs
@@ -12,6 +12,7 @@ use ockam_api::cli_state::StateDirTrait;
 use ockam_api::nodes::models::base::NodeStatus;
 use ockam_api::nodes::BackgroundNode;
 
+use crate::node::get_default_node_name;
 use crate::output::Output;
 use crate::terminal::OckamColor;
 use crate::util::{api, node_rpc};
@@ -46,17 +47,24 @@ async fn run_impl(
     // one in config, we update the pid stored in the config.
     // This should only happen if the node has failed in the past,
     // and has been restarted by something that is not this CLI.
-    let mut default = String::new();
     let node_names: Vec<_> = {
         let nodes_states = opts.state.nodes.list()?;
-        // default node
-        if let Ok(state) = opts.state.nodes.default() {
-            default = state.name().to_string();
-        }
         nodes_states.iter().map(|s| s.name().to_string()).collect()
     };
 
+    let nodes = get_nodes_info(ctx, &opts, node_names).await?;
+    print_nodes_info(&opts, nodes)?;
+
+    Ok(())
+}
+
+pub async fn get_nodes_info(
+    ctx: Context,
+    opts: &CommandGlobalOpts,
+    node_names: Vec<String>,
+) -> Result<Vec<NodeListOutput>> {
     let mut nodes: Vec<NodeListOutput> = Vec::new();
+    let default_node_name = get_default_node_name(&opts.state);
     for node_name in node_names {
         let node = BackgroundNode::create(&ctx, &opts.state, &node_name).await?;
 
@@ -99,10 +107,17 @@ async fn run_impl(
             node_status.node_name.to_string(),
             node_status.status.to_string(),
             node_status.pid,
-            node_status.node_name == default,
+            node_status.node_name == default_node_name,
         ));
     }
 
+    Ok(nodes)
+}
+
+pub fn print_nodes_info(
+    opts: &CommandGlobalOpts,
+    nodes: Vec<NodeListOutput>,
+) -> miette::Result<()> {
     let plain = opts
         .terminal
         .build_list(&nodes, "Nodes", "No nodes found on this system.")?;
@@ -110,6 +125,7 @@ async fn run_impl(
     let json = serde_json::to_string_pretty(&nodes).into_diagnostic()?;
 
     opts.terminal
+        .clone()
         .stdout()
         .plain(plain)
         .json(json)


implementations/rust/ockam/ockam_command/src/node/show.rs
```diff
@@ -52,10 +52,51 @@ async fn run_impl(
     ctx: Context,
     (opts, cmd): (CommandGlobalOpts, ShowCommand),
 ) -> miette::Result<()> {
-    let node_name = get_node_name(&opts.state, &cmd.node_name);
+    if let Some(node_name) = cmd.node_name {
+        show_node(&opts, &ctx, &node_name).await?;
+    } else if !opts.terminal.can_ask_for_user_input() {
+        show_node(&opts, &ctx, &get_default_node_name(&opts.state)).await?;
```
adrianbenavides: These two blocks could be reduced to one, and you can also remove the `else` block to avoid an unnecessary indentation level (which makes the code easier to read).

```rust
if !opts.terminal.can_ask_for_user_input() {
    show_node(&opts, &ctx, &get_node_name(&opts.state, &cmd.node_name)).await?;
    return Ok(());
}

let node_names = opts.state.nodes.list_items_names()?;
...
```

implementations/rust/ockam/ockam_command/src/node/show.rs
```diff
@@ -52,10 +52,51 @@ async fn run_impl(
...
+            opts.terminal
+                .stdout()
+                .plain("No nodes selected to show")
+                .write_line()?;
+            return Ok(());
+        }
+
+        if opts.terminal.confirm_interactively(format!(
+            "Would you like to show these items : {:?}?",
+            selected_node_names
```
adrianbenavides: You can use the following notation here to make it less verbose: `format!("Would you like to show these items: {selected_node_names:?}?")`

implementations/rust/ockam/ockam_command/src/node/show.rs
```diff
@@ -52,10 +52,51 @@ async fn run_impl(
...
+        )) {
+            for selected_node_name in selected_node_names {
+                show_node(&opts, &ctx, &selected_node_name).await?;
```
adrianbenavides: You should use the `NodeListOutput` struct to print the list of nodes, which shows less information per item. Perhaps the `show_node` function extraction is not necessary if you do this.

implementations/rust/ockam/ockam_command/src/node/show.rs
```diff
@@ -52,10 +52,51 @@ async fn run_impl(
...
+    } else {
+        let node_names = opts.state.nodes.list_items_names()?;
+        if node_names.is_empty() {
+            opts.terminal
+                .stdout()
+                .plain("There are no nodes to show")
+                .write_line()?;
+            return Ok(());
+        }
+
```
adrianbenavides: If the `node_names` list only contains one item, it should be shown right away without interaction. This would mean that you have three cases: the user has selected either 0, 1 or >1. So maybe something like this would be more explicit:

```rust
match node_names.len() {
    0 => {
        opts.terminal
            .stdout()
            .plain("There are no nodes to show")
            .write_line()?;
    }
    1 => {
        show_node(&opts, &ctx, &node_names[0]).await?;
    }
    _ => {
        let selected_node_names = opts.terminal.select_multiple(
            "Select one or more nodes that you want to show".to_string(),
            node_names,
        );
        ...
    }
}
```

adrianbenavides (commented):
Hi @mickm3n, thanks for submitting another PR 🙏 

I've added some comments on how to improve the structure and support some edge cases. Other than that looks great!

implementations/rust/ockam/ockam_command/src/node/show.rs
mickm3n: Hi @adrianbenavides, I have a question the result between `node list` and `node show` is quite different. 

`node list`
image

`node show`
image

I'd like to clarify whether it is expected to see the `node list` result when selecting multiple nodes in `node show`?
If so, should I directly use the function and struct from list.rs in show.rs? Thank you.

implementations/rust/ockam/ockam_command/src/node/show.rs
```diff
@@ -52,10 +52,54 @@ async fn run_impl(
...
+            if selected_node_names.is_empty() {
+                opts.terminal
+                    .stdout()
+                    .plain("No nodes selected to show")
+                    .write_line()?;
+                return Ok(());
+            }
+
+            if opts.terminal.confirm_interactively(format!(
```
adrianbenavides: I've been testing the command manually and I think this check here should be removed. In this case we are just showing data, it's not like in the `node delete` command where we want the user to confirm the action.

implementations/rust/ockam/ockam_command/src/node/show.rs
```diff
@@ -52,10 +52,54 @@ async fn run_impl(
     ctx: Context,
     (opts, cmd): (CommandGlobalOpts, ShowCommand),
 ) -> miette::Result<()> {
-    let node_name = get_node_name(&opts.state, &cmd.node_name);
+    if !opts.terminal.can_ask_for_user_input() {
+        show_node(&opts, &ctx, &get_node_name(&opts.state, &cmd.node_name)).await?;
+        return Ok(());
+    }
+
```
adrianbenavides, line 60: We forgot about one case here, after this `if`:

```rust
if let Some(node_name) = &cmd.node_name {
    show_node(&opts, &ctx, node_name).await?;
    return Ok(());
}
```

Perhaps we could even merge both, since, if the user passes the `node_name` arg, then we want to show it without any interactivity:

```rust
if cmd.node_name.is_some() || !opts.terminal.can_ask_for_user_input() {
    show_node(&opts, &ctx, &get_node_name(&opts.state, &cmd.node_name)).await?;
    return Ok(());
}
```

adrianbenavides (commented):
We are pretty close now! I found a couple of issues to fix, we can merge after that 👍

implementations/rust/ockam/ockam_command/src/node/show.rs
mickm3n: I'm good with this. It's the user flow described in https://github.com/build-trust/ockam/issues/6471 by @nazmulidris 
I'll create a commit first.

implementations/rust/ockam/ockam_command/src/node/show.rs
mickm3n: fixed in [aa2ff2f](https://github.com/build-trust/ockam/pull/6665/commits/aa2ff2f392027e321082531c603b5644eb5372bf)

implementations/rust/ockam/ockam_command/src/node/show.rs
mickm3n: commit [e029ade](https://github.com/build-trust/ockam/pull/6665/commits/e029adef4b45f8961acba268e2cefae018c4317d).

implementations/rust/ockam/ockam_command/src/node/show.rs
adrianbenavides: Sorry for the delay in answering this question. Yeah, I think we should use the `list` format when showing multiple nodes. Good catch!

adrianbenavides (commented):
It works perfectly now @mickm3n , we can merge after solving the last comment we have left.

implementations/rust/ockam/ockam_command/src/node/show.rs
nazmulidris: @adrianbenavides I agree 👍🏽 ; I think removing the confirmation prompt might make make more sense in this case, since the operation is non destructive. In the case where the operation is destructive it does make sense to ask for confirmation.

I can update the existing issues to reflect this change in user flow that @mickm3n was referring to (from the issues).

> - I currently have 2 variants of tuify issues. One where in the final step (after getting user confirmation) a checkbox is displayed showing whether the operation succeeded or not. And another where no checkbox is displayed.
> - I will add another variant, where the operation that is performed is non-destructive, so there might not be any need to get confirmation from the user before proceeding.

implementations/rust/ockam/ockam_command/src/node/show.rs
mickm3n: Hi @adrianbenavides, to print `node list` format in `node show`, I made some `node::list` functions public. Please help review again. Thanks!

adrianbenavides (approved):
Hey @mickm3n, sorry for not reviewing it sooner. Looks perfect now, thank you so much for your help with this PR, it will get merged soon! 🙏

metaclips (approved):

PR #6728, opened 2023-10-23
hide desktop ui from application dock

metaclips:
This PR allows us to hide the ockam app from the application dock and make it only visible on the menubar. [We use the LSUIElement attribute from Swift](https://developer.apple.com/documentation/bundleresources/information_property_list/lsuielement)
- <h3>Before</h3>
Screenshot 2023-10-23 at 2 00 14 AM

- <h3>After</h3>

Screenshot 2023-10-23 at 2 01 13 AM


davide-baldo (approved):
super nice!

PR #6730, opened 2023-10-23
Build and CI for swift app

davide-baldo:
- Added Makefile
- Added Package DMG creation with **bundled** command, custom background for installer (both SVG and PNG). 
    full package with `make swift_package`
- Added swift package build to the release flow


metaclips (approved):
🚀

PR #6760, opened 2023-10-25
feat(rust): polishing desktop app paper cuts

davide-baldo:
Removed few paper cuts of the application:

- using colored logo (only for the app icon, not tray)
- relay name is embedding the identifier to avoid conflicts whenever we reset or have multiple applications on the same account
- pending invitations: showing a orange circle when a invitation and a notification when a new comes in (as well as during bootstrap)
- change ui view for the service group
- added log handling in ~/.ockam/node/ockam_app/stdout.log
- reset visible only when pressing option


adrianbenavides:
Closing in favor of: https://github.com/build-trust/ockam/pull/6773

PR #6771, opened 2023-10-25
feat: use a dynamic relay name on the app based on the default identifier

adrianbenavides:
Based on https://github.com/build-trust/ockam/pull/6760

- fix: relay name is now based on the default identifier
- fix: message shown when a new project is created
- feat: improve ui of services dropdown


nazmulidris (approved):
LGTM! 🎉

PR #6783, opened 2023-10-26
add openurlview file to list of build file in swift

metaclips:
This PR adds the openurlview file to list of swift build


nazmulidris (approved):
👍🏽

PR #6809, opened 2023-10-30
feat(rust): using identifier as relay name and related fixes

davide-baldo:
fixed relay related issues:

relay connection was kept alive even after reset
relay delete didn't remove the monitoring (it was automatically re-spanwed)
relay check is now a scheduled service like others in the app


implementations/rust/ockam/ockam_app_lib/src/shared_service/relay/create.rs
```diff
@@ -13,29 +14,44 @@ use tracing::{debug, info, trace, warn};
...
+        let context = self.context();
+
         if !self.is_enrolled().await.unwrap_or(false) {
+            self.update_orchestrator_status_if(
+                OrchestratorStatus::Disconnected,
+                vec![
+                    OrchestratorStatus::Connected,
+                    OrchestratorStatus::Connecting,
+                ],
+            );
```
adrianbenavides, line 31: If we want to switch to `Disconnected`, why not do it unconditionally and remove that function call?

implementations/rust/ockam/ockam_app_lib/src/shared_service/relay/create.rs
```diff
@@ -47,28 +63,26 @@ impl AppState {
...
+                    self.publish_state().await;
                     debug!(project = %project.name(), "Creating relay at project");
                     let project_route = format!("/project/{}", project.name());
                     let project_address = MultiAddr::from_str(&project_route).into_diagnostic()?;
                     let relay = node_manager
                         .create_relay(
                             context,
                             &project_address,
-                            Some(relay_alias(project.config())),
+                            Some(relay_name(cli_state)?),
```
adrianbenavides: Just to double-check, the full relay name is `forward_to_{alias}`, but when you call the `create_relay` function you must pass the `alias` only, without the `forward_to` prefix.

Perhaps the function is stripping the prefix just in case?

adrianbenavides (commented):
Looks good, I left a couple of comments/questions 👍

implementations/rust/ockam/ockam_app_lib/src/shared_service/relay/create.rs
adrianbenavides: I checked the logs and we are creating relays with the following name: `key=relay-forward_to_forward_to_Ibf5f10fc2add4dc091cf194813ccf32173aa5715`

implementations/rust/ockam/ockam_app_lib/src/shared_service/relay/create.rs
davide-baldo: if the relay refresh gets called during an enrollment, the enrollment phases would be overwritten by the `Disconnected` status.
Ideally, the connection status and the enrollment status should be separated, it's unified for simplicity.

adrianbenavides: Ok got it!

implementations/rust/ockam/ockam_app_lib/src/shared_service/relay/create.rs
davide-baldo: nice catch! thanks

adrianbenavides (approved):
🚀

PR #6818, opened 2023-10-31
feat(rust): restarts the whole swift application after a reset

davide-baldo:
To avoid issues of broken state after a reset, simply restart the whole application.


adrianbenavides (approved):
great!

PR #6819, opened 2023-10-31
fix(rust): loads app state from the database when initializing the app

adrianbenavides:
The `ModelState` wasn't being loaded from the `ModelStateRepository` in the `AppState` initialization.


davide-baldo (approved):
Looks good!

PR #6823, opened 2023-11-01
feat(rust): add `timeout` arg to `tcp-inlet create`

adrianbenavides:
The timeout of the `tcp-inlet create` command can be controlled in two ways:
- `timeout`: it controls the high-level timeout of the request that the CLI does to the node handling the request.
- `connection-wait`: it controls the timeout for resolving the inlet's target route, which includes creating the secure channel to the project and create a tcp connection to the service. An example of this route could be `/project/default/service/forward_to_other_node/secure/api/service/myservice`

Additionally, we have the `retry_wait` argument, which can be used to loop until the tcp-inlet creation succeeds (it can be disabled by setting it to `0`.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
@@ -67,6 +67,10 @@ pub struct CreateCommand {
     /// Time to wait before retrying to connect to outlet.
     #[arg(long, display_order = 900, id = "RETRY", default_value = "20s", value_parser = duration_parser)]
     retry_wait: Duration,
+
+    /// Override default timeout
+    #[arg(long, value_parser = duration_parser)]
+    timeout: Option<Duration>,
 }
 
 pub(crate) fn default_from_addr() -> SocketAddr {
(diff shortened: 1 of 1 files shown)


implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
```diff
@@ -67,6 +67,10 @@ pub struct CreateCommand {
     /// Time to wait before retrying to connect to outlet.
     #[arg(long, display_order = 900, id = "RETRY", default_value = "20s", value_parser = duration_parser)]
     retry_wait: Duration,
+
+    /// Override default timeout
+    #[arg(long, value_parser = duration_parser)]
+    timeout: Option<Duration>,
```
davide-baldo, line 73: What if we also specify the default by referencing the node manager default timeout?

davide-baldo (approved):

implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
adrianbenavides: This might not be straightforward to fix, I'll address it in  separate PR.

PR #6824, opened 2023-11-01
feat(rust): setup app's logs with the same features we use in the cli

adrianbenavides:
The app now supports file rotation and can be customized with the log env variables we use in the CLI:

https://github.com/build-trust/ockam/blob/1ca0e18cf019d0f6e6702cd800901ea7652fef30/implementations/rust/ockam/ockam_command/src/environment/static/env_info.txt#L14-L17

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/logs/env.rs
+++ b/implementations/rust/ockam/ockam_api/src/logs/env.rs
@@ -1,0 +1,21 @@
+use super::LogFormat;
+use ockam_core::env::{get_env, get_env_with_default};
+
+pub fn log_level() -> Option<String> {
+    get_env("OCKAM_LOG").unwrap_or_default()
+}
+
+pub fn log_max_size_bytes() -> u64 {
+    let default = 100;
+    get_env_with_default("OCKAM_LOG_MAX_SIZE_MB", default).unwrap_or(default) * 1024 * 1024
+}
+
+pub fn log_max_files() -> usize {
+    let default: u64 = 60;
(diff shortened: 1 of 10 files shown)


implementations/rust/ockam/ockam_api/src/logs/env.rs
```diff
@@ -0,0 +1,21 @@
+use super::LogFormat;
+use ockam_core::env::{get_env, get_env_with_default};
+
+pub fn log_level() -> Option<String> {
+    get_env("OCKAM_LOG").unwrap_or_default()
+}
+
+pub fn log_max_size() -> u64 {
+    let default = 100;
+    get_env_with_default("OCKAM_LOG_MAX_SIZE_MB", default).unwrap_or(default)
```
davide-baldo: We could convert the limit into bytes at this level

implementations/rust/ockam/ockam_app_lib/src/state/mod.rs
```diff
@@ -71,6 +72,7 @@ pub struct AppState {
     refresh_inlets_scheduler: Arc<OnceLock<Scheduler>>,
     refresh_relay_scheduler: Arc<OnceLock<Scheduler>>,
     last_published_snapshot: Arc<Mutex<Option<ApplicationState>>>,
+    pub tracing_guard: Arc<OnceLock<WorkerGuard>>,
```
davide-baldo: `pub` shouldn't be needed

davide-baldo (dismissed):

nazmulidris (approved):
LGTM 🎉

PR #6839, opened 2023-11-03
refactor(rust): use `BackgroundNode` to handle tcp-inlets within the app

adrianbenavides:
- Add an `Inlets` trait that is implemented by the `BackgroundNode` struct
- Refactor crud `tcp-inlet` commands to use the `BackgroundNode`
- Refactor crud `tcp-inlet` CLI usages from the app to use the `BackgroundNode`
- Refactor `node delete` CLI usages from the app


etorreborre (approved):
One more step towards the full removal of the command-line!

PR #6841, opened 2023-11-03
feat(elixir): option to terminate tcp transports due to inactivity

polvorin:
by default transport are not terminated until the underling tcp socket is closed.  This PR adds an option to close the socket and terminate the transport after a configured period of inactivity.  It's mainly a safety valve on server apps, in cases where a client leaks connections.

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/handler.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/handler.ex
@@ -35,6 +35,11 @@ defmodule Ockam.Transport.TCP.Handler do
     authorization = Keyword.get(handler_options, :authorization, [])
     tcp_wrapper = Keyword.get(handler_options, :tcp_wrapper, Ockam.Transport.TCP.DefaultWrapper)
 
+    # There are cases where we want to detect and close tcp connections after a long
+    # period without activity.  Mainly to handle the case of a server receiving connections
+    # from a long running client that "leak" them without closing.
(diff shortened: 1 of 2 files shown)


implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/handler.ex
```diff
@@ -35,6 +35,11 @@ defmodule Ockam.Transport.TCP.Handler do
     authorization = Keyword.get(handler_options, :authorization, [])
     tcp_wrapper = Keyword.get(handler_options, :tcp_wrapper, Ockam.Transport.TCP.DefaultWrapper)
 
+    # There are cases where we want to detect and close tcp connections after a long
+    # period without activity.  Mainly to handle the case of a server receiving connections
+    # from a long running client that "leak" them without closing.
+    idle_timeout = Keyword.get(handler_options, :idle_timeout, :infinity)
```
davide-baldo, line 41: I wonder if we should avoid `:infinity` and use a big timeout instead, like 30 days of inactivity.

davide-baldo (approved):

implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/handler.ex
polvorin: 🤔  it might be a good idea indeed, but since we don't have explicit expectations about how transports (or channels) are used, I didn't want to change the existing behavior in any way unless configured to do so.    Apps using it can put these 30 days on their config.  (in most cases I bet 30 days won't even be reached as nodes are restarted, etc)

PR #6842, opened 2023-11-03
feat(rust): automatically drop tcp workers when dropping secure clients

davide-baldo:
When resolving a `MultiAddr` to a `Route` it usually involves creating a new TCP connection. In most cases, this connection (or its relative worker) is not being tracked and properly stopped after use.

This is a partial solution for at least secure clients: controller, authority, project and an unused generic.

More should be done to address every instantiation of a `Route`, this is just a first step to avoiding socket leaks in the application which frequently create secure clients.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cloud/secure_clients.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/secure_clients.rs
@@ -1,17 +1,19 @@
 use std::str::FromStr;
 use std::time::Duration;
+use tokio::spawn;
 
 use ockam::identity::{Identifier, SecureChannel, SecureChannels, SecureClient, DEFAULT_TIMEOUT};
 use ockam_core::compat::sync::Arc;
 use ockam_core::env::{get_env, get_env_with_default, FromString};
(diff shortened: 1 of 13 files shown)


implementations/rust/ockam/ockam_api/src/cloud/secure_clients.rs
```diff
@@ -87,75 +96,116 @@ impl NodeManager {
     }
 
     pub async fn controller_node(
+        context: &Context,
```
etorreborre: @davide-baldo I think that you could store the `TcpTransport` inside the client directly since it already contains a copy of the `Context`. Then there's no need to pass a context everywhere to instantiate the clients. It was my original intention that we would only need the `Context` when using the clients, not creating them (at least not explicitly since it is implicitly provided via the Transport...)

davide-baldo: I could store the transport directly, but I would need to clone it anyway, I can reduce the operation by cloning only the context from the transport, so I can have the best of both world!

etorreborre (approved):
🏥

PR #6843, opened 2023-11-03
Secure Channel Protocol

SanjoDeundiak:
1. Add dedicated struct for SecureChannel message type
2. Implement Close message
3. Implement credentials refresh
4. Use more compact CBOR representation of some Identity-related structs (arrays instead of maps)
5. Use full SHA256 as `Identifier` and `ChangeHash`
6. Add `data_type` field to `VersionedData` struct


implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_state_machine.rs
```diff
@@ -116,77 +116,144 @@ impl CommonStateMachine {
...
+                    PurposePublicKey::CredentialSigning(_) => {
+                        return Err(IdentityError::InvalidKeyType.into())
+                    }
                 }
             }
-            PurposePublicKey::CredentialSigning(_) => {
-                return Err(IdentityError::InvalidKeyType.into())
+            (None, None) => {}
+            _ => {
+                return Err(IdentityError::InvalidKeyData.into());
```
etorreborre: It would help debugging to have a logging line saying in which case we ended-up: `(Some(_), None)` or `(None, Some(_))`

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_state_machine.rs
```diff
@@ -116,77 +116,144 @@ impl CommonStateMachine {
...
+        if let Some(trust_policy) = trust_policy {
+            // check our TrustPolicy
+            let trust_info = SecureChannelTrustInfo::new(their_identifier.clone());
+            let trusted = trust_policy.check(&trust_info).await?;
+            if !trusted {
+                // TODO: Shutdown? Communicate error?
+                return Err(IdentityError::SecureChannelTrustCheckFailed.into());
+            }
+            debug!(
+                "Initiator checked trust policy for SecureChannel from: {}",
```
etorreborre: Can we pass the specific role here?

implementations/rust/ockam/ockam_identity/src/secure_channel/options.rs
```diff
@@ -11,6 +11,12 @@ use core::fmt;
 use core::fmt::Formatter;
 use core::time::Duration;
 
+/// This is the interval before a credential expiration when we'll query for a new credential.
+pub const DEFAULT_REFRESH_CREDENTIAL_TIME_GAP: Duration = Duration::from_secs(60);
+
+/// Minimal interval before
```
etorreborre: before? Should it say "Minimum duration between 2 credential refreshes" (to avoid too frequent refreshes I guess)

implementations/rust/ockam/ockam_identity/src/secure_channel/options.rs
```diff
@@ -21,6 +27,8 @@ pub struct SecureChannelOptions {
     pub(crate) trust_context: Option<TrustContext>,
     pub(crate) credentials: Vec<CredentialAndPurposeKey>,
     pub(crate) timeout: Duration,
+    pub(crate) min_refresh_credential_interval: Duration,
+    pub(crate) credential_refresh_time_gap: Duration,
 }
```
etorreborre, line 33: For consistency I propose to name that field `refresh_credential_time_gap`

implementations/rust/ockam/ockam_identity/src/secure_channels/secure_channels.rs
```diff
@@ -113,7 +147,9 @@ impl SecureChannels {
             purpose_key,
             options.trust_policy,
             access_control.decryptor_outgoing_access_control,
-            options.credentials,
+            credentials,
+            options.min_refresh_credential_interval,
```
etorreborre: I wonder if we shouldn't pass a `SecureChannelTrustOptions` struct containing all the options for credentials and trust (and have `SecureChannelListenerOptions` and `SecureChannelOptions` contain that struct)

implementations/rust/ockam/ockam_identity/tests/credentials_refresh.rs
```diff
@@ -0,0 +1,373 @@
...
+use ockam_identity::models::{CredentialAndPurposeKey, CredentialSchemaIdentifier};
+use ockam_identity::secure_channels::secure_channels;
+use ockam_identity::utils::AttributesBuilder;
+use ockam_identity::{
+    AuthorityService, Credentials, CredentialsRetriever, Identifier, IdentityError,
+    SecureChannelListenerOptions, SecureChannelOptions, TrustContext,
+};
+use ockam_node::Context;
+
+struct LocalCredentialsRetriever {
```
etorreborre: I wouldn't mind reading this code after the tests in order to have more context :-)
Also one option would be to have:

- a `LocalCredentialsRetriever` returning credentials with no issue, exposing its call counter and used in test 2 and test 3
- a `FailingLocalCredentialsRetriever` wrapping the `LocalCredentialsRetriever` just to add the failing behaviour based one counters

implementations/rust/ockam/ockam_identity/tests/credentials_refresh.rs
```diff
@@ -0,0 +1,373 @@
...
+            route!["listener"],
+            SecureChannelOptions::new()
+                .with_trust_context(trust_context1)
+                .with_credential_refresh_time_gap(Duration::from_secs(1)),
+        )
+        .await?;
+
+    ctx.sleep(Duration::from_secs(10)).await;
+
+    assert_eq!(call_counter1.load(Ordering::Relaxed), 4);
```
etorreborre: I would be good to have a comment to tell why those values should be `4` and `3`.

implementations/rust/ockam/ockam_identity/tests/credentials_refresh.rs
```diff
@@ -0,0 +1,373 @@
...
+        .added();
+
+    ctx.sleep(Duration::from_millis(3_100)).await;
+    let added4 = attributes_reader
+        .get_attributes(&client1)
+        .await?
+        .unwrap()
+        .added();
+
+    assert!(added4 > added3);
```
etorreborre: Can we reorder the assertions as
```
assert!(added1 < added2);
assert_eq!(added2, added3);
assert!(added3 < added4);
```

implementations/rust/ockam/ockam_identity/tests/credentials_refresh.rs
```diff
@@ -0,0 +1,373 @@
...
+}
+
+#[ockam_macros::test]
+async fn autorefresh(ctx: &mut Context) -> Result<()> {
+    let secure_channels = secure_channels();
+    let identities = secure_channels.identities();
+    let identities_creation = identities.identities_creation();
+    let credentials = identities.credentials();
+
+    let authority = identities_creation
```
etorreborre: Since it takes a lot of vertical space we could have a function `create_identity(identities_creation)` returning an `Identifier`.

implementations/rust/ockam/ockam_identity/tests/credentials_refresh.rs
```diff
@@ -0,0 +1,373 @@
...
+                .credentials
+                .credentials_creation()
+                .issue_credential(&self.authority, &self.client, attributes, self.ttl)
+                .await?;
+            self.call_counter.fetch_add(1, Ordering::Relaxed);
+            Ok(credential)
+        }
+    }
+
+    let call_counter2 = Arc::new(AtomicU8::new(0));
```
etorreborre: Since we're creating `client1` and `client2` in order it would be good to do the same for their trust contexts.
And possibly have a function creating a trust context just based on a credentials retriever.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -158,36 +163,46 @@ impl HandshakeWorker {
...
         remote_route: Option<Route>,
         timeout: Option<Duration>,
         role: Role,
     ) -> Result<()> {
         let vault = secure_channels.identities.vault().secure_channel_vault;
         let identities = secure_channels.identities();
+
+        let min_credential_expiration = credentials
+            .iter()
+            .filter_map(|credential| credential.credential.get_versioned_data().ok())
```
etorreborre: It would be nice to have a function `expires_at()` which does all of this on `CredentialAndPurposeKey` directly

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -110,14 +206,76 @@ impl EncryptorWorker {
...
+            self.remote_route.clone(),
+            msg,
+            self.addresses.encryptor.clone(),
+        )
+        .await?;
+
+        Ok(())
+    }
+
+    async fn schedule_credentials_refresh(&mut self, ctx: &Context, is_retry: bool) -> Result<()> {
```
etorreborre, line 249: It would be nice to have a textual explanation of the logic in this function

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -110,14 +206,76 @@ impl EncryptorWorker {
...
+            let mut credential_refresh_event =
+                DelayedEvent::create(ctx, self.addresses.encryptor_internal.clone(), ()).await?;
+
+            let now = now()?;
+
+            let duration = if min_credential_expiration < now + self.refresh_credential_time_gap {
+                self.min_refresh_credential_interval
+            } else {
+                Duration::from_secs(
+                    *(min_credential_expiration
```
etorreborre: It seems that we can compute this quantity once and use it in the `if` condition and in that branch. 
Otherwise I don't fully understand the logic. If `min_credential_expiration < now + self.refresh_credential_time_gap`, doesn't that mean that we are past the time for refreshing the credential and we should then refresh now and not in `min_refresh_credential_interval`?

I also don't get the meaning of `is_retry`. What are we retrying with this boolean?

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -91,14 +120,81 @@ impl EncryptorWorker {
...
+
+        let credential = if let Some(credentials_retriever) = &self.credentials_retriever {
+            match credentials_retriever
+                .credential(ctx, &self.my_identifier)
+                .await
+            {
+                Ok(credential) => credential,
+                Err(err) => {
+                    info!(
+                        "Credentials refresh failed for {} and is rescheduled in {}",
```
etorreborre: `in {} seconds` would be more precise

implementations/rust/ockam/ockam_identity/src/secure_channel/options.rs
```diff
@@ -11,6 +11,12 @@ use core::fmt;
 use core::fmt::Formatter;
 use core::time::Duration;
 
+/// This is the interval before a credential expiration when we'll query for a new credential.
```
etorreborre: I know it might be obvious to you but it could be good to add a comment about why we need this kind of parameter.
My guess is that we want to make sure that we don't end-up with an expired credential.

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -110,14 +206,76 @@ impl EncryptorWorker {
...
+        .await?;
+
+        Ok(())
+    }
+
+    async fn schedule_credentials_refresh(&mut self, ctx: &Context, is_retry: bool) -> Result<()> {
+        if let Some(min_credential_expiration) = self.min_credential_expiration {
+            self.credential_refresh_event = None;
+
+            let mut credential_refresh_event =
```
etorreborre: We could create this event closer to L.274

etorreborre (dismissed):
I left some comments on ideas for refactoring the code and the tests.
More importantly I would appreciate more comments / explanation around the scheduling logic to understand all the various cases.

One possibility would be to extract that logic as a function of `now`, `min_refresh_credential_interval`, `refresh_credential_time_gap`, `is_retry`, returning a `Duration` to schedule the `DelayedEvent`. 
Then have some unit tests showing the various outcomes depending on combination of inputs.

implementations/rust/ockam/ockam_identity/tests/credentials_refresh.rs
SanjoDeundiak: The plan is to change the return type to `Identifier` in all existing public functions. I just didn't get to this change yet

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_state_machine.rs
SanjoDeundiak: I changed the signature to avoid that case at all

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_state_machine.rs
SanjoDeundiak: That's a little bit problematic due to delegate macro you used. I'll remove the role from the message instead.

implementations/rust/ockam/ockam_identity/tests/credentials_refresh.rs
SanjoDeundiak: My C++ background doesn't allow me to declare something after it's used :) I'll move it
But I don't see a point of having 2 implementations, looks like excess almost copy&paste

implementations/rust/ockam/ockam_identity/tests/credentials_refresh.rs
SanjoDeundiak: Renamed to client and server

implementations/rust/ockam/ockam_identity/tests/credentials_refresh.rs
etorreborre: There are already 2 implementations in the test code :-) ([L.49](https://github.com/build-trust/ockam/pull/6843/files#diff-3a42c5c069a583c53235d3eb1ddc5ed61966d334ce64ad9bcfda2667354e374eR49) and [L.105](https://github.com/build-trust/ockam/pull/6843/files#diff-3a42c5c069a583c53235d3eb1ddc5ed61966d334ce64ad9bcfda2667354e374eR105)).
My proposal is to wrap one implementation the other so that there's no duplication of behaviour:
```rust
struct FailingLocalCredentialsRetriever {
    credential_retriever: LocalCredentialsRetriever,
    fail_iteration: Option<u8>,
    failed_call_counter: Option<Arc<AtomicU8>>,
}
```

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
SanjoDeundiak: This quantity is used only here and only in the branch that guarantees we don't go into negatives

SanjoDeundiak: Yes, I think we can put 0 in the other branch here. However, I would like to avoid situation when the credential retriever returns an error really fast, and we end up calling it very many times in a loop. In this case is_retry is true and we add the delay before calling retriever again

implementations/rust/ockam/ockam_identity/src/secure_channels/secure_channels.rs
SanjoDeundiak: That's an interesting idea. I'll explore it when refactoring TrustContext

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
SanjoDeundiak: I actually don't like either of the approaches. We should parse the data only once and then carry that information with some struct

etorreborre: > We should parse the data only once and then carry that information with some struct

Yes I'm all for it 🎸

implementations/rust/ockam/ockam_identity/tests/credentials_refresh.rs
SanjoDeundiak: @etorreborre the duplication was a mistake and is now removed :D

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -91,33 +141,154 @@ impl EncryptorWorker {
...
+        } else {
+            // Refresh in self.refresh_credential_time_gap before the expiration
+            Duration::from_secs(
+                *(min_credential_expiration
+                    - self.refresh_credential_time_gap.as_secs().into()
+                    - now),
+            )
+        };
+
+        let duration = if duration < self.min_credential_refresh_interval && is_retry {
```
etorreborre: That's a nitpick but I think that would read even better as 
```rust
// Avoid too many request to the credential_retriever, the refresh can't be sooner than
// self.min_credential_refresh_interval if it's a retry
let duration = 
    if is_retry {
        min(self.min_credential_refresh_interval, duration)
    } else { 
        duration 
    };

File: implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs [hunk 18]
```diff
@@ -91,33 +141,154 @@ impl EncryptorWorker {
...
+                // Do nothing if there is no expiration
+                return Ok(());
+            };
+
+        // Cancel the old event
+        self.credential_refresh_event = None;
+
+        let now = now()?;
+
+        let duration = if min_credential_expiration < now + self.refresh_credential_time_gap {
```
etorreborre, line 263: Why not simply
```rust
let duration = max(min_credential_expiration - self.refresh_credential_time_gap - now, Duration::from_secs(0));
```

SanjoDeundiak: Isn't that gonna panic if  self.refresh_credential_time_gap + now > min_credential_expiration?

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
SanjoDeundiak: You tricked me, should be `max` instead of `min` 😃

etorreborre: yes, `max` :-)

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
etorreborre: In that case we take a duration of `0`. What would panic?

SanjoDeundiak: @etorreborre https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6237699b4dafeefc6fe2206869cfdf25

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
SanjoDeundiak: But I'm not doing it as a part of that PR :D

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
etorreborre: I see thanks. It's a bad inherited habit of working with signed integers.

etorreborre (dismissed):

polvorin (approved):

etorreborre (approved):

PR #6844, opened 2023-11-03
feat(rust): always using enum when representing the inlet connection status

davide-baldo:
Using the `ConnectionStatus` enum for the `InletStatus` for better readability and API clarity

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/session/sessions.rs
+++ b/implementations/rust/ockam/ockam_api/src/session/sessions.rs
@@ -5,7 +5,9 @@ use std::fmt::Formatter;
 use std::time::Duration;
 
 use minicbor::{Decode, Encode};
+use serde::{Deserialize, Serialize};
 
+use crate::error::ApiError;
 use ockam_core::compat::rand;
 use ockam_core::{Error, Route};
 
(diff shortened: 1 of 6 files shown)


implementations/rust/ockam/ockam_api/src/session/sessions.rs
```diff
@@ -21,24 +22,38 @@ pub type Replacer = Box<dyn FnMut(Route) -> Replacement + Send>;
...
+    }
+}
+
+impl From<String> for ConnectionStatus {
+    fn from(value: String) -> Self {
+        match value.as_str() {
+            "down" => ConnectionStatus::Down,
+            "degraded" => ConnectionStatus::Degraded,
+            "up" => ConnectionStatus::Up,
+            _ => ConnectionStatus::Down,
```
etorreborre: I'd rather have a `TryFrom` that is failing when the input string is incorrect. Otherwise we might have a silent failure in a bad situation where we just think that the node is currently down.

etorreborre (dismissed):
🥇 awesome!

There's just the question of parsing a `ConnectionStatus` from string.

implementations/rust/ockam/ockam_api/src/session/sessions.rs
davide-baldo: makes perfectly sense! done!

etorreborre (approved):

PR #6859, opened 2023-11-07
feat(rust): add lock file to cli state items

adrianbenavides:
This PR uses the `fs2` crate to use a lock file on every CliState item to mitigate the side effects of having concurrent writes through the `overwrite` or `persist` methods of the CliState traits.


davide-baldo (approved):
Robust and simple fix 👍

PR #6860, opened 2023-11-07
feat(rust): fixes many papercuts and better avoid and handle state resets

davide-baldo:
Fixes many paper cuts:

- invitation ignore taking a lot of time to refresh
- allow for application initialization to fail: in this case, the app shows a dialogue asking for the action to take (see screenshot)
- added `OCKAM_NO_AUTOMATIC_RESET` for both command and app, when set the application won't perfom any automatic reset
- fixed a bug when ignoring an invitation, the services view would get blank
- fixed a bug not showing correctly the "pending invitation" circle
- fixed a bug of not allowing CMD+V in fields (didn't allow pasting emails)

Screenshot 2023-11-07 at 14 07 27

### DIFF
--- a/implementations/swift/ockam/ockam_app/Ockam/MainView.swift
+++ b/implementations/swift/ockam/ockam_app/Ockam/MainView.swift
@@ -154,6 +154,14 @@ struct MainView: View {
         .onReceive(timer) { time in
             optionPressed = NSEvent.modifierFlags.contains(.option)
         }
+        .onReceive(state.$groups) { _ in
+            // the selected group could have been deleted, if so, reset the selection
+            if selectedGroup != "" {
+                if !state.groups.contains(where: { $0.id == selectedGroup }) {
+                    selectedGroup = ""
+                }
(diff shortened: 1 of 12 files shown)


implementations/swift/ockam/ockam_app/Ockam/MainView.swift
```diff
@@ -154,6 +154,14 @@ struct MainView: View {
...
             optionPressed = NSEvent.modifierFlags.contains(.option)
         }
+        .onReceive(state.$groups) { _ in
+            // the selected group could have been deleted, if so, reset the selection
+            if selectedGroup != "" {
+                if !state.groups.contains(where: { $0.id == selectedGroup }) {
+                    selectedGroup = ""
+                }
+            }
+        }
```
adrianbenavides, line 164: This change here is related to the `fixed a bug when ignoring an invitation, the services view would get blank` fix?

adrianbenavides (approved):

PR #6884, opened 2023-11-09
refactor(rust): add traits to hide the details of `show` and `delete` tui commands

adrianbenavides:
The idea is to simplify as much as possible to logic of commands so that we don't have to repeat all the branching required to implement the TUI experience for show and delete commands.


nazmulidris (approved):
LGTM 🎉

PR #6887, opened 2023-11-09
feat(rust): added persistent state for incoming services in the app

davide-baldo:
Accepted invitation aka Incoming services were a bit chaotic, mainly because the full "service" instance it's composed of three sources of truth: 

1. accepted invitation
2. inlet status
3. (added) persistent storage

This was leading to extra complexity and the inability to store extra state if the inlet wasn't active.
In this PR i introduce persistence (enable/disabled) while refactoring the inlet handling: rather than handling them "raw" from accepting invitation, i created a new `IncomingService` concept at the `AppState`, which keeps every incoming service in-memory, these services are loaded when the `accepted_invitations` comes in.
The idea is to allow the user to modify accepted services by changing name, schema, and maybe also offer the possibility to have a fixed port.
I also refactored `invitations` module, splitting it by invitations-only operation and incoming service operations.
Minor refactor and fixes was also performed in the swift side, by moving the `IncomingServiceView` in separated file and fixed some enable/disable view issues.


implementations/rust/ockam/ockam_app_lib/src/state/mod.rs
```diff
@@ -79,11 +81,18 @@ pub struct AppState {
 
 impl AppState {
     /// Create a new AppState, if it fails you can assume it's because the state cannot be loaded
+    /// when `cli_state` is `None` the default initialization will be used
```
etorreborre, line 94: I understand that this is done to pass [a specific `CliState` when testing](https://github.com/build-trust/ockam/pull/6887/files#diff-fa4b824b627a0cee53c515fa420446a03f60247cf83549352a805aae013969dcR261). I'd rather have separate constructors for each use case instead (because there are less ways to misuse the API):
```rust
impl AppState {
    // called from the swift app 
    pub fn new(
        application_state_callback: ApplicationStateCallback,
        notification_callback: NotificationCallback) -> Result<AppState> {
       let cli_state = CliState::initialize()?;
       Self::make(application_state_callback, notification_callback, cli_state)
    }

    // called from tests
    pub fn test(cli_state: CliState) -> Result<AppState> {
       Self::make(None, None, cli_state)
    }

    // internal constructor
    fn make_app_state(
        application_state_callback: Option<ApplicationStateCallback>,
        notification_callback: Option<NotificationCallback>,
        cli_state: CliState) -> Result<AppState> {
       
    }
}

File: implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs [hunk 2]
```diff
@@ -0,0 +1,365 @@
...
+        let bare_relay_name = self.access_details.shared_node_identity.to_string();
+        format!("forward_to_{bare_relay_name}")
+    }
+
+    /// Returns the name of the inlet within the node, for now it's a constant
+    pub fn inlet_name(&self) -> &str {
+        "app-inlet"
+    }
+
+    pub fn service_route(&self) -> ockam::Result<String> {
```
etorreborre: I think it would be better to parse all the information we need when we get the `ServiceInvitationDetails` and create an `IncomingService`. This way we can return / log errors sooner rather than having incorrect or incomplete data and only discovering it on usage.

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
```diff
@@ -0,0 +1,365 @@
...
+                    name: None,
+                });
+                self.incoming_services.last_mut().unwrap()
+            }
+        }
+    }
+}
+
+impl AppState {
+    pub async fn load_services_from_invites(
```
etorreborre: s/invites/invitations

implementations/rust/ockam/ockam_app_lib/src/incoming_services/commands.rs
```diff
@@ -0,0 +1,212 @@
...
+impl AppState {
+    pub(crate) async fn refresh_inlets(&self) -> crate::Result<()> {
+        info!("Refreshing inlets");
+
+        // for each invitation it checks if the relative node is running
+        // if not, it deletes the node and re-create the inlet
+
+        let services_arc = self.incoming_services();
+        let services = {
+            // reduce locking as much as possible to make UI consistently responsive
```
etorreborre, line 25: Thanks for the comment!

implementations/rust/ockam/ockam_app_lib/src/incoming_services/commands.rs
```diff
@@ -0,0 +1,212 @@
...
+        if self.state().await.nodes.exists(inlet_node_name) {
+            let mut inlet_node = self.background_node(inlet_node_name).await?;
+            inlet_node.set_timeout(Duration::from_secs(5));
+
+            if let Ok(inlet) = inlet_node
+                .show_inlet(&self.context(), service.inlet_name())
+                .await?
+                .success()
+            {
+                if inlet.status == ConnectionStatus::Up {
```
etorreborre: Isn't this a no-op? If the inlet is up we don't even have to return a port. I'm wondering if it wouldn't be more direct to to implement an `is_running` operation on `BackgroundNodeClient` so that:
```rust
if inlet_node.is_running(&self.context(), service.inlet_name()).await? {
    Ok(None)
} else {
    self.create_inlet(background_node_client, service).await.map(Some)
}
```

implementations/rust/ockam/ockam_app_lib/src/incoming_services/commands.rs
```diff
@@ -0,0 +1,212 @@
...
+            .map(Some)
+    }
+
+    /// Create the tcp-inlet for the accepted invitation
+    /// Returns the inlet SocketAddr
+    async fn create_inlet(
+        &self,
+        background_node_client: Arc<dyn BackgroundNodeClient>,
+        service: &IncomingService,
+    ) -> crate::Result<u16> {
```
etorreborre: I'd be happier with a `Port` datatype :-). This would also make the method more self-documenting (BTW the comment on L.98 contradicts the return type)

implementations/rust/ockam/ockam_app_lib/src/incoming_services/commands.rs
```diff
@@ -0,0 +1,212 @@
...
+        debug!(
+            service_name = service.name(),
+            "Creating TCP inlet for accepted invitation"
+        );
+
+        let local_node_name = service.local_node_name()?;
```
etorreborre: This line could go closer to L116

implementations/rust/ockam/ockam_app_lib/src/incoming_services/commands.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        let bind_address = match service.address() {
```
etorreborre: This line could go closer to L.135

implementations/rust/ockam/ockam_app_lib/src/incoming_services/commands.rs
```diff
@@ -0,0 +1,212 @@
...
+            .await?;
+        Ok(bind_address.port())
+    }
+
+    pub(crate) async fn disable_tcp_inlet(&self, invitation_id: &str) -> crate::Result<()> {
+        // first change the in-memory state
+        let service = {
+            let incoming_services_arc = self.incoming_services();
+            let mut writer = incoming_services_arc.write().await;
+            let mut service = writer.find_mut_by_id(invitation_id);
```
etorreborre, line 161: Maybe we could get rid of the case where the service is not found right at the beginning, rather that here and on line 164

implementations/rust/ockam/ockam_app_lib/src/incoming_services/commands.rs
```diff
@@ -0,0 +1,212 @@
...
+                service.disable();
+            }
+            service.cloned()
+        };
+
+        if let Some(service) = service {
+            // this is an async operation, let's give feedback to the user as soon as possible
+            self.publish_state().await;
+            // change the persistent state
+            self.model_mut(|model| {
```
etorreborre: I'm not a great fan of this interface to the model state. Why not have a function like:
```rust
impl ModelState {
    pub async fn disable_service(&mut self, invitation_id: &str) -> Result<()> {
       // and use model_mut internally
    }
}
```
I think this would encapsulate a bit better the modifications that we expect to do on a `ModelState`.

implementations/rust/ockam/ockam_app_lib/src/state/mod.rs
```diff
@@ -581,33 +596,17 @@ impl AppState {
...
-                                    source_name: access_details
-                                        .service_name()
-                                        .unwrap_or("unknown".to_string()),
-                                    address: inlet.map(|inlet| inlet.socket_addr.ip().to_string()),
-                                    port: inlet.map(|inlet| inlet.socket_addr.port()),
-                                    scheme: None,
-                                    available: inlet.is_some(),
-                                    enabled: inlet.map(|inlet| inlet.enabled).unwrap_or(true),
-                                }
+                            .map(|service| Service {
```
etorreborre, line 627: Naming the variable `service` to `incoming_service` would be slightly clearer. Otherwise it seems that we are transforming a `Service` into a `Service`. A better alternative IMO would be to have a `Service::from_incoming_service` function to pass to `map`.

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
```diff
@@ -0,0 +1,365 @@
...
+                target_id: "target_id".to_string(),
+                ignored: false,
+            },
+            service_access_details: None,
+        };
+
+        // Skipping an invitation without service access details
+        let app_state = AppState::new(None, None, Some(CliState::test().unwrap())).unwrap();
+
+        app_state.context().runtime().block_on(async move {
```
etorreborre: Can we run this test as a `tokio::test` instead of blocking here?

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
```diff
@@ -0,0 +1,365 @@
...
+    use ockam_api::cloud::share::{
+        InvitationWithAccess, ReceivedInvitation, RoleInShare, ServiceAccessDetails, ShareScope,
+    };
+    use ockam_api::config::lookup::ProjectLookup;
+    use ockam_api::identity::EnrollmentTicket;
+    use std::str::FromStr;
+
+    #[test]
+    fn test_inlet_data_from_invitation() {
+        let mut invitation = InvitationWithAccess {
```
etorreborre: This could be moved closer to L.265. I would also advise to factor out the code creating data outside of the test as utility functions so that:

 - the structure of the test become clearer
 - we can reuse those functions for other tests

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
```diff
@@ -0,0 +1,365 @@
...
+            assert_eq!(1, services.len(), "One service should be loaded");
+
+            let service = &services[0];
+            assert_eq!(
+                "invitation_id",
+                service.id()
+            );
+            assert_eq!(
+                "I12ab34cd56ef12ab34cd56ef12ab34cd56ef12ab",
+                service.access_details.shared_node_identity.to_string()
```
etorreborre: I think that this test breaks encapsulation. The behaviour we want to test is fully covered by L.322 and L324

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
```diff
@@ -0,0 +1,365 @@
...
+                    grant_role: RoleInShare::Admin,
+                    owner_email: "owner_email".to_string(),
+                    scope: ShareScope::Project,
+                    target_id: "target_id".to_string(),
+                    ignored: false,
+                },
+                service_access_details: invitation.service_access_details.clone(),
+            };
+
+            // let's load another invitation, but but a persistent state for it already exists
```
etorreborre: s/but but/but

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
```diff
@@ -0,0 +1,365 @@
...
+                    "No service access details found for accepted invitations {}",
+                    invite.invitation.id
+                );
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct IncomingService {
```
etorreborre, line 147: A bit of description would be nice. In particular to explain that:

 - `id` and `access_details` come from the invitation
 - `name` and `enabled` can be changed by the user of the service
 - `port` is set when the service is running

Those 3 groups have very different lifecycles.

etorreborre: Also the information that what makes each service unique is its service id, and that comes from the invitation id.

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
```diff
@@ -0,0 +1,365 @@
...
+            // let's load another invitation, but but a persistent state for it already exists
+            app_state.model_mut(|m| {
+                m.incoming_services.push(PersistentIncomingServiceState {
+                    invitation_id: "second_invitation_id".to_string(),
+                    enabled: false,
+                    name: Some("custom_user_name".to_string()),
+                })
+            }).await.unwrap();
+
+            app_state
```
etorreborre: One slightly confusing thing with this test is that both services share the same access details. So I wondered if that was an objective of the test, testing what happens when there are 2 invitations to the same service. But actually I think that this is just about testing what happens when some state (service name / service enabled) has already been persisted.

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
```diff
@@ -0,0 +1,365 @@
...
+use ockam_api::error::ApiError;
+use ockam_api::identity::EnrollmentTicket;
+use serde::{Deserialize, Serialize};
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use tracing::warn;
+
+#[derive(Clone, Debug, Decode, Encode, Serialize, Deserialize, PartialEq)]
+#[rustfmt::skip]
+#[cbor(map)]
+pub struct PersistentIncomingServiceState {
```
etorreborre, line 16: In terms of overall structure I wonder if we wouldn't be better off if we:
 
 - grab all the invitations from the controller
 - transform them to `IncomingService` with additional `custom_name / enabled` fields for every new invitation received
 - persist the `IncomingServices` directly
 - recreate/restart the inlets directly from the persisted data

This would get rid of one data structure but also give more reliability to the app which would also be able to work even if the controller has hiccups when retrieving invitations.

etorreborre (dismissed):
Hi @davide-baldo I agree with the overall intent but I left plenty (too much :-)?) of ideas for refactoring / adjusting the code.

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
davide-baldo: ~~It may be superfluous, but I don't understand why would it break the encapsulation?~~ got it 👍

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
davide-baldo: Access details are just carried over and wasn't the objective the test, i will try to better explain the test goal by renaming and adding some comments

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
davide-baldo: I thought the same, but with the objective to lower the application bootstrap latency and avoid the "empty services" in the first second.

My arguments against it are:
The application only makes sense when it's connected to the orchestrator, if the controller is down or having issues, we want direct feedback, and by serializing more information, we increase the possibility to break back-compatibility. I don't think it's worth doing it during this initial phase.

implementations/rust/ockam/ockam_app_lib/src/incoming_services/commands.rs
davide-baldo: It's to reduce the lock scope, I've added another
```// we want to reduce the scope of the guard as much as possible``` 
:)

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
etorreborre: > The application only makes sense when it's connected to the orchestrator,

Let's bring that point up in our next meeting

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
etorreborre: That's the because the internal structure of service is accessed directly: `access_details.shared_node_identity`.
If we want to evolve the structure of the service, that's one more place where we need to change code.

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
davide-baldo: no, creating the context of the ockam app would create another tokio runtime resulting in a panic

implementations/rust/ockam/ockam_app_lib/src/incoming_services/commands.rs
davide-baldo: Makes sense, but I would like to take a bit more time for this, I've created an issue for it!

implementations/rust/ockam/ockam_app_lib/src/state/mod.rs
davide-baldo: > Otherwise it seems that we are transforming a Service into a Service

Actually we are.. from one internal representation with the full state to a simplified, C-compatible one.
Maybe adding the whole module would remove some of the confusion, what do you think?

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
etorreborre: I've created a branch with a commit showing how we can use the `ockam::test` macro to have a proper `Context` being provided to run the test if you want to have a look at it: `davide-baldo/app-persistent-service-state-with-ockam-test`

davide-baldo: I can't find the branch

davide-baldo:
> Hi @davide-baldo I agree with the overall intent but I left plenty (too much :-)?) of ideas for refactoring / adjusting the code.

There are two things that give me confidence about my code: tests and your review! I really appreciate the time you spend and the attention to details, it's never too much 😃 

PS: i've pushed a temporary commit to the changes, i'll squash it before merge

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
etorreborre: Sorry @davide-baldo I must have forgotten to push. Please try again.

etorreborre (dismissed):

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
etorreborre: We can change that aspect in the future if we want.

implementations/rust/ockam/ockam_app_lib/src/state/mod.rs
etorreborre: Yes, that would help understand that this is indeed transformed data. We can merge and do that if we come back for a second pass.

etorreborre (approved):

PR #6910, opened 2023-11-16
fix(rust): fixed enrollment when not successfully completed on first try

davide-baldo:
This PR fixes two cases:

1. When email is not verified, the enrollment would show up a notification but would continue, failing
2. When en enrollment is interrupted and another one is performed with a different account, it would fail because the new `user_info` wouldn't be the default

### DIFF
--- a/implementations/rust/ockam/ockam_app_lib/src/enroll/enroll_user.rs
+++ b/implementations/rust/ockam/ockam_app_lib/src/enroll/enroll_user.rs
@@ -14,6 +14,12 @@ use ockam_api::enroll::oidc_service::OidcService;
 use crate::state::{AppState, NODE_NAME, PROJECT_NAME};
 use crate::Result;
 
+enum EnrollmentOutcome {
+    AlreadyEnrolled,
+    PendingValidation,
+    Successful,
+}
+
 impl AppState {
     /// Enroll a user.
     ///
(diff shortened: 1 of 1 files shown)


implementations/rust/ockam/ockam_app_lib/src/enroll/enroll_user.rs
```diff
@@ -61,10 +70,10 @@ impl AppState {
         Ok(())
     }
 
-    async fn enroll_with_token(&self) -> Result<()> {
+    async fn enroll_with_token(&self) -> Result<bool> {
         if self.is_enrolled().await.unwrap_or_default() {
             debug!("User is already enrolled");
-            return Ok(());
+            return Ok(false);
```
etorreborre: I'd be tempted to use an enum there for a slightly better understanding:
```
enum EnrollmentStatus {
    AlreadyEnrolled,
    NewlyEnrolled,
}

etorreborre (dismissed):

implementations/rust/ockam/ockam_app_lib/src/enroll/enroll_user.rs
davide-baldo: I was tempted as well, but it would have been a superset of a Result, like

```rust
enum EnrollmentResult<T> {
  AlreadyEnrolled,
  PendingValidation,
  Successful,
  Err(T)
}
  ```

etorreborre (approved):

PR #6913, opened 2023-11-17
feat(rust): persist application data in a database

etorreborre:
This PR replaces the current storage of persistent data in separated files with the storage of data is a SQLite database.

# Motivation

There is a large amount of data need to run an Ockam node or application:

 - project, user
 - identity
 - node name
 - trust context
 - TCP listener
 - TCP outlet
 - etc...

All this data is inter-related:

 - a node has a default identity
 - a trust context can be created via a project
 - a node has a TCP listener
 - a TCP outlet can be created on a given node
 - etc...

At the moment, everything is stored in files (and some information just lives in memory, like a started TCP inlet).
Managing those files separately makes it hard to process the relationships between the data and creates some operational issues (there's no way to have atomic transactions to modify data for example), so a database fits better our data model.

# Changes

This is a large PR because the access to data, via the `CliState` struct was not very well encapsulated. As a result it was not possible to just replace one persistence solution with another to leave most of the code unchanged.

The following sections describe the major changes, from low-level crates to high-level ones.

## `ockam_node`

In this crate we introduce a `storage/database` folder with an `SqlxDatabase`. [`sqlx`](https://github.com/launchbadge/sqlx) is the Rust library used to handle the database connections and queries. It has been chosen because it supports several "drivers", SQLite, but also Postgres, if we need to migrate to Postgres in the future.

When an `SqlxDatabase` is started on a given path, a migration file is applied to create or migrate all the SQL tables we need to persist data. The migration files are available in `src/storage/database/migration`.

Additionally there are traits and conversion functions to allow the data types used in Ockam: `Identifier`, `ChangeHistory`, `PurposeKey` etc... to be mapped to simpler SQLite types when inserting data.

Note that there was no real attempt to make the current schema optimum. But with migrations we should be able to evolve it gracefully.

## `ockam_vault` 

In that crate we replaced the persistence of secrets in JSON file with key/values to a table in the database.
Note that the access to secrets is done via an interface, `SecretsRepository`, and that the implementation currently uses one single database for all data but we could as well persist this information in a separated database file since we never do queries across the table storing secrets and other tables.

## `ockam_identity`

In that crate we have now 3 repositories (each of them backed by a `SqlxDatabase`):

 - a `ChangeHistoryRepository` to store the `ChangeHistories`. Note that we don't store `VerifiedChanges` in order to avoid someone changing the history of an identity by just accessing the database file. When a `ChangeHistory` is loaded from the database it has to be verified in order to get a full `Identity`
 - an `IdentityAttributesRepository` to store the attributes associated to an identifier (I reunited the read and the write interface because I don't think that the split is that useful)
 - a `PurposeKeysRepository` to store keys and their associated purpose (secure channel or credentials)

Then most of the changes come from:

 - the split of the previous `IdentitiesRepository` into `ChangeHistoryRepository` + `IdentityAttributesRepository`
 - some utility functions added for trust contexts, for recreating an identity from a change history, or getting credential data from a `Credential`
 - the removal of the previous storage structs + the initial attempt at using SQLite to store policies and identities

## `ockam_abac`

A new `PolicyRepository` interface has been added to that crate, with the corresponding `PolicySqlxDatabase` implementation.
The previous implementations have been removed.

## `ockam_api`

This is the crate with the largest amount of changes. The `CliState` struct holds all the data that we are interested in via a set of repositories:

 - `IdentitiesRepository` to store named identities, i.e. the fact that we can associate a name to an identity and set a default identity
 - `CredentialsRepository` to store named credentials
 - `VaultsRepository` to store named vaults
 - `NodesRepository` to store data about background nodes (their name, their pid, etc...)
 - `TrustContextsRepository` to store data about trust contexts
 - `ProjectsRepository` to store project data
 - `SpacesRepository` to store space data (note that the association between `Space` and `Project` is not exposed at the repository level but is implicit with the `space_name` field in `Project`)
 - `EnrollmentsRepository` to keep track of enrolling information for users
 - `UsersRepository` to store user info as retrieved after authentication
 - `ModelStateRepository` to store data for the desktop application: created inlets and outlets

The functions exposed by the repositories cannot be accessed directly. They are hidden behind proper methods on the `CliState`. For example creating a `Node` involves several repositories and this logic is hidden behind the `CliState::create_node` function (or one of its variants).

The main `CliState` functions are provided in files, domain by domain:

 - `nodes`
 - `credentials`
 - `projects`
 - `spaces`
 - `identities`
 - `vaults`
 - `nodes`
 - `trust_contexts`
 - `users`
 - `enrollments`
 - `policies`
 - `secure_channels`

## `ockam_command`

Most of the functions changed in that crate now call a "facade" function on `CliState` to access or modify the local state.
In other cases like getting the list of all projects an `InMemoryNode` is started to first synchronize the local state with the `Controller` state. This kind of code could be moved to the `CliState` as well since it has all the necessary information without having to start an `InMemoryNode`.

## `ockam_app_lib`

The changes here are very similar to the changes to the `ockam_command` crate. Most of the places where `CliState` was accessed to get or modify data, I had to modify calls.
One notable addition is the `ModelStateRepository` to store "outlet statuses" and "incoming services" in the database.


implementations/rust/ockam/ockam/Cargo.toml
```diff
@@ -87,7 +87,7 @@ hex = { version = "0.4", default-features = false }
 minicbor = { version = "0.20.0", features = ["alloc", "derive"] }
 ockam_abac = { path = "../ockam_abac", version = "^0.38.0", default_features = false, optional = true }
 ockam_core = { path = "../ockam_core", version = "^0.93.0", default-features = false }
-ockam_identity = { path = "../ockam_identity", version = "^0.92.0", default_features = false }
+ockam_identity = { path = "../ockam_identity", optional = true, version = "^0.92.0", default_features = false }
```
SanjoDeundiak: I thought ockam is meant to be an umbrella library with all the public API, so how is identity optional then?

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -88,17 +88,19 @@ fn identities_ref() -> NifResult<Arc<Identities>> {
...
-    *IDENTITY_MEMORY_VAULT.write().unwrap() = Some(identity_vault.clone());
-    *SECURE_CHANNEL_MEMORY_VAULT.write().unwrap() = Some(secure_channel_vault.clone());
-    let builder = ockam_identity::Identities::builder().with_vault(Vault::new(
-        identity_vault,
-        secure_channel_vault,
-        Vault::create_credential_vault(),
-        Vault::create_verifying_vault(),
-    ));
-    *IDENTITIES.write().unwrap() = Some(builder.build());
+    block_future(async move {
```
SanjoDeundiak, line 91: Why is `block_future` required here?

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
```diff
@@ -0,0 +1,187 @@
+CREATE TABLE identity
```
SanjoDeundiak, line 8: I wonder if it's even possible to let every crate manage its own sql scripts

SanjoDeundiak: The fact that I can see now everything that is persisted is fascinating though, it was so hard before

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
```diff
@@ -0,0 +1,187 @@
...
+
+CREATE TABLE vault
+(
+    name       TEXT PRIMARY KEY,
+    path       TEXT NOT NULL,
+    is_default INTEGER,
+    is_kms     INTEGER
+);
+
+CREATE TABLE signing_secret
```
SanjoDeundiak: This is a little bit overkill. All we need for a secret, is its binary and type

SanjoDeundiak: handle is computed for signing keys, no need to store it

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
```diff
@@ -0,0 +1,187 @@
+CREATE TABLE identity
+(
+    identifier     TEXT NOT NULL UNIQUE,
```
SanjoDeundiak, line 10: Should we use binary for the Identifier?

SanjoDeundiak: And change_history?

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
```diff
@@ -0,0 +1,187 @@
...
+    change_history TEXT NOT NULL
+);
+
+CREATE TABLE identity_enrollment
```
SanjoDeundiak, line 25: Would be great to have a small comment for each TABLE (and optionally fields, where it's not obvious what they mean)

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
```diff
@@ -0,0 +1,187 @@
...
+(
+    identifier  TEXT    NOT NULL UNIQUE,
+    enrolled_at INTEGER NOT NULL
+);
+
+CREATE TABLE identity_attributes
+(
+    identifier  TEXT PRIMARY KEY,
```
SanjoDeundiak: Should it be or should it not be a foreign key? (I really don't know, requires consideration)

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
```diff
@@ -0,0 +1,187 @@
...
+(
+    name                     TEXT PRIMARY KEY,
+    trust_context_id         TEXT    NOT NULL,
+    is_default               INTEGER NOT NULL,
+    credential               TEXT,
+    authority_change_history TEXT,
+    authority_route          TEXT
+);
+
+CREATE TABLE tcp_outlet_status
```
SanjoDeundiak: Does it need to be persisted?

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
```diff
@@ -0,0 +1,187 @@
...
+    updated_at     TEXT    NOT NULL,
+    email_verified INTEGER NOT NULL,
+    is_default     INTEGER NOT NULL
+);
+
+CREATE TABLE credential
+(
+    name                  TEXT PRIMARY KEY,
+    issuer_identifier     TEXT NOT NULL,
+    issuer_change_history TEXT NOT NULL,
```
SanjoDeundiak: That should probably be in the `identity` table

implementations/rust/ockam/ockam/Cargo.toml
etorreborre: That's a mistake due to me trying to fix the `no_std` job. I've reverted it.

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
etorreborre: Otherwise the database is complaining that an async context is missing to create a connection pool.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
etorreborre: > I wonder if it's even possible to let every crate manage its own sql scripts
Maybe, I thought about it (at least in a 2 layer thing: below or above `ockam`) but opted for the simplest thing for now.

> The fact that I can see now everything that is persisted is fascinating though, it was so hard before

It is great for debugging! There is also a risk of introducing coupling at the database layer so we need to be careful with that.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
etorreborre: How can we retrieve a signing secret if we don't store its handle?

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
etorreborre: I think that we should start favoring debuggability unless there are established performance concerns. Having the possibility to read text values compared to having blobs is really useful when you're diagnosing an incident or a failed test.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
etorreborre: On it :-).

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
etorreborre: For the desktop application, yes

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
etorreborre: Maybe, I will leave that for data model improvements.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
etorreborre: Yes, I did more or less direct mapping but we could enforce a bit of normalization.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
SanjoDeundiak: That's true, but you can always print it as a string. SQL GUIs usually show it as a hex anyways

SanjoDeundiak: I suspect you can convert it to hex inside the SQL SELECT statement if you wish

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
SanjoDeundiak: Good point, we can compute it, but we don't want to do it on every select for the whole table :)

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
SanjoDeundiak: That's weird. There is no `.await` inside this block, but from what you say there is something async called somewhere under the hood. That's not good, would you like to figure it out?

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
etorreborre: > That's true, but you can always print it as a string. SQL GUIs usually show it as a hex anyways
> I suspect you can convert it to hex inside the SQL SELECT statement if you wish

When you are debugging complicated things, you generally don't want to write additional code or commands to compare for example the content of the database to a line in a log file. Unless there is a strong need to do otherwise I think we get lots of value to be able to see all the data directly as much as we can.

implementations/rust/ockam/ockam_abac/src/storage/mod.rs
```diff
@@ -1,26 +1,7 @@
+mod policy_repository;
 #[cfg(feature = "std")]
-pub mod lmdb_storage;
```
SanjoDeundiak, line 2: 🥳

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
```diff
@@ -0,0 +1,223 @@
...
+    type Target = SqlitePool;
+
+    fn deref(&self) -> &Self::Target {
+        &self.pool
+    }
+}
+
+impl SqlxDatabase {
+    /// Constructor for a database persisted on disk
+    pub async fn create<P: AsRef<Path>>(path: P) -> Result<Self> {
```
SanjoDeundiak: I prefer
```
pub async fn create(path: impl AsRef<Path>) -> Result<Self> {
```
Just in case you like it too

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        // creating a new database might be failing a few times
+        // if the files are currently being held by another pod which is shutting down.
+        // In that case we retry a few times, between 1 and 10 seconds.
+        let retry_strategy = FixedInterval::from_millis(1000)
+            .map(jitter) // add jitter to delays
+            .take(10); // limit to 10 retries
+
+        let db = Retry::spawn(retry_strategy, || async {
+            Self::create_and_migrate(path.as_ref()).await
```
SanjoDeundiak: Should we retry connection only? Not including migration?

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
etorreborre: This is related to the fact that some storage is in-memory storage which eventually uses `SqlxDatabase::create` which blocks on the async `sqlx` functions for creating an in-memory connection pool and migrate data.
Do you think there's a better way to handle this?

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
```diff
@@ -0,0 +1,223 @@
...
+        })
+        .await?;
+        Ok(db)
+    }
+
+    /// Constructor for an in-memory database
+    /// The implementation blocks during the creation of the database
```
SanjoDeundiak, line 62: But does it make much sense if using the storage still requires async?

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
```diff
@@ -0,0 +1,223 @@
...
+            let pool = Self::create_in_memory_connection_pool()
+                .await
+                .expect("creating an in-memory connection should work");
+            let db = SqlxDatabase { pool };
+            db.migrate().await.expect("the migration should work");
+            db
+        })
+    }
+
+    async fn create_and_migrate(path: &Path) -> Result<Self> {
```
SanjoDeundiak: I would split that into 2 functions. It looks like there are multiple ways of creating the db, but only one function to migrate

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
SanjoDeundiak: Also, the caller always can block, if they want to

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    use crate::database::ToSqlxType;
+
+    use super::*;
+
+    /// This is a sanity check to test that the database can be created with a file path
+    /// and that migrations are running ok, at least for one table
+    #[tokio::test]
+    async fn test_create_identity_table() -> Result<()> {
+        // let db = create_database().await?;
```
SanjoDeundiak: Should be removed probably

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_types.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// This type represents arbitrary bytes in the database
+    Blob(Vec<u8>),
+    /// This type represents ints, signed or unsigned
+    Integer(i64),
+    /// This type represents floats
+    #[allow(unused)]
+    Real(f64),
+}
+
+/// The SqlType implements the Type<Sqlite> trait from sqlx to allow its values to be serialized
```
SanjoDeundiak: Could you please elaborate on how this is used? I have problems understanding why we return `Vec<u8>::type_info()` here. (I understand that it's a static method and we can't return a different type info depending on the self value)

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_types.rs
```diff
@@ -0,0 +1,200 @@
...
+///     }
+/// }
+///
+/// let timestamp = TimestampInSeconds(10000000);
+/// let query = query_as("SELECT * FROM identity WHERE created_at >= $1").bind(timestamp.as_sql());
+///
+///
+pub trait ToSqlxType {
+    /// Return the appropriate sql type
+    fn to_sql(&self) -> SqlxType;
```
SanjoDeundiak, line 88: I would prefer `fn to_sqlx_type(&self) -> SqlxType;`. WDYT?

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_types.rs
```diff
@@ -0,0 +1,200 @@
...
+            1.to_sql()
+        } else {
+            0.to_sql()
+        }
+    }
+}
+
+impl ToSqlxType for u64 {
+    fn to_sql(&self) -> SqlxType {
+        SqlxType::Integer(*self as i64)
```
SanjoDeundiak, line 115: That's potentially unsafe

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_types.rs
```diff
@@ -0,0 +1,200 @@
...
+
+impl ToSqlxType for u8 {
+    fn to_sql(&self) -> SqlxType {
+        SqlxType::Integer(*self as i64)
+    }
+}
+
+impl ToSqlxType for usize {
+    fn to_sql(&self) -> SqlxType {
+        SqlxType::Integer(*self as i64)
```
SanjoDeundiak: That's also potentially unsafe

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
SanjoDeundiak: Could you please elaborate? Maybe with a small example? Because in my opinion, we should pick the most suitable way of storing data. In this case we have binary data and SQL supports binary fields. I don't understand how picking the more compact format for storing makes querying/displaying/logging data and debugging application harder.

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
etorreborre: Yes, I changed it

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
SanjoDeundiak: Ok, so this is actually related to my other comment at the creation of in memory db :)
I think generally you need to propagate async code to the top and let the caller decide if they need to block (usually they don't)

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
etorreborre: Yes, good idea, and it goes with your comment about splitting `create_and_migrate`.

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
etorreborre: 👍

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
etorreborre: 👍

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
SanjoDeundiak: Otherwise, you end up with exactly the problem you encountered here: you call an ordinary sync function, but it errors (or even panics?) because there is an async code somewhere inside. That's completely unexpected for the caller, also there are potentially performance concerns in this case

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
etorreborre: Setting this as `async` has a large knock-on effect to the builders for identities and secure channels. I think that they need to be revisited at some stage anyway so I propose that we do this at that time.

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_types.rs
etorreborre: Yes I understand the concern. I decided to keep it short for building queries, which is the thing we do the most. When we build queries we don't even see the `SqlxType` mentioned so there's no discrepancy there.

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_types.rs
etorreborre: My understanding is that this is just a default value in this case, which is overridden by the `Encode` instance on line 42. This is what the doc says about this function:
```
When binding arguments, this is used to tell the database what is about to be sent; which, the database then uses to guide query plans. This can be overridden by Encode::produces
```

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_types.rs
etorreborre: What do you propose to do instead?

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
etorreborre: For example a binary blob in TablePlus (SQL client for MacOS) shows up like this
image
If identifiers are stored as blob it will be hard to see what they are, especially if we want to correlate with logs.

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_types.rs
SanjoDeundiak: Let's introduce an unsigned integer type to `ToSqlxType` and test that it indeed is correct for values more than 2^32 after a store/retrieve cycle.
Not sure what to do with a usize, do we consider platforms with more than 64 arch?

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
SanjoDeundiak: That should not be a problem, everything ockam-related is already called from an async context

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
SanjoDeundiak: You made me download the app 
Screenshot 2023-11-21 at 21 04 58

SanjoDeundiak: Printing it as a hex to the log is easy, selecting it from the database in hex format is also easy

etorreborre: It's possible, it's not easy. When I'm debugging an issue I can be browsing a lot of information at the same time to get visual cues. Having to click 2 times to read a value is not practical in that case.

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
etorreborre: I agree on propagating the async context in general. In that case I did not do it to limit the amount of changes in that PR. I can give a go at changing that if you feel strongly about it right now.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
etorreborre: Done

implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
etorreborre: Done

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
etorreborre: Done

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_types.rs
etorreborre: I don't see what you mean. Can you please add a commit showing your idea?

etorreborre: Maybe we can use `i64::try_from(self)` and raise a proper panic message if we ever overflow?

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
SanjoDeundiak: I agree, but that complain should likely be directed to the creator of TablePlus. Both GUIs I use (Intellij IDEs' embedded sql tool and DBeaver) have no problem displaying blobs as hex without any extra clicks.

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_types.rs
SanjoDeundiak: I'll need to play with it for a bit today. If that's the only thing left in this PR - let's merge it and create an issue for that

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
etorreborre: That's good I didn't know that some GUIs could be configured for this. However:

 - this is not the case for all of them, for example DB Browser doesn't do it as far as I can see
 - if we need to ssh somewhere and the only thing we have to diagnose an issue is the command-line `sqlite`, we're stuck

implementations/rust/ockam/ockam_abac/src/storage/policy_repository_sql.rs
```diff
@@ -0,0 +1,151 @@
...
+        Ok(Arc::new(Self::new(
+            SqlxDatabase::in_memory("policies").await?,
+        )))
+    }
+}
+
+#[async_trait]
+impl PoliciesRepository for PolicySqlxDatabase {
+    async fn get_policy(&self, resource: &Resource, action: &Action) -> Result<Option<Expr>> {
+        let query = query_as("SELECT * FROM policy WHERE resource=$1 and action=$2")
```
SanjoDeundiak, line 35: Should we either use `LIMIT` or create a unique index if we expect only one row? Maybe even both

implementations/rust/ockam/ockam_vault/src/storage/secrets_repository_sql.rs
```diff
@@ -0,0 +1,319 @@
...
+
+#[async_trait]
+impl SecretsRepository for SecretsSqlxDatabase {
+    async fn store_signing_secret(
+        &self,
+        handle: &SigningSecretKeyHandle,
+        secret: SigningSecret,
+    ) -> Result<()> {
+        let secret_type: String = match handle {
+            SigningSecretKeyHandle::EdDSACurve25519(_) => "EdDSACurve25519".into(),
```
SanjoDeundiak: Would be cool to have it as a const

implementations/rust/ockam/ockam_vault/src/storage/secrets_repository_sql.rs
```diff
@@ -0,0 +1,319 @@
...
+        match self {
+            SigningSecret::EdDSACurve25519(k) => k.key().to_sql(),
+            SigningSecret::ECDSASHA256CurveP256(k) => k.key().to_sql(),
+        }
+    }
+}
+
+impl ToSqlxType for SigningSecretKeyHandle {
+    fn to_sql(&self) -> SqlxType {
+        match self {
```
SanjoDeundiak: self.handle().to_sql() if you wish

implementations/rust/ockam/ockam_abac/src/storage/policy_repository_sql.rs
etorreborre: Good idea, I've added a unique index.

implementations/rust/ockam/ockam_vault/src/storage/secrets_repository_sql.rs
etorreborre: Indeed, I've added consts.

implementations/rust/ockam/ockam_vault/src/storage/secrets_repository_sql.rs
etorreborre: Thanks.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -441,43 +421,49 @@ pub async fn spawn_background_node(
...
+        cmd.trust_context_opts.project_name.clone(),
         cmd.logging_to_file(),
-    )?;
+    )
+    .await?;
 
     Ok(())
 }
+
+async fn check_if_node_is_not_already_running(opts: &CommandGlobalOpts, cmd: &CreateCommand) {
```
adrianbenavides: _Nitpick_

A function prefixed with `check_if` makes me think that it's going to return a bool. I would use the `guard` concept here and rename it to `guard_node_is_not_already_running`

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -187,12 +175,14 @@ pub(crate) async fn background_mode(
     ctx: Context,
     (opts, cmd): (CommandGlobalOpts, CreateCommand),
 ) -> miette::Result<()> {
-    let node_name = &parse_node_name(&cmd.node_name)?;
```
adrianbenavides, line 190: If we remove this function we lose the ability to accept both `/node/n1` and `n1` as valid node names. Is that something we want to do? I would vote for just accepting `n1`.

implementations/rust/ockam/ockam_command/src/node/mod.rs
```diff
@@ -81,82 +79,3 @@ pub struct NodeOpts {
     #[arg(global = true, id = "at", value_name = "NODE_NAME", long)]
     pub at_node: Option<String>,
 }
-
-/// If the required node name is the default node but that node has not been initialized yet
-/// then initialize it
-pub fn initialize_node_if_default(opts: &CommandGlobalOpts, node_name: &Option<String>) {
```
adrianbenavides, line 87: _Observation_

We'll lose some of the default behaviors if we remove this (unless I'm missing its replacement), but I can see how removing it helps simplify the changes done in the command layer.

implementations/rust/ockam/ockam_command/src/project/create.rs
```diff
@@ -50,33 +50,14 @@ async fn run_impl(
...
     let operation_id = project.operation_id.clone().unwrap();
+    let controller = node.create_controller().await?;
     check_for_completion(&opts, ctx, &controller, &operation_id).await?;
     let project = check_project_readiness(&opts, ctx, &node, project).await?;
-    opts.state
-        .projects
-        .overwrite(&project.name, project.clone())?;
-    opts.state
-        .trust_contexts
-        .overwrite(&project.name, project.clone().try_into()?)?;
```
adrianbenavides, line 68: _Appreciation_

Love how the state management is now hidden behind the controller call!

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -189,97 +188,89 @@ pub async fn enroll_with_node(
...
+                    .clone()
+                    .color(OckamColor::PrimaryResource.color())
+            ))?;
+            space.clone()
+        }
     };
-    opts.state
-        .spaces
-        .overwrite(&default_space.name, SpaceConfig::from(&default_space))?;
+    opts.state.set_space_as_default(&space.id).await?;
```
adrianbenavides: Maybe we could add the `set_X_as_default` call inside the `create_X` function, so we just use the `set_X_as_default` function when we want to explicitly change the default resource.

adrianbenavides (commented):
I've done a general pass over the `ockam_command` crate. I'll do a second review in a bit exploring the `ockam_api` crate.

implementations/rust/ockam/ockam_command/src/node/create.rs
etorreborre: This is now done inside the `Background::create` function, to avoid doing it everywhere. I think this is wrong though! We should rather have a `NodeName` struct and an associated parser which is able to parse from either `/node/n1` or `n1`.

implementations/rust/ockam/ockam_command/src/node/mod.rs
etorreborre: You're correct. The default node data is always created and stored when you need it but the default node itself is not started automatically. The reasons for doing this are:

 - that simplifies the calls as you mention
 - in most (all?) the scripts I've seen we create a node first before creating an outlet, or another resource on it. In other places, where we get information doing, `ockam credential get` with a default node created in the background makes little sense
 - since spawning a default creates a resource that takes a real process on a machine I think it is better to be explicit about what we want to be running or not
 
 Please tell me if I am missing a scenario where the default spawning of a node is really necessary.

implementations/rust/ockam/ockam_command/src/enroll/command.rs
etorreborre: This line is indeed redundant. When we create the first projet / space / identity etc... we set it as the default in `CliState::create_xxx`.

implementations/rust/ockam/ockam_identity/src/identities/identities_creation.rs
```diff
@@ -182,7 +197,7 @@ impl IdentitiesCreation {
     pub async fn update_identity(&self, identity: &Identity) -> Result<()> {
```
SanjoDeundiak, line 199: Do you think this function should be moved to the Repository implementation itself, and be wrapped into a transaction?

implementations/rust/ockam/ockam_command/tests/fixtures/user.enrollment.ticket
```diff
@@ -1 +1 @@
...
...6e6177732e636f6d2f7463702f343031352f736572766963652f617069222c227573657273223a5b5d2c2273706163655f6964223a22222c226964656e74697479223a224932396530356139383863363233336237303432643239396562343632626537626237333134343633222c22617574686f726974795f6163636573735f726f757465223a222f646e73616464722f6b38732d6875622d6e67696e78696e672d613631306264343233622d633135313863323965623936633463312e656c622e75732d776573742d312e616d617a6f6e6177732e636f6d2f7463702f343031362f736572766963652f617069222c22617574686f726974795f6964656e74697479223a2238316132303135383362613230313031303235383335613430323832303138313538323038373063346337366161393838323231383931653536323935613035353131613338323665306637353166646234353239373336353332323364623930326161303366343034316136353464376330653035316137383139376630653032383230313831353834306165636138653538393635396366316164373533336461396538363366633761653765643236323131333763343662303131366534396633383865363838343362323236313534366334316464363963636533393563343535643263353664393263306433366130353535376163353161666266303239373038666263363038222c2276657273696f6e223a6e756c6c2c2272756e6e696e67223a6e756c6c2c226f7065726174696f6e5f6964223a6e756c6c2c22757365725f726f6c6573223a5b5d7d7d
```
SanjoDeundiak: Why did it change?

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_types.rs
SanjoDeundiak: Ok, so what actually happens when you have a u64 and its values doesn't fit into i64, is that it gets overflowed (e.g., u64::MAX -> -1i64). Then when you retrieve it from the db, you convert it again and it is overflowed again, and now has the correct value again. I think it's far from ideal, but it works 😅 I don't have a better alternative atm

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_types.rs
SanjoDeundiak: I suggest we remove it. I don't see a reasoning of why we would store a type of an architecture-dependent size in the db. Let's force anyone who is trying to save it to do a proper conversion to a fixed size integer (and make sure it's correct).

implementations/rust/ockam/ockam_identity/src/identities/identities_creation.rs
SanjoDeundiak: The problem here is that Repository would need to depend on a VerifyingVault to do that...

implementations/rust/ockam/ockam_identity/src/identities/identities_creation.rs
```diff
@@ -159,6 +142,7 @@ impl IdentitiesCreation {
             return Err(IdentityError::WrongSecretKey.into());
         }
 
+        self.repository.store_change_history(&identity).await?;
```
SanjoDeundiak: I think it's part of `self.update_identity` call few lines above. This line can be removed

implementations/rust/ockam/ockam_identity/src/models/credential_and_purpose_key.rs
```diff
@@ -13,3 +20,82 @@ pub struct CredentialAndPurposeKey {
     /// [`Credential`] and will be used to verify it
     #[n(2)] pub purpose_key_attestation: PurposeKeyAttestation,
 }
+
+impl CredentialAndPurposeKey {
+    /// Encode the credential as a hex String
+    pub fn encode_as_string(&self) -> Result<String> {
```
SanjoDeundiak, line 25: I wonder if we can auto-implement these 4 functions for all types implementing `Encode` and `Decode`? Also, I would consider more specific functions name, something like `encode_as_cbor_bytes`

SanjoDeundiak: P.S. This is a very good change

implementations/rust/ockam/ockam_api/src/bootstrapped_identities_store.rs
```diff
@@ -183,7 +137,7 @@ impl From<HashMap<Identifier, AttributesEntry>> for PreTrustedIdentities {
 }
 
 #[async_trait]
-impl IdentityAttributesReader for PreTrustedIdentities {
```
SanjoDeundiak, line 186: I'll be getting rid of this mess related to "Bootstrap" mechanism soon

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_command/src/node/create.rs
etorreborre: Good idea. Done.

implementations/rust/ockam/ockam_command/tests/fixtures/user.enrollment.ticket
etorreborre: Because I simplified the `EnrollmentTicket` to use the `Project` data type directly. Is this going to be a problem with the Controller maybe?

implementations/rust/ockam/ockam_identity/src/identities/identities_creation.rs
etorreborre: Done

implementations/rust/ockam/ockam_command/tests/fixtures/user.enrollment.ticket
SanjoDeundiak: I don't think so, but let's ask @polvorin

implementations/rust/ockam/ockam_identity/src/models/credential_and_purpose_key.rs
etorreborre: > I wonder if we can auto-implement these 4 functions for all types implementing Encode and Decode? Also, I would consider more specific functions name, something like encode_as_cbor_bytes

Yes that could be some useful extension methods. Let's do this in another PR.

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_types.rs
etorreborre: Done. There was an `id` field in `ProjectUserRole` using `usize` but I replaced it with `u64`.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
davide-baldo: @etorreborre 
I think the application should have its own data structure without relying on current status of the outlet.
We need to associate information beyond the simple (alias, ockam_addess, socket_address): like URL components (schema), and probably also some kind of check-alive, maybe even some graphical representation (just making a point that it's a higher-level data structure IMO).

implementations/rust/ockam/ockam_api/src/cli_state/storage/identities_repository_sql.rs
```diff
@@ -0,0 +1,304 @@
...
+        ))
+    }
+
+    async fn delete_identity(&self, name: &str) -> Result<Option<Identifier>> {
+        let named_identity = self.get_named_identity(name).await?;
+        let is_default = named_identity
+            .clone()
+            .map(|i| i.is_default())
+            .unwrap_or(false);
+        let query = query("DELETE FROM named_identity WHERE name=?").bind(name.to_sql());
```
davide-baldo, line 73: Why are you sometimes using `?` instead of `$1`?

implementations/rust/ockam/ockam_api/src/cli_state/storage/credentials_repository.rs
```diff
@@ -0,0 +1,26 @@
...
+use ockam_core::Result;
+
+/// This repository support the storage of credentials retrieved from the command line
+/// A credential is associated with a name and its issuer for later retrieval
+#[async_trait]
+pub trait CredentialsRepository: Send + Sync + 'static {
+    /// Store a CredentialAndPurposeKey under a given name
+    /// The issuer of the credential is also stored in order to be able to validate the credential
+    /// later on
+    async fn store_credential(
```
davide-baldo, line 14: I don't understand how a credential name would be used

implementations/rust/ockam/ockam_api/src/cli_state/storage/enrollments_repository.rs
```diff
@@ -0,0 +1,30 @@
...
+/// This trait stores the enrollment status for identities
+/// If an identity has been enrolled it is possible to retrieve:
+///
+///  - its name (if it has one)
+///  - if this the default identity
+///  - if an identity was enrolled and when the local node was informed
+///
+///
+#[async_trait]
+pub trait EnrollmentsRepository: Send + Sync + 'static {
```
davide-baldo, line 16: From the description, it isn't clear whether it's about local identities enrollments or if it's the storage of an authority keeping track of enrolled identities.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
davide-baldo: I will update the structure after the merge

implementations/rust/ockam/ockam_api/src/cli_state/storage/identities_repository_sql.rs
etorreborre: Both can be used. I have indeed not been very consistent. I see if I can do a pass to adjust this.

implementations/rust/ockam/ockam_api/src/cli_state/storage/credentials_repository.rs
etorreborre: A credential name is used in a few places in the command line API. For example, when creating a node you can pass a credential name for a credential you previously stored.

implementations/rust/ockam/ockam_api/src/cli_state/storage/enrollments_repository.rs
etorreborre: This is mostly for storing a flag that we keep locally to know if an identifier has been enrolled on the default project. In a way this can also be seen as a cache for something that happens on a remote authority. Please suggest something if you have a better name.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20231006100000_create_database.sql
etorreborre: @davide-baldo certainly the desktop application data types can be reworked a bit, both in terms of content and naming.

implementations/rust/ockam/ockam_api/src/cli_state/storage/enrollments_repository.rs
davide-baldo: I think just adding `local` to this sentence should be enough:

`This trait stores the enrollment status for local identities`

etorreborre: Done.

implementations/rust/ockam/ockam_api/src/cli_state/storage/identities_repository_sql.rs
etorreborre: Now, I remember think about it along the lines of "when there are more than 5 or so parameters having indices can be useful to quickly count them or check their position". And otherwise use `?`

implementations/rust/ockam/ockam_command/tests/fixtures/user.enrollment.ticket
etorreborre: Since the secure channels protocol changes also changes the ticket format, we should know with that PR first.

polvorin (dismissed):

polvorin (approved):

PR #6926, opened 2023-11-20
feat: added `About` dialogue when pressing option

davide-baldo:
Added `about` window to keep track of the version being used. This window is accessible via a new `About` menu entry, when pressing the option key, just above the `Reset` entry (see below screenshot).

It optionally also shows `OCKAM_HOME` `OCKAM_CONTROLLER_ADDR` and `OCKAM_CONTROLLER_IDENTITY_ID` when they are set.

Screenshot 2023-11-20 at 17 40 44
Screenshot 2023-11-20 at 17 40 56
Screenshot 2023-11-20 at 17 50 03

The information is retrieved using a new API called `runtime_information()`, in this new API I also implemented the relative call `free_runtime_information()`, and by doing so, I realized there was a bogus `libc` dependency which I removed. 
Eventually (when the structure is stable-ish), we'll have to implement the `free_*` version of the application status and notifications as well.


nazmulidris (approved):
LGTM! 🎉

PR #6931, opened 2023-11-21
feat(rust): `tcp-inlet create` `--to` works passing either a multiaddr or the relay name

adrianbenavides:
When we are creating tcp-inlets targeting a service that is behind a project relay, most of the `--to` argument's content is boilerplate when using default values. We usually run:

```
ockam tcp-inlet create --to /project/default/service/forward_to_"$NAME"/secure/api/service/outlet ...
```

This PR introduces an argument parser so we can additionally define the `--to` argument by passing just the relay name, which will transform it to the previous MultiAddr route.

```
ockam tcp-inlet create --to $NAME ...
```

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
@@ -5,13 +5,15 @@ use std::time::Duration;
 
 use clap::Args;
 use colorful::Colorful;
-use miette::miette;
+use miette::{miette, IntoDiagnostic};
+use miette::{Result, WrapErr};
 use tokio::sync::Mutex;
 use tokio::try_join;
 use tracing::log::trace;
 
 use ockam::identity::Identifier;
 use ockam::Context;
+use ockam_api::cli_state::{CliState, StateDirTrait};
 
 use ockam_api::nodes::models::portal::InletStatus;
 use ockam_api::nodes::service::portals::Inlets;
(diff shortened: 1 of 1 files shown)


implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
```diff
@@ -82,6 +83,29 @@ fn default_to_addr() -> MultiAddr {
...
+        Ok(ma) => Ok(ma),
+        Err(_) => {
+            if addr_or_relay_name.contains('/') {
+                return Err(clap::Error::raw(
+                    ErrorKind::InvalidValue,
+                    "The relay name can't contain '/'",
+                ));
+            }
+            MultiAddr::from_str(&format!(
+                "/project/default/service/forward_to_{addr_or_relay_name}/secure/api/service/outlet"
```
etorreborre: I wonder if this would be a good idea to also have an optional `--project` option for this command to keep the possibility to use a different project. Especially if we start giving random names to projects.

Alternatively we can avoid using a `--project` option for now but use the name of the default project to build the `MultiAddr`. In that case we need a value parser parsing a value of type `MultiAddrOrRelayName`:
```rust
enum MultiAddrOrRelayName {
  MultiAddress(MultiAddr),
  RelayName(String),
}
```
and later on we build the route using the default project name. This way, we are a bit more robust if the default project name changes.

adrianbenavides: I can see the usefulness of having the `--project` argument. And definitely, I should read the default project name from the state. We can add the `--project` argument in a separate PR.

adrianbenavides: I've changed the approach on how we parse the arguments. The problem is that ALL the Clap parsers are evaluated before running the command, no matter the command you are running. In this case, to retrieve the default project name, we have to initialize the CliState, which can be a costly operation, so I've moved that away from the Clap parsers into a function of the command. This also allows us to handle errors better.

etorreborre (commented):
🏎️

etorreborre (approved):

PR #6936, opened 2023-11-21
feat(rust): add an env var to control the ockam authenticator's base url

adrianbenavides:


implementations/rust/ockam/ockam_api/src/enroll/oidc_service.rs
```diff
@@ -138,7 +142,7 @@ impl OidcService {
             authorization_code.code
         );
         self.request_code(
-            Url::parse("https://account.ockam.io/oauth/token").unwrap(),
+            Url::parse(&format!("{}/oauth/token", Self::authenticator_endpoint())).unwrap(),
```
adrianbenavides, line 145: @etorreborre is there any reason to have this URL hardcoded here instead of using the one from `self.provider().token_request_url`?

implementations/rust/ockam/ockam_api/src/enroll/oidc_service.rs
```diff
@@ -230,7 +234,7 @@ impl OidcService {
 
                     // Note that a code 303 does not properly redirect to the success page
                     let response = Response::empty(302).with_header(
-                        Header::from_str("Location: https://account.ockam.io/device/success")
+                        Header::from_str(&format!("Location: {}/device/success", Self::authenticator_endpoint()))
```
adrianbenavides, line 237: Similarly, should this and the next urls be part of the `OidcProvider` trait?

implementations/rust/ockam/ockam_api/src/enroll/ockam_oidc_provider.rs
```diff
@@ -1,11 +1,19 @@
+use ockam_core::env::get_env;
 use ockam_core::Result;
 use std::time::Duration;
 use url::Url;
 
 use crate::enroll::oidc_provider::OidcProvider;
 
+pub fn authenticator_endpoint() -> String {
+    get_env("OCKAM_AUTHENTICATOR_ENDPOINT")
```
snandam: Not sure if it's relavent, but can this be added to [this page](https://github.com/build-trust/ockam/blob/744bd831d5efd77bb3c15c8136dec243a0e608bb/implementations/rust/ockam/ockam_command/src/environment/static/env_info.txt#L19) as well similar to `OCKAM_CONTROLLER_ADDR` and `OCKAM_CONTROLLER_IDENTITY_ID`  - Those are the two values we set explicitly during development environment testing. 

And i'm assuming the default value for `OCKAM_AUTHENTICATOR_ENDPOINT` will be set to https://account.ockam.io rite? Asking because I didn't see that set any place.

adrianbenavides: Sure! I wasn't sure about doing it because it's on a separate crate... but it's also the only file where we keep track of the env vars.

adrianbenavides: > And i'm assuming the default value for OCKAM_AUTHENTICATOR_ENDPOINT will be set to https://account.ockam.io/ rite? Asking because I didn't see that set any place.

I thought `https://account.ockam.io/` was going to be replaced by something else. Is that still the default value?

snandam: Yes `https://account.ockam.io/` is the default value to compile the binaries with. During testing, I shall export `OCKAM_AUTHENTICATOR_ENDPOINT` to another value to match our development environment. Similar `OCKAM_CONTROLLER*` values we override during development environment testing

polvorin (approved):

implementations/rust/ockam/ockam_api/src/enroll/oidc_service.rs
etorreborre: The `token_request_url` for the `OktaOidcProvider` has a different format. It is `Url::parse(format!("{}/v1/token", &self.okta.tenant_base_url).as_str())`. This lines looks like a bug indeed and the `tenant_base_url` function should probably be called.

implementations/rust/ockam/ockam_api/src/enroll/oidc_service.rs
etorreborre: In that case, that's the landing page when the authentication was successful. So it should be the same regardless of the provider.

PR #6937, opened 2023-11-22
feat(elixir): identity and secure channel protocol updates

polvorin:
* Implement credential refresh, close and data messages as defined in the secure channel messaging protocol.

* Adapt to CBOR encoding changes around identities.

* Add cbor encoding support for enums with inner values (`Payload(payload_msg)` , etc), compatible with minicbor

* Make use of OS-assigned listening ports on tests to avoid conflicts, add a way to obtain the port that was assigned.


implementations/elixir/ockam/ockam/test/ockam/secure_channel/key_establishment_protocol/xx/protocol_test.exs
```diff
@@ -1,103 +1,4 @@
 defmodule Ockam.SecureChannel.KeyEstablishmentProtocol.XX.Protocol.Tests do
   use ExUnit.Case, async: true
   doctest Ockam.SecureChannel.KeyEstablishmentProtocol.XX.Protocol
-
-  alias Ockam.SecureChannel.KeyEstablishmentProtocol.XX.Protocol
```
SanjoDeundiak: We definitely can't remove them. We should either keep them, or bring them back soon after the merge

SanjoDeundiak (approved):

implementations/elixir/ockam/ockam/test/ockam/secure_channel/key_establishment_protocol/xx/protocol_test.exs
polvorin: These vectors where from public ones that don't work anymore due to the protocol name change.   You mean to setup vectors for whatever we produce now,  or make an option to use the previous  name and use that option for this test?

PR #6944, opened 2023-11-23
feat(rust): Add the possibility to ignore an accepted service

davide-baldo:
The idea is to allow removal of incoming services by ignoring the originating invite, it's far from ideal, but it's working with current invitations implementation.
When clicking `ignore`, there is a popup confirmation.

image
image

### DIFF
--- a/implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
+++ b/implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
@@ -29,6 +29,12 @@ impl IncomingServicesState {
         self.services.iter().find(|s| s.id == id)
     }
 
+    pub(crate) fn remove_by_id(&mut self, id: &str) {
+        if let Some(index) = self.services.iter().position(|s| s.id == id) {
+            self.services.remove(index);
+        }
+    }
+
     pub(crate) fn find_mut_by_id(&mut self, id: &str) -> Option<&mut IncomingService> {
         self.services.iter_mut().find(|s| s.id == id)
     }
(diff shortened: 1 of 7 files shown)


nazmulidris (approved):
LGTM

implementations/swift/ockam/ockam_app/Ockam/IgnoreService.swift
```diff
@@ -0,0 +1,56 @@
...
+struct IgnoreServiceView: View {
+    @Environment(\.presentationMode) var presentationMode: Binding<PresentationMode>
+    @State var service: Service
+
+    var body: some View {
+        VStack {
+            Text(
+                "Once you click Ignore, the service '\(service.sourceName)' will no longer show up here.\n" +
+                "Are you sure you want to do this?\n" +
+                "Once ignored, the only way to get this back is to have the person who sent you the invite, to send another one."
```
adrianbenavides, line 12: Perhaps this could be reworded/reorganized a little, for example:

```
"Are you sure you want to do this?\n"}
"The service '\(service.sourceName)' you will no longer have access to it and you will have to request a new invite to the service's owner."
```

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
```diff
@@ -165,6 +185,9 @@ pub struct IncomingService {
     // this enrollment ticket is modified to avoid conflicts with 'default' name
     // the name of the project is re-set to 'project_id'
     enrollment_ticket: EnrollmentTicket,
+    // When the invitation is removed, the service is marked as removed
+    // to clean up the resources before removing the service from the list
+    removed: bool,
```
adrianbenavides, line 190: This variable is a little bit misleading. I'd rename it to something more explicit, like `marked_to_be_removed`.

implementations/rust/ockam/ockam_app_lib/src/incoming_services/state.rs
```diff
@@ -233,6 +261,16 @@ impl IncomingService {
         self.port = None;
     }
 
+    /// True when the service is marked as removed
+    pub fn removed(&self) -> bool {
```
adrianbenavides, line 265: I would also rename this to `is_marked_to_be_removed` to keep it consistent with the other comment.

adrianbenavides (commented):
Looks good! I just left some minor comments.

PR #6978, opened 2023-11-28
fix(rust): fix feature config for aws-sdk-kms

polvorin:
This fix the aws kms vault,  that was broken after a recent update of the `aws-sdk-rust` dependency that contained a  a breaking change.

From https://github.com/awslabs/aws-sdk-rust/blob/main/CHANGELOG.md

> Client creation now panics if retries or timeouts are enabled without an async sleep implementation set on the SDK config. If you're using the Tokio runtime and have the rt-tokio feature enabled (which is enabled by default),  

We are disabling default features,  so this was breaking and getting a panic at runtime.   My first attempt was to just enable the `rt-tokio`  feature and while that did work,  it has the side effect of breaking our no_std checks.   So instead, I  copy-pasted the `AsyncSleep`  implementation that would have been available if that feature was on,  and tweaked the imports a bit to get to tokio through `ockam::compat`
(so, adding a dependency from ockam_vault_aws ->  ockam ,   that was not present before).

### DIFF
--- a/implementations/rust/ockam/ockam_vault_aws/Cargo.toml
+++ b/implementations/rust/ockam/ockam_vault_aws/Cargo.toml
@@ -57,6 +57,7 @@ credentials-sso = ["aws-config/sso"]
 aws-config = { version = "1.0.1", default-features = false, features = ["rustls"] }
 aws-sdk-kms = { version = "0.39.0", default-features = false, features = ["rustls"] }
 delegate = "0.10.0"
+ockam = { path = "../ockam", version = "^0.105.0", default_features = false }
 ockam_core = { path = "../ockam_core", version = "^0.93.0", default_features = false }
 ockam_macros = { path = "../ockam_macros", version = "^0.32.0", default-features = false }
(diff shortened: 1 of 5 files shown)


davide-baldo (approved):

implementations/rust/ockam/ockam_vault_aws/Cargo.toml
```diff
@@ -57,6 +57,7 @@ credentials-sso = ["aws-config/sso"]
 aws-config = { version = "1.0.1", default-features = false, features = ["rustls"] }
 aws-sdk-kms = { version = "0.39.0", default-features = false, features = ["rustls"] }
 delegate = "0.10.0"
+ockam = { path = "../ockam", version = "^0.105.0", default_features = false }
```
etorreborre, line 60: We can avoid using this dependency by using
```
ockam_node::tokio::time::sleep
```
in `TokioSleep`:
```rusnt
Sleep::new(ockam_node::tokio::time::sleep(duration))
```

etorreborre (approved):

PR #6979, opened 2023-11-28
Refactor attribute types

etorreborre:
Attributes are used to create credentials for a given identity. When a credential is received and validated, its attributes are stored by a node. Then resources can have associated policies declaring that an identity can access it only if the attributes associated to that identity satisfies a logic expression.

The `ockam_abac` crate defines the exact type of values which can be used in expressions: string, bool, int, float, and a sequence of such values. Moreover, throughout the code attributes names are always expected to be strings (which is reasonable).

Yet we model attributes in our code as `name: Vec<u8>` and `value: Vec<u8>`. This is unnecessarily general. By having proper types for attribute names and values we signal to the library what can be really used and we avoid making mistakes by passing incompatible values.

This PR changes the type of `AttributesEntry` so that it contains a map of `AttributeName` -> `AttributeValue` pairs.

### Notes

 - the actual encoding / decoding of data exchanged with the Controller is still left as a pair of `Vec<u8>` where effectively those bytes only ever represent UTF-8 encoded strings at the moment (even for values)

 - the `AttributeValue` data type has been introduced to separate the declarations / persistence of attributes in the `ockam_identity` repository from their potential use as `Expr`essions in the `ockam_abac` crate (since there is a dependency from `ockam_abac` to `ockam_identity`)


etorreborre:
Closing for now

PR #6980, opened 2023-11-28
feat(rust): add `--relay` argument to `project ticket` command

adrianbenavides:
Fixes https://github.com/build-trust/codebase/issues/616

### DIFF
--- a/implementations/rust/ockam/ockam_command/tests/bats/portals_orchestrator.bats
+++ b/implementations/rust/ockam/ockam_command/tests/bats/portals_orchestrator.bats
@@ -77,12 +77,12 @@ teardown() {
   blue_identifier=$($OCKAM identity show blue)
 
   fwd="$(random_str)"
-  run_success "$OCKAM" node create green --project "$PROJECT_NAME" --identity green --relay $fwd
+  run_success "$OCKAM" node create green --project "$PROJECT_NAME" --identity green
   run_success "$OCKAM" node create blue --project "$PROJECT_NAME" --identity blue
 
   # Green isn't enrolled as project member
   OCKAM_HOME=$ENROLLED_OCKAM_HOME
-  run_success "$OCKAM" project ticket --member "$blue_identifier" --attribute role=member
+  run_success "$OCKAM" project ticket --member "$blue_identifier" --attribute role=member --relay $fwd
 
   OCKAM_HOME=$NON_ENROLLED_OCKAM_HOME
   run_success "$OCKAM" tcp-outlet create --at /node/blue --to 127.0.0.1:5000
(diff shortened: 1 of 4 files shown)


implementations/rust/ockam/ockam_command/tests/bats/portals_orchestrator.bats
```diff
@@ -79,6 +79,7 @@ teardown() {
   run_success "$OCKAM" node create green --project-path "$PROJECT_JSON_PATH" --identity green
   run_success "$OCKAM" node create blue --project-path "$PROJECT_JSON_PATH" --identity blue
 
+  -- TO_REVIEW --
```
adrianbenavides, line 82: What happens in this kind of scenario, where the admin enrolls a new identity as a member of the project? Should we also include the `--allowed-relay` argument here?

polvorin:
The intention for this is to be entirely a UX side thing.    There is no change on authority to support another field,  nor on abac rules to enforce that other field.   Instead of using lets say

```
ockam ticket --attribute location=ny --attribute ockam-relay-name=somename
```

we would allow to do

```
ockam ticket --attribute location=ny --relay-name somename
```

with the latter be effectively the _equivalent_ to the former,  only that not requiring the caller to remember the exact attribute name that is used for enforcing this (`ockam-relay-name`). 

(a similar shortcut should be done for the `ockam-role` attribute, it was discussed but not sure we reached a conclusion on how to name it, I'd say let it to be done on another PR later)

adrianbenavides:
Awesome, thanks @polvorin for the clarification. Will update the PR ASAP!

adrianbenavides:
@polvorin I have a question about how this attribute is used by the different parts. 

Let's say the admin creates a ticket for a user, and the ticket now has the `ockam-relay` attribute. Now, when the member that used the ticket to enroll wants to create a relay, how/when is that attribute checked? Does the name of the attribute must match a name that is checked in the backend?

polvorin:
> @polvorin I have a question about how this attribute is used by the different parts.
> 
> Let's say the admin creates a ticket for a user, and the ticket now has the `ockam-relay` attribute. Now, when the member that used the ticket to enroll wants to create a relay, how/when is that attribute checked? Does the name of the attribute must match a name that is checked in the backend?

It's checked and enforced by the backend.     (at the moment.. only for elixir backend).   it's is [here on the open PR](https://github.com/build-trust/ockam/pull/6948/files#diff-3dd693993cd73aa0e1200fa005362ee33b3c2199b755e7363b72e4273656ee1bR97-R108), but need to adapt it as it is using the wrong attribute name ("allow_relay_address").

Other than to enforce this on the server side,   on the CLI it _can_ be used as the default relay name to use when creating a relay, if one is not specified.   That means the cli can request its credential, check for the presence of this attribute, and use it as the relay name to create if one was not provided.  However, that has some drawbacks:
* It must differentiate between intention to create a relay in a project' node,   or on a random standalone node (where credentials aren't involved.. this differentiation is already done for other reasons.. but still)
* It need to interpret the "*" wildcard as "well, you can use _any_ relay name you want, so you must explicitly tell us the name to use when creating one"

adrianbenavides:
> Other than to enforce this on the server side, on the CLI it can be used as the default relay name to use when creating a relay, if one is not specified.

For this part, instead of going back and forth between the CLI and the backend, couldn't the backend just fallback to the relay name available in the attribute when the CLI doesn't specify a relay name?

etorreborre (approved):

PR #6988, opened 2023-11-29
fix(rust): honor the `timeout` arg value on `status` command

adrianbenavides:
The `timeout` argument was not being used in the `status` command. This PR modifies the `InMemoryNode` so that in passes down its timeout value to both the `secure_channel` creation and the calls done to the node.


etorreborre (approved):
⏲️

PR #6989, opened 2023-11-29
ui redesign consistent spacing and style, minor usability improvements

davide-baldo:
This PR addresses mostly UI and usability issues:

- Removed creation & share in same view [part of ID_25]
- Use `...` only when the focus is going to change (usually a new window pops up) [ID_26]
- UI refactor: used a consistent spacing, moved profile pictures to the right, spacing instead of dividers, consistent font sizes and colors [ID_40]
- Using "decline" invitation instead of "ignore" [ID_41]
- Changed the accent color to Ockam accent color [ID_43]
- Refactored views into their own file and relative preview

Also, this library was adopted: [fluid menu bar extra](https://github.com/lfroms/fluid-menu-bar-extra/tree/main) to avoid focus issues and glitches when resizing the window. The library should be fetched automatically by XCode from the [URL](https://github.com/build-trust/ockam/pull/6989/files#diff-700a2fd0aea6a9bbff8d903a15409f99345dda0f360142ee6ea4a041a8e5dbefR488).
I plan to either fully integrate the library into the source code or to fork the library inside `build-trust` org.

The library re-implement the menubar extra from scratch, bypassing SwiftUI APIs and avoiding many pitfalls. By using lower level, with relative minor modification of the library, we should also have better control of when and how to show the main popover.

Unfortunately, it introduces a **regression**: dismissing the main popover only works once; So, subsequent clicks on "copy" or "learn more..." won't close automatically the popover. [ID_44]


implementations/swift/ockam/ockam_app/Ockam/Constants.swift
```diff
@@ -0,0 +1,7 @@
+import SwiftUI
+
+let VerticalSpacingUnit: CGFloat = 8
+let HorizontalSpacingUnit: CGFloat = 8
+
+let OckamDarkerBackground = Color.black.opacity(0.1)
+let OckamSecondaryTextColor = Color.primary.opacity(0.8)
```
etorreborre, line 7: Just to double-check: Ockam "secondary" but Color "primary"?

etorreborre (approved):

implementations/swift/ockam/ockam_app/Ockam/Constants.swift
davide-baldo: Yes, the only way to stay consistent with the theme is to modify the pre-defined colors. In this case, I started from the primary.

PR #6991, opened 2023-11-29
Fix persistence issues

etorreborre:
This PR fixes a number of regressions after the merge of the [database persistence PR](https://github.com/build-trust/ockam/pull/6913):

 - incorrect use of the space id instead of the space name in a command
 - create the authority identity for the identity node if it doesn't exist
 - correctly stop nodes in the bats test suite when running concurrently
 - create a default identity only once when using the function accepting optional values

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
@@ -49,17 +49,40 @@ impl CliState {
         self.create_identity_with_name_and_vault(name, &vault.name())
             .await
     }
+
     /// Create an identity associated with an optional name and an optional vault name
-    /// If the vault name is not specified then the default vault is used
+    ///  - if the vault name is not specified then the default vault is used
+    ///  - if no name is specified:
+    ///     - if there is no default identity, create an identity with a random name
+    ///     - if there is a default identity with the correct vault, return it
+    ///     - otherwise create a new identity
+    ///  - if a name is given create a new identity with the proper vault (unless it has already been created)
     pub async fn create_identity_with_optional_name_and_optional_vault(
         &self,
         name: &Option<String>,
         vault_name: &Option<String>,
     ) -> Result<NamedIdentity> {
-        let name = name.clone().unwrap_or_else(random_name);
-        let vault = self.get_named_vault_or_default(vault_name).await?.name();
-        self.create_identity_with_name_and_vault(&name, &vault)
-            .await
+        let vault_name = self.get_named_vault_or_default(vault_name).await?.name();
+        match name {
+            Some(name) => {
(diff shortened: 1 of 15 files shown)


implementations/rust/ockam/ockam_identity/src/error.rs
```diff
@@ -1,17 +1,19 @@
+use ockam_core::compat::string::String;
 use ockam_core::{
     errcode::{Kind, Origin},
     Error,
 };
 
 /// Identity crate error
+#[repr(u8)]
```
SanjoDeundiak, line 8: Why is that required?

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_identity/src/error.rs
etorreborre: Because I have added a `String` to the `InvalidIdentifier` error.

implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
```diff
@@ -49,17 +49,40 @@ impl CliState {
...
+                    .await
+            }
+            None => {
+                let default_identity = self
+                    .identities_repository()
+                    .await?
+                    .get_default_named_identity()
+                    .await?;
+                if let Some(default_identity) = default_identity {
+                    if default_identity.vault_name == vault_name {
```
davide-baldo: What if the user didn't provide a vault name? I don't think we need to check that the default user belongs to the default vault.

etorreborre: I can indeed skip this check. At a fundamental level we have an issue with our command line options and actions. We can define a default identity, a default vault, other identities and vaults. Then we can use `ockam identity default` and `ockam vault default` to change either the default identity or the default vault so that they end-up being inconsistent with each other.

I think that the proper fix is to separate the domains and never allow to pass a `--vault` option when `--identity` is specified. And if we pass `--identity` we always expect it to exist.

davide-baldo: I agree, passing vault when we are already passing an identity is confusing. I like your proposal of mutually exclude vaults and identities. The proposed solution would also prevent us from lazily creating identities. A middle ground could be change `--vault` with `--vault-when-creating-identity` (maybe with better wording: )

etorreborre: I have now simply removed that function which was just used in 2 places so we don't get to deal with all the combinations.

SanjoDeundiak (dismissed):

davide-baldo (dismissed):

polvorin (approved):

PR #7000, opened 2023-11-30
feat(rust): dynamically parse the `--at` argument for `relay create` command

adrianbenavides:
In the `relay create` command, replace the hardcoded `/project/default` value used in the `--at` argument by `/project/<default_name>`, where the default project name is resolved dynamically.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/relay/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/relay/create.rs
@@ -2,7 +2,6 @@ use std::str::FromStr;
 
 use clap::Args;
 use colorful::Colorful;
-use miette::Context as _;
 use miette::{miette, IntoDiagnostic};
 use tokio::sync::Mutex;
 use tokio::try_join;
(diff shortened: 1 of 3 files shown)


implementations/rust/ockam/ockam_command/src/relay/create.rs
```diff
@@ -40,63 +39,96 @@ pub struct CreateCommand {
...
+    at: String,
 
     /// Authorized identity for secure channel connection
-    #[arg(long, id = "AUTHORIZED", display_order = 900)]
+    #[arg(long, id = "AUTHORIZED")]
     authorized: Option<Identifier>,
 }
 
+fn default_at_addr() -> String {
+    "/project/$PROJECT_NAME".to_string()
```
davide-baldo: What about using a more predictable name, like `$DEFAULT_PROJECT_NAME`?

davide-baldo (dismissed):
👍 
I still would love to see `--at-project` and `--at-node` though

adrianbenavides:
> I still would love to see --at-project and --at-node though

That's something we have to finish discussing, but I'm 100% in with that.

davide-baldo (approved):

PR #7001, opened 2023-11-30
fix(rust): use the same in-memory database for the vault used in nif

etorreborre:
This PR:

 - fixes the creation of the software vault in NIF code by making sure that the in-memory database is shared by all special purpose vaults (signing, secure channels, etc...)
 
 - uses a random URL for the in-memory database and specify that the shared mode should be used to avoid the issue described [here](https://github.com/launchbadge/sqlx/issues/2510)

### DIFF
--- a/implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
+++ b/implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
@@ -12,11 +12,14 @@ use ockam_identity::{
     utils::AttributesBuilder,
     Identifier, Identities, Vault,
 };
+use ockam_node::database::SqlxDatabase;
 use ockam_vault::{
     EdDSACurve25519SecretKey, HandleToSecret, SigningKeyType, SigningSecret,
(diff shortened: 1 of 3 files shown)


implementations/elixir/ockam/ockly/native/ockly/src/lib.rs
```diff
@@ -89,17 +92,20 @@ fn identities_ref() -> NifResult<Arc<Identities>> {
 
 fn load_memory_vault() -> bool {
     block_future(async move {
-        let identity_vault = SoftwareVaultForSigning::create().await.unwrap();
-        let secure_channel_vault = SoftwareVaultForSecureChannels::create().await.unwrap();
+        let database = SqlxDatabase::in_memory("in-memory-vault").await.unwrap();
```
davide-baldo, line 95: why no connection pool is created for the volatile vault?

polvorin (approved):

polvorin:
closing it, replaced by https://github.com/build-trust/ockam/pull/7003

PR #7005, opened 2023-11-30
fix(rust): use the transaction api properly with sqlx

etorreborre:
This PR fixes the use of transactions with `sqlx`

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/storage/identities_repository_sql.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/storage/identities_repository_sql.rs
@@ -40,7 +40,6 @@ impl IdentitiesRepository for IdentitiesSqlxDatabase {
         name: &str,
         vault_name: &str,
     ) -> Result<NamedIdentity> {
-        let transaction = self.database.begin().await.into_core()?;
         let is_already_default = self
             .get_default_named_identity()
             .await?
@@ -53,8 +52,6 @@ impl IdentitiesRepository for IdentitiesSqlxDatabase {
             .bind(vault_name.to_sql())
             .bind(is_already_default.to_sql());
         query.execute(&self.database.pool).await.void()?;
-
-        transaction.commit().await.void()?;
 
         Ok(NamedIdentity::new(
             identifier.clone(),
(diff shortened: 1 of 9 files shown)


implementations/rust/ockam/ockam_identity/src/identities/storage/identity_attributes_repository_sql.rs
```diff
@@ -74,17 +74,13 @@ impl IdentityAttributesRepository for IdentityAttributesSqlxDatabase {
         attribute_name: Vec<u8>,
         attribute_value: Vec<u8>,
     ) -> Result<()> {
-        let transaction = self.database.begin().await.into_core()?;
```
SanjoDeundiak, line 77: Why do you think transaction is not required here?

implementations/rust/ockam/ockam_api/src/cli_state/storage/identities_repository_sql.rs
```diff
@@ -40,7 +40,6 @@ impl IdentitiesRepository for IdentitiesSqlxDatabase {
         name: &str,
         vault_name: &str,
     ) -> Result<NamedIdentity> {
-        let transaction = self.database.begin().await.into_core()?;
```
SanjoDeundiak, line 43: Why do you think transaction is not required here?

implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
```diff
@@ -49,14 +49,14 @@ impl ProjectsSqlxDatabase {
 #[async_trait]
 impl ProjectsRepository for ProjectsSqlxDatabase {
     async fn store_project(&self, project: &Project) -> Result<()> {
-        let transaction = self.database.begin().await.into_core()?;
```
SanjoDeundiak, line 52: Why do you think transaction is not required here?

SanjoDeundiak (changes requested):

implementations/rust/ockam/ockam_api/src/cli_state/storage/identities_repository_sql.rs
etorreborre: In the 3 examples you pointed out, the code is reusing an existing query which is not executed with the transaction that was just opened. Some better code could be to include the query (not its execution) inside the transaction. But maybe some better behaviour would be to forbid an update for a `NamedIdentity` altogether. 
I think it's worth spending some time improving the storage semantics of those named/default entities but I propose that we do it in another PR.

SanjoDeundiak (approved):

PR #7006, opened 2023-11-30
refactor(rust):slim down the NodeManagerWorker (s_ch)

Darioazzali:
Refactor the NodeWorkerManager for secure channel and secure channel listener

## Current behavior

The `NodeWorkerManager` is responsable for too many tasks.

## Proposed changes

1. Created typed interface for message received from node _clients_ for:
- `/node/secure_channel`.
- `/node/secure_channel_listener`.
routes.  
Those typed struct can be renamed as the request type
(e.g:`CreateSecureChannelRequest`).  
In this way there are no _untyped_
requests.
In order for this to work all the calls to the node api should be changed.  
I've grepped all the files containing the requests calls.

2. Changed the `CreateSecureChannelResponse` constructor to accept
   `SecureChannel` `struct`
3. For coherence renamed `SecureChannelListenerList` to `ListSecureChannelListenerResponse`
4. Changed the implementations of the `From<ockam_api::Error>` trait for the
   `Response<Error>` in order to discern between a not found error or an
   internal error:
```rust
impl From<crate::Error> for Response<Error> {
    fn from(e: crate::Error) -> Self {
        match e.code().kind {
            Kind::NotFound => Response::not_found_no_request(e.into()),
            _ => Response::internal_error_no_request(&e.to_string()),
        }
    }
}
```
 In this way the error from the `NodeManager` method can be used to build the
 response.
 The `not_found_no_request` associated function has been introduced for return a
 not found error without requiring the request header.

5. Changed the various signature of `NodeWorkerManager` implementations to
   receive the typed requests instead of a `Decoder` and let the
   `From<minicbor::decore::Error>` implementation for t`Response<Error>` takes
   care of the wrong parsing.
6. Changed `get_secure_channel` and `get_secure_channel_listener` methods on
   the `NodeManager` to return `Result` instead of `Option`.
   Now, if the secure channel(secure_channel_listener) with the given address is
   not found the function returns a not found error

   ```rust
   Error::new(
     Origin::Api,
     Kind::NotFound,
     format!("Secure channel with address, {}, not found", addr),
   )
   ```

   This helps the `From<Error>` implementation for the `Response<Error>` so the
   `NodeWorkerManager` could automatically convert the error.

7. Changed `delete_secure_channel` and `delete_secure_channel_listener` methods
   on `NodeManager` to return explicitly a `Kind::NotFound` if the address of
   the deleting secure channel(secure channel listener) is not found

   ```rust
   Error::new(
   Origin::Api,
   Kind::NotFound,
   format!("Secure channel with address, {}, not found", addr),
   )

   ```

   The implementation consideration is the same for the deleting functions
   above.

8. An example to how the client request should be made from the cli client is given
   for:
   - create secure channel

**This PR won't pass the tests**, it's only showing how i would implement it, if it's ok for you @etorreborre I can change the given files for making the requests typed.

## Files to changes

Here I've simply grepped all the _Untyped Request_ to be changed if this
refactoring is ok for you.

#### CreateSecureChannelRequest

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs:236: .body(CreateSecureChannelRequest::new(
implementations/rust/ockam/ockam_api/src/nodes/service/relay.rs:452: let body = CreateSecureChannelRequest::new(

#### CreateSecureChannelListenerRequest

implementations/rust/ockam/ockam_command/src/secure_channel/listener/create.rs:65: CreateSecureChannelListenerRequest::new(
implementations/rust/ockam/ockam_command/src/util/api.rs:96: let payload = models::secure_channel::CreateSecureChannelListenerRequest::new(

#### DeleteSecureChannelRequest

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs:4: CreateSecureChannelRequest, CreateSecureChannelResponse, DeleteSecureChannelRequest,
implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs:277: .body(DeleteSecureChannelRequest::new(encryptor_address))

#### DeleteSecureChannelListenerRequest

implementations/rust/ockam/ockam_command/src/util/api.rs:117:) -> Request<models::secure_channel::DeleteSecureChannelListenerRequest> {
implementations/rust/ockam/ockam_command/src/util/api.rs:118: let payload = models::secure_channel::DeleteSecureChannelListenerRequest::new(addr);

#### ShowSecureChannelRequest

implementations/rust/ockam/ockam_command/src/util/api.rs:85:) -> Request<models::secure_channel::ShowSecureChannelRequest> {
implementations/rust/ockam/ockam_command/src/util/api.rs:86: let payload = models::secure_channel::ShowSecureChannelRequest::new(addr);

#### ShowSecureChannelListenerRequest

implementations/rust/ockam/ockam_command/src/util/api.rs:125:) -> Request<models::secure_channel::ShowSecureChannelListenerRequest> {
implementations/rust/ockam/ockam_command/src/util/api.rs:126: let payload = models::secure_channel::ShowSecureChannelListenerRequest::new(addr);

## Question
What to do with the various macro `info!("Handling the request to ...)`.
Up until now they were called only for two function at the
`NodeWorkerManager` level. only for some methods.

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/auth.rs
+++ b/implementations/rust/ockam/ockam_api/src/auth.rs
@@ -56,7 +56,8 @@ impl Server {
 
         let res = match req.method() {
             Some(Method::Get) => match req.path_segments::<2>().as_slice() {
-                [""] => Response::ok(&req)
+                [""] => Response::ok()
+                    .with_headers(&req)
                     .body(
                         self.identity_attributes_repository
                             .list_attributes_by_identifier()
@@ -70,7 +71,7 @@ impl Server {
                         .get_attributes(&identifier)
                         .await?
                     {
-                        Response::ok(&req).body(a).to_vec()?
+                        Response::ok().with_headers(&req).body(a).to_vec()?
                     } else {
                         Response::not_found(&req, &format!("identity {} not found", id)).to_vec()?
                     }
--- a/implementations/rust/ockam/ockam_api/src/authenticator/direct/authenticator.rs
+++ b/implementations/rust/ockam/ockam_api/src/authenticator/direct/authenticator.rs
@@ -88,17 +88,17 @@ impl Worker for DirectAuthenticator {
                     let add: AddMember = dec.decode()?;
                     self.add_member(&from, add.member(), add.attributes())
                         .await?;
-                    Response::ok(&req).to_vec()?
+                    Response::ok().with_headers(&req).to_vec()?
                 }
                 (Some(Method::Get), ["member_ids"]) => {
                     let entries = self.list_members().await?;
                     let ids: Vec<Identifier> = entries.into_keys().collect();
-                    Response::ok(&req).body(ids).to_vec()?
+                    Response::ok().with_headers(&req).body(ids).to_vec()?
                 }
                 (Some(Method::Get), [""]) | (Some(Method::Get), ["members"]) => {
                     let entries = self.list_members().await?;
 
-                    Response::ok(&req).body(entries).to_vec()?
+                    Response::ok().with_headers(&req).body(entries).to_vec()?
                 }
                 (Some(Method::Delete), [id]) | (Some(Method::Delete), ["members", id]) => {
                     let identifier = Identifier::try_from(id.to_string())?;
@@ -106,7 +106,7 @@ impl Worker for DirectAuthenticator {
                         .delete(&identifier)
                         .await?;
 
-                    Response::ok(&req).to_vec()?
+                    Response::ok().with_headers(&req).to_vec()?
                 }
 
                 _ => Response::unknown_path(&req).to_vec()?,
--- a/implementations/rust/ockam/ockam_api/src/authenticator/enrollment_tokens/acceptor.rs
+++ b/implementations/rust/ockam/ockam_api/src/authenticator/enrollment_tokens/acceptor.rs
@@ -70,7 +70,7 @@ impl EnrollmentTokenAcceptor {
             return Ok(Response::internal_error(req, "attributes storage error").to_vec()?);
         }
 
-        Ok(Response::ok(req).to_vec()?)
+        Ok(Response::ok().with_headers(req).to_vec()?)
     }
 }
 
--- a/implementations/rust/ockam/ockam_api/src/authenticator/enrollment_tokens/issuer.rs
+++ b/implementations/rust/ockam/ockam_api/src/authenticator/enrollment_tokens/issuer.rs
@@ -86,7 +86,7 @@ impl Worker for EnrollmentTokenIssuer {
                         .issue_token(&from, att.into_owned_attributes(), duration, ttl_count)
                         .await
                     {
-                        Ok(otc) => Response::ok(&req).body(&otc).to_vec()?,
+                        Ok(otc) => Response::ok().with_headers(&req).body(&otc).to_vec()?,
                         Err(error) => {
                             Response::internal_error(&req, &error.to_string()).to_vec()?
                         }
--- a/implementations/rust/ockam/ockam_api/src/nodes/models/secure_channel.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/models/secure_channel.rs
@@ -3,7 +3,7 @@ use std::time::Duration;
 use minicbor::{Decode, Encode};
 use serde::Serialize;
 
-use ockam::identity::{Identifier, DEFAULT_TIMEOUT};
+use ockam::identity::{Identifier, SecureChannel, DEFAULT_TIMEOUT};
 use ockam_core::flow_control::FlowControlId;
 use ockam_core::{route, Address, Result};
 use ockam_multiaddr::MultiAddr;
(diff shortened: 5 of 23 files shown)


etorreborre:
Thanks @Darioazzali. This overall looks good to me except for the implicit transformation of a `NodeManager` error into a `Response` without using the `RequestHeader`. The `RequestHeader` is used to provide a request id which we can use for tracking requests/responses on the client side (and in logs potentially). It would be better to avoid dropping this piece of information when we can.

Darioazzali:
@etorreborre thanks for the review.
Ok, maybe I've misunderstood, but isn't the request header handled directly by the caller, namely  `handle_message`?
On [this](https://github.com/build-trust/ockam/blob/5be4151f869b517169c97be39c6395831af39b73/implementations/rust/ockam/ockam_api/src/nodes/service.rs#L780C20-L780C20) line I see that the response is sent back to the client formatting the error and automatically inserting the request header. Isn't that correct?


```rust
        let r = match self.handle_request(ctx, &req, &mut dec).await {
            Ok(r) => r,
            Err(err) => {
                error! {
                    target: TARGET,
                    re     = %req.id(),
                    method = ?req.method(),
                    path   = %req.path(),
                    code   = %err.code(),
                    cause  = ?err.source(),
                    "failed to handle request"
                }
                Response::internal_error(&req, &format!("failed to handle request: {err} {req:?}"))
                    .to_vec()?
            }
        };

```

etorreborre:
The error management in `handle_request` is for the case where `encode_response` returned an `Err`:
```rust
pub(crate) fn encode_response<T: Encode<()>>(
    res: std::result::Result<Response<T>, Response<ockam_core::api::Error>>,
) -> Result<Vec<u8>> {
    let v = match res {
        Ok(r) => r.to_vec()?,
        Err(e) => e.to_vec()?,
    };

    Ok(v)
}
```
In `encode_response` we take care of serializing a `Response` whether it has a successful body, or if this is a  failure (`NotFound`, `BadRequest`, etc...). In these case we want the `Response` to keep the request id. 
That being said, there might be an opportunity for a further refactoring where we don't even try to pass the `RequestHeader` to the various `NodeManagerWorker` methods to build `Responses` but we let `encode_response` do it, right before serializing the `Response`.

Darioazzali:
Ok, thanks for the clarification 🙏.
I've added a prototype of how could be if the encode function has now the responsibility to add the request id to the response too.
1. The methods on `NodeWorkerManager`doesn't needs to have the request parameter, I've simply added the `success` constructor, although with the coexistence with the `ok` method (and with the various `.._no_response` would be probably better named something like `ok_no_request`).
```rust
    pub(super) async fn create_secure_channel(
        &mut self,
        create_secure_channel: CreateSecureChannel,
        ctx: &Context,
    ) -> Result<Response<CreateSecureChannelResponse>, Response<Error>> {
        let CreateSecureChannel {
            addr,
            authorized_identifiers,
            timeout,
            identity_name: identity,
            credential_name,
            ..
        } = create_secure_channel;

        let response = self
            .node_manager
            .create_secure_channel(
                ctx,
                addr,
                identity,
                authorized_identifiers,
                credential_name,
                timeout,
            )
            .await
            .map(|secure_channel| {
                Response::success().body(CreateSecureChannelResponse::new(secure_channel))
            })?;
        Ok(response)
    }
```
2. The `encode_response` now take the additional request parameter
```rust
pub(crate) fn encode_response<T: Encode<()>>(
    req: &RequestHeader,
    res: std::result::Result<Response<T>, Response<ockam_core::api::Error>>,
) -> Result<Vec<u8>> {
    let request_id = req.id();
    let v = match res {
        Ok(r) => r.re(request_id).to_vec()?,
        Err(e) => e.re(request_id).to_vec()?,
    };

    Ok(v)
}
```
Is this ok for you @etorreborre ?

etorreborre:
@Darioazzali yes I think that's the right thing to do. I would also consider getting rid of `Response::ok(req).body(...)` and replacing it directly with `Response::ok().body(...)` because I don't think that `Response::ok(req).body(...)` will be used at all given what you're doing.
But maybe this kind of cleanup can be done in a separate step.

Darioazzali:
I can do it at once while finishing the refactoring, so this will avoid ugly `success()` or `ok_no_response()`, is that okay for you @etorreborre?

etorreborre:
Yes, excellent.

etorreborre:
@Darioazzali note that there are [some restrictions on the commit messages](https://github.com/build-trust/ockam/actions/runs/7115068013/job/19372719168?pr=7006). Can you please rename them? (You can squash them as one if you want)

Darioazzali:
@etorreborre rebased right now, but I don't know why I have conflict to resolve (I've already pulled before pushing). As soon as are resolved I'll push again, sorry for that

etorreborre:
@Darioazzali I guess that's because I just merged a PR around those source files but I don't know exactly why there would be conflicts. Please ask if you find it difficult to fix the conflicts.

etorreborre:
Also don't forget to remove capital letters in your commit messages! (and there need to be a space after the `:` in `refactor(rust):slim`)

Darioazzali:
The tcp refactoring has been merged some second before me, resolved (the other) conflicts Re sending the test right now, as soon as it's finished I'll push again

etorreborre (approved):
That's great, your PR now closes this refactoring epic. Thank you!

Darioazzali:
Thanks to you for the help

PR #7008, opened 2023-12-01
feat(rust): execute an identity update in a transaction

etorreborre:
This PR fixes a gap where it would be possible in a concurrent context to store some invalid identity updates.

### DIFF
--- a/implementations/rust/ockam/ockam_identity/src/identities/storage/change_history_repository_sql.rs
+++ b/implementations/rust/ockam/ockam_identity/src/identities/storage/change_history_repository_sql.rs
@@ -1,5 +1,7 @@
 use core::str::FromStr;
 
+use sqlx::query::Query;
+use sqlx::sqlite::SqliteArguments;
 use sqlx::*;
 use tracing::debug;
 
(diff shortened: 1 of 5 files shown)


SanjoDeundiak (dismissed):
Cool

implementations/rust/ockam/ockam_identity/src/identities/storage/change_history_repository_sql.rs
```diff
@@ -35,15 +37,50 @@ impl ChangeHistorySqlxDatabase {
...
+            .bind(identity.identifier().to_sql());
+        let row: Option<ChangeHistoryRow> =
+            query1.fetch_optional(&mut *transaction).await.into_core()?;
+
+        let do_insert = match row {
+            Some(row) => {
+                let known_identity = Identity::import_from_change_history(
+                    Some(identity.identifier()),
+                    row.change_history()?,
+                    Vault::create_verifying_vault(),
```
SanjoDeundiak, line 52: Do you think it should be injected dependency? Not a huge difference and I don't think we're going to have a different VerifyingVault, but still

etorreborre: I wondered about that too but for now this would make things more complicated I think for no obvious gain. This is just a bunch of functions after all. They could almost be regular functions associated to `Signature`, something like
```
impl Signature {
   pub fn verify(&self, data: &[u8], verifying_public_key: &VerifyingPublicKey) -> Result<bool> {
       ...
   }
}
```

SanjoDeundiak (dismissed):

SanjoDeundiak (approved):

PR #7011, opened 2023-12-01
fix: fixed app reset and about dialogue size

davide-baldo:


implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
```diff
@@ -132,6 +132,12 @@ impl CliState {
             .await?;
 
         if let Some(pid) = node.pid() {
+            // avoid killing the current process, return successfully instead.
```
etorreborre, line 135: 🚀

etorreborre (approved):

PR #7018, opened 2023-12-04
fix(rust): filter out non-admin projects when retrieving them from orchestrator

adrianbenavides:
When using the [`Controller`'s `get_projects` function](https://github.com/build-trust/ockam/blob/0ca66af1ed508f65e127a500cf1b18819e91ae1f/implementations/rust/ockam/ockam_api/src/nodes/service/projects.rs#L88), we receive all the projects the user is part of (admin and non-admin) and they are stored in the state.

This can cause collisions when retrieving the app's projects, which are all named "default" currently.

The changes are based on the assumption that a user doesn’t get any benefit from listing non-admin projects, as they can’t do anything with them; and that the default user can only be the one created when running the `ockam enroll` command.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/projects.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/projects.rs
@@ -85,12 +85,17 @@ impl Projects for InMemoryNode {
             .await?)
     }
 
-    async fn get_projects(&self, ctx: &Context) -> miette::Result<Vec<Project>> {
+    async fn get_admin_projects(&self, ctx: &Context) -> miette::Result<Vec<Project>> {
         let projects = self.create_controller().await?.list_projects(ctx).await?;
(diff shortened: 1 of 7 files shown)


implementations/rust/ockam/ockam_api/src/nodes/service/projects.rs
```diff
@@ -87,10 +87,15 @@ impl Projects for InMemoryNode {
 
     async fn get_projects(&self, ctx: &Context) -> miette::Result<Vec<Project>> {
```
etorreborre: I suggest that we rename the function as well, so that we keep the possibility in the API to have a function returning all projects, including the ones we are just a member of.

adrianbenavides: done!

etorreborre (approved):

PR #7019, opened 2023-12-04
fix(rust): `reset` command do not fail when deleting orchestrator resources

adrianbenavides:
- When it fails to reach the orchestrator, prompt the user to continue instead of exiting with an error
- Retrieve available spaces from the controller instead of reading them from the local state, to avoid using an outdated state

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/reset.rs
+++ b/implementations/rust/ockam/ockam_command/src/reset.rs
@@ -1,13 +1,14 @@
 use clap::Args;
 use colorful::Colorful;
-use miette::miette;
+use miette::{miette, WrapErr};
+use ockam_api::cloud::space::Spaces;
 
 use ockam_api::nodes::InMemoryNode;
 use ockam_node::Context;
 
 use crate::terminal::ConfirmResult;
 use crate::util::node_rpc;
-use crate::{fmt_ok, CommandGlobalOpts};
+use crate::{color, fmt_ok, CommandGlobalOpts, OckamColor};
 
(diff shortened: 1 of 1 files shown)


implementations/rust/ockam/ockam_command/src/reset.rs
```diff
@@ -77,3 +59,45 @@ async fn run_impl(ctx: &Context, opts: CommandGlobalOpts, cmd: ResetCommand) ->
...
+        .as_ref()
+        .map(|s| s.set_message("Deleting spaces from the Orchestrator.."));
+    for space in opts.state.get_spaces().await? {
+        spinner.as_ref().map(|s| {
+            s.set_message(format!(
+                "Deleting space {}...",
+                color!(space.name, OckamColor::PrimaryResource)
+            ))
+        });
+        if controller.delete_space(ctx, &space.id).await.is_ok() {
```
etorreborre: @adrianbenavides there are functions at the `InMemoryNode` level [encapsulating the calls to the controller + cli state](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/cloud/space.rs#L110).

adrianbenavides: Yeah, I'm reviewing this part already, because I saw that we could change a few calls in this function. Thanks for the heads-up!

adrianbenavides: done!

etorreborre (dismissed):

davide-baldo (approved):

PR #7020, opened 2023-12-04
refactor(rust): remove re-entrant calls in repository implementations

etorreborre:
To avoid enforcing a pattern that can lead to deadlocks

I also removed the `put_attributes` function on the `AttributesRepository` because its semantic (adding attribute name/value pairs) was questionable (that function was only introduced to make support the credentials example)

### DIFF
--- a/implementations/rust/ockam/ockam_identity/src/identities/storage/attributes_entry.rs
+++ b/implementations/rust/ockam/ockam_identity/src/identities/storage/attributes_entry.rs
@@ -1,7 +1,9 @@
 use crate::models::{Identifier, TimestampInSeconds};
+use crate::utils::now;
 use minicbor::{Decode, Encode};
 use ockam_core::compat::borrow::ToOwned;
 use ockam_core::compat::{collections::BTreeMap, vec::Vec};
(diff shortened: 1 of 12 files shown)


implementations/rust/ockam/ockam_identity/src/identities/storage/attributes_entry.rs
```diff
@@ -56,3 +58,21 @@ impl AttributesEntry {
         self.attested_by.to_owned()
     }
 }
+
+impl AttributesEntry {
+    /// Create an AttributesEntry with just one name/value pair
+    pub fn singleton(
```
SanjoDeundiak: The name is a little bit misleading

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_identity/src/identities/storage/attributes_entry.rs
etorreborre: This is a common name in Haskell-land for this sort of thing. What do you suggest then?

SanjoDeundiak (approved):

PR #7032, opened 2023-12-04
feat(rust): delete `TrustContext`

SanjoDeundiak:
### DONE
1. Delete `TrustContext`. Replace with notion of Trusted Authority Identifier and CredentialRetriever
2. Remove CredentialsServer and related code
3. Use dedicated Members sqlite-backed storage for Authority node
4. Cache Credential in sqlite
5. Rework Credentials command
6. Migrations
7. Test migration
8. Test migration automatically
9. Support getting current time without std for ockam_node (migration storage) and ockam_identity (`now`)
10. Update Ockam App to use the new approach
11. Remove all mentions of trust context from the code
12. Update bats tests
13. Credential locking mechanism
14. Add online authority support outside of project
15. Improve authority node creation (checking if it's already running, fix logging)
16. Check enrollers can't create enrollers
17. Add additional identity service - identities verification


implementations/rust/ockam/ockam_abac/src/attribute_access_control.rs
```diff
@@ -54,21 +57,36 @@ impl AbacAccessControl {
...
             identity_attributes_repository,
+            authority,
             Policy::new(expression),
             Env::new(),
         )
     }
+
+    /// Create an AccessControl which will verify that the sender of
+    /// a message has an authenticated credential without checking any attributes
+    pub fn create_empty(
```
etorreborre: I'm not sure if `empty` really carries the meaning set in the comment. What could be a better name? `check_credential_only`?

implementations/rust/ockam/ockam_abac/src/attribute_access_control.rs
```diff
@@ -77,56 +95,67 @@ impl AbacAccessControl {
...
-                    Err(_) => {
+            Some(attrs) => {
+                for (key, value) in attrs.attrs() {
+                    let key = match from_utf8(key) {
+                        Ok(key) => key,
+                        Err(_) => {
+                            log::warn! {
+                                policy = %self.policy,
+                                id     = %id,
+                                "attribute key is not utf-8"
```
etorreborre, line 111: If only we had well-typed attributes :-)

implementations/rust/ockam/ockam_api/src/authenticator/common.rs
```diff
@@ -0,0 +1,91 @@
...
+use ockam_core::{Result, Routed};
+use ockam_node::Context;
+use std::collections::BTreeMap;
+use std::sync::Arc;
+
+/// Return a response on the return route stating that a secure channel is needed to access
+/// the service
+pub async fn secure_channel_required(c: &mut Context, m: Routed<Vec<u8>>) -> Result<()> {
+    // This was, actually, already checked by the access control. So if we reach this point
+    // it means there is a bug.  Also, if it' already checked, we should receive the Peer'
```
etorreborre: s/it'/it's

etorreborre: I think that the whole support for `secure_channel_required` should be reworked to:

 - Avoid duplication where it is used.
 - Avoid the double parsing or requests (by having better message types on the workers).

This can be done in another PR though.

implementations/rust/ockam/ockam_api/src/authenticator/common.rs
```diff
@@ -0,0 +1,91 @@
...
+impl EnrollerAccessControlChecks {
+    pub(crate) fn check_str_attributes_is_enroller(attributes: &BTreeMap<String, String>) -> bool {
+        attributes.get(OCKAM_ROLE_ATTRIBUTE_KEY)
+            == Some(&OCKAM_ROLE_ATTRIBUTE_ENROLLER_VALUE.to_string())
+    }
+
+    pub(crate) fn check_bin_attributes_is_enroller(
+        attributes: &BTreeMap<Vec<u8>, Vec<u8>>,
+    ) -> bool {
+        if let Some(val) = attributes.get(OCKAM_ROLE_ATTRIBUTE_KEY.as_bytes()) {
```
etorreborre, line 30: Arguably a one-liner? 
```rust
attributes
    .get(OCKAM_ROLE_ATTRIBUTE_KEY.as_bytes())
    .map_or(false, |v| {
        v == OCKAM_ROLE_ATTRIBUTE_ENROLLER_VALUE.as_bytes()
    })
```

implementations/rust/ockam/ockam_api/src/authenticator/common.rs
```diff
@@ -0,0 +1,91 @@
...
+            if val == OCKAM_ROLE_ATTRIBUTE_ENROLLER_VALUE.as_bytes() {
+                return true;
+            }
+        }
+
+        false
+    }
+
+    pub(crate) fn check_member_attributes(
```
etorreborre: The 2 functions `check_member_attributes` and `check_member` are only used locally and, IMO, `check_identifier` can be simplified to (with no need for auxiliary functions, and one calling the other):
```rust
    pub(crate) async fn check_identifier(
        members: Arc<dyn AuthorityMembersRepository>,
        identifier: &Identifier,
    ) -> Result<EnrollerCheckResult> {
        if let Some(member) = members.get_member(identifier).await? {
            Ok(EnrollerCheckResult {
                is_member: true,
                is_enroller: Self::check_bin_attributes_is_enroller(member.attributes()),
                is_pre_trusted: member.is_pre_trusted(),
            })
        } else {
            Ok(EnrollerCheckResult::default())
        }
    }
```

implementations/rust/ockam/ockam_api/src/authenticator/common.rs
SanjoDeundiak: Yeah, a lot of code was just moved, this is one of the examples. Improving all the pieces that I move would make me work on it for another month)

implementations/rust/ockam/ockam_api/src/authenticator/credential_issuer/credential_issuer.rs
```diff
@@ -0,0 +1,87 @@
...
+pub const TRUST_CONTEXT_ID: &[u8] = b"trust_context_id";
+
+/// Identifier for the schema of a project credential
+pub const PROJECT_MEMBER_SCHEMA: CredentialSchemaIdentifier = CredentialSchemaIdentifier(1);
+
+/// Maximum duration for a valid credential in seconds (30 days)
+pub const DEFAULT_CREDENTIAL_VALIDITY: Duration = Duration::from_secs(30 * 24 * 3600);
+
+/// This struct runs as a Worker to issue credentials based on a request/response protocol
+pub struct CredentialIssuer {
```
etorreborre, line 20: It would be good to add a comment explaining how each part of the `CredentialIssuer` is used (or comment directly on the attributes definitions)

implementations/rust/ockam/ockam_api/src/authenticator/credential_issuer/credential_issuer_worker.rs
```diff
@@ -0,0 +1,81 @@
...
+                credential_ttl,
+            ),
+        }
+    }
+}
+
+#[ockam_core::worker]
+impl Worker for CredentialIssuerWorker {
+    type Context = Context;
+    type Message = Vec<u8>;
```
etorreborre, line 44: I wrote earlier but we should stop introducing workers with untyped messages (unless I am missing a reason why this can not be the case for the RPC protocol?)

implementations/rust/ockam/ockam_api/src/authenticator/direct/client.rs
```diff
@@ -0,0 +1,82 @@
...
+        &self,
+        ctx: &Context,
+        identifier: Identifier,
+        attributes: BTreeMap<String, String>,
+    ) -> miette::Result<()> {
+        let req = Request::post("/").body(AddMember::new(identifier).with_attributes(attributes));
+        self.get_secure_client()
+            .tell(ctx, DefaultAddress::DIRECT_AUTHENTICATOR, req)
+            .await
+            .into_diagnostic()?
```
etorreborre, line 45: Maybe we should implement some syntactic sugar to stop this `.into_diagnostic().success().into_diagnostic()` dance?

implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator.rs
```diff
@@ -0,0 +1,155 @@
...
+};
+
+/// Identity attribute key that indicates the role of the subject
+pub const OCKAM_ROLE_ATTRIBUTE_KEY: &str = "ockam-role";
+
+/// Identity attribute value that indicates the enroller role of the subject
+/// the corresponding key is [`OCKAM_ROLE_ATTRIBUTE_KEY`]
+pub const OCKAM_ROLE_ATTRIBUTE_ENROLLER_VALUE: &str = "enroller";
+
+pub struct DirectAuthenticatorErrorString(pub String);
```
etorreborre: Why call it `XXXString` and not simply `XXX`?

implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator.rs
```diff
@@ -0,0 +1,155 @@
...
+        &self,
+        enroller: &Identifier,
+        id: &Identifier,
+        attrs: &BTreeMap<String, String>,
+    ) -> Result<Either<(), DirectAuthenticatorErrorString>> {
+        let check =
+            EnrollerAccessControlChecks::check_identifier(self.members.clone(), enroller).await?;
+
+        if !check.is_enroller {
+            warn!("Not enroller {} is trying to add a member {}", enroller, id);
```
etorreborre: The wording is a bit strange. I would go `{} is trying to add a member {}, but {} is not an enroller`.

implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator.rs
```diff
@@ -0,0 +1,155 @@
...
+                )));
+            }
+        }
+
+        let member = AuthorityMember::new(id.clone(), attrs, Some(enroller.clone()), now()?, false);
+
+        if let Err(err) = self.members.add_member(member).await {
+            warn!("Error adding member {} directly: {}", id, err);
+            return Ok(Either::Right(DirectAuthenticatorErrorString(
+                "Error adding member".to_string(),
```
etorreborre, line 77: Why not return the `err` to the caller?

implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator.rs
```diff
@@ -0,0 +1,155 @@
...
+    pub fn new(members: Arc<dyn AuthorityMembersRepository>) -> Self {
+        Self { members }
+    }
+
+    pub async fn add_member(
+        &self,
+        enroller: &Identifier,
+        id: &Identifier,
+        attrs: &BTreeMap<String, String>,
+    ) -> Result<Either<(), DirectAuthenticatorErrorString>> {
```
etorreborre: Maybe a type alias `type DirectAuthenticatorResult<a> = Either<a, DirectAuthenticatorError>` would make the function signatures slightly better?

implementations/rust/ockam/ockam_command/src/credential/list.rs
```diff
@@ -1,50 +1,64 @@
...
+    let node_name = match cmd.node_opts.at_node.clone() {
+        Some(name) => name,
+        None => opts.state.get_default_node().await?.name(),
+    };
+    opts.state.set_node_name(node_name.clone());
+
+    let database = opts.state.database();
+    let storage = CachedCredentialSqlxDatabase::new(database);
+
+    let credentials = storage.get_all().await.into_diagnostic()?;
```
adrianbenavides: @etorreborre ideally this should be accessed through the top-level functions of the CliState right to abstract the DB handling from the command? I'll note it to address it in a separate PR

implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    pub async fn delete_member(
+        &self,
+        enroller: &Identifier,
+        identifier: &Identifier,
+    ) -> Result<Either<(), DirectAuthenticatorErrorString>> {
+        let check_enroller =
+            EnrollerAccessControlChecks::check_identifier(self.members.clone(), enroller).await?;
+
+        if !check_enroller.is_enroller {
```
etorreborre, line 122: This code, up to L.126 could be refactored for the 3 methods.

implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator_worker.rs
```diff
@@ -0,0 +1,104 @@
...
+                        Response::ok().with_headers(&req).body(entries).to_vec()?
+                    }
+                    Either::Right(error) => Response::forbidden(&req, &error.0).to_vec()?,
+                }
+            }
+            (Some(Method::Delete), [id]) | (Some(Method::Delete), ["members", id]) => {
+                let identifier = Identifier::try_from(id.to_string())?;
+                let res = self.authenticator.delete_member(&from, &identifier).await?;
+
+                match res {
```
etorreborre, line 92: Can we refactor the `Left/Right` treatment on each path?

implementations/rust/ockam/ockam_api/src/authenticator/enrollment_tokens/acceptor.rs
```diff
@@ -1,108 +1,72 @@
...
+};
 
-pub struct EnrollmentTokenAcceptor(
-    pub(super) EnrollmentTokenAuthenticator,
-    pub(super) Arc<dyn IdentityAttributesRepository>,
-);
+pub struct EnrollmentTokenAcceptorErrorString(pub String);
+
+pub struct EnrollmentTokenAcceptor {
+    pub(super) tokens: Arc<dyn AuthorityEnrollmentTokenRepository>,
```
etorreborre, line 18: Nice to see that tokens are persisted now!

implementations/rust/ockam/ockam_api/src/authenticator/enrollment_tokens/acceptor_worker.rs
```diff
@@ -0,0 +1,71 @@
...
+impl Worker for EnrollmentTokenAcceptorWorker {
+    type Context = Context;
+    type Message = Vec<u8>;
+
+    async fn handle_message(&mut self, c: &mut Context, m: Routed<Self::Message>) -> Result<()> {
+        let secure_channel_info = match IdentitySecureChannelLocalInfo::find_info(m.local_message())
+        {
+            Ok(secure_channel_info) => secure_channel_info,
+            Err(_e) => {
+                secure_channel_required(c, m).await?;
```
etorreborre: One other thing about `secure_channel_required`. I am not a big fan that it is doing a silent reply to the sender.
I would be more comfortable with an overall structure that looks like
```rust
let res = with_secure_channel(m, handle_request(m));
c.send(m.return_route(), res).await
```

implementations/rust/ockam/ockam_api/src/authenticator/one_time_code.rs
```diff
@@ -82,6 +83,12 @@ impl<'de> Deserialize<'de> for OneTimeCode {
     }
 }
 
+impl ToSqlxType for OneTimeCode {
+    fn to_sql(&self) -> SqlxType {
+        SqlxType::Text(self.to_string())
```
etorreborre: This could be `self.to_string().to_sql()` (that's the pattern for other instances)

implementations/rust/ockam/ockam_api/src/authenticator/common.rs
SanjoDeundiak: Actually, I had a one-liner for `check_str_attributes_is_enroller`. But now I think that the long version is the only way to both avoid additional allocations during type conversion and make it simple

implementations/rust/ockam/ockam_api/src/authenticator/pre_trusted_identities.rs
```diff
@@ -0,0 +1,43 @@
...
+use serde_json as json;
+
+use ockam::identity::utils::now;
+use ockam::identity::{AttributesEntry, Identifier};
+use ockam_core::compat::{collections::HashMap, string::String};
+use ockam_core::errcode::{Kind, Origin};
+use ockam_core::Result;
+
+#[derive(Default, Clone, Debug, Serialize, Deserialize)]
+pub struct PreTrustedIdentities(pub HashMap<Identifier, AttributesEntry>);
```
etorreborre: `pub(crate) HashMap` should be enough

implementations/rust/ockam/ockam_api/src/authenticator/direct/client.rs
adrianbenavides: The easy way to do that would be to add miette's `Diagnostic` derive to our errors or implement some `From<miette::?>`, but we probably don't want to do that. We could also not use miette in the ockam_api crate... but sounds out of scope.

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_enrollment_token_repository.rs
```diff
@@ -0,0 +1,20 @@
+use crate::authenticator::one_time_code::OneTimeCode;
+use crate::authenticator::Token;
+use ockam::identity::TimestampInSeconds;
+use ockam_core::async_trait;
+use ockam_core::compat::boxed::Box;
+use ockam_core::Result;
+
+/// This repository stores enrollment tokens on the Authority node
+#[async_trait]
+pub trait AuthorityEnrollmentTokenRepository: Send + Sync + 'static {
```
etorreborre, line 10: I would remove `Authority` in the name (same thing for the table name).

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_enrollment_token_repository_sql.rs
```diff
@@ -0,0 +1,234 @@
...
+}
+
+#[async_trait]
+impl AuthorityEnrollmentTokenRepository for AuthorityEnrollmentTokenSqlxDatabase {
+    async fn use_token(
+        &self,
+        one_time_code: OneTimeCode,
+        now: TimestampInSeconds,
+    ) -> Result<Option<Token>> {
+        let query1 =
```
etorreborre, line 45: It would be good to add a comment to explain why we are doing this (purge the database of expired tokens)

implementations/rust/ockam/ockam_api/src/authenticator/credential_issuer/credential_issuer_worker.rs
SanjoDeundiak: I didn't introduce it, I moved it. But I think the main problem here is that we use CBOR for RPC but BARE for the messages. BTW, we could totally switch messages inside a node to use CBOR without breaking any backwards compatibility

implementations/rust/ockam/ockam_api/src/authenticator/direct/client.rs
SanjoDeundiak: I think we have all the diversity of different error types. Good problem to be solved

implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator.rs
SanjoDeundiak: I think we should avoid returning bare errors to the user in case of RPC. It's hard to guarantee that they are meaningful and not leak too much detail.

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_member.rs
```diff
@@ -0,0 +1,79 @@
+use ockam::identity::{Identifier, TimestampInSeconds};
+use ockam_core::compat::collections::BTreeMap;
+use ockam_core::compat::str::FromStr;
+use ockam_core::{Error, Result};
+
+/// Project member stored on the Authority node
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AuthorityMember {
```
etorreborre, line 8: Still wondering about the naming. The name is `AuthorityMember` but all the comments talk about a "project member". I would call simply call this entity a `Member` (which kind of begs the question "a member of what?". The what is not really encoded here, maybe only in `added_by`).

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_member.rs
```diff
@@ -0,0 +1,79 @@
...
+    identifier: Identifier,
+    attributes: BTreeMap<Vec<u8>, Vec<u8>>,
+    added_by: Option<Identifier>,
```
etorreborre: Do we have a use case for `added_by: None`? It doesn't seem so from the code so I would restrict that option for now.

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_member.rs
```diff
@@ -0,0 +1,79 @@
...
+    added_at: TimestampInSeconds,
+    // Was provided by TrustedIdentities argument during the Authority startup
+    // pre-trusted identities can't be deleted using [`MembersStorage::delete_member()`]
+    is_pre_trusted: bool,
```
etorreborre, line 15: At this level, I would call this flag something like `permanent` as a way to indicate that it cannot be deleted.

implementations/rust/ockam/ockam_api/src/authenticator/pre_trusted_identities.rs
SanjoDeundiak: Refactored a little bit

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_members_repository_sql.rs
```diff
@@ -0,0 +1,241 @@
...
+            .bind(member.identifier().to_sql())
+            .bind(member.added_by().clone().map(|x| x.to_sql()))
+            .bind(member.added_at().to_sql())
+            .bind(member.is_pre_trusted().to_sql())
+            .bind(minicbor::to_vec(member.attributes())?.to_sql());
+
+        query.execute(&*self.database.pool).await.void()
+    }
+
+    async fn bootstrap_pre_trusted_members(
```
etorreborre, line 71: I don't see a good reason for not having this loop outside the database code, inside `authority.rs`.
This way we separate the general notion of "member", from the fact that there are identities that we pre-trust.

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_token.rs
```diff
@@ -0,0 +1,43 @@
+use crate::authenticator::one_time_code::OneTimeCode;
```
etorreborre: Can we call the file `token.rs`, like the data type it contains?

implementations/rust/ockam/ockam_api/src/cli_state/repositories.rs
```diff
@@ -7,85 +7,65 @@ use ockam_abac::{PoliciesRepository, PolicySqlxDatabase};
...
 
-    pub(super) async fn purpose_keys_repository(&self) -> Result<Arc<dyn PurposeKeysRepository>> {
-        Ok(Arc::new(PurposeKeysSqlxDatabase::new(self.database())))
+    pub(super) fn purpose_keys_repository(&self) -> Arc<dyn PurposeKeysRepository> {
+        Arc::new(PurposeKeysSqlxDatabase::new(self.database()))
     }
 
-    pub(super) async fn secrets_repository(&self) -> Result<Arc<dyn SecretsRepository>> {
-        Ok(Arc::new(SecretsSqlxDatabase::new(self.database())))
+    pub(super) fn secrets_repository(&self) -> Arc<dyn SecretsRepository> {
```
etorreborre, line 36: I don't see why I did this in the first place. Good call.

implementations/rust/ockam/ockam_api/src/cli_state/trust.rs
```diff
@@ -0,0 +1,132 @@
...
+use ockam_core::errcode::{Kind, Origin};
+use ockam_core::{Error, Result};
+use ockam_multiaddr::MultiAddr;
+use ockam_vault::SoftwareVaultForVerifyingSignatures;
+
+impl CliState {
+    pub async fn retrieve_trust_options(
+        &self,
+        project_name: &Option<String>,
+        authority_identity: &Option<String>,
```
etorreborre, line 17: I would be nicer to have `TrustOpts` return an `Option<AuthorityOpts>` from the beginning where
```rust
struct AuthorityOpts {
  identity: String,
  route: MultiAddr,
}
```
and put the fact that identity / route must be defined together inside the `clap` validation.

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_member.rs
SanjoDeundiak: Pre trusted is what it's used to be called. Not saying it's perfect, but permanent also doesn't reflect how it's working - it's overwritten on every start of the Authority node with fresh values

implementations/rust/ockam/ockam_api/src/cli_state/trust.rs
```diff
@@ -0,0 +1,132 @@
...
+                    Origin::Api,
+                    Kind::NotFound,
+                    "Invalid authority identity hex",
+                )
+            })?;
+            let authority_identifier = identities_verification
+                .import(None, &authority_identity)
+                .await?;
+
+            let trust_options = if let Some(authority_multiaddr) = authority_multiaddr {
```
etorreborre: Those are much nicer `TrustOptions`!

implementations/rust/ockam/ockam_api/src/cli_state/trust.rs
```diff
@@ -0,0 +1,132 @@
+use crate::nodes::service::{NodeManagerCredentialRetrieverOptions, NodeManagerTrustOptions};
+use crate::{multiaddr_to_transport_route, CliState, DefaultAddress};
+use ockam::identity::{IdentitiesVerification, RemoteCredentialRetrieverInfo};
+use ockam_core::errcode::{Kind, Origin};
+use ockam_core::{Error, Result};
+use ockam_multiaddr::MultiAddr;
+use ockam_vault::SoftwareVaultForVerifyingSignatures;
+
+impl CliState {
+    pub async fn retrieve_trust_options(
```
etorreborre, line 14: A quick comment summarizing the code and the various options would be good.

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_members_repository_sql.rs
SanjoDeundiak: Or maybe we should go the other way round? Make sure that pretrused members can only be added here? Make `add_member` reject such members. Thoughts?

implementations/rust/ockam/ockam_api/src/cloud/secure_clients.rs
```diff
@@ -220,44 +219,30 @@ impl NodeManager {
...
+        multiaddr_to_transport_route(&multiaddr).ok_or_else(|| {
+            ApiError::core(format!(
+                "Couldn't convert MultiAddr to route: multiaddr={multiaddr}"
+            ))
+        })
     }
 }
 
 pub struct AuthorityNodeClient {
-    pub(crate) secure_client: SecureClient,
```
etorreborre, line 244: Nice.

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_token.rs
SanjoDeundiak: I'll rename the file and the struct to EnrollmentToken actually. Token is too broad

implementations/rust/ockam/ockam_api/src/nodes/service/policy.rs
```diff
@@ -150,7 +140,7 @@ impl Policies for BackgroundNodeClient {
         }
 
         let policy = {
-            let expr = eq([ident("subject.trust_context_id"), str(project_id)]);
+            let expr = t();
```
etorreborre: Can we have a better name than `t`?

implementations/rust/ockam/ockam_api/src/util.rs
```diff
@@ -423,19 +426,16 @@ pub mod test_utils {
...
-
         let vault = cli_state
             .get_or_create_default_named_vault()
             .await?
             .vault()
             .await?;
         let identities = cli_state.make_identities(vault).await?;
 
+        let attributes = AttributesBuilder::with_schema(PROJECT_MEMBER_SCHEMA)
+            .with_attribute("role".as_bytes(), "user".as_bytes())
```
etorreborre: I would rather have attributes passed as parameters to `start_manager_for_tests` so that it becomes a bit more obvious how those attributes relate to a specific test being executed.
I also don't currently see where this attribute is used in the tests.

implementations/rust/ockam/ockam_api/tests/direct_authenticator.rs
```diff
@@ -0,0 +1,590 @@
...
+use ockam_api::authenticator::direct::{
+    OCKAM_ROLE_ATTRIBUTE_ENROLLER_VALUE, OCKAM_ROLE_ATTRIBUTE_KEY,
+};
+use ockam_core::Result;
+use ockam_node::Context;
+use std::collections::BTreeMap;
+
+mod common;
+
+#[ockam_macros::test]
```
etorreborre, line 14: Nice tests!

implementations/rust/ockam/ockam_command/src/authority/create.rs
```diff
@@ -33,30 +32,42 @@ long_about = docs::about(LONG_ABOUT),
...
-    /// `authority create` started a child process to run this node in foreground.
-    #[arg(long, hide = true)]
-    pub child_process: bool,
+    /// Name of the Identity that the authority will use
+    #[arg(long = "identity", value_name = "IDENTITY_NAME")]
+    identity: Option<String>,
+
+    /// Identifier of the project associated to this authority node on the Orchestrator
+    #[arg(long, value_name = "PROJECT_IDENTIFIER")]
+    project_identifier: String,
```
etorreborre, line 71: We should be able to parse an `Identifier` directly here.

implementations/rust/ockam/ockam_command/src/authority/create.rs
```diff
@@ -68,16 +79,6 @@ pub struct CreateCommand {
...
 
-    /// List of the trusted identities, and corresponding attributes to be preload in the attributes storage.
-    /// Format: {"identifier1": {"attribute1": "value1", "attribute2": "value12"}, ...}
-    #[arg(group = "trusted", long, value_name = "JSON_OBJECT", value_parser = parse_trusted_identities)]
-    trusted_identities: Option<TrustedIdentities>,
-
-    /// Path of a file containing trusted identities and their attributes encoded as a JSON object.
-    /// Format: {"identifier1": {"attribute1": "value1", "attribute2": "value12"}, ...}
-    #[arg(group = "trusted", long, value_name = "PATH")]
-    reload_from_trusted_identities_file: Option<PathBuf>,
```
etorreborre, line 79: The help text for the command needs to be changed as well.

implementations/rust/ockam/ockam_command/src/credential/list.rs
etorreborre: @adrianbenavides yes indeed

implementations/rust/ockam/ockam_command/src/credential/store.rs
```diff
@@ -51,41 +43,71 @@ async fn run_impl(
...
+        .await?
+        .name();
     let mut state = opts.state.clone();
     state.set_node_name(node_name);
 
-    opts.terminal.write_line(&fmt_log!(
-        "Storing credential {}...\n",
-        cmd.credential_name.clone()
-    ))?;
+    let database = state.database();
```
etorreborre, line 54: @adrianbenavides same thing here.

implementations/rust/ockam/ockam_command/src/credential/verify.rs
```diff
@@ -96,47 +99,48 @@ pub async fn verify_credential(
...
-            .vault()
-            .await?;
-        let identities = match opts.state.make_identities(vault).await {
-            Ok(i) => i,
-            Err(e) => {
-                *is_finished.lock().await = true;
-                return Err(e)?;
-            }
-        };
+        let change_history_repository = ChangeHistorySqlxDatabase::new(state.database());
```
etorreborre, line 102: @adrianbenavides there too.

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_member.rs
SanjoDeundiak: There was, but good poing. I made it non optional

implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator.rs
SanjoDeundiak: I don't think there is much to refactor here, it's very simple

implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator_worker.rs
SanjoDeundiak: That would be pretty tricky given it has optional body and sometimes you need to do extra processing on it as in `member_ids`

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_enrollment_token_repository.rs
SanjoDeundiak: I think the problem is that we create the same set of tables on each machine, therefore, I wanted to emphasize that this table is only to be used for the Authority node.

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_member.rs
SanjoDeundiak: I think that's because the only Authority we now have is the Project Authority. Let's make it clearer when new ones appear?

implementations/rust/ockam/ockam_api/src/authenticator/enrollment_tokens/acceptor_worker.rs
SanjoDeundiak: Makes sense to just remove it altogether

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_member.rs
etorreborre: Hmm, ok. I think that the thing that annoys me a bit is the presence of "business logic" inside the storing logic. I'd rather see this rule one level up.

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_members_repository_sql.rs
etorreborre: I would leave only storage concerns at this level, and put all the member logic one level up.

implementations/rust/ockam/ockam_command/src/node/util.rs
```diff
@@ -109,38 +105,19 @@ pub async fn spawn_node(
...
+    if let Some(authority_identity) = authority_identity {
+        args.push("--authority-identity".to_string());
+        args.push(authority_identity);
     }
 
-    if let Some(project_name) = project_name {
-        args.push("--project".to_string());
-        args.push(project_name.to_string());
+    if let Some(authority_multiaddr) = authority_multiaddr {
+        args.push("--authority-multiaddr".to_string());
```
etorreborre: This argument is generally called `authority_route` elsewhere

implementations/rust/ockam/ockam_command/src/policy/mod.rs
```diff
@@ -74,12 +74,11 @@ pub(crate) async fn add_default_project_policy(
     node_name: &str,
     ctx: &Context,
     opts: &CommandGlobalOpts,
-    project_id: String,
     resource: &Resource,
 ) -> miette::Result<()> {
     let node = BackgroundNodeClient::create_to_node(ctx, &opts.state, node_name).await?;
 
-    let expr = eq([ident("subject.trust_context_id"), str(project_id)]);
+    let expr = t();
```
etorreborre: A constant like `Expr::TRUE` would be nicer IMO.

implementations/rust/ockam/ockam_command/src/project/ticket.rs
```diff
@@ -65,16 +66,16 @@ impl TicketCommand {
...
             let mut parts = attr.splitn(2, '=');
             let key = parts.next().ok_or(miette!("key expected"))?;
             let value = parts.next().ok_or(miette!("value expected)"))?;
-            attributes.insert(key, value);
+            attributes.insert(key.to_string(), value.to_string());
         }
-        if let Some(relay_name) = &self.allowed_relay_name {
-            attributes.insert("ockam-relay", relay_name);
+        if let Some(relay_name) = self.allowed_relay_name.clone() {
+            attributes.insert("ockam-relay".to_string(), relay_name);
```
etorreborre: This should be a constant with a comment explaining how this is used on the Controller side.

implementations/rust/ockam/ockam_command/src/project/util.rs
```diff
@@ -177,7 +176,7 @@ async fn check_project_node_accessible(
         .as_ref()
         .ok_or(miette!("Project identity is not set."))?;
     let project_node = node
-        .create_project_client(project_identity, &project_route, None)
+        .create_project_client(false, project_identity, &project_route, None)
```
etorreborre: This call site makes me think that an enum `CredentialsEnabled::Off` would be more descriptive (and maybe put this argument at the end of the list)

implementations/rust/ockam/ockam_command/tests/bats/credentials.bats
```diff
@@ -29,20 +27,14 @@ teardown() {
...
+  run_success "$OCKAM" credential store --issuer "$idt1_short" --credential-path "$OCKAM_HOME/credential"
 
   run_success "$OCKAM" credential list
-  assert_output --partial "Credential: smart_nyc_cred"
-  assert_output --partial "{\"application\": \"Smart Factory\", \"city\": \"New York\""
+  assert_output --partial "{\"application\":\"Smart Factory\",\"city\":\"New York\""
 }
 
-@test "credential - verify and store reject invalid credentials" {
+@test "credential - verify reject invalid credentials" {
```
etorreborre: s/reject/rejects

implementations/rust/ockam/ockam_command/tests/bats/trust.bats
```diff
@@ -0,0 +1,118 @@
...
+  $OCKAM credential issue --as authority --for "$alice_identifier" --attribute city="New York" --encoding hex >"$OCKAM_HOME/alice.cred"
+  run_success "$OCKAM" credential store --at alice --issuer "$authority_identifier" --credential-path "$OCKAM_HOME/alice.cred"
+
+  # issue and store credential for bob
+  $OCKAM credential issue --as authority --for "$bob_identifier" --attribute city="New York" --encoding hex >"$OCKAM_HOME/bob.cred"
+  run_success "$OCKAM" credential store --at bob --issuer "$authority_identifier" --credential-path "$OCKAM_HOME/bob.cred"
+
+  msg=$(random_str)
+
+  # Create a node for alice that trust authority as a credential authority
```
etorreborre: s/that trust/that trusts

implementations/rust/ockam/ockam_core/src/compat.rs
```diff
@@ -321,12 +321,38 @@ pub mod vec {
 #[cfg(feature = "std")]
 pub mod time {
     pub use std::time::*;
+
+    /// Create a new timestamp using the system time
+    pub fn now() -> crate::Result<u64> {
```
etorreborre, line 326: Maybe it's time (no pun intended :-)) to put the `TimestampInSeconds` data type from `ockam_identity` in that crate.

implementations/rust/ockam/ockam_identity/src/credentials/credentials_verification.rs
```diff
@@ -44,16 +44,32 @@ impl CredentialsVerification {
...
+            self.verifying_vault.clone(),
+            expected_subject,
+            authorities,
+            credential_and_purpose_key,
+        )
+        .await
+    }
+
+    /// Verify a [`Credential`]
+    pub async fn verify_credential_static(
```
etorreborre, line 64: I think that this function should not be necessary if the code in `ockam_command/.../verify.rs` was just calling an appropriate interface on `CliState`. But we can refactor that in another PR

implementations/rust/ockam/ockam_identity/src/credentials/retriever/cache_retriever.rs
```diff
@@ -0,0 +1,63 @@
...
+use crate::{CachedCredentialsRepository, CredentialRetriever, Identifier, TimestampInSeconds};
+use async_trait::async_trait;
+use ockam_core::compat::boxed::Box;
+use ockam_core::compat::sync::Arc;
+use ockam_core::Result;
+use ockam_node::Context;
+use tracing::{debug, error};
+
+/// Credentials retriever for credentials located in a local cache
+pub struct CacheCredentialRetriever {
```
etorreborre: s/CacheCredentialRetriever/CachedCredentialRetriever

implementations/rust/ockam/ockam_identity/src/credentials/retriever/remote_retriever.rs
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        Ok(credential)
+    }
+
+    async fn check_for_cached_credential(
+        &self,
+        for_identity: &Identifier,
+    ) -> Result<Option<CredentialAndPurposeKey>> {
+        let cache = self
```
etorreborre: This code is duplicated with the `CachedCredentialRetriever`. We could reuse it here.

implementations/rust/ockam/ockam_identity/src/credentials/retriever/remote_retriever.rs
```diff
@@ -0,0 +1,209 @@
...
+            error!(
+                "Error caching credential for {} from {}. Err={}",
+                for_identity, &self.info.issuer, err
+            );
+        }
+
+        Ok(credential)
+    }
+
+    async fn check_for_cached_credential(
```
etorreborre: `get_valid_cached_credential` would be a better name I think here.

implementations/rust/ockam/ockam_identity/src/credentials/retriever/remote_retriever.rs
```diff
@@ -0,0 +1,209 @@
...
+        let cached_credential = self.check_for_cached_credential(for_identity).await?;
+        debug!(
+            "Rechecking cached credential for: {} from: {}",
+            for_identity, self.info.issuer
+        );
+        if let Some(cached_credential) = cached_credential {
+            return Ok(Some(cached_credential));
+        }
+
+        *lock = true;
```
etorreborre: I don't understand what is the point of setting the lock to values that are not read anyway?

implementations/rust/ockam/ockam_identity/src/credentials/retriever/remote_retriever.rs
```diff
@@ -0,0 +1,209 @@
...
+        if let Some(cached_credential) = cached_credential {
+            return Ok(Some(cached_credential));
+        }
+
+        let mut lock = self.retrieving_credential.lock().await;
+
+        // Recheck in case other thread retrieved a credential
+        let cached_credential = self.check_for_cached_credential(for_identity).await?;
+        debug!(
+            "Rechecking cached credential for: {} from: {}",
```
etorreborre: The message is not very clear here. The `debug` call should be one line above and say `Trying again to get a cached credential for`.

implementations/rust/ockam/ockam_identity/src/identities/storage/cached_credentials_repository.rs
```diff
@@ -0,0 +1,28 @@
+use crate::models::CredentialAndPurposeKey;
+use crate::{Identifier, TimestampInSeconds};
+use async_trait::async_trait;
+use ockam_core::compat::boxed::Box;
+use ockam_core::Result;
+
+/// This trait supports the persistence of cached credentials
+#[async_trait]
+pub trait CachedCredentialsRepository: Send + Sync + 'static {
```
etorreborre: I think I would simply called that one `CredentialsRepository`. The fact that we are using it as a cache is another concern.

implementations/rust/ockam/ockam_identity/src/identities/storage/cached_credentials_repository_sql.rs
```diff
@@ -0,0 +1,189 @@
...
+use crate::{CachedCredentialsRepository, TimestampInSeconds};
+
+/// Implementation of `CachedCredentialRepository` trait based on an underlying database
+/// using sqlx as its API, and Sqlite as its driver
+#[derive(Clone)]
+pub struct CachedCredentialSqlxDatabase {
+    database: SqlxDatabase,
+}
+
+impl CachedCredentialSqlxDatabase {
```
etorreborre: Regardless of `Cached` or not (see my previous comment), the name here should be `CredentialsSqlxDatabase` to match `CredentialsRepository`

implementations/rust/ockam/ockam_identity/src/secure_channel/options.rs
```diff
@@ -63,21 +68,30 @@ impl SecureChannelOptions {
...
+        Ok(self)
     }
 
-    /// Adds a single credential
-    pub fn with_credential(mut self, credential: CredentialAndPurposeKey) -> Self {
-        self.credentials.push(credential);
-        self
+    /// Set credential
+    pub fn with_credential(mut self, credential: CredentialAndPurposeKey) -> Result<Self> {
+        if self.credential_retriever.is_some() {
```
etorreborre: This could be directly:
```rust
self.with_credential_retriever(Arc::new(MemoryCredentialRetriever::new(credential)))
```

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database/migration_support.rs
```diff
@@ -0,0 +1,35 @@
...
+        if let Some(count_raw) = count_raw {
+            let count: i64 = count_raw.get(0);
+            Ok(count != 0)
+        } else {
+            Ok(false)
+        }
+    }
+
+    pub(crate) async fn mark_as_migrated(pool: &SqlitePool, migration_name: &str) -> Result<()> {
+        let now = now()?;
```
etorreborre, line 24: If we have `now()` returning a `TimestampInSeconds` we can store it directly because it already has a `ToSqlxType` instance.

etorreborre (dismissed):
This is quite the PR! I think that there are opportunities for refactoring / renaming but this can be done as subsequent PRs

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_enrollment_token_repository.rs
etorreborre: This is why there is the `SqlxDatabase::create_with_migration` function. To be able to start an instance of a database with a different set of tables. If we made it a list of `Migration`s instead of an `Option<Migration>`, we could share some table definitions between a normal node and an authority node + have some authority-specific tables.

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_member.rs
etorreborre: That's ok.

implementations/rust/ockam/ockam_api/src/authenticator/storage/authority_member.rs
SanjoDeundiak: I don't think there is a right answer here. Modern sql dbs have a lot of things to support business logic, like stored functions and triggers. NoSQL seems to avoid that

implementations/rust/ockam/ockam_api/src/cli_state/trust.rs
SanjoDeundiak: There is  a use case that we might want to support when you know the Authority Identity, so you can verify credentials, but you don't need to obtain a credential because either you're trusted identifier in the system or you just use manually stored credential in the cache. In this case identity would be Some but route will be None

implementations/rust/ockam/ockam_command/src/authority/create.rs
SanjoDeundiak: That's actually not an Identity Identifier.

implementations/rust/ockam/ockam_api/src/util.rs
SanjoDeundiak: It's not used, I can remove it

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database/migration_support.rs
SanjoDeundiak: The problem is that `TimestampInSeconds` is in `ockam_identity` crate, and I'm not sure we should move it

implementations/rust/ockam/ockam_identity/src/secure_channel/options.rs
SanjoDeundiak: Sure, but it's less convenient IMO. Also requires you to know about `MemoryCredentialRetriever` existence

implementations/rust/ockam/ockam_identity/src/credentials/retriever/remote_retriever.rs
SanjoDeundiak: Good point. Now that we know that `_lock` won't be dropped :)

implementations/rust/ockam/ockam_identity/src/identities/storage/cached_credentials_repository.rs
SanjoDeundiak: Good point, `CredentialRepository` then I guess? (singular)

implementations/rust/ockam/ockam_core/src/compat.rs
SanjoDeundiak: I'm not sure it's a great idea. We intentionally kept everything identity-related in one place. Maybe a new type can be created and then converted to `TimestampInSeconds` inside `ockam_identity`

implementations/rust/ockam/ockam_identity/src/secure_channel/options.rs
etorreborre: I meant "as the implementation of the `with_credential`" method.

SanjoDeundiak: Oh, sure

implementations/rust/ockam/ockam_core/src/compat.rs
etorreborre: `TimestampInSeconds` is not identity-related in my mind. I don't why that type shouldn't go into `ockam_core`.

etorreborre (approved):

PR #7037, opened 2023-12-05
feat(rust): keep the same port for incoming services

davide-baldo:
[ID_48] Don’t change the port number for a service shared with me (remember the first assigned port number)


adrianbenavides (commented):
Looks good but we should also add a fallback mechanism when re-attaching the port to a service, in case the port is not available anymore. Or do we have that already?

davide-baldo:
> Looks good but we should also add a fallback mechanism when re-attaching the port to a service, in case the port is not available anymore. Or do we have that already?

I don't think there is a way to discriminate between a connection failure and a port already being used

adrianbenavides (approved):

PR #7040, opened 2023-12-05
refactor(rust): slim down the `NodeManagerWorker` for `node / tcp`

rbran:
This is my try at solving https://github.com/build-trust/ockam/issues/6708

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -563,10 +563,10 @@ impl NodeManagerWorker {
                 encode_response(self.get_tcp_connection(req, address.to_string()).await)?
             }
             (Post, ["node", "tcp", "connection"]) => {
-                encode_response(self.create_tcp_connection(req, dec, ctx).await)?
+                encode_response(self.create_tcp_connection(ctx, req, dec.decode()?).await)?
             }
             (Delete, ["node", "tcp", "connection"]) => {
-                encode_response(self.delete_tcp_connection(req, dec).await)?
+                encode_response(self.delete_tcp_connection(req, dec.decode()?).await)?
             }
 
             // ==*== Tcp Listeners ==*==
@@ -575,10 +575,10 @@ impl NodeManagerWorker {
                 encode_response(self.get_tcp_listener(req, address.to_string()).await)?
             }
             (Post, ["node", "tcp", "listener"]) => {
-                encode_response(self.create_tcp_listener(req, dec).await)?
+                encode_response(self.create_tcp_listener(req, dec.decode()?).await)?
             }
             (Delete, ["node", "tcp", "listener"]) => {
-                encode_response(self.delete_tcp_listener(req, dec).await)?
+                encode_response(self.delete_tcp_listener(req, dec.decode()?).await)?
             }
 
             // ==*== Credential ==*==
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
@@ -1,97 +1,112 @@
 use std::net::SocketAddr;
 
-use minicbor::Decoder;
-
 use ockam::Result;
 use ockam_core::api::{Error, RequestHeader, Response};
-use ockam_core::Address;
 use ockam_node::Context;
-use ockam_transport_tcp::{
-    TcpConnectionOptions, TcpListenerInfo, TcpListenerOptions, TcpSenderInfo, TcpTransport,
-};
+use ockam_transport_tcp::{TcpConnectionOptions, TcpListenerOptions};
 
+use super::{NodeManager, NodeManagerWorker};
 use crate::nodes::models::transport::{
-    CreateTcpConnection, CreateTcpListener, DeleteTransport, TransportList, TransportMode,
-    TransportStatus, TransportType,
+    CreateTcpConnection, CreateTcpListener, DeleteTransport, TransportList, TransportStatus,
 };
-use crate::nodes::service::ApiTransport;
 
-use super::NodeManagerWorker;
-
-impl NodeManagerWorker {
-    fn find_connection(tcp: &TcpTransport, address: String) -> Option<TcpSenderInfo> {
-        match address.parse::<SocketAddr>() {
-            Ok(socket_address) => tcp
+impl NodeManager {
+    fn get_tcp_connections(&self) -> TransportList {
+        TransportList::new(
+            self.tcp_transport
                 .registry()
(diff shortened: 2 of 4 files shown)


implementations/rust/ockam/ockam_api/src/nodes/models/transport/request.rs
```diff
@@ -4,13 +4,13 @@ use minicbor::{Decode, Encode};
 #[derive(Debug, Clone, Decode, Encode, PartialEq, Eq)]
 #[rustfmt::skip]
 #[cbor(map)]
-pub struct CreateTcpConnection {
+pub struct CreateTcpConnection<'a> {
```
etorreborre: Can you please remove the lifetime parameter? We have actually been removing lifetime parameters on other requests (but not all yet).

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
```diff
@@ -16,253 +16,265 @@ use crate::nodes::models::transport::{
 };
 use crate::nodes::service::ApiTransport;
 
-use super::NodeManagerWorker;
+use super::{NodeManager, NodeManagerWorker};
+
+fn find_connection_socketaddr(
```
etorreborre: Those functions (`find_connection_socketaddr`, `find_connection`, `find_listener_socketaddress`) are indeed very useful. Could you please make them methods on `TcpTransport`?
Also I suggest a renaming of `find_connection_socketaddr` to `find_connection_by_socketaddr` which, I think, is less misleading (same thing for `find_listener_socketaddress`).

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
```diff
@@ -16,253 +16,265 @@ use crate::nodes::models::transport::{
...
-            TransportStatus::new(ApiTransport {
-                tt: TransportType::Tcp,
-                tm: (*info.mode()).into(),
-                socket_address: info.socket_address(),
-                worker_address: info.address().to_string(),
-                processor_address: info.receiver_address().to_string(),
-                flow_control_id: info.flow_control_id().clone(),
-            })
-        };
+fn tcp_sender_to_status(sender: &TcpSenderInfo) -> TransportStatus {
```
etorreborre: Could you please put these 2 functions in the `nodes/models/transports/response.rs` file, near `TransportStatus`?

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -563,7 +563,7 @@ impl NodeManagerWorker {
             // ==*== Tcp Connection ==*==
             (Get, ["node", "tcp", "connection"]) => self.get_tcp_connections(req).await.to_vec()?,
             (Get, ["node", "tcp", "connection", address]) => {
-                encode_response(self.get_tcp_connection(req, address.to_string()).await)?
+                encode_response(self.get_tcp_connection(req, address).await)?
             }
             (Post, ["node", "tcp", "connection"]) => {
                 encode_response(self.create_tcp_connection(req, dec, ctx).await)?
```
etorreborre: Is it possible to do the decoding of `dec` in there directly? We're trying to do the same in other places like [here](https://github.com/build-trust/ockam/pull/7040/files#diff-fee061b9c2629c91dcb8dcd8412526980626c18ba53562340f2040cab470469aR680).

etorreborre (commented):
Thanks for your contribution! I have made some comments to suggest refactorings and make things more consistent with the rest.

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
rbran: Should I implement `impl From<TcpSenderInfo> for TransportStatus` and `impl From<TcpListenerInfo> for TransportStatus` instead? Should I also replace the `TransportStatus::new` function body with `impl From<ApiTransport> for TransportStatus`?

etorreborre: That's a good idea, yes.

implementations/rust/ockam/ockam_api/src/nodes/models/transport/request.rs
rbran: reverted at https://github.com/build-trust/ockam/pull/7040/commits/0c34d523ee16ceb5326d5c78e92487d4e4847dc9

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -566,10 +566,10 @@ impl NodeManagerWorker {
                 encode_response(self.get_tcp_connection(req, address.to_string()).await)?
             }
             (Post, ["node", "tcp", "connection"]) => {
-                encode_response(self.create_tcp_connection(req, dec, ctx).await)?
+                encode_response(self.create_tcp_connection(req, dec.decode()?, ctx).await)?
```
etorreborre: I forgot, could you please put the arguments in the same order as in most other places: `ctx, req, dec.decode()?`. Thanks!

rbran: done

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
rbran: I made those function methods of `TcpTransport`, I didn't know that is the best file to put the impl block, so I added it to [implementations/rust/ockam/ockam_transport_tcp/src/transport/lifecycle.rs](https://github.com/build-trust/ockam/pull/7040/files#diff-cdb18b7b2dcb1672aa278a7ded04661ae51b740a7c25fe965c09b574e459518c)

implementations/rust/ockam/ockam_api/src/nodes/service/transport.rs
rbran: now the map function is `TransportStatus::from` and a new `TransportStatus` can be created simply using `into` any type that implement the `From` trait for it.

etorreborre:
@rbran we're almost there. You need to rename your commits to avoid capital letters, then we should be good

rbran:
> @rbran we're almost there. You need to rename your commits to avoid capital letters, then we should be good

Should I squash the commits? Or just rename it?

etorreborre:
@rbran can you please open a separate PR to accept the CLA: https://github.com/build-trust/ockam/blob/develop/.github/CONTRIBUTORS.csv? Thank you.

etorreborre:
@rbran sorry I missed your CLA PR! Let's merge and re-run

adrianbenavides (approved):
🚂

PR #7049, opened 2023-12-06
Slim down the node manager worker

etorreborre:
This PR refactors the `NodeManagerWorker` so that it only deals with getting parameters from requests and making `Responses` with results from the `NodeManager` for the following services:

 - workers
 - send message
 - inlets
 - kafka services
 - services start: credentials, hop, uppercase etc...
 - credentials
 - node status

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/message.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/message.rs
@@ -3,10 +3,9 @@ use std::sync::Arc;
 use std::time::Duration;
 use tracing::trace;
 
-use minicbor::Decoder;
 use minicbor::{Decode, Encode};
 
-use ockam_core::api::{RequestHeader, Response};
+use ockam_core::api::{Error, RequestHeader, Response};
 use ockam_core::{self, async_trait, AsyncTryClone, Result};
 use ockam_multiaddr::MultiAddr;
 use ockam_node::{Context, MessageSendReceiveOptions};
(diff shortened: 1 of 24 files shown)


implementations/rust/ockam/ockam_api/src/nodes/service/message.rs
```diff
@@ -44,21 +43,20 @@ impl NodeManagerWorker {
...
             .node_manager
             .send_message(ctx, &multiaddr, msg, None)
             .await;
         match res {
-            Ok(r) => Ok(Response::ok(req).body(r).to_vec()?),
+            Ok(r) => Ok(Response::ok(req).body(r)),
             Err(err) => {
                 error!(target: TARGET, ?err, "Failed to send message");
-                Ok(Response::internal_error(req, "Failed to send message").to_vec()?)
+                Err(Response::internal_error(req, "Failed to send message"))
```
adrianbenavides, line 59: I'm finding very confusing the fact that we can return an `Ok(())`, an `Ok(Response::some_error())` and an `Err(Response::some_err())`. 

Perhaps, now that the different clients handle all the interaction internally (InMemoryNode, Controller, etc), this is not necessary anymore and we could remove all the `Ok(Response::some_error())` instances in a future PR? And perhaps also change the types to avoid mixing them again in the future.

adrianbenavides (approved):
Awesome! 🚀

implementations/rust/ockam/ockam_api/src/nodes/service/message.rs
etorreborre: > I'm finding very confusing the fact that we can return an Ok(()), an Ok(Response::some_error()) and an Err(Response::some_err()).

I don't understand where exactly we do this. In general the `NodeManagerWorker` results have the form `Result<Response<T>, Response<Error>>`. Where are the cases with `Ok(Response::error(...))`? (if you also account for the 2 other PRs which are aligning the rest of the code).

adrianbenavides: I'm talking about things that have been there for a long time, it's not anything particular from these PRs, but I'm pointing this out in case now is a good time to reconsider how we handle certain responses.

This is one example, but there are more with `NotFound` and `InternalError` responses: 

https://github.com/build-trust/ockam/blob/f6ef683b51140eb975718edb02970800cf34f766/implementations/rust/ockam/ockam_api/src/authenticator/enrollment_tokens/acceptor.rs#L41

etorreborre: Ah, yes! It would be nice indeed to stop returning `Vec<u8>` everywhere :-)

PR #7054, opened 2023-12-06
fix(rust): don't create default node when retrieving it and doesn't exist

adrianbenavides:
Fixes a problem where running commands that use a `BackgroundNode` would panic if no node is available. The code was storing in the database the node info, but there was no actual background node.

Now, if a background node is created without an explicit name and there is no default node, an error will be returned.


etorreborre (approved):
Why remove the commands for getting/setting a default node since this concept still exists at the CliState / database level?

adrianbenavides:
> Why remove the commands for getting/setting a default node since this concept still exists at the CliState / database level?

Those commands are outdated. We have the `ockam node default` to replace both of them.

PR #7055, opened 2023-12-06
test(rust): fix the bats test by properly importing a project when trying to enroll a member

etorreborre:
This PR does several things:

 1. it reworks the `ockam project import` command to only take a JSON file exported with `ockam project show --output json` so that we don't forget to persist any field when using that command in tests
 2. it allows projects to be listed even if the Controller cannot be accessed if the project data is stored locally
 3. it stores the space data associated to an imported project, even if the space has not been retrieved yet from the Controller (for consistency)
 4. it uses the optional identity name to enroll (this is a fix) 
 5. it properly checks that the project data provided in an enrollment ticket is persisted locally
 6. it only imports project data when creating a new `OCKAM_HOME` directory in bats tests


implementations/rust/ockam/ockam_command/tests/bats/projects_orchestrator.bats
```diff
@@ -59,6 +60,7 @@ teardown() {
   OCKAM_HOME=$NON_ENROLLED_OCKAM_HOME
 
   # Green' identity was added by enroller
+  run_success "$OCKAM" project import --project-file $ENROLLED_OCKAM_HOME/project.json
```
polvorin: Just checking if IIUC,  instead of 
[ockam project enroll --identity $identity --project-path $project_json_path](https://github.com/build-trust/ockam/blob/9ac68511f23e80834e6ce70a440eab676f5b060f/implementations/rust/ockam/ockam_command/tests/bats/projects_orchestrator.bats#L60)

we now have to do
[  ockam project import --project-file $project_json_path
  ockam project enroll --identity $identity](https://github.com/build-trust/ockam/blob/75b78164dfbe6630b8fe74715b0bcf4714d88a9f/implementations/rust/ockam/ockam_command/tests/bats/projects_orchestrator.bats#L79)

there is likely many places where that is needed?

polvorin (dismissed):
There are other places where we have the same issue, some examples

https://github.com/build-trust/ockam/blob/75b78164dfbe6630b8fe74715b0bcf4714d88a9f/implementations/rust/ockam/ockam_command/tests/bats/portals_orchestrator.bats#L72

https://github.com/build-trust/ockam/blob/75b78164dfbe6630b8fe74715b0bcf4714d88a9f/implementations/rust/ockam/ockam_command/tests/bats/portals_orchestrator.bats#L105

https://github.com/build-trust/ockam/blob/75b78164dfbe6630b8fe74715b0bcf4714d88a9f/implementations/rust/ockam/ockam_command/tests/bats/use_cases.bats#L78

implementations/rust/ockam/ockam_command/tests/bats/projects_orchestrator.bats
etorreborre: I think it was only needed in that place because in other places we do the copy right at the beginning of the test.

implementations/rust/ockam/ockam_command/tests/bats/projects_orchestrator.bats
```diff
@@ -25,32 +25,35 @@ teardown() {
...
+  run_failure "$OCKAM" project enroll --identity green
 
+  setup_home_dir
+  BLUE_OCKAM_HOME=$OCKAM_HOME
   run_success "$OCKAM" identity create blue
   blue_identifier=$($OCKAM identity show blue)
-
-  # They haven't been added by enroller yet
+  # blue hasn't been added by enroller yet
   run_failure "$OCKAM" project enroll --identity green
```
polvorin: green doesn't exist here IIUC,  it exists on GREEN_OCKAM_HOME.   Probably you mean blue

implementations/rust/ockam/ockam_command/tests/bats/projects_orchestrator.bats
```diff
@@ -59,13 +62,20 @@ teardown() {
   OCKAM_HOME=$NON_ENROLLED_OCKAM_HOME
 
   # Green' identity was added by enroller
+  OCKAM_HOME=$GREEN_OCKAM_HOME
+  run_success "$OCKAM" project import --project-file $PROJECT_PATH
```
polvorin: this is strange,  the failure on line 50 was because green wasn't enrolled?,  or because the project wasn't imported first?

implementations/rust/ockam/ockam_command/tests/bats/projects_orchestrator.bats
```diff
@@ -74,26 +84,25 @@ teardown() {
...
 
   # Add node as a member
   OCKAM_HOME=$ENROLLED_OCKAM_HOME
-  run_success "$OCKAM" project ticket --member "$green_identifier" --attribute role=member
+  run_success "$OCKAM" project ticket --member "$green_identifier" --attribute role=member --relay $fwd
 
   # The node can now access the project's services
   OCKAM_HOME=$NON_ENROLLED_OCKAM_HOME
-  fwd=$(random_str)
   run_success "$OCKAM" relay create "$fwd"
```
polvorin, line 109: do this create a new node with default identity and creates the relay there,    or use the "green" node created at line 94?. 
Maybe I'm wrong but I think it's creating a new node with default identity?   that's why it was not failing before this fix:    `run_failure "$OCKAM" relay create --identity green "$fwd"`  worked (command failied) because the node was launched with `green` as default identity.   While `run_success "$OCKAM" relay create "$fwd"`  worked because it was using the default identity,  that is the one that have `*` permissions on relays.   So it worked with nothing to do with the $green_identifier being enrolled.

Would be good to confirm what's going on,  and either remove  the `node create green` or add some comment that on line 98 and 106 it's using the `green` background node.

implementations/rust/ockam/ockam_command/tests/bats/projects_orchestrator.bats
etorreborre: Fixed

implementations/rust/ockam/ockam_command/tests/bats/projects_orchestrator.bats
etorreborre: That was indeed because the project wasn't imported first. Fixed.

implementations/rust/ockam/ockam_command/tests/bats/projects_orchestrator.bats
etorreborre: > it uses the "green" node created at line 94?

Yes because this is the default node.

> run_failure "$OCKAM" relay create --identity green "$fwd"

That failed simply because `--identity` is not an option for `relay create`. 

> run_success "$OCKAM" relay create "$fwd" worked because it was using the default identity, that is the one that have * permissions on relays. So it worked with nothing to do with the $green_identifier being enrolled.

Yes, in that case, the default identity was the first one created in the database that was copied at the beginning of the setup. And it was the initially enrolled identity that has a `*` permission indeed!

etorreborre: I have added some comments to explain what's going on with the green node and the green identity.

polvorin (approved):

PR #7064, opened 2023-12-07
Correctly report node status on `node list`

davide-baldo:
when checking for node liveness, `ockam node list` would check the presence of the pid, which was always included.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/list.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/list.rs
@@ -8,6 +8,7 @@ use tokio::try_join;
 
 use ockam::Context;
 use ockam_api::cli_state::nodes::NodeInfo;
+use ockam_api::NodeProcessStatus;
 
 use crate::output::Output;
 use crate::terminal::OckamColor;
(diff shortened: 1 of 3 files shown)


implementations/rust/ockam/ockam_command/src/node/list.rs
```diff
@@ -125,15 +125,19 @@ impl NodeListOutput {
...
-            "Running"
+        let status;
+        let pid;
+        if node_info.is_running() {
+            status = "Running";
+            pid = node_info.pid();
         } else {
-            "Not running"
+            status = "Not running";
+            pid = None;
```
adrianbenavides: Wouldn't this report a false negative when the process is running but in a bad state (dead/zombie)? Is there an easy way to reproduce the wrong node status reporting issue?

etorreborre: I think we should return an enum describing all the situations:

- no pid
- a pid but the process is not running
- a pid, the process is running, but in a zombie state

davide-baldo: I've added the `NodeProcessStatus` structure which explicitly handle the zombie case

adrianbenavides (approved):

PR #7065, opened 2023-12-07
refactor(rust): clean up `project ticket` command

adrianbenavides:


implementations/rust/ockam/ockam_api/src/nodes/service/projects.rs
```diff
@@ -90,14 +90,17 @@ impl Projects for InMemoryNode {
         let user = self.cli_state.get_default_user().await?;
         for project in &projects {
             let mut project = project.clone();
-            if !project.has_admin_with_email(&user.email) {
+            // If the project has no admin role, the name is set to the project id
+            // to avoid collisions with other projects with the same name that
+            // belong to other spaces.
+            if !project.has_admin_role(&user.email) {
```
etorreborre: I think that this can only be a temporary workaround because we rename a project in the back of a user who has no clue that the name they were using do not exist anymore. 

If project names can overlap across spaces and if we can collect projects from different spaces, I suggest that:

 - we report an ambiguity when there are 2 projects with the same name belonging to different spaces
 - we ask the user to disambiguate a project argument with `space-name/project-name`

adrianbenavides: I agree 100%. Right now the limitation is on how we process `MultiAddr` instances. To do what you propose, we need to modify the way we parse the routes to take into account the spaces.

This PR is just a refactor, so we could plan these changes and do them in a separate PR.

etorreborre (approved):

PR #7077, opened 2023-12-08
fix(rust): memory leak caused by the rolling file appender used in the logging setup

adrianbenavides:
Our implementation of the rolling file appender was causing a memory leak. This PR replaces it with the implementation provided by the `tracing_appender` crate.

Before this commit, running the following would create a node that leaks about 0.5mb of ram every 5-10 seconds:

```
# starting from an enrolled state
ockam node create

# create some resources on the node to trigger more logs
ockam tcp-outlet create --to 10000
ockam relay create
ockam tcp-inlet create --from 10001
```

In this branch, the memory remains stable.

### TODO's

A few things left to do:

- The log files names have changed a little and now the current log file has also the date appended... which breaks the `node log` command. What do we do about it?
- Removing [this block here](https://github.com/build-trust/ockam/blob/f6ef683b51140eb975718edb02970800cf34f766/implementations/rust/ockam/ockam_command/src/node/util.rs#L159-L179) blocks the bats tests... not sure why but looks related to dangling file descriptors


implementations/rust/ockam/ockam_command/src/logs/mod.rs
```diff
@@ -64,14 +64,13 @@ pub fn setup_logging(
...
         Some(log_path) => {
-            let r = RollingFileAppender::new(
-                log_path,
-                RollingConditionBasic::new()
-                    .daily()
-                    .max_size(log_max_size_bytes()),
-                log_max_files(),
-            )
-            .expect("Failed to create rolling file appender");
+            let log_dir = log_path.parent().expect("Failed to get log directory");
```
etorreborre: Can we easily refactor the logging logic between the app and the command line?

implementations/rust/ockam/ockam_command/src/node/create/background.rs
```diff
@@ -0,0 +1,113 @@
+use colorful::Colorful;
```
etorreborre: Good idea to split into 2 files

implementations/rust/ockam/ockam_command/src/logs/mod.rs
```diff
@@ -64,14 +64,13 @@ pub fn setup_logging(
...
+            let log_name = log_path
+                .file_name()
+                .expect("Failed to get log file name")
+                .to_str()
+                .unwrap();
+            let r = RollingFileAppender::new(Rotation::MINUTELY, log_dir, log_name);
```
etorreborre: I wonder if `MINUTELY` is a bit too fined-grained, and if that will force us to jump too much between files when debugging. Let's see.

etorreborre (dismissed):

implementations/rust/ockam/ockam_command/src/logs/mod.rs
adrianbenavides: oops I changed that while testing some stuff. It should be daily.

implementations/rust/ockam/ockam_command/src/logs/mod.rs
adrianbenavides: Probably, I'll give it a try

adrianbenavides: Done!

etorreborre (dismissed):

davide-baldo (approved):

PR #7091, opened 2023-12-11
refactor(rust): simplify the creation of the identities builder

etorreborre:
This is a small experiment to check if a deadlock occurs on CI


etorreborre:
This issue was fixed with #7329

PR #7112, opened 2023-12-12
 Fix the stopping of the router + a node start refactoring

etorreborre:
This PR fixes 2 issues:

 1. an `ockam node create` command might stay stuck
 2. the setting of a node pid and tcp listener should be done on node start, not when the node info is retrieved

### `ockam node create`

During testing it has been observed that `ockam node create` could stay stuck if:

 - the corresponding background node is not responding
 - the TCP workers of the foreground node detect the connection drop and clean themselves up
 - the `node create` task starts to close its `Context`
 - then a Router `cluster`, containing the list of all the TCP workers end up returning no address and the `Context` never gets closed

The issue is fixed by considering that there is no cluster to return on shutdown if it contains no active address.

### Setting the node pid and TCP listener

The setting of the node current pid was done in several places and when retrieving the node from the database. 
Same thing for setting the current TCP listener address. 

This PR refactors that code to make sure we only do it in one place.

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/router/record.rs
+++ b/implementations/rust/ockam/ockam_node/src/router/record.rs
@@ -164,24 +164,29 @@ impl InternalMap {
     /// Retrieve the next cluster in reverse-initialisation order
     /// Return None if there is no next cluster or if the cluster
     /// contained no more active addresses
-    pub(super) fn next_cluster(&mut self) -> Option<Vec<&mut AddressRecord>> {
-        let name = self.cluster_order.pop()?;
-        let addrs = self.clusters.remove(&name)?;
-        let active_addresses: Vec<&mut AddressRecord> = self
-            .address_records_map
-            .iter_mut()
-            .filter_map(|(primary, rec)| {
-                if addrs.contains(primary) {
-                    Some(rec)
-                } else {
-                    None
-                }
-            })
-            .collect();
-        if active_addresses.is_empty() {
-            None
-        } else {
-            Some(active_addresses)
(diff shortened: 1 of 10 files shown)


implementations/rust/ockam/ockam_node/src/router/record.rs
```diff
@@ -162,21 +162,27 @@ impl InternalMap {
...
+                } else {
+                    None
+                }
+            })
+            .collect();
+        if active_addresses.is_empty() {
+            None
+        } else {
+            Some(active_addresses)
+        }
```
adrianbenavides: > then a Router cluster, containing the list of all the TCP workers end up returning no address and the Context never gets closed

Just to make sure I'm following up, this change here fixes the problem above?

etorreborre: Yes

etorreborre: Actually, I'm fixing this method. It is incorrect as it is because it wouldn't skip to the next non-empty cluster.

etorreborre: @adrianbenavides I pushed a fix and a test to make sure that `next_cluster` really does what it should.

adrianbenavides: Nice catch there! It makes perfect sense to me now

adrianbenavides (dismissed):
🚀

SanjoDeundiak (approved):

PR #7120, opened 2023-12-13
ci: cross compile aarch64 application for mac

davide-baldo:
On aarm64, the app wasn't properly cross-compiling.


metaclips (dismissed):

metaclips (approved):
Nice

PR #7121, opened 2023-12-13
refactor(rust): split `node create` command code into separate files for background/foreground modes

adrianbenavides:
It also:
- Fixes the `node logs` command to return the current log file
- Remove the `stderr` log file, which is not used anymore 

Full [CI tests](https://github.com/build-trust/ockam-artifacts/actions/runs/7194879891)


etorreborre (approved):
Excellent!

PR #7125, opened 2023-12-13
Simplify the use of vaults

etorreborre:
This PR simplifies the use of vaults:

# `--vault` option consistency

This PR removes the `--vault` option when it is implied by the `--identity` option (except on credentials commands which are going to be removed)

# Default vault

This PR removes the management of default vaults. A vault is simply the default vault if there is only one vault. Otherwise any attempt to return the "default" vault will return an error asking to specify which vault must be used.

This works with all of our examples without changing the current UX.

# Consistency

This PR it checks if any identity is using a vault before deleting it and returns an error if it is the case.
Additionally if we delete a vault and that vault is the default one, the tables used by the vault are now deleted as well (and not only the vault metadata).

In this PR we also make sure that we don't mix-up vault creation and vault update at the repository level. The `VaultsRepository::store` function can _only_ store a new vault. 

Then the only use case for updating a vault is moving its backend file if it is has not been stored in the main database.
In order to do this:

 - a command `ockam vault move vault-name --path new-path` has been added
 - an `update` function has been added on the `VaultsRepository` to update the vault path

Some more unit tests, at the level of `CliState` have been added, to provide a better coverage of the main functions.


implementations/rust/ockam/ockam_command/src/vault/move_vault.rs
```diff
@@ -0,0 +1,56 @@
+use clap::Args;
```
SanjoDeundiak, line 1: Shouldn't the file be called just `move.rs`?

implementations/rust/ockam/ockam_api/src/cli_state/vaults.rs
```diff
@@ -112,6 +174,18 @@ impl CliState {
...
+
+    /// Move a vault file to another location if the vault is not the default vault
+    /// contained in the main database
+    pub async fn move_vault(&self, vault_name: &str, path: &Path) -> Result<()> {
+        let repository = self.vaults_repository().await?;
+        let vault = self.get_named_vault(vault_name).await?;
+        if vault.path() == self.database_path() {
+            return Err(ockam_core::Error::new(Origin::Api, Kind::Invalid, format!("The vault at path {:?} cannot be moved to {path:?} because this is the default vault", vault.path())).into());
+        };
+        std::fs::rename(vault.path(), path)?;
```
SanjoDeundiak: Should the process be the following?
1. Copy the file to the new location
2. Update the path in the db
3. Remove the old file

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_command/src/vault/move_vault.rs
etorreborre: It can be called `move` because this is a reserved keyword.

implementations/rust/ockam/ockam_api/src/cli_state/vaults.rs
etorreborre: Great idea, I updated the code.

SanjoDeundiak (approved):

PR #7146, opened 2023-12-15
docs(rust): improve the implementations/rust README 

etorreborre:
This PR merges the `DEVELOP.md` and `README.md` files in `implementations/rust` to just a `README` file.
It provides instructions to build / test / lint the code with:

 1. the regular `cargo` tools
 2. shortcut commands using the provided `make` file
 3. shortcut commands using the provided `make` file and a `nix` environment where all the necessary tools and cargo plugins are already installed

The last option is quite practical since it gives access to all the development commands in a fully configured environment. For example:
```
# run the cargo tests
make nix_rust_test

# run the bats tests
make nix_rust_bats
```


Makefile
```diff
@@ -9,14 +9,19 @@ elixir_%:
...
-	$(MAKE) -C implementations/rust $(@:rust_%=%)
+	$(MAKE) -f implementations/rust/Makefile $(@:rust_%=%)
+
+# run a rust command in a nix environment, so that all tools are installed
+nix_rust_%:
+	nix develop ./tools/nix#rust --command make rust_$*
 
 swift_%:
-	$(MAKE) -C implementations/swift $(@:swift_%=%)
+	$(MAKE) -f implementations/swift/Makefile $(@:swift_%=%)
```
metaclips: Let's revert this change as it breaks release, using the -C flag ensures we cd to the swift directory before running the make commands, which is crucial for building the desktop app
https://github.com/build-trust/ockam/blob/4862e4ae1a9cb9dcf0a2604482be2b05cb676c63/implementations/swift/Makefile#L2

metaclips (dismissed):

Makefile
etorreborre: Done

SanjoDeundiak (approved):

PR #7167, opened 2023-12-18
refactor binary release workflow

metaclips:
- Ensure that we can release without needing a -ff merge to retain git commit SHA
- We can now release crates even where there are no changes to the ockam, ockam_command, and ockam_app_lib crate
- We now use same version of ockam, ockam_app_lib and ockam_command

How release workflow runs
- Create a PR that bumps the version of our crates, after PR is created, script waits for PR to be merged
- PR merged and script automatically detects and start release from the develop branch

### DIFF
--- a/.github/workflows/release-publish-crates.yml
+++ b/.github/workflows/release-publish-crates.yml
@@ -11,7 +11,7 @@ on:
         required: false
       release_branch:
         description: Release Branch Where Recent Bump Occurred
-        required: true
+        required: false
       ockam_publish_recent_failure:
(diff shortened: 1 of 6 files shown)


davide-baldo:
@metaclips 
Why do we need to bump one of those three crates in order to release?

mrinalwadhwa (approved):

metaclips:
> @metaclips Why do we need to bump one of those three crates in order to release?
We need the version desktop app and CLI to be same

PR #7169, opened 2023-12-18
Shautvast/add cbor validation

shautvast:
## Current behavior

Currently structs are serialized using CBOR, but the messages are not validated against a CDDL schema definition

## Proposed changes
fix for issue https://github.com/build-trust/ockam/issues/6690

cddl validation added for:
* nodes/models/base/NodeStatus
* nodes/models/credentials/GetCredentialRequest
* nodes/models/credentials/PresentCredentialRequest
* nodes/models/flow_controls/AddConsumer
* nodes/models/policy/PolicyList
* nodes/models/policy/Expression
The cddl schema file was extended where necessary

No validation added for the following structs since they seem to have been removed since the issue was created
* nodes/models/forwarder/CreateForwarder
* nodes/models/forwarder/ForwarderInfo
* nodes/models/identity/LongIdentityResponse
* nodes/models/identity/ShortIdentityResponse
* nodes/models/policy/Policy

Also: as mentioned in the issue as an optional refactor, CowStr and it's lifetime have been removed

## Checks

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/models/policy.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/models/policy.rs
@@ -1,7 +1,7 @@
 use minicbor::{Decode, Encode};
 use ockam_abac::{Action, Expr};
 
-#[derive(Debug, Decode, Encode)]
+#[derive(Debug, Clone, Decode, Encode)]
 #[rustfmt::skip]
 #[cbor(map)]
 pub struct PolicyList {
@@ -18,7 +18,7 @@ impl PolicyList {
     }
 }
 
-#[derive(Debug, Decode, Encode)]
+#[derive(Debug, Clone, Decode, Encode)]
 pub struct Expression {
     #[n(1)]
     action: Action,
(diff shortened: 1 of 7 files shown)


etorreborre:
Hi @shautvast. Thanks for your contribution! Can you please [adjust your commit messages to follow the convention](https://github.com/build-trust/ockam/actions/runs/7251043509/job/19756070397?pr=7169)?

implementations/rust/ockam/ockam_api/src/schema/schema.cddl
```diff
@@ -249,3 +249,48 @@ activate_subscription = {
...
+
+multi_addr = bytes
+
+;;; policy ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+policy_list = {
+    1: [* expression]                                   ;; expressions
+}
+
+expression = [action expr]                              ;; Expression
```
etorreborre, line 293: This grammar does not follow exactly the struct definitions for `Expr`:

 - `action` is `(nil text)`. Why `nil`?
 - `1..7 expr_data` instead of `* expr_data`
 - no distinction between `Expr::Seq` and `Expr::List`
 - there is no `Expr::Ident`

I am not really familiar with CDDL and there might be good reasons why you did it this way. Don't you think we should spend more time to make the grammar more precise?

implementations/rust/ockam/ockam_api/src/nodes/models/policy.rs
```diff
@@ -39,3 +39,80 @@ impl Expression {
...
+    }
+
+    quickcheck! {
+        fn policy(g: Expression) -> TestResult {
+            validate_with_schema("expression", g)
+        }
+    }
+
+    impl Arbitrary for PolicyList {
+        fn arbitrary(g: &mut Gen) -> Self {
```
etorreborre, line 65: It would be good to compose an `Arbitrary for Action` and an `Arbitrary for Expr` to create this `Arbitrary`

implementations/rust/ockam/ockam_api/src/nodes/models/policy.rs
```diff
@@ -39,3 +39,80 @@ impl Expression {
...
+                ],
+            }
+        }
+    }
+
+    impl Arbitrary for Expression {
+        fn arbitrary(g: &mut Gen) -> Self {
+            Expression {
+                action: Action::new(String::arbitrary(g).as_str()),
+                expr: Expr::List(vec![
```
etorreborre, line 111: I would suggest to extract an `Arbitrary for Expr` and extend it to other types of `Expr` (`Str`, `Ident`, etc...).

implementations/rust/ockam/ockam_api/src/schema/schema.cddl
shautvast: #1 why nil: stupid argument: it won't work without it :). The nil is there in the binary cbor (246_u8 to be precise) . This may even be an bug in mini-cbor as far as I know. NB this expresssion is not a choice, it's a ccdl list, so nil followed by text is how the Action struct is serialized. Maybe this is just the way tuple structs are handled, because they don't have named fields. 

#2 I don't see [* expr_data]. You're referring to [+ expr_data]? I find that recursive definition quite accurate as it reflects the one in Expr

#3 in cbor representation there is only the enum constant (6 or 7) so that's how it's distinguished when (de)serializing. The enum data is the same. 

#4 It's there as enum constant 5. Contents are same as Str, (String -> text) so that's the reason it's not mentioned explicitly

Yes I suspect that the expr cddl can be somewhat more precise. Will look into that

shautvast: 1. The cddl is now both clearer and more strict. Please have a look.
2. I found the actual reason for Nil

This is the definition:
pub struct Expression {
    #[n(1)]
    action: Action,
    #[n(2)]
    expr: Expr,
}

which is encoded as a 0-indexed list, but the first attribute (Action) hs #[n(1)], so Nil is inserted at index 0 to put action at index 1 in the list. One way to fix this is to replace 1 and 2 by 0 and 1, so that Nil can just disappear from the CDDL and you save 1 byte in the serialized message. Love to hear your thoughts on that

implementations/rust/ockam/ockam_api/src/nodes/models/policy.rs
shautvast: agree, but the _validate_with_schema_ test util is in ockam_api and Expr and Action are in ockam_abac

I see no easy and elegant solution to this (but maybe my rust knowledge is incomplete here)
I tried to move the util to ockam_abac, but is it possible to have dependencies on test modules?
otherwise I'd have to copy, which I don't like, obviously

implementations/rust/ockam/ockam_api/src/nodes/models/policy.rs
shautvast: same as above. Expression is in ockam_api and the rest is in ockam_abac

PR #7190, opened 2023-12-19
App - handling email validation as en enrollment state

davide-baldo:
Added `WaitForEmailValidation` as an enrollment phase, with its relative view during enrollment.
Removed email validation notification.

Waiting for 10 seconds instead of 5 to keep the API rate low enough to avoid failure, this is valid for both command and portals.


mrinalwadhwa (approved):

mrinalwadhwa:
@davide-baldo that worked great. I changed some text and added the start enrollment again button.

SanjoDeundiak (approved):

PR #7200, opened 2023-12-20
docs: fix the make swift_build command

etorreborre:
The current `make swift_build` command does not work because `xcodebuild` expects an `ockam` executable in `target/release`, even though we should building the `debug` version.

This PR:

 - makes sure that all the Rust code is recompiled before building the Swift code since we eventually need the `ockam` executable
 - copies the `ockam` executable is copied to the right place in debug mode. It feels like the wrong to do it because I couldn't find a way to change the XCode project to pick either the `debug` or the `release` executable when building the Portals application
 - adds 2 convenience commands `make swift_run` and `make_swift_run_release` to build and start the application at once

### DIFF
--- a/implementations/swift/Makefile
+++ b/implementations/swift/Makefile
@@ -1,25 +1,53 @@
 ROOT_DIR := $(realpath $(dir $(firstword $(MAKEFILE_LIST)))/../..)
 
-build: ockam_app_lib
-    # the xcode project expect the ockam executable to be found in the release directory
-	cp $(ROOT_DIR)/target/debug/ockam $(ROOT_DIR)/target/debug/release
-	xcodebuild archive -project ockam/ockam_app/Ockam.xcodeproj/ -scheme "Portals, by Ockam" -configuration Debug -archivePath build/Ockam.xcarchive
-	@echo "Build complete.  See build/Ockam.xcarchive for the built app."
-	@echo "To execute run: build/Ockam.xcarchive/Products/Applications/Portals,\ by\ Ockam.app/Contents/MacOS/Portals,\ by\ Ockam"
-build_release: ockam_app_lib_release xcode_build
-xcode_build:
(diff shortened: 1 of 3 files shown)


implementations/swift/Makefile
```diff
@@ -10,11 +14,11 @@ xcode_build:
 xcode_signed_build:
 	xcodebuild archive -project ockam/ockam_app/Ockam.xcodeproj/ -scheme "Portals, by Ockam" -configuration Release -archivePath build/Ockam.xcarchive -arch $(ARCH)  CODE_SIGN_IDENTITY="${CODE_SIGN_IDENTITY}"  PROVISIONING_PROFILE_SPECIFIER="${PROVISIONING_PROFILE_SPECIFIER}" DEVELOPMENT_TEAM="${DEVELOPMENT_TEAM}"
 ockam_app_lib_release:
-	$(MAKE) -f ../rust/Makefile build_release_ockam_app_lib
+	$(MAKE) -f ../rust/Makefile build_release
```
davide-baldo: wouldn't be easier to move the `ockam_command_release` dependency to `xcode_build`?

etorreborre: I also think it's a bit convoluted. I refactored it (see [3c2be02](https://github.com/build-trust/ockam/pull/7200/commits/3c2be027a6ea9ad0ec06f1ec92e004f5f4f7eaa7)), tell me what you think.

implementations/swift/Makefile
```diff
@@ -1,21 +1,53 @@
...
+# build the Portals application in Release mode
+build_release: build_ockam_release
+	make xcode_build_Release
+
+# build the Portals application in Release mode and sign it
+build_signed: build_ockam_release
+	make xcode_build_signed
+
+# build the Swift code in Debug or Release mode
+xcode_build_%:
```
davide-baldo, line 34: Could the wildcard intercept the `xcode_build_signed` as well?

davide-baldo (approved):

implementations/swift/Makefile
etorreborre: Yes I tested it.

PR #7203, opened 2023-12-21
fix(rust): update unsafe-libyaml dependency

adrianbenavides:


metaclips (approved):
Thanks @adrianbenavides

PR #7205, opened 2023-12-21
Use an `EmailAddress` newtype

etorreborre:
This PR introduce an `EmailAddress` newtype in order to:

 - make sure we don't pass a string with another meaning (project id, user id etc....) to an email address field
 - enforce the same case-insensitive comparison for email addresses (this is especially important since we currently use email addresses as identifiers)
 - parse email addresses in the Portals application (when sharing a service) with the same logic as the logic used in the Controller. This makes sure that the email address will not be silently rejected 

Note: I started by using the `email_address` crate but eventually removed that dependency in favor of doing a regex validation that is similar to the one we do on the controller.


implementations/rust/ockam/ockam_api/src/cloud/email_address.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+/// Lowercase comparison for email addresses
+impl PartialEq for EmailAddress {
+    fn eq(&self, other: &Self) -> bool {
+        self.0.to_lowercase() == other.0.to_lowercase()
+    }
+}
+
+impl Default for EmailAddress {
```
SanjoDeundiak: Is it useful? I would remove it. Not an obvious implementation

implementations/rust/ockam/ockam_api/src/cloud/email_address.rs
```diff
@@ -0,0 +1,210 @@
...
+    fn default() -> Self {
+        EmailAddress("test@ockam.io".to_string())
+    }
+}
+
+impl Display for EmailAddress {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0.clone())
```
SanjoDeundiak: I'm surprised you need to clone

implementations/rust/ockam/ockam_api/src/cloud/email_address.rs
```diff
@@ -0,0 +1,210 @@
...
+                    "x@example.com",
+                    "long.email-address-with-hyphens@and.subdomains.example.com",
+                    "user.name+tag+sorting@example.com",
+                    "name/surname@example.com",
+                    "admin@example",
+                    "example@s.example",
+                    "mailhost!username@example.org",
+                    "user%example.com@example.org",
+                    "user-@example.org",
+                    // "\" \"@example.org",
```
SanjoDeundiak, line 156: Did you intentionally leave this?

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_api/src/cloud/email_address.rs
etorreborre: It is required because another struct is using `EmailAddress` and deriving `Default`.

implementations/rust/ockam/ockam_api/src/cloud/email_address.rs
etorreborre: You're right, I removed it.

implementations/rust/ockam/ockam_api/src/cloud/email_address.rs
etorreborre: Yes, I explain this in the comment [here](https://github.com/build-trust/ockam/pull/7205/files/7309b9834914ff3cd210f279f076040eb560abc2#diff-c85e5db2e533f3560c0ea50cb110aeb59ba0ac50b08e8e134347264d66c255e3R142). I'm going to bring that comment closer.

implementations/rust/ockam/ockam_api/src/cloud/email_address.rs
SanjoDeundiak: Which one? Maybe it should not?

etorreborre: This is done for `UserInfo` but I did not realize that it was only done for tests. So I also added a `cfg(test)` attribute there and maybe we will do a refactoring for tests to remove those default instances (I'd rather have arbitraries for tests).

etorreborre: Ok, I was wrong. None of this is necessary.

SanjoDeundiak (approved):

PR #7219, opened 2023-12-22
fix(rust): call the right function when enrolling with okta

etorreborre:
This PR fixes

 - [an incorrect function call](https://github.com/build-trust/ockam/pull/7219/files#diff-5fd702823db300c905c424872daadede22639f8c2bff15f22ba4b53570f6f4f1R115) when enrolling with Okta
 - a duplication of rows in the `okta_config` table

It also adds some logging to help diagnose issues when checking a token


etorreborre:
This has been fixed in another PR

PR #7222, opened 2023-12-22
feat(rust): close unneeded tcp connections in various clients

SanjoDeundiak:


implementations/rust/ockam/ockam_api/src/nodes/service/background_node_client.rs
```diff
@@ -182,30 +203,30 @@ impl BackgroundNode {
...
+            .modify()
+            .prepend(tcp_connection.sender_address().clone());
         debug!("Sending requests to {route}");
-        Ok(route)
+        Ok((tcp_connection, route))
     }
 
     /// Make a response / request client connected to the node
-    pub async fn make_client(&self) -> miette::Result<Client> {
+    pub(crate) async fn make_client(&self) -> miette::Result<(TcpConnection, Client)> {
```
etorreborre, line 219: Is there a way to make this a bit more ergonomic, either by

 - passing a function using the `Client` and closing the connection when the function is done (Rust type system might not be super-accommodating with this)
 - return some kind of Droppable struct which will close the connection when done (`async` might be an issue there)
 - any other idea?

etorreborre: BTW it's better to separate the renaming in a separate commit for this kind of fix :-)

SanjoDeundiak: I think there is a room for better ergonomics, but as you already mentioned it's tricky, so I decided to submit a fix instead of digging into the ergonomics

SanjoDeundiak: @etorreborre do you want me to split it now, or is it ok?

mrinalwadhwa (approved):

PR #7300, opened 2023-12-31
Node-isolated attributes

SanjoDeundiak:
Introduce `node_name` field in identity_attributes sqlite table to isolate attributes between nodes. Otherwise, if we have multiple nodes with different projects talking to the same Identifier - attributes are overwritten

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/vaults.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/vaults.rs
@@ -364,8 +364,10 @@ impl NamedVault {
         }
     }
 
-    async fn database(&self) -> Result<Arc<SqlxDatabase>> {
-        Ok(Arc::new(SqlxDatabase::create(self.path.as_path()).await?))
+    async fn database(&self) -> Result<SqlxDatabase> {
+        // FIXME: We should really have one instance of the SqlxDatabase per process, also the node_name is incorrect,
+        //  but we don't use it yet for the vault storage
+        Ok(SqlxDatabase::create(self.path.as_path(), "vault_node_name".to_string()).await?)
     }
 }
 
@@ -504,7 +506,7 @@ mod tests {
     async fn test_move_vault() -> Result<()> {
         let db_file = NamedTempFile::new().unwrap();
         let cli_state_directory = db_file.path().parent().unwrap().join(random_name());
-        let cli = CliState::create(cli_state_directory.clone()).await?;
+        let cli = CliState::create(cli_state_directory.clone(), "test_node".to_string()).await?;
 
         // create a vault
         let _ = cli.get_or_create_named_vault("vault1").await?;
--- a/implementations/rust/ockam/ockam_command/src/lib.rs
+++ b/implementations/rust/ockam/ockam_command/src/lib.rs
@@ -244,7 +244,8 @@ pub struct CommandGlobalOpts {
 impl CommandGlobalOpts {
     pub fn new(global_args: GlobalArgs) -> Self {
         let terminal = Terminal::from(&global_args);
-        let state = match CliState::with_default_dir() {
+        // This node_name is a placeholder and should be set to a proper value when creating a node
+        let state = match CliState::with_default_dir("ockam_command".to_string()) {
             Ok(state) => state,
             Err(_) => {
                 terminal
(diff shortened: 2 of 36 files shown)


implementations/rust/ockam/ockam_command/src/lib.rs
```diff
@@ -244,7 +244,7 @@ pub struct CommandGlobalOpts {
 impl CommandGlobalOpts {
     pub fn new(global_args: GlobalArgs) -> Self {
         let terminal = Terminal::from(&global_args);
-        let state = match CliState::with_default_dir() {
+        let state = match CliState::with_default_dir("test".to_string() /* FIXME */) {
```
adrianbenavides: This node name here doesn't matter because it's changed later at https://github.com/build-trust/ockam/pull/7300/files#diff-11d072980aa61e0db956f9b7339d26c69ed4db11a2e0714c635ddd610485d283R63 Is that right?

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
```diff
@@ -101,8 +111,13 @@ impl SqlxDatabase {
     }
 
     async fn migrate(&self) -> Result<()> {
+        Self::migrate_tables(&self.pool).await?;
+        self.migrate_attributes_node_name().await
```
adrianbenavides, line 115: I have some concerns about this migration strategy. SQL and rust migrations should be paired so we can run/test step-by-step migrations. I haven't looked if this is possible with sqlx, but we could give it a try if we agree on doing something like this.

implementations/rust/ockam/ockam_api/src/cli_state/vaults.rs
```diff
@@ -364,8 +364,10 @@ impl NamedVault {
         }
     }
 
-    async fn database(&self) -> Result<Arc<SqlxDatabase>> {
-        Ok(Arc::new(SqlxDatabase::create(self.path.as_path()).await?))
+    async fn database(&self) -> Result<SqlxDatabase> {
+        // FIXME: We should really have one instance of the SqlxDatabase per process, also the node_name is incorrect,
+        //  but we don't use it yet for the vault storage
+        Ok(SqlxDatabase::create(self.path.as_path(), "vault_node_name".to_string()).await?)
```
adrianbenavides, line 370: Perhaps having a `OnceLock` static + adding a `SqlxDatabase::new` method would do the trick? Or are you thinking about something else?

SanjoDeundiak: I think that `SqlxDatabase` shouldn't be created here at all. It should be created once per db file / per process (and we currently have only one db file). Then that instance can be cloned and used everywhere else. But this has nothing to do with this PR

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
SanjoDeundiak: Approach in this PR is from this PR by @etorreborre https://github.com/build-trust/ockam/pull/6979
Pairing migrations would be an improvement, I agree. Not sure how easy it will be to achieve with sqlx

implementations/rust/ockam/ockam_command/src/lib.rs
SanjoDeundiak: Short answer: yes
Long answer:
The problem is that CliState struct is designed (and used) as something that is not node-dependent, which has been fine so far, since it really manages the database which we have one per machine. But there are multiple reasons to isolate data between nodes. In this PR I isolate attributes a node has received form other nodes (inside credentials) using a dedicated field in the corresponding table, but I think:
1. There are probably other cases like that, so those tables can now easily use the same approach, since `node_name` value is already available to them.
2. I think there is also room for performance improvements - each node can have dedicated db file(s).
That being said, I see benefits in `CliState` being node-dependent. However, that transition wasn't that easy to do, since node name is not available at the stage where we instantiate `CliState`, ideally we should improve that, but for now this approach should work

implementations/rust/ockam/ockam_api/src/cli_state/vaults.rs
SanjoDeundiak: Maybe we could achieve that with OnceLock inside the SqlxDatabase, but I was thinking we could achieve the same by choosing the correct place where we instantiate the object

adrianbenavides (approved):
Looks good to me. I'll create a couple of issues from the comments I left for further improvements/refactorings.

PR #7314, opened 2024-01-02
refactor(rust): do not force node_name on database creation

adrianbenavides:


SanjoDeundiak (approved):
Does any of website examples need update because of the credential command change?

adrianbenavides:
> Does any of website examples need update because of the credential command change?

Nope, as the website examples stored credentials that don't include identity attributes. That is, the identity attributes repository/database table is not used.

PR #7369, opened 2024-01-08
feat(rust): catch execution panics and exit the process

etorreborre:
When we execute async functions using the `Executor` we might be left hanging if a panic is raised.
This PR:
 
 - catches the panic information
 - displays it
 - and exits the process

It is implemented as a panic hook, at the level of the `NodeBuilder` which is the point where we create the application `Context` and `Executor`, which are used throughout a whole command invocation or portal application use.

I tested this change by:

 - adding a `panic` in various places in the code (like the `InMemoryNode` creation), 
 - calling a command, like `ockam status`
 - checking that we would exit the process

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/node.rs
+++ b/implementations/rust/ockam/ockam_node/src/node.rs
@@ -19,6 +19,7 @@ impl ockam_core::Worker for NullWorker {
 /// builder API to customise the underlying node that is created.
 pub struct NodeBuilder {
     logging: bool,
+    exit_on_panic: bool,
 }
 
 impl Default for NodeBuilder {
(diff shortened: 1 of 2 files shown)


SanjoDeundiak (approved):

implementations/rust/ockam/ockam_node/src/node.rs
```diff
@@ -45,6 +60,20 @@ impl NodeBuilder {
...
 
+        // building a node should happen only once per process
+        // to create the Context and the Executor (containing the Router)
+        // Since the Executor is used to run async functions we need to catch
+        // any panic raised by those functions and exit the current process in case this happens.
+        // Otherwise the Executor might stay blocked on the Router execution.
+        #[cfg(feature = "std")]
+        if self.exit_on_panic {
+            std::panic::set_hook(Box::new(|panic_info| {
+                println!("A fatal error occurred: {panic_info}.");
```
SanjoDeundiak, line 71: Is there a reason to use print instead of logging?

SanjoDeundiak (approved):

implementations/rust/ockam/ockam_node/src/node.rs
etorreborre: I assumed this is what assumes the least in terms of what's working and what's not.

PR #7370, opened 2024-01-08
Fix `RemoteCredentialRetriever`

SanjoDeundiak:
`RemoteCredentialsRetriever` has a route to the Authority node. Every time we ask it for a credential, it's meant to create a fresh tcp connection and a fresh secure channel. However, the tcp connection is created beforehand somewhere else, and is handed to the `RemoteCredentialsRetriever` as a resolved TCP Worker Address in the route. Therefore, when that tcp connection dies (for any reason, like connectivity problem or if it's unused for some time), it won't be recreated and we won't be able to get a credential.

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/context/transports.rs
+++ b/implementations/rust/ockam/ockam_node/src/context/transports.rs
@@ -1,6 +1,7 @@
+use ockam_core::compat::collections::HashMap;
 use ockam_core::compat::sync::Arc;
 use ockam_core::errcode::{Kind, Origin};
-use ockam_core::{Error, Result, Route, TransportType};
+use ockam_core::{Address, Error, Result, Route, TransportType};
 use ockam_transport_core::Transport;
 
 use crate::Context;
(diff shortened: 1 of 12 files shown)


implementations/rust/ockam/ockam_node/src/context/transports.rs
```diff
@@ -24,6 +25,18 @@ impl Context {
...
         let transports = self.transports.read().unwrap().clone();
 
+        let (resolved_route, _resolved_address) =
+            Self::resolve_transport_route_static(route, transports).await?;
+        Ok(resolved_route)
+    }
+
+    /// For each address handled by a given transport in a route, for example, (TCP, "127.0.0.1:4000")
+    /// Create a worker supporting the routing of messages for this transport and replace the address
+    /// in the route with the worker address. Also, returns the connection worker address
```
etorreborre: It would be good to specify in the comment in which case there is going to be a return connection worker address or not, so that we don't have to look at the code to know.

etorreborre (dismissed):

etorreborre (approved):

PR #7372, opened 2024-01-08
Add docs bats test

### LINKED ISSUE #7373: A reusable workflow that can run the docs tests
We have a PR https://github.com/build-trust/ockam/pull/7372 that creates a workflow that runs docs test on the ockam repository, we should ensure that this workflow is reusable so that we can use across multiple repository

### LINKED ISSUE #7374: Run ockam documentation tests for every PR merged in the ockam repository
We have a workflow in place to check if our docs tests is outdated, we should ensure that we run this workflow on our docs right after a PR is merged to ensure that our documentation is not outdated/broken because of a release.

metaclips:
Adds bats tests of our CLI examples in the docs.ockam.io website

### DIFF
--- a/implementations/rust/ockam/ockam_command/tests/bats/load/base.bash
+++ b/implementations/rust/ockam/ockam_command/tests/bats/load/base.bash
@@ -85,7 +85,7 @@ teardown_home_dir() {
     if [[ -z "$BATS_TEST_COMPLETED" && -z "$BATS_TEST_SKIPPED" ]]; then
       # Copy the CLI directory to $HOME/.bats-tests so it can be inspected.
       # For some reason, if the directory is moved, the teardown function gets stuck.
-      echo "Failed test dir: $OCKAM_HOME $BATS_TEST_COMPLETED $BATS_TEST_SKIPPED" >&3
+      echo "Failed test dir: $OCKAM_HOME" >&3
       cp -r "$OCKAM_HOME/." "$HOME/.bats-tests"
     fi
     run $OCKAM node delete --all --force --yes
--- a/implementations/rust/ockam/ockam_command/tests/bats/load/base.bash
+++ b/implementations/rust/ockam/ockam_command/tests/bats/load/base.bash
@@ -81,10 +81,11 @@ teardown_home_dir() {
   for dir in "${DIRS[@]}"; do
     OCKAM_HOME="$dir"
     # If BATS_TEST_COMPLETED is not set, the test failed.
-    if [[ -z "$BATS_TEST_COMPLETED" ]]; then
+    # If BATS_TEST_SKIPPED is not set, then the test was not skipped
+    if [[ -z "$BATS_TEST_COMPLETED" && -z "$BATS_TEST_SKIPPED" ]]; then
       # Copy the CLI directory to $HOME/.bats-tests so it can be inspected.
       # For some reason, if the directory is moved, the teardown function gets stuck.
-      echo "Failed test dir: $OCKAM_HOME" >&3
+      echo "Failed test dir: $OCKAM_HOME $BATS_TEST_COMPLETED $BATS_TEST_SKIPPED" >&3
       cp -r "$OCKAM_HOME/." "$HOME/.bats-tests"
     fi
     run $OCKAM node delete --all --force --yes
--- a/implementations/rust/ockam/ockam_command/tests/bats/load/orchestrator.bash
+++ b/implementations/rust/ockam/ockam_command/tests/bats/load/orchestrator.bash
@@ -14,6 +14,13 @@ function skip_if_orchestrator_tests_not_enabled() {
   fi
 }
 
+function skip_if_docs_tests_not_enabled() {
+  # shellcheck disable=SC2031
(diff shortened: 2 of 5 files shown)


implementations/rust/ockam/ockam_command/tests/bats/load/base.bash
```diff
@@ -81,10 +81,11 @@ teardown_home_dir() {
   for dir in "${DIRS[@]}"; do
     OCKAM_HOME="$dir"
     # If BATS_TEST_COMPLETED is not set, the test failed.
-    if [[ -z "$BATS_TEST_COMPLETED" ]]; then
+    # If BATS_TEST_SKIPPED is not set, then the test was not skipped
+    if [[ -z "$BATS_TEST_COMPLETED" && -z "$BATS_TEST_SKIPPED" ]]; then
       # Copy the CLI directory to $HOME/.bats-tests so it can be inspected.
       # For some reason, if the directory is moved, the teardown function gets stuck.
-      echo "Failed test dir: $OCKAM_HOME" >&3
+      echo "Failed test dir: $OCKAM_HOME $BATS_TEST_COMPLETED $BATS_TEST_SKIPPED" >&3
```
etorreborre: Could you please add some strings before the variables so we understand what they mean?

implementations/rust/ockam/ockam_command/tests/bats/use_cases.bats
```diff
@@ -16,12 +15,16 @@ setup() {
 }
 
 teardown() {
+  kill_kafka_contents || true
+  kill_flask_server || true
+  kill_telegraf_instace || true
```
etorreborre: s/kill_telegraf_instace/kill_telegraf_instance

implementations/rust/ockam/ockam_command/tests/bats/use_cases.bats
```diff
@@ -73,38 +73,395 @@ teardown() {
...
-  $OCKAM project enroll $x_token --identity x_identity
+  $OCKAM project enroll "$ADMIN_HOME/x.ticket" --identity x_identity
   $OCKAM node create x --identity x_identity
   $OCKAM policy create --at x --resource tcp-inlet --expression '(= subject.component "control")'
   $OCKAM tcp-inlet create --at /node/x --from "127.0.0.1:$port_2" --to "$fwd"
   run curl --fail --head --max-time 10 "127.0.0.1:$port_2"
   assert_failure 28 # timeout error
 }
+
+start_python_server() {
```
etorreborre: This test file is already large. Could you please move all the support code, for python, kafka, influxdb/telegraf, to separate files? (maybe you can even create a use-cases directory)

etorreborre: Then, maybe, what we can do in the docs repository is to:

 1. Query the `use-case` directory for the last commit which changed anything in that directory.
 2. If that commit is different from the last commit we stored inside the docs repository, we raise an error in the docs repository CI.
 3. Then we fix the docs if necessary and write down the new `ockam` commit that we think is now the correct one for the docs.

implementations/rust/ockam/ockam_command/tests/bats/load/orchestrator.bash
```diff
@@ -14,6 +14,13 @@ function skip_if_orchestrator_tests_not_enabled() {
   fi
 }
 
+function skip_if_docs_tests_not_enabled() {
```
etorreborre, line 17: Could you please add a section in the README to explain how to run those tests?
One interesting part is the setup. Do they all work in a nix shell? That would be fantastic.

implementations/rust/ockam/ockam_command/tests/bats/load/base.bash
metaclips: Was for debugging purposes, I'll revert now

implementations/rust/ockam/ockam_command/tests/bats/load/orchestrator.bash
metaclips: We use Nix for most of the setup, I had to use the [supported Docker image of Influx](https://docs.influxdata.com/influxdb/v2/install/?t=Docker#persist-data-outside-the-influxdb-container) as Nix Influxd wasn't working accurately. I added how to run the Docs in the readme section.

implementations/rust/ockam/ockam_command/tests/bats/use_cases.bats
metaclips: Currently experimenting on github status badge https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge

metaclips: @etorreborre I added a workflow that queries the latest commit of the use-case directory and the documentation repository and compares them against a [registry on the artifacts repository](https://github.com/build-trust/ockam-artifacts/blob/docker/docs_test_data.json), if the registry doesn't have stored the latest commits, it'll alert us on the development channel

etorreborre (approved):

PR #7387, opened 2024-01-08
feat(rust): make authority issued credentials TTL configurable

polvorin:
CREDENTIAL_TTL_SECS env variable can be used to specify the TTL of the credential issued by the authority. If not present, a hardcoded default one is used, as has been done so far.

DO NOT MERGE AS-IS.   It changes the API for credential_issuer (instantiation),  that is part of examples/docs.      Reading the env variable from credential_issuer' directly is not possible due to no-std on the create where it's located.


SanjoDeundiak (approved):

SanjoDeundiak:
Changing examples should not prevent us from merging to develop. However we should make sure we update examples before the release

PR #7388, opened 2024-01-09
feat(elixir): refresh credentials msgs are accepted

polvorin:
there were some difference between rust and elixir on the serialization of refresh credentials message.  Update to match rust encoding, and add some tests cases.


ckhrysze (approved):
Looks good to me!

metaclips (approved):

PR #7412, opened 2024-01-12
feat(rust): unify creation and retry connection for portal and relay

davide-baldo:


davide-baldo:
it was split into https://github.com/build-trust/ockam/pull/7483 and https://github.com/build-trust/ockam/pull/7481

PR #7414, opened 2024-01-15
fix(elixir): correct secure channel close signaling

polvorin:
on some circustances it was using a very old approach, that don't exist on rust side anymore.  Replace with the close message available on the secure channel protocol.
Also fix the encoding of :enum variants that have arguments, needed for compatibility with rust' minicbor encoding.


ckhrysze (approved):
Simple and tested, makes sense to me.

SanjoDeundiak (approved):

PR #7427, opened 2024-01-18
chore: applied clippy fixes of rust 1.74.1

davide-baldo:


etorreborre (dismissed):

metaclips:
@davide-baldo can we also update the [version of rust in this file](https://github.com/build-trust/ockam/blob/develop/rust-toolchain.toml)?

davide-baldo:
@metaclips Done!

davide-baldo:
@metaclips Any idea on why it's not finding any version?
```error: No stable 1.74.* is available```

metaclips (approved):

PR #7437, opened 2024-01-22
refactor(rust): change parsing strategy and config structure of `run` command

adrianbenavides:
⚠️ You can **check config examples** [here](https://github.com/build-trust/ockam/pull/7437/files#diff-9328db86d0a329154760c69c4f71eddd69e54ee9c8ecceea2838b4d2f1d3df0c).

This PR rewrites the config parser used in the `ockam run` command to have more flexibility and more control before actually running the commands:
- The config file is flat now, meaning there is no hierarchy between different resources (see the example below).
- The parser only knows the type of resources beforehand (projects, nodes, policies), but it doesn’t expect a specific set of arguments for each type. That is, the parser just sees key/value pairs. This gives us more flexibility, as we support all the commands’ arguments out of the box without having an explicit 1:1 match between the parser and the commands.
- After this initial parsing, the raw arguments are parsed into clap command instances, so that we have a typed struct with all the arguments populated (default values or given through the config file). With these objects, we can now do whatever checks we want before running the command.
- Additionally, the env vars are now resolved, the commands are run in the same process as the “run” command (no more sub-shelling), and the sidecar commands that were using the old config files have been updated.


implementations/rust/ockam/ockam_command/src/run/parser/mod.rs
```diff
@@ -0,0 +1,339 @@
+mod nodes;
```
etorreborre, line 1: Maybe this is only a personal preference but I'd rather see no "domain" code in a `mod` file, and instead have a `config.rs` file + a `resources.rs` file inside the `parser` folder.

implementations/rust/ockam/ockam_command/src/run/runner.rs
```diff
@@ -0,0 +1,51 @@
+use crate::run::parser::{ArgsToCommands, Config};
+use crate::CommandGlobalOpts;
+
+use miette::Result;
+use ockam_node::Context;
+
+pub struct ConfigRunner;
+
+impl ConfigRunner {
+    pub async fn go(ctx: &Context, opts: CommandGlobalOpts, contents: &str) -> Result<()> {
```
etorreborre: `run_commands` would be more explicit than `go`

implementations/rust/ockam/ockam_command/src/run/runner.rs
```diff
@@ -0,0 +1,51 @@
+use crate::run::parser::{ArgsToCommands, Config};
+use crate::CommandGlobalOpts;
+
+use miette::Result;
+use ockam_node::Context;
+
+pub struct ConfigRunner;
+
+impl ConfigRunner {
```
etorreborre, line 18: A general comment explaining the overall strategy for running commands described in files would be great here:

 - we're parsing some yaml content, where the keys are expected to be resource + command names (with an example of how `ockam node create` would appear in the file)
 - we're parsing that yaml content as a command as if it was read from the command line
 - we're executing all the commands in the file in the proper order (nodes are created before inlets/outlets for example)

etorreborre (dismissed):
Excellent. I just left some suggestions.

implementations/rust/ockam/ockam_command/src/run/parser/mod.rs
adrianbenavides: Thanks for the feedback 🙏  I'll leave this one for the follow-up PR because I'm making a lot of changes here. I'll address the other two comments in this PR.

implementations/rust/ockam/ockam_command/src/run/runner.rs
adrianbenavides: I've added some details, but I'll extend them in the next PR, as I'm adding some changes that will need an update of the docs.

etorreborre (dismissed):

SanjoDeundiak (approved):

Issue #7053, opened 2023-12-06
Use a minimal schema for separate vaults


etorreborre:
When secrets are created for the default vault they are persisted inside the `signing_secret` and `x25519_secret` tables. Those tables are currently created inside the same database file as the rest of the persisted data (identities, nodes, projects, etc...).

However when we create a separate named vault with `ockam vault create my_vault` a new database file is created. 
Then all the migrations are run on this database file and its ends up containing a lot more tables than necessary.

The purpose of this issue is to configure the [database migration](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs#L103) and schema declaration so that we only create the necessary tables for a new Vault database. 

One difficulty might be to avoid declaring twice the schema of the `secrets` tables. One possibility is to add a new migration file called `<timestamp>_vault.sql` with `create table if not exists` instructions containing the `secrets` tables declarations. Then point to this file when creating a database for a vault.

SanjoDeundiak:
I think it's not only about Vault. There are multiple parts of the system that would like to have a separate db with required tables only

etorreborre:
I suppose that the same design will apply when we will have to do it. The question is mostly about where we keep the migration files and have the ability to use different ones.

PR #7438, opened 2024-01-22
feat(rust): add opentelemetry tracing and logging support

etorreborre:
This PR introduces a way to produce OpenTelemetry traces and logs. This feature is not yet fully operational but this PR allows to us to start experimenting with such a support.

The following sections describe the main modifications.

## OpenTelemetry tracing

The tracing configuration has been extended with additional layers in order to send spans and log records to an [OpenTelemetry collector](https://opentelemetry.io/docs/collector).

Locally an OpenTelemetry collector can be started by:

 - Checking out the project `https://github.com/open-telemetry/opentelemetry-rust.git`
 - Running the command `docker run -v $(pwd)/collector.yaml:/etc/otelcol-contrib/config.yaml -p 4317:4317 otel/opentelemetry-collector-contrib:latest` with the appropriate `collector.yaml` file.

A minimal configuration for the `collector.yaml` file is:
```
receivers:
  otlp:
    protocols:
      grpc:

exporters:
  debug:
    verbosity: detailed

service:
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [debug]
    logs:
       receivers: [otlp]
       exporters: [debug]
```
This allows traces to appear in the CLI. If you want to export traces to Honeycomb, please [refer to the corresponding documentation](https://docs.honeycomb.io/getting-data-in/otel-collector/).

### Command names

In order to produce spans, a root span has been added to the `run` function inside the `ockam_command` crate.
Every time a command is executed, a new trace is started with the command name as the root span name.
Many files are modified in the PR, only to give a textual name to all our commands.

### API functions

Some functions have been instrumented using the [`#instrument` macro](https://docs.rs/tracing/latest/tracing/attr.instrument.html), but this PR does not cover the entirety of a node API.

## Trace context propagation

We want to produce distributed traces, across different components:

 - The CLI
 - A background node
 - The Controller
 - A project node
 - An authority node

In order to do this we need to send, on every remote call, some tracing context data. This PR implements some support for this:

 - During node creation, so that the command `ockam node create` starting a background node can have tracing data from the original `ockam node create` command started by a user. (Note: there is currently a race condition and some spans might be missing)
 - During a RPC call, using the node RPC support. For example when getting the status of a background node

There is no tracing context propagation yet for the Ockam protocol itself.

## Initialization

The initialization of a Tracer requires an async runtime. In order to configure tracing even before an internal node `Context` is created, I had to change access to the tokio runtime and change all the `node_rpc` functions to accept a runtime parameter. This is also why many of the `ockam_command` files had to be touched in this PR.

## Journeys

The concept of a `HostJourney` and a `ProjectJourney` have been introduced. Those are pseudo-traces that we generate with major events for a given host, or a given project:

 - When a node has been created
 - When a user has just enrolled
 - When a TCP outlet has been created
 - etc...

The `Host` journey collects events for a given host, and has a random trace id, while the `Project` journey collects events for a given project and uses a trace id derived from the project id.

There are pros and cons to this approach:

 - pros: we can easily visualize the main commands (and possibly the main errors) happening for a given installation or project

image

 - cons: those are not real traces and will hit a size limit for a given trace (32000 spans for Honeycomb). This is not necessarily a problem for a Host journey because we could roll another trace but this could be a problem for a project journey since there might not be a way to drop old spans from a given trace.

This approach is still experimental and might be replaced in the future. A better option would be to have better querying and visualization tools across several traces. At the moment such a query in Honeycomb is not fantastic:
image
For example, it is not possible to sort events by their application timestamp.

### Application database

In order to store data which is not delete when running `ockam reset`, an additional database has been added.
That database is initialized with its own migration file (see `ApplicationMigration`). This opens the door to creating additional Vault databases with their own migration file, [instead of creating the full node database schema](https://github.com/build-trust/ockam/issues/7053).

## Standard attributes

The attribute names set on spans and log records are generally used to query events, and possibly to set metrics and alerts. One consequence is that we need to make sure that the names we use are stable. This PR follows the [OpenTelemetry semantic conventions](https://opentelemetry.io/docs/concepts/semantic-conventions/) where possible and defines Ockam-specific constants otherwise.

## Testing traces / logs export

The test `test_log_and_traces` shows how we can use some in-memory `SpansExporter` and `LogsExporter` to capture the spans and logs that are created when we run a command.

This test support is also used to make sure we create the right events for a host/project journey.

## Environment variables

There are now more ways to configure logging and tracing for an Ockam application:

 - `OCKAM_LOG_LEVEL` sets the log level to apply for both logging and tracing (there is no distinction at the moment)
 - `OCKAM_LOGGING` must be set (to any value) in order to enable logging
 - `OCKAM_TRACING` must be set (to any value) in order to enable tracing
 - `OTEL_EXPORTER_OTLP_ENDPOINT` must be set to the URL of the OpenTelemetry collector (`http://localhost:4317` by default)
 - `OTEL_EXPORTER_OTLP_HEADERS` if the OTLP endpoint is directly set to the Honeycomb API. Then that header can be used to set the Honeycomb API key

The old `OCKAM_LOG` environment variable is now deprecated, because it controls both the level filtering that is used for logs and traces _and_ if tracing/logging is on. That env. variable is still available but does not set the tracing on if `OCKAM_TRACING` is not set.


implementations/rust/ockam/ockam_api/src/session/mod.rs
```diff
@@ -78,7 +78,6 @@ impl Medic {
     async fn go(mut self, ctx: Context, mut rx: mpsc::Receiver<Message>) {
         let ctx = Arc::new(ctx);
         loop {
-            log::trace!("check sessions");
```
etorreborre, line 81: I removed this trace because it was very noisy.

implementations/rust/ockam/ockam_api/src/logs/setup.rs
```diff
@@ -0,0 +1,492 @@
...
+        // sending them to an OpenTelemetry collector using gRPC
+        let span_exporter = Executor::execute_future(async move {
+            opentelemetry_otlp::new_exporter()
+                .tonic()
+                .with_endpoint(get_tracing_endpoint())
+                .with_metadata(get_otlp_headers())
+                .build_span_exporter()
+                .expect("failed to create the span exporter")
+        })
+        .expect("can't create a span exporter");
```
adrianbenavides: Is there any side effect of creating the exporters if they end up not being used when `!tracing_configuration.is_enabled()`?

implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
```diff
@@ -0,0 +1,286 @@
...
+fn log_level(default_log_level: LevelFilter) -> LevelFilter {
+    let log_level = match get_env::<String>("OCKAM_LOG_LEVEL").unwrap_or(None) {
+        Some(s) => LevelFilter::from_str(&s).unwrap_or(default_log_level),
+        None => default_log_level,
+    };
+
+    // If we end-up with a log level that is not defined but tracing is on
+    // then we still need to provide a LevelFilter, since since parameter is used for both logs and traces
+    if tracing_enabled() == TracingEnabled::On && log_level == LevelFilter::OFF {
+        LevelFilter::TRACE
```
adrianbenavides, line 241: Maybe a stupid question but what happens with `trace!` logs that could contain sensitive information? Are they conceptually different from span traces? Is there a way to differentiate between them or should we review all the current `trace!` logs that we have in the codebase?

implementations/rust/ockam/ockam_api/src/logs/setup.rs
```diff
@@ -0,0 +1,492 @@
...
+    pub fn force_flush(&self) {
+        self.logger_provider.force_flush();
+        self.tracer_provider.force_flush();
+    }
+
+    pub fn shutdown(&self) {
+        self.force_flush();
+        global::shutdown_tracer_provider();
+        global::shutdown_logger_provider();
+    }
```
adrianbenavides, line 307: I've seen `flush` is used when shutting down a foreground node. Should this be used explicitly at the end of any command execution?

adrianbenavides (commented):
Awesome, pretty excited about this!

implementations/rust/ockam/ockam_api/src/logs/setup.rs
etorreborre: You're right, that should be inside the `if`.

implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
etorreborre: The `trace` logs appear as "events" on spans, like any log message, if `OCKAM_LOGGING` is enabled. For example 
image
This means that we indeed have to review if there is sensitive information.

implementations/rust/ockam/ockam_api/src/logs/setup.rs
etorreborre: `flush` is otherwise done [via the `TracingGuard` at the end of `run`](https://github.com/build-trust/ockam/pull/7438/files#diff-f5a30aef85b96ee0c58c26d9c22690c484ea212154879d380cae8bd4b27ce018R455). I have also added an explicit `global::shutdown_tracer_provider();` at the end of that function but I don't think it is necessary. This is something I can revisit with my next PR.

etorreborre:
@adrianbenavides I rebased and did the change that you suggested (not creating unnecessary exporters).

implementations/rust/ockam/ockam_api/src/logs/setup.rs
adrianbenavides: OK got it, I didn't see that piece of code, thanks.

adrianbenavides (approved):

PR #7452, opened 2024-01-23
feat(rust): improve ockam project ticket, ockam project enroll ux out…

nazmulidris:
Fixes: https://github.com/build-trust/codebase/issues/708


implementations/rust/ockam/ockam_command/src/docs.rs
```diff
@@ -128,7 +128,11 @@ impl FencedCodeBlockHighlighter<'_> {
         }
     }
 
+    // TODO: fix the fenced code block highlighter (currently disabled), as it does not work on macOS or Linux
```
davide-baldo: completely remove the code rather than leave dead code, you can easily recover it anyway.

davide-baldo (dismissed):

implementations/rust/ockam/ockam_command/src/docs.rs
nazmulidris: Unfortunately, there are lots of depenendencies which will have to be re-created if I remove all the dead code. Instead:
1) I've removed the call to the broken function here. 
2) And put an explicit TODO at the (future) call site, 
3) and at the broken function definition (below). 

This will also facilitate the community making a fix for this, when I create an issue for this.

implementations/rust/ockam/ockam_command/tests/bats/authority.bats
```diff
@@ -86,7 +86,7 @@ EOF
   run_success bash -c "$OCKAM project import --project-file $OCKAM_HOME/project.json"
 
   # m1 is a member (its on the set of pre-trusted identifiers) so it can get it's own credential
-  run_success "$OCKAM" project enroll --identity m1
+  run_success "$OCKAM" project enroll --identity m1 --output plain
```
adrianbenavides: This change is not necessary, since we are not adding any new output format for this command. Removing this change will also make explicit that we are not forcing anyone to use the `plain` format in scripts

implementations/rust/ockam/ockam_command/src/project/static/enroll/after_long_help.txt
```diff
@@ -1,7 +1,17 @@
...
-$ ticket=$(ockam project ticket --attribute component=user)
+$ TICKET=$(ockam project ticket --attribute component=user)
+
+# From the user machine, enroll the local identity to the project using the ticket
+$ ockam project enroll $TICKET --identity control_identity
+
+# 2) Use file enrollment ticket:
+
+# From the admin machine, generate an enrollment ticket, save to a file
+$ ockam project ticket --attribute component=user > ticket.txt
```
adrianbenavides: The convention we've been using is to name enrollment tickets like `$NAME.ticket`

implementations/rust/ockam/ockam_command/src/project/enroll.rs
```diff
@@ -109,18 +117,37 @@ impl EnrollCommand {
...
+            "The identity has the following credential in this project",
+            "The attributes below are attested by the project's membership authority"
+        ))?;
+
+        // Output the credential and purpose keys to stdout.
         opts.terminal
             .clone()
             .stdout()
-            .plain(CredentialAndPurposeKeyDisplay(credential))
+            .plain(CredentialAndPurposeKeyDisplay(credential.clone()))
```
adrianbenavides: The `terminal.clone()` and the `credential.clone()` are not necessary. I expected `clippy` to complain about this 🤔

implementations/rust/ockam/ockam_command/src/project/enroll.rs
```diff
@@ -109,18 +117,37 @@ impl EnrollCommand {
...
+                .await?;
+            project.name()
+        };
 
+        // Display success message to stderr.
+        opts.terminal.write_line(&fmt_ok!(
+            "Successfully enrolled identity to the {} project. {}. {}.",
+            color_primary(project_name),
+            "The identity has the following credential in this project",
+            "The attributes below are attested by the project's membership authority"
```
adrianbenavides: Should we print these two messages in separate lines?

implementations/rust/ockam/ockam_command/src/docs.rs
davide-baldo: nice!

implementations/rust/ockam/ockam_command/src/project/enroll.rs
nazmulidris: great idea, done

implementations/rust/ockam/ockam_command/src/project/enroll.rs
nazmulidris: i added some code earlier (which i then removed) ... these clones came about from that effort. good catch. removing.

implementations/rust/ockam/ockam_command/src/project/static/enroll/after_long_help.txt
nazmulidris: done

implementations/rust/ockam/ockam_command/tests/bats/authority.bats
nazmulidris: done

davide-baldo (approved):

PR #7456, opened 2024-01-24
feat(rust): add support for more commands to `run` command 

adrianbenavides:
- Added support for `project create`, `vault create` and `identity create` commands
- Added support for boolean and int types on argument values to express correctly all the commands' arguments
- Refactored some data types of the parser to support more ways of expressing resources in the config file (single name, list or map). The tests of the `resources` module contain examples of the different ways to express resources.
- Renamed some structs and added more docs to the main data types


etorreborre (approved):
🏇

PR #7458, opened 2024-01-25
feat(rust): allow running `reset` command even if the database is in an invalid state

adrianbenavides:
The database is always loaded before running a command. When we switch to a previous/older version with a different database schema (with differences in the migrations), we have to manually remove the database because we couldn't run the `reset` command.

This PR adds new logic in the command crate that allows us to remove the local state when running the `reset` command, even if it fails to load the database.

It also changes the output that we show when we detect the corrupted state:

image

## How to test

```
ockam node create

# Modify one of the migrations at implementations/rust/ockam/ockam_node/src/storage/database/migrations
ockam node create # this should fail and show the above error

ockam reset # this should work and reset the state

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
@@ -89,6 +89,12 @@ impl CliState {
         self.delete()
     }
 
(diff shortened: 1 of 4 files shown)


etorreborre (approved):
Nice!

PR #7459, opened 2024-01-25
Simplify the ockam test macro and make arguments tests more robust

etorreborre:
This PR:

 - Makes sure that the test `Context` provided by the `ockam::test` macro is always closed, without the user having to do it.
 
 - Returns successfully when a command is just testing the parsing of arguments. This avoids some potential later issues related to the initialization of `CommandGlobalOpts`


adrianbenavides (approved):
🚀

PR #7460, opened 2024-01-25
Send command errors as trace events

etorreborre:
This PR makes sure that we create a span, and export it, each time a command returns an error.

The main changes are in `journey.rs` and in the (ex) `node_rpc` execution which does not exit in case of an error anymore.
Instead, results are propagated up to the `main` function which then exits with an error code.

Since this PR required a change throughout all the commands to change the signature of the `node_rpc` function, I took the opportunity to make all that code more consistent. All the calls now look like:
```rust
pub fn run(self, opts: CommandGlobalOpts) -> miette::Result<()> {
    async_cmd(&self.name(), opts.clone(), |ctx| async move {
        self.async_run(&ctx, opts).await
    })
}
```

Where `async_run` is a function of the command we are implementing.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/tcp/outlet/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/tcp/outlet/create.rs
@@ -22,7 +22,7 @@ use crate::node::util::initialize_default_node;
 use crate::policy::{add_default_project_policy, has_policy};
 use crate::tcp::util::alias_parser;
 use crate::terminal::OckamColor;
-use crate::util::node_rpc;
+use crate::util::async_cmd;
 use crate::util::parsers::socket_addr_parser;
 use crate::{display_parse_logs, fmt_log};
 use crate::{docs, fmt_ok, CommandGlobalOpts};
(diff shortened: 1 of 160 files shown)


adrianbenavides:
@etorreborre I did a rebase fixing the conflicts that existed right before merging sanjo's trust context PR.

etorreborre:
Thanks @adrianbenavides. Unfortunately I saw your comment too late, and did the full rebase anyway...

implementations/rust/ockam/ockam_command/src/tcp/outlet/create.rs
```diff
@@ -142,7 +148,7 @@ pub fn default_from_addr() -> String {
     "/service/outlet".to_string()
 }
 
-pub async fn run_impl(
+pub async fn async_run(
     ctx: Context,
     (opts, cmd): (CommandGlobalOpts, CreateCommand),
 ) -> miette::Result<()> {
```
adrianbenavides: I think most of these functions (the old `run_impl` defined outside of the command struct) are now unused. Clippy doesn't complain because they are public, but we can do the cleanup after landing this PR.

implementations/rust/ockam/ockam_command/src/trust_context/list.rs
```diff
@@ -0,0 +1,45 @@
+use clap::Args;
```
adrianbenavides: All the `trust-context` commands were removed at https://github.com/build-trust/ockam/pull/7032

adrianbenavides (commented):
Awesome! I love how the commands' functions look now 🏄

etorreborre:
Thanks @adrianbenavides, there were indeed some leftovers for trust context and credentials from the rebase.
I also remove the redundant function in `outlet/create` and I don't think there are any more of the sort.

adrianbenavides (approved):

PR #7464, opened 2024-01-26
test(rust): run tests for logging / tracing as integration tests

etorreborre:
Since the setup of spans / logs exporters is done globally, there is no (easy) way to run different tests exporting different spans/logs concurrently. So I move those tests to integration tests which are executed sequentially, as far as I understand.


implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
```diff
@@ -83,22 +83,27 @@ impl LoggingConfiguration {
...
+                    .with_default_directive(self.level().into())
+                    .parse_lossy(
+                        crates
+                            .iter()
+                            .map(|c| format!("{c}={}", self.level()))
+                            .collect::<Vec<_>>()
+                            .join(","),
+                    )
+            }
+            None => EnvFilter::default().add_directive(LevelFilter::INFO.into()),
```
adrianbenavides: Shouldn't this use the `self.level()` function instead of `LevelFilter::INFO`?

etorreborre: Yes, indeed. I fixed it, thanks.

adrianbenavides (dismissed):

adrianbenavides (approved):

PR #7465, opened 2024-01-26
feat(rust): restart a project journey if project is deleted

etorreborre:
This way we can properly track events even when several projects are being tried out.


adrianbenavides (approved):
🚂

Issue #4789, opened 2023-04-25
Auto-reconnect `replacer` code is deduplicated and easy to read for Inlet and Relay


davide-baldo:
### Current behavior

Inside `ockam_api` crate, an auto-reconnect mechanism is present, currently named `Session`.
The idea is simple: the route is pinged every few seconds, if there is no answer to the ping the route gets “replaced”.
For further description of the initial implementation: https://github.com/build-trust/ockam/pull/3497

There are 2 main pain points with the current implementation: the first is the `Replacer` [object](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/session/sessions.rs#L21) is an asynchronous lambda and since it's not a first citizen the creation is a bit [convoluted](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs#L471).
The second is the need to duplicate the code for the first connection and for the replacement.

### Desired behavior

Instead of async lambdas `Replacer` object, we can use a simpler async trait, reducing the complexity in creating and using it.
Instead of duplicating the code, we can change the semantic a bit and have the trait act both as an initializer **and** as a replacer.

0xkelvin:
hello @davide-baldo can i take this

etorreborre:
Hi @0xkelvin please do.

0xkelvin:
i hope this one still valid, i can work on for now
i was busy with my own stuff for awhile

PR #7481, opened 2024-01-29
Unify creation and retry connection for portal and relay

davide-baldo:
Implements https://github.com/build-trust/ockam/issues/4789

The Medic callback was refactored into a more usable trait called `Replacer`. Its responsibility is to spawn and keep track of resources of relays and portals (or possible other elements in the future).

The differentiation between "first connection code" and later attempts was removed, less repetitive code and more tested. This should result in more robust implementations.
Both relays and portals didn't handle the concept of connection being down, every field was always returned even if it was old/stale data of a closed connection. This was addressed, and the answers now have optional fields when the information is not available due to the connection being down.
In portal creation there is an extra flag called `--no-validation` was added. This flag skips the synchronous validation, and allows the operator to create an inlet even if the route is not fully reachable.


implementations/rust/ockam/ockam_command/src/relay/create.rs
```diff
@@ -50,6 +49,10 @@ pub struct CreateCommand {
     /// Authorized identity for secure channel connection
     #[arg(long, id = "AUTHORIZED")]
     pub authorized: Option<Identifier>,
+
+    /// Relay address to use. By default, inherits the relay name.
+    #[arg(long)]
+    relay_address: Option<String>,
```
adrianbenavides, line 55: In what scenario is this useful? IIUC, this is to define the name of the relay that will be created while the alias is only used in the node to store it in the registry?

adrianbenavides (commented):
I've done a first pass over the `ockam_command` crate and a little around how the inlets are created. There will be some conflicts with sanjo's policy PR, but they should be easy to fix.

implementations/rust/ockam/ockam_api/src/session/sessions.rs
```diff
@@ -79,57 +154,86 @@ impl Session {
     ///
     /// # Arguments
     ///
-    /// * `ping_route` - The route to send pings to
     /// * `key` - The key to identify the session, usually adding the kind of the session
     ///           with the key used within the service registry is the way to gos
```
adrianbenavides: This is outdated now right?

implementations/rust/ockam/ockam_api/src/session/sessions.rs
```diff
@@ -79,57 +154,86 @@ impl Session {
...
-    pub fn new(ping_route: Route, key: String) -> Self {
+    pub fn new(replace: impl SessionReplacer) -> Self {
```
adrianbenavides: Typo? `replace` -> `replacer`

implementations/rust/ockam/ockam_api/src/nodes/service/relay.rs
```diff
@@ -122,81 +122,89 @@ impl NodeManager {
...
+                    }
+                })?;
+
+        if self.registry.relays.contains_key(&alias).await {
+            let message = format!("A relay with the name '{alias}' already exists");
+            return Err(ockam_core::Error::new(
+                Origin::Node,
+                Kind::AlreadyExists,
+                message,
+            ));
```
adrianbenavides: Why is the relay created before checking if it already exists? In the `create_inlet` function it's doing the check first.

implementations/rust/ockam/ockam_command/src/relay/create.rs
davide-baldo: By default the alias defines both the name of the relay and the address of the relay.
Normally, you would only use alias, but if you want to use the same address for multiple projects relaying toward the same node, then you need relay_address.
Another simple use case is to keep compatibility with a previously named address while renaming the relay itself.

implementations/rust/ockam/ockam_api/src/session/sessions.rs
davide-baldo: yes! deleting...

implementations/rust/ockam/ockam_api/src/nodes/service/relay.rs
davide-baldo: no particular reason, it makes sense to delay the creation after full validation

adrianbenavides (approved):

PR #7488, opened 2024-01-30
feat(rust): improve credentials management

SanjoDeundiak:
Currently, Secure Channel Encryptor is responsible for triggering its CredentialRetriever refresh, which is suboptimal. In this PR RemoteCredentialRetriever will run on its own and refresh the credential in the background according to the expiration date of the previous one. Upon refreshing, RemoteCredentialRetriever will notify all EncryptorWorkers, which will trigger them to send the new IdentityPayload to the other side, therefore keeping all Secure Channels with the most recent credentials all the time


implementations/rust/ockam/ockam_app_lib/src/incoming_services/commands.rs
```diff
@@ -169,7 +169,26 @@ impl AppState {
 
         // Recreate the node using the project name
         debug!(node = %local_node_name, "Creating node to host TCP inlet");
-        let _ = self.delete_background_node(&local_node_name).await;
+
+        // We may be creating the node, in this case you don't need to back up logs and delete the node
```
etorreborre, line 173: This comment is not clear to me. Do you mean that we should rather detect when there's not actual node to delete and skip backing-up the logs + deleting the node? In that case why not test if the node exists?

implementations/rust/ockam/ockam_command/src/node/start.rs
```diff
@@ -174,6 +174,7 @@ async fn run_node(
         None,          // No project information available
         None,          // No Authority Identity
         None,          // No Authority Address
+        false,         // No cache credentials
```
etorreborre, line 177: I'm starting to think that a builder pattern would make the code self-explanatory here.

implementations/rust/ockam/ockam_command/tests/bats/trust.bats
```diff
@@ -45,10 +45,10 @@ teardown() {
   authority_identity=$($OCKAM identity show authority --full --encoding hex)
 
   # Create a node for alice that trustt authority as a credential authority
```
etorreborre, line 47: I know it's not code from this PR but can you please fix the typo `trustt`/`trusts`? (Same issue 2 lines below).

implementations/rust/ockam/ockam_node/src/context/context_lifecycle.rs
```diff
@@ -115,8 +116,38 @@ impl Context {
...
-        tokio::time::sleep(dur).await;
+    pub async fn sleep(&self, duration: Duration) {
+        tokio::time::sleep(duration).await;
+    }
+
+    /// Utility function to sleep tasks for long periods of time (seconds precision)
+    /// Difference between this and `sleep` is that this sleeps in 1 second intervals and recalculates time left,
+    /// which account for the time the device was in sleep state
+    #[doc(hidden)]
+    pub async fn sleep_long_until(&self, deadline_timestamp_seconds: u64) {
```
etorreborre, line 127: While we're here I would move the `TimestampInSeconds` data type from `ockam_identity` to `ockam_node`:

-  It's not a concern specifically tied to Identity
- This way we can give a clearer type to the `deadline_timestamp` parameter

etorreborre (approved):
This all seem to be working but I think that it could be good to attempt a refactoring where we separate:

 - The notion of retrieving credentials: can be cached, in memory, remote
 - The notion of refreshing credentials, and then subscribing to refresh notifications

The fact that the `subscribe/unsusbcribe` methods have dummy values in 2 implementations out of 3 is a sign of this. I think that this could also remove the need for an additional layer of `CredentialRetriever`_Creator_.

I can give it a go later on to see how it goes.

implementations/rust/ockam/ockam_app_lib/src/incoming_services/commands.rs
SanjoDeundiak: Right. Unfortunately, I didn't spend time adding that check which wasn't there in the first place

implementations/rust/ockam/ockam_command/src/node/start.rs
SanjoDeundiak: Yeah, we need to do it eventually

PR #7492, opened 2024-01-30
fix(rust): fixes for okta authenticator

polvorin:
* call the correct function from the cmd cli to enroll with okta
* fix the okta worker to store the new enrolled member on the member' table

Note:  the `AuthorityMember` has an `added_by` mandatory field.    There is no enroller here, so not sure what to put there (and didn't want to go to modify the schema to allow None there).  I completed it with the same member identifier, which is probably not what we want (maybe putting the authority' identity itself?).


SanjoDeundiak (approved):

mrinalwadhwa:
@polvorin could you change the commit message to the imperative style.
`fix X` `change Y` etc. instead of `fixes for X`

PR #7493, opened 2024-01-31
fix(rust): fix identity attributes expiration

SanjoDeundiak:


etorreborre (commented):
This obviously works but I would do a bit of encapsulation to put the expiration logic out of the repository (and use this struct in `DirectAuthenticator` for example):
```rust
struct IdentityAttributes {
    repository: Arc<dyn IdentityAttributesRepository>
}

impl IdentityAttributes {
   /// takes care of setting `now`
   pub fn save_attributes(
        attrs: BTreeMap<Vec<u8>, Vec<u8>>,
        expires: Option<TimestampInSeconds>,
        attested_by: Option<Identifier>) -> Result<()>)

    /// Takes care of calling repository.delete_expired_attributes(now) before getting
    /// attributes from the database
    pub fn get_attributes(&self, identifier: Identifier) -> Result<AttributesEntry> {
       ...
    }
}
```
This way there is less risk of messing up with the passing of the "now" datetime. The control of expiration is only ever done in one place.

Additionally, I think we should stop using `TimestampInSeconds` to represent a `DateTime`. Because we end up manipulating entities which are not the same. An expiration date is a `DateTime` while a TTL is a `Duration`, but we represent it the same, as a `TimestampInSeconds`.

implementations/rust/ockam/ockam_identity/src/identities/storage/identity_attributes_repository_sql.rs
```diff
@@ -167,34 +158,22 @@ mod tests {
             .await?;
         assert_eq!(result, Some(attributes1.clone()));
 
-        let result = repository.list_attributes_by_identifier().await?;
```
SanjoDeundiak, line 170: Let's check identifier2 attributes few lines above as well

polvorin (approved):

PR #7498, opened 2024-01-31
refactor(rust): separate transport messages from local messages

etorreborre:
This PR extracts `TransportMessages` from `LocalMessages` so that we precisely know which message type is used and when:

 - `LocalMessages` are used between workers inside a node
 - `TransportMessages` can be serialized / deserialized and they are communicated between nodes

This PR prepares the work for propagating the tracing context (and setting it on `TransportMessages`).


implementations/rust/ockam/ockam_core/src/routing/message/local_message.rs
```diff
@@ -1,116 +1,210 @@
...
+/// - replace_front_onward_route: replace the first address of the onward route with another address
+/// - push_front_onward_route: add an address at the front of the onward route
+/// - prepend_front_onward_route: prepend a whole route at the front of the onward route
+/// - set_onward_route: set a new route for the onward route
+///
+/// There are similar functions for return routes. All modification functions can be composed. For example:
+///
+/// self.pop_front_onward_route()?.prepend_front_return_route(&new_route)
 ///
 #[derive(Serialize, Deserialize, Debug, Clone, Hash, Ord, PartialOrd, Eq, PartialEq, Message)]
```
SanjoDeundiak, line 38: I wonder what if we switch to CBOR here. Will it... just work after modifying a dozen lines of code? Following should be updated I guess:
1. Serialization functions in `ockam_node`code
2. Untyped workers that deserialize messages themselves

implementations/rust/ockam/ockam_core/src/routing/message/local_message.rs
```diff
@@ -1,116 +1,210 @@
...
+    }
+
+    /// Prepend a route to the return route
+    pub fn prepend_front_return_route(mut self, route: &Route) -> Self {
+        self.return_route.modify().prepend_route(route.clone());
+        self
+    }
+
+    /// Remove the first address on the onward route and push another address on the return route
+    pub fn forward(self, address: &Address) -> Result<Self> {
```
SanjoDeundiak: Would you consider renaming it? I expect `forward` to actually forward message, while here we only update addresses

implementations/rust/ockam/ockam_core/src/routing/message/local_message.rs
```diff
@@ -1,116 +1,210 @@
...
+        self.local_info.clear()
     }
 
-    /// Replace all [`LocalInfo`] entries matching the type identifier
-    /// of the given `LocalInfo` with itself.
-    pub fn replace_local_info(&mut self, local_info: LocalInfo) {
-        self.clear_local_info(local_info.type_identifier());
-        self.local_info.push(local_info)
+    /// Create a [`LocalMessage`] from a decoded [`TransportMessage`]
+    pub fn from_transport_message(transport_message: TransportMessage) -> LocalMessage {
```
SanjoDeundiak, line 181: Should this be `From` trait?

implementations/rust/ockam/ockam_core/src/routing/message/local_message.rs
```diff
@@ -1,116 +1,210 @@
...
+            vec![],
+        )
     }
 
-    /// Clear all [`LocalInfo`] entries with the given type identifier.
-    pub fn clear_local_info(&mut self, type_identifier: &str) {
-        self.local_info
-            .retain(|x| x.type_identifier() != type_identifier)
+    /// Create a [`TransportMessage`] from a [`LocalMessage`]
+    pub fn into_transport_message(self) -> TransportMessage {
```
SanjoDeundiak, line 189: This could be `From` trait as well

implementations/rust/ockam/ockam_core/src/routing/message/local_message.rs
```diff
@@ -1,116 +1,210 @@
...
+        TransportMessage::v1(self.onward_route(), self.return_route(), self.payload())
     }
 }
 
 impl LocalMessage {
     /// Create a new `LocalMessage` from the provided transport message and local information.
-    pub fn new(transport_message: TransportMessage, local_info: Vec<LocalInfo>) -> Self {
+    pub fn new(
```
SanjoDeundiak: Do you think this could be more friendly like this:
```
let msg = LocalMessage::new()
    .with_onward_route(onward_route)
    .with_return_route(return_route)
    .with_payload(payload)
    .with_local_info(local_info);
```

SanjoDeundiak: The main concern here is that you won't confuse onward and return route, since they go one after another to the constructor and have the same type

implementations/rust/ockam/ockam_node/src/context/send_message.rs
```diff
@@ -280,35 +280,44 @@ impl Context {
         Ok(())
     }
 
+    /// Forward a transport message to its next routing destination
+    /// This removes the message first onward address (considered as processed)
+    /// and appends the current Context address to the message return route
+    pub async fn forward(&self, local_msg: LocalMessage) -> Result<()> {
```
SanjoDeundiak: I think we need a clear definitions around `forward` and `send`. Previously:
1. `send` is to create a new message and send from the current address
2. `forward` move an existing message along the route. Also, caller is responsible for moving all the Addresses inside onward and return route

In this PR I'm concerned it's no longer clear what's exactly the logic behind the naming

SanjoDeundiak: Maybe, everything that takes a `LocalMessage` should be `forward` and all the addresses should be manipulated by the caller prior to calling `forward`. Thoughts?

SanjoDeundiak: I'm also concerned that we change the behaviour of the previously existing function with the exact same name. It's easy to break something this way

SanjoDeundiak (dismissed):
Left few comments, otherwise very good change!

implementations/rust/ockam/ockam_core/src/routing/message/local_message.rs
etorreborre: Yes, hopefully that will be very few changes. Why point n.2 ?

implementations/rust/ockam/ockam_core/src/routing/message/local_message.rs
etorreborre: Yes, I was also uncertain about the name. I called it `step_forward` because `step` makes you think about the fact that we `step` the onward_route and `forward` the fact that we do this to forward the message next.

implementations/rust/ockam/ockam_core/src/routing/message/local_message.rs
SanjoDeundiak: Yeah, I guess payload is serialized separately, so should be changed separately

implementations/rust/ockam/ockam_core/src/routing/message/local_message.rs
etorreborre: I'm a bit on the fence with using `From` when there's only one conversion in one direction and / or the other. Because with implicit conversions it's not immediately apparent what are things transformed into. It seems to me that it's better left when there are lots of conversions `X -> Y` and then having many functions `from_x1`, `from_x2`, `from_x3`,... is not that great.

implementations/rust/ockam/ockam_core/src/routing/message/local_message.rs
etorreborre: Great idea.

implementations/rust/ockam/ockam_node/src/context/send_message.rs
etorreborre: I reverted back the naming. I think I am annoyed that we have 2 APIs for sending messages, one with `send` and one with `forward`, especially with a comment that says that `forward` is low-level, yet used in something as simple as a `Hop` worker. I'm going to leave it like it was for now. Hopefully we can come back to this API.

implementations/rust/ockam/ockam_core/src/routing/message/local_message.rs
etorreborre: I'm leaving it for now. Please re-open if you really prefer a `From` instance.

SanjoDeundiak (approved):

PR #7499, opened 2024-01-31
feat(rust): `run` support defining resources without names so they are assigned a random name

adrianbenavides:
- Resources like tcp inlets can now be also defined without names, and a random name will be assigned. Check out [this test](https://github.com/build-trust/ockam/blob/c995923c78f95d0f945ea7e6febc7737653b220a/implementations/rust/ockam/ockam_command/src/run/parser/tcp_inlets.rs#L37) for an example
- Env vars are now resolved as part of a string, e.g. `'(= subject.from-$VAR1 "$VAR2")'` is now supported

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/run/parser/resources.rs
+++ b/implementations/rust/ockam/ockam_command/src/run/parser/resources.rs
@@ -52,7 +52,11 @@ pub struct NamedResources {
 }
 
 impl NamedResources {
-    pub fn into_commands<C, F>(self, get_subcommand: F, name_arg: Option<&str>) -> Result<Vec<C>>
+    pub fn into_commands<C, F>(
+        self,
+        get_subcommand: F,
+        name_arg_key: Option<&str>,
+    ) -> Result<Vec<C>>
     where
         C: ClapArgs,
         F: Fn(&[String]) -> Result<C>,
@@ -60,7 +64,7 @@ impl NamedResources {
         self.items
             .into_iter()
             .map(|(n, a)| {
-                let mut args = match name_arg {
+                let mut args = match name_arg_key {
                     None => vec![n],
                     Some(arg) => vec![as_keyword_arg(&arg.to_string()), n],
                 };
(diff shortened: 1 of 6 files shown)


implementations/rust/ockam/ockam_command/src/run/parser/resources.rs
```diff
@@ -183,11 +197,10 @@ pub fn as_keyword_arg(k: &ArgKey) -> String {
...
         ArgValue::String(v) => {
-            if v.starts_with('$') {
-                let v = v.trim_start_matches('$');
-                if let Ok(v) = std::env::var(v) {
-                    return v;
-                }
+            if v.contains('$') {
+                return shellexpand::env(v)
+                    .expect("Failed to resolve environment variables")
+                    .to_string();
```
nazmulidris, line 203: Is it ok to panic w/ expect here, or are we supposed to return an error that will be wrapped into some other miette error and reported to the user (bubbling up via `run_config()`)?

implementations/rust/ockam/ockam_command/src/run/parser/resources.rs
```diff
@@ -212,3 +225,22 @@ pub fn parse_cmd_from_args(cmd: &str, args: &[String]) -> Result<OckamSubcommand
 pub trait ArgsToCommands<T> {
     fn into_commands(self) -> Result<Vec<T>>;
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn resolve_env_vars() {
```
nazmulidris, line 234: Also should there be test coverage for this error case (where env var can't be found) here

implementations/rust/ockam/ockam_command/src/run/parser/resources.rs
adrianbenavides: I'd keep it like this for now. This will panic if the env vars are unknown.

I'll improve this in a separate PR.

nazmulidris: Ok sounds good 👍🏽

nazmulidris (approved):

PR #7508, opened 2024-01-31
feat(rust): improve ockam tcp-outlet commands ux output, help, logs, …

nazmulidris:
Fixes: https://github.com/build-trust/codebase/issues/713
Demo video: https://drive.google.com/file/d/1YvSVzeEUhP1lcnu6hqRFqyGgbFKjCNdj/view

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/logs/setup.rs
+++ b/implementations/rust/ockam/ockam_api/src/logs/setup.rs
@@ -89,10 +89,8 @@ impl LoggingTracing {
                 app_name,
             )
         };
-        info!(
-            "tracing initialized. {} {}",
-            logging_configuration, tracing_configuration
-        );
+        info!("tracing initialized");
+        debug!("{:#?} {:#?}", logging_configuration, tracing_configuration);
         result
     }
 
--- a/implementations/rust/ockam/ockam_command/src/tcp/outlet/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/tcp/outlet/create.rs
@@ -10,44 +10,59 @@ use tokio::try_join;
 
 use ockam::Context;
 use ockam_abac::Resource;
-use ockam_api::address::extract_address_value;
 use ockam_api::journeys::{
     JourneyEvent, NODE_NAME, TCP_OUTLET_ALIAS, TCP_OUTLET_AT, TCP_OUTLET_FROM, TCP_OUTLET_TO,
 };
 use ockam_api::nodes::models::portal::{CreateOutlet, OutletStatus};
 use ockam_api::nodes::BackgroundNodeClient;
+use ockam_api::{address::extract_address_value, random_name};
 use ockam_core::api::Request;
 
-use crate::fmt_log;
 use crate::node::util::initialize_default_node;
-use crate::policy::{add_default_project_policy, has_policy};
 use crate::tcp::util::alias_parser;
-use crate::terminal::OckamColor;
 use crate::util::async_cmd;
 use crate::util::parsers::socket_addr_parser;
 use crate::{docs, fmt_ok, CommandGlobalOpts};
+use crate::{
+    fmt_info,
+    policy::{add_default_project_policy, has_policy},
+};
+use crate::{fmt_log, terminal::color_primary};
 
 const AFTER_LONG_HELP: &str = include_str!("./static/create/after_long_help.txt");
+const LONG_ABOUT: &str = include_str!("./static/create/long_about.txt");
 
-/// Create a TCP Outlet
+/// Create a TCP Outlet that runs adjacent to a TCP server
 #[derive(Clone, Debug, Args)]
-#[command(after_long_help = docs::after_help(AFTER_LONG_HELP))]
+#[command(
+    long_about = docs::about(LONG_ABOUT),
+    after_long_help = docs::after_help(AFTER_LONG_HELP)
+)]
 pub struct CreateCommand {
-    /// Node on which to start the tcp outlet.
-    #[arg(long, display_order = 900, id = "NODE_NAME", value_parser = extract_address_value)]
-    pub at: Option<String>,
-
-    /// Address of the tcp outlet.
-    #[arg(long, display_order = 901, id = "OUTLET_ADDRESS", default_value_t = default_from_addr(), value_parser = extract_address_value)]
-    pub from: String,
-
-    /// TCP address to send raw tcp traffic.
-    #[arg(long, display_order = 902, id = "SOCKET_ADDRESS", value_parser = socket_addr_parser)]
+    /// TCP address where your TCP server is running. Your Outlet will send raw TCP traffic to it
+    #[arg(long, display_order = 900, id = "SOCKET_ADDRESS", value_parser = socket_addr_parser)]
     pub to: SocketAddr,
 
-    /// Assign a name to this outlet.
-    #[arg(long, display_order = 900, id = "ALIAS", value_parser = alias_parser)]
-    pub alias: Option<String>,
+    /// Assign a name to your Outlet. This name must be unique. If you don't provide it, a
+    /// random name will be generated. This alias name is the resource name that you will
+    /// need if you declare an access control policy for this Outlet using `ockam policy
+    /// create`
+    #[arg(long, display_order = 901, id = "ALIAS", default_value_t = random_name(), value_parser = alias_parser)]
+    pub alias: String,
+
+    /// Address of your TCP Outlet, which is part of a route that is used in other
+    /// commands. This address must be unique. This address identifies the TCP Outlet
+    /// worker, on the node, on your local machine. Examples are `/service/my-outlet` or
+    /// `my-outlet`. If you don't provide it, `/service/outlet` will be used. You will
+    /// need this address when you create a TCP Inlet (using `ockam tcp-inlet create --to
+    /// <OUTLET_ADDRESS>`)
+    #[arg(long, display_order = 902, id = "OUTLET_ADDRESS", default_value = "/service/outlet", value_parser = extract_address_value)]
+    pub from: String,
+
+    /// Your TCP Outlet will be created on this node. If you don't provide it, the default
+    /// node will be used
+    #[arg(long, display_order = 903, id = "NODE_NAME", value_parser = extract_address_value)]
+    pub at: Option<String>,
 }
 
 impl CreateCommand {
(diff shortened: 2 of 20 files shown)


implementations/rust/ockam/ockam_api/src/logs/setup.rs
```diff
@@ -90,7 +90,7 @@ impl LoggingTracing {
             )
         };
         info!(
-            "tracing initialized. {} {}",
+            "tracing initialized. {:#?} {:#?}",
             logging_configuration, tracing_configuration
```
adrianbenavides: Perhaps we could split this in two to avoid polluting too much the `info` channel:

```
info!("tracing initialized");
debug!(%logging_configuration, %tracing_configuration);
```

implementations/rust/ockam/ockam_command/src/tcp/outlet/create.rs
```diff
@@ -80,28 +88,38 @@ impl CreateCommand {
 
         let is_finished: Mutex<bool> = Mutex::new(false);
 
+        // If no alias is supplied, then create a random one (with a friendly name).
+        let alias = match &self.alias {
+            None => random_name().to_string(),
+            Some(alias) if alias.is_empty() => random_name().to_string(),
+            Some(alias) => alias.clone(),
+        };
```
adrianbenavides: We can avoid this block here removing the `Option` and using clap attributes:

```
#[arg(long, display_order = 901, id = "ALIAS", value_parser = alias_parser, default_value_t = random_name(), hide_default_value = true)]
pub alias: String,
```

implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs
```diff
@@ -9,31 +9,38 @@ use ockam_api::nodes::BackgroundNodeClient;
...
 use crate::util::async_cmd;
-use crate::{color, docs, fmt_ok, CommandGlobalOpts, OckamColor, Terminal, TerminalStream};
+use crate::{docs, fmt_ok, CommandGlobalOpts, Terminal, TerminalStream};
+use crate::{node::NodeOpts, terminal::color_primary};
 
 const AFTER_LONG_HELP: &str = include_str!("./static/delete/after_long_help.txt");
+const LONG_ABOUT: &str = include_str!("./static/delete/long_about.txt");
 
-/// Delete a TCP Outlet
+/// Delete a TCP Outlet.
```
adrianbenavides: Just a heads up: clap trims the stops from docstrings (commands' short docs + commands' args help text)

implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs
```diff
@@ -127,9 +134,9 @@ impl DeleteCommandTui for DeleteTui {
         self.terminal()
             .stdout()
             .plain(fmt_ok!(
-                "Outlet with alias {} on Node {} has been deleted",
-                color!(item_name, OckamColor::PrimaryResource),
-                color!(node_name, OckamColor::PrimaryResource)
+                "Outlet with alias {} on node {} has been deleted",
+                color_primary(item_name),
+                color_primary(&node_name)
```
adrianbenavides, line 139: Something to keep in mind for a separate PR: we should remove the `color!` macro in favor of the `color_x()` functions.

implementations/rust/ockam/ockam_command/src/tcp/outlet/list.rs
```diff
@@ -59,11 +61,21 @@ impl ListCommand {
...
-        let list = opts.terminal.build_list(
-            &outlets.list,
-            &format!("Outlets on Node {}", node.node_name()),
-            &format!("No TCP Outlets found on node {}.", node.node_name()),
-        )?;
+        let list: String = {
+            let empty_message = fmt_info!(
+                "No TCP Outlets found on node {}",
+                color_primary(node.node_name())
+            );
```
adrianbenavides, line 68: To enforce consistent formatting, we could pass a plain string to `build_list` for the `empty_message` argument and always build an `fmt_info` message with it.

implementations/rust/ockam/ockam_command/src/tcp/outlet/list.rs
```diff
@@ -59,11 +61,21 @@ impl ListCommand {
...
+            match outlets.list.is_empty() {
+                true => empty_message,
+                false => opts.terminal.build_list(
+                    &outlets.list,
+                    &format!("TCP Outlets on node {}", color_primary(node.node_name())),
+                    &empty_message,
+                )?,
+            }
```
adrianbenavides, line 76: Similarly, to avoid this match, we should change the `build_list` behavior to not print the header when the list is empty.

With these two changes to the `build_list` function, you wouldn't need to change the client code in any command and have a consistent output.

implementations/rust/ockam/ockam_command/src/tcp/outlet/show.rs
```diff
@@ -94,13 +100,44 @@ impl ShowTui {
...
+        if outlets.is_empty() {
+            tui.terminal()
+                .stdout()
+                .plain(fmt_info!(
+                    "No TCP Outlets found on node {}",
+                    color_primary(node_name.clone())
+                ))
+                .write_line()?;
+            return Ok(());
+        }
```
adrianbenavides: This is pretty much what the `tui.show()` function does in its first lines. We use that `Tui` trait exactly to avoid repeating this in every show command. Would it be OK to change the message [here](https://github.com/build-trust/ockam/blob/04cf73ab070197a6502e547da66e26adbe063494/implementations/rust/ockam/ockam_command/src/terminal/tui.rs#L26) instead?

implementations/rust/ockam/ockam_command/src/terminal/tui.rs
```diff
@@ -14,7 +16,21 @@ pub trait ShowCommandTui {
...
+                    .plain(fmt_warn!(
+                        "Failed to show {} {}",
+                        Self::ITEM_NAME.singular(),
+                        color_primary(item_name)
+                    ))
+                    .write_line()?;
+            }
+        }
+        Ok(())
+    }
```
adrianbenavides, line 33: For a separate PR:

I'd propose a different approach here. Since we want to show multiple items with the same structure as when showing a single item, we should change [this line](https://github.com/build-trust/ockam/blob/04cf73ab070197a6502e547da66e26adbe063494/implementations/rust/ockam/ockam_command/src/terminal/tui.rs#L77) to iterate the list and call `show_single`.

Then, remove the `show_multiple` method and add a new one `show_list()` that internally calls the `list_items_names` + `build_list`. That way we can reuse the implemented trait for both `show` and `list` commands.

adrianbenavides (commented):
Looks really nice, I love how all the new details in the output! I left some comments on things to use the traits/utilities we have to avoid adding logic to the commands.

implementations/rust/ockam/ockam_command/src/terminal/tui.rs
adrianbenavides: For a separate PR:

I'd propose a different approach here. Since we want to show multiple items with the same structure as when showing a single item, we should change [this line](https://github.com/build-trust/ockam/blob/04cf73ab070197a6502e547da66e26adbe063494/implementations/rust/ockam/ockam_command/src/terminal/tui.rs#L77) to iterate the list and call `show_single`.

Then, remove the `show_multiple` method and add a new one `show_list()` that internally calls the `list_items_names` + `build_list`. That way we can reuse the implemented trait for both `show` and `list` commands.

implementations/rust/ockam/ockam_command/src/tcp/outlet/static/create/long_about.txt
```diff
@@ -0,0 +1,5 @@
+Create a TCP Outlet that runs adjacent to a TCP server. The Outlet unwraps Ockam messages and delivers the raw TCP messages to the TCP server.
+
+You must specify the TCP address of the server, that your Outlet should send raw TCP traffic to. You can also name your Outlet by giving it an alias.
+
+When you create a TCP Outlet, on an Ockam node, running on your local machine, it makes the TCP service available from a worker address, to the corresponding TCP Inlet (see `ockam tcp-inlet`).
```
davide-baldo: In other parts, you refer to `TCP server` instead

implementations/rust/ockam/ockam_command/src/terminal/tui.rs
davide-baldo: Lists are supposed to be a less verbose version compared to the single element. If we always assume the same structure we lose this aspect.
Maybe we can have a `show_single(verbose: bool)`?

davide-baldo (dismissed):

implementations/rust/ockam/ockam_command/src/tcp/outlet/static/create/long_about.txt
nazmulidris: great catch, thank you. done!

implementations/rust/ockam/ockam_api/src/logs/setup.rs
nazmulidris: done

implementations/rust/ockam/ockam_command/src/tcp/outlet/create.rs
nazmulidris: great catch; done

implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs
nazmulidris: great point; i tend to add trailing period, just for readability; i will strip them +1

implementations/rust/ockam/ockam_command/src/tcp/outlet/delete.rs
nazmulidris: great point; i will make an issue for this

implementations/rust/ockam/ockam_command/src/tcp/outlet/list.rs
nazmulidris: great idea; done

implementations/rust/ockam/ockam_command/src/tcp/outlet/list.rs
nazmulidris: great catch! done

implementations/rust/ockam/ockam_command/src/tcp/outlet/show.rs
nazmulidris: great idea! done

implementations/rust/ockam/ockam_command/src/terminal/tui.rs
nazmulidris: @adrianbenavides - done

@davide-baldo - the mental model here is the the show command can be given 1 item to show. when it is given many items to show, it just shows each of those items, one after the other. the ux for the following use cases should be the same:
- `ock tcp-outlet show foo` && `ock tcp-oulet show bar`
- `ock tcp-ocket show` (then select foo and bar, press enter, and the output should be the same as above)

show / list: show is meant to display lots of info (eg: project, which has lots of info), and list is meant to show summary info (eg: just a few fields from project are displayed).

adrianbenavides (approved):
🚀

nazmulidris:
> 🚀

Thank you so much for your guidance & help!

PR #7509, opened 2024-02-01
fix(rust): don't display log messages when showing the help

etorreborre:
This PR doesn't not create an error when the `--help` or the `-h` argument is passed on the command line.


adrianbenavides (approved):
Nice!

PR #7513, opened 2024-02-01
feat(rust): add policy migration that removes trust_context_id

SanjoDeundiak:


implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -239,7 +240,9 @@ impl NodeManager {
                 .await?
                 .is_none()
             {
-                let expression = expression.unwrap_or(Expr::CONST_TRUE);
+                let check_credential_expression =
+                    Expr::Ident(format!("{}.{}", SUBJECT_KEY, ABAC_HAS_CREDENTIAL_KEY));
```
adrianbenavides, line 244: So this is the "fallback" expression? Why is that? A comment with the context or the relation with [this code block](https://github.com/build-trust/ockam/pull/7513/files#diff-fe4b8061645846b4e6d042a8d8b682b21f8e7d70e50657f62ae0e2fdbb7a2513R120-R123) would be helpful.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_20240111100002_delete_trust_context.rs
```diff
@@ -0,0 +1,327 @@
...
+            .fetch_all(&mut *transaction)
+            .await
+            .into_core()?;
+
+        for row in rows {
+            let expression = row.expression;
+
+            let expression: Expr = minicbor::decode(&expression)?;
+
+            let expression = Self::update_expression(&expression);
```
adrianbenavides, line 56: Nice!

implementations/rust/ockam/ockam_abac/src/attribute_access_control.rs
```diff
@@ -94,18 +103,30 @@ impl AbacAccessControl {
...
         match self
             .identities_attributes
             .get_attributes(&id, &self.authority)
             .await?
         {
             Some(attrs) => {
+                environment.put(
+                    format!("{}.{}", SUBJECT_KEY, ABAC_HAS_CREDENTIAL_KEY),
+                    Expr::CONST_TRUE,
+                );
```
adrianbenavides, line 123: What does this mean? Is it explained elsewhere how we use this expression?

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_20240111100002_delete_trust_context.rs
```diff
@@ -0,0 +1,327 @@
+use crate::database::{FromSqlxError, SqlxDatabase, ToSqlxType, ToVoid};
+use core::fmt;
+use minicbor::{Decode, Encode};
+use ockam_core::Result;
+use regex::Regex;
+use sqlx::sqlite::SqliteRow;
+use sqlx::*;
+
+/// This struct adds a node name column to the identity attributes table
```
adrianbenavides: Typo? `to the identity attributes table` -> `to the identity policy table`

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_20240111100002_delete_trust_context.rs
```diff
@@ -0,0 +1,327 @@
...
+pub struct PolicyTrustContextId;
+
+impl PolicyTrustContextId {
+    /// Duplicate all attributes entry for every known node
```
adrianbenavides: Typo? `attributes` -> `policies`

adrianbenavides (commented):
Looks good! I've left some questions about the fallback expression.

implementations/rust/ockam/ockam_api/src/nodes/service.rs
SanjoDeundiak: It is default policy and because I changed ABAC implementation, this
```
let check_credential_expression = Expr::Ident(format!("{}.{}", SUBJECT_KEY, ABAC_HAS_CREDENTIAL_KEY));
```
now behaves exactly like this
```
let expression = expression.unwrap_or(Expr::CONST_TRUE);
```
was behaving before.

Therefore, nothing really changed here

implementations/rust/ockam/ockam_abac/src/attribute_access_control.rs
SanjoDeundiak: There is a small comment on these constants if you click on them. So, if subject has a valid credential, we'll put `(subject.has_credential, true)` into the environment. So later on it can be checked in ABAC rules

adrianbenavides: Got it, thanks!

adrianbenavides (approved):

PR #7519, opened 2024-02-02
fix(rust): allow a policy to have only the `subject.identifier` attribute

davide-baldo:


davide-baldo:
This PR actually breaks a security check: every ABAC rules implies a credential must exists. closing.

PR #7525, opened 2024-02-05
fix(rust): store policies isolated by node and resource

adrianbenavides:
Main changes:
- Add a new column "node" to the `policy` table to store the node name associated to the policy/resource
- Replace global policies (`tcp-inlet` and `tcp-outlet`) with policies named after the resource (the alias, random or given, of the inlet/outlet)

Tested:
- [x] "Basic Web App" example
- [ ] ~Kafka example~
- [x] App

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service.rs
@@ -57,7 +57,6 @@ pub(crate) mod policy;
 pub mod portals;
 mod projects;
 pub mod relay;
-pub mod resources;
 mod secure_channel;
 mod transport;
 pub mod workers;
@@ -221,39 +220,38 @@ pub struct IdentityOverride {
 impl NodeManager {
     async fn access_control(
         &self,
-        resource: &Resource,
-        action: &Action,
+        resource: Resource,
+        action: Action,
         authority: Option<Identifier>,
-        custom_default: Option<&Expr>,
+        expression: Option<Expr>,
     ) -> Result<Arc<dyn IncomingAccessControl>> {
         if let Some(authority) = authority {
             // Populate environment with known attributes:
             let mut env = Env::new();
-
             env.put("resource.id", str(resource.as_str()));
             env.put("action.id", str(action.as_str()));
 
-            // Check if a policy exists for (resource, action) and if not, then
-            // create or use a default entry:
-            if self.cli_state.get_policy(resource, action).await?.is_none() {
-                let fallback = match custom_default {
-                    Some(e) => e.clone(),
-                    None => Expr::CONST_TRUE,
-                };
+            // Check if a policy exists for this (node, resource, action) and if not,
+            // create a policy with the given expression or the default one.
+            if self
+                .cli_state
+                .get_policy(&resource, &action)
+                .await?
+                .is_none()
+            {
+                let expression = expression.unwrap_or(Expr::CONST_TRUE);
+                let policy = Policy::new(expression);
                 self.cli_state
-                    .set_policy(resource, action, &Policy::new(fallback))
+                    .set_policy(&resource, &action, &policy)
                     .await?;
             }
             let policy_access_control = self
                 .cli_state
-                .make_policy_access_control(resource, action, env, authority)
+                .make_policy_access_control(&resource, &action, env, authority)
                 .await?;
             Ok(Arc::new(policy_access_control))
         } else {
-            warn!(
-                "no policy access control set for resource '{}' and action: '{}'",
-                &resource, &action
-            );
+            warn!("no policy access control set for resource '{resource}' and action: '{action}'");
             Ok(Arc::new(AllowAll))
         }
     }
@@ -357,7 +355,6 @@ impl NodeManager {
             transport_options.api_transport_flow_control_id.clone(),
         );
 
-        debug!("create the identity repository");
         let mut cli_state = general_options.cli_state;
         cli_state.set_node_name(general_options.node_name.clone());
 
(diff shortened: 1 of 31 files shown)


implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -221,39 +220,48 @@ pub struct IdentityOverride {
...
+            // Build expression
+            let mut expression = expression.unwrap_or(Expr::CONST_TRUE);
+
+            // TODO: verify that this is needed
+            // Add project identifier to the expression, if available
+            if let Ok(project) = self.cli_state.get_node_project(&self.node_name).await {
+                let project_identifier = project.identifier()?.to_string();
+                let project_expression = eq([ident("subject.identifier"), str(project_identifier)]);
+                expression = or([expression, project_expression]);
+            }
```
adrianbenavides: !

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -221,39 +220,48 @@ pub struct IdentityOverride {
...
+
+            // Check if a policy exists for this (node, resource, action) and if not,
+            // create a policy with the given expression or the default one.
+            if self
```
SanjoDeundiak, line 236: We can build expression after this check returned None

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -221,39 +220,48 @@ pub struct IdentityOverride {
...
-                let fallback = match custom_default {
-                    Some(e) => e.clone(),
-                    None => Expr::CONST_TRUE,
-                };
+            // Build expression
+            let mut expression = expression.unwrap_or(Expr::CONST_TRUE);
+
+            // TODO: verify that this is needed
+            // Add project identifier to the expression, if available
+            if let Ok(project) = self.cli_state.get_node_project(&self.node_name).await {
```
SanjoDeundiak: Let's not add it for all resources, only for Kafka

implementations/rust/ockam/ockam_abac/src/policy.rs
```diff
@@ -16,6 +17,7 @@ use tracing as log;
 /// Attributes come from a pre-populated environment and are augmented
 /// by subject attributes from credential data.
 pub struct PolicyAccessControl {
+    node_name: String,
```
SanjoDeundiak: Let's use node_name from the SqlxDatabase

implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
```diff
@@ -52,8 +54,16 @@ pub struct CreateCommand {
...
 
     /// Assign a name to this inlet.
-    #[arg(long, display_order = 900, id = "ALIAS", value_parser = alias_parser)]
-    pub alias: Option<String>,
+    #[arg(long, display_order = 900, id = "ALIAS", value_parser = alias_parser, default_value_t = random_name(), hide_default_value = true)]
+    pub alias: String,
+
+    #[arg(
+        hide = true,
+        long = "access-control",
```
SanjoDeundiak: Let's use name `policy` here, and maybe align with Mrinal at some point

implementations/rust/ockam/ockam_command/src/tcp/outlet/create.rs
```diff
@@ -79,19 +83,17 @@ impl CreateCommand {
     pub async fn async_run(self, ctx: &Context, opts: CommandGlobalOpts) -> miette::Result<()> {
         initialize_default_node(ctx, &opts).await?;
 
-        let node_name = opts.state.get_node_or_default(&self.at).await?.name();
```
SanjoDeundiak, line 82: I like that it know happens on the target node and not on the node running the command

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20240202100000_add_node_name_to_policies.sql
```diff
@@ -0,0 +1,14 @@
+ALTER TABLE policy RENAME TO policy_old;
+
+-- Add a new column to the table "node" to isolate policies by node
+-- The rust migration will handle the data migration between the old and new table
+CREATE TABLE policy
+(
+    node       TEXT NOT NULL, -- node name
```
SanjoDeundiak, line 7: Let's align how to merge our migrations

implementations/rust/ockam/ockam_api/src/nodes/service/kafka_services.rs
```diff
@@ -292,7 +295,8 @@ impl InMemoryNode {
                 // project identifier
                 self.cli_state
                     .set_policy(
-                        &resources::INLET,
+                        &self.node_name,
+                        &Resource::new(&project),
```
SanjoDeundiak: Won't it clash with another Inlet we create like this or anything else? Should we pick a more specific Resource name?

SanjoDeundiak (changes requested):

implementations/rust/ockam/ockam_api/src/nodes/service.rs
adrianbenavides: Done

implementations/rust/ockam/ockam_abac/src/policy.rs
adrianbenavides: Done

implementations/rust/ockam/ockam_api/src/nodes/service/kafka_services.rs
adrianbenavides: I think it shouldn't matter in this case, since the policy is hardcoded and not configurable from the client code. 

More specifically, the policy would be shared with the kafka consumers and kafka producers that the current node is running.

SanjoDeundiak (approved):

PR #7536, opened 2024-02-05
fix(rust): ockam relay create shows remote address and worker address in correct order

nazmulidris:
Fixes: https://github.com/build-trust/codebase/issues/721

ockam-documentation updates needed: https://github.com/build-trust/codebase/issues/721#issuecomment-1928450397


adrianbenavides (approved):
Great! 🚂

PR #7537, opened 2024-02-06
feat(rust): add `--enroller` flag to ockam project ticket command

nazmulidris:
Fixes: https://github.com/build-trust/codebase/issues/719

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/project/ticket.rs
+++ b/implementations/rust/ockam/ockam_command/src/project/ticket.rs
@@ -7,7 +7,9 @@ use miette::{miette, IntoDiagnostic};
 
 use ockam::identity::Identifier;
 use ockam::Context;
-use ockam_api::authenticator::direct::Members;
+use ockam_api::authenticator::direct::{
+    Members, OCKAM_ROLE_ATTRIBUTE_ENROLLER_VALUE, OCKAM_ROLE_ATTRIBUTE_KEY,
+};
 use ockam_api::authenticator::enrollment_tokens::TokenIssuer;
 use ockam_api::cli_state::enrollments::EnrollmentTicket;
 use ockam_api::cli_state::CliState;
@@ -23,6 +25,7 @@ use crate::{
     util::api::{CloudOpts, TrustOpts},
 };
 use crate::{terminal::color_primary, util::duration::duration_parser};
+use tracing::debug;
 
 const LONG_ABOUT: &str = include_str!("./static/ticket/long_about.txt");
 const AFTER_LONG_HELP: &str = include_str!("./static/ticket/after_long_help.txt");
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_command/src/project/ticket.rs
```diff
@@ -31,6 +32,9 @@ const AFTER_LONG_HELP: &str = include_str!("./static/ticket/after_long_help.txt"
 /// equal to the value of that attribute. If the value is `*` then any name is allowed
 pub const OCKAM_RELAY_ATTRIBUTE: &str = "ockam-relay";
 
+pub const OCKAM_ENROLLER_ATTRIBUTE_KEY: &str = "ockam-role";
+pub const OCKAM_ENROLLER_ATTRIBUTE_VALUE: &str = "enroller";
```
adrianbenavides: Couldn't we reuse the values from 

https://github.com/build-trust/ockam/blob/99a4b8ee78057c856f3e1a2112b854f0510517f5/implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator.rs#L14

implementations/rust/ockam/ockam_command/src/project/ticket.rs
```diff
@@ -142,12 +156,16 @@ impl TicketCommand {
...
         if let Some(id) = &self.member {
+            let attributes = self.attributes()?;
+            debug!("Attributes {:?}", attributes);
             authority_node_client
-                .add_member(ctx, id.clone(), self.attributes()?)
+                .add_member(ctx, id.clone(), attributes)
                 .await?
         } else {
+            let attributes = self.attributes()?;
+            debug!("Attributes {:?}", attributes);
```
adrianbenavides: Since the attributes are used in both branches, we could extract this right before the `if let Some(...)`.

Also, I'd write the log line as `debug!(attributes = ?a);` or optionally set a message `debug!(attributes = ?a, "passed attributes");`.

- `%variable_name`: prints the `Display` version
- `?variable_name`: prints the `Debug` version

implementations/rust/ockam/ockam_command/src/project/ticket.rs
```diff
@@ -76,8 +80,12 @@ pub struct TicketCommand {
     usage_count: Option<u64>,
 
     /// Name of the relay that the identity using the ticket will be allowed to create. This name is transformed into attributes to prevent collisions when creating relay names. For example: `--relay foo` is shorthand for `--attribute ockam-relay=foo`
-    #[arg(long = "relay", value_name = "ENROLEE_ALLOWED_RELAY_NAME")]
+    #[arg(long = "relay", value_name = "ENROLLEE_ALLOWED_RELAY_NAME")]
     allowed_relay_name: Option<String>,
+
+    /// Add the enroller role to your ticket. If you specify it, this flag is transformed into the attributes `--attribute ockam-role=enroller`
```
etorreborre: In the long help text it would be good to specify what is the effect of setting this attribute. Why would anyone want to do this?

implementations/rust/ockam/ockam_command/src/project/ticket.rs
nazmulidris: great catch! i didn't know these constants already existed in the codebase. will run a find across files next time

implementations/rust/ockam/ockam_command/src/project/ticket.rs
nazmulidris: great suggestion! good to know about the `%` and `?` meanings

implementations/rust/ockam/ockam_command/src/project/ticket.rs
nazmulidris: great idea. i add the "why" content.

etorreborre (approved):

PR #7553, opened 2024-02-08
fix(rust): use the correct policies in inlets/outlets created by kafka services

### LINKED ISSUE #7544: Delete all related data from SQL storage when deleting a node
### Observed behavior

Currently, when a node is deleted, only data in `node` table is deleted.

### Steps to reproduce

Create a node, create a related to node data (attributes entry, policy, etc.)

### Desired behavior

All data tied to that node should be deleted from the sql storage. Currently, the tables with such data are: `credential`, `identity_attributes`, `node_project`, `policy`, 

### Ockam Version

0.115.0

adrianbenavides:
- `kafka consumer` and `kafka producer` commands expect a `project_route` argument, which implies that the services created in those commands will have a special policy to check the project id
- `kafka outlet` and `kafka direct` commands do not expect any project related data, so it uses a policy depending on the target route 

Additionally, it closes https://github.com/build-trust/ockam/issues/7544

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
+++ b/implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
@@ -145,7 +145,7 @@ pub(crate) struct KafkaSecureChannelControllerImpl<F: RelayCreator> {
     inner: Arc<Mutex<InnerSecureChannelControllerImpl<F>>>,
 }
 
-//had to manually implement since #[derive(Clone)] doesn't work well in this situation
+// had to manually implement since #[derive(Clone)] doesn't work well in this situation
 impl<F: RelayCreator> Clone for KafkaSecureChannelControllerImpl<F> {
     fn clone(&self) -> Self {
         Self {
@@ -155,11 +155,12 @@ impl<F: RelayCreator> Clone for KafkaSecureChannelControllerImpl<F> {
 }
 
 /// Describe to reach the consumer node:
-/// either directly or through a relay with a relay
+/// either directly or through a relay
 #[derive(Clone)]
 pub(crate) enum ConsumerNodeAddr {
-    Direct(Option<MultiAddr>),
+    Direct(MultiAddr),
     Relay(MultiAddr),
+    None,
 }
 
 type TopicPartition = (String, i32);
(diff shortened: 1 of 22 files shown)


implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
```diff
@@ -155,10 +155,10 @@ impl<F: RelayCreator> Clone for KafkaSecureChannelControllerImpl<F> {
 }
 
 /// Describe to reach the consumer node:
-/// either directly or through a relay with a relay
+/// either directly or through a relay
 #[derive(Clone)]
 pub(crate) enum ConsumerNodeAddr {
-    Direct(Option<MultiAddr>),
+    Direct(MultiAddr),
```
davide-baldo, line 161: In direct mode, you can take the two roles, consumer or producer, the producer will specify the consumer in the `consumer_route` argument, but the consumer won't have anything to specify (unless both nodes have both roles).
Maybe we can add a third `None` option?

implementations/rust/ockam/ockam_api/src/nodes/service/kafka_services.rs
```diff
@@ -181,29 +170,39 @@ impl InMemoryNode {
...
+        let outlet_policy_expression = if consumer_route.starts_with(Project::CODE) {
+            let project_identifier = self
+                .cli_state
+                .get_node_project(&self.node_name)
+                .await?
+                .identity
+                .ok_or(ApiError::core("NodeManager has no project"))?;
+            Some(kafka_policy_expression(&project_identifier))
+        } else {
+            None
```
davide-baldo: In case we are not passing through a project, we can add an explicit policy. This policy should be the default one, but it's subject to change, and it would be more robust to repeat it here: `let check_credential_expression =
                    Expr::Ident(format!("{}.{}", SUBJECT_KEY, ABAC_HAS_CREDENTIAL_KEY))`

implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map.rs
adrianbenavides: Okey, I didn't know that. I think adding a third option will make things more understandable.

implementations/rust/ockam/ockam_api/src/nodes/service/kafka_services.rs
adrianbenavides: Will do that!

davide-baldo (approved):

PR #7561, opened 2024-02-09
feat(rust): add `variables` section to the `run` config

adrianbenavides:
This new section allows you to define local variables. It will also resolve (and give preference to) externally defined env vars. 

For example, if we externally set `MY_ENV_VAR=my_env_value` and `var_s=external`, the following yaml:

```yaml
variables:
  var_s: local
  var_b: true
  var_i: 1

nodes: 
  - ${MY_ENV_VAR}-${var_s}
  - is_${var_b}
  - num_${var_i}
```

would get resolved to:

```yaml
nodes: 
  - my_env_value-external
  - is_true
  - num_1
```

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/run/parser/config.rs
+++ b/implementations/rust/ockam/ockam_command/src/run/parser/config.rs
@@ -5,6 +5,7 @@ use crate::run::parser::projects::Projects;
 use crate::run::parser::relays::Relays;
 use crate::run::parser::tcp_inlets::TcpInlets;
 use crate::run::parser::tcp_outlets::TcpOutlets;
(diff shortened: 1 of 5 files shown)


etorreborre (approved):
Should we make the variable names non case sensitive? Because the way it is now you are forced to use the same casing in ENV variables and Yaml ones if you want to override something like my_var.

adrianbenavides:
> Should we make the variable names non case sensitive? Because the way it is now you are forced to use the same casing in ENV variables and Yaml ones if you want to override something like my_var.

I didn't think about this, but it looks like it depends on the OS. UNIX env var names are case-sensitive; they are not on Windows.

PR #7572, opened 2024-02-12
feat(rust): support policies for resource types

adrianbenavides:
- Add support for global policies in the `policy create` command. The information about whether the resource is global is derived from the resource name to guarantee backward compatibility. The global policy is used within a given node across resources of the same type (i.e. each node has its global policies).
- Change `--expression` argument on `tcp-inlet|outlet create` to `--allow`.
- The `--alias` argument of `tcp-outlet create` has been removed in favor of `--from`
- Refactor internals of policy commands to have the same structure and produce the same output as other commands
- Refactor struct returned by policy API endpoints.


implementations/rust/ockam/ockam_command/src/policy/show.rs
```diff
@@ -33,10 +37,74 @@ impl ShowCommand {
...
+            .await?
+            .into_iter()
+            .map(|i| i.resource().to_string())
+            .collect())
+    }
+
+    async fn show_single(&self, item_name: &str) -> miette::Result<()> {
+        let policy = self
+            .node
+            .show_policy(&self.ctx, &Resource::new(item_name), &self.cmd.action)
```
SanjoDeundiak: Is that going to show a policy that is global but is applied to that resource&action pair?

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
```diff
@@ -614,6 +615,7 @@ impl SessionReplacer for InletSessionReplacer {
 
             self.node_manager
                 .access_control(
+                    Some(GlobalResource::TcpInlet),
```
SanjoDeundiak: Can we improve the clarity here? It's a little bit misleading to have Resource&GlobalResource. Maybe Resource=Address and GlobalResource=ResourceType? WDYT?

implementations/rust/ockam/ockam_abac/src/storage/policy_repository_sql.rs
```diff
@@ -112,9 +129,14 @@ impl PolicyRow {
...
     }
+}
+
+impl TryFrom<PolicyRow> for Policy {
+    type Error = ockam_core::Error;
 
-    pub(crate) fn policy(&self) -> Result<Policy> {
-        Ok(Policy::new(self.expression()?))
+    fn try_from(row: PolicyRow) -> Result<Self, Self::Error> {
+        let resource = GlobalResource::strip_prefix(&row.resource).into();
```
SanjoDeundiak: Could we use an extra field instead of the prefix? That would:
1. Allow to use it in queries
2. Avoid ambiguity/security issues with concatenating such sensitive strings

implementations/rust/ockam/ockam_abac/src/storage/policy_repository.rs
```diff
@@ -14,36 +13,55 @@ pub trait PoliciesRepository: Send + Sync + 'static {
...
-    async fn get_policies_by_resource(&self, resource: &Resource) -> Result<Vec<(Action, Policy)>>;
+    async fn get_policies_by_resource(&self, resource: &Resource) -> Result<Vec<Policy>>;
 }
 
 #[derive(Debug, Decode, Encode, PartialEq, Eq)]
 #[rustfmt::skip]
 #[cbor(map)]
 pub struct Policy {
-    #[n(1)] expression: Expr,
+    #[n(1)] resource: Resource,
```
SanjoDeundiak: That makes sense. Previous version of this struct was useless

implementations/rust/ockam/ockam_abac/src/policy.rs
```diff
@@ -57,20 +61,42 @@ impl PolicyAccessControl {
             environment: env,
         }
     }
+
+    pub async fn prepend_global_expression(&self, scoped_expression: Expr) -> Result<Expr> {
```
SanjoDeundiak: I think the function name is misleading. Also, @etorreborre suggested a different way to implement such things, which I we should follow IMO. See `IdentitiesAttributes` as an example.

The structure is like this:
1. Have a simple SQL layer that only works with the storage - `PoliciesRepository` trait and `PolicySqlxDatabase` implementaiton
2. Have a layer on top, that can work with the storage, but also has business logic like verifying, aggregating data.  In this case it can be `Policies` struct,  which could have `Policies::get_policy` that will be given all the details about the resource, it would check for both specific and global policy in the storage, and return the Policy in effect (if any).

WDYT?

SanjoDeundiak (changes requested):
Left comments, let's have some discussion here before landing

implementations/rust/ockam/ockam_abac/src/storage/policy_repository_sql.rs
adrianbenavides: The `prefix` thing is no longer necessary. It now does a hard check of the resource name, meaning that a resource named "tcp-outlet" is always going to be parsed as a global resource name.

implementations/rust/ockam/ockam_abac/src/resource.rs
```diff
@@ -0,0 +1,105 @@
...
+                return Err(Error::new(
+                    Origin::Application,
+                    Kind::Misuse,
+                    "Resource or resource type must be provided",
+                ))
+            }
+        })
+    }
+
+    fn as_resource_name(&self) -> ResourceName {
```
etorreborre: This function seems a bit dangerous to me and since it is only used for display, I would inline it there.

implementations/rust/ockam/ockam_abac/src/policy.rs
```diff
@@ -82,10 +126,134 @@ impl IncomingAccessControl for PolicyAccessControl {
...
+        action: &Action,
+    ) -> Result<Expr> {
+        let mut expression = if let Some(policy) = self
+            .resources_repository
+            .get_policy(&resource.resource_name, action)
+            .await?
+        {
+            policy.into_expression()
+        } else {
+            return Err(Error::new(
```
SanjoDeundiak: I'm not sure I understand this one. So if I have a resource type policy defined, but don't have a policy for a specific resource - this function errors? I would expect we just use resource type policy in this case

implementations/rust/ockam/ockam_abac/src/policy.rs
```diff
@@ -82,10 +126,134 @@ impl IncomingAccessControl for PolicyAccessControl {
...
+                    resource.resource_name, action
+                ),
+            ));
+        };
+        if let Some(policy) = self
+            .resource_types_repository
+            .get_policy(&resource.resource_type, action)
+            .await?
+        {
+            expression = and([policy.into_expression(), expression]);
```
SanjoDeundiak: I'm not saying it doesn't make sense but for some reason I was expecting resource policy to just overwrite the resource type policy (if any). Let's maybe think which behavior makes sense more

implementations/rust/ockam/ockam_abac/src/policy.rs
```diff
@@ -82,10 +126,134 @@ impl IncomingAccessControl for PolicyAccessControl {
...
+        }
+        Ok(expression)
+    }
+
+    pub async fn set_policy(
```
SanjoDeundiak: I would just split these functions into 2: 1 for names and 1 for types. But up to you

implementations/rust/ockam/ockam_api/src/cli_state/repositories.rs
```diff
@@ -42,8 +42,11 @@ impl CliState {
         Arc::new(NodesSqlxDatabase::new(self.database()))
     }
 
-    pub(super) fn policies_repository(&self) -> Arc<dyn PoliciesRepository> {
-        Arc::new(PolicySqlxDatabase::new(self.database()))
+    pub(super) fn policies_repository(&self) -> Policies {
```
SanjoDeundiak: Should be renamed to policies() probably

implementations/rust/ockam/ockam_abac/src/policy.rs
adrianbenavides: The current approach allows us to combine the expression defined for the resource type and the one defined for the resource. I can imagine a use case where we want to have a common policy for all the outlets, but then have a specific rule that applies to each of them uniquely.

implementations/rust/ockam/ockam_abac/src/policy.rs
adrianbenavides: At this point I was expecting that we would have at least the `check credential` policy that we use as a fallback. I'm worried we could be hiding a misconfiguration bug of the resources if we fallback to the resource type policy.

implementations/rust/ockam/ockam_api/src/cli_state/repositories.rs
adrianbenavides: done

implementations/rust/ockam/ockam_abac/src/policy.rs
adrianbenavides: Done + some other changes derived from this idea.

implementations/rust/ockam/ockam_abac/src/policy.rs
adrianbenavides: I've also changed this and it now behaves as you said.

implementations/rust/ockam/ockam_abac/src/policy.rs
adrianbenavides: Done

implementations/rust/ockam/ockam_command/src/policy/show.rs
adrianbenavides: The commands now have a more predictable behavior. If you query a resource, it will return the whole policy, including the global one that applies to it, if any. If you query a resource type, it will just return its policy.

implementations/rust/ockam/ockam_abac/src/policy.rs
SanjoDeundiak:
```
 I can imagine a use case where we want to have a common policy for all the outlets, but then have a specific rule that applies to each of them uniquely.
```
Sure. Let's say you have an outlet outlet1 with policy policy1, also all outlets have policy policy2. So what should be the policy for outlet1 in the end? policy1 or policy1&policy2? I would expect the first one, but I see that your code doing the latter. So which one we should pick?

implementations/rust/ockam/ockam_abac/src/policies/storage.rs
```diff
@@ -0,0 +1,146 @@
...
+        &self,
+        resource_name: &ResourceName,
+        action: &Action,
+    ) -> ockam_core::Result<Option<ResourcePolicy>> {
+        self.resources_repository
+            .get_policy(resource_name, action)
+            .await
+    }
+
+    pub async fn get_policies_for_resource_name(
```
SanjoDeundiak: How is that different from `get_policy_for_resource_name`? Does it return policies for all possible actions?

implementations/rust/ockam/ockam_abac/src/policies/storage.rs
```diff
@@ -0,0 +1,146 @@
...
+    Action, Env, Expr, PolicyAccessControl, ResourceName, ResourcePoliciesRepository,
+    ResourcePolicy, ResourceType, ResourceTypePoliciesRepository,
+};
+use ockam_core::compat::sync::Arc;
+use ockam_core::compat::vec::Vec;
+use ockam_identity::{Identifier, IdentitiesAttributes};
+use tracing::{debug, instrument};
+
+#[derive(Clone)]
+pub struct Policies {
```
SanjoDeundiak: Honestly, now I'm not sure why we need that struct

implementations/rust/ockam/ockam_abac/src/storage/resource_policy_repository.rs
```diff
@@ -0,0 +1,39 @@
...
+/// This repository stores policies for resources.
+/// A policy is an expression which can be evaluated against an environment (a list of attribute
+/// names and values) in order to determine if a given action can be performed on a given resource.
+#[async_trait]
+pub trait ResourcePoliciesRepository: Send + Sync + 'static {
+    /// Set a policy for a given resource and action
+    async fn set_policy(
+        &self,
+        resource_name: &ResourceName,
+        resource_type: Option<&ResourceType>,
```
SanjoDeundiak: Could you please explain why we need resource_type in this repository?

implementations/rust/ockam/ockam_node/build.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    // Trigger a recompile if the migrations directory has changed.
```
SanjoDeundiak, line 2: Nice

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20240212100000_split_policies.sql
```diff
@@ -0,0 +1,15 @@
+-- Rename `policy` table
+ALTER TABLE policy RENAME TO resource_policy;
+ALTER TABLE resource_policy ADD COLUMN resource_type TEXT NULL;
+DROP INDEX IF EXISTS policy_index;
+CREATE UNIQUE INDEX resource_policy_index ON resource_policy (node_name, resource, action);
+
+-- Create a new table for resource type policies
+CREATE TABLE resource_type_policy
+(
+    resource   TEXT NOT NULL, -- resource type
```
SanjoDeundiak: Can we rename it to `type` or `resource_type`?

implementations/rust/ockam/ockam_node/src/storage/database/migrations/sqlx_migration.rs
```diff
@@ -37,7 +43,68 @@ impl NodesMigration {
         NodeNameIdentityAttributes::migrate_attributes_node_name(pool).await?;
         AuthorityAttributes::migrate_authority_attributes_to_members(pool).await?;
         PolicyTrustContextId::migrate_update_policies(pool).await?;
+        SplitPolicies::migrate_policies(pool).await?;
+        Ok(())
+    }
+
+    #[cfg(test)]
+    pub(crate) async fn migrate_schema_before(
```
SanjoDeundiak, line 52: Nice

implementations/rust/ockam/ockam_node/src/storage/database/migrations/sqlx_migration.rs
```diff
@@ -37,7 +43,68 @@ impl NodesMigration {
...
+                break;
+            }
+            transaction.execute(&*m.sql).await.into_core()?;
+        }
+        transaction.commit().await.into_core()?;
+        Ok(())
+    }
+
+    #[cfg(test)]
+    pub(crate) async fn migrate_schema_single(
```
SanjoDeundiak, line 72: Is the result the same as calling `migrate_schema_before`?

implementations/rust/ockam/ockam_abac/src/policies/storage.rs
adrianbenavides: > Does it return policies for all possible actions?

Exactly, this is the method we use from the `policy list` command. I think it's more useful this way so that you don't have to go action by action to see all the policies you have for a resource.

implementations/rust/ockam/ockam_abac/src/storage/resource_policy_repository.rs
adrianbenavides: We know the resource_name + resource_type pair only when creating an outlet or an inlet. Once we are done with that operation, how do we know what global policy should apply to the policy attached to an outlet? Before this change, we only had this information inside the `create_inlet|outlet` functions.

Having this extra column allows us to figure out the final expression for a resource policy from anywhere in the codebase. This is especially helpful from the command line: without it, the user would have to specify both the resource name and the resource type (and I don't know how they would figure this out) to retrieve the correct policy expression for that resource.

implementations/rust/ockam/ockam_abac/src/policies/storage.rs
adrianbenavides: I'm using it as a simplified "view" of both `ResourcePolicy` and `ResourceTypePolicy`. These two types can be converted to `Policy`. In that conversion, we compute the right policy (whatever it is, compounded or exclusive). It also helps to unify the output of the NodeManager functions used by the command.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/20240212100000_split_policies.sql
adrianbenavides: Sure, I'll do that

implementations/rust/ockam/ockam_node/src/storage/database/migrations/sqlx_migration.rs
adrianbenavides: No, this runs only one migration. This is used once you have the old data in place and you want to apply the migration you want to test. [Here is](https://github.com/build-trust/ockam/blob/d8fe696ad1e1491b70018ffe81ad0574b7a42f95/implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_20231231100000_node_name_identity_attributes.rs#L90) an example of that.

SanjoDeundiak: @adrianbenavides yes, good example. So if in that example you replace `migrate_schema_single` call with `migrate_schema_before` call, will the result be the same? If yes, do we really need both?

adrianbenavides: That will fail because we can't run the same migration more than once with these functions. They run the migrations manually, and not through the `Migrator`, so they are not added to the `_sqlx_migrations` table.

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_to_resource_type.rs
```diff
@@ -0,0 +1,21 @@
+use crate::{Resource, ResourceName, ResourceType};
+use ockam_core::async_trait;
+use ockam_core::compat::boxed::Box;
+use ockam_core::Result;
+
+/// This repository stores the relationship between resources and resource types.
+#[async_trait]
+pub trait ResourceToResourceTypeRepository: Send + Sync + 'static {
```
etorreborre: I would simply call this `ResourceRepository` where we store the data associated to a given resource. This will make a natural place to store future attributes of a resource if any, without the emphasis on the resource type in particular.

I would also enforce that resource names are unique (for a given node), regardless of their types, and uses this trait for the repository:
```rust
pub trait ResourceRepository: Send + Sync + 'static {
   async fn store_resource(&self, resource: &Resource) -> Result<()>,
   async fn get_resource(&self, resource_name: &ResourceName) -> Result<Option<Resource>>,
   async fn delete_resource(&self, resource_name: &ResourceName) -> Result<()>,
}
```

If you want to access the resource type for a resource you write
```rust
repository.get_resource(name).await?.map(|r| r.resource_type())
```

implementations/rust/ockam/ockam_abac/src/policy/policies.rs
```diff
@@ -0,0 +1,219 @@
...
+    resources_repository: Arc<dyn ResourcePoliciesRepository>,
+    resource_types_repository: Arc<dyn ResourceTypePoliciesRepository>,
+    resource_to_resource_type_repository: Arc<dyn ResourceToResourceTypeRepository>,
+}
+
+impl Policies {
+    pub fn new(
+        resources_repository: Arc<dyn ResourcePoliciesRepository>,
+        resource_types_repository: Arc<dyn ResourceTypePoliciesRepository>,
+        resource_to_resource_type_repository: Arc<dyn ResourceToResourceTypeRepository>,
```
etorreborre: I think that better names for those parameters would be:
```rust
resources_policies_repository: Arc<dyn ResourcePoliciesRepository>,
resource_types_policies_repository: Arc<dyn ResourceTypePoliciesRepository>,
```

I would also extract the `ResourcesRepository` from there, and use it at the place where resources are created, not when their policy is set.

implementations/rust/ockam/ockam_abac/src/policy/policies.rs
```diff
@@ -0,0 +1,219 @@
...
+        &self,
+        resource_name: &ResourceName,
+        action: &Action,
+    ) -> Result<()> {
+        self.resources_repository
+            .delete_policy(resource_name, action)
+            .await
+    }
+
+    pub async fn delete_all_policies_for_resource_name(
```
etorreborre: If the intent is to delete a resource, I think that this should rather go to a `Resources` component where:

- There is a `delete_resource` function, delegating the deletion to the `ResourcesRepository`
- Where the db also deletes the policies in one db transaction

implementations/rust/ockam/ockam_abac/src/policy/policies.rs
```diff
@@ -0,0 +1,219 @@
...
+        self.resources_repository
+            .delete_all_policies(resource_name)
+            .await?;
+        self.resource_to_resource_type_repository
+            .delete_all(resource_name)
+            .await?;
+        Ok(())
+    }
+
+    pub async fn derive_expression_for_resource(
```
etorreborre: "derive" is bit of a strong word for "this or that" IMO. That makes me think of a more complicated computation going on.

implementations/rust/ockam/ockam_abac/src/policy/policies.rs
```diff
@@ -0,0 +1,219 @@
...
+        resource_type: &ResourceType,
+        action: &Action,
+        expression: &Expr,
+    ) -> Result<()> {
+        self.resource_types_repository
+            .set_policy(resource_type, action, expression)
+            .await
+    }
+
+    pub async fn set_default_policy_for_resource_type(
```
etorreborre: It looks like this function does not need to be public.

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_policy_repository_sql.rs
```diff
@@ -0,0 +1,199 @@
...
+
+use crate::{Action, Expr, ResourceName, ResourcePoliciesRepository, ResourcePolicy};
+
+#[derive(Clone)]
+pub struct ResourcePolicySqlxDatabase {
+    database: SqlxDatabase,
+}
+
+impl ResourcePolicySqlxDatabase {
+    /// Create a new database for policies keys
```
etorreborre: Simply "for policies"?

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_to_resource_type_sql.rs
```diff
@@ -0,0 +1,142 @@
...
+
+use crate::{Resource, ResourceName, ResourceToResourceTypeRepository, ResourceType};
+
+#[derive(Clone)]
+pub struct ResourceToResourceTypeSqlxDatabase {
+    database: SqlxDatabase,
+}
+
+impl ResourceToResourceTypeSqlxDatabase {
+    /// Create a new database for policies keys
```
etorreborre: This comment should say "for resources".

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_policy_repository_sql.rs
```diff
@@ -0,0 +1,199 @@
...
+    pub fn new(database: SqlxDatabase) -> Self {
+        debug!("create a repository for policies");
```
etorreborre: The log message could be more precise, like `for resource policies` and `for resource types policies`, `for resources` for the other 2 repositories.

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_to_resource_type_sql.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub fn new(database: SqlxDatabase) -> Self {
+        debug!("create a repository for policies");
```
etorreborre: The log message needs to be changed.

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_type_policy_repository.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    /// Return the policy associated to a given resource and action
+    async fn get_policy(
+        &self,
+        resource_type: &ResourceType,
+        action: &Action,
+    ) -> Result<Option<ResourceTypePolicy>>;
+
+    /// Return the list of all the policies associated to a given resource
+    async fn get_policies_by_resource(
```
etorreborre: `by_resource_type`

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_type_policy_repository.rs
```diff
@@ -0,0 +1,36 @@
...
+pub trait ResourceTypePoliciesRepository: Send + Sync + 'static {
+    /// Set a policy for a given resource and action
+    async fn set_policy(
+        &self,
+        resource_type: &ResourceType,
+        action: &Action,
+        expression: &Expr,
+    ) -> Result<()>;
+
+    /// Return the policy associated to a given resource and action
```
etorreborre: "to a given resource type"

implementations/rust/ockam/ockam_abac/src/resource.rs
```diff
@@ -0,0 +1,74 @@
...
+#[rustfmt::skip]
+#[cbor(index_only)]
+pub enum ResourceType {
+    #[n(1)] TcpInlet,
+    #[n(2)] TcpOutlet,
+    #[n(3)] Echoer,
+}
+
+impl ResourceType {
+    pub const VALUES: &'static [Self] = &[Self::TcpInlet, Self::TcpOutlet, Self::Echoer];
```
etorreborre: It could be a good idea to add a unit test checking that there are 3 values in that list. Like this, if we add a new resource in the enum, we can't forget to update the list.

implementations/rust/ockam/ockam_api/src/cli_state/resources.rs
```diff
@@ -0,0 +1,12 @@
+use super::Result;
+use crate::CliState;
+use ockam_abac::ResourceName;
+
+impl CliState {
+    pub async fn delete_resource(&self, resource_name: &ResourceName) -> Result<()> {
+        self.policies()
```
etorreborre: This could be `self.resources().delete_resource(resource_name)` with what I wrote in one of my previous comments.

implementations/rust/ockam/ockam_api/src/nodes/models/policies.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    pub fn expression(&self) -> &Expr {
+        &self.expression
+    }
+}
+
+impl From<ResourceTypePolicy> for Policy {
+    fn from(policy: ResourceTypePolicy) -> Self {
+        Policy {
+            resource_name: policy.resource_type.to_string(),
```
etorreborre: I would resource the existing enum `ResourceTypeOrName` rather than disguising a type as a name. This is more predictable and consistent with the API request.

implementations/rust/ockam/ockam_api/src/nodes/models/policies.rs
```diff
@@ -0,0 +1,112 @@
...
+                return Err(Error::new(
+                    Origin::Application,
+                    Kind::Misuse,
+                    "Resource or resource type must be provided",
+                ))
+            }
+        })
+    }
+
+    fn as_resource_name(&self) -> ResourceName {
```
etorreborre: I'd rather inline that function in `Display` rather than making the reader think that a type could be a name.

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
```diff
@@ -222,8 +222,8 @@ pub(crate) struct Registry {
     pub(crate) kafka_services: RegistryOf<Address, KafkaServiceInfo>,
     pub(crate) hop_services: RegistryOf<Address, HopServiceInfo>,
     pub(crate) relays: RegistryOf<String, RegistryRelayInfo>,
-    pub(crate) inlets: RegistryOf<Alias, InletInfo>,
-    pub(crate) outlets: RegistryOf<Alias, OutletInfo>,
+    pub(crate) inlets: RegistryOf<String, InletInfo>,
```
etorreborre, line 225: It would be good to have a comment explaining why inlets are indexed by a random string, while outlets are indexed by a worker address.

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
```diff
@@ -289,3 +289,99 @@ impl<K: Clone, V: Clone> RegistryOf<K, V> {
...
+
+    #[tokio::test]
+    async fn outlet_registry_generate_worker_address() {
+        let registry = Registry::default();
+        let outlet_info = |worker_addr: Address| -> OutletInfo {
+            OutletInfo::new(&SocketAddr::from(([127, 0, 0, 1], 0)), Some(&worker_addr))
+        };
+
+        // No worker address passed
+        let worker_addr = registry.outlets.generate_worker_addr(None).await;
```
etorreborre, line 319: I would expand the comment to explain that a default address is returned if this the first time we generate an address for an outlet.

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
```diff
@@ -289,3 +289,99 @@ impl<K: Clone, V: Clone> RegistryOf<K, V> {
...
+
+        // Same worker address passed, should return the same address and not a random one
+        let worker_addr = registry
+            .outlets
+            .generate_worker_addr(Some(passed_addr.clone()))
+            .await;
+        assert_eq!(worker_addr, passed_addr.clone());
+
+        // Start from scratch
+        let registry = Registry::default();
```
etorreborre, line 359: Make a separate test then?

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
etorreborre: Eventually all this data will go inside a repository!

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -221,40 +219,38 @@ pub struct IdentityOverride {
...
+                    resource,
+                    action,
+                    env,
+                    authority,
+                )
                 .await?;
             Ok(Arc::new(policy_access_control))
         } else {
-            warn!("no policy access control set for resource '{resource}' and action: '{action}'");
+            warn!("no policy access control set for resource '{resource_name}' and action: '{action}'");
```
etorreborre: Why not print the resource type as well? This could be helpful when debugging client issues.

implementations/rust/ockam/ockam_app_lib/src/shared_service/relay/create.rs
```diff
@@ -109,28 +109,28 @@ impl AppState {
...
         .into_iter()
-        .find(|r| r.alias() == relay_name))
+        .find(|r| r.alias() == relay_alias))
 }
 
-async fn relay_name(cli_state: &CliState) -> ockam::Result<String> {
-    let bare_relay_name = bare_relay_name(cli_state).await?;
+async fn relay_remote_address(cli_state: &CliState) -> ockam::Result<String> {
+    let bare_relay_name = relay_alias(cli_state).await?;
     Ok(format!("forward_to_{bare_relay_name}"))
```
etorreborre, line 130: I did not realize that the `forward_to` convention was spread out across several files. This is definitely something to refactor in the future.

implementations/rust/ockam/ockam_app_lib/src/invitations/commands.rs
```diff
@@ -189,22 +190,22 @@ impl AppState {
     pub async fn create_service_invitation_by_alias(
         &self,
         recipient_email: EmailAddress,
-        alias: &str,
+        outlet_worker_addr: &Address,
```
etorreborre, line 193: Maybe `outlet_addr` is enough? Do we need to mention that there's a worker at that address?

implementations/rust/ockam/ockam_command/src/policy/mod.rs
```diff
@@ -54,6 +54,18 @@ impl PolicyCommand {
     }
 }
 
-pub(crate) fn policy_path(r: &Resource, a: &Action) -> String {
-    format!("/policy/{r}/{a}")
+pub(crate) fn resource_type_parser(input: &str) -> miette::Result<ResourceType> {
+    let resource = input.to_string().replace('_', "-");
```
etorreborre: Why do we do this replacement?

implementations/rust/ockam/ockam_command/src/policy/mod.rs
```diff
@@ -54,6 +54,18 @@ impl PolicyCommand {
...
+    ResourceType::VALUES
+        .iter()
+        .find(|global_resource| global_resource.to_string() == resource)
```
etorreborre: s/global_resource/resource_type

implementations/rust/ockam/ockam_command/src/policy/mod.rs
```diff
@@ -54,6 +54,18 @@ impl PolicyCommand {
     }
 }
 
-pub(crate) fn policy_path(r: &Resource, a: &Action) -> String {
-    format!("/policy/{r}/{a}")
+pub(crate) fn resource_type_parser(input: &str) -> miette::Result<ResourceType> {
```
etorreborre, line 58: I would put this parser close to the `ResourceType` data type so that it can be reused.

implementations/rust/ockam/ockam_command/src/policy/show.rs
```diff
@@ -33,10 +42,89 @@ impl ShowCommand {
...
+}
+
+impl ShowTui {
+    pub async fn run(
+        ctx: &Context,
+        opts: CommandGlobalOpts,
+        cmd: ShowCommand,
+    ) -> miette::Result<()> {
+        let node = BackgroundNodeClient::create(ctx, &opts.state, &cmd.at).await?;
+        let resource_type = if cmd.resource_type.is_none() && cmd.resource.is_none() {
```
etorreborre: I haven't seen one example/test where we don't specify resource or resource type. I don't really see why having a default value matters here.

implementations/rust/ockam/ockam_node/build.rs
etorreborre: Indeed! 🚀

implementations/rust/ockam/ockam_abac/src/policy/policies.rs
```diff
@@ -0,0 +1,219 @@
...
+            self.resources_repository
+                .set_policy(&resource.resource_name, action, expression)
+                .await?;
+        }
+        // Set resource to resource type mapping.
+        self.resource_to_resource_type_repository
+            .set(resource)
+            .await?;
+        // Set default policy for resource_type if none exists.
+        if self
```
SanjoDeundiak: Can this be a sqlx migration that on the startup sets default expression for every resource type?

implementations/rust/ockam/ockam_command/src/policy/list.rs
```diff
@@ -40,21 +44,25 @@ impl ListCommand {
         let node = BackgroundNodeClient::create(ctx, &opts.state, &self.at).await?;
         let is_finished: Mutex<bool> = Mutex::new(false);
 
-        let resource = &self.resource;
+        let resource_type = if self.resource_type.is_none() && self.resource.is_none() {
+            Some(&ResourceType::TcpOutlet)
```
SanjoDeundiak: Could you please elaborate why is that true?

adrianbenavides: All the list commands have a default target. I've added this fallback value to keep the UX consistent and not return an error when the user runs `ockam policy list`.

implementations/rust/ockam/ockam_abac/src/policy/policies.rs
adrianbenavides: If we prepopulate this table from a migration we will be coupling the SQL code with the values of both `ResourceType` and `Action`. We don't even have a defined set of values for `Action`. So, every time we add a new value for any of those objects, we'd need a new migration to add them to the table. I think that's prone to errors and harder to maintain.

implementations/rust/ockam/ockam_command/src/policy/mod.rs
adrianbenavides: I think it's not a good idea now that we show the user the valid values. I'll remove it.

implementations/rust/ockam/ockam_command/src/policy/show.rs
adrianbenavides: The reason for that is pretty much the same as [here](https://github.com/build-trust/ockam/pull/7572#discussion_r1497222491). From the UX perspective, I think it's better to just show a policy rather than return an error). It's also how the rest of the `show` commands behave, so I didn't want to add a special behavior on this one.

implementations/rust/ockam/ockam_command/src/policy/mod.rs
adrianbenavides: I'll still need a wrapper around the parser in `ockam_command` to add the miette error to return a better message to the user. I'll add both.

implementations/rust/ockam/ockam_app_lib/src/invitations/commands.rs
adrianbenavides: It helps to identify if we are talking about the outlet as a target route (MultiAddr like /project/default/...) or as a local service (that could be either a simple Multiaddr like /service/outlet or a plain Address) without having to figure it out from the code.

implementations/rust/ockam/ockam_app_lib/src/shared_service/relay/create.rs
adrianbenavides: I tried to pull the string but it's bigger than I expected. The relay also has the problem of having an alias and a worker address that by default share the same value. This alone makes things more complex than necessary.

Also, this problem with the `forward_to` has generated bugs more than once.

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
adrianbenavides: Inlets should also be indexed by Address and we'll do that refactor as soon as we can. I don't think it's worth it to justify why inlets can't be indexed by address right now.

implementations/rust/ockam/ockam_api/src/nodes/models/policies.rs
adrianbenavides: Totally agree, thanks!

implementations/rust/ockam/ockam_abac/src/policy/policies.rs
adrianbenavides: It was more complex in the first iteration when we were doing the combination of policies. I'm going to replace "derive" by "get".

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_to_resource_type.rs
adrianbenavides: If we check only the resource name and node, the user won't be able to create an inlet and an outlet with the same name, and that limitation wouldn't make much sense. What am I missing here?

implementations/rust/ockam/ockam_command/src/policy/show.rs
etorreborre: I see. I'm not sure I agree about the UX side. I think it makes sense for commands creating entities by default but not for policies. Maybe we can involve @mrinalwadhwa and @nazmulidris for their thoughts on this (but this doesn't have to hold up the PR).

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_to_resource_type.rs
etorreborre: I thought about that too and:

 - I am not sure it makes sense to use the same name, on the same node (did we ever have an example of this?)
 - It could even be the sign of a configuration error where the same name is reused but shouldn't

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_repository.rs
```diff
@@ -0,0 +1,17 @@
+use crate::{Resource, ResourceName};
+use ockam_core::async_trait;
+use ockam_core::compat::boxed::Box;
+use ockam_core::Result;
+
+/// This repository stores the relationship between resources and resource types.
```
etorreborre: The comment could just say now "stores resources".

etorreborre (dismissed):
Wow, what a PR :-).

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_repository.rs
adrianbenavides: so many comments and doc strings 🤯

etorreborre (dismissed):

implementations/rust/ockam/ockam_command/src/policy/list.rs
sanjoD: Shouldn't we list all policies (for all resources and resource types) in that case? I think we discussed that

SanjoDeundiak: Shouldn't we list all policies (for all resources and resource types) in that case? I think we discussed that

implementations/rust/ockam/ockam_abac/src/policy/policies.rs
SanjoDeundiak: Your points are valid, but I still would consider that option. Doing it right here is 1. Inefficient 2. Can lead to it not being called at all (what if we didn't call `store_policy_for_resource` for that resource? the default policy won't be set for that resource type at all?)
That being said, the migration is not perfect place, but AFAIK the goal is to set default policies for all resource types in the very beginning of the program execution, right? That's what sql migration could do

SanjoDeundiak: The alternative could be to return the default policy for a resource type in source code inside the `get_policy_for_resource_type`, but I like it less

adrianbenavides: Maybe we can do both?
- Adding a known set of resource types in the migration
- Keep the logic in `store_policy_for_resource`

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
```diff
@@ -181,20 +178,25 @@ impl NodeManager {
         &self,
         ctx: &Context,
         socket_addr: SocketAddr,
-        worker_addr: Address,
-        alias: String,
+        worker_addr: Option<Address>,
         reachable_from_default_secure_channel: bool,
         access_control: Option<Arc<dyn IncomingAccessControl>>,
```
SanjoDeundiak: I think only one of the two arguments can be `Some(...)` at the same time, right? I think it's worth either introducing a type for that, or at least have a runtime check that both of them aren't present at the same time

implementations/rust/ockam/ockam_command/src/policy/list.rs
adrianbenavides: True, I'll include that.

implementations/rust/ockam/ockam_abac/src/policy/policies.rs
adrianbenavides: Another problem I didn't mention of adding values to the migration. The table has a `node_name` column.

SanjoDeundiak: We'll attempt to insert the default policies on node creation

adrianbenavides: @SanjoDeundiak ready!

implementations/rust/ockam/ockam_core/src/error/mod.rs
```diff
@@ -145,6 +145,17 @@ impl ErrorTrait for Error {
     }
 }
 
+impl From<strum::ParseError> for Error {
```
SanjoDeundiak, line 148: Please add `#[track_caller]`

implementations/rust/ockam/ockam_api/src/nodes/service.rs
```diff
@@ -221,40 +218,47 @@ pub struct IdentityOverride {
 impl NodeManager {
     async fn access_control(
```
SanjoDeundiak, line 215: The flow of this function is a mess honestly, but that's out of scope of this PR

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_api/src/nodes/service.rs
adrianbenavides: I take note of this

etorreborre (approved):

PR #7574, opened 2024-02-12
Propagate the tracing context at the level of ockam messages

etorreborre:
This PR adds an optional `tracing_context` field to Ockam messages:

   - That field is serializes as a String [a propagated OpenTelemetry context ](https://opentelemetry.io/docs/specs/otel/context/api-propagators/).
   - The field is read on the Elixir side when a message arrives and is used as the root of the trace which processes the message (same thing when an Elixir node sends a message to the Rust side).
   - The similar field which was present on Request/Responses (one level above Ockam messages) is now removed.   

One scenario used to test this PR was the sending of an Ockam message to a background node. Here is an example of a trace: 
image

What we see is that:

 - The `Messages::send_message` function calls the `TcpSender` worker to send a message.
 - This message is received by the `TcpRcvProcessor` on the background node and forwarded to the `Echoer` service.
 - And the message is finally routed back to the caller.

## How does it work?

The main idea is to be able to serialize a proper tracing context in a `TransportMessage`. In order to do this, we need to solve 3 issues:

 1. The `OpenTelemetryContext` must contain the current trace id and the _latest_ span id. 
 2. The tracing context must be preserved across async boundaries, for example between command invocation (synchronous) and command execution (asynchronous)
 3. When we make a `LocalMessage` (for example containing a `Request`) we need to set the proper tracing context, _and make sure that this tracing context is staying up to date_ until we create a `TransportMessage`.

### Correct `OpenTelemetry` context 

This is not entirely obvious because of some discrepancies between the `tracing` crate and the `opentelemetry` crate(s). Please read the comments [here](https://github.com/build-trust/ockam/blob/4cd4f3a236a44b734b9ff52d14e793e0d69e5e9d/implementations/rust/ockam/ockam_node/src/opentelemetry.rs#L35) for an explanation on how this was solved in a previous PR.

### Passing the tracing context across async boundaries

This is solved by using the `with_context` function, for example [here](https://github.com/build-trust/ockam/pull/7574/files#diff-721ad94ae6660f5847f6671ca3e9d1843cc9e9f804e41b6ebf4350c645990ddbR86).

Note that we pass an `opentelemetry::Context` which is retrieved from the `OpenTelemetry::current()` function which makes sure that the last created span id is referenced.

There are currently other places in the code where `.with_context` is used, for example in `ockam_command/src/space/list.rs`. In this case we might not have the latest span id but this is not a problem since this code is called via the `embedded_node` function which sets the current context properly.

### Passing the tracing context intra-node

There are 3 crucial places where the tracing context needs to be set and updated:

 - `LocalMessage::into_transport_message` when we create a `TransportMessage` from a `LocalMessage` containing the tracing context. Note that we call `local_message.tracing_context.update()` to make sure that the tracing context really contains the latest span id
 - `Context::send_from_address_impl`. This is where we create a `LocalMessage` when a worker sends a payload on a given route (the `.update()` method is also called here)
 - `WorkerRelay::recv_message`. Here we get the tracing context conveyed by the `LocalMessage` and:
     - set it on the worker context
     - use it (with `with_context) on the async call for `worker.handle_message` 

If any of these things are missing then we are breaking traces.

## Testing

I have added 2 tests sending a "hello" message across 2 nodes, to and `Echoer` and back (with and without a secure channel).

These tests check that we end-up with one trace, showing spans from both nodes, in a correct parent/child relationship:
```rust
root
├── create tcp transport
├── create tcp transport
└── TcpSendWorker::handle_message
    └── TcpRecvProcessor::process
        └── Echoer::handle_message
            └── TcpSendWorker::handle_message
                └── TcpRecvProcessor::process
```

These kind of tests have one drawback, they break if code is refactored, methods renamed etc... However without these tests it is very easy to break tracing and not realize it.

## Additional implementations notes

 - The `OpenTelemetry` code has been moved to the `ockam_core` crate to be close to the Ockam message protocol.
 - A `Messages` interface has been refactored for both a `BackgroundNodeClient` and the `NodeManager` in order to instrument a span showing the invocation of the `send_message` function when running `ockam message send`
 - Similarly I extracted an `Outlets` trait to add a span for the outlet creation (this is [something that we needed to do anyway](https://github.com/build-trust/ockam/issues/6698))


davide-baldo:
As far as I can see, we are tracing everything at low-level, worker by worker.  What does it happen when we are sending gigabytes over the wire? Won't tracing get in the way?

implementations/rust/ockam/ockam_api/src/nodes/service/messages.rs
```diff
@@ -1,40 +1,76 @@
...
+    ) -> miette::Result<Vec<u8>> {
+        let msg_length = message.len();
+        let connection_ctx = Arc::new(ctx.async_try_clone().await.into_diagnostic()?);
+        let connection = self
+            .make_connection(connection_ctx, to, self.identifier(), None, timeout)
+            .await
+            .into_diagnostic()?;
+        let route = connection.route().into_diagnostic()?;
+
+        trace!(target: TARGET, route = %route, msg_l = %msg_length, "sending message");
```
adrianbenavides: We started using this `target: TARGET` pattern for no specific reason back when we started writing the `ockam_api`. Does it really make sense to use it? Maybe it shortens the prefix of a log line, not sure if that's true or even worth it if it is.

implementations/rust/ockam/ockam_api/tests/trace_context_propagation.rs
```diff
@@ -0,0 +1,281 @@
...
+
+EncryptorWorker::shutdown
+└── TcpSendWorker::handle_message
+    └── TcpRecvProcessor::process
+        └── DecryptorWorker::handle_message
+            └── handle_decrypt
+
+TcpRecvProcessor::process
+
+TcpRecvProcessor::process
```
adrianbenavides, line 119: Is there any situation where these `TcpRecvProcessor::process` traces have spans?

implementations/rust/ockam/ockam_core/src/routing/message/opentelemetry.rs
```diff
@@ -1,35 +1,67 @@
...
 impl OpenTelemetryContext {
     /// Recover an OpenTelemetry context from the currently serialized data
-    pub fn extract(&self) -> opentelemetry::Context {
+    pub fn extract(&self) -> Context {
         global::get_text_map_propagator(|propagator| propagator.extract(self))
     }
 
     /// Serialize the current OpenTelemetry context as OpenTelemetryContext
-    pub fn inject(context: &opentelemetry::Context) -> Self {
+    pub fn inject(context: &Context) -> Self {
```
adrianbenavides: This function doesn't need to be public I think.

implementations/rust/ockam/ockam_core/src/routing/message/opentelemetry.rs
```diff
@@ -68,10 +100,58 @@ impl OpenTelemetryContext {
...
+        };
+        OpenTelemetryContext::current()
+    }
+
     fn empty() -> Self {
         Self(HashMap::new())
     }
 
+    /// Start a new span within the current opentelemetry context
+    pub fn start_tracer(&self, span_name: String) -> Context {
```
adrianbenavides: This function is not used

implementations/rust/ockam/ockam_core/src/routing/message/opentelemetry.rs
```diff
@@ -68,10 +100,58 @@ impl OpenTelemetryContext {
         result.unwrap_or_else(|| OpenTelemetryContext::inject(&opentelemetry::Context::current()))
     }
 
+    /// Set this OpenTelemetry context as the new parent context
+    pub fn set_as_parent_context(self) {
```
adrianbenavides: This one could also be private

adrianbenavides (commented):
Nice, the tests were really helpful to understand what's going on. I've left some minor comments.

implementations/rust/ockam/ockam_api/src/nodes/service/messages.rs
etorreborre: I'm removing it for now, at least here. I'll revisit when I will have a look at some exact traces for the portal use case.

implementations/rust/ockam/ockam_api/tests/trace_context_propagation.rs
etorreborre: They would have if we called any instrumented function before grabbing the tracing context. One such possibility I think is if there's a decoding error.

implementations/rust/ockam/ockam_core/src/routing/message/opentelemetry.rs
etorreborre: Done.

implementations/rust/ockam/ockam_core/src/routing/message/opentelemetry.rs
etorreborre: Gone.

implementations/rust/ockam/ockam_core/src/routing/message/opentelemetry.rs
etorreborre: Done.

etorreborre:
Thanks @adrianbenavides I applied your suggestions and squashed/rebased.

adrianbenavides (approved):
🚀

PR #7587, opened 2024-02-13
feat(rust): tcp inlet creation will always optional validate unless `--no-connection-wait` is used`

davide-baldo:
- default create inlet behavior is to always validate but to keep trying anyway
- `--no-validation` is now `--no-connection-wait`

Screenshot_20240215_133906

Screenshot_20240215_133035

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
@@ -22,7 +22,7 @@ use ockam_api::journeys::{
 use ockam_api::nodes::models::portal::InletStatus;
 use ockam_api::nodes::service::portals::Inlets;
 use ockam_api::nodes::BackgroundNodeClient;
-use ockam_api::random_name;
+use ockam_api::{random_name, ConnectionStatus};
 use ockam_core::api::{Reply, Status};
 use ockam_multiaddr::proto::Project;
 use ockam_multiaddr::{MultiAddr, Protocol as _};
@@ -33,8 +33,7 @@ use crate::terminal::OckamColor;
 use crate::util::duration::duration_parser;
 use crate::util::parsers::socket_addr_parser;
 use crate::util::{async_cmd, find_available_port, port_is_free_guard, process_nodes_multiaddr};
-use crate::Error;
-use crate::{docs, fmt_log, fmt_ok, CommandGlobalOpts};
+use crate::{docs, fmt_info, fmt_log, fmt_ok, fmt_warn, CommandGlobalOpts, Error};
 
 const AFTER_LONG_HELP: &str = include_str!("./static/create/after_long_help.txt");
 
@@ -77,9 +76,9 @@ pub struct CreateCommand {
     #[arg(long, value_parser = duration_parser)]
     pub timeout: Option<Duration>,
 
-    /// Avoid synchronosly validation of the outlet node connection
-    #[arg(long, id = "NO_VALIDATION", default_value = "false")]
-    no_validation: bool,
+    /// Create the inlet without waiting for the outlet to connect
+    #[arg(long, default_value = "false")]
+    no_connection_wait: bool,
 }
 
 pub(crate) fn default_from_addr() -> SocketAddr {
@@ -135,7 +134,7 @@ impl CreateCommand {
                         &cmd.authorized,
                         &cmd.policy_expression,
                         cmd.connection_wait,
-                        !cmd.no_validation,
+                        !cmd.no_connection_wait,
                     )
                     .await?;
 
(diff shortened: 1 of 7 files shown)


implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
```diff
@@ -203,7 +203,21 @@ impl CreateCommand {
 
         opts.terminal
             .stdout()
-            .plain(
+            .plain(if let Some(last_failure) = inlet.last_failure.as_deref() {
+                fmt_err!(
```
adrianbenavides: Since this error it's not really a failure that forces the user to redo any action or do anything about it, I would use `fmt_warn`

implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
```diff
@@ -203,7 +203,21 @@ impl CreateCommand {
...
+                    &cmd.from
+                        .to_string()
+                        .color(OckamColor::PrimaryResource.color()),
+                    &node.node_name().color(OckamColor::PrimaryResource.color()),
+                    &cmd.to
+                        .to_string()
+                        .color(OckamColor::PrimaryResource.color())
+                ) + &fmt_log!(
+                    "Error: {}\n",
+                    last_failure.color(OckamColor::Failure.color())
```
adrianbenavides: This message can be misleading and confusing. We could log it (you already [added](https://github.com/build-trust/ockam/pull/7587/files#diff-5b3f34a98f0adf81c262908c1b6ee444f6e61e4242075b18860c9a57ea986e3fR409) this right?) but not write it to stdout.

implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
```diff
@@ -215,8 +229,21 @@ impl CreateCommand {
                     &cmd.to
                         .to_string()
                         .color(OckamColor::PrimaryResource.color())
-                ),
-            )
+                )
+            } else {
+                fmt_ok!(
+                    "TCP Inlet {} on node {} will automatically connect\n",
```
adrianbenavides: To keep the messages consistent, we should use `inlet` (in lowercase). For example: `The inlet {} on node {} will automatically connect`.

I would also rephrase this to `The inlet {} on node {} will automatically connect when the outlet becomes available`.

adrianbenavides:
Also, could you add a simple bats test as an example of this? Something along these lines:

```
ockam node create n2
ockam tcp-inlet create --at /node/n2 --from 127.0.0.1:6000 --to /ip4/127.0.0.1/tcp/4000/service/outlet
ockam node create n1 -t 127.0.0.1:4000
ockam tcp-outlet create --at /node/n1 --to 127.0.0.1:5000
sleep 5
curl 127.0.0.1:6000
```

implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
davide-baldo: Maybe i can show the "last failure" when listing/showing the inlets

davide-baldo: After some experimentation, I'm completely removing `last_failure`. The errors are too rough to for the stdout.

davide-baldo:
@adrianbenavides 
I've pushed an updated version and updated the screenshots as well

implementations/rust/ockam/ockam_command/tests/bats/portals.bats
```diff
@@ -228,3 +228,17 @@ teardown() {
...
+
+  run_success "$OCKAM" node create n1
+  run_success "$OCKAM" node create n2
+
+  run_success "$OCKAM" tcp-inlet create --at /node/n2 --from "127.0.0.1:$port" --to /node/n1/service/outlet
+  run_success "$OCKAM" tcp-outlet create --at /node/n1 --to 127.0.0.1:5000
+
+  sleep 5
+
+  run_success curl --fail --head --max-time 20 "127.0.0.1:$port"
```
adrianbenavides: I think the n1 must be created after the inlet, otherwise it will connect right away.

implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
```diff
@@ -215,8 +214,32 @@ impl CreateCommand {
...
+                    &cmd.from
+                        .to_string()
+                        .color(OckamColor::PrimaryResource.color()),
+                    &node.node_name().color(OckamColor::PrimaryResource.color())
+                ) + &fmt_log!(
+                    "to the outlet at {}",
+                    &cmd.to
+                        .to_string()
+                        .color(OckamColor::PrimaryResource.color())
+                )
```
adrianbenavides: I just realized from checking the screenshots that this produces a weird message due to the second line (`"to the outlet at {}",`).

Can we change it to the following?

```
fmt_ok!(
    "The inlet {} on node {} will automatically connect when the outlet at {} is available\n",
    &cmd.from
        .to_string()
    &node.node_name().color(OckamColor::PrimaryResource.color()),
    &cmd.to
        .to_string()
        .color(OckamColor::PrimaryResource.color())
)
```

adrianbenavides (approved):
🚀

PR #7592, opened 2024-02-13
feat(rust): improve authority debug-ability

SanjoDeundiak:


polvorin (dismissed):

implementations/rust/ockam/ockam_api/src/authenticator/enrollment_tokens/acceptor.rs
```diff
@@ -46,14 +46,24 @@ impl EnrollmentTokenAcceptor {
...
+                    "Unknown enrollment token".to_string(),
+                )));
+            }
+            Err(err) => {
+                warn!(
+                    "Error using an enrollment token received from {}. Error: {}",
+                    from, err
+                );
+                return Ok(Either::Right(EnrollmentTokenAcceptorError(
+                    "Error using the enrollment token".to_string(),
```
etorreborre, line 61: Why not surface the `err` here, so that we don't necessarily have to look for the logs?

etorreborre (dismissed):
Nice and useful changes 🏎️.
The only thing missing would be a comment to justify why we need the reference field.

implementations/rust/ockam/ockam_api/src/authenticator/enrollment_tokens/acceptor.rs
SanjoDeundiak: Returning unstructured server errors is a bad UX and potentially dangerous (can leak sensitive-ish data)

etorreborre (approved):

Issue #5452, opened 2023-07-28
Add an option to enroll without creating a space nor a project


etorreborre:
# Description

The current `ockam enroll` command is very practical because it does 3 things at once:

 1. it executes an authentication flow to enroll a user (you can think about that as "create an account on the Orchestrator")
 2. it creates a default `Space` for the user (a `Space` is a collection of `Projects` at the level of an organization)
 3. it creates a default `Project` in that `Space` for the user

Independent commands actually exist to create a `Space` and create a `Project` for a given user, and could be executed independently of `ockam enroll`. This can be particularly useful to test the creation of user accounts in test scenarios since the creation of space and specifically projects involve the time-consuming deployment of nodes inside the Orchestrator.

# Feature

The purpose of this feature is add an option to `ockam enroll` so that only the first step is executed:
```
ockam enroll --user-account-only
```
(please @mrinalwadhwa propose another flag name if you have a better idea)

# Implementation notes

 - the `Enroll` command and its options is defined [here](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/enroll/command.rs#L40).
 - the code retrieving and / or creating a `Space` and a `Project` is defined [`here`](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/src/enroll/command.rs#L86). Note that we just want to skip space/project creation. If they already exist for a given user they should still be retrieved and be set as its defaults

---

<i>We love helping new contributors! ❤️
If you have questions or need help as you explore, please join us on [Discord](https://discord.gg/bewvnm6zqS). If you're looking for other issues to contribute to, please checkout our [good first issues](https://github.com/build-trust/ockam/labels/good%20first%20issue).</i>

aleksandarskrbic:
I would like to take this one :)

etorreborre:
@aleksandarskrbic all yours!

polvorin:
hi @aleksandarskrbic  :wave:  are you still interesting on working on this issue?,  do you need guidance?

nazmulidris:
Hi @aleksandarskrbic 👋 . Thank you, since you’ve not had a chance we’re opening the issue up for other contributors.

haruska:
I was going to take a stab at this as a first contribution. If someone else is attempting, let me know.

nazmulidris:
[...]

PR #7598, opened 2024-02-15
feat(rust): add `--if-not-enrolled` flag to `enroll` command

adrianbenavides:
When this flag is passed, we check if the identity is enrolled and if we have the user info available. In that case, we skip the OIDC process.

I've also removed some checks to force getting the spaces/projects from the cloud instead of the local database so we always get an up-to-date state of those resources.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/enrollments.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/enrollments.rs
@@ -31,6 +31,21 @@ impl CliState {
             .await?)
     }
 
+    #[instrument(skip_all)]
+    pub async fn identity_should_enroll(&self, name: &Option<String>, force: bool) -> Result<bool> {
+        if force {
+            return Ok(true);
+        }
+
+        // Force enrollment if there are no spaces or projects in the database
+        if self.get_spaces().await?.is_empty() || self.get_projects().await?.is_empty() {
+            return Ok(true);
+        }
+
+        // Force enrollment if the identity is not enrolled
+        Ok(!self.is_identity_enrolled(name).await?)
+    }
+
     #[instrument(skip_all, fields(identifier = %identifier))]
     pub async fn set_identifier_as_enrolled(&self, identifier: &Identifier) -> Result<()> {
         Ok(self
@@ -165,3 +180,6 @@ impl EnrollmentTicket {
         Ok(hex::encode(serialized))
     }
 }
+
+#[cfg(test)]
+mod tests {}
--- a/implementations/rust/ockam/ockam_api/src/cloud/space.rs
+++ b/implementations/rust/ockam/ockam_api/src/cloud/space.rs
@@ -6,6 +6,7 @@ use ockam_core::api::Request;
 use ockam_core::async_trait;
 use ockam_node::Context;
 
+use crate::cloud::project::{Project, Projects};
 use crate::cloud::{ControllerClient, HasSecureClient};
 use crate::nodes::InMemoryNode;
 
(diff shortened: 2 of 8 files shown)


implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -49,6 +47,10 @@ pub struct EnrollCommand {
     #[arg(long)]
     pub authorization_code_flow: bool,
 
+    /// Skip the enrollment process if the Identity is already enrolled
+    #[arg(long)]
```
etorreborre, line 55: It would good to explain in the long help, when and why this argument should be used. 
Also the name is a bit confusing since it forces you to think in terms of double negation. I would understand better something like `force-enrollment` or simply `force`.

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -190,6 +159,62 @@ impl EnrollCommand {
...
     }
+
+    async fn enroll_identity(
+        &self,
+        ctx: &Context,
+        opts: &CommandGlobalOpts,
+        node: &InMemoryNode,
+    ) -> miette::Result<UserInfo> {
+        // Force enrollment if there are no spaces or projects
+        let force_enrollment = opts.state.get_spaces().await?.is_empty()
```
etorreborre: This piece of logic, for deciding if enrollment should happen or not, could go into the `CliState` (with a parameter to force the enrollment or not).

implementations/rust/ockam/ockam_command/src/enroll/command.rs
adrianbenavides: Actually, the "correct" name for the argument would be "skip_user_enrollment_process". Also the "user_account_only" should be renamed to "skip_orchestrator_resources_creation" to be more precise.

Now, I'm not sure how to name them from the user perspective. Perhaps `--skip-user-enrollment` and `--skip-resources-creation`? The "force" behaviour is the default right now, which just goes through all the proceses.

etorreborre: What if `force = false` was the default then? It seems to me that, by default, we don't want enrollment to happen if it's unnecessary, unless you truly want to re-enroll, possibly to use a different account (is there another use-case?). Then `force` makes sense to me.

I think that the second flag can be called `--skip-resource-creation`. I forgot that [I was the one proposing this feature ](https://github.com/build-trust/ockam/issues/5452) :-). This flag should probably stay hidden because, outside of our own testing, I'm not sure who would want to use it.

As usual naming is a bit subjective, so that's only my take on it :-).

implementations/rust/ockam/ockam_command/src/enroll/command.rs
adrianbenavides: Done!

implementations/rust/ockam/ockam_command/src/enroll/command.rs
adrianbenavides: I've added this change in a separate commit. I think the UX is better this way, as we avoid prompting the user to do an action outside of the terminal when it's not necessary.

implementations/rust/ockam/ockam_api/src/cli_state/enrollments.rs
```diff
@@ -31,6 +31,25 @@ impl CliState {
...
+            return Ok(true);
+        }
+
+        // Force enrollment if there are no spaces or projects in the database
+        if self.get_spaces().await?.is_empty() || self.get_projects().await?.is_empty() {
+            return Ok(true);
+        }
+
+        // Force enrollment if the identity is not enrolled
+        if !self.is_identity_enrolled(name).await? {
```
etorreborre: Those last lines could be simply:
```rust
Ok(!self.is_identity_enrolled(name).await?)
```

implementations/rust/ockam/ockam_api/src/cloud/space.rs
```diff
@@ -112,6 +113,18 @@ impl Spaces for InMemoryNode {
 
     #[instrument(skip_all, fields(space_id = space_id))]
     async fn delete_space(&self, ctx: &Context, space_id: &str) -> miette::Result<()> {
+        let space_projects = self
```
etorreborre, line 116: 💯. Ideally, we should have some tests for this kind of behavior that is distributed across the local state and the Controller. Since we have an interface for the Controller, that should be doable.

etorreborre (dismissed):
💮

etorreborre (approved):

PR #7606, opened 2024-02-19
feat: enable tracing by default

etorreborre:
This PR enables tracing by default.

The biggest changes consists in:

 1. Checking if the OpenTelemetry collector endpoint is accessible before deciding to enable tracing.
 2. Working around some possible deadlocks in the tracing library around shutdown behaviour.
 3. Refactoring the setup of commands so that
    - The setup code is easier to navigate.
    - Environment variables are only read once to build the logging/tracing configuration.

**The `OCKAM_TRACING` environment variable is to set to `false` by default in that PR. Once we will have tested tracing well enough, I will open a PR to set it to `true`**.

## Endpoint accessibility

We want to both enable tracing by default, and make sure that the command line stays responsive if the endpoint is not accessible or misconfigured.

Tracing is only enabled if the endpoint can be accessed with a TCP connection:

 - A default connection timeout of `500ms` is used. This is the result of experimenting with
     - An accessible endpoint, which should shown as available in that time or earlier.
     - An inaccessible endpoint, which should be shown as unavailable in that time or earlier.
 - If `quiet` is set to `true` then a message is displayed when tracing is not available.
   
## Deadlocks

The deadlock issue is opened [here](https://github.com/open-telemetry/opentelemetry-rust/issues/1395). The workaround in this PR is to have a `LogProcessor` and a `SpanProcessor` which just flush on shutdown.

### `ockam_api::config`

That module has been refactored to:

 - Remove the unused code for `atomic`, `ConfigValues`, `Config`.
 - Put in `ockam_api::config` some environment variables types:
     - `DurationVar` to parse `Durations`.
     - `UrlVar` to parse `Url`s.
     - `BooleanVar` to parse boolean values.
     - `LevelVar` to parse log levels.

### `ockam_api::logs`

That module has been refactored into many other modules:

- `current_span` convenience functions to set attributes on the current span
- `default_values` constant values used by default to configure logging / tracing
- `env_variables` list of environment variables which can be used to configure logging / tracing
- `log_exporters` (resp. `span_exporters`) implementations for the `LogExporter` trait. So far there is only a `DecoratedLogExporter` implementation which is useful to quickly intercept exported logs
- `log_processors` (resp. `span_processors`) implementations for the `LogProcessor` trait. It contains a `NonBlockingLogProcessor` which decorates any `LogProcessor` to avoid doing a shutdown, since this can trigger deadlocks
- `logging_configuration` data type describing all the possible values for configuring logging and where to get them from.
- `logging_options` data types describing some configuration options for logging
- `setup` main place where we configure the logging and tracing layers
- `tracing_guard` maintains the logging / tracing resources alive while a command is executed, or a background node running.

### `ockam_command`

The top-level code for creating and running an `OckamCommand` was becoming too bloated. The following files have been created / modified:

- `arguments` contains helper methods to parse input arguments
- `command_events` creates user events when a command executes
- `command_global_opts` extracts in its own file the `CommandGlobalOpts` data type
- `global_args` extracts in its own file the `GlobalArgs` data type 
- `ockam_command` create the `CommandGlobalOpts`, sets up logging/tracing, and run the subcommand
- `ockam_subcommand` enum with all the top level commands: `node`, `project`, `identity`, etc...


examples/rust/get_started/src/token.rs
```diff
@@ -20,6 +20,7 @@ pub async fn create_token(attribute_name: &str, attribute_value: &str) -> Result
             format!("{attribute_name}={attribute_value}").as_str(),
         ])
         .env_remove("OCKAM_LOG") // make sure that OCKAM_LOG is not set, otherwise the output will contain more than the token
+        .env_remove("OCKAM_LOGGING") // make sure that OCKAM_LOG is not set, otherwise the output will contain more than the token
```
adrianbenavides: typo in the comment

implementations/rust/ockam/ockam_api/src/config/boolean_var.rs
```diff
@@ -0,0 +1,19 @@
...
+    fn from_string(s: &str) -> ockam_core::Result<Self> {
+        Ok(BooleanVar(s.to_string()))
+    }
+}
+
+impl BooleanVar {
+    pub fn is_true(&self) -> bool {
+        !FALSE_VALUES.contains(&self.0.as_str())
+    }
+}
```
adrianbenavides: Perhaps for another PR, but we could move these types to `ockam_core::env`? We have some code there to parse bools/strings/etc

implementations/rust/ockam/ockam_api/src/logs/default_values.rs
```diff
@@ -0,0 +1,49 @@
...
+
+/// Maximum size in Mb for a log file
+pub(crate) const DEFAULT_LOG_MAX_SIZE_MB: u64 = 100;
+
+/// Maximum number of files to create for a given node
+pub(crate) const DEFAULT_LOG_MAX_FILES: u64 = 60;
+
+/// Default endpoint for the OpenTelemetry collector
+pub(crate) const DEFAULT_OTEL_EXPORTER_OTLP_ENDPOINT: &str =
+    "http://k8s-opentele-otelcoll-aa527132c8-70cbeef1b85b559b.elb.us-west-1.amazonaws.com:4317/";
```
adrianbenavides, line 19: Is there any risk if we expose this endpoint? Could anyone point to it and send random data? Just curious

implementations/rust/ockam/ockam_api/src/logs/logging_options.rs
```diff
@@ -0,0 +1,90 @@
...
+        match self {
+            LoggingEnabled::On => f.write_str("on"),
+            LoggingEnabled::Off => f.write_str("off"),
+        }
+    }
+}
+
+impl FromString for LoggingEnabled {
+    fn from_string(_s: &str) -> ockam_core::Result<Self> {
+        Ok(LoggingEnabled::On)
```
adrianbenavides: Why is this hardcoded to On? Do we need to implement this anyways?

implementations/rust/ockam/ockam_api/src/logs/span_exporters.rs
```diff
@@ -0,0 +1,27 @@
...
+/// This exporter can be used to intercept the spans sent to an OpenTelemetry collector
+#[derive(Debug)]
+struct DecoratedSpanExporter<S: SpanExporter> {
+    exporter: S,
+}
+
+#[async_trait]
+impl<S: SpanExporter> SpanExporter for DecoratedSpanExporter<S> {
+    fn export(&mut self, batch: Vec<SpanData>) -> BoxFuture<'static, ExportResult> {
+        // debug!("exporting {} spans", batch.len());
```
adrianbenavides: To be removed?

implementations/rust/ockam/ockam_api/src/logs/tracing_options.rs
```diff
@@ -0,0 +1,24 @@
...
+        match self {
+            TracingEnabled::On => f.write_str("on"),
+            TracingEnabled::Off => f.write_str("off"),
+        }
+    }
+}
+
+impl FromString for TracingEnabled {
+    fn from_string(_s: &str) -> ockam_core::Result<Self> {
+        Ok(TracingEnabled::On)
```
adrianbenavides: Same question as with the `LoggingEnabled`.

implementations/rust/ockam/ockam_api/tests/journeys.rs
```diff
@@ -40,7 +40,7 @@ fn test_create_journey_event() {
                 let cli = CliState::create(cli_state_directory)
                     .await
                     .unwrap()
-                    .set_tracing_enabled();
+                    .set_tracing_enabled(true);
```
adrianbenavides, line 43: I'd rather have an `enable_tracing()/disable_tracing()` pair of functions, but I'm not sure we are following that pattern anywhere in the codebase.

implementations/rust/ockam/ockam_command/src/arguments.rs
```diff
@@ -0,0 +1,44 @@
+/// Return true if the list of arguments contains a help flag
+pub fn has_help_flag(input: &[String]) -> bool {
+    !input.contains(&"-h".to_string()) && !input.contains(&"--help".to_string())
```
adrianbenavides: Seeing how this function is used, I think the name of the function is the opposite of what it does right?

implementations/rust/ockam/ockam_command/src/ockam_run.rs
```diff
@@ -0,0 +1,46 @@
+use clap::Parser;
```
adrianbenavides: The `ockam_command` and `ockam_run` file names are confusing to me. Just reading the names, I'd say the main file is `ockam_command` since it's named after the crate.

Maybe `ockam_run` could be named `entry_point` and `ockam_command` just `command`. We can discuss it separately and do it in another PR.

implementations/rust/ockam/ockam_api/src/logs/default_values.rs
etorreborre: Yes, it is opened for now. The next step will be to export data using Ockam (as soon as a project exist, so that we can make a secure channel).

examples/rust/get_started/src/token.rs
etorreborre: done

implementations/rust/ockam/ockam_api/src/config/boolean_var.rs
etorreborre: I'm going to refactor in this PR since I'm introducing new unnecessary stuff.

implementations/rust/ockam/ockam_api/src/logs/logging_options.rs
etorreborre: I'll fix it to read the value from `FromString<bool>`

implementations/rust/ockam/ockam_command/src/arguments.rs
etorreborre: Totally right!

implementations/rust/ockam/ockam_command/src/ockam_run.rs
etorreborre: I agree. I renamed to `entry_point`, `command` and `subcommand` (for `ockam_subcommand`).

implementations/rust/ockam/ockam_api/src/logs/tracing_options.rs
etorreborre: Same treatment :-).

implementations/rust/ockam/ockam_api/tests/journeys.rs
etorreborre: This is useful because we pass a boolean in `CommandGlobalOpts`:
```rust
 state.set_tracing_enabled(tracing_configuration.is_enabled()),
```
Otherwise we would have to surround with a more verbose `if/else` expression.

etorreborre: One thing to consider in a future refactoring would be to put enough config data directly when creating `CliState` rather than modifying it after creating it. I didn't do that because I would have touched more places than I wanted to touch in this PR.

adrianbenavides (approved):

PR #7615, opened 2024-02-19
Accepted contribution terms Ockam 

YorickdeJong:


.github/CONTRIBUTORS.csv
```diff
@@ -1,4 +1,5 @@
 Accepted CLA,Github User,Name,Emails
+yes,YorickdeJong,Yorick de Jong,<info@bluewaterfall.nl>
```
adrianbenavides: Can you move the new entry to the bottom of the list, please? That way we can keep the list in insertion order.

adrianbenavides (dismissed):

PR #7616, opened 2024-02-19
feat(rust): update enroll ux with new help text, display, and log progress status messages for provisioning identity and vault

nazmulidris:
Issue/epic: https://github.com/build-trust/codebase/issues/737

## Screenshots
`rm -rf ~/.ockam && ockam enroll`
Image

`ockam enroll` (this this again)
Image

`ockam enroll --identity my_id` (my_id does not exist)
Image

`ockam identity create my_id && ockam enroll --identity my_id`
Image
Image

UX String docs:
- [ockam enroll UX strings](https://docs.google.com/document/d/1rbJCBaZRihy4GVr_Ei1V5zCrNk3vOfQ1Z8umKVz6OU8/edit)
- [ockam project ticket|enroll UX strings](https://docs.google.com/document/d/1oLOYIAom9qFYjsAcjRd3_pgU9ukxaVrMF7c9vtfWwZc/edit#heading=h.o1swyzd1zhy1)
- [ockam tcp-outlet UX strings](https://docs.google.com/document/d/1NlJiJNDpjxZQUr_j5aV5CvQR_izuoKpnvXeBP059Kqs/edit#heading=h.o1swyzd1zhy1)

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
@@ -1,5 +1,6 @@
 use std::path::{Path, PathBuf};
 
+use colorful::{Colorful, RGB};
 use rand::random;
 
 use cli_state::error::Result;
@@ -7,10 +8,11 @@ use ockam::SqlxDatabase;
 use ockam_core::env::get_env_with_default;
 use ockam_node::database::application_migration::ApplicationMigration;
 use ockam_node::Executor;
+use tokio::sync::broadcast::{channel, Sender};
 
-use crate::cli_state;
-use crate::cli_state::CliStateError;
+use crate::cli_state::{self, CliStateError};
 use crate::logs::TracingEnabled;
+use crate::{ReportingChannelMessageType, REPORTING_CHANNEL_CAPACITY};
 
 /// The CliState struct manages all the data persisted locally.
 ///
@@ -41,6 +43,14 @@ pub struct CliState {
     database: SqlxDatabase,
     application_database: SqlxDatabase,
     tracing_enabled: TracingEnabled,
+    pub(crate) progress_bar_reporting_channel_sender: Sender<ReportingChannelMessageType>,
+}
+
+pub fn color_primary(text: &str) -> String {
+    // This is a copy of `OckamColor::PrimaryResource` which is the `ockam_command` crate.
+    // #4FDAB8 is rgb(79,218,184).
+    let color = RGB::new(79, 218, 184);
+    text.color(color).to_string()
 }
 
 impl CliState {
@@ -168,6 +178,8 @@ impl CliState {
             "Opened the application database with options {:?}",
             application_database
         );
+        let (progress_bar_reporting_channel_sender, _) =
+            channel::<ReportingChannelMessageType>(REPORTING_CHANNEL_CAPACITY);
         let state = Self {
             dir,
             database,
@@ -177,6 +189,7 @@ impl CliState {
             // the function set_tracing_enabled can be used to enable tracing, which
             // is eventually used to trace user journeys.
             tracing_enabled: TracingEnabled::Off,
+            progress_bar_reporting_channel_sender,
         };
         Ok(state)
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4236,6 +4236,7 @@ dependencies = [
  "bytes 1.5.0",
  "cddl-cat",
  "chrono",
+ "colorful",
  "either",
  "fake",
  "fs2",
--- a/implementations/rust/ockam/ockam_api/Cargo.toml
+++ b/implementations/rust/ockam/ockam_api/Cargo.toml
@@ -34,6 +34,7 @@ aws-config = { version = "1.1.6", default-features = false, features = ["rustls"
 base64-url = "2.0.2"
 bytes = { version = "1.5.0", default-features = false, features = ["serde"] }
 chrono = { version = "0.4" }
+colorful = "0.2"
 either = { version = "1.10.0", default-features = false }
 fs2 = { version = "0.4.3" }
 futures = { version = "0.3.30", features = [] }
--- a/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
@@ -4,7 +4,10 @@ use ockam_core::errcode::{Kind, Origin};
 use ockam_core::Error;
 use ockam_vault::{HandleToSecret, SigningSecretKeyHandle};
 
-use crate::cli_state::{random_name, CliState, Result};
+use crate::{
+    cli_state::{random_name, CliState, Result},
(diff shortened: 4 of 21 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
```diff
@@ -41,6 +41,32 @@ pub struct CliState {
...
+    tx: Option<tokio::sync::broadcast::Sender<ChannelMessageType>>,
+}
+
+const DEFAULT_CHANNEL_CAPACITY: usize = 16;
+pub type ChannelMessageType = String;
+pub const CHANNEL_POISON_PILL_MESSAGE: &str = "exit";
+pub const CHANNEL_MESSAGE_DISPLAY_DELAY_MS: u64 = 1000;
+
+// Broadcast channel support.
+impl CliState {
```
etorreborre: Since this is a new concern for the `CliState` I'd rather see it in its own file.

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
```diff
@@ -41,6 +41,32 @@ pub struct CliState {
     database: SqlxDatabase,
     application_database: SqlxDatabase,
     tracing_enabled: TracingEnabled,
+    tx: Option<tokio::sync::broadcast::Sender<ChannelMessageType>>,
```
etorreborre: This deserves a more explicit name than `tx` :-).
Also why not import `tokio::sync::broadcast` at the top of the file?

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
```diff
@@ -41,6 +41,32 @@ pub struct CliState {
...
+}
+
+const DEFAULT_CHANNEL_CAPACITY: usize = 16;
+pub type ChannelMessageType = String;
```
etorreborre: This should be an enum eventually, and `exit` one of the cases.

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
```diff
@@ -41,6 +41,32 @@ pub struct CliState {
...
+pub const CHANNEL_POISON_PILL_MESSAGE: &str = "exit";
+pub const CHANNEL_MESSAGE_DISPLAY_DELAY_MS: u64 = 1000;
```
etorreborre: This should be a `Duration`

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
```diff
@@ -41,6 +41,32 @@ pub struct CliState {
...
+
+// Broadcast channel support.
+impl CliState {
+    pub fn open_channel(&mut self) -> tokio::sync::broadcast::Sender<ChannelMessageType> {
```
etorreborre: I would expect the channel to be opened for the life of the `CliState` and be closed only when `CliState` is dropped. Right now we have various places where the channel is closed, which is going to kill the channel. This is going to create issues if we call `get_named_identity_or_default` twice for example.

I would propose instead to:

 - Start the broadcast channel when `CliState` is started
 - Close it when `CliState` is dropped
 - When running a command, concurrently race that command + a small delay with a process that displays event as they arrive.
    - When the command + delay finished, we stop both computations
    - The spinner / display computation runs indefinitely, with no need for a poison pill. It just stops when the command has finished executing.

etorreborre (commented):
@nazmulidris I left some comments which are more indications for when we want to move towards a real PR for this feature.

More importantly I think we don't have to use a poison pill and we should only race command + a delay with an infinite spinner. Please give it a go and tell me if that works.

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
adrianbenavides: Another thing I would keep in mind is having a way to support processing the messages in different ways:
1. Overwrite the spinner message
2. Close the previous spinner to persist the previous message and print the current one in a new spinner
3. Print the received message without any spinner
4. Ideally, we would also need another kind of message to print an error message and exit the program

With these kinds of messages we would cover most, if not all, the types of messages we show to the user.

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
nazmulidris: done

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
nazmulidris: done

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
nazmulidris: agreed; also removed "exit", since im not closing channels anymore

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
nazmulidris: done

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
nazmulidris: > I would expect the channel to be opened for the life of the `CliState` and be closed only when `CliState` is dropped. Right now we have various places where the channel is closed, which is going to kill the channel. This is going to create issues if we call `get_named_identity_or_default` twice for example.
> 
> I would propose instead to:
> 
> * Start the broadcast channel when `CliState` is started
> * Close it when `CliState` is dropped
> * When running a command, concurrently race that command + a small delay with a process that displays event as they arrive.
>   
>   * When the command + delay finished, we stop both computations
>   * The spinner / display computation runs indefinitely, with no need for a poison pill. It just stops when the command has finished executing.

I did as you suggested and it worked :)

nazmulidris: > Another thing I would keep in mind is having a way to support processing the messages in different ways:
> 
> 1. Overwrite the spinner message
> 2. Close the previous spinner to persist the previous message and print the current one in a new spinner
> 3. Print the received message without any spinner
> 4. Ideally, we would also need another kind of message to print an error message and exit the program
> 
> With these kinds of messages we would cover most, if not all, the types of messages we show to the user.

Great suggestion. We can move these concerns into another PR to generalize the progress reporting and applying it more broadly to all commands.

nazmulidris:
> @nazmulidris I left some comments which are more indications for when we want to move towards a real PR for this feature.
> 
> More importantly I think we don't have to use a poison pill and we should only race command + a delay with an infinite spinner. Please give it a go and tell me if that works.

Thanks for the comments. I've addressed all of them. Adrian had some great suggestions about more generalized use cases for this, which can all go into a different PR after the release is done.

adrianbenavides (approved):

PR #7633, opened 2024-02-22
feat(rust): refactor `Project`-related code

SanjoDeundiak:
- split project model and project struct
 - introduce `Projects` struct with logic
 - store project's and authority's change history when storing a project


implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -212,9 +212,10 @@ impl EnrollCommand {
 
         // Enroll the identity with the Orchestrator
         let controller = node.create_controller().await?;
-        enroll_with_node(&controller, ctx, token)
+        let res = enroll_with_node(&controller, ctx, token)
             .await
-            .wrap_err("Failed to enroll your local Identity with Ockam Orchestrator")?;
+            .wrap_err("Failed to enroll your local Identity with Ockam Orchestrator");
+        res?;
```
adrianbenavides: Isn't this the same as before?

implementations/rust/ockam/ockam_api/src/cli_state/projects.rs
```diff
@@ -1,56 +1,103 @@
...
+            self.identities_verification
+                .update_identity_ignore_older(project_identity)
+                .await?;
+        }
+
+        if let Ok(authority_identity) = project.authority_identity() {
+            self.identities_verification
+                .update_identity_ignore_older(authority_identity)
+                .await?;
+        }
```
adrianbenavides, line 50: Could we add a test for this?

implementations/rust/ockam/ockam_api/src/cloud/project/projects.rs
```diff
@@ -0,0 +1,62 @@
+use crate::cloud::project::models::OrchestratorVersionInfo;
+use crate::cloud::project::Project;
+use ockam_core::async_trait;
+use ockam_node::Context;
+
+#[async_trait]
+pub trait Projects {
```
adrianbenavides: Perhaps this could be named `ProjectsOrchestratorApi`.

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
```diff
@@ -12,7 +12,7 @@ use ockam_abac::{Action, Expr, Resource, ResourceType};
 use ockam_core::api::{Error, Reply, Request, RequestHeader, Response};
 use ockam_core::errcode::{Kind, Origin};
 use ockam_core::{async_trait, route, AsyncTryClone, Route};
-use ockam_multiaddr::proto::Project;
+use ockam_multiaddr::proto::Project as ProjectProto;
```
adrianbenavides, line 15: We could do that rename entirely. So many things called just `Project` 😆

implementations/rust/ockam/ockam_api/src/cloud/project/models.rs
```diff
@@ -0,0 +1,296 @@
...
+use crate::cloud::share::{RoleInShare, ShareScope};
+use crate::minicbor_url::Url;
+use minicbor::{Decode, Encode};
+use ockam::identity::Identifier;
+use serde::{Deserialize, Serialize};
+use std::fmt::{Display, Formatter};
+
+#[derive(Encode, Decode, Serialize, Deserialize, Debug, Default, Clone, Eq, PartialEq)]
+#[cbor(map)]
+pub struct ProjectModel {
```
adrianbenavides, line 12: As we get closer to CQRS/DDD, perhaps we could start introducing more concepts to disambiguate between domain and application layers, especially in this crate. Maybe this could be named `ProjectViewModel`?

implementations/rust/ockam/ockam_api/src/cli_state/projects.rs
SanjoDeundiak: @adrianbenavides oh, nice that you caught it. I actually created a test for this but I didn't commit it accidentally. Thank you

implementations/rust/ockam/ockam_api/src/cloud/project/models.rs
SanjoDeundiak: Could you please elaborate why it deserves `View` part? After this refactoring the only one responsibility of this struct is to parse Controller's response.

implementations/rust/ockam/ockam_api/src/cloud/project/projects.rs
SanjoDeundiak: Yeah, I just moved it as is, but renaming is a good suggestion

SanjoDeundiak: We have a lot of similar names that we need to change

implementations/rust/ockam/ockam_api/src/nodes/service/portals.rs
SanjoDeundiak: Then I'll need to rename all the other protos... Let's either not do it now or create a separate PR

implementations/rust/ockam/ockam_command/src/enroll/command.rs
SanjoDeundiak: Yeah, I'll revert it

implementations/rust/ockam/ockam_api/src/cloud/project/models.rs
adrianbenavides: This is subjective and open to discussion, but "model" feels too generic to me, and in some paradigms a model can contain logic. The "view" part can enforce that this model is just data with no logic.

adrianbenavides (approved):
🚀

PR #7635, opened 2024-02-22
Major optimizations

davide-baldo:
In this PR, I address the slowest pieces in ockam portals, extra memory copies and slow serialization/deserialization with serde bare.
The Routed no longer contains the decoded body, but gets decoded only when `Routed::into_body()` is called. This avoid an extra copy of the same message both in form of deserialized and serialized form. (this also reduces the peak memory usage).

Every BARE message involved with TCP portals is now manually serialized/deserilized, in specific:
- TCP portal processor/worker 
- Ockam TCP messaging protocol (`TransportMessage`)

TCP Portal message and Secure Channel messages, where possible/beneficial, are  now borrow buffer rather than owning the `Vec<u8>` to avoid extra copies (~zero copy deserialization).

The encoding/decoding primitives for serde were implemented with `no_std` within the `ockam_core` crate.

The secure channel encryptor/decryptor were already using CBOR instead (which is much faster than `serde_cbor`,  manually serializing/deserializing doesn't make much sense), but there is an extra serde encoding of a `Vec<u8>`. To avoid the serde-encoding pass and an extra buffer copy, a hack was implemented. Since the serialization of a `Vec<u8>` in bare is just the length followed the raw buffer, by calculating beforehand the length of the resulting encrypted data we can directly encrypt inside the same buffer. We have to calculate the encrypted length beforehand since the length of the array in bare is variable depending on the value. @SanjoDeundiak I'm NOT sure if we can actually assume the encrypted size is always payload + overhead, and that the overhead is a static number.

```
amd64 laptop - 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz

before optimizations:
one node portal: 7.16 GBytes  1.02 Gbits/sec
two nodes portal: 5.73 GBytes   819 Mbits/sec
message latency: 101.066µs

after optimizations:
one node portal: 42.2 GBytes  6.04 Gbits/sec - delta: ~5.9x
two nodes portal: 40.8 GBytes  5.83 Gbits/sec - delta: ~7.1x
message latency: 86.445µs - delta: ~0.855x
```

```
arm64 laptop - Apple M2 Pro 12 Cores

before optimizations:
one node portal: 4.72 GBytes   675 Mbits/sec
two nodes portal: 4.84 GBytes   692 Mbits/sec
message latency: 91.512µs

after optimizations:
one node portal: 7.10 GBytes  1.02 Gbits/sec - delta: ~1.7x
two nodes portal: 7.01 GBytes  1.00 Gbits/sec - delta: ~1.44x
message latency: 82.323µs - delta: ~0.90x
```


implementations/rust/ockam/ockam_core/src/bare.rs
```diff
@@ -0,0 +1,154 @@
+#![allow(missing_docs)]
```
SanjoDeundiak, line 1: Are there missing docs in that file?

SanjoDeundiak: Are there missing docs in that file?

implementations/rust/ockam/ockam_core/src/bare.rs
```diff
@@ -0,0 +1,154 @@
...
+
+/// Defines a set of functions to encode and decode bare encoding
+///
+/// This module is not dependent on std or any other crate
+use crate::compat::vec::Vec;
+
+/// Read a dynamically sized slize from the given cursor
+pub fn read_slice<'de>(slice: &'de [u8], index: &mut usize) -> Option<&'de [u8]> {
+    let length = read_variable_length_integer(slice, index)?;
```
SanjoDeundiak: You can cast it to usize right away to avoid multiple casts below

implementations/rust/ockam/ockam_core/src/bare.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+    result
+}
+
+/// Read a variable length integer from the given cursor (ULEB128)
+/// returns None if the buffer is too short
+pub fn read_variable_length_integer(slice: &[u8], index: &mut usize) -> Option<u64> {
+    let mut result = 0;
+    let mut shift = 0;
+    loop {
```
SanjoDeundiak, line 66: Should we check for overflows in that loop?

SanjoDeundiak: And I wonder if minicbor or serde_bare have an implementation for that somewhere that we could just copy... or is that already it?

implementations/rust/ockam/ockam_core/src/bare.rs
```diff
@@ -0,0 +1,154 @@
...
+            byte |= 0b1000_0000;
+        }
+        destination.push(byte);
+        if value == 0 {
+            break;
+        }
+    }
+}
+
+mod test {
```
SanjoDeundiak, line 99: Fuzzy testing would be very useful here

implementations/rust/ockam/ockam_core/src/message.rs
```diff
@@ -279,53 +272,14 @@ impl<M: Message> Routed<M> {
...
 impl<M: Message + Debug> Debug for Routed<M> {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        self.inner.fmt(f)
-    }
-}
-
-impl<M: Message + Display> Display for Routed<M> {
-    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        self.inner.fmt(f)
+        self.phantom.fmt(f)
```
SanjoDeundiak: What does it print? I'm honestly curious

implementations/rust/ockam/ockam_core/src/routing/address.rs
```diff
@@ -284,38 +284,75 @@ impl Distribution<Address> for Standard {
...
+            addr,
+            Address {
+                tt: TransportType::new(1),
+                inner: "remote_friend".as_bytes().to_vec(),
+            }
+        );
+    }
+
+    #[test]
+    #[should_panic(expected = "Failed to parse address type:")]
```
SanjoDeundiak, line 352: Should never panic honestly, but I wonder how big of a change this would be

implementations/rust/ockam/ockam_core/src/routing/message/transport_message.rs
```diff
@@ -125,3 +124,139 @@ impl Display for TransportMessage {
         )
     }
 }
+
+impl Encodable for TransportMessage {
+    fn encode(self) -> crate::Result<Encoded> {
+        let mut encoded = Vec::new();
```
SanjoDeundiak: Most efficient way would be to first calculate the resulting length, and create the vector with that capacity

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor.rs
```diff
@@ -70,11 +70,10 @@ impl Encryptor {
             .aead_encrypt(&self.key, payload, &nonce, &[])
             .await?;
 
-        let mut res = Vec::new();
-        res.extend_from_slice(&small_nonce);
-        res.append(&mut cipher_text);
+        destination.extend_from_slice(&small_nonce);
```
SanjoDeundiak: Would be cool to reserve the space first

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -28,7 +28,8 @@ pub(crate) struct SecureChannelSharedState {
 }
 
 pub(crate) struct EncryptorWorker {
-    role: &'static str, // For debug purposes only
```
SanjoDeundiak: Let's move it back :) It's about `role` field

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
```diff
@@ -60,7 +60,7 @@ pub(crate) struct KafkaPortalWorker {
 
 #[ockam::worker]
 impl Worker for KafkaPortalWorker {
-    type Message = PortalMessage;
+    type Message = NeutralMessage;
```
SanjoDeundiak, line 63: For a long time I have a feeling that workers should not have a message type declared at all... WDYT?

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
```diff
@@ -78,7 +78,9 @@ impl Worker for KafkaPortalWorker {
         let onward_route = routed_message.onward_route();
         let return_route = routed_message.return_route();
         let local_info = routed_message.local_message().local_info().to_vec();
```
SanjoDeundiak: BTW, that's calling `.to_vec()` on a `Vec` instance

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
```diff
@@ -78,7 +78,9 @@ impl Worker for KafkaPortalWorker {
...
-        let portal_message = routed_message.as_body();
+        // TODO: avoid cloning, rebuild the message when necessary
+        let body = routed_message.payload().to_vec();
```
SanjoDeundiak: I think you can avoid `.to_vec()` here

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
```diff
@@ -255,16 +257,13 @@ impl KafkaPortalWorker {
...
     ) -> Result<Option<Bytes>, InterceptError> {
         let mut encoded_buffer: Option<BytesMut> = None;
 
         for complete_kafka_message in self
             .decoder
-            .decode_messages(
-                BytesMut::from(encoded_message.as_slice()),
-                self.max_message_size,
-            )
+            .decode_messages(BytesMut::from(encoded_message), self.max_message_size)
```
SanjoDeundiak: I suspect that it's possible to avoid cloning here. `decode_messages` should be completely fine having its input as slice. Also, if we reserve few additional bytes in the result of this function, we may optimize `length_encode` to be allocation free. (See `encode_transport_message` for context)

SanjoDeundiak: Also, can we pick a better name for the function?

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
```diff
@@ -255,16 +257,13 @@ impl KafkaPortalWorker {
...
             .map_err(InterceptError::Ockam)?
         {
             let transformed_message = match self.receiving {
```
SanjoDeundiak, line 267: Is there a reason to return a value from that function? If I'm not mistaken, we always return the argument back

SanjoDeundiak: Also, would it be possible to return messages from this function one by one instead of putting them in one large buffer?

implementations/rust/ockam/ockam_core/src/message.rs
```diff
@@ -203,21 +203,20 @@ impl<M: Message> Routed<M> {
     /// Create a new `Routed` message wrapper from the given message,
     /// message address and a local message that contains routing
     /// information.
-    pub fn new(inner: M, msg_addr: Address, src_addr: Address, local_msg: LocalMessage) -> Self {
+    pub fn new(
+        phantom: PhantomData<M>,
```
SanjoDeundiak: Let's remove that argument

implementations/rust/ockam/ockam_transport_tcp/src/portal/portal_message.rs
```diff
@@ -161,8 +162,7 @@ mod test {
         let decoded = PortalMessageV1::decode(&encoded).unwrap();
         assert!(matches!(decoded, PortalMessageV1::Disconnect));
 
-        let encoded =
-            PortalMessage::encode(PortalMessage::Payload(payload.clone(), Some(123))).unwrap();
+        let encoded = PortalMessage::encode(PortalMessage::Payload(&payload, Some(123))).unwrap();
```
SanjoDeundiak, line 165: Do you think we should have test vectors here and in other places where you used manual encoding?

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -66,9 +67,34 @@ impl EncryptorWorker {
...
+    }
+
+    async fn encrypt_to(
+        &mut self,
+        ctx: &Context,
+        destination: &mut Vec<u8>,
+        payload: &[u8],
+    ) -> Result<()> {
+        // by reserving the capacity beforehand, we can avoid copying memory later
+        destination.reserve(8 + 8 + 8 + payload.len());
```
SanjoDeundiak: Didn't we already reserve before calling that function?

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor.rs
```diff
@@ -70,11 +70,10 @@ impl Encryptor {
             .aead_encrypt(&self.key, payload, &nonce, &[])
```
SanjoDeundiak: Let's actually pass the destination buffer to this function to avoid another allocation

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -141,16 +172,46 @@ impl EncryptorWorker {
...
+            // variable length integer followed by the raw bytes.
+            // The idea is first to calculate the size of the encrypted payload,
+            // so we can calculate the size of the variable length integer.
+            // The goal is to prepend the variable length integer to the buffer
+            // before it's actually written, so we can write the whole encrypted
+            // payload without any extra copies.
+
+            let encoded_payload = minicbor::to_vec(&msg)?;
+            // we assume this calculation is exact
+            let encrypted_payload_size = 8 + 8 + 8 + encoded_payload.len();
```
SanjoDeundiak: It's 8 bytes nonce + 16 bytes auth tag. The result is correct but let's create corresponding public constants in the `ockam_vault` crate

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -66,9 +67,34 @@ impl EncryptorWorker {
     }
 
     /// Encrypt the message
-    async fn encrypt(&mut self, ctx: &Context, msg: SecureChannelMessage) -> Result<Vec<u8>> {
-        match self.encryptor.encrypt(&minicbor::to_vec(&msg)?).await {
-            Ok(encrypted_payload) => Ok(encrypted_payload),
+    async fn encrypt(&mut self, ctx: &Context, msg: SecureChannelMessage<'_>) -> Result<Vec<u8>> {
+        let payload = minicbor::to_vec(&msg)?;
+        // capacity is nonce + tag + payload
+        let mut buffer = Vec::with_capacity(8 + 8 + payload.len());
```
SanjoDeundiak: I think it's 8 + 16

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -66,9 +67,34 @@ impl EncryptorWorker {
     }
 
     /// Encrypt the message
-    async fn encrypt(&mut self, ctx: &Context, msg: SecureChannelMessage) -> Result<Vec<u8>> {
-        match self.encryptor.encrypt(&minicbor::to_vec(&msg)?).await {
-            Ok(encrypted_payload) => Ok(encrypted_payload),
+    async fn encrypt(&mut self, ctx: &Context, msg: SecureChannelMessage<'_>) -> Result<Vec<u8>> {
```
SanjoDeundiak, line 69: What is the difference between `encrypt` and `encrypt_to`? Aren't they doing the same thing?

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -66,9 +67,34 @@ impl EncryptorWorker {
...
+    async fn encrypt_to(
+        &mut self,
+        ctx: &Context,
+        destination: &mut Vec<u8>,
+        payload: &[u8],
+    ) -> Result<()> {
+        // by reserving the capacity beforehand, we can avoid copying memory later
+        destination.reserve(8 + 8 + 8 + payload.len());
+
+        match self.encryptor.encrypt(destination, payload).await {
```
SanjoDeundiak, line 85: Can we pass the buffer here?

SanjoDeundiak (commented):
Very nice changes, will be a lot less clones. Found few more potential improvements

implementations/rust/ockam/ockam_core/src/bare.rs
davide-baldo: It's expecting a general description of the module, i can add it since now it's public :+1:

implementations/rust/ockam/ockam_core/src/bare.rs
davide-baldo: I've added a check on the shift, which indirectly counts the number of loops, we could copy the serde implementation but i think this is more readable.

implementations/rust/ockam/ockam_core/src/bare.rs
davide-baldo: i also using `try_into()`, i thought about 32bits embedded, on them a cast could silently fail

implementations/rust/ockam/ockam_core/src/bare.rs
davide-baldo: Added a fuzzy method for each one!

implementations/rust/ockam/ockam_core/src/message.rs
davide-baldo: Just phantom<data type>. I've manually implemented something a bit more verbose with extra info.

implementations/rust/ockam/ockam_core/src/routing/address.rs
davide-baldo: ~50 usages,  but 97+% are direct usages without any possibility of having a '#', hence they can't fail.
Maybe we can assert local service only ('#0') and have a dedicated call if you want to parse the protocol(should be an handful).

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor.rs
davide-baldo: The destination vector is a provided parameter, and in the common `encrypt()` flow it's already reserved.

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
davide-baldo: IDE doing weird stuff :/ fixed

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
davide-baldo: I think we could move the parameter declarion to the `message.into_body()` call and by doing so, we could also allow messages that borrows data and avoid the manual encoding/decoding calls.
On the other hand It's simpler for the avarage user to think a worker with a fixed type.
Maybe we could move the message type from the worker to the incoming addresses instead, but i don't have a clear picture on how it would look like.

implementations/rust/ockam/ockam_transport_tcp/src/portal/portal_message.rs
davide-baldo: Where manually encoding is used i've added tests to check compatibility with serde own serialization, i think it's enough to cover sane cases (if serde own encoding changes, everything would crumble). On the long run we might want to start some schema of sort.

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
davide-baldo: Yes! But another user might not reserve the needed memory, this should be a no-op in most cases.

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
davide-baldo: Yes they are, one takes a writable `Vec` argument the other returns it. Could delete the returning one and force every caller to create the destination beforehand, but it would make the code a bit less readable.

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
davide-baldo: I'm not sure what you meant

implementations/rust/ockam/ockam_core/src/bare.rs
SanjoDeundiak: Shouldn't the exit condition be `if shift == 63`?

implementations/rust/ockam/ockam_core/src/bare.rs
```diff
@@ -0,0 +1,226 @@
...
+        if shift >= 64 {
+            return None;
+        }
+    }
+
+    Some(result)
+}
+
+/// Write a variable length integer to the given buffer (ULEB128)
+pub fn write_variable_length_integer(destination: &mut Vec<u8>, mut value: u64) {
```
SanjoDeundiak, line 85: Just in case you like, you can do this way:
```rs
pub fn write_variable_length_integer(destination: &mut Vec<u8>, mut value: u64) {
    loop {
        let mut byte = (value & 0b0111_1111) as u8;
        value >>= 7;
        if value != 0 {
            destination.push(byte | 0b1000_0000);
        } else {
            destination.push(byte);
            break;
        }
    }
}
```

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
SanjoDeundiak: Why not at least call `encrypt_to` from `encrypt`?

SanjoDeundiak (approved):

PR #7643, opened 2024-02-23
feat: time-limited journey traces

etorreborre:
This PR improves the host and project journeys. Since they are represented as traces, it is desirable to limit the growth of those traces.

This PR let spans accumulate for 5 days for a given journey (not configurable for now). After that date:

 - A new project/host journey is started
 - Its top span contains a link to the previous journey so that we can navigate to the previous data
 - The journey still contains the project id / host name so that it is still possible to query for all the traces related to a given project if necessary
 - The trace id incorporates the date as `YYMMDD` so that we can quickly identify when the actions took place
 - The `YYMMDD` dates are created in 5 days buckets so that the creation of journeys is very deterministic, even 
   if user actions are executed from different nodes on different days.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/journeys/attributes.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/journeys/attributes.rs
@@ -2,11 +2,13 @@ use crate::journeys::{
     APPLICATION_EVENT_OCKAM_GIT_HASH, APPLICATION_EVENT_OCKAM_HOME, APPLICATION_EVENT_OCKAM_VERSION,
 };
 use crate::Version;
+use chrono::{DateTime, Datelike, Utc};
 use gethostname::gethostname;
-use opentelemetry::trace::TraceId;
+use opentelemetry::trace::{SpanId, TraceId};
 use opentelemetry::Key;
 use opentelemetry_sdk::trace::{IdGenerator, RandomIdGenerator};
 use std::collections::HashMap;
+use std::fmt::Write;
 use std::process::Command;
 
 /// This function returns the default attributes to set on each user event: OCKAM_HOME, `ockam` version and git hash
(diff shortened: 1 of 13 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/journeys/attributes.rs
```diff
@@ -25,35 +27,118 @@ pub fn default_attributes<'a>() -> HashMap<&'a Key, String> {
...
+            let random_id_generator = RandomIdGenerator::default();
+            random_id_generator.new_span_id()
+        }
+    }
+}
+
+/// Return a string containing enough data to uniquely identify a host
+/// That string is hexadecimal string which can be part of a trace id
+pub(crate) fn make_host() -> String {
+    let host = match (get_mac_address(), get_ip_address()) {
```
davide-baldo, line 91: for privacy, can't we hash it out? (hostname may contain user name)

davide-baldo: That would also make the length consistent (we can truncate the hash however we want)

implementations/rust/ockam/ockam_api/src/cli_state/journeys/journeys.rs
```diff
@@ -195,92 +200,181 @@ impl CliState {
     }
 
     /// Return a list of journeys for which we want to add spans
```
davide-baldo, line 202: Why do we write to host and project journey at the same time?

etorreborre: The two journeys end-up being different:

 - The project journey groups events possibly coming from different machines from different users.
 - The host journey is machine centric and can contain events that are emitted before a project even exists (when toying around with local nodes for example).

implementations/rust/ockam/ockam_api/src/cli_state/journeys/attributes.rs
etorreborre: Good idea.

davide-baldo (approved):

PR #7649, opened 2024-02-25
refactor(rust): remove `--resource` and `--resource-type` args from `policy show|list|delete`

adrianbenavides:
In the `policy show|list|delete` commands, additionally to the resource name, we were also forcing the user to specify what kind of resource that name is associated to. E.g. `ockam policy show --resource-type tcp-outlet`.

This PR simplifies those commands so the user only has to pass the resource's name. Whether it's a `resource-type` or a `resource` is done automatically by checking if the passed name can be first converted to a valid `ResourceType`, otherwise it's assumed to be a `ResourceName`. This is also the behavior of the `tuified` behavior, so now the UX is the same in both tty and non-tty cases. 

It also improves the json format to display the policies details properly:

```json
# Before
{
  "resource": {
    "Type": "echoer"
  },
  "action": "handle_message",
  "expression": {
    "Ident": "subject.has_credential"
  }
}

# Now
{
  "resource": "myoutlet",
  "action": "handle_message",
  "expression": "(= subject.c 123)"
}
```


implementations/rust/ockam/ockam_abac/Cargo.toml
```diff
@@ -30,15 +30,18 @@ std = [
...
 
 [dependencies]
+cfg-if = "1.0.0"
 either = { version = "1.10.0", default-features = false }
 minicbor = { version = "0.20.0", features = ["derive", "alloc"] }
 ockam_core = { version = "0.102.0", path = "../ockam_core", default-features = false }
 ockam_identity = { version = "0.104.0", path = "../ockam_identity", default-features = false }
 ockam_node = { version = "0.109.0", path = "../ockam_node", default-features = false }
 once_cell = { version = "1.19.0", default-features = false, features = ["alloc"] }
+serde = { version = "1.0.197", default-features = false, features = ["derive"] }
```
davide-baldo: specifying the major only avoids dependency conflicts.

implementations/rust/ockam/ockam_command/src/policy/create.rs
```diff
@@ -51,25 +52,15 @@ impl CreateCommand {
...
-                        color_primary("--resource tcp-outlet"),
-                        color_primary("--resource-type tcp-outlet")
-                    ))?;
-                    self.resource_type = Some(ResourceType::TcpOutlet);
-                }
-                _ => self.resource = Some(resource),
+        if let Some(resource) = self.resource.as_ref() {
+            if let Ok(resource_type) = ResourceType::from_str(resource.as_str()) {
+                let resource_type_str = resource_type.to_string();
+                opts.terminal.write_line(fmt_warn!(
```
davide-baldo, line 58: super smooth deprecation, love it! :rocket:

implementations/rust/ockam/ockam_command/tests/bats/policies_orchestrator.bats
```diff
@@ -0,0 +1,86 @@
...
+  # This will fail because the resource type policy is not satisfied
+  run_failure curl --head --retry-connrefused --max-time 3 "127.0.0.1:$inlet_port"
+
+  # Update resource type policy and try again. Now the policy is satisfied
+  export OCKAM_HOME=$DB_OCKAM_HOME
+  run_success $OCKAM policy create --resource-type tcp-outlet --expression '(= subject.component "web")'
+  run_success curl --head --retry-connrefused --retry 2 --max-time 5 "127.0.0.1:$inlet_port"
+
+  # Update the policy for the outlet and try again. It will fail because the local policy is not satisfied
+  run_success $OCKAM policy create --resource outlet --expression '(= subject.component "NOT_web")'
```
davide-baldo, line 84: shouldn't the explicit outlet expression win over the policy? 
My expectation would be:
1. outlet expression
2. policy
3. type

implementations/rust/ockam/ockam_abac/Cargo.toml
adrianbenavides: I didn't know that, thanks for the heads up!

davide-baldo (approved):

PR #7667, opened 2024-02-27
feat(rust): rework migrations

SanjoDeundiak:


implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
```diff
@@ -0,0 +1,284 @@
...
+        for migration in all_migrations.iter().map(|(_version, m)| m) {
+            match migration {
+                NextMigration::Sql(sql_migration) => {
+                    if sql_migration.migration_type.is_down_migration() {
+                        return Ok(());
+                    }
+
+                    match applied_migrations.get(&sql_migration.version) {
+                        Some(applied_migration) => {
+                            if sql_migration.checksum != applied_migration.checksum {
```
davide-baldo, line 176: Do we really need the checksum check?

implementations/rust/ockam/ockam_macros/src/lib.rs
```diff
@@ -168,3 +168,18 @@ fn to_compile_errors(errors: Vec<syn::Error>) -> proc_macro2::TokenStream {
 fn to_compile_error(error: syn::Error) -> TokenStream {
     error.to_compile_error().into()
 }
+
+#[proc_macro]
+pub fn migrate(input: TokenStream) -> TokenStream {
```
davide-baldo, line 173: I don't get why this needs to be a macro

SanjoDeundiak: Because it should be run during the compilation. Sql scripts don't exist during the runtime

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
SanjoDeundiak: Sqlx has that, so I decided to preserve it

implementations/rust/ockam/ockam_macros/src/lib.rs
davide-baldo: Could we use a standard [include_str](https://doc.rust-lang.org/std/macro.include_str.html) instead?

SanjoDeundiak: How would you iterate over the files in the directory with include_str?

SanjoDeundiak: Also, we actually use sqlx macro under the hood, which you also can't do without a macro

davide-baldo: got it

adrianbenavides (approved):

PR #7673, opened 2024-02-27
Added manual tests to measure latency

davide-baldo:


adrianbenavides:
Have you considered using a benchmarking tool like [criterion](https://docs.rs/criterion/latest/criterion/) to run these tests? Maybe we could keep a history of the benchmarks so we can check how new versions compare to older releases.

davide-baldo:
@adrianbenavides 
I did try to use criterion, but the async support it's pretty minimal, there is the possibility to set up the benchmark but nothing to bring it down. There are also limitations on what you can move inside async closures.

Other frameworks didn't even bother to implement the async support, and the rust standard is still unstable.
And I couldn't find a framework that supports the possibility to manually provide the data.

adrianbenavides (approved):

PR #7679, opened 2024-02-27
handle project admin credentials on authority

polvorin:
* Add a new, optional  `--account-authority` argument to `authority create`.  This is a full change history, in hex format.
* If the above argument is given,  the authority will accept credentials signed by that identity and,  if it has a `project` attribute equals to this authority' project id,   it will consider the subject as a valid administrator (as long as the credential doesn't expire, as always)
* Having such credential means:
    * It's allowed to create/remove enrollers
    * It's itself a member,  so can retrieve his own credential (allowing it to use any relay name, `relay=*`).   Implication of this is that after creating a project/enrolling with orchestrator,   that identity can start to use the project without any further action.   
    * No one can remove it

* Enrollers are not allowed to create/remove other enrollers,  altough as a stopgap we still allow it. It will be removed on a future client release.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator.rs
+++ b/implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator.rs
@@ -1,4 +1,5 @@
 use either::Either;
+use ockam::identity::IdentitiesAttributes;
 use std::collections::{BTreeMap, HashMap};
 
 use ockam::identity::utils::now;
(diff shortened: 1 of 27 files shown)


implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator.rs
```diff
@@ -23,11 +24,48 @@ pub type DirectAuthenticatorResult<T> = Either<T, DirectAuthenticatorError>;
 
 pub struct DirectAuthenticator {
     members: Arc<dyn AuthorityMembersRepository>,
+    account_authority: Option<AccountAuthorityInfo>,
+}
+#[derive(Clone)]
+pub struct AccountAuthorityInfo {
+    identities_attributes: Arc<IdentitiesAttributes>,
```
SanjoDeundiak, line 31: I think that shouldn't be a dependency here. I' would like to refactor it. WDYT?

implementations/rust/ockam/ockam_api/src/authority_node/authority.rs
```diff
@@ -39,6 +39,7 @@ pub struct Authority {
     secure_channels: Arc<SecureChannels>,
     members: Arc<dyn AuthorityMembersRepository>,
     tokens: Arc<dyn AuthorityEnrollmentTokenRepository>,
+    account_authority: Option<AccountAuthorityInfo>,
```
SanjoDeundiak, line 42: I wonder if we should reuse `load_controller_identifier` somehow in the future

implementations/rust/ockam/ockam_api/tests/token_enrollment.rs
```diff
@@ -50,7 +50,7 @@ async fn admin_can_issue_token(ctx: &mut Context) -> Result<()> {
 }
 
 #[ockam_macros::test]
-async fn admin_cant_accept_token(ctx: &mut Context) -> Result<()> {
+async fn admin_can_accept_token(ctx: &mut Context) -> Result<()> {
```
SanjoDeundiak, line 53: Should it? It'll still get credential as if it's an admin, so it doesn't effect anything?

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_api/src/authenticator/direct/direct_authenticator.rs
polvorin: sure.   I needed some way to access to the identity attributes (attested by the account credential).   But agree that it's subtle.. if one isn't careful can end up being confused between this and the member table

implementations/rust/ockam/ockam_api/src/authority_node/authority.rs
polvorin: I wasn't aware of that function 🤔 ,  it could be used indeed,  but not sure if it makes the UX simpler/easier to understand,   as this is basically a third-party authority that _this_ authority trust to emit project membership' credentials.   It's true than on orchestrator that is the controller identity,  but could be something else (so the name could be confusing if we use controller_identifier).   Honestly I think the only places where you use something different there are either on tests (to not depend on controller),  as on custom deployments.. problably won't use admin credentials at all.

implementations/rust/ockam/ockam_api/tests/token_enrollment.rs
polvorin: Good point!, right 🤔 ,   this probably was before adding the change so admins get issued credentials too, and didn't revisit.
It's a bit tricky however, we could make that before adding a member or accepting an enrollment token,  we verify the identify trying to do that isn't a valid admin.   But admin' credentials are pretty short-lived so from the user point of view, will be that sometimes it's allowed, and sometimes it isn't.

It's a good think to think of,  how we want to handle this.

SanjoDeundiak (approved):

PR #7680, opened 2024-02-28
refactor(rust): minor structural changes to the custom migrator

adrianbenavides:
- move the logic of how to sort and apply migrations into the `NextMigration` enum
- add struct to abstract the behavior of version intervals (from, to, incude_to)

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
+++ b/implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
@@ -1,7 +1,7 @@
-use ockam_core::compat::collections::{HashMap, HashSet};
+use ockam_core::compat::collections::HashSet;
 use ockam_core::compat::time::now;
 use ockam_core::errcode::{Kind, Origin};
-use sqlx::migrate::{Migrate, Migration};
+use sqlx::migrate::{AppliedMigration, Migrate, Migration as SqlxMigration};
 use sqlx::sqlite::SqliteRow;
 use sqlx::{query, Row, SqliteConnection, SqlitePool};
 use std::cmp::Ordering;
@@ -67,29 +67,10 @@ impl Migrator {
 }
 
 impl Migrator {
-    fn is_inside_version_range(
-        version: i64,
-        from_version: i64, // including
-        to_version: i64,
-        including_to: bool,
-    ) -> bool {
-        if from_version <= version && version < to_version {
-            return true;
-        }
-
-        if version == to_version && including_to {
-            return true;
-        }
-
-        false
-    }
-
     async fn run_migrations(
         &self,
         connection: &mut SqliteConnection,
-        from_version: i64,  // including
-        to_version: i64,    // not including
-        run_last_sql: bool, // will run sql migration with verion == to_version
+        up_to: Version,
     ) -> Result<()> {
         connection.ensure_migrations_table().await.into_core()?;
 
@@ -102,6 +83,26 @@ impl Migrator {
             ));
         }
 
+        let migrations = {
+            let sql_iterator = self.sql_migrator.migrations.iter().filter_map(|m| {
+                if m.version <= up_to {
+                    Some(NextMigration::Sql(m))
+                } else {
+                    None
+                }
+            });
+            let rust_iterator = self.rust_migrations.iter().filter_map(|m| {
+                if m.version() <= up_to {
+                    Some(NextMigration::Rust(m.as_ref()))
+                } else {
+                    None
+                }
+            });
+            let mut migrations: Vec<NextMigration> = sql_iterator.chain(rust_iterator).collect();
+            migrations.sort();
+            migrations
+        };
+
         // sqlx Migrator also optionally checks for missing migrations (ones that had been run and
         // marked as migrated in the db but now don't exist). Skipping that check for now.
         // WARNING: the check if re-enabled can potentially fail because of renaming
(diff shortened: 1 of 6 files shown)


implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
```diff
@@ -67,29 +67,11 @@ impl Migrator {
...
-    }
-
     async fn run_migrations(
         &self,
         connection: &mut SqliteConnection,
-        from_version: i64,  // including
-        to_version: i64,    // not including
-        run_last_sql: bool, // will run sql migration with verion == to_version
+        version_interval: VersionInterval,
+        run_last_sql_migration: bool,
```
etorreborre: Why not extend the `VersionInterval` data type so that either:
 
 - It has a function `set_last_version_inclusive(self) -> Self`
 - It has  2 constructors `from_to` and `from_to_inclusive`

This way we directly decide if the last version should be in when building the range.

adrianbenavides: I didn't do that because of how it's used later: see lines 87-105. I'm not entirely sure why we treat rust/sql differently in that block.

adrianbenavides: I think it's because how it's used in tests. I'm going to try to change a little how we use it so we can define the interval properly.

adrianbenavides: done!

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
```diff
@@ -266,19 +187,262 @@ impl Migrator {
...
+        let mut migrations = vec![
+            NextMigration::Sql(&sql_2),
+            NextMigration::Sql(&sql_1),
+            NextMigration::Rust(rust_1.as_ref()),
+            NextMigration::Rust(rust_3.as_ref()),
+            NextMigration::Rust(rust_2.as_ref()),
+        ];
+        migrations.sort();
+
+        for m in migrations.iter() {
```
SanjoDeundiak: If you like, you can do
```
for m in &migrations {
```

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
```diff
@@ -266,19 +187,262 @@ impl Migrator {
...
+                }
+
+                async fn migrate(&self, _connection: &mut SqliteConnection) -> Result<bool> {
+                    Ok(true)
+                }
+            }
+        };
+    }
+
+    impl_rust_migration!(DummyRustMigration1, 1);
```
SanjoDeundiak: Wouldn't it be easier to define `DummyRustMigration` that takes one number in the constructor? But up to you

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
```diff
@@ -266,19 +187,262 @@ impl Migrator {
...
+    }
+}
+
+trait Migration {
+    fn version(&self) -> Version;
+}
+
+type Version = i64;
+
+pub(crate) struct VersionInterval {
```
SanjoDeundiak: That's a nice improvement, but at this point I wonder if we could simply use `core::ops::Range`?

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
```diff
@@ -266,19 +187,262 @@ impl Migrator {
...
         pool: &SqlitePool,
-        version: i64, // not including
-        run_last_sql: bool,
+        version_interval: VersionInterval,
+    ) -> Result<()> {
+        self.migrate_partial(pool, version_interval).await
+    }
+}
+
+trait Migration {
```
SanjoDeundiak: It's weird that clippy didn't notice that, but it's unused

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
```diff
@@ -102,6 +83,36 @@ impl Migrator {
...
+                    None
+                }
+            });
+
+            // When running a test for a rust migration, we start the test running all the
+            // migrations right before the rust migration we want to test. This includes the sql
+            // migration that exists for the same version. We modify `version_interval` to ensure
+            // that only the sql migration is run, so that we can then populate the database with
+            // test data, and finally run the rust migration that is being tested.
+            #[cfg(test)]
```
SanjoDeundiak: The code that behaves differently depending if `test` is enabled is misleading. Twice misleading when you try to test it from another crate and notice that `test` isn't propagated

SanjoDeundiak: A much better approach would be to either have more clear public API to invoke desired behaviour, or to create a dedicated function (with clear naming) that is only available with `test` flag

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
```diff
@@ -110,94 +121,8 @@ impl Migrator {
...
-                    if Self::has_migrated(connection, rust_migration.name()).await? {
-                        continue;
-                    }
-                    if rust_migration.migrate(connection).await? {
-                        Self::mark_as_migrated(connection, rust_migration.name()).await?;
-                    }
-                }
-            }
+        for migration in migrations.into_iter() {
+            migration.apply(connection, &applied_migrations).await?;
```
SanjoDeundiak: I would inline the implementation here. Otherwise, it's unclear that `applied_migrations` only include sql migrations, while rust migrations also have their own mechanism of checking if they were applied.

SanjoDeundiak (changes requested):
Thank you for these improvements! Left a few comments

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
adrianbenavides: Totally, I did this thinking that I might need a unique name as well and didn't review it after writing the tests. Thanks!

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
adrianbenavides: Ok, I'll try that!

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
adrianbenavides: Oh, I didn't have this in mind, I think using that would simplify it, indeed.

adrianbenavides: I've removed this completely. Now that we store what migrations have been migrated and skip them, we were always using `0` as the lower bound of the interval, which makes using a range unnecessary.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
adrianbenavides: I've moved the logic of skipping the last rust migration up to the higher level, at the `Migrator`'s test function [here](https://github.com/build-trust/ockam/blob/01f455a066689606d8039c4394cdb2e4a5803a16/implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs#L195).

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
```diff
@@ -266,19 +185,192 @@ impl Migrator {
...
-        &self,
+    /// Run migrations up to the specified version (inclusive) but skip the last rust migration
+    pub(crate) async fn migrate_up_to_skip_last_rust_migration(
+        mut self,
         pool: &SqlitePool,
-        version: i64, // not including
-        run_last_sql: bool,
+        up_to: Version,
+    ) -> Result<()> {
+        self.rust_migrations.retain(|m| m.version() < up_to);
```
SanjoDeundiak, line 200: Would be enough to just remove one rust migration where `m.version() == up_to` so that you don't repeat the filtering logic twice

SanjoDeundiak (approved):

PR #7694, opened 2024-02-29
fix(rust): make sure all the handshake spans end up in the same trace with the secure channel use

etorreborre:
This PR makes sure that:

 - The `trace_code` function sets-up the initial `Context` with the root tracing context.
 - When a `Context` is duplicated, its tracing context is updated with the most recent span.

This way, the spans corresponding to starting the secure channel end up under the same root span as the code using the secure channel
```
root
└── send_echo_message_over_secure_channel
    ├── create tcp transport
    ├── create tcp transport
    ├── TcpSendWorker::handle_message
    |   └── TcpRecvProcessor::process
    |       └── HandshakeWorker::handle_message
    |           └── TcpSendWorker::handle_message
    |               └── TcpRecvProcessor::process
    |                   └── HandshakeWorker::handle_message
    |                       └── TcpSendWorker::handle_message
    |                           └── TcpRecvProcessor::process
    |                               └── HandshakeWorker::handle_message
    └── MessageSender::handle_message
        └── EncryptorWorker::handle_message
            └── handle_encrypt
                └── TcpSendWorker::handle_message
                    └── TcpRecvProcessor::process
                        └── DecryptorWorker::handle_message
                            └── handle_decrypt
                                └── Echoer::handle_message
                                    └── EncryptorWorker::handle_message
                                        └── handle_encrypt
                                            └── TcpSendWorker::handle_message
                                                └── TcpRecvProcessor::process
                                                    └── DecryptorWorker::handle_message
                                                        └── handle_decrypt
```

**UPDATE**: I've added a fix for the creation of a user journey. I don't know but there was a regression there. 
**UPDATE**: I've added another fix the selection of crates when logging is off.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
+++ b/implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
@@ -7,7 +7,7 @@ use std::path::PathBuf;
 use tracing_core::Level;
 use tracing_subscriber::EnvFilter;
 
-use super::{is_tracing_set, Colored, GlobalErrorHandler, LoggingEnabled};
+use super::{Colored, GlobalErrorHandler, LoggingEnabled};
 use crate::logs::LogFormat;
 
 /// List of all the configuration parameters relevant for configuring the logs
@@ -128,6 +128,22 @@ impl LoggingConfiguration {
         }
     }
 
+    /// Set the default crates
+    pub fn set_default_crates(self) -> LoggingConfiguration {
+        LoggingConfiguration {
+            crates: Some(LoggingConfiguration::default_crates()),
+            ..self
+        }
+    }
+
+    /// Set the all crates
+    pub fn set_all_crates(self) -> LoggingConfiguration {
+        LoggingConfiguration {
+            crates: None,
+            ..self
+        }
+    }
+
     /// Create an EnvFilter which keeps only the log messages
     ///
     ///  - for the configured level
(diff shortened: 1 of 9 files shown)


implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
```diff
@@ -151,18 +167,19 @@ impl LoggingConfiguration {
     }
 
     /// Return a LoggingConfiguration which doesn't log anything
-    pub fn off() -> LoggingConfiguration {
-        LoggingConfiguration::new(
+    /// We still retrieve the
```
adrianbenavides: Some text is missing here?

implementations/rust/ockam/ockam_node/src/context/context_lifecycle.rs
```diff
@@ -119,7 +119,7 @@ impl Context {
             self.transports.clone(),
             &self.flow_controls,
             #[cfg(feature = "std")]
-            self.tracing_context(),
+            OpenTelemetryContext::current(),
```
adrianbenavides, line 122: What's the rationale behind this change? I'm trying to understand when we need to get the "global" context or just copy the one available at `self`.

implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
etorreborre: Oops, I must have be interrupted. I forced-push the rest of the comment.
But I am thinking that I will probably refactor that part to extract the configuration that's used by both logging and tracing to make things clearer.

implementations/rust/ockam/ockam_node/src/context/context_lifecycle.rs
etorreborre: That's all due to the complexity of:

 - Having to pass the current tracing context between async processes when working with our workers. We work around this by setting a tracing context on the worker Context
 
 - Having a different notion of "tracing context" between the tracing crate, which creates spans using the `#[instrument]` macro, and the opentelemetry crate which creates the span data sent to the collector at a different point in time (when exiting the instrumented function). It's also the `opentelemetry` crate which generates span ids. So in order to get the latest span id, the one corresponding to the span created by the latest `#[instrument]`ed method, we need to reactualize the OpenTelemetry context using all the code in `OpenTelemetryContext::current()`.

SanjoDeundiak (approved):

PR #7702, opened 2024-03-01
chore(elixir): do not include secure channel return address when not needed

polvorin:
the peer address is discovered at secure channel initialization, after that it's known and fixed.
https://github.com/build-trust/ockam/pull/7701

### DIFF
--- a/implementations/elixir/ockam/ockam/test/ockam/secure_channel_test.exs
+++ b/implementations/elixir/ockam/ockam/test/ockam/secure_channel_test.exs
@@ -33,9 +33,16 @@ defmodule Ockam.SecureChannel.Tests do
 
   defp man_in_the_middle(callback, initiator, n) do
     receive do
-      %Message{return_route: ^initiator} = message when n > 2 ->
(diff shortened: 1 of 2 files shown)


implementations/elixir/ockam/ockam/test/ockam/secure_channel_test.exs
```diff
@@ -76,7 +83,7 @@ defmodule Ockam.SecureChannel.Tests do
 
       Router.route(message)
 
-      if rem(i, 2) == 1 do
+      if rem(i, 2) == 0 do
```
etorreborre, line 86: I mildly shocked to see that there's no `is_even` function in Elixir? Would it make sense to make it a utility function somewhere?

etorreborre (approved):

implementations/elixir/ockam/ockam/test/ockam/secure_channel_test.exs
polvorin: there is one macro on Integer,  didn't check before writting the test

PR #7705, opened 2024-03-03
feat(rust): `node create` can be run given a configuration file

adrianbenavides:
## Main changes

- I've reorganized the files under `./parser`, and most of the changes are due to this.
- I've added a [trait for the configuration building blocks](https://github.com/build-trust/ockam/pull/7705/files#diff-e36ffaaf7fea8785172ea93999149bc7f02b7d42fd3e8b5e3703cae48b4dd0abR8) to hide some common details of the implementation. I've also added a [trait for commands](https://github.com/build-trust/ockam/pull/7705/files#diff-c1790a45fa6e1ede4e83172fe334835604a8d860075229ad51f563af3b8f1573R37), that is used by the previous trait to auto-implement some of its internal logic.
- The `Config::run` method now builds the commands first to bubble up any validation error that may come up when parsing the arguments into commands, and then it runs all the commands. This will avoid leaving an invalid state due to validation errors.
- Finally, the [`NodeConfig`](https://github.com/build-trust/ockam/pull/7705/files#diff-9e0e3be7253b2e190afe7acf73d7082c7ccd72176c297102f9d61c3b7b2bb161R23) is similar to the regular `Config` struct but with some differences to merge into it the arguments of the `node create` command. This is run in the main thread (depending on the `child_process` arg) to print commands' output to the user.

## Other changes

- There are lots of files with a single change because I've modified the `name()` function of all the commands so that they match the exact command name. This is done so we can use the returned string to parse it back into the correct command.
- I've removed the `configuration` commands, since they are not relevant anymore.

## How to test

```
# Create a default node
ockam node create

# From the project root, create the node with the config file
ockam node create ./implementations/rust/ockam/ockam_command/src/node/create/demo_config_files/1.portal.single-machine.yaml

# Check that the config resources have been attached to the second node, and not to the default one
```


implementations/rust/ockam/ockam_command/src/project/addon/configure_okta.rs
```diff
@@ -87,7 +87,7 @@ impl AddonConfigureOktaSubcommand {
     }
 
     pub fn name(&self) -> String {
-        "configure okta addon".into()
+        "project addon configure influxdb".into()
```
etorreborre: This should be `project addon configure okta`

implementations/rust/ockam/ockam_command/src/project/enroll.rs
```diff
@@ -31,8 +31,8 @@ long_about = docs::about(LONG_ABOUT),
 after_long_help = docs::after_help(AFTER_LONG_HELP)
 )]
 pub struct EnrollCommand {
-    /// Path to an enrollment ticket file, or hex encoded enrollment ticket
-    #[arg(display_order = 800, group = "authentication_method", value_name = "ENROLLMENT TICKET PATH | ENROLLMENT TICKET", value_parser = parse_enroll_ticket)]
+    /// Path, URL or inlined hex-encoded enrollment ticket
+    #[arg(display_order = 800, group = "authentication_method", value_name = "ENROLLMENT TICKET", value_parser = parse_enrollment_ticket)]
     pub enroll_ticket: Option<EnrollmentTicket>,
```
etorreborre: Shouldn't this argument be called `enrollment_ticket` instead for consistency?

implementations/rust/ockam/ockam_command/src/run/parser/building_blocks.rs
```diff
@@ -1,59 +1,91 @@
...
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
 use std::fmt::Display;
 
-static BINARY_PATH: Lazy<String> = Lazy::new(|| {
-    std::env::args()
-        .next()
-        .expect("Failed to get the binary path")
-});
+pub trait ArgsToCommands: Sized {
```
etorreborre, line 22: It would be nice to have:

- A comment describing the purpose of that trait.
- A comment for the functions and their arguments.
- And maybe a short concrete example. 
 
Otherwise it is quite abstract to read as such.

implementations/rust/ockam/ockam_command/src/run/parser/resource/node.rs
```diff
@@ -0,0 +1,135 @@
...
+    pub project: Option<ArgValue>,
+}
+
+impl Node {
+    pub const NAME_ARG: &'static str = "name";
+}
+
+#[async_trait]
+impl ConfigRunner<CreateCommand> for Node {
+    fn len(&self) -> usize {
```
etorreborre: This function does not seem to be used anywhere.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -96,58 +109,82 @@ impl Default for CreateCommand {
...
         }
     }
+}
 
-    pub async fn async_run(self, ctx: &Context, opts: CommandGlobalOpts) -> miette::Result<()> {
-        let ctx = ctx.async_try_clone().await.into_diagnostic()?;
-        if self.foreground {
-            self.foreground_mode(&ctx, opts).await
+impl CreateCommand {
+    pub fn tracing_name(&self) -> String {
```
etorreborre: I don't think that this function is necessary anymore. That name is only used now in a debug message. I think that the distinction was previously used to create a span name but this is not the case anymore, so we can use the regular "name()" function.

etorreborre (commented):
I left some initial comments then I realized that I would structure the traits differently. Here is a commit with my proposal: `adrian/node-create-config-refactor`

implementations/rust/ockam/ockam_command/src/run/parser/building_blocks.rs
adrianbenavides: Done!

implementations/rust/ockam/ockam_command/src/run/config.rs
```diff
@@ -0,0 +1,462 @@
...
+            self.tcp_inlets.parse_commands(overrides)?.into(),
+        ];
+
+        // Run commands
+        for cmd in commands {
+            cmd.run(ctx, opts).await?
+        }
+        Ok(())
+    }
+    // Build commands and return validation errors before running any command.
```
etorreborre: This comment is out of place.

etorreborre (dismissed):

implementations/rust/ockam/ockam_command/src/run/config.rs
adrianbenavides: Oops, right! Solved 🙏

etorreborre (approved):

PR #7710, opened 2024-03-05
feat: tracing improvements

etorreborre:
This PR improves a few things related to tracing:

 - No logging layer is set up if logging is disabled. This improves the runtime performance when logging is off but tracing is on.
 - The blocking behavior observed during the export of spans and log records was due to a missing crate feature (This is described in one of the comments of the opentelemetry crate).
 - Some recent bats tests generate test files in the current path. I removed them.
 - There was a bug on the creation of a trace id based on the project name (that would only show up in the bats tests).
 - I renamed some methods and env variables to use the word "EXPORT" rather than "TRACING" since the main difference with using OpenTelemetry and not using it is the export of spans and log records.
 - I tweaked the timeouts for checking if a node is up so that we don't retry too often when checking if a node is up. When opentelemetry exporting is on, this helps in reducing the overhead.

I ran the full bats tests suite (with `--jobs 4`), with and without opentelemetry exporting:

 - Without: 5 min 04s
 - Without: 6 min 34s 

Given the amount of commands executed, this doesn't look so bad.


implementations/rust/ockam/ockam_api/src/cli_state/journeys/attributes.rs
```diff
@@ -99,6 +100,17 @@ pub(crate) fn make_host() -> String {
...
+/// Fill or trim a string so that it makes exactly the desired size
+fn adjust(s: String, desired_size: usize, filler: char) -> String {
+    let mut result = s;
+    let current_size = result.len();
+    // make sure that there exactly the desired number of characters
+    if current_size < desired_size {
+        result.extend(std::iter::repeat(filler).take(desired_size - current_size));
+    };
+    result[0..25].to_string()
+}
```
adrianbenavides, line 106: Have you checked [this](https://doc.rust-lang.org/std/fmt/#fillalignment)? I think you could replace this function with a call to `format!()`

implementations/rust/ockam/ockam_api/src/cli_state/journeys/attributes.rs
```diff
@@ -35,7 +35,7 @@ pub fn default_attributes<'a>() -> HashMap<&'a Key, String> {
 ///  - The last 6 characters are the 'now' date as YYMMDD
 ///
 pub(crate) fn make_host_trace_id(now: DateTime<Utc>) -> TraceId {
-    let mut machine = make_host();
+    let mut machine = adjust(make_host(), 25, '1');
```
adrianbenavides: The code after this (lines 39-43) is duplicated right?

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
```diff
@@ -45,7 +45,7 @@ pub struct CliState {
     dir: PathBuf,
     database: SqlxDatabase,
     application_database: SqlxDatabase,
-    tracing_enabled: TracingEnabled,
+    tracing_enabled: ExportingEnabled,
```
adrianbenavides: Should this be renamed to `exporting_enabled`? Or is it equivalent?

implementations/rust/ockam/ockam_command/src/node/show.rs
```diff
@@ -218,44 +223,72 @@ pub async fn is_node_up(
...
-            tokio::time::sleep(timeout_duration).await;
-            continue;
+        if total_time >= IS_NODE_ACCESSIBLE_TIMEOUT || !wait_until_ready && !total_time.is_zero() {
+            return Ok(false);
+        };
+        if node_client.is_accessible(ctx).await.is_ok() {
+            trace!(%node_name, "node is accessible");
+            return Ok(true);
         }
+        trace!(%node_name, "node is not accessible");
```
adrianbenavides, line 257: Instead of doing the `!wait_until_ready && !total_time.is_zero()` above, wouldn't it be the same to check `if !wait_until_ready { return Ok(false) }` right after this line? And the same for the `is_node_ready` function.

implementations/rust/ockam/ockam_api/src/cli_state/journeys/attributes.rs
etorreborre: Thanks @adrianbenavides, that's so much better than me reinventing the wheel!

implementations/rust/ockam/ockam_api/src/cli_state/journeys/attributes.rs
etorreborre: 🤦 yes.

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
etorreborre: Yes I missing that one

implementations/rust/ockam/ockam_command/src/node/show.rs
etorreborre: We still need to make sure that we run the loop at least once, which is what `!total_time.is_zero()`  is testing.

implementations/rust/ockam/ockam_api/src/cli_state/journeys/attributes.rs
etorreborre: @adrianbenavides I eventually reverted to the original code because the string can be filled in some cases but trimmed in others and `format!` does not seem to support this use case.

adrianbenavides (approved):

PR #7711, opened 2024-03-05
feat(rust): update display, log output in frequently used commands

nazmulidris:
- Make the changes described in the doc below to `ockam enroll`. 
- Fix error reporting (don't display duplicate error output).
- Fix --version / -V to display indented output.

Issue / Epic: https://github.com/build-trust/codebase/issues/726
Docs: https://docs.google.com/document/d/16Cq2RQU95hcQ1ddNu2ETtd9qkpCEToQW7ikv12LxEVo/edit
Video (WIP): https://drive.google.com/file/d/19mGya91w0qNS8PBveiharMgqWdP_8zQw/view?usp=drive_link

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/entry_point.rs
+++ b/implementations/rust/ockam/ockam_command/src/entry_point.rs
@@ -1,10 +1,13 @@
+use std::process::exit;
+
 use clap::Parser;
+
 use miette::IntoDiagnostic;
 use tracing_core::Level;
 
 use crate::{
-    add_command_error_event, has_help_flag, pager, replace_hyphen_with_stdin, ErrorReportHandler,
-    OckamCommand,
+    add_command_error_event, fmt_log, has_help_flag, has_version_flag, pager,
+    replace_hyphen_with_stdin, util::exitcode, version::Version, ErrorReportHandler, OckamCommand,
 };
 use ockam_api::cli_state::CliState;
 use ockam_api::logs::{
@@ -22,6 +25,10 @@ pub fn run() -> miette::Result<()> {
         .collect::<Vec<_>>();
 
     let _ = miette::set_hook(Box::new(|_e| Box::new(ErrorReportHandler::new())));
+
+    if has_version_flag(&input) {
+        print_version_and_exit();
+    }
 
     match OckamCommand::try_parse_from(input.clone()) {
         Err(help) => {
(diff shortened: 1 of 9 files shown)


implementations/rust/ockam/ockam_command/src/entry_point.rs
```diff
@@ -23,37 +24,46 @@ pub fn run() -> miette::Result<()> {
...
-                    .iter()
-                    .map(|s| s.to_string())
-                    .collect::<Vec<String>>()
-                    .join(" ");
+    if has_version_flag(&input) {
+        let version_msg = Version::long();
+        let version_msg_vec = version_msg.split('\n').collect::<Vec<_>>();
+        eprintln!("{}", fmt_log!("ockam {}", version_msg_vec[0]));
+        for line_num in 1..version_msg_vec.len() {
+            eprintln!("{}", fmt_log!("{}", version_msg_vec[line_num]));
```
adrianbenavides: Why do we need to handle the version flags ourselves (outside of clap)?

If have to do it, I'd avoid the if/else to improve the readability:

```rust
if has_version_flag(&input) {
    print_version_and_exit();
}

match OckamCommand::try_parse_from ...
```

nazmulidris: excellent suggestion. 
this [slack thread](https://ockamteam.slack.com/archives/CR11KJ4CW/p1709597847844499) has context for this change.

adrianbenavides: I still don't get it. Are we replacing [clap's version](https://github.com/build-trust/ockam/blob/aad7f0001a9c3fc7c0ef9cfd3d2ca77c7a028f3e/implementations/rust/ockam/ockam_command/src/command.rs#L38-L39) handling? Couldn't we use those attributes to define the function we want to call to display the version instead of manually handling the version arguments?

nazmulidris: if we use clap, we can't inject spaces before "ockam". we only control the string that comes after it. 

so in the display output, if we wanted the following, then we can use clap:
`"ockam ░░░░version: XYZ"`

however, since we want the spaces before the command name `ockam`, we have to bypass clap to get:
`"░░░░ockam version: XYZ"`

adrianbenavides: ooh! I get it now, thanks for showing an example, I didn't see that we wanted to add the margin before the "ockam" string. All clear now!

nazmulidris: @adrianbenavides Awesome! It is a tricky one that made no sense (to me either) w/out using that example :)

adrianbenavides (approved):

davide-baldo (approved):

PR #7714, opened 2024-03-06
feat(rust): change behavior of how the target route is handled in "tcp-inlet create"

adrianbenavides:
- The `to` argument accepts either a MultiAddr route or a short string to define the service name. Before this PR, the short string was used to define the relay name.
- Add a new optional argument `via` to define the relay name.

When `to` is passed as a service name, and `via` is also passed, the route is processed as follows:

`/project/$DEFAULT_PROJECT_NAME/service/forward_to_$RELAY_NAME/secure/api/service/$SERVICE_NAME`

which simplifies its definition to just two "words":

```
# before
ockam tcp-inlet create --to /project/myproject/service/forward_to_myrelay/secure/api/service/myservice

# now
ockam tcp-inlet create --via myrelay --to myservice
```

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
@@ -24,7 +24,7 @@ use ockam_api::nodes::service::portals::Inlets;
 use ockam_api::nodes::BackgroundNodeClient;
 use ockam_api::{random_name, ConnectionStatus};
 use ockam_core::api::{Reply, Status};
-use ockam_multiaddr::proto::Project;
+use ockam_multiaddr::proto;
 use ockam_multiaddr::{MultiAddr, Protocol as _};
 
 use crate::node::util::initialize_default_node;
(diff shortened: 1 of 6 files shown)


implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
```diff
@@ -273,95 +277,160 @@ impl CreateCommand {
...
+            ),
+        ];
+        for (to, via, expected) in cases {
+            let res = CreateCommand::parse_arg_to(&state, &to, Some(&via.to_string()))
+                .await
+                .unwrap();
+            assert_eq!(res, expected.to_string());
+        }
+
+        // "via" is ignored if "to" is passed as a full route
```
etorreborre: Should it be ignored or raise an error? Because it's likely to be a configuration error in that case.

etorreborre (dismissed):
That's a very nice UX improvement. I left a question and it would be nice to enrich the long help text as well.

implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
adrianbenavides: Good point, I think it's better to return an error, yes.

adrianbenavides: done!

adrianbenavides:
@etorreborre I've updated the docs for the `to` and `via` arguments. Not sure if they are too verbose now, but I think the examples can help a lot in understanding the different use cases.

etorreborre:
@adrianbenavides thanks for expanding the help! However I think that arguments help should be kept short and the verbose explanation should go to the after_long_help text instead.

etorreborre (approved):

PR #7716, opened 2024-03-06
feat: bounded traces

etorreborre:
This PR makes sure that we don't send the tracing context across more than one node at the time:

 - When a `TransportMessage` is created we start a new trace before it is sent to another node. 
 - The trace that lead to creating this `TransportMessage` and the new trace are both linked together so it becomes possible to navigate traces and analyze, end-to-end, the processing of a given message.

This way, we:

 - Remove the possibility to correlate messages entering a node with messages leaving a node when they are part of the same communication route
 - Yet, we keep traces with spans that represent the full processing of messages, from the moment they are received (or from the moment that a user executes a command) to the moment they leave a node.

In this PR also:

 - A few more spans were introduced to display more activity for portals (it's far from complete though).
 - The node name, if available, is added to all spans in order to better understand traces as they cross several nodes.

**NOTE** the `tracing_context` field is still not enabled with this PR. We need first to finish testing the integration with the Controller and deploy a new version

### DIFF
--- a/implementations/rust/ockam/ockam_core/src/routing/message/transport_message.rs
+++ b/implementations/rust/ockam/ockam_core/src/routing/message/transport_message.rs
@@ -1,9 +1,15 @@
 #[cfg(feature = "std")]
 use crate::OpenTelemetryContext;
-use crate::{compat::vec::Vec, Message, Route};
+use crate::{compat::vec::Vec, Message, Route, OCKAM_TRACER_NAME};
 #[cfg(feature = "std")]
 use cfg_if::cfg_if;
 use core::fmt::{self, Display, Formatter};
+#[cfg(feature = "std")]
+use opentelemetry::{
+    global,
+    trace::{Link, SpanBuilder, TraceContextExt, Tracer},
+    Context,
+};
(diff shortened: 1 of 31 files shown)


implementations/rust/ockam/ockam_core/src/routing/message/transport_message.rs
```diff
@@ -59,8 +67,23 @@ impl TransportMessage {
     pub fn set_tracing_context(self, _tracing_context: OpenTelemetryContext) -> Self {
```
davide-baldo: Maybe it should be called `link_tracing_context` since we no longer retain the original context

davide-baldo (dismissed):

implementations/rust/ockam/ockam_core/src/routing/message/transport_message.rs
etorreborre: That's good idea. I renamed the function to `start_new_tracing_context` to give the idea that we are starting a brand new trace and left a more explicit comment.

davide-baldo (approved):

PR #7728, opened 2024-03-07
feat(rust): propagating the errors from api clients to the command

davide-baldo:
In this PR I mapped the error code from the api clients to a miette Report via a new method `miette_success`. This new method performs both `success()` and `into_diagnostic()` but without loosing any information (status code).
Change every method signature involved with the error passing to `miette::Result`.

Extras:
- Completely dropped anyhow (unused)
- Changed the formatting of errors in the custom miette error handler, otherwise `wrap_error()` completly obscure the previous error. 
- Delete duplicate error printing in `local_cmd()` (The method printing the error is command main)

See screenshot for thefinal  result.

ockam_new_error_reporting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4192,7 +4192,6 @@ dependencies = [
 name = "ockam_api"
 version = "0.60.0"
 dependencies = [
- "anyhow",
  "aws-config",
  "base64-url",
  "bytes 1.5.0",
@@ -4293,7 +4292,6 @@ dependencies = [
 name = "ockam_command"
 version = "0.117.0"
 dependencies = [
- "anyhow",
  "arboard",
  "assert_cmd",
  "async-trait",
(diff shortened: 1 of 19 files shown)


adrianbenavides (commented):
Awesome, it looks great! 

If you rebase on top of latest develop, I think it's safe now to revert the changes from https://github.com/build-trust/ockam/pull/7720. I manually tested it and the error doesn't show twice.

adrianbenavides:
Also, to keep the error codes consistent, we should rename the codes from [here](https://github.com/build-trust/ockam/blob/6ba710fe0101971ba7b1e48d1a3240078907ddf6/implementations/rust/ockam/ockam_command/src/error.rs#L16), removing the `OCK` prefix. What do you think?

davide-baldo:
@adrianbenavides
The alternative is prepend `OCK` the HTTP codes. Depends if we want to keep aligned only with HTTP errors or want to use it for every kind error in the future (like rust compiler error codes).

nazmulidris (dismissed):

nazmulidris (approved):

PR #7730, opened 2024-03-07
Account credentials

SanjoDeundiak:
Support getting and presenting project admin credentials to the Authority node


implementations/rust/ockam/ockam_identity/src/credentials/retriever/remote_retriever/info.rs
```diff
@@ -1,27 +1,72 @@
...
+#[derive(Debug, Clone)]
 pub struct RemoteCredentialRetrieverInfo {
     /// Issuer identity, used to validate retrieved credentials
     pub issuer: Identifier,
     /// Route used to establish a secure channel to the remote node
     pub route: Route,
-    /// Address of the credentials service on the remote node
-    pub service_address: Address,
+    /// Address of the credentials service on the remote node, e.g. "credential_issuer" or "accounts"
+    pub service_address: String,
```
adrianbenavides, line 48: Could we encode this as an enum?

implementations/rust/ockam/ockam_api/src/cli_state/trust.rs
```diff
@@ -1,12 +1,177 @@
...
 use ockam::identity::{IdentitiesVerification, RemoteCredentialRetrieverInfo};
 use ockam_core::errcode::{Kind, Origin};
 use ockam_core::{Error, Result};
 use ockam_multiaddr::MultiAddr;
 use ockam_vault::SoftwareVaultForVerifyingSignatures;
 
 impl CliState {
+    async fn retrieve_trust_options_explicit_project_authority(
+        &self,
+        authority_identity: &str,
```
adrianbenavides, line 16: Since we are passing this identity from the command, maybe we could add a type to make explicit that this is an encoded identity. That way we can also avoid duplicating the `hex::encode/decode` calls.

implementations/rust/ockam/ockam_command/src/util/api.rs
```diff
@@ -157,7 +157,7 @@ pub struct TrustOpts {
 
     /// Expect credential manually saved to the storage
     #[arg(long)]
-    pub expect_cached_credential: bool,
+    pub credential_scope: Option<String>,
```
adrianbenavides, line 160: What's the difference between the free-form scopes and the ones from `CredentialScope`?

implementations/rust/ockam/ockam_command/src/credential/store.rs
```diff
@@ -21,6 +21,9 @@ pub struct StoreCommand {
     #[arg(long = "issuer", value_name = "IDENTIFIER", value_parser = identity_identifier_parser)]
     pub issuer: Identifier,
 
+    #[arg(long = "scope", value_name = "CREDENTIAL_SCOPE")]
+    pub scope: String,
```
adrianbenavides, line 28: Can you add some context to this argument explaining how/why is this used?

implementations/rust/ockam/ockam_api/src/cli_state/trust.rs
SanjoDeundiak: Good idea, since there are a lot of commands like this I created an issue
https://github.com/build-trust/ockam/issues/7819

implementations/rust/ockam/ockam_command/src/util/api.rs
SanjoDeundiak: In the database they're strings and can have any value as long as it doesn't conflict with another credential. For example in bats tests I use "test" value. However, for project admin and project membership credentials I use a predefined format so we can identify where a credential belongs given its scope value. To be honest I don't think it's a perfect design, but we don't yet have a good use case for it either (except the bats tests)

implementations/rust/ockam/ockam_command/src/credential/store.rs
SanjoDeundiak: Sure

implementations/rust/ockam/ockam_identity/src/credentials/retriever/remote_retriever/info.rs
SanjoDeundiak: I would say that it can be potentially arbitrary value, but putting constants we use so far in an enum is a good idea

adrianbenavides (dismissed):

adrianbenavides (approved):

PR #7748, opened 2024-03-12
fix(rust): decode a transport message even without a tracing_context field

etorreborre:
This PR makes it possible to decode transport messages produced by nodes which are not yet emitting a `tracing_context` field.

*NOTE*: This PR does not enable the `tracing_context` field yet. For now:

  - It produces transport messages with protocol `v1` and no `tracing_context` field
  - Yet it can consume protocol `v2` messages (with a `tracing_context`) field

So once this PR is deployed in a new Orchestrator version, we will be able to switch to `v2` as the default protocol version. 

# New protocol version

In this PR we introduce a new protocol version, `v2`, where transport messages in `v2` have an additional `tracing_context` field.

That change was tested by deploying locally the code from this PR (using the [orchestrator docker-compose setup](https://github.com/build-trust/ockam-orchestrator/blob/main/compose/README.md)) and by sending a message using a secure channel to the project node, with:

 - A version of `ockam` based on `develop`, which only knows about transport messages in version `v1`.
 - This version of `ockam` which can read both `v1` and `v2`

Indeed, clients which understand only protocol v1 can communicate with:

 - The controller and the project node which, with orchestrator#555, just discard the `tracing_context` field and emit v1 messages
 - The authority node (based on this PR) which 
    - Accepts v1 messages and reply with v1 messages
    - Accepts v2 messages and reply with v2 messages 

## Implementation notes

In order to know which protocol version needs to be used to reply to a v1 or v2 messages, we add a new field to `LocalMessage` and make sure that each new `LocalMessage` that is propagated inside a node keep this information.
This is very similar to what we do for the `tracing_context` and that explains why we also need to add a `protocol_version` field to the `Context` that is setup for a worker.

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/wire.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/wire.ex
@@ -12,7 +12,7 @@ defmodule Ockam.Wire do
   require DecodeError
   require EncodeError
 
-  @default_implementation Ockam.Wire.Binary.V1
+  @default_implementation Ockam.Wire.Binary.Versions
 
   @doc """
   Encodes a message into a binary.
--- a/implementations/elixir/ockam/ockam/lib/ockam/wire/binary/versions.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/wire/binary/versions.ex
@@ -55,7 +55,7 @@ defmodule Ockam.Wire.Binary.Versions do
         context ->
           :bare.encode(
             %{
-              version: @latest_version,
+              version: @version_1,
               onward_route: normalize_route(onward_route),
               return_route: normalize_route(return_route),
               payload: payload,
--- a/implementations/elixir/ockam/ockam/lib/ockam/wire/binary/versions.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/wire/binary/versions.ex
@@ -1,4 +1,4 @@
-defmodule Ockam.Wire.Binary.V1 do
+defmodule Ockam.Wire.Binary.Versions do
   @moduledoc false
 
   @behaviour Ockam.Wire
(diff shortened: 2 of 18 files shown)


implementations/elixir/ockam/ockam/lib/ockam/wire.ex
```diff
@@ -12,7 +12,7 @@ defmodule Ockam.Wire do
   require DecodeError
   require EncodeError
 
-  @default_implementation Ockam.Wire.Binary.V1
+  @default_implementation Ockam.Wire.Binary.Versions
```
ckhrysze, line 15: I realize that I probably started us down this path, but in thinking through how this needs to be deployed now, and how a version 3 might work, I believe this should remain V1, the file v1.ex should be reverted to before I changed it, and that the majority of the changes we made should be in a v2.

implementations/elixir/ockam/ockam/lib/ockam/wire/binary/versions.ex
```diff
@@ -78,7 +79,7 @@ defmodule Ockam.Wire.Binary.V1 do
   def decode(encoded) do
     ## Expect first byte to be the version
     case encoded do
-      <<@version, _rest::binary>> ->
+      <<@version_1, _rest::binary>> ->
```
ckhrysze, line 82: I didn't pick up on this at first, but in furtherance of my comment above and the general goal, I think we should pick this byte off earlier in the processing and pick the implementation as appropiate.

ckhrysze (commented):
In both the rust at the elixir aspects of this change the use of 'latest' makes me nervous. There is some utility to the idea conceptually, but I feel it should be minimized to maybe a config or a default impl, and largely the specific version should appear in use. Its hard to imagine a version 3 right now without a lot of changes, but ideally a version 3 would be almost purely additive with just a single line of configuration marking it as latest.

polvorin (approved):

PR #7758, opened 2024-03-13
feat(rust): add command to import an identity

adrianbenavides:


polvorin (dismissed):
It looks fine to me. 

Instead of creating yet another `import_from_*` in implementations/rust/ockam/ockam_identity/src/identities/identities_verification.rs ,   I'd have   use `import_from_change_history(ChangeHistory::import_from_string(arg))` from command directly.   After all,  from command we call `ChangeHistory::export_as_string()` when exporting,  so  feels natural to call the opposite at the same abstraction level when importing.

implementations/rust/ockam/ockam_command/src/identity/import.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    async fn async_run(&self, opts: CommandGlobalOpts) -> miette::Result<()> {
+        let named_vault = opts.state.get_named_vault_or_default(&self.vault).await?;
+        let vault = named_vault.vault().await?;
+        let identifier = opts
+            .state
+            .make_identities(vault)
+            .await?
+            .identities_verification()
```
SanjoDeundiak: It can be instantiated with `SoftwareVaultForVerifyingSignatures`, unless there is purpose in initializing the full Vault instance here

implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
```diff
@@ -55,6 +55,18 @@ impl CliState {
             .await
     }
 
+    #[instrument(skip_all, fields(name = %name, identifier = %identifier, vault_name = %vault_name))]
+    pub async fn create_identity_with_identifier(
```
SanjoDeundiak: What is the difference between this one and `store_named_identity`? Is it that we instantiate the Vault here? What is the idea behind that?

SanjoDeundiak (changes requested):
Nice! Left few comments

implementations/rust/ockam/ockam_api/src/cli_state/identities.rs
adrianbenavides: I didn't see that method, I'll remove `create_identity_with_identifier`.

implementations/rust/ockam/ockam_command/src/identity/import.rs
adrianbenavides: No, there isn't. I thought we had to use the AWS vault to verify the identity.

SanjoDeundiak: No, AWS only implements `VaultForSigning` trait, and there is no such dependency for that operation 🙂

adrianbenavides: Got it, I've added that change.

SanjoDeundiak (approved):
Thank you!

PR #7763, opened 2024-03-15
Authority commands

SanjoDeundiak:
Add CRUD commands to work with a remote Project Membership Authority node

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/project_member/add.rs
+++ b/implementations/rust/ockam/ockam_command/src/project_member/add.rs
@@ -1,0 +1,95 @@
+use clap::Args;
+use colorful::Colorful;
+
+use ockam::identity::Identifier;
+use ockam::Context;
+use ockam_api::authenticator::direct::Members;
+use ockam_api::nodes::InMemoryNode;
+use ockam_multiaddr::MultiAddr;
+
+use crate::project_member::{create_authority_client, create_member_attributes, get_project};
+use crate::util::api::IdentityOpts;
+use crate::util::async_cmd;
+use crate::{docs, fmt_ok, CommandGlobalOpts};
+
+const LONG_ABOUT: &str = include_str!("./static/add/long_about.txt");
+const AFTER_LONG_HELP: &str = include_str!("./static/add/after_long_help.txt");
+
+/// This attribute in credential allows member to create a relay on the Project node, the name of the relay should be
+/// equal to the value of that attribute. If the value is `*` then any name is allowed
+pub const OCKAM_RELAY_ATTRIBUTE: &str = "ockam-relay";
+
+/// Add members to a Project, as an authorized enroller directly
+#[derive(Clone, Debug, Args)]
+#[command(
+long_about = docs::about(LONG_ABOUT),
+after_long_help = docs::after_help(AFTER_LONG_HELP),
+)]
+pub struct AddCommand {
+    #[command(flatten)]
+    identity_opts: IdentityOpts,
+
+    /// Which project add member to
+    #[arg(long, short, value_name = "ROUTE_TO_PROJECT")]
+    to: Option<MultiAddr>,
+
+    #[arg(value_name = "IDENTIFIER")]
+    member: Identifier,
+
+    /// Attributes in `key=value` format to be attached to the member. You can specify this option multiple times for multiple attributes
+    #[arg(short, long = "attribute", value_name = "ATTRIBUTE")]
+    attributes: Vec<String>,
+
+    /// Name of the relay that the identity will be allowed to create. This name is transformed into attributes to prevent collisions when creating relay names. For example: `--relay foo` is shorthand for `--attribute ockam-relay=foo`
+    #[arg(long = "relay", value_name = "ENROLLEE_ALLOWED_RELAY_NAME")]
+    allowed_relay_name: Option<String>,
+
+    /// Add the enroller role. If you specify it, this flag is transformed into the attributes `--attribute ockam-role=enroller`. This role allows the Identity using the ticket to enroll other Identities into the Project, typically something that only admins can do
+    #[arg(long = "enroller")]
+    enroller: bool,
+}
+
+impl AddCommand {
+    pub fn run(self, opts: CommandGlobalOpts) -> miette::Result<()> {
+        async_cmd(&self.name(), opts.clone(), |ctx| async move {
+            self.async_run(&ctx, opts).await
+        })
+    }
+
+    pub fn name(&self) -> String {
+        "project-member add".into()
+    }
+
+    async fn async_run(&self, ctx: &Context, opts: CommandGlobalOpts) -> miette::Result<()> {
+        let project = get_project(&opts.state, &self.to).await?;
+
+        let node = InMemoryNode::start_with_project_name(
+            ctx,
+            &opts.state,
+            Some(project.name().to_string()),
+        )
+        .await?;
+
+        let authority_node_client =
+            create_authority_client(&node, &opts.state, &self.identity_opts, &project).await?;
+
+        authority_node_client
+            .add_member(
+                ctx,
+                self.member.clone(),
+                create_member_attributes(
+                    &self.attributes,
+                    &self.allowed_relay_name,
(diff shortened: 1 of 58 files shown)


implementations/rust/ockam/ockam_command/src/project_member/add.rs
```diff
@@ -0,0 +1,107 @@
...
+
+impl AddCommand {
+    pub fn run(self, opts: CommandGlobalOpts) -> miette::Result<()> {
+        async_cmd(&self.name(), opts.clone(), |ctx| async move {
+            self.async_run(&ctx, opts).await
+        })
+    }
+
+    pub fn name(&self) -> String {
+        "create project ticket".into()
```
adrianbenavides: This should be exactly the name of the command as we type it in the console (`project-member add`), so we can quickly identify the command name in the logs/tracing and also to properly parse the command when working with configuration files used by `run` and `node create` commands.

implementations/rust/ockam/ockam_command/src/project_member/add.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    async fn async_run(&self, ctx: &Context, opts: CommandGlobalOpts) -> miette::Result<()> {
+        if opts.global_args.output_format == OutputFormat::Json {
+            return Err(miette::miette!(
+            "This command only outputs a hex encoded string for 'ockam project enroll' to use. \
+            Please try running it again without '--output json'."
+        ));
+        }
```
adrianbenavides: This is not relevant for this command right?

implementations/rust/ockam/ockam_command/src/project_member/add.rs
```diff
@@ -0,0 +1,107 @@
...
+    identity_opts: IdentityOpts,
+
+    /// Which project add member to
+    #[arg(
+        long,
+        short,
+        default_value = "/project/default",
+        value_name = "ROUTE_TO_PROJECT"
+    )]
+    to: MultiAddr,
```
adrianbenavides: To avoid having the `/project/default` hardcoded (at some point we'll use a random default name), we could use here an option, and then in the command handler use this to get the real default name:

```rust
let project_name = self
    .to
    .as_ref()
    .map(|ma| {
        ma.iter()
            .next()
            .map(|p| p.cast::<proto::Project>().to_string())
    })
    .flatten();
opts.state
    .projects()
    .get_project_by_name_or_default(&project_name);
```

And maybe we could extract the `project_name` extraction, I'm pretty sure we'll use it in more commands.

implementations/rust/ockam/ockam_command/src/project_member/delete.rs
```diff
@@ -0,0 +1,83 @@
...
+
+impl DeleteCommand {
+    pub fn run(self, opts: CommandGlobalOpts) -> miette::Result<()> {
+        async_cmd(&self.name(), opts.clone(), |ctx| async move {
+            self.async_run(&ctx, opts).await
+        })
+    }
+
+    pub fn name(&self) -> String {
+        "create project ticket".into()
```
adrianbenavides: ⚠️

implementations/rust/ockam/ockam_command/src/project_member/delete.rs
```diff
@@ -0,0 +1,83 @@
...
+)]
+pub struct DeleteCommand {
+    #[command(flatten)]
+    identity_opts: IdentityOpts,
+
+    /// Which project's member to delete
+    #[arg(
+        long,
+        short,
+        default_value = "/project/default",
```
adrianbenavides: ⚠️

implementations/rust/ockam/ockam_command/src/project_member/delete.rs
```diff
@@ -0,0 +1,83 @@
...
+        "create project ticket".into()
+    }
+
+    async fn async_run(&self, ctx: &Context, opts: CommandGlobalOpts) -> miette::Result<()> {
+        if opts.global_args.output_format == OutputFormat::Json {
+            return Err(miette::miette!(
+                "This command only outputs text. \
+            Please try running it again without '--output json'."
+            ));
+        }
```
adrianbenavides: I think we should use the `TUI` traits in these commands to have the same experience as with other CRUD commands. Did you find any inconvenient or something like that? I can add this change in a separate PR.

implementations/rust/ockam/ockam_command/src/project_member/list.rs
```diff
@@ -0,0 +1,98 @@
...
+)]
+pub struct ListCommand {
+    #[command(flatten)]
+    identity_opts: IdentityOpts,
+
+    /// Which project members to request
+    #[arg(
+        long,
+        short,
+        default_value = "/project/default",
```
adrianbenavides: ⚠️

implementations/rust/ockam/ockam_command/src/project_member/list.rs
```diff
@@ -0,0 +1,98 @@
...
+
+impl ListCommand {
+    pub fn run(self, opts: CommandGlobalOpts) -> miette::Result<()> {
+        async_cmd(&self.name(), opts.clone(), |ctx| async move {
+            self.async_run(&ctx, opts).await
+        })
+    }
+
+    pub fn name(&self) -> String {
+        "list project members".into()
```
adrianbenavides: This should be exactly the name of the command as we type it in the console (`project-member list`)

implementations/rust/ockam/ockam_command/src/project_member/list_ids.rs
```diff
@@ -0,0 +1,101 @@
...
+
+impl ListIdsCommand {
+    pub fn run(self, opts: CommandGlobalOpts) -> miette::Result<()> {
+        async_cmd(&self.name(), opts.clone(), |ctx| async move {
+            self.async_run(&ctx, opts).await
+        })
+    }
+
+    pub fn name(&self) -> String {
+        "create project ticket".into()
```
adrianbenavides: ⚠️

implementations/rust/ockam/ockam_command/src/project_member/list_ids.rs
```diff
@@ -0,0 +1,101 @@
...
+)]
+pub struct ListIdsCommand {
+    #[command(flatten)]
+    identity_opts: IdentityOpts,
+
+    /// Which project members to request
+    #[arg(
+        long,
+        short,
+        default_value = "/project/default",
```
adrianbenavides: ⚠️

implementations/rust/ockam/ockam_command/src/project_member/delete.rs
SanjoDeundiak: I copied it from `project ticket` command

implementations/rust/ockam/ockam_command/src/project_member/add.rs
SanjoDeundiak: You mean the message should be updated?

implementations/rust/ockam/ockam_command/src/project_member/add.rs
SanjoDeundiak: Good point

implementations/rust/ockam/ockam_command/src/project_member/delete.rs
adrianbenavides: Ok, I'll take a look and see if we can add that.

implementations/rust/ockam/ockam_command/src/project_member/add.rs
adrianbenavides: What I meant is that this command is not outputting any hex encoded string, nor it has a json output set up. Actually, in [this line](https://github.com/build-trust/ockam/blob/ceb34eaa3f2912f52751745d4d712852e740d546/implementations/rust/ockam/ockam_command/src/project_member/add.rs#L100), it should call `opts.terminal.stdout().plain(&fmt_ok!(...))`.

adrianbenavides (approved):

PR #7784, opened 2024-03-20
feat(rust): add one second cache for incoming and outgoing access control

davide-baldo:
ABAC Policy can have a "slow" (few milliseconds) evaluation, adding a cache for 1 second is more than enough to remove any performance penality.
The cache is only available on `std` due to the `Instant` dependency. I could have used `now()` but it wouldn't be as robust.

### DIFF
--- a/implementations/rust/ockam/ockam_core/src/access_control/cache.rs
+++ b/implementations/rust/ockam/ockam_core/src/access_control/cache.rs
@@ -1,0 +1,349 @@
+use crate::compat::sync::Mutex;
+use crate::{
+    Address, IncomingAccessControl, LocalInfo, OutgoingAccessControl, RelayMessage, Route,
+};
+use alloc::vec::Vec;
+use async_trait::async_trait;
+use core::fmt::Debug;
+use std::time::Instant;
+
+const CACHE_MAX_SIZE: usize = 10;
+const CACHE_DURATION_SECS: u64 = 1;
+
+#[derive(Debug)]
+struct CacheEntry {
+    source: Address,
+    destination: Address,
+    onward_route: Route,
+    return_route: Route,
+    local_info: Vec<LocalInfo>,
+    timestamp: Instant,
+}
+
+impl CacheEntry {
+    fn from(relay_message: &RelayMessage) -> Self {
+        Self {
+            source: relay_message.source().clone(),
+            destination: relay_message.destination().clone(),
+            onward_route: relay_message.onward_route().clone(),
+            return_route: relay_message.return_route().clone(),
+            local_info: relay_message.local_message().local_info(),
+            timestamp: Instant::now(),
+        }
+    }
+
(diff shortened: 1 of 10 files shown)


implementations/rust/ockam/ockam_core/src/access_control/cache.rs
```diff
@@ -0,0 +1,351 @@
...
+            let entry = cache_guard.get_mut(position).unwrap();
+            if entry.is_expired() {
+                cache_guard.remove(position);
+                false
+            } else {
+                true
+            }
+        } else {
+            false
+        }
```
adrianbenavides: An alternative:

```rust
cache_guard
    .iter()
    .position(|entry| entry.matches(relay_message))
    .map(|position| {
        if cache_guard[position].is_expired() {
            cache_guard.remove(position);
            false
        } else {
            true
        }
    })
    .unwrap_or(false);
```

implementations/rust/ockam/ockam_core/src/access_control/cache.rs
```diff
@@ -0,0 +1,351 @@
...
+
+    /// Adds the relay message to the cache.
+    pub fn add_authorized(&self, relay_message: &RelayMessage) {
+        let mut cache_guard = self.cache.lock().unwrap();
+        let position = cache_guard
+            .iter()
+            .position(|entry| entry.matches(relay_message));
+        if let Some(position) = position {
+            cache_guard.remove(position);
+        }
```
adrianbenavides, line 88: Just for readability, I'd write:

```rust
cache_guard
    .iter()
    .position(|entry| entry.matches(relay_message))
    .map(|position| cache_guard.remove(position));
```

adrianbenavides (dismissed):
Great, thanks for adding this 🙏  I left a couple of minor comments, feel free to ignore them.

implementations/rust/ockam/ockam_core/src/access_control/cache.rs
davide-baldo: so much better!

implementations/rust/ockam/ockam_core/src/access_control/cache.rs
davide-baldo: Having side effects on `map()` without reading the output is dangerous, since it could be lazy and never be evaluated (and it's already lazy on iterators). Ideally, `inspect()` would be a better alternative, but it's [not stable yet ](https://github.com/rust-lang/rust/issues/91345).

adrianbenavides: Oh right! I didn't try to compile my proposal, but makes sense that is not a good idea because we are iterating and removing from the same source right? I don't think it even compiles. Good to know about that `inspect` function, never seen it before.

adrianbenavides (approved):

PR #7791, opened 2024-03-21
refactor(rust): simplify `ProgressDisplay` to remove the external mutex used to stop the message recv end

adrianbenavides:
- Simplify `ProgressDisplay` to remove the external mutex used to stop the message recv end
- Use `ProgressDisplay` in the `identity create` command to show the vault creation messages


etorreborre (approved):
Nice change! 🖌️

PR #7808, opened 2024-03-25
tests(rust): add bats coverage for `node create ./config.yaml` command

adrianbenavides:
- Add bats coverage for `node create ./config.yaml` command. All bats passing locally ✅ 
- Fixes passing URLs for `enrollment ticket` and `config file` without scheme (defaults to `http://`), i.e. `--enrollment ticket 127.0.0.1:1234/my.ticket` will be processed as `--enrollment ticket http://127.0.0.1:1234/my.ticket`
- Using an enrollment ticket twice doesn't return an error now. This is useful so that the user doesn't have to modify the config file or command after enrolling once.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/enroll/enrollment.rs
+++ b/implementations/rust/ockam/ockam_api/src/enroll/enrollment.rs
@@ -32,7 +32,11 @@ pub trait Enrollment {
         token: OidcToken,
     ) -> miette::Result<()>;
 
-    async fn present_token(&self, ctx: &Context, token: &OneTimeCode) -> miette::Result<()>;
+    async fn present_token(
+        &self,
+        ctx: &Context,
+        token: &OneTimeCode,
+    ) -> miette::Result<EnrollStatus>;
 
     async fn issue_credential(&self, ctx: &Context) -> miette::Result<CredentialAndPurposeKey>;
 }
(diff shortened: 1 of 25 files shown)


implementations/rust/ockam/ockam_api/src/enroll/enrollment.rs
```diff
@@ -110,14 +118,31 @@ impl Enrollment for SecureClient {
...
+        match self
+            .tell(ctx, DefaultAddress::ENROLLMENT_TOKEN_ACCEPTOR, req)
             .await
             .into_diagnostic()?
-            .success()
-            .into_diagnostic()
+        {
+            Reply::Successful(_) => Ok(EnrollStatus::EnrolledSuccessfully),
+            Reply::Failed(e, s) => match (e.message(), s) {
+                // TODO: the `authenticator` should return proper error codes
```
davide-baldo, line 135: The authenticator could return a `201` when adding it, and a `200` when it already exists

davide-baldo (approved):

implementations/rust/ockam/ockam_api/src/enroll/enrollment.rs
adrianbenavides: Yeah, I briefly talked about this with @SanjoDeundiak. The main problem is that it requires a big refactor to change how results/errors are handled in that module.

PR #7816, opened 2024-03-26
feat(rust): switch to `aws-lc-rs` library for encryption/decryption

davide-baldo:
`aws-lc-rs` library (which is based on an OpenSSL fork) is ~8 times faster than our current `aes-gcm`.
With some extra adjustments, the performance goes from `~6 Gbits` to `~15 Gbits` on my amd64 laptop.

**This library requires `cmake` and a C compiler (either `gcc` or `clang`).**
I've put the `aws-lc` library behind the `ockam_vault/aws-lc` feature and enabled it on CI builds.


implementations/rust/ockam/ockam_vault/Cargo.toml
```diff
@@ -30,14 +30,13 @@ disable_default_noise_protocol = []
...
 OCKAM_XX_25519_AES128_GCM_SHA256 = []
 OCKAM_XX_25519_ChaChaPolyBLAKE2s = []
+aws-lc = ["dep:aws-lc-rs"]
 
 # Feature (enabled by default): "std" enables functionality expected to
 # be available on a standard platform.
 std = [
   "ockam_macros/std",
   "ockam_node/std",
-  "aes-gcm/alloc",
```
SanjoDeundiak, line 39: Could you please elaborate on this? Is there any reason we don't want `alloc` and `std` for `aes-gcm`?

implementations/rust/ockam/ockam_vault/Cargo.toml
```diff
@@ -30,14 +30,13 @@ disable_default_noise_protocol = []
 OCKAM_XX_25519_AES256_GCM_SHA256 = []
 OCKAM_XX_25519_AES128_GCM_SHA256 = []
 OCKAM_XX_25519_ChaChaPolyBLAKE2s = []
+aws-lc = ["dep:aws-lc-rs"]
```
SanjoDeundiak, line 33: What are the downside for using aws lib? If it's better, should we use it by default?

implementations/rust/ockam/ockam_vault/Cargo.toml
```diff
@@ -73,6 +72,7 @@ storage = ["ockam_node/storage", "sqlx"]
 [dependencies]
 aes-gcm = { version = "0.10", default-features = false, features = ["aes"] }
```
SanjoDeundiak: Should we exclude that dependency if aws-lc is used?

implementations/rust/ockam/ockam_vault/src/software/vault_for_secure_channels/types.rs
```diff
@@ -44,9 +44,6 @@ cfg_if! {
         /// AEAD Secret Length.
         pub const AEAD_SECRET_LENGTH: usize = AES256_SECRET_LENGTH;
 
-        /// AES-GCM nonce length
-        pub const AES_NONCE_LENGTH: usize = 12;
```
SanjoDeundiak: Could you please elaborate what are the reasons to move it?

implementations/rust/ockam/ockam_vault/src/software/vault_for_secure_channels/aes_aws_lc.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+impl AesGen {
+    fn encrypt_in_place_detached(
+        &self,
+        nonce: Nonce,
+        aad: &[u8],
+        buffer: &mut [u8],
+    ) -> Result<Tag, Unspecified> {
+        let unbound_key = UnboundKey::new(&AES_TYPE, &self.0 .0).unwrap();
```
SanjoDeundiak, line 89: I'm concerned if memory is properly zeroized here... We do that for our `AeadSecret` type, but that data is copied here and I don't see if `UnboundKey` or `LessSafeKey` is ever zeroized. Could you please check it?

implementations/rust/ockam/ockam_vault/Cargo.toml
davide-baldo: It requires `cmake` and `libclang`, also it doesn't support `no_std`

implementations/rust/ockam/ockam_vault/Cargo.toml
davide-baldo: I couldn't find a way to use “negative” dependency.
Either you make the library user choose one of the two features (`aws-lc` or `aes-gcm`) or you always build the default one.

implementations/rust/ockam/ockam_vault/src/software/vault_for_secure_channels/aes_aws_lc.rs
davide-baldo: When the pointer is dropped, it'll call OpenSSL free key function, and inside it `BN_clear_free` [1][2] will "overwrite" the memory before freeing. 

[1] https://github.com/openssl/openssl/blob/master/crypto/ec/ec_key.c#L100C5-L100C18
[2] https://www.openssl.org/docs/man1.1.1/man3/BN_clear_free.html

implementations/rust/ockam/ockam_vault/src/software/vault_for_secure_channels/types.rs
davide-baldo: It was needed only by the `aes-gcm` implementation but I've just noticed it get exposed publicly and moved it back :+1:

implementations/rust/ockam/ockam_vault/Cargo.toml
davide-baldo: The code doesn't work without `alloc` so in the proposal i've moved it to the default feature. Then changed my mind to address this in the future with other `nostd` related changes, but forgot to restore it.
(`aws-gcm/alloc` was also a repetition already covered by the `alloc` feature)

implementations/rust/ockam/ockam_vault/src/software/vault_for_secure_channels/aes_aws_lc.rs
davide-baldo: I've also explicit passed `zeroize` to `aes-gcm`, just in case.

davide-baldo:
I've added another change: switching from `native-tls` to `rustls` with optional `aws-lc-rs` dependency

implementations/rust/ockam/ockam_vault/src/software/vault_for_secure_channels/aes_aws_lc.rs
SanjoDeundiak: It happens inside `OPENSSL_cleanse`. ec_key is for elliptic curve keys. But anyways, it happens and that's good.

implementations/rust/ockam/ockam_vault/Cargo.toml
SanjoDeundiak: Oh, the repetition part makes total sense

SanjoDeundiak (approved):

PR #7818, opened 2024-03-27
fix(rust): kms identity can be used in regular api nodes

adrianbenavides:
The KMS vault was using the stateless implementation for the `VaultForSecureChannels` and it wasn't persisting necessary information (e.g. the `x25519 secrets`) in the local database.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
+++ b/implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
@@ -249,7 +249,7 @@ pub fn logging_configuration(
     crates: CratesFilter,
 ) -> ockam_core::Result<LoggingConfiguration> {
     let is_legacy_variable_set = is_set::<LevelVar>(OCKAM_LOG)?;
-    let enabled = if is_legacy_variable_set {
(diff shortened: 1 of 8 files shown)


implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
```diff
@@ -249,7 +249,7 @@ pub fn logging_configuration(
     crates: CratesFilter,
 ) -> ockam_core::Result<LoggingConfiguration> {
     let is_legacy_variable_set = is_set::<LevelVar>(OCKAM_LOG)?;
-    let enabled = if is_legacy_variable_set {
+    let enabled = if is_legacy_variable_set || preferred_log_level.is_some() {
```
adrianbenavides, line 252: This change here fixes the following behavior: the following won't print any logs, which is confusing, because the user is already showing the intent of enabling logging.

```rust
unset OCKAM_LOGGING
ockam status -vvv 
```

We shouldn't force the user to set both the `OCKAM_LOGGING` env var AND pass the verbose flag.

etorreborre (approved):

SanjoDeundiak (approved):

PR #7820, opened 2024-03-29
feat(elixir): added packet counter support

davide-baldo:
Adding packet counter support to the elixir side, the idea is to have a 2-step deployment, first the elixir side and then “unlock” the rust side.

To have a more lenient parser, and to better match the rust side, I've added the  `tuple` schema in BARE parsing.

I haven't written in Erlang in years and never used Elixir before, I'm pretty sure I made silly mistakes!

### DIFF
--- a/implementations/elixir/ockam/ockam/lib/ockam/protocol/bare_extended.ex
+++ b/implementations/elixir/ockam/ockam/lib/ockam/protocol/bare_extended.ex
@@ -81,17 +81,20 @@ defmodule Ockam.Bare.Variant do
   The tag is encoded as bare enum, optionally followed by the field value in case
   the variant has one.
   """
+
+  alias Ockam.Bare.Extended
+
   @type schema :: :bare.spec()
   @type extended_schema() :: schema() | {:variant, [atom() | {atom(), schema()}]}
 
   @spec encode(any(), extended_schema()) :: binary()
   def encode(value, {:variant, ss} = schema) do
-    type = :bare.encode(enum_member(value), to_bare_schema(schema))
+    type = encode(enum_member(value), to_bare_schema(schema))
     value = encode_value(enum_value(value), List.keyfind(ss, enum_member(value), 0))
     <<type::binary, value::binary>>
   end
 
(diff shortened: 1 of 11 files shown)


implementations/elixir/ockam/ockam/lib/ockam/protocol/bare_extended.ex
```diff
@@ -126,17 +129,100 @@ defmodule Ockam.Bare.Variant do
...
+
+  @spec decode(binary(), extended_schema()) :: {:ok, any()} | {:error, any()}
+  def decode(data, {:tuple, [type | types]} = _schema) do
+    case decode(data, type) do
+      {:ok, value, rest} ->
+        with {:ok, values, ""} <- decode(rest, {:tuple, types}) do
+          {:ok, [value] ++ values, <<>>}
+        end
+
+      {:ok, value} ->
```
polvorin: something is fishy here.    If decode/2  returns `{:ok, value}` one would assume it would be because there is no more input.    In that case,  we shouldn't accept the thing,  _unless_  there are also no remaining non-optional fields on the schema being decoded (`types == []`).  Otherwise   things like `decode( encode([1,2], {:tuple, [:int, :int]}),   {:tuple, [:int, :int, :int, :int]})`  would return a 2 element tuple even when it should be 4 and none is optional.

*However,  that is not happening*.    I suspect because decode/2  _never_ returns `{:ok, value}` .   It returns `{:ok, value, ""}`.    That's what  `:bare.decode/2` returns.    

I'd suggest re-checking if that's the case,  if so fix the typespec for `decode/2` and remove this clause.

polvorin (changes requested):

implementations/elixir/ockam/ockam/lib/ockam/protocol/bare_extended.ex
davide-baldo: You were right! The clause was unnecessary, i've also updated the typespec to reflect the binary being returned.

polvorin (approved):

SanjoDeundiak (approved):

PR #7822, opened 2024-03-29
feat(rust): support TLS for outlets

etorreborre:
This PR makes it possible to point tcp-outlets toward an endpoint supporting TLS.
This way, a TCP portal can be configured to send data to `https://myservice:443`.

# Implementation notes 

## Certificates

The certificates are loaded from the system certificates using the [rustls-native-certs](https://crates.io/crates/rustls-native-certs) crate.

## Configuration

The configuration parameter for the `tcp-outlet create` command `--to` option is now a `HostnamePort` which is simply a host and a port, instead of a socket address. This is necessary in order to keep the domain name (`otelcoll.orchestrator.ockam.io` for example) as long as possible before being resolved to a socket address. This domain name is then used to check certificates.

I did not propagate this configuration to every place, in particular not to the Kafka configuration yet. I provided instead a conversion function from `SocketAddr` to `HostnamePort`.

### DIFF
--- a/implementations/rust/ockam/ockam_transport_tcp/src/transport/common.rs
+++ b/implementations/rust/ockam/ockam_transport_tcp/src/transport/common.rs
@@ -1,11 +1,22 @@
-use crate::TcpConnectionMode;
+use crate::{HostnamePort, TcpConnectionMode};
+use cfg_if::cfg_if;
 use core::fmt;
 use core::fmt::Formatter;
+use native_tls::Certificate;
+use native_tls::TlsConnector as NativeTlsConnector;
 use ockam_core::compat::net::{SocketAddr, ToSocketAddrs};
+use ockam_core::errcode::{Kind, Origin};
 use ockam_core::flow_control::FlowControlId;
-use ockam_core::{Address, Result};
+use ockam_core::{Address, Error, Result};
 use ockam_node::Context;
 use ockam_transport_core::TransportError;
+use socket2::{SockRef, TcpKeepalive};
+use std::time::Duration;
+use tokio::io::{ReadHalf, WriteHalf};
+use tokio::net::tcp::{OwnedReadHalf, OwnedWriteHalf};
+use tokio::net::TcpStream;
+use tokio_native_tls::TlsStream;
+use tracing::{debug, instrument};
 
 /// Result of [`TcpTransport::connect`] call.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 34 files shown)


implementations/rust/ockam/ockam_transport_tcp/src/transport/hostname_port.rs
```diff
@@ -0,0 +1,128 @@
...
+        match values.next() {
+            Some(port) => match port.parse::<u16>().ok() {
+                Some(port) => {
+                    let hostname = values.map(|v| v.to_string()).rev().collect::<Vec<_>>();
+                    let hostname = if hostname.is_empty()
+                        || (hostname.len() == 1 && hostname.join("").is_empty())
+                    {
+                        "127.0.0.1".to_string()
+                    } else {
+                        hostname.join(":")
```
davide-baldo: This parsing is very lenient and accepts something ambiguous, like  `example.com:80:443`.
Since no hostname will ever contain “:”, we should be stricter to avoid any possible exploitation.

implementations/rust/ockam/ockam_transport_tcp/src/transport/hostname_port.rs
```diff
@@ -0,0 +1,128 @@
+use crate::resolve_peer;
+use core::fmt::{Display, Formatter};
+use core::str::FromStr;
+use minicbor::{Decode, Encode};
+use ockam_core::errcode::{Kind, Origin};
+use std::net::SocketAddr;
+
+/// Hostname and port
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+pub struct HostnamePort {
```
davide-baldo, line 10: What about `Endpoint` instead of `HostnamePort`?

implementations/rust/ockam/ockam_transport_tcp/src/transport/common.rs
```diff
@@ -154,6 +165,116 @@ pub(super) fn parse_socket_addr(s: &str) -> Result<SocketAddr> {
...
+#[allow(clippy::type_complexity)]
+#[instrument(skip_all)]
+pub(crate) async fn connect_tls(
+    hostname_port: &HostnamePort,
+) -> Result<(
+    ReadHalf<TlsStream<TcpStream>>,
+    WriteHalf<TlsStream<TcpStream>>,
+)> {
+    // Don't try to establish a TLS connection locally
+    if ["127.0.0.1", "localhost"].contains(&hostname_port.hostname().as_str()) {
```
davide-baldo: what about IPv6?

implementations/rust/ockam/ockam_transport_tcp/tests/portal.rs
```diff
@@ -55,7 +59,7 @@ async fn read_should_timeout(stream: &mut TcpStream) {
 }
 
 #[allow(non_snake_case)]
-#[ockam_macros::test(timeout = 5000)]
+#[ockam_macros::test(timeout = 5000000)]
```
davide-baldo: I think you forgot to revert it

implementations/rust/ockam/ockam_transport_tcp/src/transport/common.rs
etorreborre: Funny that you see this, I have removed those lines since.

implementations/rust/ockam/ockam_transport_tcp/src/transport/hostname_port.rs
etorreborre: I chose a name that was simply describing anything that comes before the last `:` and the port. There is already an `Endpoint` in the `tonic` library which is much richer than what I need to describe here.

implementations/rust/ockam/ockam_transport_tcp/src/transport/hostname_port.rs
etorreborre: Thanks, I have changed this with some regex parsing to be a bit more strict.

davide-baldo (dismissed):

.github/workflows/release-draft-binaries.yml
```diff
@@ -182,169 +182,169 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        build: [linux_arm64, linux_86, linux_armv7, macos_silicon, macos_86]
+        build: [ linux_arm64, linux_86, linux_armv7, macos_silicon, macos_86 ]
```
metaclips: Please undo changes made to this file

metaclips (approved):

PR #7825, opened 2024-03-29
Release improvement

metaclips:
Reduces time required to build our binaries by using macos m1 runners and latest version of Cross. Build time got reduced from >30min to <16min


etorreborre (approved):
Nice one!

PR #7849, opened 2024-04-04
docs(rust): improve the example for connecting to an ai model

etorreborre:
This PR improves the example for connecting to an AI model using a portal:

 - Fixed some typos.
 - Don't download `ockam` on cleanup
 - Check if the instance type required to run the model is available in the current region (otherwise return the list of regions where it is supported).

### DIFF
--- a/examples/command/portals/ai/amazon_ec2/aws_cli/ai_corp/run.sh
+++ b/examples/command/portals/ai/amazon_ec2/aws_cli/ai_corp/run.sh
@@ -28,7 +28,7 @@ run() {
 
     # Create a security group to allow:
     #   - TCP egress to the Internet
-    #   - Postgess ingress only from withing our two subnets.
+    #   - SSH ingress only from withing our two subnets.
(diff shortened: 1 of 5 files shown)


examples/command/portals/ai/amazon_ec2/aws_cli/ai_corp/run.sh
```diff
@@ -64,10 +67,58 @@ run() {
...
             nohup node api.mjs &>output.log &
+            echo "AI API started"
 EOS
 }
 
+# This function checks if an instance type is available in the currently configured region
+# If not, then the list of all regions supporting that instance type are returned
+check_if_instance_type_available() {
+    instance_type=$1
+    current_region=$(get_configured_region)
```
mrinalwadhwa, line 81: I learnt recently that this is the best way to get region because the it gets it from an api response instead of deciphering it from config and env vars

```
region=$(aws ec2 describe-availability-zones --query 'AvailabilityZones[0].[RegionName]')
```

etorreborre: Thanks, I adjusted the script.

mrinalwadhwa (approved):

PR #7862, opened 2024-04-08
test(rust): fix curl command on telegraf-influx bats test

adrianbenavides:


etorreborre (dismissed):

metaclips (changes requested):
There are other command that we didn't append the --header flag 
https://github.com/build-trust/ockam/blob/3af5c9a132c34f147bb5805e50807e1cd17f24d4/implementations/rust/ockam/ockam_command/tests/bats/orchestrator/use_cases.bats#L93-L98

adrianbenavides:
Fixed @metaclips

metaclips (approved):

PR #7884, opened 2024-04-10
feat(rust): added kafka-inlet command and relative config side

davide-baldo:


implementations/rust/ockam/ockam_command/src/kafka/inlet/create.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// The address of the kafka bootstrap broker
+    #[arg(long)]
+    pub bootstrap_server: Option<String>,
+    /// Local port range dynamically allocated to kafka brokers, must not overlap with the
+    /// bootstrap port
+    #[arg(long)]
+    pub brokers_port_range: Option<PortRange>,
+    /// The route to the Kafka outlet node, either the project in ockam orchestrator or a rust node, expected something like /project/<name>
+    #[arg(long)]
+    pub to: Option<MultiAddr>,
```
adrianbenavides, line 55: If only `bootstrap_server` or `to` can be used at the same time, you could add `conflicts_with = "bootstrap_server"` here.

implementations/rust/ockam/ockam_command/src/kafka/inlet/create.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// The route to the Kafka consumer node, valid only when --bootstrap-server is specified
+    #[arg(long)]
+    pub consumer: Option<MultiAddr>,
```
adrianbenavides, line 58: You can add `requires = "bootstrap_server"` to enforce that dependency between args.

implementations/rust/ockam/ockam_command/src/node/create/demo_config_files/2.kafka.single-machine.yaml
```diff
@@ -0,0 +1,8 @@
+name: kafka-single-machine
```
adrianbenavides: Maybe `kafka-portal` or `kafka-node` would be a more descriptive name for the node?

implementations/rust/ockam/ockam_command/src/run/demo_config_files/4.kafka.inlet.yaml
```diff
@@ -0,0 +1,6 @@
+name: kafka_inlet_node
```
adrianbenavides: In the `run` command nodes are described as regular resources:

```yaml
nodes: kafka_inlet_node
```

The `parse_demo_config_files` test won't fail because it will skip the `name` field, but if you look into what it got parsed, you'll see no nodes.

implementations/rust/ockam/ockam_command/src/run/demo_config_files/4.kafka.outlet.yaml
```diff
@@ -0,0 +1,8 @@
+name: kafka_outlet_node
+
+ticket: kafka.ticket
+
+tcp-listener-address: 127.0.0.1:6000
```
adrianbenavides: Similarly, all the node's args shoud be defined in its own block:

```yaml
nodes:
  kafka_outlet_node:
    tcp-listener-address: 127.0.0.1:6000
```

adrianbenavides (dismissed):

adrianbenavides (approved):

PR #7891, opened 2024-04-11
feat(rust): support progress_bar in command notifications

adrianbenavides:
It also fixes the messages formatting generated when creating the default identity and vault.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/ui/terminal/notification.rs
+++ b/implementations/rust/ockam/ockam_api/src/ui/terminal/notification.rs
@@ -9,12 +9,36 @@ use tokio::select;
 use tokio::sync::broadcast::Receiver;
 use tokio::time::sleep;
 
-use tracing::info;
-
 const REPORTING_CHANNEL_POLL_DELAY: Duration = Duration::from_millis(100);
-const REPORTING_CHANNEL_MESSAGE_DISPLAY_DELAY: Duration = Duration::from_millis(1_000);
 
-pub type Notification = String;
+#[derive(Debug, Clone, PartialEq)]
+pub enum Notification {
+    Message(String),
+    ProgressSet(String),
+    ProgressFinish(Option<String>),
+}
+
+impl Notification {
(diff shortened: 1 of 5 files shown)


implementations/rust/ockam/ockam_api/src/ui/terminal/notification.rs
```diff
@@ -9,12 +9,52 @@ use tokio::select;
...
 use tokio::time::sleep;
 
-use tracing::info;
-
 const REPORTING_CHANNEL_POLL_DELAY: Duration = Duration::from_millis(100);
-const REPORTING_CHANNEL_MESSAGE_DISPLAY_DELAY: Duration = Duration::from_millis(1_000);
 
-pub type Notification = String;
+#[derive(Debug, Clone)]
+pub struct Notification {
```
etorreborre: I think it is better to model this as an enum from the start:
```rust
pub enum Notification {
   Message(String),
   ProgressSet(String),
   ProgressFinish(Option<String>),
}
```
This is more precise and then we don't have to worry down the road if the "content" exists or not, we can just pattern match on the 3 cases.

adrianbenavides: I started with that, but it didn't feel very good in this [function](https://github.com/build-trust/ockam/blob/e30bcfe7a71fbd0541cb41ce78435ea46eadff08/implementations/rust/ockam/ockam_api/src/ui/terminal/notification.rs#L118) with lots of `if let Some(c) = ...`.

I'll revert back to that.

adrianbenavides: done! I don't know what I did in my first attempt with the `content` field, but it now looks as clean as it can be 😆

etorreborre (dismissed):

adrianbenavides:
For some reason, the local bats tests are hanging now in this PR. I'll check it out

etorreborre (approved):

PR #7912, opened 2024-04-16
feat(rust): add the possibility to pass a node configuration inline

etorreborre:
This PR also simplifies the parsing of commands and removes the `CommandsParser ` which IMO is a premature abstraction since:

 - We don't use it as a trait.
 - It requires a parameter for the default node which is only used in some cases.

I also fixed the logic to use the node name defined in the configuration file on relays/inlets/outlets only if a user has not already specified a value for the corresponding `to`, `at` fields. @adrianbenavides please tell me if I understood correctly this piece of logic. What prompted to do this was the fact that, when I tested the inline configuration, I was not ending up with the node name used for the `at` field for an outlet.

Also I started enriching the help text for `ockam node create` but we know that it deserves a lot more.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create/config.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create/config.rs
@@ -6,6 +6,7 @@ use crate::run::parser::Version;
 use crate::value_parsers::async_parse_path_or_url;
 use crate::CommandGlobalOpts;
 use ockam_api::cli_state::journeys::APPLICATION_EVENT_COMMAND_CONFIGURATION_FILE;
+use ockam_api::cli_state::random_name;
 use ockam_node::Context;
 use serde::{Deserialize, Serialize};
 use tracing::{instrument, Span};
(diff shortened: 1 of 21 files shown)


implementations/rust/ockam/ockam_command/src/node/create/config.rs
```diff
@@ -83,9 +96,6 @@ impl NodeConfig {
         }
 
         // Merge the node arguments from the config with the cli command args.
-        if self.node.name.is_none() {
-            self.node.name = Some(ArgValue::String(random_name()));
-        }
```
adrianbenavides, line 88: It took me a while, but I finally remembered the reason behind this. This is the scenario where this was necessary: if you already have a running node and then you want to create a new node given a configuration file that doesn't define a node name, the resources will be attached to the first node (the default node) instead of to the node you are creating.

So, if you have this config file:
```
tcp-outlets:
  to: 18080
```

and you run it twice, the outlets will be attached to the first node because it's the default node. Since the config file is run in the context of the node being created, the config file should be aware of that and create the resources for that node, hence we need to pass an explicit node name to the other commands so they don't fallback to the default node. Does that make sense?

We should definitely add a test for this.

etorreborre: @adrianbenavides thanks for spotting this. Please see my latest commit.

adrianbenavides (dismissed):
Awesome, this is a great improvement! 🚀

adrianbenavides (approved):

PR #7928, opened 2024-04-18
feat(rust): improve output of `node create` command

adrianbenavides:
## Current output

It always returns the same:

image

And when creating a foreground node, it doesn't show anything at all.

## New output

When creating a node on a clean state (no vaults, no identities):

image

When creating a node afterward:

image

Finally, there is also new content when we create a foreground node, in line with what we see in background nodes:

image

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/manager.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/manager.rs
@@ -288,6 +288,10 @@ impl NodeManager {
         transport_options: NodeManagerTransportOptions,
         trust_options: NodeManagerTrustOptions,
     ) -> ockam_core::Result<Self> {
+        let mut cli_state = general_options.cli_state;
+        let node_name = general_options.node_name.clone();
+        cli_state.set_node_name(&node_name);
+
         debug!("create transports");
         let api_transport_id = random_alias();
         let mut transports = BTreeMap::new();
@@ -296,22 +300,14 @@ impl NodeManager {
             transport_options.api_transport_flow_control_id.clone(),
         );
 
-        let mut cli_state = general_options.cli_state;
-        cli_state.set_node_name(general_options.node_name.clone());
-
-        let secure_channels = cli_state
-            .secure_channels(&general_options.node_name)
-            .await?;
+        let secure_channels = cli_state.secure_channels(&node_name).await?;
 
         let registry = Arc::new(Registry::default());
         debug!("start the medic");
         let medic_handle = MedicHandle::start_medic(ctx, registry.clone()).await?;
 
         debug!("retrieve the node identifier");
-        let node_identifier = cli_state
-            .get_node(&general_options.node_name)
-            .await?
-            .identifier();
+        let node_identifier = cli_state.get_node(&node_name).await?.identifier();
 
         debug!("create default resource type policies");
         cli_state
@@ -377,7 +373,7 @@ impl NodeManager {
 
         let mut s = Self {
             cli_state,
-            node_name: general_options.node_name,
+            node_name,
             node_identifier,
             api_transport_flow_control_id: transport_options.api_transport_flow_control_id,
             tcp_transport: transport_options.tcp_transport,
@@ -388,7 +384,7 @@ impl NodeManager {
             medic_handle,
         };
 
-        debug!("retrieve the node identifier");
+        debug!("initializing services");
         s.initialize_services(ctx, general_options.start_default_services)
             .await?;
         info!("created a node manager for the node: {}", s.node_name);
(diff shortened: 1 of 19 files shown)


implementations/rust/ockam/ockam_api/src/nodes/service/manager.rs
```diff
@@ -288,30 +288,27 @@ impl NodeManager {
         transport_options: NodeManagerTransportOptions,
         trust_options: NodeManagerTrustOptions,
     ) -> ockam_core::Result<Self> {
+        let mut cli_state = general_options.cli_state;
+        let node_name = general_options.node_name.clone();
+        cli_state.set_node_name(&node_name);
+
         debug!("create transports");
+        tokio::time::sleep(Duration::from_secs(1)).await;
```
etorreborre: Why do we need a `sleep` here? Doesn't that mean that we are adding a latency of `1 second` to many commands using an `InMemoryNode`?

implementations/rust/ockam/ockam_command/src/command_global_opts.rs
```diff
@@ -139,10 +139,6 @@ impl CommandGlobalOpts {
         cmd: &OckamSubcommand,
         is_tty: bool,
     ) -> miette::Result<LoggingConfiguration> {
-        if global_args.quiet {
```
etorreborre, line 142: I think we should write a good description, and possibly some bats tests for the interactions between `quiet`, `-vvv`, `export OCKAM_LOGGING=true`.

implementations/rust/ockam/ockam_command/src/node/create/foreground.rs
```diff
@@ -46,18 +50,30 @@ impl CreateCommand {
...
             .into_diagnostic()?;
-
-        debug!(
-            "set the node {node_name} listener address to {:?}",
+        debug!(%node_name,
+            "listener address set to {:?}",
             tcp_listener.socket_address()
         );
 
         // Set node_name so that node can isolate its data in the storage from other nodes
```
etorreborre: This is a hack that I'd like to fix at some stage because the `CliState` is not supposed to be node-specific and we could have data races down the line.

implementations/rust/ockam/ockam_api/src/nodes/service/manager.rs
adrianbenavides: Sorry, that's something I forgot to remove. I was testing spinner messages in the nodemanager init function.

implementations/rust/ockam/ockam_command/src/command_global_opts.rs
adrianbenavides: The current [description](https://github.com/build-trust/ockam/blob/33160ea84a6bbf4e0c4d162cb20a5ca99f192a95/implementations/rust/ockam/ockam_command/src/global_args.rs#L21) is pretty accurate in my opinion. The `quiet` flag should only control whether we are allowed to write to stderr.

Would you add something else?

implementations/rust/ockam/ockam_command/src/node/create/foreground.rs
adrianbenavides: Do we have an issue for that? I can open one so we don't forget

implementations/rust/ockam/ockam_command/src/command_global_opts.rs
etorreborre: Actually I would:

 - Enforce and specify the following priorities
    - file configuration variable < env. variable < command line parameter
 - Use some naming conventions:
    - option in configuration file: `an_option`
    - same option as an environment variable:  `OCKAM_AN_OPTION`
    - same option as a command-line option: `an-option`

That means that:

 - We introduce a `--logging` option corresponding to `OCKAM_LOGGING`, set to `false` by default.
 - We introduce a `--confirm` option set to `true` by default.
 - We specify that `--quiet` is a flag meaning `--logging false --confirm false` as a shortcut.
 - We similarly explain that `-vvv` is a flag meaning `--logging true --log-level info` as a shortcut.

implementations/rust/ockam/ockam_command/src/node/create/foreground.rs
etorreborre: I'm giving it a go right now since I'm kind of blocked.

implementations/rust/ockam/ockam_command/src/command_global_opts.rs
adrianbenavides: I like the idea of adding the `logging` and `log-level` flags. But note that the `quiet` has its own meaning: disable interactivity + displaying less output, and it doesn't have anything to do with logging, it's only CLI-related.

etorreborre: Ah, ok. that was not clear to me from the description: `Do not print any log messages` :-).

etorreborre (approved):

PR #7929, opened 2024-04-18
TCP portals over UDP puncture

SanjoDeundiak:
1. Rewrite UDP transport
2. Rewrite UDP puncture
3. Add high-level puncture negotiation protocol
4. Add possibility to use UDP puncture for TCP inlet <-> outlet communication
5. Add possibility to pause TCP inlets
6. Improve transport imports


examples/rust/get_started/examples/04-udp-transport-responder.rs
```diff
@@ -14,11 +14,18 @@ async fn main(ctx: Context) -> Result<()> {
     let udp = node.create_udp_transport().await?;
 
     // Create a UDP listener and wait for incoming datagrams.
-    udp.listen("127.0.0.1:4000").await?;
+    let bind = udp
```
etorreborre, line 17: I'm wondering, maybe we could leave `listen` as an alias for this expression here, in order to make the example short and sweet?

examples/rust/get_started/examples/05-secure-channel-over-two-udp-hops-middle.rs
```diff
@@ -0,0 +1,37 @@
+// This node creates a tcp connection to a node at 127.0.0.1:4000
```
etorreborre: This should say UDP (listener, connection).

examples/rust/get_started/examples/05-secure-channel-over-two-udp-hops-responder.rs
```diff
@@ -0,0 +1,44 @@
+// This node starts a tcp listener, a secure channel listener, and an echoer worker.
```
etorreborre: This should say UDP

implementations/rust/ockam/ockam_api/src/kafka/inlet_controller.rs
```diff
@@ -129,6 +129,8 @@ impl KafkaInletController {
             suffix,
             None,
             true,
+            false,
```
etorreborre, line 132: It could be worth to start exposing some enums instead of booleans when there are long list of parameters.

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
```diff
@@ -54,16 +54,23 @@ pub struct CreateInlet {
     /// The identifier to be used to create the secure channel.
     /// If not set, the node's identifier will be used.
     #[n(10)] pub(crate) secure_channel_identifier: Option<Identifier>,
+    /// Enable UDP NAT puncture.
+    #[n(11)] pub enable_udp_puncture: bool,
```
etorreborre, line 58: Coming back to my comment about exposing enums, this would restrict the number of valid states. For example here `disable_tcp_fallback` is only useful when `enable_udp_puncture` is true.

implementations/rust/ockam/ockam_api/src/nodes/service/manager.rs
```diff
@@ -547,13 +579,22 @@ pub struct ApiTransport {
 pub struct NodeManagerTransportOptions {
     api_transport_flow_control_id: FlowControlId,
     tcp_transport: TcpTransport,
+    enable_udp: bool,
+    udp_transport: UdpTransport,
```
etorreborre: Does it make sense to have a `UdpTransport` if `enable_udp` is `false`? Why not pass an `Option<UdpTransport>` instead?

implementations/rust/ockam/ockam_transport_tcp/src/portal/inlet_listener.rs
```diff
@@ -1,14 +1,20 @@
...
 use ockam_core::{async_trait, compat::boxed::Box};
 use ockam_core::{Address, Processor, Result, Route};
 use ockam_node::Context;
-use ockam_transport_core::TransportError;
+use ockam_transport_core::{HostnamePort, TransportError};
 use tokio::net::TcpListener;
 use tracing::{debug, error, instrument};
 
+#[derive(Debug, Clone)]
+pub struct OutletSharedState {
```
etorreborre: I don't quite understand the "shared" part. "shared" between what and what?

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets.rs
```diff
@@ -345,6 +125,9 @@ impl NodeManager {
         authorized: Option<Identifier>,
         wait_connection: bool,
         secure_channel_identifier: Option<Identifier>,
+        enable_udp_puncture: bool,
+        // TODO: Introduce mode enum
```
etorreborre, line 129: Nice, I can see your comment here.

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets.rs
```diff
@@ -614,10 +404,181 @@ struct InletSessionReplacer {
...
+        Arc<dyn OutgoingAccessControl>,
+    )> {
+        let authority = {
+            if let Some(p) = self.outlet_addr.first() {
+                if let Some(p) = p.cast::<ProjectProto>() {
+                    let projects = self
+                        .node_manager
+                        .cli_state
+                        .projects()
+                        .get_projects_grouped_by_name()
```
etorreborre: It would be nicer to use a `get_project_by_name` function here.

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets.rs
```diff
@@ -614,10 +404,181 @@ struct InletSessionReplacer {
...
+
+impl InletSessionReplacer {
+    async fn access_control(
+        &self,
+    ) -> Result<(
+        Arc<dyn IncomingAccessControl>,
+        Arc<dyn OutgoingAccessControl>,
+    )> {
+        let authority = {
+            if let Some(p) = self.outlet_addr.first() {
```
etorreborre, line 440: I think it would be nice to have a `MultiAddr` method `first_as(code)` to avoid a bit of nesting. There are a few places that would benefit from that.

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets.rs
```diff
@@ -614,10 +404,181 @@ struct InletSessionReplacer {
...
+                            info!("Updating route to UDP: {}", route);
+
+                            // TODO: PUNCTURE monitor this side as well?
+                            //  Also monitored on the UDP puncture level
+                            // TODO: PUNCTURE handle error
+                            let _res = sc_side.update_remote_node_route(route.clone());
+
+                            let _res = inlet.unpause(route![sc_side.clone()]);
+                        // TODO: PUNCTURE handle error
+                        } else if disable_tcp_fallback {
```
etorreborre, line 547: It would be good to refactor the `disable_tcp_fallback` and `!disable_tcp_fallback` branches in the 2 `match` cases.

implementations/rust/ockam/ockam_transport_core/src/resolve_peer.rs
```diff
@@ -0,0 +1,87 @@
...
+        if let Some(p) = iter.find(|x| x.is_ipv6()) {
+            return Ok(p);
+        }
+    }
+
+    // Nothing worked, return an error
+    Err(TransportError::InvalidAddress)?
+}
+
+pub fn parse_socket_addr(s: &str) -> Result<SocketAddr> {
```
etorreborre, line 30: I think that this function could be made private so that client code is forced to use `resolve_peer` all the time. And in that case the `test_parse_socket_address` should be a `test_resolve_peer`.

implementations/rust/ockam/ockam_transport_tcp/src/transport/connection.rs
```diff
@@ -1,10 +1,87 @@
...
+use core::fmt;
+use core::fmt::Formatter;
+use ockam_core::flow_control::FlowControlId;
 use ockam_core::{Address, Result};
+use ockam_node::Context;
+use ockam_transport_core::resolve_peer;
+use std::net::SocketAddr;
 use tracing::debug;
 
+/// Result of [`TcpTransport::connect`] call.
```
etorreborre, line 13: We should come back to this data type at some stage since it seems to accept 2 modes, incoming and outgoing, but in practice we only use outgoing.

implementations/rust/ockam/ockam_transport_tcp/src/transport/portals.rs
```diff
@@ -189,20 +187,37 @@ impl TcpInlet {
...
 
-        let old_route = remote_route.clone();
+        outlet_state.route = Self::build_new_full_route(new_route, &outlet_state.route)?;
 
-        let their_outlet_address = old_route.recipient()?;
-        let new_route = route![new_route, their_outlet_address];
+        Ok(())
+    }
+
+    /// Pause TCP Inlet, all incoming TCP streams will be dropped.
```
etorreborre, line 208: I think it would be good to improve the general documentation of `TcpInlet/TcpOutlet` to specify:

 - The lifecycle: start, stop, pause, unpause.
 - The fact that they accept TCP streams from the "outside" but that the "inside routing" is using Ockam routing, which can mean a number of things: TCP, UDP puncture, secure channels, etc...

implementations/rust/ockam/ockam_transport_udp/src/puncture/negotiation/listener.rs
```diff
@@ -0,0 +1,102 @@
...
+        address: impl Into<Address>,
+        udp: &UdpTransport,
+        rendezvous_route: Route,
+        options: UdpPunctureNegotiationListenerOptions,
+    ) -> Result<()> {
+        let address = address.into();
+
+        let access_control = options.incoming_access_control.clone();
+
+        options.setup_flow_control_for_udp_negotiation_listener(ctx.flow_controls(), &address);
```
etorreborre: Since `options` is already a `UdpPunctureNegotiationListenerOptions` I think we could use a shorter name there.

implementations/rust/ockam/ockam_transport_udp/src/puncture/negotiation/message.rs
```diff
@@ -0,0 +1,29 @@
+use minicbor::{Decode, Encode};
+use ockam_core::{Decodable, Encodable, Message, Result};
+
+#[derive(Encode, Decode, Debug, Clone)]
+#[rustfmt::skip]
+pub(crate) enum UdpPunctureNegotiationMessage {
+    #[n(0)] Initiate {
+        #[n(0)] initiator_udp_public_address: String,
```
etorreborre, line 11: Could this be a proper UDP address instead of a `String`?

implementations/rust/ockam/ockam_transport_udp/src/puncture/negotiation/message.rs
```diff
@@ -0,0 +1,29 @@
...
+        #[n(1)] initiator_remote_address: Vec<u8>,
```
etorreborre, line 12: Could this be an `Address` instead?

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
```diff
@@ -54,16 +54,23 @@ pub struct CreateInlet {
...
+    /// Disable fallback to tcp via relay
```
adrianbenavides: Is this feature explained somewhere else? Perhaps adding a link to the struct that explains it in a deeper layer would be nice, or adding some more context here.

implementations/rust/ockam/ockam_transport_udp/src/puncture/negotiation/options.rs
```diff
@@ -0,0 +1,102 @@
+use ockam_core::flow_control::{FlowControlId, FlowControls};
+use ockam_core::{Address, AllowAll, IncomingAccessControl, OutgoingAccessControl};
+use std::sync::Arc;
+
+/// Trust Options for an Outlet
```
etorreborre: This comment needs to be updated

implementations/rust/ockam/ockam_transport_udp/src/puncture/negotiation/options.rs
```diff
@@ -0,0 +1,102 @@
...
+        mut self,
+        access_control: Arc<dyn OutgoingAccessControl>,
+    ) -> Self {
+        self.outgoing_access_control = access_control;
+        self
+    }
+
+    /// Mark that this UDP Negotiation listener is a Consumer for to the given [`FlowControlId`]
+    /// Also, in this case spawned workers will be marked as Consumers with [`FlowControlId`]
+    /// of the message that was used to create the Outlet
```
etorreborre: "to create the NegotiationWorker"?

implementations/rust/ockam/ockam_transport_udp/src/puncture/rendezvous_service/messages.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+impl Message for RendezvousRequest {}
+
+/// Response type for UDP Puncture Rendezvous service
+#[derive(Encode, Decode, Debug)]
+#[rustfmt::skip]
+pub enum RendezvousResponse {
+    #[n(0)] Pong,
+    #[n(1)] GetMyAddress(#[n(0)] String),
```
etorreborre, line 33: Why not use a proper `Address` here? Is it because we want to eventually remove the transport type from the `Address`?

implementations/rust/ockam/ockam_transport_udp/src/puncture/rendezvous_service/rendezvous.rs
```diff
@@ -0,0 +1,208 @@
...
+    use super::RendezvousServiceWorker;
+    use crate::puncture::rendezvous_service::{RendezvousRequest, RendezvousResponse};
+    use crate::{RendezvousService, UdpBind, UdpBindArguments, UdpBindOptions, UdpTransport, UDP};
+    use ockam_core::{route, Result, Route, TransportType};
+    use ockam_node::Context;
+
+    #[test]
+    fn parse_route() {
+        assert_eq!(
+            route![(UDP, "c")],
```
etorreborre: That's a minor thing, but I think we should stick to `assert_eq!(actual, expected)` like [most of the rust docs](https://doc.rust-lang.org/std/vec/struct.Vec.html#examples-16), even if it's not officially mandated.

implementations/rust/ockam/ockam_transport_udp/src/puncture/negotiation/message.rs
```diff
@@ -0,0 +1,29 @@
+use minicbor::{Decode, Encode};
+use ockam_core::{Decodable, Encodable, Message, Result};
+
+#[derive(Encode, Decode, Debug, Clone)]
+#[rustfmt::skip]
+pub(crate) enum UdpPunctureNegotiationMessage {
```
adrianbenavides, line 6: It would be helpful to have a brief description of the meaning of these messages / how are they used conceptually.

implementations/rust/ockam/ockam_transport_udp/src/puncture/negotiation/negotiation.rs
```diff
@@ -0,0 +1,57 @@
+use crate::puncture::negotiation::worker::UdpPunctureNegotiationWorker;
+use crate::puncture::rendezvous_service::RendezvousClient;
+use crate::{UdpBindArguments, UdpBindOptions, UdpTransport};
+use ockam_core::{Address, Result, Route};
+use ockam_node::Context;
+use tokio::sync::broadcast;
+use tokio::sync::broadcast::{Receiver, Sender};
+
+/// Allows to negotiate a UDP puncture to the other node by communicating
+/// with its [`UdpPunctureListener`] via some side channel (e.g. Relayed connection though the
```
adrianbenavides: typo: `though`/`through`

implementations/rust/ockam/ockam_transport_udp/src/workers/receiver.rs
```diff
@@ -69,17 +65,36 @@ impl Processor for UdpListenProcessor {
...
-            self.sender_addr.clone(),
-            Address::new(UDP, addr.to_string()),
-            msg.return_route(),
-        ];
-        msg = msg.set_return_route(new_route);
+        if msg.onward_route_ref().is_empty() {
+            return Ok(true);
+        }
+
+        let return_route = RouteBuilder::default().append(self.addresses.sender_address().clone());
```
etorreborre, line 72: `route![self.addresses.sender_address().clone()]` should work there.

implementations/rust/ockam/ockam_transport_udp/src/puncture/negotiation/worker.rs
```diff
@@ -0,0 +1,259 @@
...
+            my_udp_public_address
+        );
+
+        self.my_udp_public_address = Some(my_udp_public_address.clone());
+
+        if self.is_initiator {
+            let onward_route = self
+                .onward_route
+                .as_ref()
+                .ok_or(PunctureError::Internal)?
```
adrianbenavides, line 105: We could have a more specific error in places like this, like`PunctureError::InitiatorBadUsage` in this case. I can see similar cases in this file that could have more specific errors.

etorreborre (dismissed):
I mostly made cosmetic comments. Overall I think it would be helpful to insert somewhere a diagram showing:

 - The various workers involved and the addresses / routes they manage.
 - A sequence for the protocol showing how and where state changes based on the exchanged messages.

examples/rust/get_started/examples/04-udp-transport-responder.rs
SanjoDeundiak: I think tcp listening and udp bind has different semantics, so we should not reuse the word. I would even say that udp bind is closer to tcp accept, rather than listen

implementations/rust/ockam/ockam_api/src/kafka/inlet_controller.rs
SanjoDeundiak: True

implementations/rust/ockam/ockam_api/src/nodes/service/manager.rs
SanjoDeundiak: It doesn't, but then I would need to handle transport being None everywhere it's used, which I would like to avoid. But I guess I need to do it

implementations/rust/ockam/ockam_transport_tcp/src/portal/inlet_listener.rs
SanjoDeundiak: Between `TcpInletListenProcessor` and `TcpInlet`. Maybe there is a confusion with naming because it's not really Outlet's state. It was information inlet knows about the outlet, before I added `is_paused` it did make some sense. Now I think it should be `InletSharedState`?

implementations/rust/ockam/ockam_transport_udp/src/workers/receiver.rs
SanjoDeundiak: Then you would need to convert it to `RouteBuilder`, which you can't because it can't reference a temporal value, so you'll need to create a new variable. I think it's better this way

implementations/rust/ockam/ockam_transport_udp/src/puncture/rendezvous_service/messages.rs
SanjoDeundiak: I don't think that's the type it is. `Address` is an Ockam Routing address. This one is a UDP address

SanjoDeundiak: It's actually `SocketAddr`. So, we could maybe wrap it inside something `Encode`-able as an improvement and use here

implementations/rust/ockam/ockam_transport_udp/src/puncture/negotiation/message.rs
SanjoDeundiak: It's always a local address, so I don't think using `Address` type is appropriate. So either we introduce a new type, of just use `Vec<u8>`

implementations/rust/ockam/ockam_transport_udp/src/puncture/negotiation/message.rs
SanjoDeundiak: That's a `SocketAddr`

implementations/rust/ockam/ockam_transport_udp/src/puncture/negotiation/listener.rs
SanjoDeundiak: A little bit shorter yes, but there also will be a function that sets up flow control for the puncture, on the same struct, so need to distinguish them by name

implementations/rust/ockam/ockam_transport_tcp/src/transport/portals.rs
SanjoDeundiak: Yes, and I had something similar when I initially implemented them, but it's probably very outdated. I feel like almost every feature needs it honestly...

implementations/rust/ockam/ockam_transport_tcp/src/transport/connection.rs
SanjoDeundiak: True, there is no way to get it for the incoming connection at the moment. But we don't need it yet

examples/rust/get_started/examples/04-udp-transport-responder.rs
etorreborre: You're right, that makes sense. Still, would there be a word that we could use as a default when no arguments/options are specified?

implementations/rust/ockam/ockam_transport_core/src/resolve_peer.rs
SanjoDeundiak: Hm, do you think calling `parse_socker_addr` doesn't make sense? For example when we create an inlet, should we accept a peer or a socket addr?

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets.rs
SanjoDeundiak: Could you please elaborate?

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets.rs
SanjoDeundiak: I quickly tried to do it, but it's tricky due to borrowing and lifetimes mess

examples/rust/get_started/examples/04-udp-transport-responder.rs
SanjoDeundiak: We can implement conversion `&str` => `UDPBindArguments` to simplify the example. Why would we want to rename the function for that?

adrianbenavides (dismissed):
Haven't finished yet, I still have to check the UDP transport files (the similar/common transport stuff), will do that later today, but feel free to move the PR forward.

implementations/rust/ockam/ockam_transport_udp/src/puncture/negotiation/worker.rs
SanjoDeundiak: I'm leaving `Internal` for errors that should never happen (to avoid using panic). I'll make other errors more detailed

implementations/rust/ockam/ockam_transport_udp/src/puncture/negotiation/message.rs
SanjoDeundiak: Fixed

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
SanjoDeundiak: Fixed

implementations/rust/ockam/ockam_api/src/nodes/service/manager.rs
SanjoDeundiak: Fixed

etorreborre (approved):

PR #7930, opened 2024-04-18
feat(rust): scope some repositories to a given node name

etorreborre:
We realized some time ago that some data needed to be segregated by node name in the database:

 - Resources (type, policies).
 - Identity attributes.
 - Credentials.

We fixed this by using mutation and allowing a database to temporarily have a `node_name` which could be used by the repositories needing it (the ones storing the data mentioned above).

This is problematic because:

 - This introduces some temporal coupling: we need to think of setting the right node name before invoking any operation that could potential touch resources, identity attributes or credentials.
 - This creates more potential states in the system: now _any_ repository can potentially use a node name to store data, while in reality only a few of them need this. 

This PR addresses this issue by:

 - Adding a `node_name` directly on the `XXXSqlDatabase` implementations needing it. For example in `CredentialSqlxDatabase` because credentials are segregated by node name in the database.
 - Changing the various builder functions so that a `node_name` is passed as a parameter as necessary. Typically, in the `NodeManager` we create storage that is scoped to a specific node.

This is still not great:

 - The fact that we share some storage, for example for the list of all identities, and that we segregate some, for example credentials, is the result of building a command line application and a `NodeManager` on top of the functions provided by  the `ockam_identities` crates. So, the constraints introduced at a "high" level in the architecture have some consequences on the "low" level of the architecture.
 
 - Concretely speaking, the `Identities` struct is mostly node agnostic but not entirely. When we use it to create an identity, that identity is accessible by any node. On the other hand if we use it to store a credential, it's node specific. One thing we could potentially do is revisit this choice and make every repository under `Identities` node specific. But this would be a much larger refactoring.

### DIFF
--- a/implementations/rust/ockam/ockam_abac/src/policy/resources.rs
+++ b/implementations/rust/ockam/ockam_abac/src/policy/resources.rs
@@ -1,0 +1,28 @@
+use crate::{Resource, ResourceName, ResourcesRepository};
+use ockam_core::compat::sync::Arc;
+use ockam_core::Result;
+
+#[derive(Clone)]
+pub struct Resources {
+    resources_repository: Arc<dyn ResourcesRepository>,
+}
+
+impl Resources {
+    pub fn new(resources_repository: Arc<dyn ResourcesRepository>) -> Self {
+        Self {
+            resources_repository,
+        }
+    }
+
+    pub async fn store_resource(&self, resource: &Resource) -> Result<()> {
+        self.resources_repository.store_resource(resource).await?;
+        Ok(())
+    }
+
+    pub async fn delete_resource(&self, resource_name: &ResourceName) -> Result<()> {
+        self.resources_repository
+            .delete_resource(resource_name)
+            .await?;
+        Ok(())
+    }
+}
--- a/implementations/rust/ockam/ockam_abac/src/policy/storage/resource_policy_repository_sql.rs
+++ b/implementations/rust/ockam/ockam_abac/src/policy/storage/resource_policy_repository_sql.rs
@@ -12,19 +12,24 @@ use crate::{Action, Expr, ResourceName, ResourcePoliciesRepository, ResourcePoli
 #[derive(Clone)]
 pub struct ResourcePolicySqlxDatabase {
     database: SqlxDatabase,
+    node_name: String,
 }
 
 impl ResourcePolicySqlxDatabase {
     /// Create a new database for resource policies
-    pub fn new(database: SqlxDatabase) -> Self {
+    pub fn new(database: SqlxDatabase, node_name: &str) -> Self {
         debug!("create a repository for resource policies");
-        Self { database }
+        Self {
+            database,
+            node_name: node_name.to_string(),
+        }
     }
 
     /// Create a new in-memory database for policies
     pub async fn create() -> Result<Self> {
         Ok(Self::new(
             SqlxDatabase::in_memory("resource_policies").await?,
+            "default",
         ))
     }
 }
@@ -44,7 +49,7 @@ impl ResourcePoliciesRepository for ResourcePolicySqlxDatabase {
         .bind(resource_name.to_sql())
         .bind(action.to_sql())
         .bind(expression.to_string().to_sql())
-        .bind(self.database.node_name()?.to_sql());
+        .bind(self.node_name.to_sql());
         query.execute(&*self.database.pool).await.void()
     }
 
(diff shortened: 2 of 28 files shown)


implementations/rust/ockam/ockam_command/src/credential/list.rs
```diff
@@ -42,11 +42,9 @@ impl ListCommand {
             Some(name) => name,
             None => opts.state.get_default_node().await?.name(),
         };
-        let mut opts = opts.clone();
-        opts.state.set_node_name(&node_name);
-
+        let opts = opts.clone();
```
SanjoDeundiak: Is clone still required?

implementations/rust/ockam/ockam_identity/src/identities/identities.rs
```diff
@@ -157,14 +157,23 @@ impl Identities {
     /// Return a builder for identities with a specific database
     #[cfg(feature = "storage")]
     pub fn create(database: SqlxDatabase) -> IdentitiesBuilder {
+        Self::create_with_node(database, "default")
```
SanjoDeundiak, line 160: I would say this function can be error-prone, it looks like node name value is not required, while it can be very important. Can we delete the function?

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_policy_repository_sql.rs
```diff
@@ -12,19 +12,24 @@ use crate::{Action, Expr, ResourceName, ResourcePoliciesRepository, ResourcePoli
...
+            database,
+            node_name: node_name.to_string(),
+        }
     }
 
     /// Create a new in-memory database for policies
     pub async fn create() -> Result<Self> {
         Ok(Self::new(
             SqlxDatabase::in_memory("resource_policies").await?,
+            "default",
```
SanjoDeundiak, line 32: Same here

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_type_policy_repository_sql.rs
```diff
@@ -13,19 +13,24 @@ use crate::{Action, Expr, ResourceType, ResourceTypePoliciesRepository};
...
+            database,
+            node_name: node_name.to_string(),
+        }
     }
 
     /// Create a new in-memory database for policies
     pub async fn create() -> Result<Self> {
         Ok(Self::new(
             SqlxDatabase::in_memory("resource_type_policies").await?,
+            "default",
```
SanjoDeundiak, line 33: +1

implementations/rust/ockam/ockam_abac/src/policy/resources.rs
```diff
@@ -0,0 +1,28 @@
+use crate::{Resource, ResourceName, ResourcesRepository};
+use ockam_core::compat::sync::Arc;
+use ockam_core::Result;
+
+#[derive(Clone)]
+pub struct Resources {
```
SanjoDeundiak, line 6: I mean, that abstraction layer itself does make total sense, but is it useful enough to exist? It's not doing anything that the repository can't do directly

etorreborre: I know, I want to keep the consistency we currently have where no repositories are directly accessible from the `CliState`, so that we can reason about the `CliState` uniformly as offering a list of services.

implementations/rust/ockam/ockam_command/src/credential/list.rs
etorreborre: I removed it.

implementations/rust/ockam/ockam_identity/src/identities/identities.rs
etorreborre: Unfortunately, this can not be removed without the larger refactoring that I mentioned it the "This is still not great" section. Another way to see this is that we currently lack a good dependency injection solution. This would be trivial if we had it because we could assemble exactly the components we need.

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_policy_repository_sql.rs
etorreborre: This is in memory so less critical.

implementations/rust/ockam/ockam_abac/src/policy/storage/resource_type_policy_repository_sql.rs
etorreborre: This is in memory so less critical.

etorreborre:
@SanjoDeundiak thanks for the review, I rebased and addressed your comments.

SanjoDeundiak (approved):

PR #7935, opened 2024-04-19
chore(rust): improve help text for `node create` arguments

adrianbenavides:


implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -48,11 +48,11 @@ long_about = docs::about(LONG_ABOUT),
...
 pub struct EnrollCommand {
-    /// The name of an existing Ockam Identity that you wish to enroll. You can use `ockam
-    /// identity list` to get a list of existing Identities. To create a new Identity, use
-    /// `ockam identity create`. If you don't specify an Identity, and you don't have a
-    /// default Identity, this command will create a default Identity for you and save it
-    /// locally in a default Vault
+    /// The name of an existing Ockam Identity that you wish to enroll.
+    /// You can use `ockam identity list` to get a list of existing Identities.
+    /// To create a new Identity, use `ockam identity create`.
+    /// If you don't specify an Identity, and you don't have a default Identity, this command
```
etorreborre: I would say "If you don't specify an Identity name"

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -65,36 +64,26 @@ pub struct CreateCommand {
...
     /// seconds to complete.
-    #[arg(long, hide = true, value_parser = parse_launch_config)]
+    #[arg(hide = true, long, value_parser = parse_launch_config)]
     pub launch_config: Option<Config>,
 
-    /// Name of the Identity that the node will use
+    /// The name of an existing Ockam Identity that this node will use.
+    /// You can use `ockam identity list` to get a list of existing Identities.
+    /// To create a new Identity, use `ockam identity create`.
+    /// If you don't specify an Identity, and you don't have a default Identity, this command
```
etorreborre: Same thing here "specify Identity name".

implementations/rust/ockam/ockam_command/src/node/create/config.rs
```diff
@@ -3,14 +3,34 @@ use crate::run::parser::building_blocks::ArgValue;
...
+
+    /// A path, URL or inlined hex-encoded enrollment ticket to use for the Ockam Identity associated to this node.
+    /// When passed, the identity will be given a project membership credential.
+    /// Check the `project ticket` command for more information about enrollment tickets.
+    #[arg(long, value_name = "ENROLLMENT TICKET", value_parser = parse_enrollment_ticket)]
+    pub enrollment_ticket: Option<EnrollmentTicket>,
+
+    /// Key-value pairs defining environment variables used in the Node configuration.
+    /// This variable can be used multiple times, each time adding a new key-value pair.
+    /// Example: `--variable KEY1=VALUE1 --variable KEY2=VALUE2`
```
etorreborre, line 31: Would that be too much to also mention that repeating an existing key will override the previous one (instead of the opposite for example). Or should we raise an error because it is likely to be a configuration issue?

implementations/rust/ockam/ockam_command/src/node/create/foreground.rs
```diff
@@ -22,9 +23,24 @@ use ockam_core::{route, LOCAL};
...
+    /// Run the node in foreground mode. This will block the current process until the node receives
+    /// an exit signal (e.g., SIGINT, SIGTERM, CTRL+C, EOF).
+    #[arg(long, short)]
+    pub foreground: bool,
+
+    /// When running a node in foreground mode, exit the process when receiving EOF on stdin.
+    #[arg(long, short, requires = "foreground")]
+    pub exit_on_eof: bool,
+
+    /// A flag to determine whether the current foreground node was started as a child process.
```
etorreborre, line 39: Maybe we could add that this flag exists for implementation purposes, even if it's hidden?

etorreborre (dismissed):

implementations/rust/ockam/ockam_command/src/node/create/config.rs
adrianbenavides: I think we should raise an error. I can't see how passing the same variable twice with different values can be intended. I'll add that.

adrianbenavides: I've added this error

image

and also an extra line to the `variable` argument

https://github.com/build-trust/ockam/blob/3531ce3fed29f947c910d6022810b7a809112191/implementations/rust/ockam/ockam_command/src/node/create/config.rs#L28

etorreborre (approved):

PR #7936, opened 2024-04-19
feat(rust): create a tcp inlet for exporting traces when a project exists

etorreborre:
This PR supports the export of OpenTelemetry traces via a portal when a project exists.

In order to do this, we:

 - Create or restart an `ockam-opentelemetry-inlet` node 
 - Create an inlet named `ockam-opentelemetry` connecting to a relay named `ockam-opentelemetry`

Another PR sets-up the other part of the `ockam-opentelemetry` portal when a new project is created.

I also introduced a few environment variables:

 - `OCKAM_OPENTELEMETRY_EXPORT_VIA_PORTAL` to enable this feature (`false` by default for now)
 - `OCKAM_OPENTELEMETRY_EXPORT_DEBUG` to debug the creation of the logger / tracer configurations (since a logger is not available at this stage)
 - `OCKAM_FOREGROUND_LOG_EXPORT_PORTAL_CUTOFF` and `OCKAM_FOREGROUND_SPAN_EXPORT_PORTAL_CUTOFF` to stop waiting for a response from the other side of a portal. If we don't do this we add too much latency to commands because we need to wait for a full portal round-trip in order to return from the command execution.

## Implementation notes

- The inlets and outlets passed in the node configuration are persisted so that they can be recreated when the node is restarted.
- The name of the data persisted for a `TcpOutlet` is not great "OutletStatus" and we should possibly identify an outlet with a better name than its worker address. But I didn't try to address this question in the PR.
- I added some `CliState` functions to manage inlets and outlets as a "business layer" on top of the pure persistence layer.

### DIFF
--- a/tools/docker/healthcheck/Dockerfile
+++ b/tools/docker/healthcheck/Dockerfile
@@ -1,5 +1,5 @@
 # Stage 1 - Build elixir release of ockam_healthcheck elixir app
-FROM ghcr.io/build-trust/ockam-elixir-builder@sha256:65bf57734875c150e7dd94cd3b4df256abfab1c1509c9d6ede15b791c2b76cd9 AS elixir-app-release-build
+FROM ghcr.io/build-trust/ockam-elixir-builder@sha256:134fc5b1163e34d66628dc8e0b4e8ce80213d0b82303c71c2a50b916df213aaf AS elixir-app-release-build
 
 COPY --from=cgr.dev/chainguard/wolfi-base /bin /bin
 COPY --from=cgr.dev/chainguard/wolfi-base /usr/bin /usr/bin
@@ -21,7 +21,7 @@ RUN set -ex; \
 
 
 # Stage 2 - Create container and copy executables in above step
-FROM ghcr.io/build-trust/ockam-elixir-base@sha256:50da5804c0aef8f98b8500f4fcbe877b95826d587d8ef0f474a8895effe6140b AS app
+FROM ghcr.io/build-trust/ockam-elixir-base@sha256:6e741652317065de0234e12d4a95563d1f6dbdd6bb95cce63e53780361f64a62 AS app
 
 COPY --from=cgr.dev/chainguard/wolfi-base /bin /bin
 COPY --from=cgr.dev/chainguard/wolfi-base /usr/bin /usr/bin


implementations/rust/ockam/ockam_api/src/cli_state/storage/tcp_portals_repository.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    /// Return the configuration of a TcpInlet for a given node name and worker address
+    async fn get_tcp_outlet(
+        &self,
+        node_name: &str,
+        worker_addr: &Address,
+    ) -> Result<Option<OutletStatus>>;
+
+    /// Delete the configuration of a TcpInlet for a given node name and worker address
+    async fn delete_tcp_outlet(&self, node_name: &str, worker_addr: &Address) -> Result<()>;
```
adrianbenavides: A couple of typos in the doc strings: `TcpInlet` -> `TcpOutlet`

implementations/rust/ockam/ockam_api/src/logs/exporting_configuration.rs
```diff
@@ -131,48 +172,120 @@ impl Display for ExportingConfiguration {
...
-        } else {
-            if !quiet {
-                eprintln!("Exporting OpenTelemetry events is disabled because the OpenTelemetry collector endpoint at {} cannot be reached after {}ms", endpoint, connection_check_timeout.as_millis());
-                eprintln!("You can disable the export of OpenTelemetry events with: `export OCKAM_OPENTELEMETRY_EXPORT=false` to avoid this connection check.");
-            }
-            Ok(ExportingEnabled::Off)
-        }
+        let endpoint_kind = match endpoint {
+            OpenTelemetryEndpoint::HttpsEndpoint(_) => "OpenTelemetry collector endpoint",
+            OpenTelemetryEndpoint::PortalEndpoint(_) => "opentelemetry inlet",
```
adrianbenavides: `opentelemetry` -> `OpenTelemetry`?

implementations/rust/ockam/ockam_api/src/logs/exporting_configuration.rs
```diff
@@ -131,48 +172,120 @@ impl Display for ExportingConfiguration {
...
 }
 
 /// Return true if the endpoint can be accessed with a TCP connection
-fn is_endpoint_accessible(url: Url, connection_check_timeout: Duration) -> bool {
+fn is_endpoint_accessible(url: &Url, connection_check_timeout: Duration) -> bool {
     match to_socket_addr(url) {
         Some(address) => {
-            std::net::TcpStream::connect_timeout(&address, connection_check_timeout).is_ok()
+            let retries = FibonacciBackoff::from_millis(100);
+            let mut total_time = Duration::from_secs(0);
```
adrianbenavides: You could use `let now = Instant::now()` and `now.elapsed()` instead?

implementations/rust/ockam/ockam_api/src/logs/exporting_configuration.rs
```diff
@@ -131,48 +172,120 @@ impl Display for ExportingConfiguration {
...
+                ));
+                if std::net::TcpStream::connect_timeout(&address, timeout_duration).is_ok() {
+                    return true;
+                } else {
+                    if total_time >= connection_check_timeout {
+                        return false;
+                    };
+                    let _ = Executor::execute_future(async move {
+                        tokio::time::sleep(timeout_duration).await
+                    });
```
adrianbenavides: If this function is called from a blocking context, we could use a blocking sleep right? It's always used through this function [here](https://github.com/build-trust/ockam/blob/33160ea84a6bbf4e0c4d162cb20a5ca99f192a95/implementations/rust/ockam/ockam_command/src/command.rs#L69).

implementations/rust/ockam/ockam_api/src/logs/exporting_configuration.rs
etorreborre: Much better, thanks!

implementations/rust/ockam/ockam_api/src/logs/exporting_configuration.rs
etorreborre: Good point.

PR #7938, opened 2024-04-19
feat(rust): added abac rules to kafka inlet and oulet

davide-baldo:
This PR adds supports to ABAC policies for Kafka Inlet and Outlet. Other commands, such as consumer, producer, and direct, were explicitly ignored with the intention of deprecating them in the near future.

Also, TLS and generic hostname support were added to the `kafka-outlet` command.

The relative example is available in this PR: https://github.com/build-trust/ockam/pull/7944


implementations/rust/ockam/ockam_abac/src/policy/manual.rs
```diff
@@ -0,0 +1,47 @@
+use crate::PolicyAccessControl;
+use core::fmt::{Debug, Formatter};
+use ockam_core::Result;
+use ockam_identity::Identifier;
+use tracing::debug;
+
+pub struct ManualPolicyAccessControl {
```
etorreborre: It would be good to add a comment here to explain what that is and why it is encapsulating a `PolicyAccessControl`.

Actually I don't see why `is_identity_authorized` cannot be added directly to `PolicyAccessControl` to provide more encapsulation. Is there a particular reason?

davide-baldo: To keep the APIs consistent: `PolicyAccessControl` is a factory, I didn't want to mix it with logic.
`ManualPolicyAccessControl` is symmetric to `IncomingPolicyAccessControl` and `OutgoingPolicyAccessControl`.

SanjoDeundiak (approved):

adrianbenavides (approved):

PR #7946, opened 2024-04-22
feat(rust): add the possibility to start a node in foreground mode with a configuration

etorreborre:
This PR allows the `ockam node create` command to take a configuration and run the node in the foreground (with the `-f` or `--foreground` flag). 

Since the execution of the command creating and starting a node eventually blocks, we run all the other commands in a separate thread. Those commands are retried because at the time when they execute the node not be up yet.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create/config.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create/config.rs
@@ -6,12 +6,17 @@ use crate::run::parser::Version;
 use crate::value_parsers::{async_parse_path_or_url, parse_enrollment_ticket, parse_key_val};
 use crate::CommandGlobalOpts;
 use clap::Args;
-use miette::miette;
+use colorful::Colorful;
+use miette::{miette, IntoDiagnostic};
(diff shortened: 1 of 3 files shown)


adrianbenavides:
How do you plan to use this exactly?  I think it's a bit misleading having to define the `foreground` and `child-process` fields in the config file.

Currently, we can do this to start a node with a config file and block the process after the node is created:

```
 ockam node create -f ./config.yaml
```

etorreborre:
This is used to start the `ockam-opentelemetry` node that is next to the authority node. That node is started in a pod and needs an inline configuration file because we can't create and store one in the deployment file.

etorreborre:
That being said I agree that it is weird to have those 2 fields in there. When I had a look at it, I couldn't find a way to not do this that would not require a serious refactoring. Do you see how to do it?

adrianbenavides:
> That node is started in a pod and needs an inline configuration file because we can't create and store one in the deployment file.

The `node create` can also read the contents from a URL. Could we use that in this environment? If not, we could pass a json equivalent config file and pass more complex configurations as a single line. This shouldn't require too many changes I think.

adrianbenavides (approved):

PR #7947, opened 2024-04-22
feat(rust): when deleting a node, wait for node's process to finish

adrianbenavides:
When deleting a node, wait until the process is fully terminated. Not doing so can lead to a variety of errors related to node resources not being freed up (e.g. ports) before trying to recreate the same node or a new one.

If the node takes more than 0.5s to stop, a progress message is shown:

```
➜ ockam node delete -y                                                
     ⠹ Waiting for node sound-whippet to stop                                                                                                                           ✔ 
     ✔ Node with name sound-whippet has been deleted
```

It normally takes less than that. In that case, the progress message is not shown.


etorreborre (approved):
Excellent.

PR #7955, opened 2024-04-23
feat(rust): improve output of `node show` command

adrianbenavides:
In the following images, we can compare the output of the same node in the current/new formats:

Current: 

image

Proposed:

image


etorreborre (approved):
That's much better in a compact and colorized form indeed

PR #7956, opened 2024-04-23
kafka instaclustr example

snandam:
This example uses instaclustr managed kafka cluster to demonstrate portal kafka tcp-inlet and tcp-outlet

### DIFF
--- a/examples/command/portals/kafka/instaclustr/docker/cluster_manager.sh
+++ b/examples/command/portals/kafka/instaclustr/docker/cluster_manager.sh
@@ -1,0 +1,172 @@
+#!/bin/bash
+
+# Function to create a Kafka cluster
+create_cluster() {
+  local response
+  local create_cluster_url="https://api.instaclustr.com/cluster-management/v2/resources/applications/kafka/clusters/v2"
+  response=$(curl -s -w "%{http_code}" -X POST -u "${INSTACLUSTR_USER_NAME}:${INSTACLUSTR_API_KEY}" "${create_cluster_url}" \
+    -H 'Content-Type: application/json' \
+    -d '{
+      "allowDeleteTopics": true,
+      "autoCreateTopics": true,
+      "clientToClusterEncryption": false,
+      "dataCentres": [
+        {
+          "cloudProvider": "AWS_VPC",
+          "name": "AWS_VPC_US_EAST_1",
+          "network": "10.0.0.0/16",
+          "nodeSize": "KFK-DEV-t4g.small-5",
+          "numberOfNodes": 3,
+          "region": "US_EAST_1"
+        }
+      ],
+      "dedicatedZookeeper": [],
+      "defaultNumberOfPartitions": 3,
+      "defaultReplicationFactor": 3,
+      "kafkaVersion": "3.6.1",
+      "name": "DemoKafkaCluster",
+      "pciComplianceMode": false,
+      "privateNetworkCluster": false,
+      "slaTier": "NON_PRODUCTION"
+    }')
+  echo "${response}"
+}
+
+# Function to wait for cluster to be fully operational
+wait_for_cluster() {
+  local cluster_id=$1
+  local status
+  local response
+
+  while true; do
+    local url="https://api.instaclustr.com/cluster-management/v2/resources/applications/kafka/clusters/v2/${cluster_id}"
+    response=$(curl -s -X GET -H "Content-Type: application/json" -u "${INSTACLUSTR_USER_NAME}:${INSTACLUSTR_API_KEY}" "${url}")
+    status=$(echo "${response}" | jq -r '.status')
+
+    case "${status}" in
+      RUNNING)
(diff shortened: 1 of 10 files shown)


examples/command/portals/kafka/instaclustr/application_team/producer.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
snandam: Having the commands inline inside `docker-compose.yml` is causing authentication errors.

examples/command/portals/kafka/instaclustr/application_team/consumer.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
```
snandam: Having the commands inline inside `docker-compose.yml` is causing authentication errors.

examples/command/portals/kafka/instaclustr/docker/run.sh
```diff
@@ -0,0 +1,116 @@
...
+}
+# Check if Ockam Command is already installed and available in path.
+# If it's not, then install it.
+if ! type ockam &>/dev/null; then
+    curl --proto '=https' --tlsv1.2 -sSfL https://install.command.ockam.io | bash
+    source "$HOME/.ockam/env"
+fi
+
+# Check that tools we we need installed.
+for c in docker curl jq; do
```
mrinalwadhwa, line 114: The likelihood that someone will have JQ is too low.

mrinalwadhwa: If we can't avoid using it - lets use it though docker since we already have that dependency.

snandam: Using jq to extract the necessary details from the API responses in [cluster_manager.sh](https://github.com/build-trust/ockam/pull/7956/files#diff-128852991c06403daaf1d4902ab9214252f8d8722c69422643c96dce8f6c6fb0) after a cluster is created in instaclustr. This script runs at the beginning of the run and no docker image is built at that time. Let me think about this.

examples/command/portals/kafka/instaclustr/docker/run.sh
```diff
@@ -0,0 +1,116 @@
...
+# Cleanup after the example - `./run.sh cleanup`
+# Remove all containers and images pulled or created by docker compose.
+cleanup() {
+    instaclustr_auth
+    ./cluster_manager.sh cleanup
+    pushd instaclustr_operator; docker compose down --rmi all --remove-orphans; popd
+    pushd application_team; docker compose down --rmi all --remove-orphans; popd
+}
+
+instaclustr_auth() {
```
adrianbenavides, line 88: It would be nice to skip this function if the user already has the `INSTACLUSTR_USER_NAME` and `INSTACLUSTR_API_KEY` vars.

examples/command/portals/kafka/instaclustr/docker/cluster_manager.sh
```diff
@@ -0,0 +1,160 @@
...
+    unset INSTACLUSTR_API_KEY
+  else
+    local response=$(create_cluster)
+    local cluster_id=$(echo "${response}" | jq -r '.id')
+    echo "Creating a trial cluster. Cluster setup usually take around 5-10 minutes"
+    echo "Cluster ID: ${cluster_id}"
+    echo "${cluster_id}" > cluster_id.txt
+    wait_for_cluster "${cluster_id}"
+    create_user "${cluster_id}"
+    add_firewall_rule "${cluster_id}"
```
adrianbenavides, line 163: I accidentally used a wrong API key and this went through without failing  with the following output:

```bash
Cluster ID: 
Current status: . Waiting for cluster to be running...
```

Perhaps we could check if the cluster id is empty before continuing.

examples/command/portals/kafka/instaclustr/docker/run.sh
snandam: Nice suggestion. Updated code to check for ENV variables. Workflow is better when it's set before running `run.sh`. Also makes `run.sh cleanup` smoother where user doesn't have to enter the creds twice

examples/command/portals/kafka/instaclustr/docker/cluster_manager.sh
snandam: Good catch. Im capturing the response_code along with the response now. In cases when username/api key is invalid, the API response is empty. So had to do more parsing to get the response code separately. Could you give it a test again please?

examples/command/portals/kafka/instaclustr/docker/run.sh
snandam: As suggested, will work on removing `jq` dependency as v2 of this example

adrianbenavides (approved):

etorreborre (approved):

PR #7961, opened 2024-04-25
feat(rust): improve output of `project enroll` command

adrianbenavides:
## Current output

It shows cryptic information that the user won't understand/use. Also the format is not aligned with the rest of the commands. 

image

## Proposed output

In this case, it shows less information:
- The identity name instead of the identifier, because it's what we use as the input in the commands
- The creation and expiration dates for the credential
- The attributes associated with the credential

image

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/project/enroll.rs
+++ b/implementations/rust/ockam/ockam_command/src/project/enroll.rs
@@ -1,11 +1,13 @@
+use std::fmt::Write;
 use std::sync::Arc;
 
 use async_trait::async_trait;
 use clap::Args;
 use colorful::Colorful;
-use miette::miette;
 use miette::Context as _;
+use miette::{miette, IntoDiagnostic};
 
+use ockam::identity::models::CredentialData;
 use ockam::Context;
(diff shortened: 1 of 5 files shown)


implementations/rust/ockam/ockam_command/src/project/enroll.rs
```diff
@@ -191,6 +183,89 @@ impl EnrollCommand {
...
+                        fmt::INDENTATION,
+                        color_primary(format!("\"{k}={v}\""))
+                    )
+                )?;
+            }
+        }
+        Ok(buf)
+    }
+
+    fn extract_credential_data(
```
etorreborre: This could be moved to the `CredentialAndPurposeKey` struct in order to be reused in other places.

etorreborre (dismissed):
This looks a lot better indeed.

implementations/rust/ockam/ockam_command/src/project/enroll.rs
adrianbenavides: We actually had a method to extract just the needed data (credential data only without the purpose key data). I've removed the unnecessary code.

etorreborre (approved):

PR #7964, opened 2024-04-25
doc: propagate the `OCKAM_DEVELOPER` environment variable in docker examples

etorreborre:
This PR:

 - Fixes a typo in examples descriptions.
 - Propagates the `OCKAM_DEVELOPER` environment variable in docker examples


davide-baldo:
By adding a line, the links in the documentation are off by 1 and needs to be adjusted.
Is there any other way we can detect a developer usage? like mark specific accounts instead?

davide-baldo (approved):

PR #7965, opened 2024-04-25
feat(rust): improve output of `node create` command

adrianbenavides:
- Updates the human-readable format
- Adds machine and json output formats
- Convert some `debug` logs to `info`, and add new `info` logs, for messages that output information visible when running `node show` (e.g. services created, TCP listener created, secure-channel listener, ...)

## How to test

```bash
if [[ -z $OCKAM ]]; then
  export OCKAM=ockam
fi

# Human output, full
$OCKAM reset -y && $OCKAM node create
$OCKAM reset -y && $OCKAM node create --foreground

# Human output, with previous state
$OCKAM node create
$OCKAM node create --foreground

# Machine output
$OCKAM node create 1> ./output && cat ./output && rm ./output

# Json output
$OCKAM node create --output json | jq

# Duplicated node error message
$OCKAM node create n1
$OCKAM node create n1 # returns: "Node n1 has been already created"
$OCKAM node create n1 --foreground # returns: "Node n1 is already running"
$OCKAM node create n1 --foreground --skip-is-running-check # doesn't fail
```


etorreborre (commented):
When I tested it the machine and json output there was also some console like `Creating Node restful-gull...`. I think this a bit unusual for this kind of flag which is generally used for automation, and not interactive use.

etorreborre:
@adrianbenavides the output looks really neat!

adrianbenavides:
> When I tested it the machine and json output there was also some console like `Creating Node restful-gull...`. I think this a bit unusual for this kind of flag which is generally used for automation, and not interactive use.

@etorreborre let me tackle this in the https://github.com/build-trust/ockam/pull/7966 PR, as the only way right now to filter out some messages would be to add them as a spinner message and in the other branch I've done some modifications that will affect the json/machine outputs.

etorreborre (dismissed):

etorreborre (approved):

PR #7966, opened 2024-04-25
feat(rust): if logging is enabled, command output will be redirected to the logs

adrianbenavides:
When running a command with logging enabled, we get interleaved messages (mixed stdout/stderr messages with log messages):

image

This PR changes that behavior so that stdout/stderr gets redirected to logs if logging is enabled:

image


implementations/rust/ockam/ockam_command/src/command_global_opts.rs
```diff
@@ -47,9 +47,10 @@ impl CommandGlobalOpts {
         global_args: &GlobalArgs,
         cmd: &OckamSubcommand,
     ) -> miette::Result<Self> {
-        let terminal = Terminal::from(global_args);
+        let mut terminal = Terminal::from(global_args);
         let logging_configuration =
             Self::make_logging_configuration(global_args, cmd, terminal.clone().stdout().is_tty())?;
+        terminal.set_logging_enabled(logging_configuration.is_enabled());
```
etorreborre: Could we swap the lines 50 and 51 and pass the `logging_configuration.is_enabled()` boolean directly to create a `Terminal` instead of having a mutable method?

implementations/rust/ockam/ockam_command/src/command_global_opts.rs
```diff
@@ -210,24 +211,25 @@ impl CommandGlobalOpts {
...
-        Self {
-            global_args,
-            state,
-            terminal,
-            rt: Arc::new(Runtime::new().expect("cannot initialize the tokio runtime")),
-            tracing_guard: None,
-        }
-    }
-}
+// #[cfg(test)]
```
etorreborre: Remove that code?

etorreborre:
@adrianbenavides it seems that there is still a newline after `Creating a node`. Can we remove it?

implementations/rust/ockam/ockam_command/src/command_global_opts.rs
adrianbenavides: I commented it to check if the compiler was happy and completely forgot about it.

implementations/rust/ockam/ockam_command/src/command_global_opts.rs
adrianbenavides: The problem is that the `make_logging_configuration` needs to check the terminal... I could recreate the instance as well to avoid the mutation.

adrianbenavides:
> @adrianbenavides it seems that there is still a newline after `Creating a node`. Can we remove it?

Done, I also removed the icons introduced by `fmt_ok, fmt_warn, ...`

etorreborre (approved):

PR #7974, opened 2024-04-29
Create an exporter project member when the authority node starts

etorreborre:
This PR let the authority node create a project member which is used with an additional node started when a project is created. That node:

 - Supports a TCP outlet directed to the OpenTelemetry collector
 - Creates a relay named `ockam-opentelemetry` and which is used by a background node running an inlet where commands are executed (see #7936)

Two command line arguments have been added to the `ockam authority create` command in order to be able to persist a default project used to make the `ockam-opentelemetry-exporter` a project member:

 - `--project-access-route` Multiaddr to the project.
 - `--project-identity-identifier-file` path to the file that contains the project identifier. That file is created by the project node when it starts for the first time.


adrianbenavides (approved):
Looks good to me!

PR #7988, opened 2024-05-01
feat(rust): improve output of `node show` and `status` commands

adrianbenavides:
- The `node show` output remains the same, but I've refactored all the data that is passed between the node and the command. Now, instead of building the output out of multiple calls to the node, the node manager has a new endpoint that returns all its resources in a single call. I'll use this in a follow up PR to access this data through an HTTP endpoint. The previous refactoring also removes a lot of redundant structs
- The `status` command now shows all the identities and nodes from the system. I've removed the `--all` argument (that was used to control whether to show enrolled identities only or all of them), as it felt unnecessary.

## `node show` output

It remains the same:

image

## `status` output

Previous output (current develop):

image

Empty state:

image

After enrolling:

image

After creating more nodes and identities:

image


etorreborre (approved):
That looks really good 🪞.

PR #7998, opened 2024-05-02
(feat): use simple boolean expressions for `allow` options

etorreborre:
This PR introduces simplified boolean expressions to describe policy expressions.

# Boolean expression

A boolean expression takes the form of names which are attribute names ("attribute" in the sense of credential attribute) combined with boolean operators: `and`, `or`, `not`. For example:
```
(a or b) and (not c)
```
Such an expression is eventually transformed into a full policy expression when stored on a `Resource` or a `ResourceType`:
```
(and (or (= subject.a "true") (= subject.b "true")) (not (= subject.c "true")))
```
As you can see each attribute name is expected to contain the value "true" when evaluated.

# Command line arguments

The commands accepting policy expressions: `tcp-inlet create`, `tcp-outlet create`, and `resource-type create` now use
the `allow` instead of the `expression` option (because it reads better if we use boolean expressions in examples). `expression` is still supported for background compatibility.

An `allow` option accepts both formats: boolean expressions and full policy expressions. If the expression fails to be parsed we return an error message explaining that both formats are supported.

Note that, in order to get a genuine error when an incorrect boolean expression is entered, like `a or1 b`, the parser for full policy expressions had to be improved. Indeed, by default the expression `a or1 b` is successfully parsed by the `Expr` parser as `List([Ident("a"), Ident("or1), Ident("b")])`. The parser now expects that the first element of an `Expr` is an expression operator, like `=`, `or`, `and`, etc... Eventually, it will be a good idea to re-implement this parser to only parse valid expressions.

# Command line help

The help for the `ockam policy create` command presents the expected format for policy expressions.

# Examples

All the examples have been migrated to the new notation. I have also updated the examples to use `ockam policy create --resource-type` instead of `ockam policy create --resource` which is deprecated.

# Migration

Because of the restriction on `Expr` parsing, the previously hard-coded policy expression `subject.has_credential` is not a valid `Expr` anymore. There is a new migration to transform those expressions into `(= subject.has_credential "true")`.

# Implementation notes

 - The type `PolicyExpression` has be introduced to represent expressions which can be parsed from either a boolean expression or a policy expression.
 - I have refactored a few places when the `has_credential` policy was redefined.

### DIFF
--- a/implementations/rust/ockam/ockam_abac/src/boolean_expr.rs
+++ b/implementations/rust/ockam/ockam_abac/src/boolean_expr.rs
@@ -1,11 +1,25 @@
-use core::fmt::{Display, Formatter};
-use ockam_core::compat::str::FromStr;
-use winnow::error::{ContextError, ErrMode, StrContext};
-use winnow::Parser;
-
 use crate::Expr;
-use crate::ParseError;
+#[cfg(feature = "std")]
+use core::str::FromStr;
+use minicbor::{Decode, Encode};
+use ockam_core::compat::boxed::Box;
+use ockam_core::compat::fmt::*;
+use ockam_core::compat::format;
+use ockam_core::compat::string::*;
+use ockam_core::compat::vec::*;
+#[cfg(feature = "std")]
+use ockam_core::Result;
+#[cfg(feature = "std")]
+use std::ops::Not;
+#[cfg(feature = "std")]
+use winnow::error::{ContextError, ErrMode, StrContext};
+#[cfg(feature = "std")]
+use winnow::Parser;
 use Expr::*;
+
+#[cfg(feature = "std")]
+const NAME_FORMAT: &str =
+    "an alphanumerical name, separated with '.', '-' or '_'. The first character cannot be a digit or a '.'";
 
 /// A BooleanExpr models a boolean expression made of:
 ///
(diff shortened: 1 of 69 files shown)


implementations/rust/ockam/ockam_abac/src/boolean_expr.rs
```diff
@@ -0,0 +1,536 @@
...
+    pub fn empty() -> BooleanExpr {
+        BooleanExpr::Empty
+    }
+
+    /// Transform this boolean expression into a policy expression
+    /// by using names as attributes and setting them to the value 'true'
+    ///
+    /// Note: there is no attempt to normalize the expression and for example
+    /// transform `not a` into `= subject.a "false"`
+    pub fn to_expression(&self) -> Expr {
```
adrianbenavides, line 163: Why not implement this through the `Into` trait?

implementations/rust/ockam/ockam_abac/src/boolean_expr.rs
```diff
@@ -0,0 +1,536 @@
...
+        match self {
+            BooleanExpr::Name(s) => List(vec![
+                Ident("=".to_string()),
+                Ident(format!("subject.{}", s)),
```
adrianbenavides: Could we use the `SUBJECT_KEY` constant here?

etorreborre: Good idea.

implementations/rust/ockam/ockam_abac/src/boolean_expr.rs
etorreborre: Because the type inference with `.into()` doesn't work in many cases. I have added an instance anyway because that can make sense when it works.

adrianbenavides (approved):
🚀

PR #8000, opened 2024-05-02
Update secure channel & portal route

SanjoDeundiak:
Allows updating the route to the other side on existing Secure Channel and TCP Portal

### DIFF
--- a/implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
+++ b/implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
@@ -1,7 +1,7 @@
 use core::sync::atomic::Ordering;
 use ockam_core::compat::sync::Arc;
 use ockam_core::compat::vec::Vec;
-use ockam_core::{Any, Result, Routed};
+use ockam_core::{route, Any, Result, Route, Routed};
 use ockam_core::{Decodable, LocalMessage};
 use ockam_node::Context;
 
@@ -10,10 +10,10 @@ use crate::secure_channel::encryptor::{Encryptor, KEY_RENEWAL_INTERVAL};
 use crate::secure_channel::handshake::handshake_state_machine::CommonStateMachine;
 use crate::secure_channel::key_tracker::KeyTracker;
 use crate::secure_channel::nonce_tracker::NonceTracker;
-use crate::secure_channel::Addresses;
+use crate::secure_channel::{Addresses, Role};
 use crate::{
     DecryptionRequest, DecryptionResponse, Identities, IdentityError,
-    IdentitySecureChannelLocalInfo, PlaintextPayloadMessage, RefreshCredentialsMessage,
+    IdentitySecureChannelLocalInfo, Nonce, PlaintextPayloadMessage, RefreshCredentialsMessage,
     SecureChannelMessage,
 };
 
@@ -25,7 +25,7 @@ use tracing_attributes::instrument;
 
 pub(crate) struct DecryptorHandler {
     //for debug purposes only
-    pub(crate) role: &'static str,
+    pub(crate) role: Role,
     pub(crate) addresses: Addresses,
     pub(crate) their_identity_id: Identifier,
     pub(crate) decryptor: Decryptor,
@@ -40,7 +40,7 @@ impl DecryptorHandler {
     pub fn new(
         identities: Arc<Identities>,
         authority: Option<Identifier>,
-        role: &'static str,
+        role: Role,
         addresses: Addresses,
         key: AeadSecretKeyHandle,
         vault: Arc<dyn VaultForSecureChannels>,
@@ -79,7 +79,7 @@ impl DecryptorHandler {
         let decrypted_payload = self.decryptor.decrypt(&request.0).await;
 
         let response = match decrypted_payload {
-            Ok(payload) => DecryptionResponse::Ok(payload),
+            Ok((payload, _nonce)) => DecryptionResponse::Ok(payload),
             Err(err) => DecryptionResponse::Err(err),
         };
 
@@ -94,7 +94,19 @@ impl DecryptorHandler {
         &mut self,
         ctx: &mut Context,
         mut msg: PlaintextPayloadMessage<'_>,
+        nonce: Nonce,
+        encrypted_msg_return_route: Route,
     ) -> Result<()> {
+        if !self.role.is_initiator() {
+            let mut remote_route = self.shared_state.remote_route.write().unwrap();
+            // Only overwrite if we know that's the latest address
+            if remote_route.last_nonce < nonce {
+                let their_decryptor_address = remote_route.route.recipient()?;
+                remote_route.route = route![encrypted_msg_return_route, their_decryptor_address];
+                remote_route.last_nonce = nonce;
+            }
+        }
+
         // Add encryptor hop in the return_route (instead of our address)
         msg.return_route
             .modify()
@@ -176,6 +188,8 @@ impl DecryptorHandler {
             &self.addresses.decryptor_remote
         );
 
+        let encrypted_msg_return_route = msg.return_route();
+
         // Decode raw payload binary
         let payload = msg.into_payload();
         let payload =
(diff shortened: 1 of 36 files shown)


implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
```diff
@@ -228,7 +245,7 @@ impl Decryptor {
     }
 
     #[instrument(skip_all)]
-    pub async fn decrypt(&mut self, payload: &[u8]) -> Result<Vec<u8>> {
+    pub async fn decrypt(&mut self, payload: &[u8]) -> Result<(Vec<u8>, u64)> {
```
etorreborre: It could be a good idea to have a newtype for `Nonce`. That would make this type signature a bit more self-documenting.

And maybe even a type for the whole result of `decrypt`. Because later on, when we write in tests `decrypt(payload).0` we don't know what we are referring to without going back to the definition (another option is to pattern match in the test `let (decrypted, _nonce) = ...`).

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -20,17 +20,34 @@ use crate::{
     PlaintextPayloadMessage, RefreshCredentialsMessage, SecureChannelMessage,
 };
 
+#[derive(Debug, Clone)]
```
etorreborre, line 28: This could be the place where we have some documentation explaining what kind of problem we are trying to solve with this data type.
Also explain what is likely to be found in `route` by giving an example, because it is not a totally arbitrary route.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -369,11 +373,10 @@ impl HandshakeWorker {
                 self.shared_state.clone(),
             );
 
-            let next_hop = self.remote_route()?.next()?.clone();
             let main_mailbox = Mailbox::new(
                 self.addresses.encryptor.clone(),
                 Arc::new(AllowAll),
-                Arc::new(AllowOnwardAddress(next_hop)),
+                Arc::new(AllowAll),
```
etorreborre, line 379: Could we / should we be more restrictive?

implementations/rust/ockam/ockam_identity/src/secure_channel/options.rs
```diff
@@ -229,23 +236,18 @@ impl SecureChannelListenerOptions {
...
-                addresses.decryptor_remote.clone(),
-                &producer_flow_control_id,
-            );
+        // Add decryptor as consumer for the same ids as the listener, so that even if the initiator
+        // updates the route - decryptor is still reachable
+        for id in flow_controls.get_flow_control_ids_for_consumer(listener_address) {
+            flow_controls.add_consumer(addresses.decryptor_remote.clone(), &id);
         }
 
+        // TODO: What if we added a listener as a consumers for new FlowControlIds, should existing
```
etorreborre: s/as a consumers/as a consumer
s/throught/through
s/this/these

etorreborre: I don't understand the exact situation mentioned by the TODO:

 - "What if we added a listener as a consumers for new FlowControlIds". In which case would that happen? "new" compared to what?
 - "should existing secure channels". "existing" compared to what?

implementations/rust/ockam/ockam_identity/src/secure_channels/common.rs
```diff
@@ -54,7 +60,37 @@ impl SecureChannel {
     /// API [`Address`] of the corresponding`EncryptorWorker` Worker that can be used to encrypt
     /// a message without sending it
     pub fn encryptor_api_address(&self) -> &Address {
-        &self.encryptor_api_address
+        &self.addresses.decryptor_api
```
etorreborre: Shouldn't that be `encryptor_api`? Shouldn't a test have broken because of this?

implementations/rust/ockam/ockam_identity/src/secure_channels/common.rs
```diff
@@ -54,7 +60,37 @@ impl SecureChannel {
...
+    }
+    /// Update route to the decrpytor node on the other side in case transport changes happened
```
etorreborre: s/decrpytor/decryptor

implementations/rust/ockam/ockam_identity/src/secure_channels/common.rs
```diff
@@ -54,7 +60,37 @@ impl SecureChannel {
...
+        let mut remote_route = self.encryptor_remote_route.write().unwrap();
+
+        let old_route = remote_route.clone();
+
+        let their_decryptor_address = old_route.route.recipient()?;
+        let new_route = route![new_route, their_decryptor_address];
+
+        remote_route.route = new_route;
+
+        if let Some(next) = next {
```
etorreborre, line 82: What do we allow the `new_route` to be an empty route? When is it useful?

implementations/rust/ockam/ockam_transport_tcp/src/portal/portal_worker.rs
```diff
@@ -310,6 +311,11 @@ impl TcpPortalWorker {
             DisconnectionReason::FailedTx => {
                 self.notify_remote_about_disconnection(ctx).await?;
                 self.stop_receiver(ctx).await?;
+                // Sleep, so that if connection is dropped on both sides at the same time, the other
+                // side had time to notify us about the closure. Otherwise, the message won't be
+                // delivered which can lead to a warning message from a secure channel (or whatever
+                // is used to delivered the message). Can be removed though
```
etorreborre: s/to delivered/to deliver

implementations/rust/ockam/ockam_transport_tcp/src/portal/portal_worker.rs
```diff
@@ -322,6 +328,11 @@ impl TcpPortalWorker {
             // We couldn't read data from the tcp connection
             // Receiver should have already notified the other end and should shut down itself
             DisconnectionReason::FailedRx => {
+                // Sleep, so that if connection is dropped on both sides at the same time, the other
+                // side had time to notify us about the closure. Otherwise, the message won't be
+                // delivered which can lead to a warning message from a secure channel (or whatever
+                // is used to delivered the message). Can be removed though
```
etorreborre: s/to delivered/to deliver

implementations/rust/ockam/ockam_transport_tcp/src/transport/portals.rs
```diff
@@ -143,3 +146,67 @@ impl TcpTransport {
...
+    ) -> Self {
+        Self {
+            socket_address,
+            processor_address,
+            outlet_listener_route,
+        }
+    }
+
+    /// Socket Address
+    pub fn socket_address(&self) -> SocketAddr {
```
etorreborre, line 183: Sometimes I wish we had different names for socket "addresses" and worker "addresses".

implementations/rust/ockam/ockam_transport_tcp/src/transport/portals.rs
```diff
@@ -143,3 +146,67 @@ impl TcpTransport {
...
+
+    /// Processor address
+    pub fn processor_address(&self) -> &Address {
+        &self.processor_address
+    }
+
+    /// Update the route to the outlet node.
+    ///  NOTE: Existing TCP connections will still use the old route,
+    ///        only newly accepted connections will use the new route.
+    pub fn update_outlet_node_route(&self, new_route: Route) -> Result<()> {
```
etorreborre, line 197: It would be good to explain in the comment who would use this method, and when it would be used.

etorreborre (dismissed):
I think I understand how this all works but I would be good to have some documentation 
showing textually what is going on:

 - "Before" topology, with worker names, stored routes, flow controls
 - "After" topology, with worker names, stored routes, flow controls

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
SanjoDeundiak: I would say that a dedicated type for this return is a little bit overkill, but Nonce type is a very great idea. Will do

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
SanjoDeundiak: I don't think that makes much sense nor adds much protection

implementations/rust/ockam/ockam_identity/src/secure_channel/options.rs
SanjoDeundiak: Consider following flow:
1. You have a secure channel listener listener1 accessible from a tcp listener tcp1.
2. A secure channel sc1 is established
3. You start TcpListener tcp2
4. You make existing listener1 accessible from tcp2
Should sc1 now be accessible from tcp2? In current implementation it won't be. That's something to consider

I'll add that example to the comment

implementations/rust/ockam/ockam_identity/src/secure_channels/common.rs
SanjoDeundiak: Nice catch. It wasn't broken because this function is unused. In the test we get that address by other means (through the Registry)

implementations/rust/ockam/ockam_identity/src/secure_channels/common.rs
SanjoDeundiak: If you create a secure channel to your own node. Maybe not super useful in real world, but used in tests all the time

implementations/rust/ockam/ockam_transport_tcp/src/transport/portals.rs
SanjoDeundiak: What do you mean? I thought that's what we have in this particular place. A dedicated type, a dedicated function name

etorreborre: Yes I don't know how to address (no pun intended) this. It just seems to me that they are so different "addresses" that they should even have different names.

etorreborre (approved):

PR #8003, opened 2024-05-06
feat(rust): add an http server to the node manager to return the node resources

adrianbenavides:
When a `NodeManager` instance is created, an HTTP server is started to serve two endpoints:
- HEAD at the root `/` to quickly verify if the node is reachable
- GET at `/show` to retrieve the node's resources (same output as `node show` command)


implementations/rust/ockam/ockam_api/src/nodes/service/http.rs
```diff
@@ -0,0 +1,119 @@
...
+use tokio::net::TcpListener;
+use tokio::task::JoinHandle;
+
+use crate::nodes::models::node::NodeResources;
+use crate::nodes::BackgroundNodeClient;
+use crate::{HttpError, Result};
+
+pub struct HttpServer {
+    ctx: Arc<Context>,
+    node_client: Arc<BackgroundNodeClient>,
```
etorreborre: Maybe I misunderstood something, but since the server is already running inside a node, it should need a client to access the node data. The `get_node_resources` function is directly available on the `NodeManager`.

implementations/rust/ockam/ockam_api/src/nodes/service/http.rs
```diff
@@ -0,0 +1,119 @@
...
+use ockam_node::Context;
+use serde::Serialize;
+use tokio::net::TcpListener;
+use tokio::task::JoinHandle;
+
+use crate::nodes::models::node::NodeResources;
+use crate::nodes::BackgroundNodeClient;
+use crate::{HttpError, Result};
+
+pub struct HttpServer {
```
etorreborre, line 25: It would be good to leave a bit of documentation to explain:

 - Why do we need a HttpServer?
 - Is it supposed to support anything more than what it does now? (all the functions of the node manager API?)
 - Are there some constraints on its configuration (I'm thinking about the port. How do we make sure that there's no conflict?)

implementations/rust/ockam/ockam_command/tests/bats/local/nodes.bats
```diff
@@ -162,3 +162,14 @@ force_kill_node() {
   run_failure "$OCKAM" node create --node-config "{name: n}" --variable MY_VAR=
   assert_output --partial "Empty value for variable 'MY_VAR'"
 }
+
+@test "node - http server is started by default" {
```
etorreborre: I would phrase this more in terms of the provided feature: "A node provides an HTTP endpoint to query its status".
And maybe "The node show command displays the address to use in order to query the HTTP endpoint" as a comment for testing the `http_server` field.
Another question is: since you wrote "started by default", is there a way to avoid that behavior? If not, then we should remove "by default".

implementations/rust/ockam/ockam_api/src/nodes/service/http.rs
adrianbenavides: I started with that approach, but I got some issues. The `NodeManager` needs to keep the `JoinHandle` of the future task that runs the `HttpServer` to shut down gracefully. If I want to keep a `NodeManager` instance inside the `HttpServer`, then I need to create first the `NodeManager`, arc it, start the `HttpServer` and mutate the original `NodeManager`. That felt too complex, and I opted to decouple them using the `BackgroundNodeClient` as the intermediary.

etorreborre: Would it be the same if you used the `NodeManagerWorker` and sent a local message instead?

etorreborre: Or maybe initialize the HttpServer with references to the `registry`, the `transport` and the `cli_state` from the `NodeManager` because that's all that is required to return the nodes resources?

implementations/rust/ockam/ockam_api/src/nodes/service/http.rs
adrianbenavides: > Are there some constraints on its configuration (I'm thinking about the port. How do we make sure that there's no conflict?)

About this, I have some open questions about the UX of the `node create` command, and I wasn't confident enough to make a decision, so I left the port to be assigned automatically. I'll work on this on a separate PR when we decide how to approach this. Some of the questions:

- Should the HTTP server be optional? (i.e. adding a flag like `--disable-http-server`?)
- How do we express this feature in the `node create` command? An argument to set the port `--http-server-port`? Default to `0` to automatically assign a free port? Will the user ever care about the HTTP server port?

implementations/rust/ockam/ockam_api/src/nodes/service/http.rs
adrianbenavides: > Would it be the same if you used the NodeManagerWorker and sent a local message instead?

That has the same problem as the `NodeManager` (you need the node manager to be initialized).

> Or maybe initialize the HttpServer with references to the registry, the transport and the cli_state from the NodeManager because that's all that is required to return the nodes resources?

I think that could be assuming too much from the HttpServer point of view. What if we also want to retrieve some data outside of those objects' scope (e.g. some project data)? The cost of keeping a BackgroundNodeClient copy is minimal and it doesn't limit us.

etorreborre: It feels really wrong to me to fully step outside of the Rust process, go through the whole transport layer, to access some information that is already there.

> What if we also want to retrieve some data outside of those objects' scope (e.g. some project data)? The cost of keeping a BackgroundNodeClient copy is minimal and it doesn't limit us.

I think that in that case we can (and we should) start making high-level APIs for the `NodeManager` instead of a monolithic set of routes redirecting to a "God" object.

implementations/rust/ockam/ockam_api/src/nodes/service/http.rs
etorreborre: > Should the HTTP server be optional? 

My gut feeling is that:

 - It should be optional,`off` by default, with a `enable-http-access` option to give access to the `NodeManager` services via `HTTP` instead of the Ockam protocol. That option becomes then relevant is some production deployment scenarios when HTTP observability is required.
 
 - The port should be `0` by default and be assignable for k8s-like deployments where we can reference the port in the YAML deployment file so it's used as a probe.

implementations/rust/ockam/ockam_api/src/nodes/service/http.rs
adrianbenavides: Ok, let me try to move some things around to include the NodeManager inside the HttpServer.

implementations/rust/ockam/ockam_api/src/nodes/service/http.rs
adrianbenavides: That sounds good to me. I would name the arguments `--enable-http-server` and `--http-server-port` perhaps?

etorreborre: Yes, that looks fine to me.

implementations/rust/ockam/ockam_api/src/nodes/service/http.rs
```diff
@@ -0,0 +1,103 @@
...
+            .path()
+            .split('/')
+            .filter(|s| !s.is_empty())
+            .collect::<Vec<_>>();
+        match (req.method(), path.as_slice()) {
+            (&Method::HEAD, []) => Ok(Response::new(Full::new(Bytes::new()).boxed())),
+            (&Method::GET, ["show"]) => {
+                Self::json_response(node_manager.get_node_resources().await?)
+            }
+            _ => Ok(Response::builder()
```
etorreborre: I could be a good idea to log the incorrect request, to help when diagnosing failures.

etorreborre (dismissed):
Thanks for the changes! I think we could add one log message but otherwise it's good to 🚢.

implementations/rust/ockam/ockam_api/src/nodes/service/http.rs
adrianbenavides: Done!

etorreborre (approved):

PR #8005, opened 2024-05-06
fix(rust): use a default value for ockam_developer in examples

etorreborre:


examples/command/portals/kafka/redpanda/docker/redpanda_operator/docker-compose.yml
```diff
@@ -15,7 +15,7 @@ services:
       dockerfile: ../ockam.dockerfile
     environment:
       ENROLLMENT_TICKET: ${ENROLLMENT_TICKET:-}
-      OCKAM_DEVELOPER: ${OCKAM_DEVELOPER:-}
+      OCKAM_DEVELOPER: ${OCKAM_DEVELOPER:-false}
```
adrianbenavides, line 18: What does this syntax mean exactly? Use the existing `OCKAM_DEVELOPER` value or fallback to false if is not set?

metaclips (approved):

examples/command/portals/kafka/redpanda/docker/redpanda_operator/docker-compose.yml
etorreborre: Yes

metaclips: It'll fall back to false if OCKAM_DEVELOPER is not set https://docs.docker.com/reference/dockerfile/#environment-replacement

PR #8031, opened 2024-05-09
feat(rust): improvements for commands' output to standardize their formatting

adrianbenavides:
## Output changes

- `list` commands: I've removed the header that was shown in a box. These commands now print the items without any header
- `tcp-inlet|outlet create`: I changed the formatting of the output to use a "narrative" format, instead of showing a list of values.
- I extended the docs for some arguments (global arguments and command specific arguments)
- All arguments that have json output now print the json struct in a single line (the `pretty-print` can be done with jq if the user needs it)

image

## Fixes

- `tcp-inlet|outlet create`: the `allow` argument was set as positional, and it should be a named argument. I also changed the alias to match the arg name we use at `policy create` (`expression` instead of `policy_expression`)

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
@@ -88,10 +88,12 @@ impl CliState {
         self.notify(Notification::progress(message));
     }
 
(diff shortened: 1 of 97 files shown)


etorreborre (approved):
Great stuff 🛝

PR #8037, opened 2024-05-10
Large secure messages

etorreborre:
When we currently access the Controller via a secure channel and request a large amount of data, the `ockam_transport_core` crate [rejects the response as being too large](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_transport_core/src/transport.rs#L33).

This PR solves this issue by splitting large secure channel message into smaller parts.

# Implementation notes

## How does it work?

When a large payload is too large:

 - A UUID is generated.
 - The payload is chunked into n smaller "payload parts":
    - Each part contains the UUID, a part number, the total number of parts.
  - Then the part is encrypted and sent as a regular message.

 When the part has been decrypted:

 - It is kept in a map, with other received parts for the same UUID.
 - When all the parts have been received for a given UUID, they are assembled and the resulting message is sent on its onward route.

## Max message size

The size used to start splitting messages is `48 * 1024` bytes which is the same size as the size used by Portals to split messages. don't know why that size has been taken exactly. I suppose that it was to keep a buffer for the additional routing information.

## Cleanup

I have chosen a maximum time of 60s for removing incomplete payloads, if more than 60s have elapsed between 2 message updates.

## Backward compatibility

The Controller can be deployed with this new code and will not break old Rust clients, as long as large messages are not being sent. In principle the CBOR encoding should be more forgiving but secure channel messages are using an enum as the top-level data type (see [the note in the `minicbor-derive` library](https://docs.rs/minicbor-derive/0.15.0/minicbor_derive/index.html) about this aspect).

## Elixir implementation

This is my first significant Elixir contribution. Please tell me how I can make the code more idiomatic if necessary.


polvorin:
@etorreborre  elixir part looks good to me,  congrats for digging into it!.  There is just some minor style or simplifications we can do,  can discuss latter (like,  instead of iterating over existing keys when re-assembling the packet,  iterate over `1..NUMNER_OF_FRAGMENTS` as we _known_ them.   Can use `IO.iodata_to_binary/1` to build a binary from a list of binaries (that falls into what is refered as [iodata](https://hexdocs.pm/elixir/IO.html#module-io-data) ) and things like that, but not big deal. 

But what I'd suggest first is to add two further limits:
* Limit the max # of fragments per packet.   As receiver needs to build it up in memory, so that would avoid pathological or attack cases.  Can still be something big like 2000,  that would be a max of ~100MB per packet iiuc right?. 
* Limit the max [# of ongoing fragmented packets](https://github.com/build-trust/ockam/blob/6d82fdc3549fbd0cab35dbbc9f8a8047937fd65b/implementations/elixir/ockam/ockam/lib/ockam/secure_channel/channel.ex#L93).  Like, no more than "10", or similar.   In most cases, when the underling transport is reliable, ordered, there will be at most 1,  and couldn't even be interrelated with other packets.

davide-baldo:
There is no limit to the amount of the memory that could be allocated using this mechanism, this could lead to a DOS just by sending a very large packet.

etorreborre:
@polvorin @davide-baldo I have added limits for receiving payload parts.

etorreborre:
Closing since we increased the message size at the transport layer (for TCP).

PR #8043, opened 2024-05-13
feat(rust): introduced stress test tool for projects

davide-baldo:
see [tools/stress-test/README.md](https://github.com/build-trust/ockam/pull/8043/files#diff-fcd98909f3eaca748db81998d65309fa0c72746171e6f01e8495826b6ad2e8c3) for a description

### DIFF
--- a/tools/stress-test/README.md
+++ b/tools/stress-test/README.md
@@ -1,0 +1,81 @@
+This tool can be used to stress test a project and see how it behaves under heavy load.
+
+It shares the same `OCKAM_HOME` of the `ockam` command, and you need to perform the
+set up using the `ockam` command before using it.
+A setup could be either an `ockam enroll` or a `ockam project import`.
+Relays are created using the same code as `ockam` command but portals are created using
+a simpler implementation which doesn't involve TCP sockets.
+
+You can also use the `deploy` script in this directory to deploy the stress test to a remote machine.
+
+```
+Usage: stress-test <COMMAND>
+
+Commands:
+run       Run the stress test
+validate  Validate the configuration file
+generate  Generate sample configuration files
+help      Print this message or the help of the given subcommand(s)
+
+Options:
+-h, --help  Print help
+```
+
+To debug failures, you can use:
+
+```
+OCKAM_LOG=error stress-test run <config.toml> --log
+```
+
+Sample configuration:
+```
+peak_portals = 20
+peak_relays = 10
+ramp_up = 60
+throughput = "1 mbits"
+project = "/project/default"
+```
+
+| Parameter     | Default          | Description                                            | Possible values                                                                          |
+|---------------|------------------|--------------------------------------------------------|------------------------------------------------------------------------------------------|
+| peak_portals  | 0                | Number of portals to create                            | positive integers, 1_000 is allowed                                                      |
+| peak_relays   | 0                | Number of relays to create, at least 1 is created      | positive integers, 1_000 is allowed                                                      |
+| ramp_up       | 0                | Time, in seconds, to create all the portals and relays | positive integers                                                                        |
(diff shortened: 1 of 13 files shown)


tools/stress-test/README.md
```diff
@@ -0,0 +1,80 @@
+This tool can be used to stress test a project and see how it behaves under heavy load.
+
+It shares the same `OCKAM_HOME` of the `ockam` command, and you need to perform the
+set up using the `ockam` command before using it.
```
etorreborre, line 4: Maybe you could specify that the user needs to run `ockam enroll --authorization-code-flow`.

tools/stress-test/README.md
```diff
@@ -0,0 +1,80 @@
...
+-h, --help  Print help
+```
+
+To debug failures you can use:
+
+```
+OCKAM_LOG=error stress-test run <config.toml> --log
+```
+
+Sample configuraition:
```
etorreborre: s/configuraition/configuration

tools/stress-test/README.md
```diff
@@ -0,0 +1,80 @@
...
+```
+peak_portals = 20
+peak_relays = 10
+ramp_up = 60
+throughput = "1mbits"
+project = "/project/default"
+```
+
+| Parameter     | Description                                           |
```
etorreborre: Several things could be documented here:

 - Are there default values? 
 - What are the possible values for `throughput`?
 - In which case would I want something else than `project/default`? What would I put instead?

tools/stress-test/README.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+| Column    | Description                                                                                 |
+|-----------|---------------------------------------------------------------------------------------------|
+| Elapsed   | Time elapsed since the start of the test                                                    |
+| Portals   | Number of portals created                                                                   |
+| Relays    | Number of relays created                                                                    |
+| M. sent   | Messages sent                                                                               |
+| M. recv   | Messages received                                                                           |
+| In-fly    | Messages in flight, sent but not yet received                                               |
```
etorreborre: Maybe that should be called `In-flight` instead?

tools/stress-test/deploy
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+# This scripts deploys a stress test client and ockam client to a remote machine accessible via ssh.
```
etorreborre: s/and ockam client/and an ockam client

tools/stress-test/src/main.rs
```diff
@@ -0,0 +1,534 @@
...
+use ockam_core::Address;
+
+use crate::config::Config;
+use crate::portal_simulator::PortalStats;
+
+mod config;
+mod portal_simulator;
+
+#[derive(Debug, Args, Clone)]
+struct RunCommand {
```
etorreborre, line 26: It would be good to split this file into several concerns:

 - Making commands
 - Executing them
 - Display data
 - ...

etorreborre (dismissed):
📊

tools/stress-test/README.md
davide-baldo: You actually don't need to use the `--authorization-code-flow`, a normal enroll or a project import is enough. I've added a note about it.

tools/stress-test/README.md
davide-baldo: `In-flight` was taking too much space (I already had to compress the table to horizontal space).
I've renamed `In-fly` to `In-fli`

tools/stress-test/src/main.rs
davide-baldo: I split the logic of main into several files: main, display, execution and stats.

tools/stress-test/README.md
davide-baldo: :+1: thanks for pointing these things out, I've added a lot of information and defaults!

polvorin (approved):

Issue #8044, opened 2024-05-14
Allow to use `self` route to specify the same node


davide-baldo:
part of https://github.com/build-trust/ockam/issues/7948

PR #8049, opened 2024-05-15
feat(rust): allow to specify 'self' in multiaddr

davide-baldo:
implements https://github.com/build-trust/ockam/issues/8044

Currently if you want to specify "the same node" either you specify an empty string "" or "/secure/api" if you need packets to pass ABAC policies.

This syntax introduces some syntatic sugar we can use to avoid the user to use confusing empty string or "/secure/api" when refering a local node.
Once we eliminate the need for "/secure/api" when referring to the same node, we can resolve "self" with an empty multiaddr.


etorreborre (approved):
🦘

PR #8052, opened 2024-05-16
simplify kafka inlets

davide-baldo:
Removed kafka consumer/producer/direct implementation, all of which can be replaced either with a single kafka inlet command or with a kafka-outlet and kafka-inlet combination.
The previous commands consumer/producer were left as a redirect to kafka-inlet (while keeping the old syntax) with a `[DEPRECATED]` description plus a warning when using them.
Kafka direct was completely removed it wasn't referenced in any example or documentation (other than the CLI manual).
Implicit parameters, consumer and publishing relay,  that were somehow deduced can now be explicitly expressed.

image


davide-baldo:
Validated against `use-case - end-to-end-encrypted-kafka` bat test and redpanda/docker example

polvorin (approved):

PR #8053, opened 2024-05-16
fix(rust): node create can be created with a configuration file in the foreground

adrianbenavides:
The issue happens when we create a foreground node passing a configuration file that has a ticket. 

```yaml
# config.yaml
name: my-node
ticket: path-to-ticket
```

```bash
ockam node create -f config.yaml
```

The commands generated from the configuration file are run in the same process, sharing the same ockam context, which means that there is a global state of workers addresses that enters in a conflict state when multiple InMemoryNodes are run. Since workers are not cleaned up properly (only the top level workers are removed AFAIK), there is no easy way to isolate InMemoryNodes within the same process.

This PR solves that problem by running the commands that create InMemoryNode instances in separate processes. The affected commands are `project enroll` and `node create`.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/message/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/message/mod.rs
@@ -2,7 +2,7 @@ use clap::{Args, Subcommand};
 
 pub use send::SendCommand;
 
-use crate::CommandGlobalOpts;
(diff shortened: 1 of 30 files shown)


etorreborre (approved):
Good one 📦

PR #8059, opened 2024-05-20
feat(rust): improve error handling for nodes created in the foreground with a config

adrianbenavides:
If a node created in the foreground with a config fails (starting the node, or creating any of the resources from the config), it will return an error and delete the node to clean everything up.

Also, I added a ctrl+c handler for this specific case to avoid leaking the node's output into the next terminal prompt. See the image below:

image

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create/config.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create/config.rs
@@ -2,23 +2,18 @@ use crate::node::show::is_node_up;
 use crate::node::CreateCommand;
 use crate::run::parser::building_blocks::ArgValue;
 use crate::run::parser::config::ConfigParser;
-
 use crate::run::parser::resource::*;
 use crate::run::parser::Version;
 use crate::value_parsers::{async_parse_path_or_url, parse_enrollment_ticket, parse_key_val};
 use crate::CommandGlobalOpts;
 use clap::Args;
-
 use miette::{miette, IntoDiagnostic};
 use ockam_api::cli_state::journeys::APPLICATION_EVENT_COMMAND_CONFIGURATION_FILE;
 use ockam_api::cli_state::{random_name, EnrollmentTicket};
-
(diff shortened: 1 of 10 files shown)


implementations/rust/ockam/ockam_command/src/node/create/config.rs
```diff
@@ -180,25 +182,32 @@ impl NodeConfig {
...
+            .into_diagnostic()?
+            .success()
+        {
+            return Err(miette!("Project enroll failed"));
+        }
 
         // Next, run the 'node create' command
-        let mut child = self.node.run_in_subprocess(opts.global_args.quiet)?;
+        let child = self.node.run_in_subprocess(opts.global_args.quiet)?;
+        ctrlc::set_handler(move || {
```
davide-baldo, line 202: an unused `move` here (feel free to ignore it in this PR)

implementations/rust/ockam/ockam_command/src/node/create/config.rs
```diff
@@ -209,36 +218,37 @@ impl NodeConfig {
...
+            self.kafka_outlet.into_parsed_commands(node_name)?.into(),
         ];
         for cmds in other_sections {
             cmds.run(ctx, opts).await?;
         }
 
         // Block on the foreground node
-        child.wait().await.into_diagnostic()?;
-        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+        child.wait_with_output().await.into_diagnostic()?;
```
davide-baldo, line 235: sleep removal! nice!

davide-baldo (approved):
I really like the attention to details and UX :+1:

implementations/rust/ockam/ockam_command/src/node/create/config.rs
adrianbenavides: Good catch! I'll remove it in the next iteration to not go over the CI checks again for this, thanks 🙏

PR #8068, opened 2024-05-23
Kafka offset

davide-baldo:


implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
```diff
@@ -278,6 +326,46 @@ impl NodeManager {
...
+        &self,
+        context: &Context,
+        address: Address,
+    ) -> Result<SecureChannelListener> {
+        // skip creation if it already exists
+        if let Some(listener) = self.registry.secure_channel_listeners.get(&address).await {
+            return Ok(listener);
+        }
+
+        let options = SecureChannelListenerOptions::new()
```
SanjoDeundiak, line 339: Should we set that it's only key exchange here?

implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
```diff
@@ -278,6 +326,46 @@ impl NodeManager {
 
 /// SECURE CHANNEL LISTENERS
 impl NodeManager {
+    pub(super) async fn start_key_exchanger_service(
```
SanjoDeundiak, line 329: Should be very close to regular function, maybe worth adding a bool argument

implementations/rust/ockam/ockam_identity/src/secure_channel/listener.rs
```diff
@@ -41,6 +41,7 @@ impl SecureChannelListenerWorker {
 
         let listener = Self::new(secure_channels.clone(), identifier.clone(), options);
 
+        // FIXME: add ABAC policies for the key_exchange_only listener?
```
SanjoDeundiak, line 44: We currently don't support ABAC at this layer, but it's OK to not do use that check here for now

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -287,7 +290,14 @@ impl HandshakeWorker {
     async fn handle_decrypt(&mut self, context: &mut Context, message: Routed<Any>) -> Result<()> {
         let decryptor_handler = self.decryptor_handler.as_mut().unwrap();
         let msg_addr = message.msg_addr();
-        if msg_addr == self.addresses.decryptor_remote {
+
+        if self.key_exchange_only {
```
SanjoDeundiak, line 294: Is that intended that API won't work for regular secure channels? Also, I would consider not creating that mailbox in the first place

implementations/rust/ockam/ockam_identity/src/secure_channel/decryptor.rs
```diff
@@ -223,15 +230,27 @@ impl DecryptorHandler {
 pub(crate) struct Decryptor {
     vault: Arc<dyn VaultForSecureChannels>,
     key_tracker: KeyTracker,
-    nonce_tracker: NonceTracker,
+    nonce_tracker: Option<NonceTracker>,
+    rekeying: bool,
```
SanjoDeundiak: I think we can use self.nonce_tracker.is_some() instead

implementations/rust/ockam/ockam_api/src/nodes/service/secure_channel.rs
davide-baldo: The intention is very different compared to when creating a typical SC, and these parameters do not apply:
```
authorized_identifiers: Option<Vec<Identifier>>,
credential: Option<CredentialAndPurposeKey>,
```

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
davide-baldo: > Is that intended that API won't work for regular secure channels?

The API worker is working on both cases, unless i missed something

implementations/rust/ockam/ockam_identity/src/secure_channel/listener.rs
davide-baldo: I think we could enforce it by passing via a typical secure channel first.
This could come in handy if we want to limit the possibility to create persistent keys.
but we can discuss it later

SanjoDeundiak: That's a nice idea, but the regular secure channel listener doesn't have that check in place as well :) (and does't support it)

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
SanjoDeundiak: My bad, but maybe it could be a little bit more easier to read if that check would be inside the existing `} else if msg_addr == self.addresses.decryptor_remote {`

SanjoDeundiak (approved):

PR #8074, opened 2024-05-24
feat(rust): allow kafka portals to anchor trust on identities

polvorin:
This is a rebase of https://github.com/build-trust/ockam/pull/8072 on top of the latest changes on https://github.com/build-trust/ockam/pull/7938.

Changes are:
* Make authority optional, instead of mandatory for evaluating abac rules
* Propagate that change everywhere
* 🔴 🔴 🔴  Hardcoded  `response_outgoing_access_control`to `AllowAll` in  `interceptor_listener` ⚠️⚠️ ⚠️.   Because couldn't make the standalone, no orchestrator,   example work otherwise. No matter what policy I tried to put there  ( `(= "true" "true")` for example), it end up failing because something not being able to find the identity that the message would be sent to (iiuc `DEBUG ockam_abac::policy::outgoing: identity identifier not found; access denied policy=(= "true3" "true3")`).  So it's something with the setup or with the code, for now I'm letting this here so others can reference and test it.

**Last item requires** 👁️ !!

### DIFF
--- a/implementations/rust/ockam/ockam_abac/src/policy/policies.rs
+++ b/implementations/rust/ockam/ockam_abac/src/policy/policies.rs
@@ -9,7 +9,7 @@ use ockam_core::compat::vec::Vec;
 use ockam_core::Result;
 use ockam_identity::{Identifier, IdentitiesAttributes};
 use strum::IntoEnumIterator;
-use tracing::{debug, instrument};
+use tracing::debug;
 
 #[derive(Clone)]
 pub struct Policies {
@@ -28,14 +28,14 @@ impl Policies {
         }
     }
 
-    #[instrument(skip_all, fields(resource = %resource, action = %action, env = %env, authority = %authority))]
+    //TODO #[instrument(skip_all, fields(resource = %resource, action = %action, env = %env, authority = %authority))]
     pub fn make_policy_access_control(
         &self,
         identities_attributes: Arc<IdentitiesAttributes>,
         resource: Resource,
         action: Action,
         env: Env,
-        authority: Identifier,
+        authority: Option<Identifier>,
     ) -> PolicyAccessControl {
         debug!(
             "set a policy access control for resource '{}' of type '{}' and action '{}'",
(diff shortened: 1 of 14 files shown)


implementations/rust/ockam/ockam_abac/src/policy/policies.rs
```diff
@@ -28,14 +28,14 @@ impl Policies {
         }
     }
 
-    #[instrument(skip_all, fields(resource = %resource, action = %action, env = %env, authority = %authority))]
+    //TODO #[instrument(skip_all, fields(resource = %resource, action = %action, env = %env, authority = %authority))]
```
SanjoDeundiak:
```
#[instrument(skip_all, fields(resource = %resource, action = %action, env = %env, authority = ?authority))]
```

implementations/rust/ockam/ockam_api/src/nodes/service/manager.rs
```diff
@@ -411,6 +411,10 @@ impl NodeManager {
                 }
             }
         } else {
+            // If no expression is given, assume it's AllowAll, but only if no authority
+            // was set neither. Why: not sure, but to behave as it was previusly if there
```
SanjoDeundiak: previously

implementations/rust/ockam/ockam_api/src/kafka/outlet_service/interceptor_listener.rs
```diff
@@ -26,7 +26,7 @@ impl OutletManagerService {
         default_secure_channel_listener_flow_control_id: FlowControlId,
         policy_expression: Option<PolicyExpression>,
         request_incoming_access_control: Arc<dyn IncomingAccessControl>,
-        response_outgoing_access_control: Arc<dyn OutgoingAccessControl>,
+        _response_outgoing_access_control: Arc<dyn OutgoingAccessControl>,
```
SanjoDeundiak, line 29: Yes, need to take a look with @davide-baldo

SanjoDeundiak (commented):
Looks good

etorreborre (approved):

PR #8077, opened 2024-05-26
switch to standard relay creation for kafka usage

davide-baldo:
Also fixes relays usage in kafka scenario in develop


etorreborre (approved):
Is there any documentation to adjust?

davide-baldo:
> Is there any documentation to adjust?

no, it's a transparent change for the user

PR #8080, opened 2024-05-26
fixes kafka reject

davide-baldo:
the key exchange only secure channel doesn't exchange any credential at all, by doing so, every attempt to validate the other identity based on credential will fail.
this pr allow initial credential exchange for key exchange only, fixing credential validation.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map/secure_channels.rs
+++ b/implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map/secure_channels.rs
@@ -2,6 +2,7 @@ use crate::kafka::secure_channel_map::controller::{
     InnerSecureChannelControllerImpl, KafkaSecureChannelControllerImpl,
 };
 use crate::kafka::ConsumerResolution;
+use crate::nodes::service::SecureChannelType;
 use crate::DefaultAddress;
 use ockam::identity::SecureChannelRegistryEntry;
 use ockam_core::errcode::{Kind, Origin};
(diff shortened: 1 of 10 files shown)


implementations/rust/ockam/ockam_api/src/kafka/secure_channel_map/secure_channels.rs
```diff
@@ -22,7 +22,7 @@ impl KafkaSecureChannelControllerImpl {
 
         let secure_channel = inner
             .node_manager
-            .create_secure_channel(context, destination, None, None, None, None)
+            .create_secure_channel(context, destination, None, None, None, None, false)
```
etorreborre: I think it would help the reader and help prevent future mistakes to introduce an enum here. For example:
```
enum SecureChannelType {
  KeyExchangeAndMessages,
  KeyExchangeOnly
}
```
The other alternative is to create a builder pattern which will also allow to name all those `None`s.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -368,7 +368,17 @@ impl HandshakeWorker {
 
         // create a separate encryptor worker which will be started independently
         {
-            let rekeying = !self.key_exchange_only;
+            let rekeying;
```
etorreborre: This could be:
```rust
let (rekeying, credential_retriever) = 
   if self.key_exchange_only {
      (false, None)
   } else {
      (true, self.credential_retriever.clone())
   }
```

SanjoDeundiak (approved):

PR #8087, opened 2024-05-27
Secure Channel persistence

SanjoDeundiak:
- Create basic persistence for Secure Channel decryptor side
- The decryption key is saved to SQLite storage
- Only decryption API mailbox is started when starting a previously persisted secure channel
- Added usage of persisted secure channels to kafka use case

### DIFF
--- a/implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
+++ b/implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
@@ -11,7 +11,8 @@ use ockam_core::{
 use ockam_core::{Result, Worker};
 use ockam_node::callback::CallbackSender;
 use ockam_node::{Context, WorkerBuilder};
-use tracing::{debug, error, info};
+use ockam_vault::AeadSecretKeyHandle;
+use tracing::{debug, error, info, warn};
 use tracing_attributes::instrument;
 
 use crate::models::Identifier;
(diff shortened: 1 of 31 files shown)


implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -251,6 +268,8 @@ impl HandshakeWorker {
         let payload = message.payload();
         if let SendMessage(send_message) = self
             .state_machine
+            .as_mut()
+            .unwrap() // FIXME
```
etorreborre: Fix me now by raising an appropriate error?

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
```diff
@@ -270,11 +289,13 @@ impl HandshakeWorker {
         };
 
         // if we reached the final state we can make a pair of encryptor/decryptor
-        if let Some(final_state) = self.state_machine.get_handshake_results() {
+        // FIXME
```
etorreborre: Fix me now by raising an appropriate error?

implementations/rust/ockam/ockam_identity/src/secure_channels/common.rs
```diff
@@ -62,6 +68,16 @@ impl SecureChannel {
     pub fn encryptor_api_address(&self) -> &Address {
         &self.addresses.encryptor_api
     }
+    /// API [`Address`] of the corresponding`DecryptorWorker` Worker that can be used to decrypt
```
etorreborre: s/corresponding`DecryptorWorker`/corresponding `DecryptorWorker`

implementations/rust/ockam/ockam_identity/src/secure_channels/common.rs
```diff
@@ -62,6 +68,16 @@ impl SecureChannel {
...
+    /// a message without sending it
+    pub fn decryptor_api_address(&self) -> &Address {
+        &self.addresses.decryptor_api
+    }
+    /// API [`Address`] of the corresponding`DecryptorWorker` Worker that can be used to decrypt
```
etorreborre: s/corresponding`DecryptorWorker`/corresponding `DecryptorWorker`

implementations/rust/ockam/ockam_vault/src/traits/vault_for_secure_channels.rs
```diff
@@ -59,6 +59,12 @@ pub trait VaultForSecureChannels: Send + Sync + 'static {
         aad: &[u8],
     ) -> Result<Vec<u8>>;
 
+    /// Persiste an existing AEAD key.
```
etorreborre: s/Persiste/Persist

implementations/rust/ockam/ockam_node/src/storage/database/migrations/node_migrations/sql/20240527100000_add_sc_persistence.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE secure_channel
+(
+    role                        TEXT NOT NULL,
+    my_identifier               TEXT NOT NULL,
+    their_identifier            TEXT NOT NULL,
+    decryptor_remote_address    TEXT NOT NULL,
+    decryptor_api_address       TEXT NOT NULL,
+    decryption_key_handle       BLOB NOT NULL,
+    -- TODO: Add date?
+    node_name                   TEXT NOT NULL     -- node name to isolate credential that each node has
```
etorreborre: This comment needs to be adjusted.

polvorin (dismissed):

etorreborre (dismissed):
All good. Two things to think about:

 1. When are we going to delete the persisted keys? We could store the creation date and regularly remove old entries when new ones are added.

 2. We could consider a refactoring to extract exactly the component needed for key exchange only + decryption, instead of keeping track of booleans and options.

implementations/rust/ockam/ockam_identity/src/secure_channel/handshake/handshake_worker.rs
SanjoDeundiak: Sure, I missed it

SanjoDeundiak:
@etorreborre exactly two questions I had...
The first one is currently hard to solve, as we don't know when we want to remove that channel. The second one is definitely true and that's what I think we should do, but that's a long-term goal I think

etorreborre (approved):

PR #8088, opened 2024-05-28
fix(rust): register secure channels recreated from db

polvorin:
register them after recreated, on the secure channel registry


SanjoDeundiak:
Thanks for debugging that!

PR #8090, opened 2024-05-28
fix(rust): avoid arithmetic overflow when calculating the range of the kafka brokers ports

adrianbenavides:


implementations/rust/ockam/ockam_command/src/kafka/mod.rs
```diff
@@ -48,7 +49,27 @@ fn kafka_default_producer_server() -> SocketAddr {
 }
 
 pub(crate) fn make_brokers_port_range(bootstrap_server: &SocketAddr) -> PortRange {
-    let boostrap_server_port = bootstrap_server.port();
+    let boostrap_server_port = bootstrap_server.port() as u32;
+    let start = min(boostrap_server_port + 1, u16::MAX as u32) as u16;
+    let end = min(boostrap_server_port + 100, u16::MAX as u32) as u16;
     // we can unwrap here because we know that range start <= range end
-    PortRange::new(boostrap_server_port + 1, boostrap_server_port + 100).unwrap()
+    PortRange::new(start, end).unwrap()
```
etorreborre, line 56: We could end-up with `start=end=u16::MAX`. Could this be a problem?

adrianbenavides: I think in that case, we wouldn't have any fallback server. This function is called in the command layer, before creating anything. Perhaps we could warn the user? Not sure if returning an error would be the right move.

etorreborre: Yes a warning would be good. In production I expect people to be explicit with the ports they are using.

adrianbenavides: Done

polvorin (approved):

PR #8093, opened 2024-05-29
fix(rust): brokers port range warning message

adrianbenavides:
Returns an error when the bootstrap port overlaps with the brokers port range.

- If `--brokers-port-range` is provided, then we use that range. This range is arbitrary, so it can have any size (e.g. 8081-8082)
- If `--brokers-port-range` is not passed, then we create a range of 100 ports right after the bootstrap server. If the bootstrap server is big enough, then we trim the range so that it doesn't overflow the max value (65535)

In any case, we want to make sure that the bootstrap port is not part of the `brokers-port-range`. If it is, then we return an error.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/kafka/inlet/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/kafka/inlet/create.rs
@@ -5,12 +5,13 @@ use std::sync::Arc;
 
 use clap::{command, Args};
 use colorful::Colorful;
+use miette::miette;
 use ockam_abac::PolicyExpression;
 use ockam_api::colors::OckamColor;
 use ockam_api::kafka::{ConsumerPublishing, ConsumerResolution};
 use ockam_api::nodes::models::services::{StartKafkaInletRequest, StartServiceRequest};
 use ockam_api::nodes::BackgroundNodeClient;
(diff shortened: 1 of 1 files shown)


implementations/rust/ockam/ockam_command/src/kafka/inlet/create.rs
```diff
@@ -101,19 +102,16 @@ impl Command for CreateCommand {
...
-                "Consider changing the bootstrap server \
-                port to a lower value that has at least 100 free ports after it\n"
-            ))?;
+
+        // The bootstrap port can't overlap with the brokers port range
+        if self.from.port() >= brokers_port_range.start()
+            && self.from.port() <= brokers_port_range.end()
+        {
+            return Err(
+                miette!("The bootstrap port can't overlap with the brokers port range").into(),
```
etorreborre: This could be slightly improved by showing the actual values.

etorreborre (dismissed):

implementations/rust/ockam/ockam_command/src/kafka/inlet/create.rs
adrianbenavides: Done, now it will read like:

```
The bootstrap port 63442 can't overlap with the brokers port range 63442-63445
```

davide-baldo (approved):

PR #8094, opened 2024-05-29
fix(rust): fixes permissions and naming of local relays

davide-baldo:


davide-baldo:
replaced by relay overhaul

PR #8097, opened 2024-05-30
feat(rust): make `default` vault reuse `SqlxDatabase` instance

SanjoDeundiak:
Currently, we always create a new `SqlxDatabase` instance for the Vault, even though typically we use the same sqlite file as the main database. This is at minimum suboptimal and I'm not fully aware what guarantees sqlite gives under these conditions, so not even sure if it's safe. In this PR I implement a naive optimization which will reuse the existing `SqlxDatabase` instance if we see that this instance and the vault point to the same file.


etorreborre (approved):
🧷

PR #8099, opened 2024-05-30
feat(rust): keep secure channel even if credentials check failed

SanjoDeundiak:
Keep secure channel even if credential check failed. The authz checks are enforced on individual resources, the fact that the credential was invalid or we didn't know the authority to check it doesn't imply that the user can't access come of the resources on our node.


etorreborre (dismissed):
Is there a not too troublesome way to add a unit test for this behavior?

etorreborre (approved):

PR #8103, opened 2024-05-30
feat(rust): fixed tls tcp outlets and kafka outlets

davide-baldo:
Fix both normal TLS outlets and kafka outlets name resolution


implementations/rust/ockam/ockam_api/src/kafka/protocol_aware/metadata_interceptor.rs
```diff
@@ -157,21 +155,9 @@ impl KafkaMessageInterceptor for OutletInterceptorImpl {
 
                 for (broker_id, metadata) in response.brokers {
                     let address = format!("{}:{}", metadata.host.as_str(), metadata.port);
-                    let socket_addr = lookup_host(&address)
```
davide-baldo, line 160: I think i lost this change somewhere during rebases

adrianbenavides (approved):

PR #8115, opened 2024-05-31
Exposed and added `ockam-relay` attribute validation for relay service

davide-baldo:
Currently, rust relays are used for demo purposes only, this PR allows rust relays to work exactly as project relays.
Rust relays are exposed (both `static_forwarding_service` and `forwarding_service`).
These services require both project membership and `ockam-relay` credential attribute, like projects.
`forward_to_` prefix is applied at service level rather than client level.

### DIFF
--- a/implementations/rust/ockam/ockam/src/relay_service/relay_service.rs
+++ b/implementations/rust/ockam/ockam/src/relay_service/relay_service.rs
@@ -1,15 +1,16 @@
+use crate::alloc::string::ToString;
 use crate::relay_service::relay::Relay;
 use crate::{Context, RelayServiceOptions};
-use core::str::from_utf8;
+use alloc::string::String;
 use ockam_core::compat::boxed::Box;
-use ockam_core::{Address, Any, DenyAll, Result, Routed, Worker};
+use ockam_core::{Address, DenyAll, Encodable, Result, Routed, Worker};
+use ockam_identity::IdentitySecureChannelLocalInfo;
 use ockam_node::WorkerBuilder;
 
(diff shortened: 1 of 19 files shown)


implementations/rust/ockam/ockam/src/relay_service/relay_service.rs
```diff
@@ -46,34 +61,75 @@ impl RelayService {
...
+                    .and_then(|a| a.attrs().get("ockam-relay".as_bytes()).cloned())
+                    .and_then(|a| String::from_utf8(a).ok())
+                {
+                    match attributes.as_str() {
+                        "*" => {
+                            // allow any relay name
+                        }
+                        allowed_name => {
+                            if allowed_name != requested_relay_name {
+                                warn!(%allowed_name, "Relay creation request not authorized, relay name does not match the attribute, dropping.");
```
etorreborre: It could be a good idea to include the requested name in order to better diagnose the issue, if there is a configuration problem.

implementations/rust/ockam/ockam/src/relay_service/relay_service.rs
```diff
@@ -46,34 +61,75 @@ impl RelayService {
...
+                                return Ok(());
+                            }
+                        }
+                    }
+                } else {
+                    warn!("Relay creation request not authorized, missing or invalid `ockam-relay` attribute, dropping.");
```
etorreborre: Maybe displaying the list of existing attribute names could help here as well.

etorreborre (dismissed):

etorreborre (approved):

PR #8164, opened 2024-06-12
feat(rust): change tcp protocol serialization

SanjoDeundiak:
1. Optimize `Address` CBOR implementation to use byte strings instead of array of bytes which saves size BREAKING CHANGE
2. Create a `TcpTransportMessage` type dedicated to exchange messages using TCP transport exclusively with optimized encoding and support for larger messages (current max allowed size is 16 MB) BREAKING CHANGE
3. Optimize Secure Channel message encoding (remove bare wrapping) + affected by change `1.` BREAKING CHANGE
4. Add Secure Channel padding
5. Remove `protocol_version` from `LocalMessage`, `RelayMessage` and `Context`


implementations/rust/ockam/ockam_api/src/kafka/protocol_aware/mod.rs
```diff
@@ -76,11 +76,10 @@ impl KafkaMessageInterceptor for InletInterceptorImpl {
 
 #[derive(Debug, Clone, Decode, Encode)]
 #[rustfmt::skip]
-#[cbor(map)]
 /// Wraps the content within every record batch
 struct MessageWrapper {
```
SanjoDeundiak, line 81: @davide-baldo does that need a change on the elixir side?

implementations/rust/ockam/ockam_core/src/routing/address.rs
```diff
@@ -139,6 +147,11 @@ impl Address {
     pub fn is_local(&self) -> bool {
         self.tt == LOCAL
     }
+
+    /// Take inner Vec
+    pub fn inner(self) -> Vec<u8> {
```
etorreborre, line 152: Instead of exposing the implementation, I would add a `to_local` method and use it [here](https://github.com/build-trust/ockam/pull/8164/files#diff-fb6b2451268d57544e5b8570ee3d4cff294dab6325a3f48426547f219d83c452R145).

implementations/rust/ockam/ockam_core/src/routing/route.rs
```diff
@@ -252,6 +252,11 @@ impl Route {
     pub fn is_local(&self) -> bool {
         self.iter().all(|a| a.is_local())
     }
+
+    /// Return inner VecDeque
```
etorreborre, line 256: This function is not used.

implementations/rust/ockam/ockam_core/src/routing/transport_type.rs
```diff
@@ -19,6 +31,11 @@ impl TransportType {
         TransportType(n)
     }
 
+    /// Raw value
```
etorreborre, line 34: This function is not used.

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
```diff
@@ -312,17 +284,35 @@ impl EncryptorWorker {
...
+            self.addresses.encryptor.clone(),
+        )
+        .await?;
 
         Ok(())
     }
+
+    fn add_padding(msg: SecureChannelMessage) -> SecureChannelPaddedMessage {
+        // Naїve padding of 0 to 255 zeros
+        // let padding_length: u8 = ockam_core::compat::rand::random();
```
etorreborre, line 306: Why is the padding commented out for now? Would that be possible to add a test to show that messages are effectively padded, to make sure that we don't have regressions there?

implementations/rust/ockam/ockam_transport_tcp/src/protocol_version.rs
```diff
@@ -0,0 +1,26 @@
...
+    }
+}
+
+impl TryFrom<u8> for TcpProtocolVersion {
+    type Error = ockam_core::Error;
+
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        match value {
+            1 => Ok(TcpProtocolVersion::V1),
+            _ => Err(TransportError::InvalidProtocolVersion)?,
```
etorreborre, line 23: It would be good to include the incorrect version in the exception for debugging.

implementations/rust/ockam/ockam_transport_tcp/src/transport_message.rs
```diff
@@ -0,0 +1,120 @@
...
+use opentelemetry::{
+    global,
+    trace::{Link, SpanBuilder, TraceContextExt, Tracer},
+    Context,
+};
+
+/// TCP transport message type.
+#[derive(Debug, Clone, Eq, PartialEq, Encode, Decode, CborLen)]
+#[rustfmt::skip]
+pub struct TcpTransportMessage<'a> {
```
etorreborre, line 11: It seems to be that most of this code could stay in the `ockam_transport` crate, otherwise we will have to duplicate it for every transport.

It looks like the specificity of TCP is [here](https://github.com/build-trust/ockam/pull/8164/files#diff-8ac45ee8284813729705924bd44521c87c21eb8eeb29fd3385b8bd2ba4d9561cR122), where we serialize the data but that is independent from the `TransportMessage` datatype.

implementations/rust/ockam/ockam_transport_tcp/src/workers/sender.rs
```diff
@@ -114,6 +118,30 @@ impl TcpSendWorker {
...
     }
+
+    fn serialize_message(&self, local_message: LocalMessage) -> Result<Vec<u8>> {
+        // Create a message buffer with prepended length
+        let transport_message = TcpTransportMessage::from(local_message);
+
+        let msg_len = minicbor::len(&transport_message);
+
+        if msg_len > MAX_MESSAGE_SIZE {
+            return Err(TransportError::MessageLengthExceeded)?;
```
etorreborre: We could put the actual size in the error message.

implementations/rust/ockam/ockam_transport_tcp/src/workers/sender.rs
```diff
@@ -114,6 +118,30 @@ impl TcpSendWorker {
...
+        }
+
+        let len = 4 /* len */ + msg_len;
```
etorreborre: Maybe instead of `/* len */` a more explicit comment saying `We need 4 bytes to serialize the size of the message`.

implementations/rust/ockam/ockam_transport_tcp/src/workers/sender.rs
```diff
@@ -133,6 +161,22 @@ impl Worker for TcpSendWorker {
             self.receiver_flow_control_id.clone(),
         ));
 
+        // First thing send our protocol version
```
etorreborre, line 177: It's nice to have the protocol version serialized first now.

implementations/rust/ockam/ockam_transport_tcp/src/workers/sender.rs
```diff
@@ -114,6 +118,30 @@ impl TcpSendWorker {
...
+        let len = 4 /* len */ + msg_len;
+
+        let msg_len_u32 =
+            u32::try_from(msg_len).map_err(|_| TransportError::MessageLengthExceeded)?;
+
+        let mut vec = vec![0u8; len];
+
+        vec[..4].copy_from_slice(&msg_len_u32.to_be_bytes());
+        minicbor::encode(&transport_message, &mut vec[4..])
+            .map_err(|_| TransportError::Encoding)?;
```
etorreborre: Can we keep the encoding error?

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
SanjoDeundiak: We should pick the strategy for padding, which I think we'll do later. For now I only added a field so that we can add padding without any compatibility issues later. Adding a test is problematic since the API for padding is not exposed

implementations/rust/ockam/ockam_transport_tcp/src/protocol_version.rs
SanjoDeundiak: I didn't want to introduce additional data to `TransportError` (currently none of its members have it)

implementations/rust/ockam/ockam_transport_tcp/src/transport_message.rs
SanjoDeundiak: I think it's fine to duplicate, different transports' on-the-wire structs are completely decoupled from each other. At the same time I would be OK sharing those structs, if they were exactly the same, which is not the case here. `TcpTransportMessage` doesn't have protocol type and uses CBOR only, while TransportMessage uses only BARE. It would be easier in my opinion to make further changes not worrying to break something in other transports

implementations/rust/ockam/ockam_core/src/routing/address.rs
SanjoDeundiak: I'm becoming a fan of exposing implementation lately to be honest. In context of rust it's a) allows to reduce number of clones in some places b) helps borrow checker figure out when we move the inner field, and not the whole struct. That being said I would even consider making fields public more often... the only thing I don't like, is that we also give mutable access to the field as a side-effect.
Speaking of this specific case, I don't think converting the Address to local makes any sense, it's only case - test code. So I vote against making that a part of public API

implementations/rust/ockam/ockam_core/src/routing/route.rs
SanjoDeundiak: I that a problem?

implementations/rust/ockam/ockam_transport_tcp/src/workers/sender.rs
SanjoDeundiak: Do you think it would be useful? Decoding errors is what can happen and having details would be useful... encoding errors are 99% bugs found during development

implementations/rust/ockam/ockam_transport_tcp/src/transport_message.rs
etorreborre: I don't see why we can't have different serialization formats for the same, or mostly the same, data types. Now the code above is duplicating [some tracing logic](https://github.com/build-trust/ockam/pull/8164/files#diff-b540e777cb15df1fe532d0c6946f4934d3f201c726d6d0eb5bd2b0ee3ce30dc1R52) that has nothing to do with the serialization.

implementations/rust/ockam/ockam_transport_tcp/src/workers/sender.rs
etorreborre: In general I think it's better to not drop information, and one day missing it when debugging a production problem, at least to be able to say with 100% certainty that a bug does not come from that line.

implementations/rust/ockam/ockam_transport_tcp/src/protocol_version.rs
etorreborre: I think that this kind of information can make the difference when debugging production issues. Because the actual value can give you clues about what when wrong.

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
etorreborre: Can't we test something that when we encrypt the same message twice, the on-wire bytes are likely to have a different size?

implementations/rust/ockam/ockam_core/src/routing/route.rs
etorreborre: That's more code to inspect when trying to understand what's going on. But I'd rather not expose implementation details which you seem to like, seeing your comment above :-).

implementations/rust/ockam/ockam_core/src/routing/address.rs
etorreborre: I'm definitely not a fan of exposing implementations, especially not as a blanket rule. Performance optimisations can sometimes require it, but I'd rather have this trade-off on a case-by-case basis.

implementations/rust/ockam/ockam_identity/src/secure_channel/encryptor_worker.rs
SanjoDeundiak: Why would it be different if we don't add any padding at the moment?

etorreborre: I meant when padding is used of course :-). I was responding to your comment:
> Adding a test is problematic since the API for padding is not exposed

SanjoDeundiak: Yes, so currently padding is always empty and you can't use it. So it's hard to test

implementations/rust/ockam/ockam_transport_tcp/src/transport_message.rs
SanjoDeundiak: That's a good point, do you think we should move that code to a common place and reuse?

etorreborre: Yes I'd like that.

davide-baldo (dismissed):

polvorin (approved):

PR #8171, opened 2024-06-13
feat(rust): add `disable-content-encryption` flag to the kafka-inlet create command

davide-baldo:


polvorin (approved):
IIUC the scenario where this is useful is if client can't connect to kafka directly,  but uses a project relay to reach it.    Messages won't be encrypted on _kafka_ ,  but in-flight messages between the two networks (client and kafka one) _will_ be as they pass through the project relay?

davide-baldo:
> IIUC the scenario where this is useful is if client can't connect to kafka directly, but uses a project relay to reach it. Messages won't be encrypted on kafka , but in-flight messages between the two networks (client and kafka one) will be as they pass through the project relay?

Exactly

PR #8191, opened 2024-06-19
refactor(rust): `project-member` commands, and adds the `show` command

adrianbenavides:


implementations/rust/ockam/ockam_command/src/project_member/delete.rs
```diff
@@ -28,78 +30,147 @@ pub struct DeleteCommand {
...
+
+#[derive(Serialize)]
+struct DeleteMemberOutput {
+    project: String,
+    identifiers: Vec<Identifier>,
+}
+
+impl Display for DeleteMemberOutput {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self.identifiers.len() {
```
etorreborre: Pattern matching on the slice would be a bit more elegant:
```
match self.identifiers[..] {
    [] => "There are no members",
    [identifier] => "Identifier {} ...",
    _ => "n members were deleted", 
}
```

implementations/rust/ockam/ockam_command/src/project_member/list.rs
```diff
@@ -24,55 +23,49 @@ pub struct ListCommand {
...
         let members = authority_node_client
             .list_members(ctx)
             .await?
             .into_iter()
-            .map(|i| MemberOutput(i.0, i.1))
-            .collect();
-
-        print_members(&opts, members)?;
+            .filter(|(_, a)| {
+                if self.enrollers {
```
etorreborre: That's some nitpicking :-) but this doesn't need an if/then/else:
```
!self.enrollers || 
a.deserialized_key_value_attrs().contains(&format!("{}={}", 
    OCKAM_ROLE_ATTRIBUTE_KEY, OCKAM_ROLE_ATTRIBUTE_ENROLLER_VALUE))                      
```

etorreborre (dismissed):

implementations/rust/ockam/ockam_command/src/project_member/delete.rs
adrianbenavides: Oh I like that, I wasn't aware of this!

implementations/rust/ockam/ockam_command/src/project_member/list.rs
adrianbenavides: Good catch, fixed too!

etorreborre (approved):

PR #8194, opened 2024-06-19
Handling hostname rather than raw socket address in portals

davide-baldo:


davide-baldo:
Superseded by https://github.com/build-trust/ockam/pull/8248

PR #8202, opened 2024-06-21
Support Postgres as a persistence backend

etorreborre:
This PR adds the possibility to use Postgres as a backend to persist data for an Ockam node.

In order to use a Postgres database simply set the following environment variables:

 - `OCKAM_POSTGRES_HOST`: for example `localhost`
 - `OCKAM_POSTGRES_PORT`: for example `5432`
 - `OCKAM_POSTGRES_DATABASE_NAME` (optional): for example `postgres`
 - `OCKAM_POSTGRES_USER` (optional): for example `admin`
 - `OCKAM_POSTGRES_PASSWORD` (optional): for example `password`

# Implementation notes

This PR touches a large number of files and the following sections will help the reviewer understanding the changes.

## Driver

The `sqlx` library provides an `Any` driver, which replaces the `SQLite` driver used before this PR.
With the `Any` driver and its associated types, `AnyConnection`, `AnyPool`, `Type<Any>` etc... we can open a connection to either a SQLite or a Postgres database by specifying the proper URL:

 - `sqlite:file:in-memory?mode=memory&cache=shared` for an in-memory SQLite database
 - `sqlite:file://my-file?mode=rwc` for an on-disk SQLite database
 - `postgres://admin:password@localhost:5432/postgres` for a Postgres database

This logic is encapsulated in a new `DatabaseConfiguration` type.

The result of this driver change is that a number of method signatures had to be changed from `... SQLite ...` to `... Any ...`.

## Queries

Some queries had to be changed in order to work with both databases.

In particular the queries of the form `INSERT OR REPLACE` don't work with Postgres. Those queries have been replaced with `INSERT ... ON CONFLICT ... DO UPDATE | NOTHING`.

## Data types serialization

The mapping of Rust types to database types is quite different between SQLite and Postgres.
To support this properly:

The `ToSqlxType` typeclass has been removed in favor of using the `Encode` and `Type` typeclasses from `sqlx` directly. While this typeclass was useful when using SQLite only, it became a source of encoding bugs with both Postgres and SQLite.

Reading boolean and optional values from the database was not working out of the box. This was fixed via:
   - A `Boolean` datatype to read a boolean value from a database column
   - A `Nullable<T>` datatype to read an optional value from a database column possibly containing NULL values
   - Two PRs to the `sqlx` library:
     - https://github.com/launchbadge/sqlx/pull/3291 (merged) to support the `Bool` type.
     - https://github.com/launchbadge/sqlx/pull/3298 (not yet merged) to track null values types.
     - This explains why the `sqlx` dependencies mentioned in the Cargo files now point to a fork.

### Repositories

All the repositories code is now tested with 3 databases: SQLite in-memory, SQLite on-disk and Postgres (if available in the environment). This test support is provided by the `sqlx_database::with_dbs` function.

### Execution

Note that you need to run the tests sequentially if you run them with a local Postgres database since it will be shared between all tests. For example:
```
cargo --locked nextest --config-file ./tools/nextest/.config/nextest.toml run --no-fail-fast --test-threads 1
```

### Clean-up

For the same reason, some system tests are now calling `TestNode::clean()` in order to make sure that the database is cleaned-up before the test runs.

### CI

I have added a CI job to run all the `nextest` test suite with a Postgres database. That job only selects test names containing `sql`, to cover all the repositories, and `cli_state` to test the high-level behavior which generally involves some persistence.

## Vaults

The modeling for vaults has changed slightly. An enum has been introduced to represent exactly a vault metadata:
```rust
pub enum VaultType {
    DatabaseVault {
        use_aws_kms: UseAwsKms,
    },
    LocalFileVault {
        path: PathBuf,
        use_aws_kms: UseAwsKms,
    },
}
```
A vault is either stored directly in the database, with some dedicated tables, or in a file (using SQLite).
The naming of `is_kms` has been changed to `use_aws_kms` to indicate that:

 - If the Vault uses a `KMS` it is specifically an AWS one
 - Not _all_ of the Vault data is stored in the KMS. Some data is still stored locally, only the signing keys are stored in the KMS.
 - I have also made some change to the output of `ockam vault list/show` to better reflect the vault types:
```
❯ ockam vault list
   │ Name: default
   │ Type: Internal

   │ Name: v1
   │ Type: External
   │ Path: /Users/etorreborre/.ockam-4/vault-v1

   │ Name: v2
   │ Type: External
   │ Path: /Users/etorreborre/.ockam-4/second-vault

   │ Name: v3
   │ Type: External
   │ Path: /Users/etorreborre/.ockam-4/vault-v3
   │ Uses AWS KMS: true
```

 - The JSON output of the `ockam vault show` command also changes. This is not necessarily the best JSON output but I didn't try to improve it for now.


implementations/rust/ockam/ockam_vault/src/storage/secrets_repository_sql.rs
```diff
@@ -174,48 +193,91 @@ impl SecretsRepository for SecretsSqlxDatabase {
...
-        }
+impl Type<Any> for SigningSecret {
+    fn type_info() -> <Any as Database>::TypeInfo {
+        <Vec<u8> as Type<Any>>::type_info()
+    }
+}
+
+impl Encode<'_, Any> for SigningSecret {
+    fn encode_by_ref(&self, buf: &mut <Any as HasArguments>::ArgumentBuffer) -> IsNull {
+        let key = match self {
```
SanjoDeundiak: let key = self.key().to_vec();

implementations/rust/ockam/ockam_vault/src/storage/secrets_repository_sql.rs
```diff
@@ -174,48 +193,91 @@ impl SecretsRepository for SecretsSqlxDatabase {
...
+    }
+}
+
+impl Encode<'_, Any> for SigningSecretKeyHandle {
+    fn encode_by_ref(&self, buf: &mut <Any as HasArguments>::ArgumentBuffer) -> IsNull {
+        <HandleToSecret as Encode<'_, Any>>::encode_by_ref(self.handle(), buf)
+    }
+}
+
+impl Type<Any> for HandleToSecret {
```
SanjoDeundiak, line 220: As a future improvement, I feel like we can derive it somehow

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
```diff
@@ -223,54 +409,79 @@ impl<T> ToVoid<T> for core::result::Result<T, sqlx::error::Error> {
...
-    use sqlx::sqlite::SqliteQueryResult;
-    use sqlx::FromRow;
-    use tempfile::NamedTempFile;
-
-    use crate::database::ToSqlxType;
+/// Create a temporary database file that won't be cleaned-up automatically
+pub fn create_temp_db_file() -> Result<PathBuf> {
+    let (_, path) = NamedTempFile::new()
+        .map_err(|e| Error::new(Origin::Core, Kind::Io, format!("{e:?}")))?
+        .keep()
```
SanjoDeundiak, line 420: Is it intended to keep the file?

SanjoDeundiak: Also I see some places still use `NamedTempFile` directly, is that intended?

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
```diff
@@ -39,41 +43,91 @@ impl Debug for SqlxDatabase {
...
 }
 
 impl SqlxDatabase {
-    /// Constructor for a database persisted on disk
-    pub async fn create(path: impl AsRef<Path>) -> Result<Self> {
-        Self::create_impl(path, Some(NodeMigrationSet)).await
+    /// Constructor for a database
+    pub async fn create(configuration: DatabaseConfiguration) -> Result<Self> {
+        Self::create_impl(
+            configuration.clone(),
```
SanjoDeundiak: Let's avoid clone here

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
```diff
@@ -114,54 +175,64 @@ impl SqlxDatabase {
...
 
-    pub(crate) async fn create_connection_pool(path: &Path) -> Result<SqlitePool> {
-        let options = SqliteConnectOptions::new()
-            .filename(path)
-            .create_if_missing(true)
-            .log_statements(LevelFilter::Trace)
-            .log_slow_statements(LevelFilter::Trace, Duration::from_secs(1));
-        let pool = SqlitePool::connect_with(options)
+    pub(crate) async fn create_connection_pool(
+        configuration: DatabaseConfiguration,
```
SanjoDeundiak: Can we use a reference instead of the value here?

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_from_row_types.rs
```diff
@@ -0,0 +1,80 @@
+use sqlx::database::HasValueRef;
+use sqlx::error::BoxDynError;
+use sqlx::postgres::any::AnyTypeInfoKind;
+use sqlx::{Any, Database, Decode, Type, Value, ValueRef};
+
+/// This type is used to map Option<T> fields for the types deriving `FromRow`
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Nullable<T>(Option<T>);
```
SanjoDeundiak, line 8: Should we also implement `From<Option<T>> for Nullable<T>` and `From<Nullable<T>> for Option<T>`?

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_from_row_types.rs
```diff
@@ -0,0 +1,80 @@
...
+
+impl<T: Clone> Nullable<T> {
+    /// Create a new Nullable value
+    pub fn new(t: Option<T>) -> Self {
+        Nullable(t)
+    }
+
+    /// Return the Option corresponding to this value in the database
+    pub fn to_option(&self) -> Option<T> {
```
SanjoDeundiak, line 29: Should we avoid cloning and use `into()` instead?

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_from_row_types.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    fn compatible(ty: &<Any as Database>::TypeInfo) -> bool {
+        <T as Type<Any>>::compatible(ty) || ty.kind() == AnyTypeInfoKind::Null
+    }
+}
+
+/// This type is used to map boolean fields for the types deriving `FrowRow`.
+/// Postgres provides a proper boolean type but SQLite maps them as integers.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Boolean(bool);
```
SanjoDeundiak, line 56: Implementing `From` in both ways would have been nice

implementations/rust/ockam/ockam_identity/src/identities/storage/credential_repository_sql.rs
```diff
@@ -87,27 +89,59 @@ impl CredentialRepository for CredentialSqlxDatabase {
...
+            .bind(issuer)
+            .bind(scope)
+            .bind(self.node_name.clone());
         query.execute(&*self.database.pool).await.void()
     }
 }
 
+// Database serialization / deserialization
+
+//
```
SanjoDeundiak: Should there be a comment?

implementations/rust/ockam/ockam_command/src/environment/static/env_info.txt
```diff
@@ -21,6 +21,20 @@ Logging
 - OCKAM_LOG_MAX_FILES: an `integer` that defines the maximum number of log files to keep per node. Default value `60`.
 - OCKAM_LOG_CRATES_FILTER: a filter for log messages based on crate names: `all`, `default`, comma-separated list of crate names. Default value: `default`, i.e. the list of `ockam` crates.
 
+Database
```
SanjoDeundiak, line 24: Didn't know we have this file :D

implementations/rust/ockam/ockam_app_lib/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![recursion_limit = "256"]
```
SanjoDeundiak: Could you please add an explanation here?

implementations/rust/ockam/ockam_api/tests/direct_authenticator.rs
```diff
@@ -54,7 +54,7 @@ async fn admin_cant_delete_admin(ctx: &mut Context) -> Result<()> {
     Ok(())
 }
 
-#[ockam_macros::test]
+#[ockam_macros::test(timeout = 6000000)]
```
SanjoDeundiak: Testing leftover?

implementations/rust/ockam/ockam_api/tests/common/common.rs
```diff
@@ -129,7 +131,7 @@ pub async fn start_authority(
     configuration.no_direct_authentication = false;
     configuration.no_token_enrollment = false;
 
-    println!(
+    debug!(
```
SanjoDeundiak, line 134: I wonder why it was like that here. @polvorin is that intentional?

implementations/rust/ockam/ockam_api/src/nodes/service/in_memory_node.rs
```diff
@@ -58,10 +58,13 @@ impl Drop for InMemoryNode {
         // because in that case they can be restarted
         if !self.persistent {
             executor::block_on(async {
-                self.node_manager
-                    .delete_node()
-                    .await
-                    .unwrap_or_else(|e| panic!("cannot delete the node {}: {e:?}", self.node_name))
+                // We need to recreate the CliState here to make sure that
```
SanjoDeundiak, line 61: Should this be fixed before landing?

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
```diff
@@ -682,10 +686,11 @@ mod test {
     }
 
     #[allow(non_snake_case)]
-    #[ockam_macros::test(timeout = 5_000)]
+    #[ockam_macros::test(timeout = 500_000)]
```
SanjoDeundiak: Is their a particular reason to increase the timeout that much?

implementations/rust/ockam/ockam_api/src/cli_state/test_support.rs
```diff
@@ -1,12 +1,30 @@
...
     /// Return a test CliState with a random root directory
+    /// Use this CliState for a simple integration test since every call to that function deletes
+    /// all previous state if the database being used is Postgres.
     pub async fn test() -> Result<Self> {
-        Self::create(Self::test_dir()?).await
+        let test_dir = Self::test_dir()?;
+
+        // clean the existing state if any
+        let db = SqlxDatabase::create(CliState::make_database_configuration(&test_dir)?).await?;
+        db.drop_all_postgres_tables().await?;
```
SanjoDeundiak, line 16: Should we hide implementation details here?

implementations/rust/ockam/ockam_vault/src/storage/secrets_repository_sql.rs
```diff
@@ -50,15 +52,21 @@ impl SecretsRepository for SecretsSqlxDatabase {
             SigningSecretKeyHandle::ECDSASHA256CurveP256(_) => EC_DSA_SHA256_CURVE_P256.into(),
         };
 
-        let query = query("INSERT OR REPLACE INTO signing_secret VALUES (?, ?, ?)")
-            .bind(handle.to_sql())
-            .bind(secret_type.to_sql())
-            .bind(secret.to_sql());
+        let query = query(
```
SanjoDeundiak, line 55: Can this affect AWS KMS vault? I'm asking because we don't have CI tests for it AFAIK, so could these changes potentially break something in a way that we don't catch it?

implementations/rust/ockam/ockam_api/src/cli_state/storage/journeys_repository_sql.rs
```diff
@@ -282,14 +296,17 @@ mod test {
...
     /// This test checks that we can store project journeys with a previous / next relationship
     #[tokio::test]
     async fn test_several_project_journeys() -> Result<()> {
-        let repository = create_repository().await?;
+        with_application_dbs(|db| async move {
+            let repository: Arc<dyn JourneysRepository> =
+                Arc::new(JourneysSqlxDatabase::new(db));
 
-        // create and store a the first host journey
+            // create and store a the first host journey
```
SanjoDeundiak, line 309: Apparently cargo fmt gave up )

implementations/rust/ockam/ockam_node/src/storage/database/migrations/node_migrations/sql/postgres/20240613100000_create_database.sql
```diff
@@ -0,0 +1,359 @@
+--------------
```
SanjoDeundiak, line 1: I hope it's correct...)
BTW: did you create it manually and went over all the sqlite migrations, or did you use some tool to reverse engineer a final state of the sqlite table?

SanjoDeundiak:
Nice!

implementations/rust/ockam/ockam_api/tests/common/common.rs
polvorin: my bad, a leftover. Some test were failing and I wondered why.   Letting it as debug is fine,  or remove it entirely as it doesn't look useful anymore now.

implementations/rust/ockam/ockam_api/src/cli_state/storage/vaults_repository_sql.rs
```diff
@@ -27,43 +28,57 @@ impl VaultsSqlxDatabase {
...
-        query.execute(&*self.database.pool).await.void()?;
-
-        Ok(NamedVault::new(name, path.into(), is_kms))
+    async fn store_vault(&self, name: &str, vault_type: VaultType) -> Result<NamedVault> {
+        let mut transaction = self.database.begin().await.into_core()?;
+
+        let query1 =
+            query_scalar("SELECT EXISTS(SELECT 1 FROM vault WHERE is_default = $1)").bind(true);
+        let default_exists: Boolean = query1.fetch_one(&mut *transaction).await.into_core()?;
+        let default_exists = default_exists.to_bool();
```
adrianbenavides, line 37: I need to check it, but now the "is_default" check is done here directly instead of in a higher layer?

implementations/rust/ockam/ockam_api/src/ui/output/utils.rs
```diff
@@ -47,6 +47,15 @@ pub fn colorize_connection_status(status: ConnectionStatus) -> CString {
     }
 }
 
+pub fn indent(indent: impl Into<String>, text: impl Into<String>) -> String {
```
adrianbenavides, line 50: I think this is not necessary. We have some formatting tools [here](https://github.com/build-trust/ockam/blob/8db3748cf20faeb29ad1dc70d60464049959b7d7/implementations/rust/ockam/ockam_api/src/ui/terminal/fmt.rs) which we use to customize the command output as needed (example [here](https://github.com/build-trust/ockam/blob/d63a5045bd400a08da4dcc57dfe9da623ca6e0b5/implementations/rust/ockam/ockam_api/src/cloud/project/project.rs#L250)).

We can land it as it is, I'll take a look once it's merged.

implementations/rust/ockam/ockam_command/src/environment/static/env_info.txt
adrianbenavides: Yeah, it's actually the output of the `ockam environment` command

implementations/rust/ockam/ockam_command/src/vault/util.rs
```diff
@@ -27,48 +28,82 @@ impl Output for VaultOutput {
         Ok(formatdoc!(
             r#"
             Vault:
-                Name: {name}
-                Type: {vault_type}
-                Path: {vault_path}
+            {vault}
```
adrianbenavides, line 31: This output is definitely super outdated 😆  I'll check it once this is landed.

adrianbenavides (commented):
Awesome! I pray to the gods that we never have to support oracle 😄

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
etorreborre: > Is it intended to keep the file?

Yes, that's the intention of that method for the case where you want to debug a test and inspect the database afterwards.

> Also I see some places still use NamedTempFile directly, is that intended?

Yes it usually is. For example to create a temporary directory in `test_log_and_traces`.

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
etorreborre: Done.

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
etorreborre: Done.

implementations/rust/ockam/ockam_vault/src/storage/secrets_repository_sql.rs
etorreborre: Done.

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_from_row_types.rs
etorreborre: This type is only used in one direction `Nullable<T>` to `Option<T>`. So we could implement `From<Nullable<T>> for Option<T>`. The problem is that this changes `to_option()` to `.clone().into()` in many places, see my next comment.

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_from_row_types.rs
etorreborre: I think we should address this is in a separate PR, because there would still be some cloning involved. There are many places where we do some cloning of individual fields to re-assemble them in the final desired type, to go from `&MyTypeRow` to `MyType`. What we gain in exchange is separate methods to parse the row fields if only a subset are needed.
We can definitely optimize this pattern but let's tackle this separately.

implementations/rust/ockam/ockam_identity/src/identities/storage/credential_repository_sql.rs
etorreborre: Fixed.

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_from_row_types.rs
etorreborre: For the same reason as above, it's not so nice because, in the current state, we need to `.clone().into()`.

implementations/rust/ockam/ockam_api/tests/direct_authenticator.rs
etorreborre: Left-over from testing. Removed.

implementations/rust/ockam/ockam_api/src/kafka/portal_worker.rs
etorreborre: Left-over from testing. Removed.

implementations/rust/ockam/ockam_app_lib/src/lib.rs
etorreborre: The compiler complained loudly about breaking some recursion level at some stage and I had to raise the recursion limit. Strangely this does not happen anymore, so I just removed it.

implementations/rust/ockam/ockam_api/src/nodes/service/in_memory_node.rs
etorreborre: Oops, this is also a left-over from testing. Thanks!

implementations/rust/ockam/ockam_api/src/cli_state/test_support.rs
etorreborre: Maybe but so far I don't really know what's the right abstraction to introduce there.

implementations/rust/ockam/ockam_api/src/cli_state/storage/journeys_repository_sql.rs
etorreborre: Indeed, I had to do the formatting everywhere `with_dbs/with_application_dbs` was called and I forgot that spot.

implementations/rust/ockam/ockam_vault/src/storage/secrets_repository_sql.rs
etorreborre: If I'm not wrong this code is not called when an AWS KMS vault is used, is it?

implementations/rust/ockam/ockam_node/src/storage/database/migrations/node_migrations/sql/postgres/20240613100000_create_database.sql
etorreborre: I went through it manually, also because I had to review the column types which changed for some of them in Postgres. Hopefully it is as correct as the test suite for `xxx_repository_sql` is correct :-).

implementations/rust/ockam/ockam_api/src/ui/output/utils.rs
etorreborre: Thanks. Yes I saw in your recent PRs that there was some support to avoid duplicating single and list items.

implementations/rust/ockam/ockam_api/src/cli_state/storage/vaults_repository_sql.rs
etorreborre: That was done incorrectly and, as a result, all the vaults were marked as default.

SanjoDeundiak (dismissed):

SanjoDeundiak (approved):

PR #8226, opened 2024-06-27
(elixir) use cbor on tcp transport serialization,  introduce support for padding on secure channel

polvorin:
* port tcp transport layer protocol changes to elixir. 
* port secure channel changes


implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/handler.ex
```diff
@@ -28,11 +28,16 @@ defmodule Ockam.Transport.TCP.Handler do
 
     {:ok, socket} = :ranch.handshake(ref, ranch_options)
 
+    # Header, protocol version "1" must be the first thing exchanged.
+    # It isn't send anymore after the initial exchange.
+    {:ok, <<1>>} = transport.recv(socket, 1, 5000)
```
SanjoDeundiak: I wound change the order of these two lines. Imagine the other side doing the same and now both parties are waiting for the version number.

SanjoDeundiak (changes requested):

implementations/elixir/ockam/ockam/lib/ockam/transport/tcp/handler.ex
polvorin: ok, I assumed the connecting side send it, and receiver responded.  But will update

SanjoDeundiak (approved):

PR #8246, opened 2024-07-02
feat(rust): display a malformed transport address

etorreborre:
When an address fails to be resolved we currently have no information about that address.

This PR add a String field to the `InvalidAddress` error in order to report what was attempted to parse.


SanjoDeundiak (dismissed):
image

SanjoDeundiak (approved):

PR #8248, opened 2024-07-02
Using hostname rather than socket addresses

davide-baldo:
Removed every manual hostname resolution except on the Inlet side, where it's resolved once during creation and the result is kept. This addresses a bootstrap failure when creating an outlet for a non-existing hostname, and also dynamically resolves the address at each connection.

Converted all `SocketAddress` usages in command into `HostnamePort`, even in inlets you can specify the local hostname, so for example, you can bind to the local network interface without knowing the IP address.
The previous code did perform some amount of hostname-level validation, but since the validation is hard to get right, (we can't rely on the hostname being an ASCII, since IDN are now a thing, and there are several rules), it's easier to just forward the hostname and fail during connection.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/kafka/inlet_controller.rs
+++ b/implementations/rust/ockam/ockam_api/src/kafka/inlet_controller.rs
@@ -1,7 +1,10 @@
 use core::str::FromStr;
 use minicbor::Decoder;
-use ockam_core::compat::net::IpAddr;
 
+use crate::kafka::kafka_outlet_address;
+use crate::nodes::models::portal::{CreateInlet, InletStatus};
+use crate::nodes::NODEMANAGER_ADDR;
+use crate::port_range::PortRange;
 use ockam::compat::tokio::sync::Mutex;
 use ockam_abac::PolicyExpression;
 use ockam_core::api::{Request, ResponseHeader, Status};
@@ -14,11 +17,7 @@ use ockam_core::{route, Error};
 use ockam_core::{Result, Route};
 use ockam_multiaddr::MultiAddr;
 use ockam_node::Context;
-
-use crate::kafka::kafka_outlet_address;
-use crate::nodes::models::portal::{CreateInlet, InletStatus};
-use crate::nodes::NODEMANAGER_ADDR;
-use crate::port_range::PortRange;
+use ockam_transport_core::HostnamePort;
 
 type BrokerId = i32;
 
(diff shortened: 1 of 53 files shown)


implementations/rust/ockam/ockam_api/src/kafka/inlet_controller.rs
```diff
@@ -33,10 +32,10 @@ pub(crate) struct KafkaInletController {
 
 #[derive(Debug)]
 struct KafkaInletMapInner {
-    broker_map: HashMap<BrokerId, SocketAddr>,
+    broker_map: HashMap<BrokerId, HostnamePort>,
     port_range: PortRange,
     current_port: u16,
-    bind_ip: IpAddr,
+    bind_hostname: String,
```
SanjoDeundiak, line 38: Maybe we could introduce `Hostname` type, and then make it a part of `HostnamePort`. We may do it later though

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_outlets.rs
```diff
@@ -104,9 +104,7 @@ impl NodeManager {
             .await;
 
         info!(
-            "Handling request to create outlet portal at {}:{} with worker {:?}",
-            &hostname_port.hostname(),
-            hostname_port.port(),
+            "Handling request to create outlet portal at {to} with worker {:?}",
```
SanjoDeundiak: "Handling request to create outlet portal to {to} with worker {:?}",

implementations/rust/ockam/ockam_api/src/kafka/inlet_controller.rs
SanjoDeundiak: There is also some overlap with `InternetAddress` type

implementations/rust/ockam/ockam_transport_core/src/hostname_port.rs
```diff
@@ -1,38 +1,58 @@
...
 use core::net::SocketAddr;
 use core::str::FromStr;
 use minicbor::{CborLen, Decode, Encode};
 use ockam_core::compat::format;
 use ockam_core::compat::string::{String, ToString};
 use ockam_core::errcode::{Kind, Origin};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+/// [`HostnamePort`]'s static counterpart usable for const values.
+pub struct StaticHostnamePort {
```
SanjoDeundiak, line 12: I see you picked up that type, do you intend to use it for constant values?

SanjoDeundiak (dismissed):

SanjoDeundiak (approved):

PR #8251, opened 2024-07-02
feat(rust): add healthcheck to `Rendezvous` server

SanjoDeundiak:


implementations/rust/ockam/ockam_command/src/rendezvous_server/start/healthcheck.rs
```diff
@@ -0,0 +1,149 @@
...
+                Origin::Application,
+                Kind::Unknown,
+                "Can't start Healthcheck because handle is present",
+            ));
+        }
+
+        let task = self.task.take().ok_or(Error::new(
+            Origin::Application,
+            Kind::Unknown,
+            "Can't start Healthcheck because handle is present",
```
adrianbenavides: Shouldn't this say something about the task?

implementations/rust/ockam/ockam_command/src/rendezvous_server/start/healthcheck.rs
```diff
@@ -0,0 +1,149 @@
...
+            handle: None,
+        })
+    }
+
+    pub async fn start(&mut self) -> Result<()> {
+        if self.handle.is_some() {
+            return Err(Error::new(
+                Origin::Application,
+                Kind::Unknown,
+                "Can't start Healthcheck because handle is present",
```
adrianbenavides: Does this mean that it can't start because it was already started? If that's the case, I'd change the message to give more higher-level details.

implementations/rust/ockam/ockam_command/src/rendezvous_server/start/healthcheck.rs
```diff
@@ -0,0 +1,149 @@
...
+use ockam_core::errcode::{Kind, Origin};
+use ockam_core::{route, AsyncTryClone, Error, Result};
+use ockam_node::Context;
+use std::net::SocketAddr;
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::net::TcpListener;
+use tokio::task::JoinHandle;
+use tracing::info;
+
+pub struct Healthcheck {
```
adrianbenavides: I'm not 100% convinced that this logic should be in this repo... This kind of logic looks more like the stuff we have in `ockam_api`, and the same thing for the errors. We can come back to this later.

We have a similar thing for the node [here](https://github.com/build-trust/ockam/blob/1d63a294ededc1a5b5498f3cd4d486bfef1702e7/implementations/rust/ockam/ockam_api/src/nodes/service/http.rs#L25), as a reference of what I have in mind.

implementations/rust/ockam/ockam_command/src/rendezvous_server/start/healthcheck.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    pub async fn stop(&mut self) -> Result<()> {
+        if let Some(handle) = self.handle.take() {
+            handle.abort();
+            Ok(())
+        } else {
+            Err(Error::new(
+                Origin::Application,
+                Kind::Unknown,
+                "Can't stop Healthcheck because handle is missing",
```
adrianbenavides: Same for this error message.

implementations/rust/ockam/ockam_command/src/rendezvous_server/start/healthcheck.rs
SanjoDeundiak: I think that's a programmer error if call start twice

implementations/rust/ockam/ockam_command/src/rendezvous_server/start/healthcheck.rs
SanjoDeundiak: Same here

implementations/rust/ockam/ockam_command/src/rendezvous_server/start/healthcheck.rs
SanjoDeundiak: I'm fine with moving that to `ockam_api`, looks like a better place. Does that work?

adrianbenavides: Yes, I think it would fit better there.

adrianbenavides (approved):

PR #8263, opened 2024-07-05
Add udp segmentation

SanjoDeundiak:


implementations/rust/ockam/ockam_transport_udp/src/messages/routing_number.rs
```diff
@@ -0,0 +1,97 @@
...
+        core::fmt::Display::fmt(&self.0, f)
+    }
+}
+
+impl RoutingNumber {
+    pub fn new() -> Self {
+        Self(random())
+    }
+
+    pub fn increment(&mut self) {
```
etorreborre, line 24: I know that this all works ok in the current implementation but I'm slightly uneasy with having the `increment` logic inside the `RoutingNumber`. I'd be more comfortable with:

- Having `RoutingNumber` being immutable.
- Have the `increment` logic as a function `RoutingNumber -> RoutingNumber` inside the `UdpSenderWorker` since this is the only place where this should be called.

etorreborre (approved):
I left one suggestion the `RoutingNumber` and I have a question: what are the drawbacks of handling out-of-order messages directly in the `PendingRoutingMessageStorage` instead of dropping them (by waiting until all of them arrive and cleaning up if this not the case after a while)? Do you another suggestion for how this should be supported?

PR #8266, opened 2024-07-07
ci(rust)::> Github Workflow to verify new crates that are being introduced

### LINKED ISSUE #2526: Create a checklist of things to verify when new crates are introduced.
Two things that were missed in the `ockam_transport_ble` review that we should be sure not to miss for new crates (for example `ockam_transport_smoltcp` in #2213):
1. No `CHANGELOG.md` file. Initially, it should be empty. It's possible this should instead be handled by the changelog generation script.
2. Use of a `path = ...` and no `version = ...` for the dependency on `ockam` (#2525). This failure manifests quite late in the process, and not during a part that handles interruption well so it's pretty important we catch this one in the future.

More broadly, there are certainly other things to check as well. We should make a list of these somewhere. It probably is too rare that we add new crates to be worth automating, but perhaps that's worthwhile too.

PsychoPunkSage:
Closes #2526 

Made a `check_crate.yml` workflow that will be triggered when `push` or `pull` request are made.
It will run simple script at each steps ti check for `Change.log` and `readme.md`

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -97,3 +97,186 @@ jobs:
             git diff;
             exit 1; \
           }
+
+  check_crates:
+    runs-on: ubuntu-latest
+    name: All - check_crates
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y jq
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@88dc2356392166efad76775c878094f4e83ff746
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+
+      - name: Install tomlq
+        run: cargo install --locked tomlq --version 0.1.0
+
+      - name: Check for CHANGELOG.md
+        run: |
+          for crate in $(find implementations/rust/ockam -name Cargo.toml); do
+
+            dir=$(dirname $crate)
+            if [ ! -f "$dir/CHANGELOG.md" ]; then
+              echo "Error: $dir/CHANGELOG.md is missing"
+              exit 1
+            fi
+          done
+
+      - name: Check for README.md
+        run: |
+          for crate in $(find implementations/rust/ockam -name Cargo.toml); do
+
+            dir=$(dirname $crate)
+            if [ ! -f "$dir/README.md" ]; then
+              echo "Error: $dir/README.md is missing"
+              exit 1
+            fi
+          done
+
+      - name: Validate Cargo.toml categories
+        run: |
+          allowed_categories="
+          accessibility
+          aerospace
+          drones
+          protocols
+          simulation
+          space-protocols
+          unmanned-aerial-vehicles
+          algorithms
+          api-bindings
+          asynchronous
+          authentication
+          caching
+          command-line-interface
+          command-line-utilities
+          compilers
+          compression
+          computer-vision
+          concurrency
+          config
+          cryptography
+          cryptocurrencies
+          data-structures
+          database
+          database-implementations
+          date-and-time
+          development-tools
+          build-utils
+          cargo-plugins
+          debugging
+          ffi
+          procedural-macro-helpers
+          profiling
+          testing
+          email
+          embedded
+          emulators
+          encoding
+          external-ffi-bindings
+          filesystem
+          finance
+          game-development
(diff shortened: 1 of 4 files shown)


PsychoPunkSage:
`CONCEPT:`
I fetch all the members from `root Cargo.toml` and use it in my further checks.
Check involves:
1. Presence of `CHANGELOG.md`
2. Presence of `README.md`
3. Whether `Cargo.toml` follows [allowed sets of categories](https://crates.io/category_slugs)

.github/workflows/check_crate.yml
```diff
@@ -0,0 +1,81 @@
...
+          toolchain: stable
+          profile: minimal
+          override: true
+
+      - name: Read member paths from main Cargo.toml
+        id: member_paths
+        run: |
+          paths=$(grep -oP '(?<=members = \[).*(?=\])' Cargo.toml | tr -d '\n')
+          paths=$(echo "$paths" | sed 's/,/\n/g' | tr -d '"')
+          echo "::set-output name=paths::$paths"
```
metaclips: Let's instead get Cargo.toml files in the `implementations/rust/ockam` directory
```bash
paths=$(find implementations/rust/ockam -name Cargo.toml)
```

.github/workflows/check_crate.yml
```diff
@@ -0,0 +1,81 @@
...
+              if [ ! -f "$dir/README.md" ]; then
+                echo "Error: $dir/README.md is missing"
+                exit 1
+              fi
+            done
+          done
+
+      - name: Validate Cargo.toml categories
+        run: |
+          allowed_categories=$(curl -s https://crates.io/category_slugs | jq -r '.[]')
```
metaclips: https://crates.io/category_slugs doesn't support JSON, so we will need to manually get all the categries and hardcode them here
allowed_categories="cryptography networking ....."

.github/workflows/check_crate.yml
```diff
@@ -0,0 +1,81 @@
...
+          paths="${{ steps.member_paths.outputs.paths }}"
+          IFS=$'\n' read -r -d '' -a path_array <<< "$paths"
+          for path in "${path_array[@]}"; do
+            for crate in $(find $path -name 'Cargo.toml'); do
+              categories=$(grep -A 10 '\[package\]' $crate | grep -m 1 'categories' | sed 's/categories = \[\(.*\)\]/\1/' | tr -d '[],' | tr -d ' ' | tr ',' '\n')
```
metaclips: We can get all categories in a toml file using cargo toml, I'll advice to use version 0.1.0 of tomlq
```bash
cargo install --locked tomlq --version 0.1.0
```
Then to get the categories of a Cargo.toml file, you can do
```bash
categories=$(tomlq package.categories -f 'directory_to_Cargo.toml')
```

.github/workflows/check_crate.yml
```diff
@@ -0,0 +1,81 @@
...
+              for category in $categories; do
+                if ! echo "$allowed_categories" | grep -q "^$category$"; then
```
metaclips:
```suggestion
                if ! echo "$allowed_categories" | grep "$category"; then
```

metaclips (changes requested):
Awesome PR @PsychoPunkSage little changes left, and we should merge this PR, also you should make the commit message
`ci: github workflow to verify new crates that are being introduced`
We do not support uppercase in our commit messages, this JS file contains code rules for commits https://github.com/build-trust/ockam/blob/develop/tools/commitlint/commitlint.config.js

.github/workflows/check_crate.yml
PsychoPunkSage: Done

.github/workflows/check_crate.yml
PsychoPunkSage: Done

.github/workflows/check_crate.yml
PsychoPunkSage: Done

.github/workflows/check_crate.yml
PsychoPunkSage: Done

PsychoPunkSage:
@metaclips 
made required changes...
Please have a look at it..

PsychoPunkSage:
Hi @metaclips 
Did you find time to review this PR?

metaclips (approved):
Thanks @PsychoPunkSage this PR found errors in our crates.

adrianbenavides (approved):

PR #8267, opened 2024-07-08
Add the possibility to use a larger timeout for retrieving credentials

etorreborre:
This PR tries to improve the reliability of the Orchestrator command tests:

1. By having the possibility to set higher timeouts when creating secure channels and making requests

2. By allowing an `InMemoryNode` to be restarted in case of a command retry
    - In order to do this we must be able to restart services with fixed address names like the "echoer" service or the `Medic` collector. 
    
3. By returning the last error for a retried command (otherwise some commands are shown as successful when they should fail)

Additionally, I:

 - Added more details to some errors.
 - Made sure that commands are logged when running the CI test suite.


adrianbenavides (approved):
🚀

PR #8269, opened 2024-07-08
rust(test): split orchestrator bats suite

adrianbenavides:
It also updates the `bats/run.sh` script so we can decide what suites to run on different CI jobs:

```
run.sh local
run.sh examples serial
```


etorreborre (approved):
Excellent.

PR #8294, opened 2024-07-11
feat(rust): add CRUD commands to manage space and project admins

polvorin:
Add commands to add/list/delete admins from spaces and projects:
- `ockam space-admin add/list/delete <email>`
- `ockam project-admin add/list/delete <email>`


implementations/rust/ockam/ockam_api/src/cli_state/enrollments.rs
```diff
@@ -78,6 +78,17 @@ impl CliState {
         }
     }
 
+    /// Return enmrollment information of the identity with the given name
```
etorreborre: s/enmrollment/enrollment

implementations/rust/ockam/ockam_api/src/cloud/project/controller_client.rs
```diff
@@ -72,6 +74,126 @@ impl ControllerClient {
...
+                "spaces",
+                Request::put(format!("/v0/{space_id}/admins/{email}")),
+            )
+            .await
+            .into_diagnostic()?
+            .miette_success("add space admins")?;
+        admins
+            .into_iter()
+            .find(|a| a.email == email.to_string())
+            .ok_or(miette!("The email {email} was not added to the space"))
```
etorreborre: Maybe "A user with email {email} was not added to space {space_id}".

But that also begs the question: why? It seems weird to me that the response on a call for adding an admin is the list of admins and not simply a success/error enum.

implementations/rust/ockam/ockam_api/src/cloud/project/controller_client.rs
```diff
@@ -72,6 +74,126 @@ impl ControllerClient {
...
+        let _admins: Vec<AdminInfo> = self
+            .get_secure_client()
+            .ask(
+                ctx,
+                "spaces",
+                Request::delete(format!("/v0/{space_id}/admins/{email}")),
+            )
+            .await
+            .into_diagnostic()?
+            .miette_success("delete space admins")?;
```
etorreborre, line 135: What can of error do we return if the email wasn't found? Can we make it so that it's not a 500?

implementations/rust/ockam/ockam_api/src/cloud/project/controller_client.rs
```diff
@@ -72,6 +74,126 @@ impl ControllerClient {
...
+            .get_secure_client()
+            .ask(
+                ctx,
+                "projects",
+                Request::put(format!("/v0/{project_id}/admins/{email}")),
+            )
+            .await
+            .into_diagnostic()?
+            .miette_success("add project admins")?;
+        admins
```
etorreborre, line 156: Same questions that I had about spaces here.

implementations/rust/ockam/ockam_command/src/project_admin/delete.rs
```diff
@@ -0,0 +1,135 @@
...
+    email: Option<EmailAddress>,
+
+    /// Name of the Project
+    name: Option<String>,
+
+    /// Confirm the deletion without prompting
+    #[arg(long, short)]
+    yes: bool,
+
+    #[arg(long, short)]
```
etorreborre: Let's remove the short option for every `--all` option, to be consistent with `reset`.

etorreborre (dismissed):
I left one or 2 comments about error management that you might to consider.

implementations/rust/ockam/ockam_api/src/cloud/project/controller_client.rs
adrianbenavides: The orchestrator returns the full list of admins in all the endpoints. Changing that is out of scope for me right now, so that's why I added this "normalization" step in the client to return 1/n/empty depending on the endpoint. 

In this case, since the orchestrator returns the whole list, this error shouldn't happen.

implementations/rust/ockam/ockam_api/src/cloud/project/controller_client.rs
adrianbenavides: We would return an error 500 if it fails on our end. Otherwise we have the status provided on the elixir side, but looks like we are not using that when parsing the `MietteError` into another of our error types.

implementations/rust/ockam/ockam_api/src/cloud/project/controller_client.rs
etorreborre: @ckhrysze can you please make a note somewhere to improve this? Does the current API make sense to you?

implementations/rust/ockam/ockam_api/src/cloud/project/controller_client.rs
etorreborre: Maybe this should be part of another PR but right now the output is:
```
❯ ockam project-admin delete me2@email.com


    Error:
    ────────────────────────────────────────────────────────────────────────────────

    The project admin me2@email.com was not found

    Error code: OCK500
    version: 0.129.0
    compiled from: 49252157cb3537a82a0af0d2275bfdff1092c7a1
```
And we shouldn't tell the user to open an issue for this.

implementations/rust/ockam/ockam_command/src/project_admin/delete.rs
```diff
@@ -0,0 +1,135 @@
...
+use ockam_api::colors::color_primary;
+use ockam_api::fmt_ok;
+use ockam_api::nodes::InMemoryNode;
+use ockam_api::terminal::{Terminal, TerminalStream};
+
+/// Delete an Admin from a Project
+#[derive(Clone, Debug, Args)]
+#[command()]
+pub struct DeleteCommand {
+    /// Email of the Admin to add
```
etorreborre: s/to add/to delete

implementations/rust/ockam/ockam_command/src/space_admin/delete.rs
```diff
@@ -0,0 +1,164 @@
...
+use ockam_api::colors::color_primary;
+use ockam_api::nodes::InMemoryNode;
+use ockam_api::terminal::{ConfirmResult, Terminal, TerminalStream};
+use ockam_api::{fmt_ok, fmt_warn};
+
+/// Delete an Admin from a Space
+#[derive(Clone, Debug, Args)]
+#[command()]
+pub struct DeleteCommand {
+    /// Email of the Admin to add
```
etorreborre: s/to add/to delete

implementations/rust/ockam/ockam_api/src/cloud/project/controller_client.rs
adrianbenavides: oooh right. I think this is easier to fix as this error is created [in the comand](https://github.com/build-trust/ockam/blob/49252157cb3537a82a0af0d2275bfdff1092c7a1/implementations/rust/ockam/ockam_command/src/terminal/tui.rs#L175). Let me have a look.

adrianbenavides: There is a tradeoff here due to the error issue... Because of the way we bubble up errors (plain miette errors) we can't say what kind of error we had, so 99% of the times we are converting the error into a code 500. If we want to return a plain message without the error template, then we would have to return an Ok(), which would be wrong from the command perspective (e.g. when using the command in a script we check the exit code).

I wouldn't change anything for now until we refactor the error handling.

etorreborre: Makes sense.

etorreborre (approved):

PR #8296, opened 2024-07-11
feat(rust): integrate space's subscription data in command

adrianbenavides:
Adds the subscription data to the `space`, `status` and `enroll` commands. This PR also improves the output construction of those commands, which led me to improve the `Output` trait.

## Changes to the `Output` trait

In this PR you can see that the `Subscription` struct is used in three different kinds of outputs (see images below for examples):

1. As a standalone item (it's individual representation)
2. As part of other structs in their individual representation (e.g. in the `space show` command)
3. As part of other structs in their list representation (e.g. in the `space list` command)

Now, we want to reuse the output of the individual representation, but there are some differences we need to keep in mind: in `2.` the subscription is padded + indented; in `3.` the subscription is padded.

The `Output` trait lets you define the output for `1.` and `3.` but not for `2.` and, in this case, the only difference between them is the left spacing. And I think this is also true for a lot of other structs. 

That being said, I've extended the `Output` trait a little to make these "left-spacing" transformations easier. You can check the [`Subscription`](https://github.com/build-trust/ockam/pull/8296/files#diff-8d38b9822d9dba78bcd7e2fe4b524aa028ef0d7872d598270242522a397c0f89R85-R121) struct as an example of how to reuse the `Display` representation and how it's also integrated in the [`Space`'s `Display` impl](https://github.com/build-trust/ockam/pull/8296/files#diff-ac92631eb44c2b0ea45d6c7ef8bb4dec67639f735f6a46d708a61b483125d2feR64).

## `enroll`

Free trial + and the space has been just created:

image

Free trial + the space already existed:

image

Not a free trial subscription:

image

## `space` and `status`

image

image

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/storage/spaces_repository_sql.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/storage/spaces_repository_sql.rs
@@ -1,13 +1,12 @@
-use sqlx::any::AnyRow;
-use sqlx::*;
-
+use super::SpacesRepository;
+use crate::cloud::space::Space;
+use crate::cloud::subscription::Subscription;
 use ockam_core::async_trait;
 use ockam_core::Result;
-use ockam_node::database::{Boolean, FromSqlxError, SqlxDatabase, ToVoid};
-
-use crate::cloud::space::Space;
-
-use super::SpacesRepository;
+use ockam_node::database::{Boolean, FromSqlxError, Nullable, SqlxDatabase, ToVoid};
+use sqlx::any::AnyRow;
+use sqlx::*;
+use time::OffsetDateTime;
 
 #[derive(Clone)]
 pub struct SpacesSqlxDatabase {
@@ -24,6 +23,15 @@ impl SpacesSqlxDatabase {
     /// Create a new in-memory database
     pub async fn create() -> Result<Self> {
         Ok(Self::new(SqlxDatabase::in_memory("spaces").await?))
+    }
+
+    async fn query_subscription(&self, space_id: &str) -> Result<Option<Subscription>> {
+        let query = query_as("SELECT space_id, name, is_free_trial, marketplace, start_date, end_date FROM subscription WHERE space_id = $1").bind(space_id);
+        let row: Option<SubscriptionRow> = query
+            .fetch_optional(&*self.database.pool)
+            .await
+            .into_core()?;
+        Ok(row.map(|r| r.subscription()))
     }
 }
 
@@ -69,6 +77,31 @@ impl SpacesRepository for SpacesSqlxDatabase {
             query4.execute(&mut *transaction).await.void()?;
         }
 
+        // store the subscription if any
+        if let Some(subscription) = &space.subscription {
+            let start_date = subscription.start_date();
+            let end_date = subscription.end_date();
+            let query = query(
+                r"
+             INSERT INTO subscription (space_id, name, is_free_trial, marketplace, start_date, end_date)
+             VALUES ($1, $2, $3, $4, $5, $6)
+             ON CONFLICT (space_id)
+             DO UPDATE SET space_id = $1, name = $2, is_free_trial = $3, marketplace = $4, start_date = $5, end_date = $6",
+            )
+                .bind(&space.id)
+                .bind(&subscription.name)
+                .bind(subscription.is_free_trial)
+                .bind(&subscription.marketplace)
+                .bind(start_date.map(|d| d.unix_timestamp()))
+                .bind(end_date.map(|d| d.unix_timestamp()));
+            query.execute(&mut *transaction).await.void()?;
+        }
+        // remove the subscription
+        else {
+            let query = query("DELETE FROM subscription WHERE space_id = $1").bind(&space.id);
+            query.execute(&mut *transaction).await.void()?;
+        }
+
         transaction.commit().await.void()
     }
 
@@ -95,6 +128,7 @@ impl SpacesRepository for SpacesSqlxDatabase {
         let row: Option<SpaceRow> = query1.fetch_optional(&mut *transaction).await.into_core()?;
         let space = match row.map(|r| r.space()) {
             Some(mut space) => {
+                // retrieve the users
                 let query2 =
                     query_as("SELECT space_id, user_email FROM user_space WHERE space_id = $1")
                         .bind(&space.id);
@@ -102,6 +136,10 @@ impl SpacesRepository for SpacesSqlxDatabase {
                     query2.fetch_all(&mut *transaction).await.into_core()?;
                 let users = rows.into_iter().map(|r| r.user_email).collect();
                 space.users = users;
+
+                // retrieve the subscription
+                space.subscription = self.query_subscription(&space.id).await?;
+
                 Some(space)
             }
             None => None,
@@ -114,16 +152,20 @@ impl SpacesRepository for SpacesSqlxDatabase {
         let mut transaction = self.database.begin().await.into_core()?;
 
         let query = query_as("SELECT space_id, space_name FROM space");
-        let row: Vec<SpaceRow> = query.fetch_all(&mut *transaction).await.into_core()?;
+        let rows: Vec<SpaceRow> = query.fetch_all(&mut *transaction).await.into_core()?;
 
         let mut spaces = vec![];
-        for space_row in row {
+        for row in rows {
             let query2 =
                 query_as("SELECT space_id, user_email FROM user_space WHERE space_id = $1")
-                    .bind(&space_row.space_id);
+                    .bind(&row.space_id);
             let rows: Vec<UserSpaceRow> = query2.fetch_all(&mut *transaction).await.into_core()?;
             let users = rows.into_iter().map(|r| r.user_email).collect();
-            spaces.push(space_row.space_with_user_emails(users))
+            let subscription = self.query_subscription(&row.space_id).await?;
+            let mut space = row.space();
+            space.users = users;
+            space.subscription = subscription;
+            spaces.push(space);
         }
 
         transaction.commit().await.void()?;
@@ -169,6 +211,9 @@ impl SpacesRepository for SpacesSqlxDatabase {
         let query2 = query("DELETE FROM user_space WHERE space_id = $1").bind(space_id);
         query2.execute(&mut *transaction).await.void()?;
 
+        let query3 = query("DELETE FROM subscription WHERE space_id = $1").bind(space_id);
+        query3.execute(&mut *transaction).await.void()?;
+
         transaction.commit().await.void()
     }
 }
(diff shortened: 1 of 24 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/storage/spaces_repository_sql.rs
```diff
@@ -42,14 +42,20 @@ impl SpacesRepository for SpacesSqlxDatabase {
...
+             VALUES ($1, $2, $3, $4)
              ON CONFLICT (space_id)
-             DO UPDATE SET space_name = $2, is_default = $3"#,
+             DO UPDATE SET space_name = $2, is_default = $3, subscription = $4"#,
         )
         .bind(&space.id)
         .bind(&space.name)
-        .bind(is_already_default);
+        .bind(is_already_default)
+        .bind(
```
etorreborre: I wonder if we should create a separate table for subscriptions. I'm afraid of some silent breakage scenarios that we might not catch in time. For example if we add a new field to subscriptions and we forget to a Rust migration, nothing will break at first. Then a user goes `ockam space list` and gets a runtime error.

This is more work but we might be possibly better off with an additional table.

implementations/rust/ockam/ockam_api/src/cloud/subscription.rs
```diff
@@ -50,14 +53,80 @@ impl ActivateSubscription {
...
 #[cbor(map)]
 pub struct Subscription {
+    #[n(1)]
+    pub name: String,
+    #[n(2)]
+    pub is_free_trial: bool,
+    #[n(3)]
+    pub marketplace: Option<String>,
+    #[n(4)]
+    pub start_date: Option<String>,
```
etorreborre: Is there no way to deserialize this field directly as `start_date: Option<OffsetDateTime>`?
Also, especially if the internal representation (`String`) is different than the external one (`OffsetDateTime`), it's probably better to leave this field private.

implementations/rust/ockam/ockam_api/src/cloud/subscription.rs
```diff
@@ -50,14 +53,80 @@ impl ActivateSubscription {
...
+
+impl Output for Subscription {
+    fn item(&self) -> crate::Result<String> {
+        Ok(self.padded_display()?)
+    }
+}
+
+#[derive(Encode, Decode, CborLen, Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+#[cbor(map)]
+pub struct SubscriptionLegacy {
```
etorreborre, line 165: It would be good to have comment here explaining:

 - Why it's legacy.
 - When we can eventually remove it.

implementations/rust/ockam/ockam_api/src/date.rs
```diff
@@ -0,0 +1,21 @@
+use crate::ApiError;
+use time::format_description::well_known::Iso8601;
+use time::OffsetDateTime;
+
+/// Transform a string that represents an Iso8601 date into a `time::OffsetDateTime`
```
etorreborre, line 5: This is kind of problematic. We should fix this on the Elixir side and have it send properly formatted dates (with a `Z`) so that we make sure we're always talking in UTC dates. 
Then have the client do a UTC->LocalTime transformation for display.

implementations/rust/ockam/ockam_api/src/ui/output/mod.rs
```diff
@@ -37,10 +39,34 @@ use std::fmt::Write;
...
+    /// Adds padding to each line of the Display output
+    fn padded_display(&self) -> std::result::Result<String, Error>
+    where
+        Self: Display,
+    {
+        Ok(self.iter_output()?.pad().to_string())
+    }
+
+    /// Returns an iterator over the lines of the Display output
+    fn iter_output(&self) -> std::result::Result<OutputIter, Error>
```
etorreborre: There are two things that I find a bit troubling:

 - The fact that we need to declare that output functions can throw errors. It seems to me that as long as I have correct data, I should always be able to display it.
 
 - The fact that we require different error types for different methods of the trait. It looks like this is due to the fact that we're trying to reuse some `Output` code to implement some `Display` instance. I think we should consider one being lower level than the other and use only code in one direction (like `Display` inside `Output` but not the opposite).

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -463,12 +463,32 @@ async fn get_user_space(
...
+        opts.terminal.write_line(fmt_log!(
+            "This Space has a {} Subscription attached to it.",
+            color_primary(&subscription.name)
+        ))?;
+        if subscription.is_free_trial {
+            if is_new {
+                opts.terminal.write_line("")?
+                    .write_line(fmt_log!("As a courtesy, we created a temporary Space for you, so you can continue to build.\n"))?
+                    .write_line(fmt_log!("Please subscribe to an Ockam plan within two weeks {}", color_uri("https://www.ockam.io/pricing")))?
+                    .write_line(fmt_log!("{}", "If you don't subscribe in that time, your Space and all Projects will be permanently deleted.".color(OckamColor::FmtWARNBackground.color())))?;
```
etorreborre: maybe "and all associated Projects"?

implementations/rust/ockam/ockam_command/src/enroll/command.rs
```diff
@@ -463,12 +463,32 @@ async fn get_user_space(
...
+            {
+                opts.terminal.write_line("")?
+                    .write_line(fmt_log!("Your free trial started on {} and will end on {}.\n", start_date, end_date))?
+                    .write_line(fmt_log!("Please subscribe to an Ockam plan before the trial ends to avoid any service interruptions {}", color_uri("https://www.ockam.io/pricing")))?
+                    .write_line(fmt_log!("{}", "If you don't subscribe in that time, your Space and all Projects will be permanently deleted.".color(OckamColor::FmtWARNBackground.color())))?;
+            }
+        }
+    } else {
+        opts.terminal.write_line(fmt_log!(
+            "This Space does not have a Subscription attached to it."
```
etorreborre: Any consequence? Should we tell the user how much cooler it would be to have a subscription attached to it?

implementations/rust/ockam/ockam_command/src/space/create.rs
```diff
@@ -35,61 +36,55 @@ pub struct CreateCommand {
...
                 ctx,
                 &self.name,
                 self.admins.iter().map(|a| a.as_ref()).collect(),
             )
-            .await?;
-
+            .await?
+        };
+        if space.is_in_free_trial_subscription() {
+            opts.terminal.write_line(fmt_log!("This Space does not have a Subscription attached to it."))?
```
etorreborre: Maybe we should factor out this code with [this one](https://github.com/build-trust/ockam/pull/8296/files#diff-23aa36ffad98b3c4e3a6e17876f299e2bf29ec2000af3cb84bcc9687a61e500cR480). This will help when tweaking the message and make sure that we call `color_warn` more consistently.

implementations/rust/ockam/ockam_command/src/space/create.rs
```diff
@@ -35,61 +36,55 @@ pub struct CreateCommand {
...
+                .write_line(fmt_log!("As a courtesy, we created a temporary Space for you, so you can continue to build.\n"))?
+                .write_line(fmt_log!("Please subscribe to an Ockam plan within two weeks {}", color_uri("https://www.ockam.io/pricing")))?
+                .write_line(fmt_log!("{}\n", color_warn("If you don't subscribe in that time, your Space and all Projects will be permanently deleted.")))?;
+        }
         opts.terminal
             .stdout()
             .plain(space.item()?)
-            .json(serde_json::json!(&space))
+            .json_obj(&space)?
```
etorreborre, line 72: Nice shortcut.

implementations/rust/ockam/ockam_command/src/status.rs
```diff
@@ -93,18 +94,21 @@ impl StatusCommand {
 struct StatusData {
     #[serde(flatten)]
     orchestrator_version: OrchestratorVersionInfo,
+    spaces: Vec<Space>,
     identities: Vec<IdentityEnrollment>,
     nodes: Vec<NodeResources>,
 }
 
 impl StatusData {
```
etorreborre, line 103: Putting the `ockam` version (+ hash) in there could be a good idea too.

etorreborre (commented):
Some of my comments can definitely be addressed in other PRs. My biggest concern is with inlining the whole `Subscription` data into one row cell.

implementations/rust/ockam/ockam_api/src/date.rs
adrianbenavides: +1 to that, I have created a story on my board to keep track of this.

implementations/rust/ockam/ockam_command/src/space/create.rs
adrianbenavides: It's half-way there, but the idea is to remove the current `json` method and rename the `json_obj` to `json`. But to do that, we first need to have properly formatted outputs in every command.

implementations/rust/ockam/ockam_api/src/cloud/subscription.rs
adrianbenavides: > Is there no way to deserialize this field directly as start_date: Option<OffsetDateTime>?

Probably, but maybe we should wait until we fix the elixir side to return dates. For now, I've made the date fields private.

implementations/rust/ockam/ockam_command/src/enroll/command.rs
adrianbenavides: > Any consequence?

For now, I'm not sure if this code is reachable, as the user should have at least the "developer-free" subscription. This is still in the works on the elixir side, so we might want to wait until that's ready.

implementations/rust/ockam/ockam_api/src/ui/output/mod.rs
adrianbenavides: > The fact that we need to declare that output functions can throw errors. It seems to me that as long as I have correct data, I should always be able to display it.

There could be cases where we need to do some data transformation to build the output. I'm not sure if we have a case like that, but I didn't want to close that door when I first introduced this trait.

> The fact that we require different error types for different methods of the trait....

Right, those functions don't need to return a result. It's a leftover from a previous prototype.

implementations/rust/ockam/ockam_api/src/cli_state/storage/spaces_repository_sql.rs
adrianbenavides: Agree, I'll extract the subscription into its own table.

implementations/rust/ockam/ockam_command/src/space/create.rs
adrianbenavides: done!

implementations/rust/ockam/ockam_api/src/ui/output/mod.rs
etorreborre: > There could be cases where we need to do some data transformation to build the output. I'm not sure if we have a case like that, but I didn't want to close that door when I first introduced this trait.

I still don't think that this should not be an error. I think I'd rather see a partial output than getting an Error 500 because a field can not be transformed into the ideal format.

etorreborre (approved):
Thanks for all the changes 🏵️

Issue #3005, opened 2022-07-21
Improve tests on examples/rust/*


neil2468:
Here are some opportunities...

1. The tests in `examples/rust/get_started` do not cover significant portions of code. 
2. Some tests in `examples/rust/*` could still clash on TCP ports when run in parallel. Maybe they should ask the OS for free ports, rather then using hardcoded values (could use [portpicker](https://crates.io/crates/portpicker)?). 
3. `example_test_helper` could allow child processes to inherit environment variables from their parent. This _might_ allow child `cargo` processes to use the mold linker in CI.
4. Maybe tests in `examples/rust/*` should use the the `cargo` `--locked` argument as is done in other parts of the Ockam code base (e.g. [here](https://github.com/build-trust/ockam/blob/f6eb4126cd4ac5f7c265118ac580b70f6eea05fb/implementations/rust/build.gradle#L88)).

---

<i>We love helping new contributors! ❤️
If you have questions or need help as you explore, please join us on [Discord](https://discord.gg/bewvnm6zqS). If you're looking for other issues to contribute to, please checkout our [good first issues](https://github.com/build-trust/ockam/labels/good%20first%20issue).</i>

PsychoPunkSage:
Hi @neil2468 
Can I work on this issue. it seems no one is working on it. Can I have a look at this?

neil2468:
Hi. Sure, that’s fine by me.

PsychoPunkSage:
Hi,
I'm planning to tackle this issue in 4 PRs. 1 PR for each pointer.
As it will allow me keep track of things that are completed and things that are left. Plus it will also motivate me to see some PRs made by me got merged.
Can I proceed in this way or should I do something else?
Thank you

adrianbenavides:
@PsychoPunkSage Absolutely, that makes total sense, each point is worth its own PR.

PR #8304, opened 2024-07-12
chore(rust)::> used portpicker crate to avoid port clash in ockam examples

PsychoPunkSage:
## Current behavior
Some tests in `examples/rust/*` could clash on TCP ports when run in parallel.

## Proposed changes
Use [`portpicker`](https://crates.io/crates/portpicker) crate to avoid port clash. Now tests will ask the OS for free ports, rather then using hardcoded values.

This PR addresses #3005 **pointer 2**

### DIFF
--- a/tools/docs/example_test_helper/src/lib.rs
+++ b/tools/docs/example_test_helper/src/lib.rs
@@ -46,6 +46,7 @@ use cfg_if::cfg_if;
 use duct::unix::HandleExt;
 use duct::{cmd, Handle};
 use regex::Regex;
+use std::net::TcpListener;
 use std::time::SystemTime;
 use std::{fs, io};
 use std::{fs::File, io::Read, thread::sleep, time::Duration};
@@ -324,6 +325,12 @@ impl CmdRunner {
         }
         Err(Error::Timeout)
     }
+}
+
+pub fn find_available_port() -> u16 {
+    let listener = TcpListener::bind("127.0.0.1:0").expect("Failed to bind to address");
+    let address = listener.local_addr().expect("Failed to get local address");
+    address.port()
 }
 
 #[cfg(test)]
--- a/examples/rust/tcp_inlet_and_outlet/tests/tests.rs
+++ b/examples/rust/tcp_inlet_and_outlet/tests/tests.rs
@@ -1,10 +1,8 @@
-use example_test_helper::{CmdBuilder, Error};
-use ockam::compat::rand;
-use ockam::compat::rand::Rng;
+use example_test_helper::{find_available_port, CmdBuilder, Error};
 
 #[test]
 fn run_01_inlet_outlet_one_process() -> Result<(), Error> {
-    let port = rand::thread_rng().gen_range(10000..65535);
+    let port = find_available_port();
     // Spawn example, wait for it to start up
     let runner = CmdBuilder::new(&format!(
         "cargo run --locked --example 01-inlet-outlet 127.0.0.1:{port} ockam.io:80"
@@ -24,8 +22,8 @@ fn run_01_inlet_outlet_one_process() -> Result<(), Error> {
 
 #[test]
 fn run_02_inlet_outlet_separate_processes() -> Result<(), Error> {
-    let routing_port = rand::thread_rng().gen_range(10000..65535);
-    let inlet_port = rand::thread_rng().gen_range(10000..65535);
+    let routing_port = find_available_port();
+    let inlet_port = find_available_port();
     // Spawn outlet, wait for it to start up
     let outlet = CmdBuilder::new(&format!(
         "cargo run --locked --example 02-outlet ockam.io:80 {routing_port}"
@@ -55,8 +53,8 @@ fn run_02_inlet_outlet_separate_processes() -> Result<(), Error> {
 #[test]
 #[ignore]
 fn run_03_inlet_outlet_separate_processes_secure_channel() -> Result<(), Error> {
-    let routing_port = rand::thread_rng().gen_range(10000..65535);
-    let inlet_port = rand::thread_rng().gen_range(10000..65535);
+    let routing_port = find_available_port();
+    let inlet_port = find_available_port();
     // Spawn outlet, wait for it to start up
     let outlet = CmdBuilder::new(&format!(
         "cargo run --locked --example 03-outlet ockam.io:80 {routing_port}"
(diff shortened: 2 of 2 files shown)


examples/rust/tcp_inlet_and_outlet/Cargo.toml
```diff
@@ -9,6 +9,7 @@ rust-version = "1.56.0"
 
 [dependencies]
 ockam = { path = "../../../implementations/rust/ockam/ockam" }
+portpicker = "0.1.1"
```
metaclips: Portpicker was last contributed to 3years ago, which means its unmaintained, can we instead copy some of the functions to the example code?

metaclips (changes requested):
can we update the commit message to `chore(rust): used portpicker crate to avoid port clash in ockam examples`

examples/rust/tcp_inlet_and_outlet/Cargo.toml
PsychoPunkSage: Oh ok,
I didn't see that...
I will search for some alternatives otherwise I will see which functions can be used from `portpicker`

adrianbenavides: There are some places where we do this without any library. Check [this](https://github.com/build-trust/ockam/blob/18929d337638ab54f063ce2c84b1a5750762cd54/implementations/rust/ockam/ockam_command/src/util/mod.rs#L123) out as an example.

PsychoPunkSage:
Hi @metaclips 
I have replaced `portpicker` crate with the solution pointed by @adrianbenavides 
I have on thing to report...
`run_04_inlet_outlet_seperate_processes_secure_channel_via_ockam_hub` is failing... due to some reasons... (not because of newly implemented `portfinder.rs`, i have verified that)

examples/rust/tcp_inlet_and_outlet/Cargo.toml
PsychoPunkSage: Thank you @adrianbenavides 
I used the soln you provided.

Resolved

examples/rust/tcp_inlet_and_outlet/tests/portfinder.rs
```diff
@@ -0,0 +1,15 @@
+use miette::{self, Context, IntoDiagnostic};
+use std::net::TcpListener;
+
+pub type Result<T> = miette::Result<T>;
+
+pub fn find_available_port() -> Result<u16> {
```
adrianbenavides: You can return the same `Result` type as the other code in the examples. If you do that, you can then remove the miette dependency, as it's not really needed in this case.

adrianbenavides: Also, you could move this function to `/Users/adrian/projects/ockam/ockam/tools/docs/example_test_helper/src/lib.rs`, which acts like a kind of "utils" file that most of the examples use.

PsychoPunkSage:
Hi @adrianbenavides 
Did the requested changes.
Please have a look at it.

tools/docs/example_test_helper/src/lib.rs
```diff
@@ -79,6 +80,10 @@ pub enum Error {
     ExitedCmd,
     #[error("failed to create temporary file for stdout")]
     StdoutTempfile,
+    #[error("failed to bind with given address")]
+    BindError(io::Error),
+    #[error("failed to get local socket address of the listener")]
+    AddrError(io::Error),
```
adrianbenavides: You can use the `Io` error instead. That way we don't lose the trace of the error and we can print/log the full details.

PsychoPunkSage: Done

PsychoPunkSage:
Hi @adrianbenavides 
Did the changes... but should we keep the custom error variants `BindError` and `AddrError` but modify them to include io::Error?? 
like this
```rust
pub enum Error {
......
    #[error("failed to bind with given address")]
    BindError(#[source] io::Error),
    #[error("failed to get local socket address of the listener")]
    AddrError(#[source] io::Error),
}

pub fn find_available_port() -> Result<u16> {
    let listener = TcpListener::bind("127.0.0.1:0").map_err(Error::BindError)?;
    let address = listener.local_addr().map_err(Error::AddrError)?;
    Ok(address.port())
}
```

adrianbenavides:
> Hi @adrianbenavides Did the changes... but should we keep the custom error variants `BindError` and `AddrError` but modify them to include io::Error?? like this
> 
> ```rust
> pub enum Error {
> ......
>     #[error("failed to bind with given address")]
>     BindError(#[source] io::Error),
>     #[error("failed to get local socket address of the listener")]
>     AddrError(#[source] io::Error),
> }
> 
> pub fn find_available_port() -> Result<u16> {
>     let listener = TcpListener::bind("127.0.0.1:0").map_err(Error::BindError)?;
>     let address = listener.local_addr().map_err(Error::AddrError)?;
>     Ok(address.port())
> }
> ```

It's not necessary in this case. Generally, you don't want to map every possible error into an enum variant, just the ones you want to act upon. In this case, we just want to convert external error types into our own and print them.

tools/docs/example_test_helper/src/lib.rs
```diff
@@ -326,6 +327,12 @@ impl CmdRunner {
     }
 }
 
+pub fn find_available_port() -> Result<u16> {
+    let listener = TcpListener::bind("127.0.0.1:0")?;
+    let address = listener.local_addr()?;
+    Ok(address.port())
+}
```
adrianbenavides: An optional improvement: since we are using this function exclusively in tests, you can move the `.expect()` calls here and return a `u16`, so that, in the tests, you can just call the function and not duplicate the `.expect("Failed to find available port");` piece.

adrianbenavides:
Thanks @PsychoPunkSage for this PR 🙏  It looks good, but you'll also have to sign your commits here before I can approve it.

https://docs.github.com/en/authentication/managing-commit-signature-verification

tools/docs/example_test_helper/src/lib.rs
PsychoPunkSage: Done

PsychoPunkSage:
Hi @adrianbenavides 
Added `Signed` commits that you requested

adrianbenavides (approved):
Looks good to merge, thanks a lot @PsychoPunkSage for your PR!! 🙏

etorreborre (approved):

metaclips (approved):

PR #8305, opened 2024-07-12
feat(rust): increase opentelemetry queue sizes

etorreborre:


davide-baldo (approved):
Validated using a load simulator

PR #8306, opened 2024-07-12
`UDP` puncture improvements

SanjoDeundiak:
### Changes
1. Rename `Medic` -> `Session`.
2. Introduce a notion of additional session: the only example for that is UDP punctured connection for TCP inlets. We need to monitor and keep alive both main connection through a relay and UDP punctured connection. Both connections are related to the same session.
3. Switch from one monitoring task for all session to:
   - Isolated tasks for each session
   - One task to send pings and recreate session (+1 for additional session)
   - One task to receive pongs (+1 for additional session)
   - One Collector worker to receive pings (+1 for additional session)
4. Fix flow controls and access controls for UDP puncture.
5. Add a lot of tests for Session.
6. Fix lifecycle of the `Connection` for Inlet and Relay session replacers (old connection wasn't cleaned before starting the new one).
7. Fix circular dependencies with `Arc` that would cause `NodeManager` to never be freed.
8. Implement additional session support for Inlets to properly support UDP puncture connection.
9. Introduced a notion of "recreating" session, which allows to avoid recreating an Inlet if connection is lost and instead just update the route to the outlet when connection is recreated.


implementations/rust/ockam/ockam/src/remote/worker.rs
```diff
@@ -52,20 +42,20 @@ impl Worker for RemoteRelay {
                     debug!("RemoteRelay received service message");
 
                     let payload = String::decode(local_message.payload_ref())
-                        .map_err(|_| OckamError::InvalidHubResponse)?;
+                        .map_err(|_| OckamError::InvalidOrchestratorResponse)?;
```
davide-baldo: The communication could also be between 2 rust nodes, `InvalidOrchestratorResponse` could be misleading.

implementations/rust/ockam/ockam_api/src/nodes/service/relay.rs
```diff
@@ -292,8 +312,15 @@ impl SessionReplacer for RelaySessionReplacer {
     }
 
     async fn close(&mut self) {
+        let node_manager = if let Some(node_manager) = self.node_manager.upgrade() {
+            node_manager
+        } else {
+            error!("Node manager is dropped. Can't close the Relay.");
```
davide-baldo: Maybe we could log something closer to the user, what about
```
warn!("A relay close was issued after node manager shut down, skipping.")
```

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/inlets_trait.rs
```diff
@@ -0,0 +1,33 @@
...
+use ockam_node::Context;
+use ockam_transport_core::HostnamePort;
+use std::time::Duration;
+
+use crate::nodes::models::portal::InletStatus;
+
+#[async_trait]
+pub trait Inlets {
+    #[allow(clippy::too_many_arguments)]
+    async fn create_inlet(
```
davide-baldo, line 15: The number of parameters for inlets is quickly getting out of hand, we should consider using `TcpInletOptions` all the way.

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer.rs
```diff
@@ -0,0 +1,390 @@
...
+
+    pub(super) connection: Option<Connection>,
+
+    pub(super) additional_secure_channel: Option<SecureChannel>,
+    pub(super) udp_puncture: Option<UdpPuncture>,
+    pub(super) additional_route: Option<Route>,
+}
+
+impl InletSessionReplacer {
+    fn enable_udp_puncture(&self) -> bool {
```
davide-baldo: The method name seems to enable the UDP puncture, but it's reading it instead.
What about `udp_puncture_enabled()` instead?

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer.rs
```diff
@@ -0,0 +1,390 @@
...
+        let main_route: Route = main_route.modify().pop_back().into();
+
+        let additional_sc_route =
+            route![main_route.clone(), DefaultAddress::SECURE_CHANNEL_LISTENER];
+
+        let additional_sc = node_manager
+            .create_secure_channel_internal(
+                &self.context,
+                additional_sc_route,
+                &node_manager.identifier(),
```
davide-baldo: Why it isn't using `self.secure_channel_identifier`?

implementations/rust/ockam/ockam_api/src/session/connection_status.rs
```diff
@@ -0,0 +1,45 @@
...
+use crate::colors::{color_error, color_ok};
+use crate::error::ApiError;
+
+use ockam_core::Result;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Encode, Decode, CborLen, Serialize, Deserialize)]
+#[rustfmt::skip]
+pub enum ConnectionStatus {
+    #[n(0)] Down,
+    #[n(1)] Up,
```
davide-baldo, line 15: Finally! `degraded` was obliterated 😄

implementations/rust/ockam/ockam_api/src/session/ping.rs
```diff
@@ -0,0 +1,34 @@
+use core::fmt;
+use minicbor::{CborLen, Decode, Encode};
+use ockam_core::{Decodable, Encodable, Error, Message, Result};
+use rand::random;
+
+#[derive(Debug, Default, Copy, Clone, Encode, Decode, CborLen, PartialEq, Eq)]
+#[cbor(transparent)]
+pub struct Ping(#[n(0)] u64);
```
davide-baldo, line 8: Do we really a payload in the ping message?

implementations/rust/ockam/ockam_api/src/session/session.rs
```diff
@@ -0,0 +1,656 @@
+use rand::random;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Mutex as SyncMutex;
```
davide-baldo, line 3: like it, we should always alias mutex 👍

implementations/rust/ockam/ockam_api/src/session/session.rs
```diff
@@ -0,0 +1,656 @@
...
+                        Err(err) => {
+                            error!(key = %key, err = %err, "failed to send additional ping")
+                        }
+                    }
+
+                    drop(pings);
+
+                    sleep(ping_interval).await;
+                }
+                _ => {
```
davide-baldo, line 591: `Down` instead of `_` should be more readable

implementations/rust/ockam/ockam_api/src/session/session.rs
```diff
@@ -0,0 +1,656 @@
...
+                            trace!(key = %key, "sent additional ping")
+                        }
+                        Err(err) => {
+                            error!(key = %key, err = %err, "failed to send additional ping")
+                        }
+                    }
+
+                    drop(pings);
+
+                    sleep(ping_interval).await;
```
davide-baldo, line 589: It did happen to me that after dropping a mutex guard exactly like this, it was still retained. I am not sure if it was a rust issue on a specific version or a deeper mechanism I'm not aware of.

implementations/rust/ockam/ockam_api/src/session/session.rs
```diff
@@ -0,0 +1,656 @@
...
+                        Err(err) => {
+                            error!(key = %key, err = %err, "failed to send ping")
+                        }
+                    }
+
+                    drop(pings);
+
+                    sleep(ping_interval).await;
+                }
+                _ => {
```
davide-baldo, line 493: `Down` instead of `_` should be more readable

implementations/rust/ockam/ockam_api/src/session/session.rs
```diff
@@ -0,0 +1,656 @@
...
+                    shared_state
+                        .is_being_replaced
+                        .store(true, Ordering::Relaxed);
+
+                    let res = shared_state.replacer.lock().await.create().await;
+
+                    match res {
+                        Ok(replacer_outcome) => {
+                            info!(key = %key, ping_route = %replacer_outcome.ping_route, "replacement is up");
+                            pings.clear();
```
davide-baldo: Pings loose meaning the moment we decide to replace the session, there shouldn't be any in this instant

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer.rs
```diff
@@ -0,0 +1,390 @@
...
+            // Switch Inlet to the main route
+            if let (Some(inlet), Some(main_route)) = (self.inlet.as_ref(), self.main_route.as_ref())
+            {
+                let res = inlet.unpause(main_route.clone());
+
+                if let Some(err) = res.err() {
+                    error!("Error switching Inlet to the main route {}", err);
+                }
+            }
+        }
```
davide-baldo, line 377: Shouldn't we call `inlet.pause()` when no fallback is available?

implementations/rust/ockam/ockam_api/src/session/session.rs
```diff
@@ -0,0 +1,656 @@
...
+    }
+
+    async fn wait_for_pings(
+        key: String,
+        mut pong_receiver: mpsc::Receiver<Ping>,
+        pings: Arc<AsyncMutex<Vec<Ping>>>,
+    ) {
+        while let Some(ping) = pong_receiver.recv().await {
+            let mut pings_guard = pings.lock().await;
+            if pings_guard.contains(&ping) {
```
davide-baldo, line 648: What's the purpose of keeping track of specific pings?

davide-baldo (dismissed):
If I understood correctly, the `pause()` and `unpause()` are modifying the behavior for new connections but not existing one, is that correct?

implementations/rust/ockam/ockam_api/src/session/session.rs
SanjoDeundiak: That would not be the same. The first match arm runs if the status is Up, AND number of pending pings is < 3. Therefore, the other arm runs if status is Down, OR (status is Up AND number of pending pings is >= 3)

implementations/rust/ockam/ockam_api/src/session/session.rs
SanjoDeundiak: That would be a problem... Do you know how to reproduce that? I created a simple example and it works as I would expect https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9187a7714c7c2d1dac5610e077061b70

implementations/rust/ockam/ockam_api/src/session/session.rs
SanjoDeundiak: Could you please elaborate?

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer.rs
SanjoDeundiak: That should result in more responsive behaviour during downtime, good suggestion

implementations/rust/ockam/ockam_api/src/session/session.rs
SanjoDeundiak: Imagine you sent pings: ping1, ping2, ping3, then you received pong3 and cleared pings vector. Then you send ping3 and ping4, and received pong2. In this case you don't need to clear pings vector, that check will help us figure out if we even wait for the ping the we just received

implementations/rust/ockam/ockam_api/src/session/session.rs
davide-baldo: Once we declare the connection as offline and close it, we don't care about previous pings anymore.
But we clear them only once the connection is up and running

implementations/rust/ockam/ockam_api/src/session/session.rs
davide-baldo: I don't know how to reproduce (if it is), I'm just saying: if you see unexpected behavior in locks, this could be the reason.

SanjoDeundiak: Thanks for the heads up

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/inlets_trait.rs
SanjoDeundiak: I think a dedicated struct for this argument set would be great, but I don't want to address that in this PR

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer.rs
SanjoDeundiak: Nice catch!

implementations/rust/ockam/ockam_api/src/session/session.rs
SanjoDeundiak: Makes sense!

davide-baldo (approved):

PR #8355, opened 2024-07-29
Kafka `disable-content-encryption` false by default

davide-baldo:


implementations/rust/ockam/ockam_command/src/kafka/inlet/create.rs
```diff
@@ -88,7 +88,7 @@ pub struct CreateCommand {
         long,
         name = "disable-content-encryption",
         value_name = "BOOL",
-        default_value_t = true
+        default_value_t = false
     )]
     pub disable_content_encryption: bool,
```
adrianbenavides, line 93: We should also consider deprecating this argument in favor of `--encrypt-content` to avoid using the `enable/disable` prefixes. We should keep both in any case.

adrianbenavides (approved):

implementations/rust/ockam/ockam_command/src/kafka/inlet/create.rs
davide-baldo: If we turn the logic to positive: `encrypt-content`, true by default, how would you disable it?

adrianbenavides: oh right, by default we want to encrypt 🤦  Then something like `--plaintext` could work? It's the official concept they use to define a non-encrypted connection. If that's confusing, maybe `--unencrypted` is another viable option.

PR #8364, opened 2024-07-30
fix(rust): make sure that there is only one space max marked as default

etorreborre:
This PR makes sure that we mark a stored space as default in the same transaction, to avoid possible race conditions.


etorreborre:
Closed in favor of https://github.com/build-trust/ockam/pull/8366.

PR #8366, opened 2024-07-30
fix(rust): do not set invited projects as default

adrianbenavides:
By ordering the projects before storing them, we make sure we don't assign an invited project as the local default project.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
@@ -66,11 +66,11 @@ impl ProjectsSqlxDatabase {
         let non_admin_emails: Vec<String> = users_roles
             .iter()
             .filter(|user_role| user_role.role != RoleInShare::Admin)
-            .map(|user_role| user_role.email.to_string())
+            .map(|user_role| user_role.email.to_string().to_lowercase())
             .collect();
 
         // Check if any of the emails are in the user table
-        let q = query_scalar(r#"SELECT EXISTS(SELECT 1 FROM "user" WHERE email IN ($1))"#)
+        let q = query_scalar(r#"SELECT EXISTS(SELECT 1 FROM "user" WHERE LOWER(email) IN ($1))"#)
             .bind(non_admin_emails.join(","));
         let shared: Boolean = q.fetch_one(transaction).await.into_core()?;
         Ok(shared.to_bool())
@@ -607,7 +607,7 @@ mod test {
                     name: "name".to_string(),
                     picture: "picture".to_string(),
                     updated_at: "2024-07-29T17:56:24.585Z".to_string(),
-                    email: "me@ockam.io".try_into().unwrap(),
+                    email: "Me@ockam.io".try_into().unwrap(),
                     email_verified: false,
                 })
                 .await
--- a/implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
@@ -1,9 +1,8 @@
-use std::str::FromStr;
-
 use sqlx::any::AnyRow;
 use sqlx::database::HasArguments;
 use sqlx::encode::IsNull;
 use sqlx::*;
+use std::str::FromStr;
 
 use ockam::identity::Identifier;
 use ockam_core::async_trait;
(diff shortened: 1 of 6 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
```diff
@@ -44,22 +44,61 @@ impl ProjectsSqlxDatabase {
...
+        // 1. The user enrolls and gets a project as an admin. There is an email from the "user"
+        //      table that matches an email in the project's "user_role" field that has role "admin".
+        // 2. Someone shares a project to the user via an invitation. There is an email from the "user"
+        //      table that matches an email in the project's "user_role" field that has role "service_user".
+        // 3. The user uses a ticket to enroll into a project. There is no email from the "user" table
+        //      that matches an email in the project's "user_role" field.
+        // This function returns true if we are in the second scenario; false otherwise.
+
+        // Get emails from user_roles that are not admins
+        let mut non_admin_emails = vec![];
```
etorreborre: Maybe it is worth promoting this as method on `ProjectModel`?

implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
```diff
@@ -44,22 +44,61 @@ impl ProjectsSqlxDatabase {
...
-            .into_core()?
-            .map(|row| row.get(0));
-        // The project is set as the default one if no other default project exists already
-        let is_default = project_id.is_none() || project_id == Some(project.id.clone());
+        let is_default: bool;
+        let mut project_name = &project.name;
+        // If it's a shared project, it can't be set as default.
+        if self.is_shared_project(project, &mut transaction).await? {
+            is_default = false;
+            // Also, the name is set to the project id  to avoid collisions with other
```
etorreborre: Just an unnecessary space in that sentence.

implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
```diff
@@ -44,22 +44,61 @@ impl ProjectsSqlxDatabase {
...
+            // projects with the same name that belong to other spaces.
+            project_name = &project.id;
```
etorreborre, line 138: This is out of this PR scope but we definitely need to come back to this because we arbitrarily change a project name for the user. This might be surprising for any UI that put on top:

 - You've been invited to the "erics-services" project
 - "show services" --> the services shared by project 8195bd43-eb4f-4e4a-9650-8fed1cc028a5 are ...

One way to solve this would be to accept the ambiguity and return an error whenever the project name is ambiguous. Then ask the user to specify the space name or space id.

implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
```diff
@@ -264,6 +303,17 @@ impl ProjectsRepository for ProjectsSqlxDatabase {
...
     async fn set_default_project(&self, project_id: &str) -> Result<()> {
         let mut transaction = self.database.begin().await.into_core()?;
+
+        if let Some(project) = self.get_project(project_id).await? {
+            if self.is_shared_project(&project, &mut transaction).await? {
+                warn!(
+                    "the project {} can't be set as default because is not owned by any local user",
+                    project_id
+                );
+                return Ok(());
```
etorreborre: I think we should return an error in that case. The user wants to set the project as a default but the action was not performed, this should not stay silent.

etorreborre (dismissed):
The project creation / update logic is indeed a bit tricky.
I just left a comment to possibly return an error instead of succeeding when a project cannot be set as a default.

implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
adrianbenavides: The non-admin projects are hidden from the user when using the command, so they shouldn't be able to interact directly with this kind of projects.

The reason I did this is that we have [some logic](https://github.com/build-trust/ockam/blob/b1f9800944f9f1cc3e7915052eb13ee4dc95c736/implementations/rust/ockam/ockam_api/src/cli_state/projects.rs#L74) where we set a new default project after deleting another project. Would you rather return an error in the repository and ignore the error in this layer?

implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
adrianbenavides: Agree. We should also stop using space/project names outside the command layer. The rest of the codebase should just work with their IDs. That should also help to remove this issue.

implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
adrianbenavides: Not sure if we are going to need to do this outside of this context. Maybe if I rewrite it using iterators would look better?


```rust
let non_admin_emails: Vec<String> = project
    .user_roles
    .iter()
    .filter(|user_role| user_role.role != RoleInShare::Admin)
    .map(|user_role| user_role.email.to_string())
    .collect();
```

implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
etorreborre: > The non-admin projects are hidden from the user when using the command, so they shouldn't be able to interact directly with this kind of projects.

I see.

> The reason I did this is that we have [some logic](https://github.com/build-trust/ockam/blob/b1f9800944f9f1cc3e7915052eb13ee4dc95c736/implementations/rust/ockam/ockam_api/src/cli_state/projects.rs#L74) where we set a new default project after deleting another project. Would you rather return an error in the repository and ignore the error in this layer?

Oh, but then the logic which tries to set a default project could select a shared project as the first one, try to set it as a default, then silently fail and we would end-up with no default project?
I think this means that the `delete_project` should take care of this logic inside `projects_repository_sql`. And this will be better because it will be inside the same transaction.

adrianbenavides: Done!

implementations/rust/ockam/ockam_api/src/cli_state/storage/projects_repository_sql.rs
etorreborre: That works for me as well. I was just suggesting this because it's possible that we might need this in the future and won't think about coming back to this code to refactor.

etorreborre (approved):
🗼

PR #8374, opened 2024-08-01
feat(rust): the config of `node create` accepts an `identity`

adrianbenavides:
This identity can be used in the config commands that accept an optional identity. For now, only the `project ticket` command supports this option.

The identity can be defined in the node configuration:

```bash
ockam node create config.yaml --identity i
```

or passed as a command arg:

```yaml
name: n
identity: i

ticket: $TICKET

relay: my-relay
```


etorreborre (approved):
🧮

PR #8354, opened 2024-07-29
Kafka refactoring

davide-baldo:
- Moved interceptor from kafka to portal, making it available for any protocol
- Heavy kafka refactoring: the core code is the same, but now it should be much more readable
  - inlet/outlet are now specular
  - flow control is better documented
  - lot of small simplifications
- Added a lot kafka bats tests based on local spawn redpanda and apache kafka
  - tests every kafka topology
  - these for now are disabled by default

PR #8380, opened 2024-08-02
Kafka: added the possibility to encrypt specific fields of a JSON record

davide-baldo:
This PR introduces the `--fields-to-encrypt` option in the kafka inlet, adding the possibility to encrypt only specific fields of a JSON message. When specified, it's assumed the payload is JSON.

To be merged after [davide-baldo/kafka-general-refactoring](https://github.com/build-trust/ockam/pull/8354)

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/kafka/key_exchange/mod.rs
+++ b/implementations/rust/ockam/ockam_api/src/kafka/key_exchange/mod.rs
@@ -1,17 +1,11 @@
+use crate::kafka::protocol_aware::KafkaEncryptedContent;
 use minicbor::{CborLen, Decode, Encode};
-
-use ockam_core::Address;
+use ockam_core::{async_trait, Address};
 use ockam_multiaddr::MultiAddr;
+use ockam_node::Context;
 
 pub(crate) mod controller;
 mod secure_channels;
-
-pub(crate) struct KafkaEncryptedContent {
-    /// The encrypted content
-    pub(crate) content: Vec<u8>,
-    /// The secure channel identifier used to encrypt the content
-    pub(crate) consumer_decryptor_address: Address,
-}
 
 /// Describe how to reach the consumer node: either directly or through a relay
 #[derive(Debug, Clone, Encode, Decode, CborLen)]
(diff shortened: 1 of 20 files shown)


mrinalwadhwa:
@davide-baldo let's rename the arg to `--encrypted-field`

davide-baldo:
Rebased and renamed the argument

implementations/rust/ockam/ockam_api/src/kafka/protocol_aware/inlet/mod.rs
```diff
@@ -12,53 +13,78 @@ use std::sync::{Arc, Mutex};
...
 
+#[cfg(test)]
+mod tests;
+
 #[derive(Clone)]
 pub(crate) struct InletInterceptorImpl {
     request_map: Arc<Mutex<HashMap<CorrelationId, RequestInfo>>>,
     uuid_to_name: TopicUuidMap,
-    key_exchange_controller: KafkaKeyExchangeController,
+    key_exchange_controller: Arc<dyn KafkaKeyExchangeController>,
```
adrianbenavides, line 23: If `KafkaKeyExchangeController` enforces the `clone` trait, and given that the `KafkaKeyExchangeControllerImpl` holds an Arc, could we avoid this Arc here? Not sure if that would make things ugly/uncomfortable.

implementations/rust/ockam/ockam_api/src/kafka/key_exchange/mod.rs
```diff
@@ -32,3 +26,43 @@ pub enum ConsumerPublishing {
...
+    /// Encrypts the content specifically for the consumer waiting for that topic name and
+    /// partition.
+    /// To do so, it'll create a secure channel which will be used for key exchange only.
+    /// The secure channel will be created only once and then re-used, hence the first time will
+    /// be slower, and may take up to few seconds.
+    async fn encrypt_content(
+        &self,
+        context: &mut Context,
+        topic_name: &str,
+        partition_id: i32,
```
adrianbenavides: To match the naming of the kafka-protocol library, shouldn't this be named `partition_index`?

implementations/rust/ockam/ockam_api/src/kafka/protocol_aware/inlet/response.rs
```diff
@@ -271,4 +261,67 @@ impl InletInterceptorImpl {
...
+        context: &mut Context,
+        record_value: Bytes,
+    ) -> Result<Vec<u8>, InterceptError> {
+        let mut record_value = serde_json::from_slice::<serde_json::Value>(&record_value)?;
+
+        if let serde_json::Value::Object(map) = &mut record_value {
+            for field in &self.encrypted_fields {
+                if let Some(value) = map.get_mut(field) {
+                    let encrypted_content = if let serde_json::Value::String(string) = value {
+                        hex::decode(string).map_err(|_| "Encrypted is not a valid hex string")?
```
adrianbenavides, line 296: I got confused at this point. Why does it expect a hex encoded string here?

implementations/rust/ockam/ockam_command/tests/bats/kafka/docker.bats
```diff
@@ -462,3 +462,87 @@ kafka_docker_end_to_end_encrypted_offset_decryption() {
...
+    --partition 0 \
+    --offset 1 \
+    --max-messages 1 --timeout-ms 30000 >"$CONSUMER_OUTPUT"
+
+  run bash -c "cat \"\$CONSUMER_OUTPUT\" | jq -r '.encrypted_field_one.key'"
+  assert_output "value"
+
+  run bash -c "cat \"\$CONSUMER_OUTPUT\" | jq -r '.encrypted_field_two[0]'"
+  assert_output "hello"
+}
```
adrianbenavides, line 543: Could you add some comments in the different sections of this test? Just to differentiate what's expected out of each group of commands.

implementations/rust/ockam/ockam_api/src/kafka/protocol_aware/inlet/mod.rs
davide-baldo: I've tried to add `Clone` to it, but it can't be called using `KafkaKeyExchangeController` since it's being used as a dynamic Trait, and you can't return Self using a trait.

implementations/rust/ockam/ockam_api/src/kafka/protocol_aware/inlet/response.rs
davide-baldo: We are encrypting singles fields, each field (regardless of the original type) gets encrypted and then encoded using a hex string.

adrianbenavides: Right, I just saw the place where it's hex-encoded, thanks!

adrianbenavides (approved):
Awesome! 🚀

PR #8395, opened 2024-08-07
feat(rust): node config improvements

adrianbenavides:
- Add validation for `node create` positional argument (name or conf): we'll now return an error if the node name is not valid (e.g. `node create config.yaml` and the config file doesn't exist.
- Refactor `ArgKey` as a struct instead of a type. This simplifies some data conversions and makes the code more readable
- The command args now have precedence over config values. This means that we now have a lot of combinations to run a node with a config file. For example: `node create n1 --identity i1 --configuration "{name: api-service-inlet}"` will use the name from the command arg, instead of the one defined in the config
- If an identity is passed to `node create`, the identity will be created if it doesn't exist, so that running the command in foreground/background/config modes, the behaviour will be the same

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer.rs
@@ -72,7 +72,10 @@ impl InletSessionReplacer {
                         .get_project_by_name(&p)
                         .await
                     {
-                        Some(p.authority_identifier()?)
+                        Some(
+                            p.authority_identifier()
+                                .ok_or(ApiError::core("no authority identifier"))?,
+                        )
                     } else {
                         None
                     }
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -8,10 +8,11 @@ use miette::Context as _;
 use miette::{miette, IntoDiagnostic};
 use opentelemetry::trace::TraceContextExt;
 use opentelemetry::KeyValue;
+use regex::Regex;
 use tracing::instrument;
 
 use ockam_api::cli_state::random_name;
-use ockam_api::colors::color_primary;
+use ockam_api::colors::{color_error, color_primary};
 use ockam_api::{fmt_log, fmt_ok};
 use ockam_core::{opentelemetry_context_parser, OpenTelemetryContext};
 use ockam_node::Context;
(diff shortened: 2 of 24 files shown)


implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -42,7 +45,7 @@ after_long_help = docs::after_help(AFTER_LONG_HELP)
 pub struct CreateCommand {
     /// Name of the node or a configuration to set up the node.
     /// The configuration can be either a path to a local file or a URL.
-    #[arg(value_name = "NAME_OR_CONFIGURATION", hide_default_value = true, default_value_t = random_name())]
+    #[arg(value_name = "NAME_OR_CONFIGURATION", hide_default_value = true, default_value = DEFAULT_NODE_NAME)]
```
etorreborre, line 48: Maybe I would not bother with a fake name and just use `Option<String>` here.

implementations/rust/ockam/ockam_command/tests/bats/orchestrator_enroll/nodes.bats
```diff
@@ -201,3 +201,15 @@ EOF
   # Use the identity to send a message
   $OCKAM message send hi --identity i1 --to "/project/default/service/forward_to_$RELAY_NAME/secure/api/service/echo"
 }
+
+@test "nodes - create with config, using the specified enrollment ticket" {
```
etorreborre, line 205: Maybe it would be good to add a bats test showing that we can override values coming from the configuration file with command-line arguments.

etorreborre (dismissed):
I hadn't noticed that we could pass a configuration file directly with `ockam node create config.yaml`.

My concern is that this is redundant with the existing `--configuration` argument. This also leads to more complicated code because now we have to guess if we're being passed a name or a configuration.

On the reflexion I would personally prefer a less ambiguous approach to command line arguments. Have either:

-  Full command line arguments 
- _OR_ a configuration file specified with `--configuration` (and even `--configuration-file` when using a file).

implementations/rust/ockam/ockam_command/src/node/create.rs
adrianbenavides: I need to investigate the side effects of doing that. I'm not sure it won't break things if we define this argument as optional. I'll try to tackle this in a separate PR in isolation

adrianbenavides:
> I hadn't noticed that we could pass a configuration file directly with ockam node create config.yaml.

Yeah, this was the intended UX when we first added the config file so the user can use the same positional argument for either a config or a node name. We should revisit it, but it will probably break some examples.

> My concern is that this is redundant with the existing --configuration argument. This also leads to more complicated code because now we have to guess if we're being passed a name or a configuration.

The `--configuration` argument was later added to accept an inline config, for use cases where we can't or don't want to store the config in a file. But same thing, we should revisit this UX.

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer.rs
```diff
@@ -72,7 +72,7 @@ impl InletSessionReplacer {
                         .get_project_by_name(&p)
                         .await
                     {
-                        Some(p.authority_identifier()?)
+                        p.authority_identifier()
```
etorreborre: Shouldn't we fail here? Because if `/project` is specified we really need one.

adrianbenavides: Right. I've also reviewed the rest of the calls to this function just in case.

implementations/rust/ockam/ockam_command/src/node/create.rs
adrianbenavides: About this, I tried and it adds a different complexity... Making the argument optional generates a new logical path around how you want to read or replace the arg value, and it's pretty error-prone. I don't think it's worth the effort right now.

etorreborre (approved):

PR #8430, opened 2024-08-29
influxdb inlet (auth proxy)

polvorin:
Assuming the identity is enrolled in a project that has influxdb addon enabled, like:
```
[configure influxdb'  addon]
$ ockam project addon configure  influxdb -e http://some.influxdb.instance:8086 -t $TOKEN -o $ORG_ID --permissions '$PERMISSIONS'
```

An http authentication inlet is started as:
```
[create a node, and a http-inlet on it.   it creates a tcp-outlet behind the scenes]
$ ockam node create n1
$ ockam http-inlet create --at /node/n1 --from 8087  --to /project/default/service/forward_to_influxdb/secure/api/service/outlet 
```
Then sending influxdb' request (writes, queries) to  `localhost:8087`  will cause the request to be sent to upstream influxdb instance (the outlet would be pointing to influxdb),  but with the request modified to include an `Authentication` header with a token obtained from the project' token lease manager.   This token is rotated as needed, without clients having to worry about that. 

The same  can be done through a node' config file as well: 
```
name: n1
http-inlets:
  influxdb:
    from: 0.0.0.0:8087
    to:  /project/default/service/forward_to_influxdb/secure/api/service/outlet
```

Notes:
* It requires a secure channel between inlet->outlet, even if both are on the same node (the first case, where we implicitly create the tcp outlet).  That's something to maybe improve, right now is mandatory.
* Right now, it doesn't support serving https,  as tcp-inlets doesn't support that _yet_ .  So influxdb' clients must be configured to use http (upstream can be http or https).
* Right now the original Token sent by clients is just discarded (it can be anything).   Perhaps we could require a custom string there instead (like "OCKAM-MANAGED" or something.  As long as clients don't complaint that that isn't a valid token when reading it from config)

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/http_auth/mod.rs
+++ b/implementations/rust/ockam/ockam_api/src/http_auth/mod.rs
@@ -1,0 +1,366 @@
+use std::io::Write;
+
+use httparse::{Header, Status};
+use ockam_core::async_trait;
+use ockam_node::Context;
+use ockam_transport_tcp::{Direction, PortalInterceptor, PortalInterceptorFactory};
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+use ockam::errcode::{Kind, Origin};
+
+use tracing::{debug, error};
+
+use crate::TokenLeaseRefresher;
+
+#[derive(Debug, Clone, PartialEq)]
+enum RequestState {
+    ParsingHeader(Option<Vec<u8>>),
+    ParsingChunkedHeader(Option<Vec<u8>>),
+    RemainingInChunk(usize),
+    RemainingBody(usize),
+}
+
+struct HttpAuthInterceptorState {
+    state: RequestState,
+}
+
+struct HttpAuthInterceptor {
+    state: Arc<Mutex<HttpAuthInterceptorState>>,
+    token_refresher: TokenLeaseRefresher,
+}
+
+impl HttpAuthInterceptor {
+    fn new(token_refresher: TokenLeaseRefresher) -> Self {
+        let state = HttpAuthInterceptorState {
+            state: RequestState::ParsingHeader(None),
+        };
+        Self {
+            state: Arc::new(Mutex::new(state)),
+            token_refresher,
+        }
+    }
+}
+
+pub struct HttpAuthInterceptorFactory {
+    token_refresher: TokenLeaseRefresher,
+}
+
+impl HttpAuthInterceptorFactory {
+    pub fn new(token_refresher: TokenLeaseRefresher) -> Self {
+        Self { token_refresher }
+    }
+}
+
+impl PortalInterceptorFactory for HttpAuthInterceptorFactory {
+    fn create(&self) -> Arc<dyn PortalInterceptor> {
+        Arc::new(HttpAuthInterceptor::new(self.token_refresher.clone()))
+    }
+}
+
+fn attach_auth_token_and_serialize_into(
+    req: &httparse::Request,
+    token: &str,
+    buffer: &mut Vec<u8>,
+) {
+    debug!("Serializing http req header");
+    write!(
+        buffer,
+        "{} {} HTTP/1.{}\r\n",
+        req.method.unwrap(),
+        req.path.unwrap(),
+        req.version.unwrap()
+    )
+    .unwrap();
+
+    write!(buffer, "Authorization: Token {}\r\n", token).unwrap();
+    for h in &*req.headers {
+        if !h.name.eq_ignore_ascii_case("Authorization") {
+            write!(buffer, "{}: ", h.name).unwrap();
+            buffer.extend_from_slice(h.value);
+            buffer.extend_from_slice(b"\r\n");
+        }
+    }
+    buffer.extend_from_slice(b"\r\n");
+}
+
+fn body_state(method: &str, headers: &[Header]) -> ockam_core::Result<RequestState> {
+    match method.to_uppercase().as_str() {
+        "POST" | "PUT" => {
+            for h in headers {
+                if h.name.eq_ignore_ascii_case("Content-Length") {
+                    if let Ok(str) = std::str::from_utf8(h.value) {
+                        return str.parse().map(RequestState::RemainingBody).map_err(|e| {
+                            ockam_core::Error::new(Origin::Transport, Kind::Invalid, e)
+                        });
+                    }
+                } else if h.name.eq_ignore_ascii_case("Transfer-Encoding")
+                    && String::from_utf8(h.value.to_vec()).is_ok_and(|s| s.contains("chunked"))
+                {
+                    return Ok(RequestState::ParsingChunkedHeader(None));
+                }
+            }
+            // Not content-length, no chunked encoding, fail.
+            Err(ockam_core::Error::new(
+                Origin::Transport,
+                Kind::Invalid,
+                "No Content-Length nor chunked Transfer-Encoding",
+            ))
+        }
+        _ => Ok(RequestState::ParsingHeader(None)),
+    }
+}
+
+impl RequestState {
+    /* Parse the incoming data,  attaching an Authorization header token to it.
+     * data is received in chunks, and there is no warranty on what we get on each:
+     * incomplete requests,  multiple requests, etc.
+     */
+    fn process_http_buffer(&mut self, buf: &[u8], token: &str) -> ockam_core::Result<Vec<u8>> {
+        let mut acc = Vec::with_capacity(buf.len());
+        let mut cursor = buf;
+        loop {
+            if cursor.is_empty() {
+                return Ok(acc);
+            }
+            match self {
+                RequestState::ParsingHeader(prev) => {
+                    let (to_parse, prev_size): (&[u8], usize) = if let Some(b) = prev {
+                        let prev_size = b.len();
+                        b.extend_from_slice(cursor);
+                        (b, prev_size)
+                    } else {
+                        (cursor, 0usize)
+                    };
+                    let mut headers = [httparse::EMPTY_HEADER; 64];
+                    let mut req = httparse::Request::new(&mut headers);
+                    match req.parse(to_parse) {
+                        Ok(httparse::Status::Partial) if prev_size == 0 => {
+                            // No previous buffered, need to copy and own the unparsed data
+                            *self = RequestState::ParsingHeader(Some(cursor.to_vec()));
+                            return Ok(acc);
+                        }
+                        Ok(httparse::Status::Partial) => {
+                            // There was a previous buffer, and we already added the newly data to it
+                            return Ok(acc);
+                        }
+                        Ok(httparse::Status::Complete(body_offset)) => {
+                            cursor = &cursor[body_offset - prev_size..];
+                            attach_auth_token_and_serialize_into(&req, token, &mut acc);
+                            *self = body_state(req.method.unwrap(), req.headers)?;
+                        }
+                        Err(e) => {
+                            error!("Error parsing header: {:?}", e);
+                            return Err(ockam_core::Error::new(
+                                Origin::Transport,
+                                Kind::Invalid,
+                                e,
+                            ));
+                        }
+                    }
+                }
+                RequestState::RemainingBody(remaining) => {
+                    if *remaining <= cursor.len() {
+                        acc.extend_from_slice(&cursor[..*remaining]);
+                        cursor = &cursor[*remaining..];
+                        *self = RequestState::ParsingHeader(None);
+                    } else {
+                        acc.extend_from_slice(cursor);
+                        *remaining -= cursor.len();
+                        return Ok(acc);
+                    }
+                }
+                RequestState::ParsingChunkedHeader(prev) => {
+                    let (to_parse, prev_size): (&[u8], usize) = if let Some(b) = prev {
+                        let prev_size = b.len();
+                        b.extend_from_slice(cursor);
+                        (b, prev_size)
+                    } else {
+                        (cursor, 0usize)
+                    };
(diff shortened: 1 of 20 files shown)


implementations/rust/ockam/ockam_api/src/http_auth/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+                        .map_err(|e| ockam_core::Error::new(Origin::Transport, Kind::Invalid, e));
+                }
+            } else if h.name.eq_ignore_ascii_case("Transfer-Encoding")
+                && String::from_utf8(h.value.to_vec()).is_ok_and(|s| s.contains("chunked"))
+            {
+                return Ok(RequestState::ParsingChunkedHeader(None));
+            }
+        }
+        // Not content-length, no chunked encoding, fail.
+        Err(ockam_core::Error::new(
```
davide-baldo: There is a third possibility: `Connection: close`, the content ends when the connection is closed. It's legacy from HTTP 1.0, so if InfluxDB clients do not use it, we can ignore it.

implementations/rust/ockam/ockam_api/src/http_auth/mod.rs
davide-baldo, line 1: Non-blocking, but I would appreciate it if you could generalize this code, to intercept HTTPs header, other uses for it are already appearing.

davide-baldo: For example, @metaclips is already modifying the `Host` header on the client side to make services work

implementations/rust/ockam/ockam_api/src/http_auth/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+
+            Direction::FromInletToOutlet => {
+                let mut guard = self.state.lock().await;
+                let token = self.token_refresher.get_token().await;
+                if token.is_none() {
+                    error!("No authorization token available");
+                }
+                let out = guard
+                    .state
+                    .process_http_buffer(buffer, &token.unwrap_or_default())?;
```
davide-baldo, line 258: In this case, I think it would be safer to close the portal by returning an error.
That should be easier to debug.

implementations/rust/ockam/ockam_api/src/influxdb_token_lease/token_lease_refresher.rs
```diff
@@ -0,0 +1,125 @@
...
+    ctx: Context,
+    project_client: ProjectNodeClient,
+) {
+    loop {
+        debug!("refreshing token");
+        let token_result = project_client.create_token(&ctx).await;
+        let now_t = now().unwrap();
+        let wait_secs = match token_result {
+            Ok(new_token) => {
+                let expires = DateTime::parse_from_rfc3339(&new_token.expires).unwrap();
```
davide-baldo: Please avoid using `wrap` where it can actually fail.

implementations/rust/ockam/ockam_api/src/influxdb_token_lease/token_lease_refresher.rs
```diff
@@ -0,0 +1,125 @@
...
+
+async fn refresh_loop(
+    token: Arc<RwLock<Option<String>>>,
+    ctx: Context,
+    project_client: ProjectNodeClient,
+) {
+    loop {
+        debug!("refreshing token");
+        let token_result = project_client.create_token(&ctx).await;
+        let now_t = now().unwrap();
```
davide-baldo: Adding ockam error to `refresh_loop` and using `now()?` should be enough to avoid unwrapping.
Unwrapping makes debugging harder and affects overall stability.

implementations/rust/ockam/ockam_api/src/influxdb_token_lease/token_lease_refresher.rs
davide-baldo, line 1: I was wondering if inverting the logic is possible, lazily fetching the token when needed.
We may end up asking for tokens that are never used.

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/node_manager_worker.rs
```diff
@@ -22,22 +28,36 @@ impl NodeManagerWorker {
...
             wait_for_outlet_duration,
             policy_expression,
             wait_connection,
             secure_channel_identifier,
             enable_udp_puncture,
             disable_tcp_fallback,
+            is_http_auth_inlet,
         } = create_inlet;
+
+        let prefix_route = if is_http_auth_inlet {
```
davide-baldo, line 40: You can avoid modifying the TCP creation by creating an endpoint which handles both the creation of the `http_auth_interceptor` and the inlet creation (by passing a proper prefix route).
This helps keep the TCP inlet creation code straightforward, regardless of the number of special cases. You can also add any number of custom parameters you need.

Kafka handles the interceptor similarly (but with some extra complexity) using services:

https://github.com/build-trust/ockam/blob/cb695cd90ba698a58e3878a062ee78bdba120ce9/implementations/rust/ockam/ockam_api/src/nodes/service/kafka_services.rs#L33

https://github.com/build-trust/ockam/blob/cb695cd90ba698a58e3878a062ee78bdba120ce9/implementations/rust/ockam/ockam_api/src/nodes/service/worker.rs#L136

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/node_manager_worker.rs
```diff
@@ -55,6 +75,27 @@ impl NodeManagerWorker {
...
+    async fn create_http_auth_interceptor(
+        &self,
+        ctx: &Context,
+        inlet_alias: &String,
+    ) -> Result<Address, Error> {
+        let token_refresher = TokenLeaseRefresher::new(ctx, self.node_manager.clone()).await?;
+        let http_interceptor_factory = Arc::new(HttpAuthInterceptorFactory::new(token_refresher));
+
+        let interceptor_address: Address = (inlet_alias.to_owned() + "_http_intercetor").into();
+        //TODO: incoming and outgoing access control?
```
davide-baldo: Yes, the worker needs to implement proper incoming and outgoing access control. You should be able to simply re-use the same policy of the relative `tcp-inlet` for both incoming and outgoing.

Here an example in the kafka code, https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/nodes/service/kafka_services.rs#L222-L241

implementations/rust/ockam/ockam_command/src/http/inlet/create.rs
```diff
@@ -0,0 +1,566 @@
...
+use ockam_multiaddr::{MultiAddr, Protocol as _};
+use ockam_node::compat::asynchronous::resolve_peer;
+
+use crate::util::parsers::duration_parser;
+use crate::util::parsers::hostname_parser;
+use crate::util::{find_available_port, port_is_free_guard, process_nodes_multiaddr};
+
+/// Create TCP Inlets
+#[derive(Clone, Debug, Args)]
+pub struct CreateCommand {
```
davide-baldo: You should be able to ""inherit"" fields by declaring a field containing the inlet `Create` command with the `#[command(flatten)]` attribute. From the clap point of view would be as-if the fields were locally declared.

davide-baldo (commented):
The implementing looks good!
There is some room for simplification around the CLI commands and on the endpoint.

Also, this way of creating inlet+interceptor is pretty new (kafka is the only user of interceptor but has many services and multiple inlets).
Ideally, we would like to keep inlet creation independent of interceptors while avoiding the re-creation of CRUD for every inlet type.

davide-baldo:
Other than InfluxDb, I can't think of other use cases where the `Authorization` header is dynamically added. Maybe we should name it `influxdb-inlet`?

implementations/rust/ockam/ockam_api/src/http_auth/mod.rs
polvorin: ha, 👍  will incorportate it anyway.  I didn't encounter it when testing with influx, but who knows

implementations/rust/ockam/ockam_api/src/http_auth/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+            write!(buffer, "{}: ", h.name).unwrap();
+            buffer.extend_from_slice(h.value);
+            buffer.extend_from_slice(b"\r\n");
+        }
+    }
+    buffer.extend_from_slice(b"\r\n");
+}
+
+fn body_state(method: &str, headers: &[Header]) -> ockam_core::Result<RequestState> {
+    if ["POST".to_string(), "PUT".to_string()].contains(&method.to_uppercase()) {
```
adrianbenavides: (nitpick) You could use a match here to make it look cleaner:

```
match method.to_uppercase().as_str() {
    "POST" | "PUT" => {
        todo!()
    }
    _ => Ok(RequestState::ParsingHeader(None)),
}
```

implementations/rust/ockam/ockam_api/src/http_auth/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+        let req = "GET /home/user/example.txt HTTP/1.1\r\n\r\n";
+        let mut data = Vec::new();
+        data.extend_from_slice(req.as_bytes());
+        data.extend_from_slice(req.as_bytes());
+
+        let mut expected = format!(
+            "GET /home/user/example.txt HTTP/1.1\r\nAuthorization: Token {}\r\n\r\n",
+            TOKEN
+        );
+        expected = expected.clone() + &expected;
```
adrianbenavides, line 353: Why is it expected to get the request twice? Could you add the context to the `process_http_buffer` function docs if that makes sense?

implementations/rust/ockam/ockam_api/src/influxdb_token_lease/token_lease_refresher.rs
```diff
@@ -0,0 +1,125 @@
...
+#[derive(Clone)]
+pub struct TokenLeaseRefresher {
+    token: Arc<RwLock<Option<String>>>,
+}
+
+impl TokenLeaseRefresher {
+    pub async fn new(
+        ctx: &Context,
+        node_manager: Arc<InMemoryNode>,
+    ) -> Result<TokenLeaseRefresher, Error> {
```
adrianbenavides, line 29: The result type should be the one from the [crate](https://github.com/build-trust/ockam/blob/cf7467f5b9d3707effbfd8c9c9cb2d5edbce7424/implementations/rust/ockam/ockam_api/src/cli_state/error.rs#L5), so you don't have to map every error manually.

implementations/rust/ockam/ockam_api/src/http_auth/mod.rs
polvorin: hum,  but these are _requests_ only, and `Connection: close` seems a response-only thing

>  5.By the server closing the connection. (Closing the connection
     cannot be used to indicate the end of a request body, since that
     would leave no possibility for the server to send back a response.)

will let it for latter,  _maybe_ at some point we need to parse responses too (at that point need to adapt other parts of the code as well)

implementations/rust/ockam/ockam_api/src/influxdb_token_lease/token_lease_refresher.rs
polvorin: it's a somewhat expensive op (in the sense that it might need to create a secure channel to the project).   In some close future it's even 2 secure channels.    So it _can_ take some time.   Would better not delay/buffer/timeout request in the meantime.

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/node_manager_worker.rs
polvorin: it's significant more boilerplate 🤷  .  Agree there should be some better way,  but we can think on generalization  once there are _more_ than one use of it

implementations/rust/ockam/ockam_api/src/http_auth/mod.rs
polvorin: the request is sent twice, that's why it's received twice as well.  Is to check that pipelined requests are parsed ok.

implementations/rust/ockam/ockam_api/src/influxdb_token_lease/token_lease_refresher.rs
polvorin: not sure I understand the differences between all the error types and what is compatible with what 😬 , but  some of the errors that could happen inside this function are `CliStateError` indeed,  but some others aren't.     In general,  what error should a failure to start this here should return?  I'm sure it' isn't a `CliStateError` (altough that could be the underling cause),   but really don't know between `ockam_core:Error `or `ockam_api::Error` ,  used the latter as it's on the ockam_api' create and seems the error returned on other modules here.

implementations/rust/ockam/ockam_api/src/http_auth/mod.rs
polvorin: you mean returning an error for the intercept() function,   would that close the portal?.

Right now this case just log a warn/err on the logs.

implementations/rust/ockam/ockam_api/src/influxdb_token_lease/token_lease_refresher.rs
adrianbenavides: Yeah good point... I see that we are using all three all over the crate, so I guess it makes sense to leave it as it is now.

adrianbenavides (approved):

polvorin:
closing in favor of https://github.com/build-trust/ockam/pull/8461

PR #8472, opened 2024-09-18
feat(rust): allow multiple values in node configuration entries

davide-baldo:


adrianbenavides (commented):
Looks good! Could you expand [this test](https://github.com/build-trust/ockam/blob/b81f3fe63c0c044d0dd8a0b03dfba0bb02fd8015/implementations/rust/ockam/ockam_command/src/run/parser/resource/kafka_inlet.rs#L59) to have an example of the new feature? It will also be useful as an example of how to use it.

davide-baldo:
Just added the field in the YAML!

adrianbenavides (approved):

PR #8490, opened 2024-09-23
test(rust): disable re-create outlet bats test

SanjoDeundiak:


polvorin (dismissed):
putting a `skip`  [instruction at the beginning of the test will do it too](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_command/tests/bats/examples/kafka.bats#L30).

polvorin (dismissed):

polvorin (dismissed):

polvorin (approved):

PR #8504, opened 2024-10-01
feat(rust): allow relay connection failure without failing relay creation

davide-baldo:
image
image
image

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/relay/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/relay/create.rs
@@ -11,15 +11,16 @@ use ockam::Context;
 use ockam_api::address::extract_address_value;
 use ockam_api::colors::color_primary;
 
+use ockam_api::nodes::models::relay::ReturnTiming;
 use ockam_api::nodes::service::relay::Relays;
 use ockam_api::nodes::BackgroundNodeClient;
-use ockam_api::{fmt_ok, CliState};
+use ockam_api::{fmt_info, fmt_ok, fmt_warn, CliState, ConnectionStatus};
 use ockam_multiaddr::proto::Project;
 use ockam_multiaddr::{MultiAddr, Protocol};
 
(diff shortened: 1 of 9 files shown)


implementations/rust/ockam/ockam_command/src/relay/create.rs
```diff
@@ -79,6 +83,11 @@ impl Command for CreateCommand {
         let cmd = self.parse_args(&opts).await?;
         let at = cmd.at();
         let alias = cmd.relay_name();
+        let return_timing = if cmd.no_connection_wait {
+            ReturnTiming::Immediately
+        } else {
+            ReturnTiming::AfterConnection
+        };
```
adrianbenavides: Should we move this as a function of `cmd`, similar to what we do in the previous two lines?

implementations/rust/ockam/ockam_command/src/relay/create.rs
```diff
@@ -53,11 +54,14 @@ pub struct CreateCommand {
     #[arg(long)]
     relay_address: Option<String>,
 
-    /// Whether the relay will be used to relay messages at a project.
-    /// By default, this information will be inferred from the `--at` argument.
+    /// Deprecated
```
adrianbenavides: We should add a warn message to display this. Something like [this](https://github.com/build-trust/ockam/blob/167a0e310cea891669fd3990068ec5f19b3c641a/implementations/rust/ockam/ockam_command/src/node/stop.rs#L42) would work (perhaps we could extract the print logic into a function, similar to [this](https://github.com/build-trust/ockam/blob/d0e14a630956b72cb27dd047c9ae3d9e9dd55a51/implementations/rust/ockam/ockam_command/src/util/mod.rs#L212)). I'd add this logic inside the `parse_args` function.

adrianbenavides (approved):
🚀

PR #8508, opened 2024-10-02
eBPF updates

SanjoDeundiak:


implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
```diff
@@ -132,6 +132,7 @@ pub struct CreateCommand {
     pub no_tcp_fallback: bool,
 
     /// Use eBPF and RawSocket to access TCP packets instead of TCP data stream.
+    /// If `OCKAM_EBPF` env variable is set, this argument will be `true`.
     #[arg(long, hide = true)]
```
adrianbenavides: You can use the [`env` attribute](https://docs.rs/clap/latest/clap/struct.Arg.html#method.env) to handle this automatically: `#[arg(long, env="OCKAM_EBPF", hide = true)]`. You won't need to override the arg value manually in the command logic.

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/raw_socket_processor.rs
```diff
@@ -0,0 +1,156 @@
...
+    async fn get_new_packet(
+        mut socket_read_handle: TransportReceiver,
+    ) -> Result<(TransportReceiver, ParsedRawSocketPacket)> {
+        // TODO: I wonder how bad it is to use blocking here
+        //  Will it be shutdown properly eventually?
+        //  Should we use socket read with timeout?
+        let (socket_read_handle, parsed_packet) = tokio::task::spawn_blocking(move || {
+            // TODO: Should we check the checksum?
+            let (packet, source_ip, destination_ip) = next_tcp_packet(&mut socket_read_handle)
+                .map_err(|_| TransportError::RawSocketReadError)?;
```
adrianbenavides: Is there any situation where the `next_tcp_packet` would run forever? If that happens the async executor would not be able to shut down right?

implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
SanjoDeundiak: Nice, tnx!

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/raw_socket_processor.rs
SanjoDeundiak: It will run forever if there is no tcp packet, but I'm not sure it won't shutdown properly, should test. However, I don't think we can do better easily. We should make this call asynchronous, which I have in my backlog

.github/workflows/rust.yml
```diff
@@ -321,10 +321,41 @@ jobs:
...
+        shell: bash
+        run: |
+          set -x
+          home_dir=$(echo ~)
+          echo "$home_dir"
+          echo "HOME_DIR=$home_dir" >> $GITHUB_ENV
+          mkdir -p ~/.bats-tests/
+
+          if sudo ls -a /root/.bats-tests; then
+            sudo tar -czvf "${home_dir}/.bats-tests/root_tests.tar.gz" -C /root/.bats-tests .
```
davide-baldo, line 339: Why are you loading/saving the bat tests state?

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/internal_processor.rs
```diff
@@ -0,0 +1,166 @@
...
+                    Some(connection) => {
+                        trace!(
+                            "Inlet Processor: Existing connection from {}:{}",
+                            parsed_packet.packet.source_ip,
+                            parsed_packet.packet.source
+                        );
+                        connection
+                    }
+                    None => {
+                        if parsed_packet.packet.flags != 2 {
```
davide-baldo: My guess is that 2 should be the `SYNC` flag? Could you make it explicit?
(with a small explanation that SYNC should be set only during connection initialization)

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/pnet_helper.rs
```diff
@@ -0,0 +1,45 @@
...
+        let len = pnet_sys::recv_from(receiver.socket.fd, &mut receiver.buffer[..], &mut caddr)?;
+
+        let src = pnet_sys::sockaddr_to_addr(&caddr, size_of::<pnet_sys::SockAddrStorage>())?;
+        let src = match src {
+            net::SocketAddr::V4(sa) => *sa.ip(),
+            net::SocketAddr::V6(_) => continue,
+        };
+
+        // FIXME: Is that guaranteed that we receive the packet fully in one read?
+        let ip_header = Ipv4Packet::new(&receiver.buffer[..]).unwrap();
```
davide-baldo: Can we return an error in case the header is corrupted?

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/raw_socket_processor.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    async fn get_new_packet(
+        mut socket_read_handle: TransportReceiver,
+    ) -> Result<(TransportReceiver, ParsedRawSocketPacket)> {
+        // TODO: I wonder how bad it is to use blocking here
+        //  Will it be shutdown properly eventually?
+        //  Should we use socket read with timeout?
+        let (socket_read_handle, parsed_packet) = tokio::task::spawn_blocking(move || {
+            // TODO: Should we check the checksum?
```
davide-baldo: Mhm. Isn't the checksum handled at the IP level?

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/raw_socket_processor.rs
```diff
@@ -0,0 +1,156 @@
...
+    async fn handle_outlet(&self, outlet: Outlet, packet: ParsedRawSocketPacket) -> Result<()> {
+        outlet.sender.send(packet).await.unwrap();
+
+        Ok(())
+    }
+
+    async fn get_new_packet(
+        mut socket_read_handle: TransportReceiver,
+    ) -> Result<(TransportReceiver, ParsedRawSocketPacket)> {
+        // TODO: I wonder how bad it is to use blocking here
```
davide-baldo, line 59: I think you can put a timeout  (SO_RCVTIMEO and SO_SNDTIMEO) inside the socket itself rather than manually implementing it. The blocking syscall should return with an error as soon as the timeout is hit.

davide-baldo (commented):
eBPF is getting cleaner and easier to follow! 👍

.github/workflows/rust.yml
SanjoDeundiak: That's uploading logs by @metaclips

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/internal_processor.rs
SanjoDeundiak: I'll add a comment. It's `SYN` flag, but we're not only checking if `SYN` flag is set, we also check that no other flag is set.

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/pnet_helper.rs
SanjoDeundiak: Yes, there are a lot of unwraps I need to get rid of in the future, this is one of them

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/raw_socket_processor.rs
SanjoDeundiak: IP checksum - yes, tcp checksum - no

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/raw_socket_processor.rs
SanjoDeundiak: I'm not yet sure what is the correct way to do it. For now, that's mostly copy&pasted code from pnet library. The library only allowed to query packets using an iterator, which was very inconvenient in our context. As the next step I would like to see how reading from a socket/file descriptor is implemented in tokio and reuse that.

implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
SanjoDeundiak: Fixed

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/internal_processor.rs
SanjoDeundiak: Fixed

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/pnet_helper.rs
SanjoDeundiak: Fixed

davide-baldo (approved):

PR #8520, opened 2024-10-07
feat(rust): compact enrollment ticket encoded format

adrianbenavides:
The enrollment tickets are now a comma-separated text line, with all the necessary information. The fields are ordered as follows: 

`<project route>,<project identifier>,<project name>,<token>,<project change history>,<authority change history>,<authority route>`

The project route will follow a similar structure to this: 

`/dnsaddr/<id>.projects.orchestrator.ockam.io/tcp/443/service/<id>/service/api`

The authority route is optional. It will derive it from the project id/route if not present. The project identifier and project name will eventually disappear from this format, but we need it there for now.

## Using the previous format

A message will be shown asking the user to use an up-to-date ockam version. It will still convert the old ticket into the new format, and use it normally to enroll.

image

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/enrollments.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/enrollments.rs
@@ -1,10 +1,4 @@
-use serde::{Deserialize, Serialize};
-use std::fmt::{Display, Formatter};
-use time::OffsetDateTime;
-
 use crate::authenticator::one_time_code::OneTimeCode;
-use ockam::identity::{Identifier, TimestampInSeconds};
-
 use crate::cli_state::Result;
 use crate::cli_state::{CliState, CliStateError};
 use crate::cloud::email_address::EmailAddress;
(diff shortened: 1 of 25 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/enrollments.rs
```diff
@@ -276,3 +277,466 @@ impl EnrollmentTicket {
...
+    }
+}
+
+impl FromStr for ExportedEnrollmentTicket {
+    type Err = ApiError;
+
+    fn from_str(contents: &str) -> std::result::Result<Self, Self::Err> {
+        // Decode as comma-separated text
+        let values: Vec<&str> = contents.split(',').collect();
+        if values.len() < Self::MANDATORY_FIELDS_NUM {
```
etorreborre, line 348: Shouldn't we raise an error also if there are more fields?

implementations/rust/ockam/ockam_api/src/cli_state/enrollments.rs
```diff
@@ -276,3 +277,466 @@ impl EnrollmentTicket {
...
+        let (
+            project_id_or_route,
+            project_name,
+            one_time_code,
+            project_change_history,
+            authority_change_history,
+        ) = (values[0], values[1], values[2], values[3], values[4]);
+        let authority_route = values.get(5).map(|r| r.to_string());
+
+        let project_id_or_route = if project_id_or_route.starts_with('/') {
```
etorreborre: I think we can call the variable in `let project_id_or_route`, simply `project_route` after we checked for `starts_with('/')`.

etorreborre (dismissed):
🎫

implementations/rust/ockam/ockam_api/src/cli_state/enrollments.rs
adrianbenavides: They will just be ignored. I don't think it's necessary to enforce that.

etorreborre (approved):

PR #8525, opened 2024-10-07
Fix flaky test by improving SQLite concurrency model

davide-baldo:
~~Changed mode to operate SQLite to `WAL` mode, see https://sqlite.org/wal.html~~
To address migration locking, an exclusive lock is used, which is skipped whenever migration isn't necessary.
To address concurrent database access WAL mode plus an operation-level retry mechanism when database deadlocks.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
@@ -289,16 +289,7 @@ impl CliState {
         let _ = std::fs::remove_dir_all(Self::make_commands_log_dir_path(root_path));
         // Delete the nodes database, keep the application database
         if let Some(path) = Self::make_database_configuration(root_path)?.path() {
-            std::fs::remove_file(path.clone())?;
-            // wal and shm files may not exist, depending on the database state
-            let wal_path = path.with_extension("sqlite3-wal");
-            if wal_path.exists() {
-                std::fs::remove_file(wal_path)?;
-            }
-            let shm_path = path.with_extension("sqlite3-shm");
-            if shm_path.exists() {
-                std::fs::remove_file(shm_path)?;
-            }
+            std::fs::remove_file(path)?;
         };
         Ok(())
     }
@@ -409,16 +400,9 @@ mod tests {
 
     /// HELPERS
     fn list_file_names(dir: &Path) -> Vec<String> {
-        let file_names: Vec<_> = fs::read_dir(dir)
+        fs::read_dir(dir)
             .unwrap()
             .map(|f| f.unwrap().file_name().to_string_lossy().to_string())
-            .collect();
-
-        // remove -wal and -shm files from the list
-        // they may or may not exist depending on the database state
-        file_names
-            .into_iter()
-            .filter(|file| !file.ends_with("-wal") && !file.ends_with("-shm"))
-            .collect::<Vec<String>>()
+            .collect()
     }
 }
--- a/implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
@@ -289,7 +289,16 @@ impl CliState {
         let _ = std::fs::remove_dir_all(Self::make_commands_log_dir_path(root_path));
         // Delete the nodes database, keep the application database
         if let Some(path) = Self::make_database_configuration(root_path)?.path() {
-            std::fs::remove_file(path)?
+            std::fs::remove_file(path.clone())?;
+            // wal and shm files may not exist, depending on the database state
+            let wal_path = path.with_extension("sqlite3-wal");
+            if wal_path.exists() {
+                std::fs::remove_file(wal_path)?;
+            }
+            let shm_path = path.with_extension("sqlite3-shm");
+            if shm_path.exists() {
+                std::fs::remove_file(shm_path)?;
+            }
         };
         Ok(())
     }
@@ -400,9 +409,16 @@ mod tests {
 
     /// HELPERS
     fn list_file_names(dir: &Path) -> Vec<String> {
-        fs::read_dir(dir)
+        let file_names: Vec<_> = fs::read_dir(dir)
             .unwrap()
             .map(|f| f.unwrap().file_name().to_string_lossy().to_string())
-            .collect()
+            .collect();
+
+        // remove -wal and -shm files from the list
+        // they may or may not exist depending on the database state
+        file_names
+            .into_iter()
+            .filter(|file| !file.ends_with("-wal") && !file.ends_with("-shm"))
+            .collect::<Vec<String>>()
     }
 }
--- a/implementations/rust/ockam/ockam_api/src/cli_state/storage/auto_retry.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/storage/auto_retry.rs
@@ -1,0 +1,628 @@
+use crate::cli_state::journeys::{Journey, ProjectJourney};
+use crate::cli_state::{
+    EnrollmentsRepository, IdentitiesRepository, IdentityEnrollment, JourneysRepository,
+    NamedIdentity, NamedVault, NodeInfo, NodesRepository, ProjectsRepository, SpacesRepository,
+    TcpInlet, TcpPortalsRepository, UsersRepository, VaultType, VaultsRepository,
+};
+use crate::cloud::email_address::EmailAddress;
+use crate::cloud::enroll::auth0::UserInfo;
+use crate::cloud::project::models::ProjectModel;
+use crate::cloud::space::Space;
+use crate::config::lookup::InternetAddress;
+use crate::nodes::models::portal::OutletStatus;
+use chrono::{DateTime, Utc};
+use ockam::identity::models::{ChangeHistory, CredentialAndPurposeKey, PurposeKeyAttestation};
+use ockam::identity::storage::PurposeKeysRepository;
+use ockam::identity::{
+    ChangeHistoryRepository, CredentialRepository, Identifier, Identity, Purpose,
+    TimestampInSeconds,
+};
+use ockam_core::{async_trait, Address};
+use ockam_vault::storage::SecretsRepository;
+use ockam_vault::{
+    AeadSecret, AeadSecretKeyHandle, SigningSecret, SigningSecretKeyHandle, X25519SecretKey,
+    X25519SecretKeyHandle,
+};
+
+macro_rules! retry {
+    ($async_function:expr) => {{
+        let mut retries = 0;
+        loop {
+            match $async_function.await {
+                Ok(result) => break Ok(result),
+                Err(err) => {
+                    if err.to_string().contains("database is locked") && retries < 100 {
+                        tokio::time::sleep(tokio::time::Duration::from_millis(10)).await;
+                    } else {
+                        break Err(err);
+                    }
+                    retries += 1;
+                }
+            }
+        }
+    }};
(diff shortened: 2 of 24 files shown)


implementations/rust/ockam/ockam_node/src/storage/database/database_configuration.rs
```diff
@@ -108,12 +110,27 @@ impl DatabaseConfiguration {
...
         DatabaseConfiguration::Sqlite {
             path: Some(path.to_path_buf()),
+            single_connection: false,
         }
     }
 
     /// Create an in-memory sqlite configuration
     pub fn sqlite_in_memory() -> DatabaseConfiguration {
-        DatabaseConfiguration::Sqlite { path: None }
+        DatabaseConfiguration::Sqlite {
```
etorreborre: I would evolve the `DatabaseConfiguration` to introduce `SqliteInMemory` and `SqliteOnDisk` so that `single_connection` only makes sense when the `path` is `Some`. 
This way we won't have to `AutoRetry` when the database is used in memory only.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
```diff
@@ -245,35 +328,73 @@ impl NextMigration<'_> {
...
         }
-        Ok(())
+    }
+
+    async fn needs_rust_migration<'a>(
+        migration: &'a dyn RustMigration,
+        connection: &mut AnyConnection,
+        _applied_migrations: &[AppliedMigration],
+    ) -> Result<bool> {
+        if Migrator::has_migrated(connection, migration.name()).await? {
```
etorreborre: `Ok(!Migrator::has_migrated(connection, migration.name()).await?)` could be a one-liner here.

implementations/rust/ockam/ockam_command/tests/bats/orchestrator/portals.bats
```diff
@@ -197,7 +197,6 @@ teardown() {
 }
 
 @test "portals - local inlet and outlet passing through a relay, removing and re-creating the outlet" {
-  skip
```
davide-baldo, line 200: forgot to revert this change, will do tomorrow

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
```diff
@@ -209,15 +258,64 @@ impl SqlxDatabase {
...
-        let pool = Pool::connect_with(options)
+
+        let max_pool_size = match configuration {
+            DatabaseConfiguration::Sqlite {
+                single_connection, ..
+            } => {
+                if *single_connection {
+                    1
+                } else {
+                    16
```
SanjoDeundiak: Could you please add a comment explaining why 16 is picked here?

implementations/rust/ockam/ockam_command/tests/bats/orchestrator/portals.bats
```diff
@@ -281,7 +280,7 @@ teardown() {
   socat_pid=$!
   sleep 2
 
-  run_success curl -sfI --retry-all-errors --retry-delay 2 --retry 10 -m 30 "127.0.0.1:${inlet_port}"
+  run_success curl -sfI --retry-all-errors --retry-delay 2 --retry 30 -m 60 "127.0.0.1:${inlet_port}"
```
SanjoDeundiak: I was actually wondering all this time we even need retries. Now I'm even more concerned why we need to increase retry attempts and time. Isn't it supposed to work from the first try here?

implementations/rust/ockam/ockam_command/tests/bats/load/orchestrator.bash
```diff
@@ -43,7 +43,14 @@ function get_project_data() {
 
 function copy_enrolled_home_dir() {
   if [ ! -z "${ORCHESTRATOR_TESTS}" ]; then
-    cp -r $OCKAM_HOME_BASE/application_database.sqlite3 $OCKAM_HOME/
-    cp -r $OCKAM_HOME_BASE/database.sqlite3 $OCKAM_HOME/
+    cp -a $OCKAM_HOME_BASE/application_database.sqlite3 $OCKAM_HOME/
+    if [ -f $OCKAM_HOME_BASE/application_database.sqlite3-wal ]; then
```
SanjoDeundiak: Hm. Is it required? I thought we only want to clone the database here

implementations/rust/ockam/ockam_api/src/cli_state/storage/auto_retry.rs
```diff
@@ -0,0 +1,628 @@
...
+    TimestampInSeconds,
+};
+use ockam_core::{async_trait, Address};
+use ockam_vault::storage::SecretsRepository;
+use ockam_vault::{
+    AeadSecret, AeadSecretKeyHandle, SigningSecret, SigningSecretKeyHandle, X25519SecretKey,
+    X25519SecretKeyHandle,
+};
+
+macro_rules! retry {
```
SanjoDeundiak, line 27: Can we instead use a wrapper around `Query::execute` that would retry under the hood, so that we don't need to re-implement all the repository traits?

SanjoDeundiak: Also, some operations here can consist of multiple queries, and I'm not sure it's guaranteed that all of them are idempotent

implementations/rust/ockam/ockam_command/tests/bats/local/reset.bats
```diff
@@ -34,6 +34,9 @@ teardown() {
   # reset the OCKAM_HOME directory
   run_success "$OCKAM" reset --yes
 
+  # it may, or may not exist, so we remove it to have a consistent state
+  run_success rm -f "$OCKAM_HOME"/application_database.sqlite3-wal
```
adrianbenavides: Should this be done as part of the `reset` command?

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
adrianbenavides: Maybe an env var with a default value could be useful in environments where we want to customize it?

implementations/rust/ockam/ockam_command/tests/bats/orchestrator/portals.bats
davide-baldo: Yes, it's supposed to work on the first try, I forgot to revert inlet-related tests changes before pushing.
I'll address these tests in a follow-up PR if i manage to discover the root cause.

implementations/rust/ockam/ockam_command/tests/bats/load/orchestrator.bash
davide-baldo: the `sqlite3-wal` file is a piece of the database, without it the database would be corrupted.

implementations/rust/ockam/ockam_api/src/cli_state/storage/auto_retry.rs
davide-baldo: > Can we instead use a wrapper around Query::execute that would retry under the hood, so that we don't need to re-implement all the repository traits?

A single transaction could issue multiple execute() statements, wrapping it directly wouldn't be enough.

> Also, some operations here can consist of multiple queries, and I'm not sure it's guaranteed that all of them are idempotent

Any operations should be wrapped around a single transaction, when that fails, everything will be reverted.
I think we can make the assumption of idempotency for database operations, if this is not the case in specific cases, we can fix it and make the operation fully atomic.

implementations/rust/ockam/ockam_command/tests/bats/local/reset.bats
davide-baldo: The application database is untouched by the ockam reset, and at the end of the reset the application database `wal` file may or may not exist.
When it exists, the `ls` would have a different output, breaking the assertion.
The alternative would be to assert either outputs with or without the `wal` file.

implementations/rust/ockam/ockam_node/src/storage/database/database_configuration.rs
davide-baldo: >I would evolve the DatabaseConfiguration to introduce SqliteInMemory and SqliteOnDisk so that single_connection only makes sense when the path is Some.

👍 

> This way we won't have to AutoRetry when the database is used in memory only.

AFAIK, deadlocks can also occur with in-memory databases, so auto-retry should also be enabled in that case.

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
davide-baldo: > Maybe an env var with a default value could be useful in environments where we want to customize it?

We could, but I don't think there is the need for fine-tuning yet.
The only reason i changed from the default 10 to 16, it's because it's closer to the real number of threads tokio is using on a typical node.

davide-baldo:
In the last revision, I've disabled the WAL mode and everything seems to be working fine after many test iterations, without any extra workaround other than ignoring `1032 maps to SQLITE_READONLY_DBMOVED` error when deleting a non-persistent node.

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
```diff
@@ -289,7 +289,7 @@ impl CliState {
         let _ = std::fs::remove_dir_all(Self::make_commands_log_dir_path(root_path));
         // Delete the nodes database, keep the application database
         if let Some(path) = Self::make_database_configuration(root_path)?.path() {
-            std::fs::remove_file(path)?
+            std::fs::remove_file(path.clone())?;
```
SanjoDeundiak: Is clone required here?

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
```diff
@@ -400,9 +400,16 @@ mod tests {
...
             .unwrap()
             .map(|f| f.unwrap().file_name().to_string_lossy().to_string())
-            .collect()
+            .collect();
+
+        // remove -wal and -shm files from the list
+        // they may or may not exist depending on the database state
+        file_names
+            .into_iter()
+            .filter(|file| !file.ends_with("-wal") && !file.ends_with("-shm"))
```
SanjoDeundiak: Do we still have these files?

implementations/rust/ockam/ockam_api/src/nodes/service/in_memory_node.rs
```diff
@@ -58,13 +58,14 @@ impl Drop for InMemoryNode {
...
-                cli_state
-                    .remove_node(&self.node_name)
-                    .await
-                    .unwrap_or_else(|e| panic!("cannot delete the node {}: {e:?}", self.node_name));
+                let result = self.cli_state.remove_node(&self.node_name).await;
+                if let Err(err) = result {
+                    // code: 1032 maps to SQLITE_READONLY_DBMOVED - meaning the database has been
+                    // moved to another directory, most likely already deleted
+                    if !err.to_string().contains("code: 1032") {
+                        panic!("cannot delete the node {}: {err:?}", self.node_name);
```
SanjoDeundiak: Should we really panic here?

davide-baldo: I've changed the panic into an error, it shouldn't happen, but even if it does, it wouldn't result in a catastrophic effect

SanjoDeundiak (approved):

PR #8529, opened 2024-10-09
feat: deleting project members

etorreborre:
This PR:

- Deletes all the members of a project before deleting the project itself. 
- Adds an endpoint to delete all the members of a project at once, instead of one by one, in order to speed up the deletion. This endpoint is not yet used, we need first to deploy it on Authority nodes.
- Simplifies the passing of a project name for member commands. Instead of having to pass `--to /project/<project name>`, where the only type of route that can be passed to `--to` is only `/project/<project name>`, we now pass directly `--project-name <project name>`.


polvorin (dismissed):
`Deletes all the members of a project before deleting the project itself.`  this part should be removed, as just add to confusion.

polvorin (approved):

adrianbenavides (approved):

PR #8531, opened 2024-10-11
feat(rust): improvements to portals commands outputs

adrianbenavides:
New behaviour:
- The plain output of commands will always be logged, regardless of the passed output format. This fixes the issue where only the machine or json output was logged, lacking the necessary context to understand the command output.

Review output of the commands:
- `project enroll`
- `relay create`
- `tcp-inlet/outlet create`
- `influxdb-inlet/outlet create`

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/relay.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/relay.rs
@@ -37,19 +37,20 @@ impl NodeManagerWorker {
     ) -> Result<Response<RelayInfo>, Response<Error>> {
         let CreateRelay {
             address,
-            alias,
+            name,
             authorized,
             relay_address,
         } = create_relay;
+
         match self
             .node_manager
-            .create_relay(ctx, &address, alias, authorized, relay_address)
+            .create_relay(ctx, &address, name.clone(), authorized, relay_address)
             .await
         {
             Ok(body) => Ok(Response::ok().with_headers(req).body(body)),
             Err(err) => Err(Response::internal_error(
                 req,
-                &format!("Failed to create relay: {}", err),
+                &format!("Failed to create relay at {address} with name {name}. {err}"),
             )),
         }
     }
--- a/implementations/rust/ockam/ockam_api/src/ui/terminal/mod.rs
+++ b/implementations/rust/ockam/ockam_api/src/ui/terminal/mod.rs
@@ -350,62 +350,33 @@ impl<W: TerminalWriter + Debug> Terminal<W, ToStdOut> {
     }
 
     pub fn write_line(mut self) -> Result<()> {
-        // Check that there is at least one output format defined
-        if self.mode.output.plain.is_none()
-            && self.mode.output.machine.is_none()
-            && self.mode.output.json.is_none()
+        let msg = match self
+            .mode
+            .output
+            .get_message(&self.output_format, self.is_tty())?
         {
-            return Err(miette!("At least one output format must be defined"))?;
-        }
-
-        let plain = self.mode.output.plain.clone();
-        let machine = self.mode.output.machine.clone();
-        let json = self.mode.output.json.clone();
-        let (pretty, jq_query) = match self.output_format.clone() {
-            OutputFormat::Json { pretty, jq_query } => (pretty, jq_query),
-            _ => (false, None),
+            Some(msg) => msg,
+            None => return Ok(()),
         };
 
-        let msg = match self.output_format {
-            OutputFormat::Plain => {
-                // If interactive, use the following priority: Plain -> Machine -> JSON
-                if self.stdout.is_tty() {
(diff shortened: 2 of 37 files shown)


implementations/rust/ockam/ockam_api/src/nodes/service/relay.rs
```diff
@@ -37,20 +37,27 @@ impl NodeManagerWorker {
...
         {
             Ok(body) => Ok(Response::ok().with_headers(req).body(body)),
-            Err(err) => Err(Response::internal_error(
-                req,
-                &format!("Failed to create relay: {}", err),
-            )),
+            Err(err) => match err.code().kind {
+                Kind::Timeout => Err(Response::internal_error(
+                    req,
+                    &format!("Failed to create relay at {address} with name {name}. Make sure you are allowed to create this relay. {err}")
```
etorreborre: How could we be even more specific there? How can a user check that they can, or not, create a relay?
The other big improvement would be to get a response telling the user that they not authorized to create a relay, instead of relying on a timeout (but that's obviously outside of the scope of this PR).

I'm also not sure what the error message brings here. This makes the line pretty long, and we still display an
error message with a code 500 (we need to fix that) below.

implementations/rust/ockam/ockam_command/src/project/ticket.rs
```diff
@@ -115,10 +115,16 @@ impl Command for TicketCommand {
...
         // Request an enrollment token that a future member can use to get a
         // credential.
-        let token = authority_node_client
-            .create_token(ctx, attributes, self.expires_in, self.usage_count)
-            .await
-            .map_err(Error::Retry)?;
+        let token = {
+            let pb = opts.terminal.progress_bar();
+            if let Some(pb) = pb.as_ref() {
+                pb.set_message("Creating enrollment ticket...");
```
etorreborre: "Creating _an_ enrollment ticket", maybe?

implementations/rust/ockam/ockam_api/src/ui/terminal/mod.rs
```diff
@@ -401,6 +402,13 @@ impl<W: TerminalWriter + Debug> Terminal<W, ToStdOut> {
             },
         };
 
+        // Log the plain message if present
+        if let Some(plain) = self.mode.output.plain.as_ref() {
+            if msg.ne(plain) {
```
etorreborre: Having to compare full messages here is a bit of a red flag. 

I would suggest to refactor some code in that function:
 
- Have a method on `self.mode.output.get_message(format, is_tty): Result<OutputMessage>`.
- That method returns an error if no output has been selected, otherwise it applies the priority rules
- `OutputMessage` is an enum with alternatives `Plain(String) | Machine(String) | Json(String)`.
- The formatting of the `Json` type can be done directly in `get_message` (or raise an error if formatting can't be done).
- Then pattern-matching on `(OutputMessage, self.output.plain.as_ref())` we can decide:
   - To log both messages if `OutputMessage` is `Json` or `Machine` and `self.output.plain` is defined.
   - To log only `OutputMessage` otherwise.

One advantage is that we can now unit test the priority rules.

etorreborre (dismissed):
🚙  I left a possible idea for a refactoring.

implementations/rust/ockam/ockam_api/src/nodes/service/relay.rs
adrianbenavides: > How can a user check that they can, or not, create a relay?

I think this would require refactoring the inners of how secure channel are created. I think at some point we know/log whether an identity is authorized to create the secure channel, but we always get a timeout as the final error.

But yeah, I agree that this change may be assuming too much about the underlying error. I think I should rollback it.

implementations/rust/ockam/ockam_api/src/ui/terminal/mod.rs
adrianbenavides: I've added this 👍

implementations/rust/ockam/ockam_api/src/ui/terminal/mod.rs
```diff
@@ -350,62 +350,33 @@ impl<W: TerminalWriter + Debug> Terminal<W, ToStdOut> {
...
-                            self.process_json_output(json, pretty, jq_query.as_ref())?
-                        }
-                        (None, None, Some(plain)) => plain,
-                        _ => unreachable!(),
-                    }
-                }
+        // Log the message in the given format and the plain message if present
+        match (&msg, self.mode.output.plain.as_ref()) {
+            // msg == plain, so we log it once
+            (OutputMessage::Plain(msg), Some(_)) => {
```
etorreborre, line 365: Small thing: here we can just go `(OutputMessage::Plain(_), Some(_))` and not even do a message comparison.

etorreborre (approved):

PR #8541, opened 2024-10-16
fix(rust): `project enroll` should not try to fetch project data from orchestrator

adrianbenavides:
- Remove `get_project()` call from `project enroll` command. A machine enrolling using an enrollment ticket won't have access to the controller, so that call will fail. In that case, a warning message is logged (the command doesn't fail).
- In that warning log, print the miette report using the display impl instead of its debug trait. The debug implementation is meant to be shown in stdout only, not in logs.


etorreborre (approved):
🙇

PR #8545, opened 2024-10-17
feat(rust): pretty json output by default, and colored if possible

adrianbenavides:
- replace `--pretty` by `--compact-output` to mimic the `jq` api
- apply colors to json output if possible (TTY and NO_COLOR is not passed)
- ux improvement: if either `--jq` or `--compact-output` args are passed, the `--output-format json` is assumed

[asciicast](https://asciinema.org/a/mB8T2sepaLHcqtT7ZevghwYME)


implementations/rust/ockam/ockam_command/src/global_args.rs
```diff
@@ -36,85 +34,55 @@ pub struct GlobalArgs {
...
     /// jq query to apply to the JSON output of the command
     #[arg(global = true, long = "jq")]
-    pub jq_query: Option<String>,
+    jq_query: Option<String>,
 
-    /// Pretty print the JSON output of the command
+    /// Compact the JSON output of the command
     #[arg(global = true, long)]
-    pub pretty: bool,
+    compact_output: bool,
```
davide-baldo, line 61: Should we add an alias for `--pretty` to keep back-compatibility?

davide-baldo (dismissed):
Looks good to me!

implementations/rust/ockam/ockam_command/src/global_args.rs
adrianbenavides: Yeah good idea, but it will have to be a separate argument. I'll add it in a bit

adrianbenavides: ready now!

etorreborre (approved):

PR #8547, opened 2024-10-18
feat(rust): added optional watchdog for tokio blocking tasks

davide-baldo:
By building with the `watchdog` feature, every node runs a watchdog that monitors tokio tasks, if one thread remains more than 50ms (default) on the same task without calling `await` it's deemed as "stuck", this triggers a multiline warning log entry and pauses the process through a SIGSTOP.

The interval can be set using the `OCKAM_WATCHDOG_INTERVAL` environment variable.
Once the process is paused, it is possible to use the external debugger `rust-lldb` to understand why the thread was stuck.

To build the watchdog feature it's necessary to enable the `tokio_unstable` configuration:
`RUSTFLAGS='--cfg tokio_unstable' cargo build -p  ockam_command  -F ockam_node/watchdog`

Screenshot 2024-10-18 at 17 41 24

^ in the screenshot a fictional example with a synchronous sleep()

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4928,6 +4928,7 @@ dependencies = [
  "heapless 0.8.0",
  "hex",
  "minicbor",
+ "nix 0.29.0",
  "ockam_core",
  "ockam_executor",
  "ockam_macros",
--- a/implementations/rust/ockam/ockam_node/Cargo.toml
+++ b/implementations/rust/ockam/ockam_node/Cargo.toml
@@ -68,6 +68,10 @@ metrics = []
 # message flows within Ockam apps.
 debugger = ["ockam_core/debugger"]
 
+# Feature: "watchdog" reports blocking task that compromise the runtime
+# neeeds to be compiled with RUSTFLAGS="--cfg tokio_unstable"
+watchdog = ["nix"]
+
 storage = [
   "std",
   "time",
@@ -87,6 +91,7 @@ fs2 = { version = "0.4.3", optional = true }
 futures = { version = "0.3.30", default-features = false }
 heapless = { version = "0.8", features = ["mpmc_large"], optional = true }
 minicbor = { version = "0.24.1", features = ["derive"] }
+nix = { version = "0.29", features = ["signal"], optional = true }
 ockam_core = { path = "../ockam_core", version = "^0.118.0", default-features = false }
 ockam_executor = { path = "../ockam_executor", version = "^0.87.0", default-features = false, optional = true }
 ockam_macros = { path = "../ockam_macros", version = "^0.35.0" }
--- a/implementations/rust/ockam/ockam_node/src/lib.rs
+++ b/implementations/rust/ockam/ockam_node/src/lib.rs
@@ -72,6 +72,9 @@ mod worker_builder;
 #[cfg(feature = "std")]
 pub mod runtime;
 
+#[cfg(feature = "watchdog")]
+mod watchdog;
+
 pub use context::*;
 pub use delayed::*;
 pub use error::*;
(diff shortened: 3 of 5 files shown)


etorreborre (dismissed):

implementations/rust/ockam/ockam_node/src/runtime.rs
```diff
@@ -14,3 +14,106 @@ pub fn take() -> Runtime {
         .take()
         .expect("Runtime was consumed")
 }
+
+#[cfg(feature = "watchdog")]
+pub(crate) mod watchdog {
```
SanjoDeundiak: Well worth a separate file IMO

implementations/rust/ockam/ockam_node/src/runtime.rs
```diff
@@ -14,3 +14,106 @@ pub fn take() -> Runtime {
...
+        fn watchdog_loop(mut self, metrics: tokio::runtime::RuntimeMetrics) {
+            info!("Starting tokio runtime watchdog");
+            const WATCHDOG_INTERVAL: Duration = Duration::from_millis(50);
+            let watchdog_interval = std::env::var("OCKAM_WATCHDOG_INTERVAL")
+                .map(|s| Duration::from_millis(s.parse().unwrap()))
+                .unwrap_or(WATCHDOG_INTERVAL);
+
+            loop {
+                // create any missing thread info
+                for worker in 0..metrics.num_workers() {
```
SanjoDeundiak: Took me a while to understand that it's tokio internal metrics and those are not Ockam workers. But not sure how we could articulate that better here

implementations/rust/ockam/ockam_node/src/runtime.rs
```diff
@@ -14,3 +14,106 @@ pub fn take() -> Runtime {
...
+#[cfg(feature = "watchdog")]
+pub(crate) mod watchdog {
+    use std::collections::HashMap;
+    use std::thread::ThreadId;
+    use std::time::Duration;
+    use tokio::runtime::Runtime;
+
+    struct ThreadInfo {
+        poll_count: u64,
+        was_parked: bool,
```
SanjoDeundiak: do you think `is_parked: bool,` makes more sense?

SanjoDeundiak: `was_parked` can be understood as `was_ever_parked`, or maybe it's just me

implementations/rust/ockam/ockam_node/src/runtime.rs
```diff
@@ -14,3 +14,106 @@ pub fn take() -> Runtime {
...
+                            You can attach the rust debugger with:
+                                rust-lldb {exe} --attach-pid {pid}
+                            and then select the thread with:
+                                thread select THREAD_ID
+                            and print the backtrace with:
+                                bt
+                            "
+                        );
+
+                        // Send SIGSTOP to self to pause the process in this precise moment in time
```
SanjoDeundiak: Should that be optional?

SanjoDeundiak (commented):
I think our runtime is a large black box and it's the first step to change that!

implementations/rust/ockam/ockam_node/src/runtime.rs
davide-baldo: I went with `parked_on_previous_iteration`

implementations/rust/ockam/ockam_node/src/runtime.rs
davide-baldo: Without attaching a debugger, you would know there is an issue, but you wouldn't have any other information. 
I couldn't find an easy way to add any meaningful information about the offending thread.
If we could at least print the stack trace, then it would absolutely make sense to make it optional. (I could only find one library, but it appears to be unmaintained)

implementations/rust/ockam/ockam_node/src/runtime.rs
davide-baldo: :) I used `tokio_metrics` and `tokio_worker`

SanjoDeundiak (approved):

PR #8548, opened 2024-10-18
fix(rust): when an item is not found an error is returned

davide-baldo:
When a show command is executed on any resource and that resource is missing, it'll simply print `There are no <type> to show`, return a SUCCESS (0) exit code.

```
$ ockam node show hello
  > There are no nodes to show
$ echo $?
0
```
This could be misleading, and an error should be printed.
Also, some tests rely on the fact that the shown command returning 0 means the resource exists.

With this patch the resource not found triggers an error instead.

```
$ target/debug/ockam node show hello


    Error:
    ────────────────────────────────────────────────────────────────────────────────

    There is no node with name hello (origin: Api, kind: NotFound, source location: implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs:311:17)
    
    Error code: OCK500
    version: 0.134.0
    compiled from: d071009f2ce6f9f7c7be1d04e89308abc8c4420a

    If you need help, please email us on support@ockam.io
    We will promptly help you get unstuck and quickly resolve any problems.

    ────────────────────────────────────────────────────────────────────────────────
$ echo $?
70
```

While it's nice to return a non-successful exit code, this error message could be too much.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/terminal/tui.rs
+++ b/implementations/rust/ockam/ockam_command/src/terminal/tui.rs
@@ -23,7 +23,7 @@ pub trait ShowCommandTui {
     async fn show(&self) -> miette::Result<()> {
         let terminal = self.terminal();
         let items_names = self.list_items_names().await?;
-        if items_names.is_empty() {
+        if items_names.is_empty() && self.cmd_arg_item_name().is_none() {
             terminal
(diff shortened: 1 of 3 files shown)


davide-baldo:
addressed by Adrian in another PR

PR #8549, opened 2024-10-18
feat(rust): tie each tcp connection inside portal to an `Identifier`

SanjoDeundiak:


implementations/rust/ockam/ockam_core/src/identity/identifier.rs
```diff
@@ -0,0 +1,14 @@
...
+/// Identifier LEN. Should be equal to [`ockam_identity::models::IDENTIFIER_LEN`]
+pub const LOCAL_INFO_IDENTIFIER_LEN: usize = 32;
+
+/// Identity SecureChannel LocalInfo unique Identifier
+pub const SECURE_CHANNEL_IDENTIFIER: &str = "SECURE_CHANNEL_IDENTIFIER";
+
+/// Copy of [`ockam_identity::models::IDENTIFIER`]. Copied for decoupling.
+#[derive(Clone, Eq, PartialEq, Hash, Encode, Decode, CborLen, Debug)]
+#[cbor(transparent)]
+pub struct LocalInfoIdentifier(
```
davide-baldo, line 12: Could we make `ockam_identity::models::IDENTIFIER` an alias of this type?

implementations/rust/ockam/ockam_transport_tcp/src/portal/portal_message.rs
```diff
@@ -14,6 +14,9 @@ pub enum PortalMessage<'de> {
     /// or from the target to the Inlet was dropped
     Disconnect,
     /// Message with binary payload and packet counter
+    // TODO: Add route_index. May not be as important as for eBPF portals, as regular portals
```
davide-baldo, line 17: Do we actually need the `route_index` when we already have a generic incremental counter?

davide-baldo (approved):
Is adding a test where the identifier changes and is properly detected possible?

implementations/rust/ockam/ockam_transport_tcp/src/portal/portal_message.rs
SanjoDeundiak: Hm, maybe we could reuse it for that purpose as well. I'm not working on that any time soon though

implementations/rust/ockam/ockam_core/src/identity/identifier.rs
SanjoDeundiak: I would not do that. Overall I wasn't very happy about leaking that abstraction into ockam_core, but didn't figure out anything better. Let's at least keep those types decoupled

PR #8554, opened 2024-10-21
feat(rust): improve parsing of node config files

adrianbenavides:
Improves the parsing logic of config files:
- Env vars defined in the `variables` section are only used in the expand step. After that, they are unset
- Improve tests in the different parsing steps
- Improve handling of the `ENROLLMENT_TICKET` env var

The following json config syntax was already supported out of the box. I added tests to make sure we don't modify this behavior. Here's an example of a valid json configuration file:

```yaml
{
    # comment
    "variables": {
        "var_b": true,
        "var_i": 1, # trailing commas are supported
    },
    "nodes": [
        "node1",
        "node2",
    ],
    "relays": {
        "r1": {
            # comment
            "at": "/project/default",
            "to": "outlet-node",
        }
    }
}
```

- Comments are supported with `#` (c-style comments are not supported)
- Trailing commas are supported

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/run/config.rs
+++ b/implementations/rust/ockam/ockam_command/src/run/config.rs
@@ -41,8 +41,6 @@ pub struct Config {
     pub relays: Relays,
 }
 
-impl ConfigParser<'_> for Config {}
-
 impl Config {
     /// Executes the commands described in the configuration to create the desired state.
     ///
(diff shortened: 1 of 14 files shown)


implementations/rust/ockam/ockam_command/src/run/config.rs
```diff
@@ -79,26 +77,29 @@ impl Config {
     pub async fn parse_and_run(
         ctx: &Context,
         opts: CommandGlobalOpts,
-        contents: &str,
+        contents: &mut String,
```
SanjoDeundiak: I think this is better unless you want to return contents back
```
mut contents: String
```

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_command/src/run/config.rs
adrianbenavides: Right, good catch! I'll update it

SanjoDeundiak (approved):

PR #8559, opened 2024-10-22
feat(rust): check capabilities before using ebpf portals

SanjoDeundiak:


implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/portals.rs
```diff
@@ -13,6 +13,22 @@ use tokio::sync::mpsc::channel;
 use tracing::instrument;
 
 impl TcpTransport {
+    fn check_capabilities() -> Result<()> {
+        use caps::CapSet;
+        use caps::Capability::{CAP_BPF, CAP_NET_RAW};
+        let caps = caps::read(None, CapSet::Effective).unwrap();
```
etorreborre: Can we avoid the `unwrap` here?

etorreborre (dismissed):

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/portals.rs
SanjoDeundiak: Sure!

etorreborre (approved):

PR #8560, opened 2024-10-22
refactor(rust): add more granular scopes for command logs

adrianbenavides:
Change the logic to define the log level and crates filter together so that:
- no verbose: no logs, will fallback to whatever is defined by env vars
- `-v`: level INFO and crates `ockam_command, ockam_api::ui::terminal`. The last crate is necessary to show the contents we write through the `Terminal` abstraction, which includes the command output
- `-vv`: level INFO and the [core crates](https://github.com/build-trust/ockam/pull/8560/files#diff-7bc1797d356e51984fef554b044acc1fbf368347d4eccb49fef297818015c306R396)
- `-vvv`: level DEBUG and the core crates
- `-vvvv`: level TRACE and the core crates
- `-vvvvv`: level TRACE and all crates, including deps

Other important considerations (no changes here, just a reminder):
- The verbose flag takes precedence over env vars
- To fully customize the behavior, we can use `OCKAM_LOG_LEVEL` and `OCKAM_LOG_CRATES_FILTER`

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
+++ b/implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
@@ -1,7 +1,7 @@
 use crate::config::LevelVar;
 use crate::logs::default_values::*;
 use crate::logs::env_variables::*;
-use ockam_core::env::{get_env, get_env_with_default, is_set, FromString};
+use ockam_core::env::{get_env, get_env_with_default, FromString};
 use std::fmt::{Display, Formatter};
 use std::path::PathBuf;
 use tracing_core::Level;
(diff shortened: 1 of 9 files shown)


implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
```diff
@@ -300,57 +269,145 @@ pub fn logging_enabled() -> ockam_core::Result<LoggingEnabled> {
...
+pub struct LogLevelWithCratesFilter {
+    pub level: Level,
+    pub crates_filter: CratesFilter,
+    /// Indicates if the log level was explicitly set by the commands' `verbose` flag
+    pub explicit_verbose_flag: bool,
+}
+
+impl LogLevelWithCratesFilter {
+    /// Create a new LogLevel using one of the environment variables
+    /// (the legacy OCKAM_LOG variable or the current OCKAM_LOG_LEVEL variable).
```
etorreborre: Maybe we can drop the `OCKAM_LOG` variable. It's been a while now.

etorreborre (dismissed):

implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
adrianbenavides: Let's do it!

etorreborre (approved):

PR #8562, opened 2024-10-22
fix(rust): usage of `ENROLLMENT_TICKET` in `node create` and `project enroll` used in a node config

adrianbenavides:
When using the `ENROLLMENT_TICKET` environment variable as part of a node configuration file, a collision was happening, producing a loop of `project enroll` calls within the `node create` command.

The configuration file is parsed correctly, the `project enroll` command was running as expected, but then, when jumping to the next command (i.e. `node create`), it was receiving the enrollment ticket as an argument set automatically from the env var. This triggered the `node create` command to be run in `config` mode, running the `project enroll` command again.

So, this was affecting config files where the enrollment ticket was set using the `ENROLLMENT_TICKET` env var:

```yaml
ticket: ${ENROLLMENT_TICKET}
...
```

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create/config.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create/config.rs
@@ -218,9 +218,31 @@ impl NodeConfig {
         identity_name: &String,
     ) -> miette::Result<()> {
         debug!("Running node config");
-        for section in self.parse_commands(node_name, identity_name)? {
+
+        // Run `project enroll` first
+        if let Some(cmd) = self
+            .project_enroll
+            .clone()
+            .into_parsed_commands(Some(identity_name))?
+            .into_iter()
+            .next()
+        {
+            cmd.run(ctx, opts).await?;
+
+            // Newline before the `node create` command
+            opts.terminal.write_line("")?;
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_command/src/node/create/config.rs
```diff
@@ -218,9 +218,31 @@ impl NodeConfig {
...
+        {
+            cmd.run(ctx, opts).await?;
+
+            // Newline before the `node create` command
+            opts.terminal.write_line("")?;
+            debug!("project enroll command finished");
+
+            // Unset the `ENROLLMENT_TICKET` env var, so that the `node create` command
+            // doesn't try to use it again
+            std::env::remove_var("ENROLLMENT_TICKET");
```
SanjoDeundiak, line 238: Could you please move it to a const in one of your PRs in the future?

SanjoDeundiak (approved):

implementations/rust/ockam/ockam_command/src/node/create/config.rs
adrianbenavides: Yup, I'm preparing another pr to clean this up and add a proper test for this.

PR #8580, opened 2024-10-28
feat(rust): `project ticket` improvements

adrianbenavides:
- add support to `project ticket` and `project enroll` to produce/consume json encoded enrollment tickets
- show a warning when a ticket is generated with high values for both usage/expiration arguments. This won't prevent the user from generating the ticket.
- add plain output to `project ticket` 

image

### DIFF
--- a/implementations/rust/ockam/ockam_app_lib/src/lib.rs
+++ b/implementations/rust/ockam/ockam_app_lib/src/lib.rs
@@ -4,6 +4,7 @@
 //!
 //! It exposes C APIs that can be used by the frontend to interact with the application.
 //!
+#![recursion_limit = "256"]
 
 use thiserror::Error;
 mod api;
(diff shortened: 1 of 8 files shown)


implementations/rust/ockam/ockam_app_lib/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 //!
 //! It exposes C APIs that can be used by the frontend to interact with the application.
 //!
+#![recursion_limit = "256"]
```
davide-baldo, line 7: Why did you need to increase the limit?

davide-baldo (dismissed):

implementations/rust/ockam/ockam_app_lib/src/lib.rs
adrianbenavides: I keep getting a pretty ugly compilation error and that's what the compiler recommended to do to avoid the issue (basically increase that value)... It was caused by a third party library (aws-smithy I think). I was hoping that this would disappear after the recent deps updates we did, but it didn't.

davide-baldo (approved):

PR #8582, opened 2024-10-29
feat(rust): support `node create` first argument as inline configuration

adrianbenavides:
Now we can use the `node create` positional argument to pass an inline configuration. So, the following are equivalents:

```
ockam node create "{name: n1}"
ockam node create --configuration "{name: n1}"
```

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -18,6 +18,7 @@ use ockam_api::{fmt_log, fmt_ok};
 use ockam_core::{opentelemetry_context_parser, OpenTelemetryContext};
 use ockam_node::Context;
 
+use crate::node::config::NodeConfig;
 use crate::node::create::config::ConfigArgs;
 use crate::node::util::NodeManagerDefaults;
 use crate::service::config::Config;
@@ -127,6 +128,7 @@ impl Default for CreateCommand {
                 configuration: None,
                 enrollment_ticket: None,
                 variables: vec![],
+                started_from_configuration: false,
             },
             tcp_listener_address: node_manager_defaults.tcp_listener_address,
             http_server: false,
@@ -200,7 +202,12 @@ impl CreateCommand {
             return false;
         }
 
-        let name_arg_is_a_config = !self.has_name_arg();
+        // Ignore the config args if the node was started from a configuration
+        if self.config_args.started_from_configuration {
+            return false;
+        }
+
+        let name_arg_is_a_config = self.name_arg_is_a_config();
 
         let no_config_args = !name_arg_is_a_config
             && self.config_args.configuration.is_none()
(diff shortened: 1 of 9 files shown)


implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -216,12 +223,14 @@ impl CreateCommand {
             || self.config_args.enrollment_ticket.is_some()
     }
 
-    /// Return true if the `name` argument is not a config file path or URL
+    /// Return true if the `name` argument is not a URL, a file path, or an inline config
```
etorreborre: The comment could say that it is just a name if it's not a URL, a file path, or an inline config.

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -216,12 +223,14 @@ impl CreateCommand {
...
+        let is_url = is_url(&self.name).is_some();
         let is_file = std::fs::metadata(&self.name)
             .map(|m| m.is_file())
             .unwrap_or(false);
-        is_url(&self.name).is_none() && !is_file
+        let is_inline_config = serde_yaml::from_str::<NodeConfig>(&self.name).is_ok();
+        !is_url && !is_file && !is_inline_config
     }
 
     fn parse_args(&self) -> miette::Result<()> {
```
etorreborre, line 240: Is `color_primary("NAME_OR_CONFIGURATION"),` on line 255 still valid?

etorreborre (dismissed):
Approved but I am not a big fan of this because this prevents us from giving precise error messages. For example an incorrect input could be a malformed path but also a configuration file having `inlet:` instead of `tcp-inlet` and we can't tell the difference in terms of error.

I would be more comfortable if we could:

 1. Determine for sure what was the intent of the user.
 2. Display an appropriate message saying: "incorrect name" or "incorrect URL", or ... And giving as info we can on why it's incorrect and how to fix it.

Overall I think it is better to reduce the ambiguity in the user intent for command line arguments, at the expense of verbosity. For example what happens if I enter a file name for a local file `test-config` that does not exist because I made a typo in my file name? Are we going to determine that this is not a file name, thus it must be a node name, and going to silently create a node with that name?

implementations/rust/ockam/ockam_command/src/node/create.rs
adrianbenavides: I think I will rename the function from `has_name_arg` to `name_arg_is_a_config`. Sounds more explicit to me. What do you think? The comment would also look less confusing.

implementations/rust/ockam/ockam_command/src/node/create.rs
adrianbenavides: what do you mean? because of the arg name?

adrianbenavides:
> Approved but I am not a big fan of this because this prevents us from giving precise error messages. For example an incorrect input could be a malformed path but also a configuration file having inlet: instead of tcp-inlet and we can't tell the difference in terms of error.

I don't like it either, but we have to be consistent for now. If the user can run `ockam create ./config.yaml` I wouldn't understand as a user why I can't just run `ockam create "{name: n1, ...}"`. I think we are still inconsistent with the behavior of the `--configuration` argument.

The `node create` positional argument (the current `name`) is slowly being replaced in importance by the configuration. Perhaps in the near future it would make more sense to have an argument for the name, and leave the positional argument for the configuration alone. But right now I think that's not clear and it's best to have more flexibility in the arguments to see what works best.

implementations/rust/ockam/ockam_command/src/node/create.rs
etorreborre: Yes, I was wondering if the `NAME_OR_CONFIGURATION` was describing the right thing, but it does.

etorreborre (approved):

PR #8584, opened 2024-10-29
fix(rust): make sure that traces are exported when a command is executed

etorreborre:
This PR fixes the sending of spans for user commands:

 - We make sure to flush the logs and spans at the end of a command.
 - The cutoff timeout for exporting is set to a higher time to make sure that all spans have the time to be exported (note for the reviewers: I haven't changed the naming of the `CUTOFF` variables but they apply even if we don't use a portal to send traces).

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/util/mod.rs
+++ b/implementations/rust/ockam/ockam_command/src/util/mod.rs
@@ -58,7 +58,7 @@ where
 {
     let (ctx, mut executor) = NodeBuilder::new()
         .no_logging()
-        .with_runtime(opts.rt)
+        .with_runtime(opts.rt.clone())
         .build();
     let res = executor.execute(
         async move {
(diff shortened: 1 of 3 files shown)


implementations/rust/ockam/ockam_command/src/util/mod.rs
```diff
@@ -76,6 +76,7 @@ where
         }
         .with_context(OpenTelemetryContext::current_context()),
     );
+    opts.force_flush();
```
adrianbenavides, line 79: Will this block indefinitely until all traces are exported? Or all the processors we use implement the cutoff logic?

adrianbenavides (approved):

implementations/rust/ockam/ockam_command/src/util/mod.rs
etorreborre: Yes the cutoff logic applies all the time. It was an additional reason why I was not getting those user events.

PR #8586, opened 2024-10-29
feat(rust): implemented key rotation and rekeying for the kafka use case

davide-baldo:
> [!IMPORTANT]
> This PR is toward `kafka-multiple-consumers` branch and not `develop` since it breaks back-compatibility as well as temporary disable the consumer validation.
> Multiple patches will be applied toward the same branch and then merged back into develop once we have solved the back-compatibility.

In this PR the producer is rotating key (a new exchange) as well performing rekeying.
Rekeying and rotation are now hard-coded relatively at 1 minutes and 5+5 minutes, these will be provided by the other party in a future patch.
The rotation mechanism has 2 timestamps: after the first, a new exchange **should** be performed, but the node will keep using the existing key is the exchange fails. The second one is the `validity` timestamp, after which the key will no longer be used no matter what.
The rekeying is much simpler, every minute the key is rotated unless a new exchange happens.

`TestNode` and `start_manager_for_tests` were enhanced to allow a local authority.
`Clock` trait along with `ProductionClock` and `TestClock` was introduced in the `ockam_core` crate for testing purposes. So far, the only way I've found to mock the clock without propagating the template was to wrap it around `Box<dyn Clock>`.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/kafka/key_exchange/controller.rs
+++ b/implementations/rust/ockam/ockam_api/src/kafka/key_exchange/controller.rs
@@ -1,17 +1,20 @@
-use crate::kafka::key_exchange::{KafkaKeyExchangeController, TopicPartition};
+use crate::kafka::key_exchange::KafkaKeyExchangeController;
 use crate::kafka::protocol_aware::KafkaEncryptedContent;
 use crate::kafka::{ConsumerPublishing, ConsumerResolution};
 use crate::nodes::models::relay::ReturnTiming;
 use crate::nodes::NodeManager;
 use ockam::identity::{
-    DecryptionRequest, DecryptionResponse, EncryptionRequest, EncryptionResponse, Identifier,
-    SecureChannels,
+    utils, DecryptionRequest, DecryptionResponse, EncryptionRequest, EncryptionResponse,
+    SecureChannels, TimestampInSeconds,
 };
 use ockam_abac::PolicyAccessControl;
+use ockam_core::compat::clock::{Clock, ProductionClock};
 use ockam_core::compat::collections::{HashMap, HashSet};
-use ockam_core::{async_trait, route, Address};
+use ockam_core::errcode::{Kind, Origin};
+use ockam_core::{async_trait, route, Address, Error};
 use ockam_node::Context;
-use std::sync::Arc;
+use std::sync::{Arc, Weak};
+use time::Duration;
 use tokio::sync::Mutex;
 
 #[derive(Clone)]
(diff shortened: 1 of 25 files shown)


implementations/rust/ockam/ockam_api/src/kafka/key_exchange/secure_channels.rs
```diff
@@ -1,173 +1,121 @@
...
+        } else {
+            Err(Error::new(
+                Origin::Transport,
+                Kind::Internal,
+                "Node Manager is not available",
+            ))
+        }
     }
 
     /// Creates a secure channel from the producer to the consumer needed to encrypt messages.
```
etorreborre, line 50: This comment needs to be adapted a bit.

implementations/rust/ockam/ockam_api/src/kafka/key_exchange/controller.rs
```diff
@@ -170,3 +283,209 @@ impl KafkaKeyExchangeControllerImpl {
...
+
+                let _consumer_secure_channel_controller = create_secure_channel_controller(
+                    test_clock.clone(),
+                    &mut consumer_node,
+                    destination.clone(),
+                    authority.node_manager.identifier(),
+                )
+                .await;
+
+                let first_key_handle = producer_secure_channel_controller
```
etorreborre: s/handle/handler ?

implementations/rust/ockam/ockam_api/src/kafka/key_exchange/secure_channels.rs
```diff
@@ -176,7 +124,103 @@ impl KafkaKeyExchangeControllerImpl {
...
+            last_rekey: now,
+            last_rotation_attempt: TimestampInSeconds(0),
+            rekey_counter: 0,
+        };
+
+        inner
+            .producer_topic_encryptor_map
+            .insert(topic_name.to_string(), encryption_key);
+
+        info!("Successfully exchange new key with {destination} for topic {topic_name}");
```
etorreborre: s/exchange/exchanged

implementations/rust/ockam/ockam_api/src/kafka/key_exchange/controller.rs
```diff
@@ -170,3 +283,209 @@ impl KafkaKeyExchangeControllerImpl {
...
+    use std::sync::Arc;
+    use std::time::Duration;
+    use tokio::runtime::Runtime;
+    use tokio::time::timeout;
+
+    #[test]
+    pub fn rekey_rotation() -> ockam_core::Result<()> {
+        let runtime = Arc::new(Runtime::new().unwrap());
+        let runtime_cloned = runtime.clone();
+        std::env::set_var("OCKAM_LOG", "none");
```
etorreborre: "OCKAM_LOG" is not used anymore, you can write `std::env::set_var("OCKAM_LOGGING", "false");` instead.
I'm going to do a PR to fully remove it from the code and comments.

implementations/rust/ockam/ockam_api/src/kafka/key_exchange/controller.rs
```diff
@@ -127,20 +121,119 @@ impl KafkaKeyExchangeController for KafkaKeyExchangeControllerImpl {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub(crate) struct TopicEncryptionKeyHandler {
```
etorreborre: Naming is hard but somehow I'm not sure if `TopicEncryptionKeyHandler` and `TopicEncryptionKey` are the best names for these (is one really a `Handler` or more like a `Handle`? Is the other one really a `Key` or some `Key` metadata, or some "rotation state"?). 
Maybe we could introduce the fact that one is for the consumer and the other one is for the producer? I'm just writing this in case you can think of alternatives.

etorreborre (approved):
🦫 I don't necessarily have the exact topology of everything in mind but that seems about right to me.

implementations/rust/ockam/ockam_api/src/kafka/key_exchange/controller.rs
davide-baldo: I agree, the naming is off, for now, `TopicEncryptionKey` became `TopicEncryptionKeyState` and `TopicEncryptionKeyHandler` became `TopicEncryptionKey`.
This code will probably change once the producer will not directly contact the consumer and the exchange happens via a different flow.

PR #8591, opened 2024-10-30
test(rust): add bats test to check we can enroll with previous versions

adrianbenavides:
- Creates an enrollment ticket with the current version (i.e. `$OCKAM`)
- Use the enrollment ticket from the previous 5 releases


etorreborre (approved):
Does it take a long time to run all of these?

adrianbenavides:
Around 1m 30s, 20s per version checked.

PR #8594, opened 2024-10-31
Async raw socket read

SanjoDeundiak:
- Remove pnet crate dependency (and others)
 - Open raw socket ourselves
 - Read/Write from/to RawSocket using tokio's AsyncFd


implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/raw_socket/async_fd_packet_reader.rs
```diff
@@ -0,0 +1,163 @@
...
+    fn parse_headers(buffer: &mut [u8]) -> Result<ParseHeadersRes> {
+        if Ipv4Info::HEADER_BASE_LEN > buffer.len() {
+            return Err(TransportError::ParsingHeaders(
+                "Buffer can't fit IPv4 header".to_string(),
+            ))?;
+        }
+
+        let mut offset = 0;
+
+        let ipv4_view = ipv4_header::View::new(&buffer[offset..offset + Ipv4Info::HEADER_BASE_LEN]);
```
davide-baldo, line 47: It would be easier to read if you would use `0` instead of `offset` in the beginning, and declare offset later on.

davide-baldo (approved):
LGTM

implementations/rust/ockam/ockam_transport_tcp/src/ebpf_portal/raw_socket/async_fd_packet_reader.rs
SanjoDeundiak: Maybe, but I wanted to make TCP header and IPv4 header read look similar

PR #8603, opened 2024-11-05
feat(rust): node's http server is enabled by default

adrianbenavides:
- Node's HTTP server is now enabled by default
- The `--http-server` flag is deprecated in favor of `--no-healthcheck`
- If no port is provided, the port 23345 will be assigned. If that port is used, a random one will be assigned

I also found a flakyness source in the command wherever we use the `find_available_port` (and likely in [ockam_api](https://github.com/build-trust/ockam/blob/944a8fd2120a516a8ea8ffd185c67222adc69c51/implementations/rust/ockam/ockam_api/src/address.rs#L58)). In bats tests, where multiple nodes will run that function at the same time, there can be collisions. That can be fixed by delaying the port assignment until we create the TcpListener. I've created a struct to express the [port](https://github.com/build-trust/ockam/pull/8603/files#diff-953f71bae7a5f6e697345f4d5e5b9a45a150590c8e37ad2a15132339b0abc5e0) usage. Here's an [example](https://github.com/build-trust/ockam/pull/8603/files#diff-eb16923d98064a371a647e3cafb769767c25817f700d5773e8415a5c215030d4R39) of its usage relevant to this PR.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/node/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/node/create.rs
@@ -1,22 +1,3 @@
-use std::fmt::Write;
-use std::{path::PathBuf, str::FromStr};
-
-use async_trait::async_trait;
-use clap::Args;
-use colorful::Colorful;
-use miette::{miette, IntoDiagnostic, WrapErr};
-use opentelemetry::trace::TraceContextExt;
-use opentelemetry::KeyValue;
-use regex::Regex;
-use tracing::instrument;
-
-use ockam_api::cli_state::random_name;
-use ockam_api::colors::{color_error, color_primary};
-use ockam_api::terminal::notification::NotificationHandler;
-use ockam_api::{fmt_log, fmt_ok};
-use ockam_core::{opentelemetry_context_parser, OpenTelemetryContext};
-use ockam_node::Context;
-
 use crate::node::config::NodeConfig;
 use crate::node::create::config::ConfigArgs;
 use crate::node::util::NodeManagerDefaults;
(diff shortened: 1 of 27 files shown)


implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -81,9 +81,13 @@ pub struct CreateCommand {
...
-    /// Enable the HTTP server at the given port.
-    #[arg(long, value_name = "PORT", conflicts_with = "http_server")]
-    pub http_server_port: Option<u16>,
+    /// Disable the node's status endpoint that serves the healthcheck endpoint.
+    #[arg(long, value_name = "BOOL", default_value_t = false)]
+    pub no_status_endpoint: bool,
+
+    /// Specify the port that the status endpoint will listen to.
+    #[arg(long, value_name = "PORT", conflicts_with = "no_status_endpoint")]
+    pub status_endpoint_port: Option<u16>,
```
etorreborre, line 95: Could we parse to `Port` directly, with `Port::Explicit` if the port is given and `Port::TryExplicitOrRandom(DEFAULT_NODE_STATUS_ENDPOINT_PORT)` as a default value?

etorreborre (dismissed):

implementations/rust/ockam/ockam_command/src/node/create.rs
adrianbenavides: It must be an option for when the `--no-status-endpoint` flag is passed. I think in the best case we could set the default Some() value, and then overwrite it if`--no-status-endpoint` is passed. Let me check

adrianbenavides: I can't find a way to use `default_value_t` with an optional argument. I'd leave it as it is for now. I've pushed a small change adding some `derives` to the `Port` struct.

etorreborre: > It must be an option for when the --no-status-endpoint flag is passed

It doesn't have to, since we would discard the value later in `get_status_endpoint_port`. Yes, `Some` would work as well.

adrianbenavides: Oh I see what you mean. Let me try that

adrianbenavides: > It doesn't have to, since we would discard the value later in get_status_endpoint_port. Yes, Some would work as well.

I was trying that and I think it has to. If it's not an option, we must provide a default value. We also need a parser function.

```
/// Specify the port that the status endpoint will listen to.
    #[arg(long, value_name = "PORT", default_value_t = status_endpoint_default_value(), value_parser = parse_status_endpoint_port)]
    pub status_endpoint_port: Port,
```

In the parser function, we can't tell whether the given input is the default value (i.e. `node create` vs `node create --status-endpoint-port 23345`) which would lead to an unexpected behavior.

etorreborre: I don't want to make you spend too much time on this but wouldn't the parser just return `Port::Explicit` if a value is passed whereas the default value is `Port::TryExplicitOrRandom(DEFAULT_NODE_STATUS_ENDPOINT_PORT)`?

adrianbenavides: The problem is that they are not independent. The parser receives the default raw value, so we always end up with a `Port::Explicit`

etorreborre: I'm not sure I understand but I trust that you've carefully explored all options (no pun intended) now :-). Thanks!

adrianbenavides: Just to clarify: in the example from above, we have an argument with a default value and a value parser:
- The default value is `Port::TryExplicitOrRandom(DEFAULT_NODE_STATUS_ENDPOINT_PORT)`: clap converts this to a string value using the Display implementation. That is, the actual default value is `"23345"`
- Then we have the parser, which will receive the `"23345"` string. At that point we no longer know whether that value was the default or the provided by the user, so we can't just return a `Port::Explicit`

Does that makes sense?

etorreborre (approved):

PR #8606, opened 2024-11-06
feat(rust): improve delete behavior on different commands

adrianbenavides:
Main changes:
- I've refactored the logic of killing nodes processes. Now, when a node receives the SIGKILL signal correctly
- I've also made some changes in the `reset` command so it doesn't fail when it fails to remove one of the elements (any vault, identity or node)

Minor improvements:
- Commands using the `DeleteCommandTui` interface can now delete the elements in parallel (only enabled for the `node delete` for now)
- I also added this logic to the `project-member delete` command. Doesn't seem to have a big effect though

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
@@ -6,9 +6,6 @@ use std::process;
 use colorful::Colorful;
 use nix::errno::Errno;
 use nix::sys::signal;
-use serde::Serialize;
-use sysinfo::{Pid, ProcessStatus, ProcessesToUpdate, System};
-
 use ockam::identity::utils::now;
 use ockam::identity::Identifier;
 use ockam::tcp::TcpListener;
@@ -16,6 +13,8 @@ use ockam_core::errcode::{Kind, Origin};
 use ockam_core::Error;
 use ockam_multiaddr::proto::{DnsAddr, Node, Tcp};
 use ockam_multiaddr::MultiAddr;
+use serde::Serialize;
+use sysinfo::{Pid, ProcessStatus, ProcessesToUpdate, System};
 
 use crate::cli_state::{random_name, NamedVault, Result};
 use crate::cli_state::{CliState, CliStateError};
@@ -125,7 +124,12 @@ impl CliState {
     pub async fn delete_all_nodes(&self) -> Result<()> {
         let nodes = self.nodes_repository().get_nodes().await?;
         for node in nodes {
-            self.delete_node(&node.name()).await?;
+            if let Err(err) = self.delete_node(&node.name()).await {
+                self.notify_message(fmt_warn!(
+                    "Failed to delete the node {}: {err}",
+                    color_primary(node.name())
+                ));
+            }
         }
         Ok(())
     }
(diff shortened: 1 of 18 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
```diff
@@ -185,27 +189,33 @@ impl CliState {
             // Avoid killing the current process, return successfully instead.
             // This is needed when deleting nodes from another node, for example, during a reset
             if pid == process::id() {
+                debug!(name=%node_name, "node is the current process, skipping sending kill signal");
                 return Ok(());
             }
 
             // Try first with SIGTERM, if it fails, try again with SIGKILL
             let mut err = None;
             for signal in [signal::Signal::SIGTERM, signal::Signal::SIGKILL] {
```
SanjoDeundiak: Could you please consider removing the loop here? I suspect it will take the same number of lines but will be so much easier to follow IMO

implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
```diff
@@ -223,55 +233,61 @@ impl CliState {
...
-                    format!("failed to stop PID `{pid}` with error `{e}`"),
-                ))
-            })?;
-        debug!(name = %node_name, %pid, %signal, "sent stop signal to node's process");
+        let _ = self.send_kill_signal(pid, signal);
 
         // Wait until the node has fully stopped
         let timeout = std::time::Duration::from_millis(100);
-        let max_attempts = std::time::Duration::from_secs(5).as_millis() / timeout.as_millis();
+        let max_attempts = std::time::Duration::from_secs(3).as_millis() / timeout.as_millis();
```
SanjoDeundiak: Could we please leave it as 5 seconds? That's pretty long, but on VMs eBPF portals sometimes take some time to shut down, sometimes even more than 5 seconds. But I would make it 5 seconds and then I will investigate why it's so long

implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
```diff
@@ -223,55 +233,61 @@ impl CliState {
...
         let mut attempts = 0;
-        let mut sys = System::new();
-        let pid = Pid::from_u32(pid.as_raw() as u32);
         loop {
-            sys.refresh_processes(ProcessesToUpdate::All);
-            if sys.process(pid).is_none() {
-                info!(name = %node_name, %pid, "node process exited");
-                self.notify_progress_finish_and_clear();
-                return Ok(());
+            match self.send_kill_signal(pid, signal) {
```
SanjoDeundiak, line 242: Is there a point to send multiple signals of the same type? @davide-baldo told me there is none, but I haven't checked :)

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
adrianbenavides: the first time it always returns OK, it just ack's that the signal was sent to an existing process. You need to keep sending it until it returns the `Errno::ESRCH` error, which signals that the process is effectively gone.

implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
adrianbenavides: Yes sorry, I forgot to roll that back

implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
adrianbenavides: done!

SanjoDeundiak (approved):

PR #8618, opened 2024-11-11
feat(rust): improvements to portals commands arguments

adrianbenavides:
- Inlet `create` commands now have a positional argument to define the portal's name. This argument replaces the current `alias` or `addr`. Old arguments are kept as deprecated flags. To ensure backward compatibility, if a deprecated flag is used, it will take precedence over the positional argument
- Outlet `create` commands also have the positional argument, as an alternative to the `--from` argument. In this case, we'll keep `--from` as a valid alternative (not deprecated).
- InfluxDB commands now have their own args definition. This fixes the issue that was showing the TCP commands docs in the InfluxDB commands docs
- The arguments using `HostnamePort` now use `SchemeHostnamePort`, allowing the user to optionally specify the scheme (tcp, tls or udp). The `--tls` and `--udp` flags have been deprecated (but kept), as this information can now be derived from the address arguments. To ensure backward compatibility, if a deprecated flag is used, it will take precedence over an address with an explicit scheme
- Other args renames (with old names as valid aliases)
- Unhide `--allow` flag in all the commands


implementations/rust/ockam/ockam_api/src/nodes/registry.rs
```diff
@@ -346,6 +346,10 @@ mod tests {
     }
 
     fn outlet_info(worker_addr: Address) -> OutletInfo {
-        OutletInfo::new(HostnamePort::new("127.0.0.1", 0), Some(&worker_addr), true)
+        OutletInfo::new(
+            HostnamePort::new("127.0.0.1", 0).unwrap(),
```
etorreborre, line 350: Maybe we could make a function `HostnamePort::localhost(u16)` that is guaranteed to succeed, instead of unwrapping here.

implementations/rust/ockam/ockam_command/src/influxdb/inlet/static/long_about.txt
```diff
@@ -1 +1 @@
-A Http inlet is a way of defining where a node should be listening for connections, and where it should forward that traffic to. It is one end (tcp-outlet being the other) of a portal, which receives Http data, attach an authorization token header to it, chunks and wraps them into Ockam Routing messages and sends them along the supplied route.
+An InfluxDB Inlet is a way of defining where a node should be listening for connections, and where it should forward that traffic to. It is one end (influxdb-outlet being the other) of a portal, which receives Http data, attach an authorization token header to it, chunks and wraps them into Ockam Routing messages and sends them along the supplied route.
```
etorreborre: "Chunks and wraps _them_" feels a bit weird to me. Shouldn't we keep "it"?

implementations/rust/ockam/ockam_command/src/influxdb/outlet/create.rs
```diff
@@ -1,24 +1,69 @@
...
+    /// Address of your InfluxDB Outlet, which is part of a route used in other commands.
+    /// This unique address identifies the InfluxDB Outlet worker on the Node on your local machine.
+    /// Examples are `/service/my-outlet` or `my-outlet`.
+    /// If not provided, `outlet` will be used, or a random address will be generated if `outlet` is taken.
+    /// You will need this address when creating a InfluxDB Inlet using `ockam influxdb-inlet create`.
+    #[arg(value_parser = extract_address_value)]
+    pub name: Option<String>,
+
+    /// Address where your InfluxDB server is running, in the form of `<scheme>://<hostname>:<port>`.
+    /// At least the port must be provided. The default scheme is `tcp` and the default address is `127.0.0.1`.
```
etorreborre: I guess it should be: "and the default hostname".

implementations/rust/ockam/ockam_command/src/influxdb/outlet/create.rs
```diff
@@ -1,24 +1,69 @@
...
+    /// You can check the fallback policy with `ockam policy show --resource-type tcp-outlet`.
+    #[arg(
+        long,
+        visible_alias = "expression",
+        display_order = 904,
+        id = "POLICY_EXPRESSION"
+    )]
+    pub allow: Option<PolicyExpression>,
+
+    /// Use eBPF and RawSocket to access TCP packets instead of TCP data stream.
```
etorreborre, line 63: I think it would be nice to either point to some documentation, or to explain that the overall goal is to support an end to end TCP connection, even across several Ockam nodes.

implementations/rust/ockam/ockam_command/src/kafka/consumer/create.rs
```diff
@@ -22,14 +22,17 @@ pub struct CreateCommand {
     /// The local address of the service
     #[arg(long, default_value_t = kafka_inlet_default_addr())]
     addr: String,
-    /// The address where to bind and where the client will connect to alongside its port, <address>:<port>.
-    /// In case just a port is specified, the default loopback address (127.0.0.1) will be used
-    #[arg(long, default_value_t = kafka_default_consumer_server(), value_parser = hostname_parser)]
-    bootstrap_server: HostnamePort,
+
+    /// The address and port where the client will connect, in the format <address>:<port>.
```
etorreborre: We should say something like like one of the arguments above, to mention "scheme" and "hostname":
```
Address where your the client will connect, in the form of `<scheme>://<hostname>:<port>`.
```

implementations/rust/ockam/ockam_command/src/kafka/inlet/create.rs
```diff
@@ -14,38 +21,33 @@ use ockam_api::nodes::models::services::{StartKafkaInletRequest, StartServiceReq
...
-    /// The local address of the service
+    /// [DEPRECATED] Use the <NAME> positional argument instead
     #[arg(long, default_value_t = kafka_inlet_default_addr())]
     pub addr: String,
 
-    /// The address where to bind and where the client will connect to alongside its port, <address>:<port>.
-    /// In case just a port is specified, the default loopback address (127.0.0.1:4000) will be used
-    #[arg(long, default_value_t = kafka_default_inlet_bind_address(), value_parser = hostname_parser)]
-    pub from: HostnamePort,
+    /// The address and port where the client will connect, in the format <address>:<port>.
```
etorreborre: Same thing here, we should mention "scheme" and "hostname".

implementations/rust/ockam/ockam_command/src/influxdb/inlet/create.rs
```diff
@@ -1,25 +1,148 @@
...
+pub struct CreateCommand {
+    /// Assign a name to this InfluxDB Inlet
+    #[arg(id = "NAME", value_parser = alias_parser)]
+    pub name: Option<String>,
+
+    /// Node on which to start the InfluxDB Inlet.
+    #[arg(long, display_order = 900, id = "NODE_NAME", value_parser = extract_address_value)]
+    pub at: Option<String>,
+
+    /// Address on which to accept InfluxDB connections, in the format <scheme>://<address>:<port>.
```
etorreborre: I think that we should keep the name `Address` for the full `<scheme>://<hostname>:<port>` format and then refer to each component separately with those names. For example "the default hostname is 127.0.0.1". 

Or maybe even say `base URL` instead of `Address`? (it's not really a URL because there's no resource path). This would leave the name `Address` for our multi-addresses like `/project/default`.

implementations/rust/ockam/ockam_command/src/kafka/outlet/create.rs
```diff
@@ -27,26 +27,35 @@ use crate::{
 /// Create a new Kafka Outlet
 #[derive(Clone, Debug, Args)]
 pub struct CreateCommand {
+    /// Address of your Kafka Outlet, which is part of a route used in other commands.
+    /// This unique address identifies the Kafka Outlet worker on the Node on your local machine.
+    /// Examples are `/service/my-outlet` or `my-outlet`.
```
etorreborre, line 34: Can those 2 be really used interchangeably? If `${name}` is going to be injected somewhere, it will have to be of one form or the other I guess? Unless we always prepend `/service` when it doesn't exist? Maybe we should have a dedicated type instead of `String` for those things.

etorreborre (dismissed):
Sorry if my comments are a bit out of order. This is a good step forward but I think that could be even clearer in the naming of "addresses" so that there no ambiguity in the description on what an "address" is. Feel free to open another PR for modification.

implementations/rust/ockam/ockam_command/src/kafka/outlet/create.rs
adrianbenavides: This command is missing a value parser actually. It should use the `extract_address_value` to parse the multiaddr into a valid `Address`.

adrianbenavides:
I've created issues for the open comments: https://github.com/build-trust/ockam/issues/8710, https://github.com/build-trust/ockam/issues/8711, https://github.com/build-trust/ockam/issues/8712

implementations/rust/ockam/ockam_command/src/influxdb/inlet/create.rs
adrianbenavides: I've updated args docs to use the same words consistently, not sure if that's all you meant here.

etorreborre (dismissed):

etorreborre (approved):

PR #8621, opened 2024-11-11
chore(rust): improve output for portals creation

SanjoDeundiak:


implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
```diff
@@ -239,36 +233,56 @@ impl Command for CreateCommand {
...
+                color_primary(&cmd.to));
+
+            msg += &fmt_info!("It will retry to connect automatically\n");
+
+            if privileged {
+                msg += &privileged_mode_str;
+            }
+
+            msg
         };
```
adrianbenavides, line 266: You could move the privileged message out here:

```
let mut plain = { ... }
if privileged {
    plain += &privileged_mode_str;
}
```

SanjoDeundiak: Nice catch, thank you

etorreborre (dismissed):

etorreborre (approved):

PR #8625, opened 2024-11-12
feat(rust): review cli node shutdown

adrianbenavides:
- Background nodes were using tracing without a cutoff value. There are now env vars to control that with default values, just as we do with foreground nodes
- Foreground nodes were exiting differently than background nodes. See [here](https://github.com/build-trust/ockam/pull/8625/files#diff-11d072980aa61e0db956f9b7339d26c69ed4db11a2e0714c635ddd610485d283L152-L159). That logic is not relevant anymore, as the `opts.shutdown();` is called at the [end](https://github.com/build-trust/ockam/blob/0d88c208cc78ee0bf851d19a784ff911835b144f/implementations/rust/ockam/ockam_command/src/command.rs#L102) of every command
- Spawned processes are now forked from the main process. This aims to fix the problems of killing processes on docker 
- Fixes an important issue when recreating nodes. In current develop, when the first node is stopped (e.g. `node create --foreground`), is left as the default node, so no node could take the default role unless manually set with the `node default` command. This is now changed so that, when a node checks if it should be the default node, it also takes into account wether the current default node is stopped
- Related to the previous, now the default node name used in `node create` will reuse the stopped node name if possible. That means that running `ockam node create --foreground`, then exiting, and again `ockam node create --foreground` will result in using the same node name.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/logs/env_variables.rs
+++ b/implementations/rust/ockam/ockam_api/src/logs/env_variables.rs
@@ -108,14 +108,20 @@ pub(crate) const OCKAM_SPAN_EXPORT_QUEUE_SIZE: &str = "OCKAM_SPAN_EXPORT_QUEUE_S
 pub(crate) const OCKAM_LOG_EXPORT_QUEUE_SIZE: &str = "OCKAM_LOG_EXPORT_QUEUE_SIZE";
 
 /// Maximum time for sending a log batch and not waiting for a response when running
-/// a foreground command and using a portal to export log records. For example: 200ms
-pub(crate) const OCKAM_FOREGROUND_LOG_EXPORT_PORTAL_CUTOFF: &str =
-    "OCKAM_FOREGROUND_LOG_EXPORT_PORTAL_CUTOFF";
+/// a foreground command export log records. For example: 200ms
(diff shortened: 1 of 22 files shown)


implementations/rust/ockam/ockam_api/src/logs/env_variables.rs
```diff
@@ -108,13 +108,23 @@ pub(crate) const OCKAM_SPAN_EXPORT_QUEUE_SIZE: &str = "OCKAM_SPAN_EXPORT_QUEUE_S
 pub(crate) const OCKAM_LOG_EXPORT_QUEUE_SIZE: &str = "OCKAM_LOG_EXPORT_QUEUE_SIZE";
 
 /// Maximum time for sending a log batch and not waiting for a response when running
-/// a foreground command and using a portal to export log records. For example: 200ms
-pub(crate) const OCKAM_FOREGROUND_LOG_EXPORT_PORTAL_CUTOFF: &str =
+/// a foreground command export log records. For example: 200ms
+pub(crate) const OCKAM_FOREGROUND_LOG_EXPORT_CUTOFF: &str =
     "OCKAM_FOREGROUND_LOG_EXPORT_PORTAL_CUTOFF";
```
etorreborre: The `PORTAL` word needs to be removed in the string as well.

adrianbenavides: done 👍

etorreborre (approved):

PR #8635, opened 2024-11-13
chore(rust): start using jemalloc to avoid memory fragmentation issues

davide-baldo:


davide-baldo:
It seems like the memory fragmentation only happens on macOS

PR #8641, opened 2024-11-14
feat(rust): rewrote router into a state container

davide-baldo:
This PR fully rewrites the Router in a simple shared state rather than being an actor.
I've also tried to simplify the structure and make it more readable.
Every unused feature was removed (like waiting for a worker/processor to become ready or the immediate shutdown) since it's unused ATM.


implementations/rust/ockam/ockam_node/src/router/record.rs
```diff
@@ -317,48 +469,50 @@ impl AddressRecord {
     }
 
     /// Signal this worker to stop -- it will no longer be able to receive messages
-    pub async fn stop(&mut self) -> Result<()> {
+    /// Since the worker is held behind a lock, we cannot hold the synchronous lock
+    /// and call an `.await` to send the signal.
+    /// To avoid this shortcoming, we return a future that can be awaited on instead.
+    pub fn stop(&self) -> Option<impl core::future::Future<Output = Result<()>>> {
+        if self.state.load(Ordering::Relaxed) != AddressState::Running as u8 {
```
SanjoDeundiak, line 476: Recently I read some docs on Ordering and it looked like using Relaxed almost everywhere like we currently do might not be safe. But I haven't finished my investigation yet. FYI

SanjoDeundiak (approved):

davide-baldo:
Superseded by this PR: https://github.com/build-trust/ockam/pull/8739

PR #8658, opened 2024-11-20
feat: extend boolean expressions to allow key-value and identifiers

adrianbenavides:
The policies boolean expression now support key-value and identifiers:

- `a=b` is transformed into `(= subject.a "b")`
- `a="the value"` is transformed into `(= subject.a "the value")`
- `I228786ae` is transformed into `(= subject.identifier \"I228786ae\")`

### DIFF
--- a/implementations/rust/ockam/ockam_abac/src/boolean_expr.rs
+++ b/implementations/rust/ockam/ockam_abac/src/boolean_expr.rs
@@ -39,12 +39,16 @@ pub enum BooleanExpr {
     #[n(0)]
     Name(#[n(0)] String),
     #[n(1)]
-    Or(#[n(0)] Box<BooleanExpr>, #[n(1)] Box<BooleanExpr>),
+    NameValue(#[n(0)] String, #[n(1)] String),
     #[n(2)]
-    And(#[n(0)] Box<BooleanExpr>, #[n(1)] Box<BooleanExpr>),
+    Identifier(#[n(0)] String),
     #[n(3)]
-    Not(#[n(0)] Box<BooleanExpr>),
+    Or(#[n(0)] Box<BooleanExpr>, #[n(1)] Box<BooleanExpr>),
     #[n(4)]
+    And(#[n(0)] Box<BooleanExpr>, #[n(1)] Box<BooleanExpr>),
+    #[n(5)]
+    Not(#[n(0)] Box<BooleanExpr>),
+    #[n(6)]
     Empty,
 }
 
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_abac/src/boolean_expr.rs
```diff
@@ -293,22 +327,48 @@ mod parsers {
...
+            // the next characters can be alphanumerical, dots, underscores or dashes
+            take_while(0.., |c| {
+                AsChar::is_alphanum(c) || c == '.' || c == '_' || c == '-'
+            }),
+        )
+            .context(StrContext::Expected(NAME_FORMAT.into()))
+            .map(|(first_char, rest): (&str, &str)| format!("{first_char}{rest}"))
+            .parse_next(input)?;
+
+        if Identifier::from_string(&name).is_ok() {
```
etorreborre, line 355: This rule can probably go before the name parsing since our Identifiers are proper names.

etorreborre (dismissed):
I'm slightly uncomfortable with the ambiguity between `Name` and `Identifier`.
This PR means that I won't be able to use an identifier as an attribute name. I don't really see why a user would want to do that but people sometimes come with clever schemes.
I guess that this is ok, as long as we can revert to a more explicit policy description with `(= subject.attribute_name attribute_value)`

implementations/rust/ockam/ockam_abac/src/boolean_expr.rs
adrianbenavides: But we first need to cut the piece of the string we want to process. For example, if `input = "I228786ae and a"`, the name parsing will grab the `I228786ae` piece.

adrianbenavides: > I'm slightly uncomfortable with the ambiguity between Name and Identifier.

I agree... I feel like this introduces some assumptions in the boolean expression that limit its usage. The user can always use the full expression to overcome those limits though.

I'm going to extend one of the tests to make explicit one limitation: the current key-value parsing logic doesn't allow quoted values like `a="the value"`. I guess we could add support for it, but I couldn't find a way to do it.

adrianbenavides: I added support for quoted values and tests for these cases, that solves the last part of my previous comment.

etorreborre (dismissed):

etorreborre (approved):

PR #8680, opened 2024-11-27
feat(rust): portals commands feedback improvements

adrianbenavides:
Improves the logs at the `-vv` level for node/relay/inlet/outlet commands.

The goal of these changes is to show the most relevant information, and move the rest to the next verbosity level.

### Log format
This PR also adds a custom log format, to show the event contents before the scope, and not the other way around:

```
# before
2024-12-10T15:22:03.822769Z DEBUG foreground_mode{node_name="n1"}:new in-memory node{node_name="n1"}:create_node_manager{node_name="n1"}: ockam_abac::policy::storage::resource_type_policy_repository_sql: create a repository for resource type policies

# now
2024-12-10T15:22:03.822769Z DEBUG create a repository for resource type [policies foreground_mode{node_name="n1"}:new in-memory node{node_name="n1"}:create_node_manager{node_name="n1"}: ockam_abac::policy::storage::resource_type_policy_repository_sql] 
```

### TCP Inlet and Relay status updates

The command now shows messages when an inlet/relay loses/restores the connection:

image

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/relay.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/relay.rs
@@ -204,7 +204,6 @@ impl NodeManager {
 
         info!(
             %alias, %address, ?authorized, ?relay_address,
-            forwarding_route = ?relay_info.forwarding_route(),
             remote_address = ?relay_info.remote_address(),
             "relay created"
         );
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/relay.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/relay.rs
@@ -130,12 +130,13 @@ impl NodeManager {
     pub async fn create_relay(
         self: &Arc<Self>,
         ctx: &Context,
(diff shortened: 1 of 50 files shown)


implementations/rust/ockam/ockam_api/src/nodes/service/relay.rs
```diff
@@ -182,29 +183,32 @@ impl NodeManager {
...
 
         self.registry
             .relays
-            .insert(alias, registry_relay_info.clone())
+            .insert(alias.clone(), registry_relay_info.clone())
             .await;
 
+        info!(
+            %alias, %address, ?authorized, ?relay_address,
+            forwarding_route = ?relay_info.forwarding_route(),
```
etorreborre: Is the forwarding route really useful? It contains a worker address which is only available at the `DEBUG` level.

adrianbenavides: true, the `remote_address` field already shows the information we need. I'll remove it

etorreborre (approved):

PR #8684, opened 2024-12-02
performance-guideline for rust

davide-baldo:


implementations/rust/performance-guideline.md
```diff
@@ -0,0 +1,344 @@
...
+  self.item = item.clone();
+}
+
+// Good
+fn example(item: Type) {
+  self.item = item;
+}
+```
+
+The same can be applied to constructors, if the instance needs to own the element, ask for an owned element in the constructor.
```
SanjoDeundiak: Maybe we can omit this. Constructor isn't different from any other function

implementations/rust/performance-guideline.md
```diff
@@ -0,0 +1,344 @@
...
+
+```rust
+// Bad
+let value = some_function().ok_or(Error::new("Error"));
+
+// Good
+let value = some_function().ok_or_else(|| Error::new("Error"));
+```
+
+Also, it seems that the compiler always implicitly converts the error when used in conjunction with `?`, triggering an allocation. When fixed, clippy may  complain of an `unnecessary_lazy_evaluations` in this case, but it's a false positive.
```
SanjoDeundiak: Double space

implementations/rust/performance-guideline.md
```diff
@@ -0,0 +1,344 @@
...
+impl Type {
+  fn new(item: Type) -> Self {
+    Self {
+      item,
+    }
+  }
+}
+```
+
+### Setters returns references
```
SanjoDeundiak: I haven't noticed that pattern anywhere before. I used to either return nothing, or a reference to self so that different setter calls could be chained. What's the idea behind returning the reference to the value being set?

SanjoDeundiak (commented):
Thanks for doing that!

adrianbenavides:
Awesome content, thanks for putting it together! 🙏

implementations/rust/performance-guideline.md
davide-baldo: Most of the time you only need a reference to an object, but if you use it in a setter you "lose it".
Now you are faced with two choices, either you add a getter or clone the object before passing it to the setter.

davide-baldo:
In the last push I also removed the section about the implicit error conversion with `ok_or`, the semantic of rust shouldn't allow any implicit allocation. Even if the current behavior is "broken", it's probably a situational compiler issue.

implementations/rust/performance-guideline.md
SanjoDeundiak: Do we have an example where you need a reference to an object? If it's not common, I would say we can just use a getter to get the reference

SanjoDeundiak: Also, given this document is about performance specifically, I would say - add a getter and that's it.

SanjoDeundiak (approved):

PR #8685, opened 2024-12-02
UDP updates

SanjoDeundiak:
1. Added some degree of robustness to UDP transport, so that messages are cached and reassembled in case they arrive out-of-order. (But no guarantee for order is added)
2. Add UDP transport to node so that 2 nodes can communicate via UDP
3. Refactor multiaddr resolution

This PR changes message format for UDP transport, so it's a breaking change for UDP


implementations/rust/ockam/ockam_api/src/multiaddr_resolver/remote_resolver.rs
```diff
@@ -0,0 +1,257 @@
...
+            connection: connection_res,
+            route: rb.into(),
+        })
+    }
+
+    async fn connect_tcp(&self, ma: &MultiAddr, peer: String) -> Result<TcpConnection> {
+        let tcp = self
+            .tcp
+            .as_ref()
+            .ok_or_else(|| invalid_multiaddr_error(ma))?;
```
adrianbenavides: Perhaps a more specific error here would be better, saying that no TCP transport is defined.

implementations/rust/ockam/ockam_api/src/multiaddr_resolver/remote_resolver.rs
```diff
@@ -0,0 +1,257 @@
...
+                    ),
+                )
+            })
+    }
+
+    async fn connect_udp(&self, ma: &MultiAddr, peer: &str) -> Result<UdpBind> {
+        let udp = self
+            .udp
+            .as_ref()
+            .ok_or_else(|| invalid_multiaddr_error(ma))?;
```
adrianbenavides: Same as above regarding the error.

implementations/rust/ockam/ockam_api/src/multiaddr_resolver/remote_resolver.rs
```diff
@@ -0,0 +1,257 @@
...
+                    Kind::Unsupported,
+                    format!("UDP hops are not allowed. Multiaddr={}", ma),
+                ));
+            }
+
+            return Ok((UDP, port.0));
+        }
+
+        // Should not happen
+        Err(invalid_multiaddr_error(ma))
```
adrianbenavides: I think it's not entirely obvious unless you backtrack the code path. Could you add in the comment some more context? Perhaps also adding an example of the expected `ma` structure would also help.

implementations/rust/ockam/ockam_api/src/multiaddr_resolver/remote_resolver.rs
```diff
@@ -0,0 +1,257 @@
...
+            let connection = self.connect_tcp(ma, peer).await?;
+            return Ok(RemoteMultiaddrResolverConnection::Tcp(connection));
+        }
+
+        if transport_type == UDP {
+            let connection = self.connect_udp(ma, &peer).await?;
+            return Ok(RemoteMultiaddrResolverConnection::Udp(connection));
+        }
+
+        // Should not happen
```
adrianbenavides: I'd say directly explain how we got here. Something like `// Unsupported transport`

implementations/rust/ockam/ockam_api/src/nodes/connection/plain_udp.rs
```diff
@@ -0,0 +1,69 @@
...
+use crate::nodes::connection::{Changes, ConnectionBuilder, Instantiator};
+use crate::{RemoteMultiaddrResolver, RemoteMultiaddrResolverConnection, ReverseLocalResolver};
+
+use crate::nodes::NodeManager;
+use ockam_core::{async_trait, Error, Route};
+use ockam_multiaddr::proto::{DnsAddr, Ip4, Ip6, Udp};
+use ockam_multiaddr::{Match, MultiAddr, Protocol};
+use ockam_node::Context;
+
+/// Creates the tcp connection.
```
adrianbenavides, line 11: typo on `tcp`. There are a couple more in the file

implementations/rust/ockam/ockam_api/src/nodes/service/manager.rs
```diff
@@ -142,13 +142,19 @@ impl NodeManager {
...
         let mut s = Self {
             cli_state,
             node_name,
             node_identifier,
-            api_transport_flow_control_id: transport_options.api_transport_flow_control_id,
-            tcp_transport: transport_options.tcp_transport,
-            udp_transport: transport_options.udp_transport.clone(),
+            api_transport_flow_control_ids,
+            tcp_transport: transport_options.tcp.transport,
+            udp_transport: transport_options.udp.as_ref().map(|u| u.transport.clone()),
```
adrianbenavides: I think you could avoid this clone with `transport_options.udp.take()` and get the reference from `s` below in line 177.

implementations/rust/ockam/ockam_api/src/nodes/service/manager.rs
```diff
@@ -657,23 +671,39 @@ pub struct ApiTransport {
     pub flow_control_id: FlowControlId,
 }
 
+#[derive(Debug)]
+pub struct NodeManagerTransportOptionsS<T> {
```
adrianbenavides: Maybe this could be renamed to something simpler like `NodeManagerTransport`

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -71,6 +71,18 @@ pub struct CreateCommand {
...
+    /// Once the node is created, its services can be accessed via this address.
+    /// By default, it binds to 127.0.0.1:0 to assign a random free port.
+    #[arg(
+        display_order = 900,
+        long,
+        short,
+        id = "SOCKET_ADDRESS_UDP",
+        default_value = "127.0.0.1:0"
+    )]
+    pub udp_listener_address: String,
```
adrianbenavides, line 84: Once we merge https://github.com/build-trust/ockam/pull/8618 we can merge the udp and tcp arguments.

implementations/rust/ockam/ockam_transport_udp/src/workers/pending_messages/mod.rs
```diff
@@ -0,0 +1,9 @@
+mod peer_pending_routing_message_storage;
+mod pending_message;
+mod pending_routing_message_storage;
+mod transport_messages_iterator;
+
+pub(crate) use peer_pending_routing_message_storage::*;
+pub(crate) use pending_message::*;
+pub(crate) use pending_routing_message_storage::*;
+pub(crate) use transport_messages_iterator::*;
```
adrianbenavides, line 9: Would it be very hard to test this module?

implementations/rust/ockam/ockam_api/src/multiaddr_resolver/reverse_local_resolver.rs
```diff
@@ -0,0 +1,31 @@
+use ockam_core::{Address, Result, Route, LOCAL};
+use ockam_multiaddr::proto::Service;
+use ockam_multiaddr::MultiAddr;
+
+use crate::error::ApiError;
+
+pub struct ReverseLocalResolver;
```
davide-baldo: To me, the name seems a bit misleading since there is no `resolution` involved, what about `conversion`?

implementations/rust/ockam/ockam_transport_udp/src/messages/routing_number.rs
```diff
@@ -57,6 +58,30 @@ impl Ord for RoutingNumber {
     }
 }
 
+impl AddAssign<u16> for RoutingNumber {
+    fn add_assign(&mut self, rhs: u16) {
+        self.0 = self.0.overflowing_add(rhs).0;
```
davide-baldo: You could simplify by using [wrapping_add](https://doc.rust-lang.org/std/primitive.u32.html#method.wrapping_add) directly.

implementations/rust/ockam/ockam_transport_udp/src/messages/routing_number.rs
```diff
@@ -57,6 +58,30 @@ impl Ord for RoutingNumber {
...
+        s += rhs;
+        s.0
+    }
+}
+
+impl Sub<Self> for RoutingNumber {
+    type Output = u16;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        self.0.overflowing_sub(rhs.0).0
```
davide-baldo: same with [wrapping_sub](https://doc.rust-lang.org/std/primitive.u32.html#method.wrapping_sub)

implementations/rust/ockam/ockam_transport_udp/src/workers/pending_messages/peer_pending_routing_message_storage.rs
davide-baldo, line 1: Creating a rust unit test should be straightforward

davide-baldo (dismissed):
Looks good to me.
It would be nice to have a basic `/udp/` transport bat test

implementations/rust/ockam/ockam_api/src/multiaddr_resolver/remote_resolver.rs
SanjoDeundiak: This code path is unreachable, since by the time this function is called we already checked that we have tcp transport, but I didn't want to use unwrap to be safe. I'll add more details to the comment. I don't think we should add a specific error for the unreachable path though

implementations/rust/ockam/ockam_transport_udp/src/workers/pending_messages/mod.rs
SanjoDeundiak: No, I don't think so. But not a part of this PR unfortunately

implementations/rust/ockam/ockam_api/src/nodes/service/manager.rs
SanjoDeundiak: Nice catch, I need to rename it

implementations/rust/ockam/ockam_transport_udp/src/workers/pending_messages/peer_pending_routing_message_storage.rs
SanjoDeundiak: Will add in a separate PR

implementations/rust/ockam/ockam_api/src/multiaddr_resolver/remote_resolver.rs
SanjoDeundiak: Actually, I read the code again, and it can happen :D

implementations/rust/ockam/ockam_transport_udp/src/workers/pending_messages/peer_pending_routing_message_storage.rs
SanjoDeundiak: Ended up adding to this PR

implementations/rust/ockam/ockam_transport_udp/src/workers/pending_messages/mod.rs
SanjoDeundiak: Ended up adding to this PR

davide-baldo (approved):

adrianbenavides (approved):

PR #8698, opened 2024-12-12
feat(rust): simplify `node create` execution

adrianbenavides:
- A foreground node with a configuration file runs all the commands within the same process.
- A new env `OCKAM_SQLITE_IN_MEMORY` and flag `--in-memory` have been added to run foreground nodes using an in-memory database
- Background nodes don't support the in-memory feature


SanjoDeundiak (changes requested):
Looks good. Few comments:
1. Could you please add new env variable to `implementations/rust/ockam/ockam_command/src/environment/static/env_info.txt`?
2. Do you think we need to add an argument to the node create command that would allow to use in-memory mode without specifying the env variable? I think we have it for other arguments

adrianbenavides:
> Do you think we need to add an argument to the node create command that would allow to use in-memory mode without specifying the env variable? I think we have it for other arguments

Specifying it via an env var is consistent with the rest of database-related options. If we add the flag, should we also add arguments to define all the postgres variables? We can add the in-memory flag as a hidden argument while we decide that

implementations/rust/ockam/ockam_api/src/logs/setup.rs
```diff
@@ -135,9 +135,6 @@ impl LoggingTracing {
             result.expect("Failed to initialize tracing subscriber");
         };
 
-        // the global error handler prints errors when exporting spans or log records fails
-        set_global_error_handler(logging_configuration);
```
SanjoDeundiak: Is this intended?

implementations/rust/ockam/ockam_command/src/command.rs
```diff
@@ -58,6 +57,17 @@ impl OckamCommand {
         // This allows us to customize how we format the error messages and their content.
         let _hook_result = miette::set_hook(Box::new(|_| Box::new(ErrorReportHandler::new())));
 
+        let command_name = self.subcommand.name();
+
+        // Set the in-memory env var if needed
+        if let OckamSubcommand::Node(cmd) = &self.subcommand {
+            if let crate::node::NodeSubcommand::Create(c) = &cmd.subcommand {
+                if c.in_memory {
+                    std::env::set_var(OCKAM_SQLITE_IN_MEMORY, "true");
```
SanjoDeundiak, line 66: Would it be hard to do it vice versa: Use the boolean flag as the source of truth and set it to true if the env variable is set? I think using environment as a global storage for variables isn't perfect

implementations/rust/ockam/ockam_api/src/logs/setup.rs
adrianbenavides: Nope, thanks for catching that. It's a leftover from something I was experimenting with

implementations/rust/ockam/ockam_command/src/command.rs
adrianbenavides: Not in this case because it's just one variable, but I was following the way we configure the database [here](https://github.com/build-trust/ockam/blob/2ddf8ac6cacdaeecba0f794b242116555e676fb5/implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs#L255). Passing this as a variable would also mean adding a new field to `CliState` that would be exposing an implementation detail unnecessarily.

adrianbenavides: It's also true that the default in-disk sqlite details are leaked into the CliState's `dir` field... and we are not consistent in defining the database options.

SanjoDeundiak: IMO the right way to use environment variables, is to pull value from it and then populate it to where it's needed. Otherwise, it's just a global variable (which is kinda a bad practice itself), but even slightly worse. It's okish to leave it that way to avoid extra work though. Up to you.

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_command/src/command.rs
adrianbenavides: Ok done! Let me know what you think 🙏

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
```diff
@@ -315,6 +351,18 @@ pub fn random_name() -> String {
     petname::petname(2, "-").unwrap_or(hex::encode(random::<[u8; 4]>()))
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum CliStateMode<'a> {
+    Persistent(Cow<'a, PathBuf>),
```
SanjoDeundiak: Can we just own it?

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_api/src/cli_state/cli_state.rs
adrianbenavides: Done. I was trying to save some clones, but it wasn't really helpful.

SanjoDeundiak (approved):

PR #8701, opened 2024-12-12
feat(rust): introduce env variables to adjust transport performance

SanjoDeundiak:


implementations/rust/ockam/ockam_core/src/env/env.rs
```diff
@@ -6,34 +6,38 @@ use std::env::VarError;
...
+    Ok(get_env::<T>(var_name)?.unwrap_or(default_value))
+}
+
+/// Get environmental value `var_name`. If value is not found returns `default_value`
+pub fn get_env_with_default_ignore_error<T: FromString>(var_name: &str, default_value: T) -> T {
+    get_env::<T>(var_name)
+        .ok()
+        .flatten()
+        .unwrap_or(default_value)
+}
```
adrianbenavides, line 39: That's nice. I wonder if we should remove the result from the return type... We can take a look after this pr is merged though.

adrianbenavides (approved):

implementations/rust/ockam/ockam_core/src/env/env.rs
SanjoDeundiak: I'll create an issue

PR #8704, opened 2024-12-16
feat(rust): add `tcp-test` tool to measure portal properties

davide-baldo:
This tool is useful to test the portal capabilities and operates as a **generic** TCP client and server.
It also implements (unsafe) TLS, which is useful for performance comparison.

### DIFF
--- a/tools/tcp-test/README.md
+++ b/tools/tcp-test/README.md
@@ -1,0 +1,89 @@
+This tool is useful to test the portal capabilities and operates as a **generic** TCP client and server.
+It also implements (unsafe) TLS, which is useful for performance comparison.
+
+# Deployment
+
+Can be easily deployed via ssh using the `deploy` script:
+```bash
+$ ./deploy user@host
+```
+
+# Actions
+
+# latency
+Measures the latency of a TCP packets to a given host and port.
+It must be used with an echo server.
+
+```bash
+$ tcp-test echo 1234
+Listening on 0.0.0.0:1234
+
+$ tcp-test latency 127.0.0.1:1234
+Connection + First RTT: 0ms 440us, Second RTT: 0ms 143us
+Connection + First RTT: 0ms 192us, Second RTT: 0ms 160us
+Connection + First RTT: 0ms 276us, Second RTT: 0ms 162us
+Connection + First RTT: 0ms 326us, Second RTT: 0ms 155us
+Connection + First RTT: 0ms 227us, Second RTT: 0ms 172us
+Connection + First RTT: 0ms 250us, Second RTT: 0ms 111us
+Connection + First RTT: 0ms 461us, Second RTT: 0ms 122us
+Connection + First RTT: 0ms 326us, Second RTT: 0ms 116us
+Connection + First RTT: 0ms 320us, Second RTT: 0ms 104us
+Connection + First RTT: 0ms 270us, Second RTT: 0ms 101us
+Average - Connection + First RTT: 0ms 309us, Second RTT: 0ms 135us
+```
+
+# flood
+Floods a given host and port with connections until it fails or reaches the provided maximum.
+It's meant to measure the maximum number of connections a portal can handle.
+It must be used with an echo server.
+
+Example:
+```bash
+$ tcp-test echo 1234
+Listening on 0.0.0.0:1234
+
+$ tcp-test flood 127.0.0.1:1234
+Failed to connect: Os { code: 49, kind: AddrNotAvailable, message: "Can't assign requested address" }
+Flooded 16344 connections
+Pinging each...
+All connections succeeded
+```
+
+# throughput
+Measures the throughput of a TCP connection to a given host and port.
+It can be used to measure the throughput of a portal, especially to compare a portal against TLS.
+Can be used in conjunction with a null server for results similar to `iperf3`, but echo can also be
+used for a full bandwidth test.
+
+```bash
+$ tcp-test null 1234
+Listening on 0.0.0.0:1234
+
+$ tcp-test throughput 127.0.0.1:1234
+Outgoing Throughput: 70.44 Gbps
+Outgoing Throughput: 75.12 Gbps
+Outgoing Throughput: 75.44 Gbps
+Outgoing Throughput: 78.43 Gbps
+Outgoing Throughput: 71.89 Gbps
+Outgoing Throughput: 69.80 Gbps
+Outgoing Throughput: 75.13 Gbps
+^C
+```
+
+## echo
+It starts a TCP server that listens to on a given port and **echoes back** any data it receives.
+It's meant to be used in conjunction with the other actions.
+
+```bash
+$ tcp-test echo 1234
+Listening on 0.0.0.0:1234
+```
+
+# null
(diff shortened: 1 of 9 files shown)


tools/tcp-test/src/main.rs
```diff
@@ -0,0 +1,324 @@
...
+                return;
+            }
+        }
+
+        if iterations > 10_000 {
+            iterations = 0;
+            let elapsed = start.elapsed();
+            let seconds = elapsed.as_secs();
+            if seconds >= 1 {
+                println!(
```
SanjoDeundiak, line 188: Did you compare to iperf3? In terms of results and in terms of how it's implemented? My concern is the fact that stream write success, doesn't guarantee that the data was received (or will be received at all) on the other end. It doesn't wait for the ack, therefore I'm not sure  how correct is the throughput computation here. I would call it "sending throughput" compared to "receiving throughput". The second one is the real one. I understand that due to how TCP works sending throughput should get closer and closer to receiving throughput eventually, but have no idea how different the values can theoretically be in the meantime

tools/tcp-test/deploy
```diff
@@ -0,0 +1,35 @@
...
+
+function ssh {
+    /usr/bin/env ssh -q -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=QUIET -o ConnectTimeout=5 "${SSH_ENDPOINT}" "$@"
+}
+
+function scp {
+    /usr/bin/env scp -q -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=QUIET -o ConnectTimeout=5 "$@"
+}
+
+TARGET_ARCHITECTURE=$(ssh uname -m)
```
SanjoDeundiak, line 29: Does that assume that the machine we deploy to uses the same architecture as the machine we're running the script on? Is that intended?

tools/tcp-test/README.md
```diff
@@ -0,0 +1,85 @@
...
+$ tcp-test echo 1234
+Listening on 0.0.0.0:1234
+```
+
+# null
+It starts a TCP server that listens to on a given port and **discards** any data it receives.
+It's meant to be used in conjunction with the other actions.
+
+```bash
+$ tcp-test echo 1234
```
SanjoDeundiak: Should the command be different from the previous somehow?

tools/tcp-test/README.md
```diff
@@ -0,0 +1,85 @@
...
+It can be used to measure the throughput of a portal, especially to compare a portal against TLS.
+Can be used in conjunction with a null server for results similar to `iperf3`, but echo can also be
+used for a full bandwidth test.
+
+```bash
+$ tcp-test null 1234
+Listening on 0.0.0.0:1234
+
+$ tcp-test throughput 127.0.0.1:1234
+Throughput: 47.56066Gbits/s
```
SanjoDeundiak: Let's leave maybe 2 digits after the comma :)

tools/tcp-test/deploy
```diff
@@ -0,0 +1,35 @@
...
+
+if [ "$#" -lt 1 ]; then
+    echo "usage: $0 <[username@]host>"
+    exit 1
+fi
+
+export SSH_ENDPOINT="$1"
+
+function ssh {
+    /usr/bin/env ssh -q -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=QUIET -o ConnectTimeout=5 "${SSH_ENDPOINT}" "$@"
```
SanjoDeundiak: I'm a little bit concerned that some arguments for ssh and scp are different, e.g. `-p` is port for ssh, but `-P` for scp. Not sure that's a real issue, just FYI

tools/tcp-test/src/server.rs
```diff
@@ -0,0 +1,158 @@
+use rustls::pki_types::pem::PemObject;
+use rustls::pki_types::{CertificateDer, PrivateKeyDer};
+use std::collections::VecDeque;
+use std::net::{IpAddr, SocketAddr};
+use std::sync::{Arc, Mutex};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
+use tokio::sync::Notify;
+
+#[derive(Clone)]
```
SanjoDeundiak: nit: might derive Copy

tools/tcp-test/src/main.rs
```diff
@@ -0,0 +1,324 @@
...
+            let result = stream_read.read(&mut incoming_buffer).await;
+            if let Err(err) = result {
+                println!("Failed to read: {:?}", err);
+                return;
+            }
+        }
+    });
+
+    // read and report the speed
+    let outgoing_buffer = [0u8; 64 * 1024];
```
SanjoDeundiak: Let's move this value (and for other buffers) to a constant or a variable. Not sure if it should be adjustable

tools/tcp-test/src/server.rs
```diff
@@ -0,0 +1,158 @@
...
+            loop {
+                let incoming_buffer = incoming_buffers.lock().unwrap().pop_front();
+                if let Some(mut incoming_buffer) = incoming_buffer {
+                    let result = stream_reader.read(&mut incoming_buffer).await;
+                    match result {
+                        Ok(bytes) => {
+                            outgoing_buffers
+                                .lock()
+                                .unwrap()
+                                .push_back((bytes, incoming_buffer));
```
SanjoDeundiak, line 125: nit: should be possible to set length on the buffer after read, and clear it before putting it back after write

SanjoDeundiak (changes requested):
That looks very useful !

tools/tcp-test/README.md
davide-baldo: yes! it was a typo!

tools/tcp-test/deploy
davide-baldo: Nope, we assume it's different, and cross is always used to build (even when the arch is the same, to avoid glibc-related issues)

tools/tcp-test/src/main.rs
davide-baldo: It's also naive, but both server and client gives you a throughput report.
Also, we won't be able to compare with iperf without zerocopy.
It's mostly meant to compare performance with and without TLS, as long as both are using the same implementation it should be good enough.

tools/tcp-test/src/server.rs
davide-baldo: We could use it, but we would need to call `resize(.capacity(), 0)` before calling `read()`, and it would fill with zeroes every time

tools/tcp-test/deploy
davide-baldo: I double checked, these parameters overlap 👍

tools/tcp-test/src/main.rs
SanjoDeundiak: Actually, now I suspect iperf3 does the same

tools/tcp-test/deploy
SanjoDeundiak: I missed the `ssh` part :)

SanjoDeundiak (approved):

PR #8716, opened 2024-12-19
feat(rust): cache and reuse tcp connections

davide-baldo:
Reuse TCP connections rather than close them directly


davide-baldo:
not worth the extra complexity for now

PR #8718, opened 2024-12-22
feat(rust): improve ockam_node

SanjoDeundiak:
1. Using Oneshot channels where possible - faster, better semantics, avoiding async when sending a signal
2. Make start_worker and stop_worker sync, which allowed to make a lot of things sync, like start a secure channel listener, outlet, create TcpTransport, etc.
3. Removing code that was leading to implicit Clones (like &str -> Address conversion). Cloning causes a lot of small allocation and memory fragmentation => therefore should be avoided => therefore should explicit for starters.
4. HashMaps and HashSet instead of BTreeMap and BTreeSet for better performance
5. More granular locks (lock individual fields instead of the whole struct) inside Router.
6. Renaming to improve vocabulary consistency:
    - `ctx.stop_worker()`, `ctx.stop_processor()` -> `ctx.stop_address()`
    - `ctx.stop()` -> `ctx.shutdown_node()`
    - `node.stop()` -> `node.shutdown()`
    - terminate was used together with stop for the same things -> removed
8. Using `Weak<Router>` for clearer ownership over Router, which helps to drop it earlier and in more predictable way. E.g., detached Context will no longer keep Router from running => we can shutdown the environment successfully and then detached Context instances that are left can be dropped safely considering they no longer has access to the Router.
9. Use tokio::Handle instead of tokio::Runtime where possible, to achieve much clearer ownership over tokio's Runtime. Owning the Runtime instance keeps runtime alive + exclusive right to terminate it, while tokio::Handle only gives access to the Runtime, but doesn't keep it alive itself.
10. Got read of `AsyncTryClone`. Cloning Context type became sync, so we no longer need AsyncTryClone. Still needed failable clone, so replaced with `TryClone`
11. Made Worker and Processor treatment by the Router/Node almost identical, which will ease future merge of this abstractions.
12. Replaced clusters with Priority shutdown. Each worker can specify its shutdown priority (or get a default one). During node shutdown workers will be stopped in batches starting with the highest priority to the lowest. Why it's useful: workers rely on other workers, e.g. secure channel worker needs tcp connection worker to deliver messages. Stopping secure channel worker first, can possible allow it to still use tcp connection worker to deliver some final message(s) to the other side (e.g., "close secure channel message")

Possible future improvements:
1. More tests
2. Merge Worker and Processor into one trait
3. Try lockless structures (HashMaps and HashSets at least)
4. Shutdown tokio Environment in cases where something hangs and can't shutdown gracefully


implementations/rust/ockam/ockam_api/tests/common/mod.rs
```diff
@@ -4,5 +4,3 @@ pub mod common;
 pub mod session;
 #[allow(dead_code)]
 pub mod test_spans;
-#[allow(dead_code)]
```
SanjoDeundiak, line 7: @etorreborre is it ok to delete it? It looks unused

implementations/rust/ockam/ockam_core/src/routing/route.rs
```diff
@@ -14,6 +15,50 @@ pub struct Route {
     #[n(0)] inner: VecDeque<Address>,
 }
 
+impl AddAssign for Route {
```
davide-baldo, line 18: ❤️

implementations/rust/ockam/ockam_node/src/context/context.rs
```diff
@@ -1,39 +1,77 @@
...
+    /// 7
+    Priority7,
+}
+
+impl WorkerShutdownPriority {
+    /// All possible values in descending order
+    pub fn all_descending_order() -> [WorkerShutdownPriority; 7] {
+        use WorkerShutdownPriority::*;
+        [
+            Priority7, Priority6, Priority5, Priority4, Priority3, Priority2, Priority1,
```
davide-baldo, line 59: Do we actually need 7 priorities?

davide-baldo: It should be easy enough to add whenever necessary

implementations/rust/ockam/ockam_node/src/relay/worker_relay.rs
```diff
@@ -125,17 +123,15 @@ where
...
-                            error!("Error encountered during '{}' message handling: {:?}", address, e);
+                            error!("Error encountered during '{}' message handling: {:?}", self.ctx.primary_address(), e);
                             #[cfg(not(feature = "debugger"))]
-                            error!("Error encountered during '{}' message handling: {}", address, e);
+                            error!("Error encountered during '{}' message handling: {}", self.ctx.primary_address(), e);
                         }
                     }
                 },
-                result = ctrl_rx.recv() => {
-                    if result.is_some() {
```
davide-baldo, line 135: Was this a bug?

implementations/rust/ockam/ockam_node/src/router/record.rs
```diff
@@ -58,17 +66,9 @@ impl InternalMap {
...
-                    address_record.increment_msg_count();
-                    Ok(sender)
-                } else {
-                    trace!("Resolving worker address '{addr}'... REJECTED; worker shutting down");
-                    Err(
-                        Error::new(Origin::Node, Kind::Shutdown, "Worker shutting down")
-                            .context("Address", addr.clone()),
-                    )
-                }
+                trace!("Resolving worker address '{addr}'... OK");
```
davide-baldo, line 72: Even when disabled traces still have an impact, IMO we should reduce as much as possible in the "normal" message path

implementations/rust/ockam/ockam_node/src/router/router.rs
```diff
@@ -0,0 +1,168 @@
...
+/// This router supports two routing modes: internal, and external.
+///
+/// Internal routing resolves `type=0` addresses to local workers.
+///
+/// External routing is supported only after a plugin component
+/// registers itself with this router.  Only one router can be
+/// registered per address type.
+pub struct Router {
+    /// Keep track of some additional router state information
+    pub(super) state: SyncRwLock<RouterState>, // TODO: Could be AtomicU8
```
davide-baldo: > // TODO: Could be AtomicU8

It's only accessed when adding new addresses, transforming it into an atomic would have reduced benefits.

implementations/rust/ockam/ockam_node/src/router/shutdown.rs
```diff
@@ -1,81 +1,128 @@
...
+use ockam_core::Result;
 
 /// Register a stop ACK
 ///
 /// For every ACK we re-test whether the current cluster has stopped.
 /// If not, we do nothing. If so, we trigger the next cluster to stop.
 
 impl Router {
-    /// Implement the graceful shutdown strategy
+    /// Implement the graceful stop strategy
```
davide-baldo: It seems like the wording `stop` and `shutdown` inverted here

implementations/rust/ockam/ockam_node/src/shutdown.rs
```diff
@@ -1,3 +1,5 @@
+// TODO: Use either stop or shutdown everywhere
```
davide-baldo: IMO `shutdown` is more self-explanatory and less ambiguous

davide-baldo (commented):
Looks good!
Everything can be addressed in a follow-up

implementations/rust/ockam/ockam_node/src/router/record.rs
SanjoDeundiak: Agree, but looks like a completely unrelated improvement

implementations/rust/ockam/ockam_node/src/relay/worker_relay.rs
SanjoDeundiak: Could you please elaborate?

davide-baldo: You removed `if result.is_some() {`, and `result` comes from the previous select!{} condition `result = self.recv_message() => {`, was this preventing a relay shutdown?

PR #8724, opened 2024-12-30
feat(rust): added the possibility to overwrite http headers in inlets

davide-baldo:
This PR generalizes the HTTP interception code used in InfluxDb, and adds the possibility to write custom http headers using `ockam tcp-inlet create --http-header "Header: value" ...`

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/http/interceptor.rs
+++ b/implementations/rust/ockam/ockam_api/src/http/interceptor.rs
@@ -1,0 +1,309 @@
+use crate::http::state::{ClientRequestWriter, RequestState};
+use crate::nodes::models::services::{DeleteServiceRequest, StartServiceRequest};
+use crate::nodes::registry::HttpHeaderInterceptorInfo;
+use crate::nodes::{NodeManager, NodeManagerWorker};
+use crate::DefaultAddress;
+use httparse::Request;
+use minicbor::{CborLen, Decode, Encode};
+use ockam_abac::{Action, PolicyAccessControl, Resource, ResourceType};
+use ockam_core::api::Response;
+use ockam_core::errcode::{Kind, Origin};
+use ockam_core::{async_trait, Address, AllowAll, IncomingAccessControl, OutgoingAccessControl};
+use ockam_node::Context;
+use ockam_transport_tcp::{
+    read_portal_payload_length, Direction, PortalInletInterceptor, PortalInterceptor,
+    PortalInterceptorFactory,
+};
+use std::io::Write;
+use std::sync::{Arc, Mutex as SyncMutex};
+
+/// An HTTP headers interceptor that rewrites the headers of the HTTP
+/// request with the statically provided headers.
+struct StaticHttpHeadersInterceptor {
+    headers: Arc<Vec<(String, String)>>,
+}
+
+impl StaticHttpHeadersInterceptor {
+    /// Starts a listener that will intercept data in a portal on the inlet side.
+    /// The listener will rewrite the headers of the HTTP request with the provided headers.
+    pub async fn start_listener(
+        context: &Context,
+        listener_address: Address,
+        headers: Vec<(String, String)>,
+        policy_access_control: Option<PolicyAccessControl>,
+    ) -> ockam_core::Result<()> {
+        let flow_control_id = context
+            .flow_controls()
+            .get_flow_control_with_spawner(&Address::from_string(
+                DefaultAddress::SECURE_CHANNEL_LISTENER,
+            ))
+            .ok_or_else(|| {
+                ockam_core::Error::new(
+                    Origin::Channel,
+                    Kind::NotFound,
+                    "Secure channel listener not found",
+                )
+            })?;
+
+        context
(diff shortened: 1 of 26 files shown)


implementations/rust/ockam/ockam_api/src/http/state.rs
```diff
@@ -0,0 +1,168 @@
...
+pub(crate) trait ClientRequestWriter {
+    fn write_headers(
+        &self,
+        request: &httparse::Request,
+        buffer: &mut Vec<u8>,
+    ) -> ockam_core::Result<()>;
+}
+
+impl RequestState {
+    /* Parse the incoming data, attaching an Authorization header token to it.
```
etorreborre: This should use `///` for commenting.

implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
```diff
@@ -186,6 +195,43 @@ impl Command for CreateCommand {
...
+                        &overwrite_http_header_address,
+                        cmd.http_header.clone(),
+                    )
+                    .await;
+
+                match result {
+                    Ok(_) => {
+                        if let Some(pb) = pb.as_ref() {
+                            let created_message = format!(
+                                "Created a new Interceptor Service bound to {}\n",
```
etorreborre: "a new HTTP Interceptor Service"

implementations/rust/ockam/ockam_command/src/util/parsers.rs
```diff
@@ -21,6 +21,17 @@ pub(crate) fn hostname_parser(input: &str) -> Result<SchemeHostnamePort> {
...
 
+/// Helper function for parsing a key-value pair from user input
+/// The input is expected to be in the format `key:value`
+pub(crate) fn http_header_parser(input: &str) -> Result<(String, String)> {
+    let parts: Vec<&str> = input.split(':').collect();
+    if parts.len() != 2 {
+        return Err(miette!("Invalid header format. Expected 'key:value'"));
+    }
+
+    Ok((parts[0].trim().to_string(), parts[1].trim().to_string()))
```
etorreborre: Ideally we could check [the specification for header names and values](https://datatracker.ietf.org/doc/html/rfc7230#page-22), which is like:

- The header name has one or more characters in this range:
`! # $ % & ' * + - .  ^ _ ` | ~ digit alpha`
```
- The value is non-empty, and does not contain newlines (`\n`, `\r\n`) or control characters (`0x00–0x1F, 0x7F`).

This way we can prevent hard-to-diagnose issues down the stream. One typical example would be to define:
```
--header token:$AUTH_TOKEN
```
Where the user makes a mistake and uses `AUTH_TOKEN` instead of `AUTHORIZATION_TOKEN`.

File: implementations/rust/ockam/ockam_api/src/http/interceptor.rs [hunk 4]
```diff
@@ -0,0 +1,300 @@
...
+};
+use std::io::Write;
+use std::sync::{Arc, Mutex as SyncMutex};
+
+struct HttpHeadersInterceptorFactory {
+    headers: Arc<Vec<(String, String)>>,
+}
+
+impl HttpHeadersInterceptorFactory {
+    pub async fn create(
```
etorreborre: The naming of these functions make things a bit hard to follow IMO.

 1. `HttpHeadersInterceptorFactory::create` creates something but we don't know what.
 2. Internally, it calls `PortalInletInterceptor::create` which also creates something but does not return it, so we don't really what it is.
 3. But then we see that to create a `PortalInletInterceptor` we need a `HttpHeadersInterceptorFactory`, so presumably step 1. creates more than just a `HttpHeadersInterceptorFactory`.
 4. We then conclude that a `PortalInletInterceptor` uses a factory which also creates an `Interceptor`, so we eventually have interceptors creating interceptors 🙃.

It would be nice to have a clearer terminology:

- With maybe `register/start` instead of `create` when there are side-effects.
- Some documentation for methods having side-effects so we don't have to look at the implementation to understand what's going on.
- Just one level of interception and not an interceptor that can create other interceptors.

etorreborre (dismissed):
🏯

implementations/rust/ockam/ockam_command/src/util/parsers.rs
davide-baldo: I've taken inspiration from the `httparse` library we're using and made the parsing much more robust!
And I've disallowed an empty value, so we can cover the empty variable resolution case.

implementations/rust/ockam/ockam_api/src/http/interceptor.rs
davide-baldo: I've added some comment and renamed create into `start_listener`, which should be much more self-explanatory.
That said, I agree that the situation is still convoluted.
Oleksandr did propose to move the interceptors from the communication protocol and place them directly adjacent to the inlet and output ports. This would drastically simplify the architecture and avoid extra decoding/encoding of the payload.

etorreborre (approved):

PR #8731, opened 2025-01-07
feat(rust): simplify the retrieval of the postgres connection url

etorreborre:
This PR uses only one environment variable instead of 5 to build a postgres connection URL.


davide-baldo:
I would expect everything except the password to be readable configuration, It would be nice to have the password as a separated environment variable to keep it secret.

davide-baldo (approved):

PR #8735, opened 2025-01-10
feat: use postgres for managed nodes

etorreborre:
This PR adds a Rust migration to import the data from a SQLite node to a shared Postgres database for managed nodes.

- The main difficulty for this PR consists in passing the legacy SQLite database path down to the migration code, hence the modifications on the `migrator.rs` structs.

- This migration only applies once per migrated node if the Postgres database is configured. For this specific migration we use the SQLite database to mark the node as migrated and there's no mention of that migration in the Postgres database.

 - I also simplified the return value of the `migrate` method which returns `Result<()>` now since the previously returned boolean was always `true`.

 - Note that this code only runs if the Postgres database is enabled.

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/storage/database/database_configuration.rs
+++ b/implementations/rust/ockam/ockam_node/src/storage/database/database_configuration.rs
@@ -30,6 +30,8 @@ pub enum DatabaseConfiguration {
     Postgres {
         /// Connection string of the form postgres://[{user}:{password}@]{host}:{port}/{database_name}
         connection_string: String,
+        /// Path to a SQLite database that needs to be migrated to the Postgres database.
+        legacy_sqlite_path: Option<PathBuf>,
     },
 }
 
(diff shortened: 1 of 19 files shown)


implementations/rust/ockam/ockam_node/src/storage/database/database_configuration.rs
```diff
@@ -30,6 +30,8 @@ pub enum DatabaseConfiguration {
     Postgres {
         /// Connection string of the form postgres://[{user}:{password}@]{host}:{port}/{database_name}
         connection_string: String,
+        /// Path to a SQLite database that needs to be migrated to the Postgres database.
+        legacy_sqlite_path: Option<PathBuf>,
```
adrianbenavides, line 34: Do we plan to remove the related logic to this field once we have completed the migrations wherever it's needed?

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
```diff
@@ -228,10 +235,124 @@ impl Migrator {
...
+        // to know if that import happened already
+        if migration.name() == InitializeFromSqlite::name() {
+            if let Some(sqlite_db) = &self.legacy_sqlite_database {
+                let mut sqlite_connection = sqlite_db.pool.acquire().await.into_core()?;
+                if !self
+                    .has_migrated(&mut sqlite_connection, migration.name())
+                    .await?
+                {
+                    migration
+                        .migrate(self.legacy_sqlite_database.clone(), connection)
```
adrianbenavides, line 335: Just to double-check, the usage of `connection` here instead of `sqlite_connection` is correct here?

adrianbenavides (approved):

implementations/rust/ockam/ockam_node/src/storage/database/database_configuration.rs
etorreborre: Potentially yes, but it's not impossible that we start some independent nodes at some stage and then want to move them to being managed nodes. So I don't know how long this code should stay in.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migrator.rs
etorreborre: Yes, we use the `connection` to insert data into Postgres there. 
Then we mark the migration as executed inside the SQLite database.

PR #8744, opened 2025-01-14
feat(rust): add binary to compile custom branding command binaries

adrianbenavides:
## Binary to compile branded `ockam` versions

There is a new script at `ockam_command/src/bin/brand.rs` that can be run to compile the `ockam` binary with some configurable options.

The script needs a configuration that can be either passed inline, or as a path to a file. The configuration is a yaml with he following format:

```yaml
bin:
  support_email: support@domain.com
  # the following fields are optional
  brand_name: Brand1 # if not set, will default to Bin
  home_dir: /home/brand1 # if not set, will default to $HOME/.bin
  orchestrator_identifier: brand1 # if not set, will default to the OCKAM_CONTROLLER_IDENTITY_ID env var
  orchestrator_address: brand1.network # if not set, will default to the OCKAM_CONTROLLER_ADDR env var
```

We can run the script as follows: `cargo run --bin brand ./config.yaml`

## Custom branding

The `ockam_command` crate now will use the compile-time env vars `OCKAM_COMMAND_BIN_NAME` and `OCKAM_COMMAND_BRAND_NAME` to replace the `ockam` and `Ockam` references in:

- Command names
- Command help texts: documentation of command and args, code examples, markdown files
- Command output

We can also customize:

- The `OCKAM_HOME` to be used, which will default to `$HOME/.$BIN_NAME`
- Support email, which is shown in the error footer (the text in dark gray)
- The Controller address
- The Controller identifier

## ⚠️  Things that are not replaced ⚠️ 

1. References to attributes that are prefixed with `ockam` like `ockam-relay` or `ockam-role`. Should we change this prefix?
2. Environment variables names
3. Logs
4. Errors (the `source location` leaks the crates names)
5. `ockam enroll` has a lot of ockam related output (links to docs, links to our github, the ascii header), but this command should only be used by admins anyway

## New Cargo features in ockam_command

I've removed the `orchestrator` feature, which wasn't being used consistently, and I've added two new features:
- `admin_commands`: enroll, create/delete spaces and projects, authority, sidecar, ...
- `advanced_commands`: experimental or not widely used commands (tcp-listener/connection, flow control, secure-channel, ...)

So, by default, we will include the core commands that are used to create nodes, portals, relays, policies, ...


etorreborre (approved):
🚗

PR #8746, opened 2025-01-14
feat(rust): update the postgres schema

etorreborre:
This PR: 

 1. Consolidates the current Postgres database schema (that schema is not yet used in production). Some tables are removed from that schema since Postgres will be used for managed nodes where some notions don't apply, like `space`, `user` and `project`.
 2. Removes the unused `project_node` table.
 3. Deactivates the automatic database upgrades on startup when Postgres is used (because we can't allow one individual managed node to trigger a full database migration).
 4. Makes sure that the `ockam reset` command cannot be used with a Postgres database.
 5. Adds some test support to run only the relevant tests with Postgres.


adrianbenavides (approved):
Nice!

PR #8747, opened 2025-01-14
feat: allow authority nodes to share a Postgres instance

etorreborre:
This PR makes sure that we can start several authority nodes using a shared Postgres database:

 - It adds an `authority_id` column to the `authority_member` table in order to isolate members from authority to authority
 - It adds some tests to show that 2 authority nodes can be started with the same Postgres database with all their services running:
   - direct authentication.
   - issue credentials.
   - issue tokens.
   - accept tokens.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/authority_node/authority.rs
+++ b/implementations/rust/ockam/ockam_api/src/authority_node/authority.rs
@@ -29,11 +29,13 @@ use crate::nodes::service::default_address::DefaultAddress;
 
 /// This struct represents an Authority, which is an
 /// Identity which other identities trust to authenticate attributes
-/// An Authority is able to start a few services
-//   - a direct authenticator
-//   - a credential issuer
-//   - an enrollment token issuer
-//   - an enrollment token acceptor
+///
+/// An Authority is able to start a few services:
+//   - a direct authenticator: can add and retrieve members.
+//   - a credential issuer: return the attributes of a member as a time-limited credential.
+//   - an enrollment token issuer: create a token attributed allowing an identity to acquire some specific attributes.
+//   - an enrollment token acceptor: create or update a member, given a token.
+#[derive(Clone)]
 pub struct Authority {
     identifier: Identifier,
     secure_channels: Arc<SecureChannels>,
@@ -169,6 +171,7 @@ impl Authority {
         }
 
         let direct = DirectAuthenticatorWorker::new(
+            &self.identifier,
             self.members.clone(),
             self.secure_channels.identities().identities_attributes(),
             self.account_authority.clone(),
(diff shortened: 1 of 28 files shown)


implementations/rust/ockam/ockam_api/src/authority_node/authority.rs
```diff
@@ -334,7 +345,331 @@ impl Authority {
...
+    ) -> Result<()> {
+        if let Err(e) = client
+            .clone()
+            .with_client_identifier(member)
+            .issue_credential(ctx)
+            .await
+        {
+            panic!("{e:?}");
+        };
+        Ok(())
```
adrianbenavides, line 607: Wouldn't be the same if the function doesn't return a Result and instead you just unwrap the function call (same for the `add_member` function)?

`client.clone().with_client_identifier(member).issue_credential(ctx).await.unwrap()`

implementations/rust/ockam/ockam_api/src/authority_node/authority.rs
```diff
@@ -334,7 +345,331 @@ impl Authority {
...
+                )
+                .await?;
+
+                // adding members must work for both authorities
+                let identities_creation = secure_channels.identities().identities_creation();
+                let member1 = identities_creation.create_identity().await?;
+                let member2 = identities_creation.create_identity().await?;
+
+                add_member(&ctx_client, &client1, &member1).await?;
+                add_member(&ctx_client, &client2, &member2).await?;
```
adrianbenavides: What's the point of doing the same operations with different members on different clients? I'd also add cases to prove the two clients are isolated by adding the same member to both of them and asserting that you can't add the same member twice in a given client.

implementations/rust/ockam/ockam_api/src/authority_node/authority.rs
etorreborre: Right, going to do this.

implementations/rust/ockam/ockam_api/src/authority_node/authority.rs
etorreborre: The purpose of the test is show that the authority nodes that are sharing the same database are able to respond to their own clients. 

> I'd also add cases to prove the two clients are isolated by adding the same member to both of them and asserting that you can't add the same member twice in a given client.

Good idea, I'm going to add this case.

adrianbenavides: > sharing the same database are able to respond to their own clients.

Ok, I got it now, thanks

etorreborre: > you can't add the same member twice in a given client.

Hmm, actually we can, this just updates the member attributes. I pushed some changes to the test showing that updating a member works ok.e

adrianbenavides (dismissed):
🚀

adrianbenavides (approved):

PR #8755, opened 2025-01-15
Optimize ockam node create

SanjoDeundiak:
1. Move tokio Runtime creation to the earliest possible during ockam command execution
2. Avoid creating tokio Runtime more than once
3. Remove Runtime singletin inside ockam_node create
4. Remove most of the blocking calls
5. Simplify commands to support only async execution
6. Add timeout to Tcp connections
7. Optimize node creation by replacing readiness check from polling to TCP callback
8. Move logging setup to the earliest possible
9. Simplify CliState


implementations/rust/ockam/ockam_command/src/command.rs
```diff
@@ -43,31 +57,198 @@ pub struct OckamCommand {
...
+            (tracing_configuration, tracing_guard, None)
+        } else {
+            let cli_state = self.init_cli_state(in_memory).await;
+            let tracing_configuration = self.make_tracing_configuration(&cli_state).await?;
+            let tracing_guard =
+                self.setup_logging_tracing(&logging_configuration, &tracing_configuration);
+            let cli_state = cli_state.set_tracing_enabled(tracing_configuration.is_enabled());
+
+            (tracing_configuration, tracing_guard, Some(cli_state))
+        };
```
adrianbenavides, line 206: what are the implications of not initializing the clistate here? Couldn't we initialize it here in both branches and remove the `match cli_state` from below?

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -45,6 +45,7 @@ after_long_help = docs::after_help(AFTER_LONG_HELP)
 pub struct CreateCommand {
     /// Name of the node or a configuration to set up the node.
     /// The configuration can be either a path to a local file or a URL.
+    /// TODO: Use Option<String>
```
adrianbenavides, line 48: What's the context of this?

implementations/rust/ockam/ockam_command/src/node/create.rs
```diff
@@ -318,16 +297,11 @@ impl CreateCommand {
             };
             if name == DEFAULT_NODE_NAME {
                 name = random_name();
-                if let Ok(default_node) = opts.state.get_default_node().await {
-                    if !default_node.is_running() {
-                        // The default node was stopped, so we can reuse the name
-                        name = default_node.name();
-                    }
-                }
```
adrianbenavides, line 326: Why are these two checks removed, to prevent the clistate calls?

implementations/rust/ockam/ockam_command/src/command.rs
SanjoDeundiak: The idea is to initialize logging as early as possible. The problem is that if tracing is enabled, we unfortunately need CliState to be able to properly bootstrap it. Therefore, I was able to initialize logging before creating the CliState only if tracing is disabled

implementations/rust/ockam/ockam_command/src/node/create.rs
SanjoDeundiak: With current approach it's impossible to see if a user has passed "_default_node_name" or no value at all. Then later in the code you can see
```
if name == DEFAULT_NODE_NAME {
        name = random_name();
}
```
Which is now a hacky way to figure out if the user has provided no value. It would be a surprise for a user who provided "_default_node_name" and then ended up with a random node name, right? Not sure that's a realistic concern in this particular situation (due to low chance someone picking "_default_node_name" as the node name), but using Option would make it a clear distinction and as a consequence - easier to read the code.

implementations/rust/ockam/ockam_command/src/node/create.rs
SanjoDeundiak: I think the whole idea of node registry in the sqlite may need some improvements, but I haven't spent much time thinking about it. My main concern is that there are too many operations to achieve simple things, like starting a node. Ideally, creating a background node should not require opening the DB inside the parent process at all. Let's at least make small steps towards that. This check looked like something that doesn't give us much usability, but would slow us down each time we create a node. We need to admit that db calls are not free - they hurt latency, and start rethinking UX based on that. Also, let's maybe attempt replacing node registry and their status by getting information about running processes from the OS. That might turn out both faster and more reliable. WDYT about all of that?

SanjoDeundiak: Also, given default Vault, Identity, etc. creation is already performed by the background node, not the parent process, can we move some of these checks to the background node as well? That should get us to "parent node doesn't need db access" relatively easy. The ideal result would be that parent process only need to perform few things:
1. Start the background ockam process
2. Print output from that process
3. Wait for the completion callback

SanjoDeundiak: Also, given default Vault, Identity, etc. creation is already performed by the background node, not the parent process, can we move some of these checks to the background node as well? That should get us to "parent node doesn't need db access" relatively easy. The ideal result would be that parent process only need to perform few things:
1. Start the background ockam process
2. Print output from that process
3. Wait for the completion callback

SanjoDeundiak: Also, given default Vault, Identity, etc. creation is already performed by the background node, not the parent process, can we move some of these checks to the background node as well? That should get us to "parent node doesn't need db access" relatively easy. The ideal result would be that parent process only need to perform few things:
1. Start the background ockam process
2. Print output from that process
3. Wait for the completion callback

implementations/rust/ockam/ockam_command/src/node/create.rs
adrianbenavides: I was checking previous versions of the command and it's never been an `Option<String>`, although conceptually it's always been optional. The only explanation I can find is that it's easier to work with a `String`, given all the logic we have around this argument.

implementations/rust/ockam/ockam_command/src/node/create.rs
adrianbenavides: > My main concern is that there are too many operations to achieve simple things, like starting a node. Ideally, creating a background node should not require opening the DB inside the parent process at all.

I agree that we can remove those checks from the `parse_args` function, but we need to make sure they exist in the foreground function.

>  Also, let's maybe attempt replacing node registry and their status by getting information about running processes from the OS. 

I don't think we can get useful info from processes in most cases. 

> Also, given default Vault, Identity, etc. creation is already performed by the background node, not the parent process, can we move some of these checks to the background node as well? 

+1

implementations/rust/ockam/ockam_command/src/node/create.rs
SanjoDeundiak: I think if it's conceptually optional, it's better make it Optional, otherwise it can bite us in a very unobvious way eventually :)

implementations/rust/ockam/ockam_command/src/node/create.rs
SanjoDeundiak:
```
I don't think we can get useful info from processes in most cases.
```
I think we can try. Some information is possible to built into the process name, e.g. node name, or even listening port. And with listening port we could actually connect to that node and get all the info from it directly.
^
||
Just an idea, not a suggestion. Needs thorough consideration

adrianbenavides: that's a good idea. I just checked and we can modify the process name in the `pre_exec` call when spawning a background node. Setting that information for a foreground node is another story I guess... but it's something we could investigate.

adrianbenavides (approved):

PR #8756, opened 2025-01-16
feat(rust): add a command to migrate a postgres database

etorreborre:
This PR adds a `migrate-postgres` command which can be used to check the migration status of a shared Postgres database and to migrate the current schema/data.

# Implementation notes

## `host_journey` table

I added the `host_journey` table to still be able to collect traces when there are command errors, even in contexts where a project does not exists. 

This allows to not change the existing code too much for the `migrate-postgres` command. Indeed that command might fail if the `OCKAM_DATABASE_CONNECTION_URL` environment variable is not defined and in that case we create a command error user event.

## MigrationStatus

Most of the change consist in reporting richer data about the status of migrations, instead of simple booleans and errors.
Note that a failed migration is reported with an `Ok(status)` and Rust `Err`ors are left for infrastructure errors (broken connection for example).

## Version

I introduced a wrapper type, `Version`, around `i64` for versions, to give that value its own meaning and make sure we don't make undue conversions.


implementations/rust/ockam/ockam_command/src/migrate_database/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+                if !self.dry_run {
+                    migrator.migrate(&db.pool).await?;
+                };
+
+                let status = migrator.migration_status(&db.pool).await?;
+                opts.terminal.stdout().plain(&status).json_obj(&status)?.machine(status.up_to_date()).write_line()?;
+
+                Ok(())
+            },
+            None => Err(Error::new(Origin::Core, Kind::NotFound, "There is no postgres database configuration, or it is incomplete. Please run ockam environment to check the database environment variables".to_string())).into_diagnostic()?,
```
adrianbenavides: I haven't tried but I think the error would look nicer in the terminal if it were just a string like `miette!("There is no...")`. Ignore this comment if you have already considered it!

adrianbenavides: Another minor improvement would be to format the command `ockam environment` with `color_primary`

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migration_result.rs
```diff
@@ -0,0 +1,70 @@
...
+use serde::Serialize;
+
+/// This enum models the result of executing one migration.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum MigrationResult {
+    /// The migration was successful
+    MigrationSuccess,
+    /// The migration had a failure
+    SomeMigrationError(MigrationFailure),
+}
```
adrianbenavides: Why not write `MigrationFailure` as an `Error` and use `Result<T, MigrationFailure>`? I can see it's easy to use it this way in other parts of the code, instead of doing the error conversion stuff, but it still feels a little bit weird.

Also, I'd remove the `Some` prefix in `SomeMigrationError` and rename `MigrationFailure` to `MigrationError` if possible.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migration_status.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// Create a new MigrationStatus::Todo
+    pub fn create_to_do(current_version: Option<Version>, next_version: Version) -> Self {
+        MigrationStatus::Todo(current_version, next_version)
+    }
+
+    /// Create a new MigrationStatus::Failed
+    pub fn create_failed(version: Version, failure: MigrationFailure) -> Self {
+        MigrationStatus::Failed(version, failure)
+    }
```
adrianbenavides: What's the benefit of adding these constructors? I'd just use the enum variants directly.

implementations/rust/ockam/ockam_command/src/migrate_database/mod.rs
etorreborre: > I haven't tried but I think the error would look nicer in the terminal if it were just a string like miette!("There is no..."). Ignore this comment if you have already considered it!

Yes it's better, I made that change.

> Another minor improvement would be to format the command ockam environment with color_primary

I agree we could highlight the env names. But I'm leaving this out of this PR because I think this means putting the whole static file containing the variables inside the code (and not as a static `.txt` file anymore). So that's a larger change that deserves its own PR.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migration_result.rs
etorreborre: I modeled it like that to represent expected "errors" as regular values while other errors can be more unexpected or transient, for example if we can't access the database. This is also why I chose to call those "failures" and not "errors".

I renamed `SomeMigrationError` to `MigrationFailure`. I actually did not realize that this would not create a clash with the `MigrationFailure` struct.

If you don't mind, let's go with this and revisit if it's really cumbersome.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/migration_support/migration_status.rs
etorreborre: You're right, removed.

adrianbenavides (approved):

PR #8758, opened 2025-01-17
refactor(rust): rename cloud module to orchestrator

adrianbenavides:
- Minor cleanup of the old "cloud" module of the ockam_command crate
- Revert command features, as this could interfere with the next release. I'm also working on a different approach that won't use cargo features


etorreborre (approved):
🍡

PR #8769, opened 2025-01-22
feat(rust): custom branding config allows you to specify which commands are included

adrianbenavides:
In the configuration passed to the brand binary, we can now pass the commands we want to include and optionally set a custom name. If none are passed, all the commands are available:

```yaml
bin_name:
  commands:
    - node: "host" # 'node' top-level command will be renamed to 'host'
    - node_list
    - node_create: "init" # 'node create' -> 'host init'
```

So far I’ve added the ability to add/rename top-level commands only (e.g. node, project, space, ...) but not subcommands.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1509,9 +1509,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.23"
+version = "4.5.27"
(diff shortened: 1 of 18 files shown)


etorreborre (approved):
🐰

PR #8772, opened 2025-01-23
feat(rust): don't return an error when a tcp processor does not receive an ockam message

etorreborre:
This PR do no fail the initialization of a TCP receiver when it receives a message which does not support the Ockam protocol.
This reduces the number of errors logged in authority nodes when we used their TCP listener to check for liveness.

### DIFF
--- a/implementations/rust/ockam/ockam_transport_tcp/src/workers/receiver.rs
+++ b/implementations/rust/ockam/ockam_transport_tcp/src/workers/receiver.rs
@@ -14,7 +14,6 @@ use ockam_core::{
 };
 use ockam_core::{Processor, Result};
 use ockam_node::{Context, ProcessorBuilder, WorkerShutdownPriority};
-use ockam_transport_core::TransportError;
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_transport_tcp/src/workers/receiver.rs
```diff
@@ -132,24 +131,24 @@ impl Processor for TcpRecvProcessor {
         let protocol_version = match self.read_half.read_u8().await {
             Ok(p) => p,
             Err(e) => {
+                trace!("Cannot read the Ockam protocol version: {:?}", &e);
                 self.notify_sender_stream_dropped(ctx, e).await?;
-                return Err(TransportError::GenericIo)?;
+                self.registry
```
SanjoDeundiak: No need to do it here, it'll be done in the shutdown. But we need to stop the processor using ctx.stop_adress

SanjoDeundiak (dismissed):

SanjoDeundiak (approved):

PR #8777, opened 2025-01-27
feat(rust): make portal handshake optional

SanjoDeundiak:
1. Add `--skip-handshake` flag to tcp inlets&outlets, that makes handshake non-blocking to optimize latency (but possibly sacrificing throughput)
2. Add `--enable-nagle` flag to tcp inlets&outlets. By default we set nodelay flag on TCP streams, but that may sometimes be unwanted (for high throughput use cases)

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
+++ b/implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
@@ -142,6 +142,7 @@ pub struct CreateCommand {
 
     /// Use eBPF and RawSocket to access TCP packets instead of TCP data stream.
     /// If `OCKAM_PRIVILEGED` env variable is set to 1, this argument will be `true`.
+    /// WARNING: This flag value should be equal on both ends of a portal (inlet and outlet)
     #[arg(long, env = "OCKAM_PRIVILEGED", value_parser = FalseyValueParser::default(), hide = true)]
     pub privileged: bool,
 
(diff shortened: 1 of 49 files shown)


implementations/rust/ockam/ockam_command/src/tcp/inlet/create.rs
```diff
@@ -153,6 +153,14 @@ pub struct CreateCommand {
     /// Requires `ockam-tls-certificate` credential attribute.
     #[arg(long, value_name = "ROUTE", hide = true)]
     pub tls_certificate_provider: Option<MultiAddr>,
+
+    /// Skip Portal handshake for lower latency, but also lower throughput
```
davide-baldo, line 158: It would be helpful for the user to know that he'll need to specify `skip_handshake` in the outlet as well (and the reverse as well)

implementations/rust/ockam/ockam_command/src/tcp/outlet/create.rs
```diff
@@ -73,6 +73,14 @@ pub struct CreateCommand {
     /// If `OCKAM_PRIVILEGED` env variable is set to 1, this argument will be `true`.
     #[arg(long, env = "OCKAM_PRIVILEGED", value_parser = FalseyValueParser::default(), hide = true)]
     pub privileged: bool,
+
+    /// Skip Portal handshake for lower latency, but also lower throughput
```
davide-baldo, line 78: Same as inlet

implementations/rust/ockam/ockam_transport_tcp/src/portal/outlet_listener.rs
```diff
@@ -81,34 +107,84 @@ impl Worker for TcpOutletListenWorker {
...
+        if self.options.skip_handshake {
+            let sender_remote = self
+                .outlet_registry
+                .started_workers
+                .read()
+                .unwrap()
+                .get(&map_key)
+                .cloned();
+            if let Some(sender_remote) = sender_remote {
+                return Self::reroute_msg(ctx, sender_remote.clone(), msg).await;
```
davide-baldo: `sender_remote.clone()` ← this clone should be avoidable

davide-baldo (dismissed):

implementations/rust/ockam/ockam_transport_tcp/src/portal/outlet_listener.rs
SanjoDeundiak: I hope compiler is clever enough to not clone it, but let's not rely on that :)

davide-baldo (approved):

PR #8779, opened 2025-01-28
feat(rust): in enrollment tickets, derive id from project route's service

adrianbenavides:
- Change logic of enrollment tickets so that the project id is extracted from the service in the route, instead of from the DNS address 
- Cleans up the branding related code in the `ockam_command` crate, to isolate the compile env vars usage and make a clear distinction between these vars and the regular runtime env vars we use in other parts of the codebase
- Adds clap to parse the brand binary arguments


etorreborre (approved):
👍 The consequence of this PR is that, as Ockam developers, we should now have `COMPILE_OCKAM_DEVELOPER=true` in our environment, right (because then `OCKAM_DEVELOPER` follows from it)?

adrianbenavides:
>👍 The consequence of this PR is that, as Ockam developers, we should now have COMPILE_OCKAM_DEVELOPER=true in our environment, right (because then OCKAM_DEVELOPER follows from it)?

Yes, exactly.

PR #8783, opened 2025-01-31
refactor(rust): enable requests to be messages

etorreborre:
This PR adjusts the serialization / deserialization of Ockam messages. 

## Motivation

The initial motivation was to be able to define `Request<T>` as a type of `Message` for workers processing messages.
Some additional objectives are to:

 - Decouple the general `Encodable / Decodable` interfaces from the `Serializable / Deserializable` implementations provided by the `serde_bare` library
 - Highlight the fact that `Requests` are currently doubly-encoded, once as a `Vec<u8>` with `minicbor` and then serialized with `serde_bare`. Fixing this issue will be a one-line (or 2) change after this PR but it is a breaking change, so I left this for later.

## Changes

The main changes are:

 1. The removal of default instances for `Encodable` and `Decodable`.
 2. The addition of an `Encodable` and `Decodable` for a `Request<T>` provided that `T: Message`.
     - As a result, all the messages being sent via requests now have their own `Message` instance (that explains the number of touched files)
 3. The stricter typing of `ask/tell/request/...` methods for both requests and responses.
 4. A two-steps decoding process of `Requests` in workers like the `NodeManager` to:
     -  First decode the request as `Request<Vec<u8>>` (now by the virtue of `Request<T>` being a `Message`) in order to read the `RequestHeader`.  
     - And then decode the request payload using a `Decodable<T>` instance we know the expected type based on the header. Note that this is done without having to know that the request body was encoded using `minicbor`.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/influxdb/lease_token.rs
+++ b/implementations/rust/ockam/ockam_api/src/influxdb/lease_token.rs
@@ -4,14 +4,16 @@ use crate::terminal::fmt;
 use crate::ApiError;
 use minicbor::{CborLen, Decode, Encode};
 use ockam::identity::Identifier;
+use ockam::Message;
 use ockam_core::compat::fmt::Error as FmtError;
+use ockam_core::{cbor_encode_preallocate, Decodable, Encodable, Encoded};
 use serde::{Deserialize, Serialize};
 use std::cmp::PartialEq;
 use std::fmt::{Display, Formatter};
 use strum::{Display, EnumString};
 use time::OffsetDateTime;
 
-#[derive(Encode, Decode, CborLen, Serialize, Deserialize, Debug, Clone)]
+#[derive(Encode, Decode, CborLen, Serialize, Deserialize, Debug, Clone, Message)]
 #[cbor(map)]
 pub struct LeaseToken {
     #[cbor(n(1))]
@@ -31,6 +33,34 @@ pub struct LeaseToken {
 
     #[cbor(n(6))]
     pub status: TokenStatus,
+}
+
+impl Encodable for LeaseToken {
+    fn encode(self) -> ockam_core::Result<Encoded> {
+        cbor_encode_preallocate(self)
+    }
+}
+
+impl Decodable for LeaseToken {
+    fn decode(e: &[u8]) -> ockam_core::Result<Self> {
+        Ok(minicbor::decode(e)?)
+    }
+}
+
+#[derive(Encode, Decode, CborLen, Serialize, Deserialize, Debug, Clone, Message)]
+#[cbor(transparent)]
+pub struct LeaseTokenList(#[n(0)] pub Vec<LeaseToken>);
+
+impl Encodable for LeaseTokenList {
+    fn encode(self) -> ockam_core::Result<Encoded> {
+        cbor_encode_preallocate(self)
+    }
+}
+
+impl Decodable for LeaseTokenList {
+    fn decode(e: &[u8]) -> ockam_core::Result<Self> {
+        Ok(LeaseTokenList(minicbor::decode(e)?))
+    }
 }
 
 #[cfg(test)]
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/node_manager_worker.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/node_manager_worker.rs
@@ -2,13 +2,13 @@ use ockam::{route, Result};
 use ockam_core::api::{Error, Response};
 use ockam_node::Context;
 
-use crate::nodes::models::portal::{CreateInlet, InletStatus};
+use crate::nodes::models::portal::{CreateInlet, InletStatus, InletStatusList};
 use crate::nodes::NodeManagerWorker;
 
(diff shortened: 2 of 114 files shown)


implementations/rust/ockam/ockam_api/src/influxdb/lease_token.rs
```diff
@@ -33,6 +35,34 @@ pub struct LeaseToken {
     pub status: TokenStatus,
 }
 
+impl Encodable for LeaseToken {
+    fn encode(self) -> ockam_core::Result<Encoded> {
+        Ok(minicbor::to_vec(self)?)
```
davide-baldo: `minicbor::to_vec()` can cause multiple memory allocations since it doesn't know the result beforehand.
You can use `cbor_encode_preallocate()` to avoid many small allocations.

implementations/rust/ockam/ockam_api/src/influxdb/lease_token.rs
```diff
@@ -33,6 +35,34 @@ pub struct LeaseToken {
     pub status: TokenStatus,
 }
 
+impl Encodable for LeaseToken {
```
davide-baldo, line 38: Have you tried to implement Encodable for every Encode?
The cons would be that you need to create a new type if the **root** message is encoded both in cbor and bare, but i think it's pretty rare

```
impl<T: Encode<()> + CborLen<()>> Encodable for T {
    fn encode(self) -> Result<Encoded> {
        cbor_encode_preallocate(self)
    }
}
````

**Review by davide-baldo — COMMENTED** [review]
There is no performance regression but maybe we can avoid manually implementing every message

File: implementations/rust/ockam/ockam_api/src/influxdb/lease_token.rs (hunk 1 above)
Comment (etorreborre) [reply implementations/rust/ockam/ockam_api/src/influxdb/lease_token.rs]: Thanks, I'll do that change.

File: implementations/rust/ockam/ockam_api/src/influxdb/lease_token.rs (hunk 2 above)
Comment (etorreborre) [reply implementations/rust/ockam/ockam_api/src/influxdb/lease_token.rs:38]: I did it on purpose in order to not tie the `Encodable` trait to any implementation. It used to be tied to `Serializable` before that PR.

We currently have a mix of `cbor` and `serde` at the moment and this way it becomes a conscious choice on how you want to serialize your message data. Also note that we can't specify both a generic encoding for `T: Encode` and `T: Serializable` in Rust.

Comment (adrianbenavides) [reply implementations/rust/ockam/ockam_api/src/influxdb/lease_token.rs:38]: Perhaps a macro derive to implement Encodable with cbor would help in this case? It would be explicit as to how the trait is implemented and it would keep the trait isolated.

File: implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/node_manager_worker.rs [hunk 3]
```diff
@@ -2,13 +2,13 @@ use ockam::{route, Result};
 use ockam_core::api::{Error, Response};
 use ockam_node::Context;
 
-use crate::nodes::models::portal::{CreateInlet, InletStatus};
+use crate::nodes::models::portal::{CreateInlet, InletStatus, InletStatusList};
 use crate::nodes::NodeManagerWorker;
 
 impl NodeManagerWorker {
-    pub(crate) async fn get_inlets(&self) -> Result<Response<Vec<InletStatus>>, Response<Error>> {
+    pub(crate) async fn get_inlets(&self) -> Result<Response<InletStatusList>, Response<Error>> {
```
adrianbenavides, line 9: This kind of change (vec<T> => TList) is now needed because we don't have a default impl right?

implementations/rust/ockam/ockam_core/src/api.rs
```diff
@@ -587,31 +576,60 @@ impl Request<()> {
...
+    }
+}
+
+impl<T: Decodable> Decodable for Request<T> {
+    fn decode(e: &[u8]) -> Result<Self> {
+        let deserialized = deserialize::<Vec<u8>>(e)?;
+        let mut dec = Decoder::new(&deserialized);
+        let header: RequestHeader = dec.decode()?;
+        if header.has_body() {
+            let body = dec.input().get(dec.position()..deserialized.len()).unwrap();
```
adrianbenavides: Since the request's body is an option, wouldn't it be better to default to None if we fail to get the body, instead of unwrapping?

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/node_manager_worker.rs
etorreborre: Yes. Since `Decodable` is abstract, you can't say something like "reuse the same decodable implementation to decode a `Vec`, than the one you used to decode the `T` inside the `Vec`".

implementations/rust/ockam/ockam_core/src/api.rs
etorreborre: I should return a better error here, but this should still be an error if we can't get the rest of the input.

etorreborre: Done.

davide-baldo (approved):

PR #8785, opened 2025-02-03
fix(rust): add missing branding replacements in ockam_command

adrianbenavides:
- replace hardcoded binary names and command names (e.g. when spawning a background node) with their corresponding compile env vars
- replace command names with their custom names in output, docs, and help texts

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/ui/command.rs
+++ b/implementations/rust/ockam/ockam_api/src/ui/command.rs
@@ -1,0 +1,92 @@
+use std::fmt::{Debug, Formatter};
+
+#[derive(Clone, Debug)]
+pub struct Commands {
+    pub(crate) commands: Vec<Command>,
+}
+
+#[derive(Clone)]
+pub(crate) struct Command {
+    pub(crate) name: &'static str,
+    pub(crate) custom_name: &'static str,
+}
+
+impl Debug for Command {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Command")
+            .field("name", &self.name)
+            .field("custom_name", &self.custom_name)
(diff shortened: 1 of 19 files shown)


implementations/rust/ockam/ockam_api/src/ui/command.rs
```diff
@@ -0,0 +1,101 @@
...
+            })
+            .collect();
+        Self { commands }
+    }
+
+    pub fn hide(&self, command_name: &'static str) -> bool {
+        if self.commands.is_empty() {
+            return false;
+        }
+        // we only support custom names for top-level commands
```
etorreborre: It might be a good idea to give an example here of what is a "top-level" command and what it's not.
The fact that we are identifying this with the presence of a whitespace seems a bit fragile to me.

etorreborre (dismissed):

implementations/rust/ockam/ockam_api/src/ui/command.rs
adrianbenavides: > It might be a good idea to give an example here of what is a "top-level" command and what it's not.
The fact that we are identifying this with the presence of a whitespace seems a bit fragile to me.

I thought it was a limitation in clap, but it turns out that you can name commands with spaces. I'm working on a different approach.

adrianbenavides: I moved that check to the brand.rs binary. Otherwise, we end up in a circular dependency situation when trying to parse the command and get the top-level commands.

https://github.com/build-trust/ockam/compare/86ad15c37431ed5e2ee94699e6ffb14127f3aea2..371f8db10659c81c4d4738495141f616161e6fe8

etorreborre (approved):
🛝

PR #8796, opened 2025-02-12
feat(rust): add custom log format for foreground nodes

adrianbenavides:
- Fix session loop updates the `initial_connect_was_called` var after reconnecting
- Add custom log format used in `node create --foreground` that will print log messages with the target `ockam_command:user` in the terminal as regular output
- Add info logs in the portal worker connect/disconnect events

https://github.com/user-attachments/assets/70a59e0a-636a-4566-95d1-ccaa2cfc10d9

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
+++ b/implementations/rust/ockam/ockam_api/src/logs/logging_configuration.rs
@@ -7,7 +7,7 @@ use std::path::PathBuf;
 use tracing_core::Level;
 use tracing_subscriber::EnvFilter;
 
-use super::{Colored, LoggingEnabled};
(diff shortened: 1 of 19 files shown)


etorreborre (commented):
I think it is a good idea to not require a `CliState` to instantiate a `Notifier` but I would not make it a singleton, for the usual mocking and testing reasons:

 - It's not easy to replace with a `Notifier` with a different behaviour
 - Any tests making some assertions on the `Notifier` cannot be executed concurrently and has to be carefully isolated.

I know that it is a currently a bit painful to pass components around because we don't have any DI support but I would still build that component and pass it around manually, first on the `CliState`, then the `NodeManager`, the `SessionReplacer` etc...

etorreborre (approved):
🎁

PR #8801, opened 2025-02-14
feat: send telemetry data via a secure channel

etorreborre:
This PR allows telemetry data (logs and spans) to be exported via a secure channel to an Ockam node and forwarded to an OpenTelemetry collector.

## Configuration

There are new environment variables to decide how to secure channel should be established:

 - `OCKAM_TELEMETRY_EXPORT_VIA_PROJECT` creates a secure channel to the project node (this is the default).
 - `OCKAM_TELEMETRY_EXPORT_VIA_AUTHORITY` creates a secure channel to the authority node if for some reason we prefer this route (this could be useful for testing differences for testing Elixir vs Rust implementations or if we want to have only one node collecting telemetry data in a scenario with one authority node and several project nodes.
 - `OCKAM_TELEMETRY_EXPORT_NODE_ROUTE` + `OCKAM_TELEMETRY_EXPORT_NODE_IDENTIFIER` creates a secure channel to an arbitrary node (this is useful when testing locally by just starting 2 nodes). 

## gRPC data interception

When a project is ready, the log and span exporters are set with a gRPC client which packs the gRPC requests into Ockam messages.

Those messages are sent to a `GrpcForwarder` worker which deserializes the requests and forwards them to an Opentelemetry collector endpoint. This PR provides 2 implementations of a `GrpcForwarder`, one in Elixir for the project node, and one in Rust for the authority node (or an arbitrary Rust node).

## Documentation

I started a `telemetry.md` document to describe the different options which can be used to export telemetry data.

## `TcpTransport`

The creation of a `TcpTransport` with the `TcpTransport::create(ctx: &Context)` method now checks if the transport has not already been created. If it has been created, it returns the same `TcpTransport` instead of returning a new one which would have a new internal registry and potentially lose some data.

This avoids the passing of both the `Context` and a `TcpTransport` when creating the `SecureClient` that will be used to send telemetry data.

## Dependencies upgrades

This PR also upgrades the `opentelemetry` Rust crates since there were recent breaking changes and [more are coming up](https://github.com/open-telemetry/opentelemetry-rust/blob/69e7e89b47e78d76e5f9b3af3f0bfcfbf4254ba5/opentelemetry-sdk/CHANGELOG.md).

## Testing

I added a way to support the testing of a Rust node sending telemetry data to an Elixir node, in order to speed-up the feedback loop when developing that feature:

 - The `start_node.exs` file starts an Ockam Elixir node with a fixed identity:
```
MIX_ENV=test iex -S mix run --no-halt -r test/start_node.exs
```
 - The identity identifier is stored in the `identity.id` file in the same directory as the `start_node.exs` script and can be used to setup the `OCKAM_TELEMETRY_EXPORT_NODE_IDENTIFIER` variable when creating a secure channel between a Rust node and the Elixir node.
 - The `test` config starts a `grpc_forwarder` service using the `OCKAM_OPENTELEMETRY_ENDPOINT` environment variable pointing to an OpenTelemetry collector. You can find how to start such a collector on your machine [here](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/examples/basic-otlp/README.md).


implementations/elixir/ockam/ockam_cloud_node/lib/cloud_node.ex
```diff
@@ -41,6 +41,9 @@ defmodule Ockam.CloudNode do
             []
         end
 
+      {nil, nil} ->
```
etorreborre, line 44: This silences some logs in test mode.

implementations/elixir/ockam/ockam_metrics/lib/application.ex
```diff
@@ -17,7 +17,11 @@ defmodule Ockam.Metrics.Application do
       else
         []
       end ++
-        telemetry_poller()
+        if Application.get_env(:ockam_metrics, :start_telemetry_poller, true) do
```
etorreborre, line 20: This silences some logs when testing.

implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
```diff
@@ -23,6 +23,8 @@ use crate::config::lookup::InternetAddress;
 
 use crate::{fmt_warn, ConnectionStatus};
 
+pub const DEFAULT_NODE_NAME: &str = "_default_node_name";
```
adrianbenavides: I don't think this will work as expected. This value is used in command as a temporary value that is replaced by either a random name or the name defined in a node configuration. The only reliable way to get the default node name would be to read it from the DB.

implementations/rust/ockam/ockam_api/src/logs/grpc_forwarder.rs
```diff
@@ -0,0 +1,84 @@
...
+            .await
+            .map_err(|e| ApiError::message(format!("cannot connect to {uri:?}: {e:?}")))?;
+
+        Ok(Self { channel })
+    }
+
+    /// Forward an http Request.
+    /// We don't wait for a response here.
+    async fn forward_http_request(&mut self, request: http::Request<BoxBody>) -> Result<()> {
+        self.ready().await.map_err(ApiError::core)?;
```
adrianbenavides, line 41: Why is this needed before forwarding every request?

implementations/rust/ockam/ockam_transport_udp/src/transport/lifecycle.rs
```diff
@@ -20,15 +21,30 @@ impl UdpTransport {
...
-    pub fn create(ctx: &Context) -> Result<Self> {
-        let udp = Self {
-            ctx: Arc::new(ctx.try_clone()?),
-        };
-        // make the UDP transport available in the list of supported transports for
-        // later address resolution when socket addresses will need to be instantiated as UDP
-        // worker addresses
-        ctx.register_transport(Arc::new(udp.clone()));
-        Ok(udp)
+    pub fn create(ctx: &Context) -> Result<Arc<UdpTransport>> {
```
adrianbenavides: Shouldn't this be named `get_or_create` ?

implementations/rust/ockam/ockam_transport_tcp/src/transport/lifecycle.rs
```diff
@@ -20,13 +20,32 @@ impl TcpTransport {
     /// # Ok(()) }
     /// ```
     #[instrument(name = "create tcp transport", skip_all)]
-    pub fn create(ctx: &Context) -> Result<Self> {
+    pub fn create(ctx: &Context) -> Result<Arc<TcpTransport>> {
```
adrianbenavides: Same as in the udp, shouldn't this be named `get_or_create` ? Then the `create_new` could be just `create`

implementations/rust/telemetry.md
```diff
@@ -0,0 +1,113 @@
...
+        grpc_endpoint: "http://opentelemetry-collector:4317",
+        authorization: [{Ockam.Worker.Authorization, :from_secure_channel}]
+      ]}
+```
+
+The value of `grpc_endpoint` must be set to the endpoint of an accessible OpenTelemetry collector.
+
+## Sending telemetry data directly via a secure channel to the authority node
+
+This behaviour is controlled by the `OCKAM_TELEMETRY_EXPORT_VIA_PROJECT` environment variable (the default is `false`).
```
adrianbenavides: Typo in the env var (should point to the authority var right?)

implementations/rust/telemetry.md
```diff
@@ -0,0 +1,113 @@
...
+In order for the forwarding to work, the authority node be configured with the `OCKAM_OPENTELEMETRY_ENDPOINT` set to
+the collector endpoint URL, for example `http://opentelemetry-collector:4317`.
+
+## Sending telemetry data directly via a secure channel to an arbitrary Ockam node
+
+This behaviour is controlled by the setting of two environment variables:
+
+- `OCKAM_TELEMETRY_EXPORT_NODE_ROUTE` a route to the node to connect. For example:
+  `/dnsaddr/localhost/tcp/4000/secure/api`. Note that the presence of `secure/api` in the address is going to trigger
+  the creation of a secure channel.
```
adrianbenavides: Perhaps a little extension to this sentence would help to understand the implications of the secure channel creation.

implementations/rust/ockam/ockam_api/src/logs/secure_client_service.rs
```diff
@@ -0,0 +1,255 @@
...
+use std::task::Poll;
+use tonic::body::BoxBody;
+use tonic::codegen::Service;
+
+/// This service implements the `tower_service::Service` trait
+/// to post http Requests to an Ockam node as Ockam messages, via a secure channel defined by the
+/// SecureClient.
+///
+/// The Context is an Option since because it is necessary to clone this struct and attempting
+/// to attempt a Context could fail if the current node is being shut down.
```
adrianbenavides: `attempt` = `clone`?

implementations/rust/ockam/ockam_api/src/cli_state/nodes.rs
etorreborre: This is a rebase error. I shouldn't have to touch that code.

implementations/rust/ockam/ockam_api/src/logs/grpc_forwarder.rs
etorreborre: This is the protocol used by the `tonic` library when calling a `Channel`.

implementations/rust/ockam/ockam_transport_udp/src/transport/lifecycle.rs
etorreborre: Yes, much better.

implementations/rust/telemetry.md
etorreborre: Good catch.

adrianbenavides (approved):

PR #8803, opened 2025-02-17
fix(rust): `node create` issues with node name and stderr/stdout pipes

adrianbenavides:


implementations/rust/ockam/ockam_command/src/node/create/background.rs
```diff
@@ -38,11 +38,19 @@ impl CreateCommand {
 
         // Run foreground node in a separate process
         // Output is handled in the foreground execution
-        let handle = spawn_node(&opts, cmd)?;
+        let mut handle = spawn_node(&opts, cmd)?;
 
         tokio::select! {
-            _ = wait_while_draining_output(&opts, handle) => { std::process::exit(1) }
-            _ = node_callback.wait_for_signal() => {}
+            res = handle.wait() => {
```
etorreborre: I guess that it's difficult to refactor that code with the same code for the authority node?

implementations/rust/ockam/ockam_command/src/node/start.rs
```diff
@@ -179,11 +180,19 @@ async fn run_node(
         cmd
     };
 
-    let handle = spawn_node(opts, cmd)?;
+    let mut handle = spawn_node(opts, cmd)?;
 
     tokio::select! {
-        _ = wait_while_draining_output(opts, handle) => { std::process::exit(1) }
-        _ = node_callback.wait_for_signal() => {}
+        res = handle.wait() => {
```
etorreborre: Same question here.

etorreborre (dismissed):
🧑‍🍳

implementations/rust/ockam/ockam_command/src/node/create/background.rs
adrianbenavides: done!

etorreborre (approved):

PR #8804, opened 2025-02-17
MPTCP support

SanjoDeundiak:
Add MPTCP support

### DIFF
--- a/examples/command/portals/mptcp/README.md
+++ b/examples/command/portals/mptcp/README.md
@@ -1,0 +1,52 @@
+## MPTCP
+
+This example shows how to create a TCP Portal that leverages MPTCP to achieve up to 25Gbits/s throughput between different AWS accounts.
+
+## Usage
+
+ - Create a working directory, place ockam binary built for aarch64-unknown-linux-gnu in release mode (by default script installs latest released version of Ockam, but expects a manually built binary in the working directory that would be used to replace the released version for easier debugging)
+ - Use 2 different terminal tabs that logged in into 2 different AWS accounts (e.g. use `export AWS_PROFILE="{PROFILE_NAME}"`)
+ - Preferably, use the same AWS region to experience the best throughput (e.g., use `export AWS_REGION="{REGION}"`)
+ - Create 2 subdirectories for 2 AWS accounts and navigate to them in corresponding terminal tabs
+ - Use the script to create the outlet ec2 machine on the 1st account:
+    ```
+   aws.sh "{ENROLLMENT TICKET}" inlet
+   ```
+ - Use the script to create the inlet ec2 machine on the 2nd account. `8` refers to the number of additional ports that the server will advertise as available for MPTCP subflows:
+   ```
+   aws.sh "{ENROLLMENT TICKET}" outlet 8
+   ```
+ - Use printed full ssh command to connect to both ec2 machines
+ - It's recommended to adjust few related variables to achieve the best throughput, e.g.:
+   - Ockam settings: `export OCKAM_OPENTELEMETRY_EXPORT=false`, `OCKAM_TCP_PORTAL_PAYLOAD_LENGTH=400000` (see `env_info.txt` for the full list)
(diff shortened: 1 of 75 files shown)


implementations/rust/ockam/ockam_transport_tcp/src/mptcp/tokio.rs
```diff
@@ -0,0 +1,105 @@
...
+/// Bind using MPTCP
+pub async fn bind_mptcp(addr: SocketAddr) -> io::Result<TcpListener> {
+    let socket = MptcpSocketBuilder::new_for_addr(addr)?
+        .set_nonblocking()?
+        .bind(addr)?;
+
+    TcpListener::from_std(socket.into())
+}
+
+#[cfg(all(test, target_os = "linux"))]
```
davide-baldo: Why are the tests linux specific?

implementations/rust/ockam/ockam_transport_tcp/src/mptcp/tokio.rs
```diff
@@ -1,27 +1,31 @@
...
     for addr in addrs {
-        match f(addr).await {
-            Ok(l) => return Ok(l),
+        let stream = {
+            let sock = MptcpSocketBuilder::new_for_addr(addr)?
+                .set_nonblocking()?
+                .connect(addr)
+                .and_then(|sock| TcpStream::from_std(sock.into()))?;
+            // Wait for the socket to be writable
+            poll_fn(|cx| sock.poll_write_ready(cx)).await?;
```
davide-baldo, line 25: I guess writable here means connection completed, could you add a comment about it?
I can't find any timeout reference, is there an implicit one?

examples/command/portals/mptcp/README.md
```diff
@@ -0,0 +1,52 @@
+## MPTCP
+
+This example shows how to create a TCP Portal that leverages MPTCP to achieve up to 25Gbits/s throughput between different AWS accounts.
+
+## Usage
+
+ - Create a working directory, place ockam binary built for aarch64-unknown-linux-gnu in release mode (by default script installs latest released version of Ockam, but expects a manually built binary in the working directory that would be used to replace the released version for easier debugging)
```
davide-baldo, line 7: It isn't clear how the user can created the expected binary.
Maybe we could get away with downloading a specific binary?
```
curl --proto '=https' --tlsv1.2 -sSfL https://install.command.ockam.io | bash /dev/stdin --binary-file aarch64-unknown-linux-gnu --no-modify-path --install-path /tmp/ && mv /tmp/bin/ockam .
```

davide-baldo (commented):
Connection timeout is the main concern, everything else looks good

implementations/rust/ockam/ockam_transport_tcp/src/mptcp/tokio.rs
SanjoDeundiak: Only linux kernel supports mptcp listeners 😅

implementations/rust/ockam/ockam_transport_tcp/src/mptcp/tokio.rs
SanjoDeundiak: That's a very good question.
1. There is no timeout neither here, nor in tokio's TCP connect, which is why I added timeout support myself a while ago (see `create_tcp_stream_timeout`). However, it's optional and we don't always specify it (maybe we should)
2. While inspecting tokio's counterpart, I found we also should check the socket for an error after it's writable. Added that part along with the comment

examples/command/portals/mptcp/README.md
SanjoDeundiak: I think when we release with that feature, last released binary would work fine (it's installed automatically on the ec2) and we won't need to copy a custom-build one. So the scp line could be easily commented-out.

davide-baldo (dismissed):

davide-baldo (approved):

PR #8805, opened 2025-02-18
fix(rust): fix the deserializaton of error messages in responses

etorreborre:
This PR fixes an error occurring when enrolling with a ticket:
```
ockam enroll
ockam project ticket > ticket
ockam node create --enrollment-ticket ticket
```

### DIFF
--- a/implementations/rust/ockam/ockam_core/src/api.rs
+++ b/implementations/rust/ockam/ockam_core/src/api.rs
@@ -654,8 +654,7 @@ impl<T: Message> Message for Request<T> {}
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Response<T = ()> {
     header: ResponseHeader,
-    body: Option<T>,
-    error: Option<Error>,
+    body: Option<core::result::Result<T, Error>>,
 }
 
 impl<T> Response<T> {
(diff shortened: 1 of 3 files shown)


implementations/rust/ockam/ockam_core/src/api.rs
```diff
@@ -697,6 +696,15 @@ impl<T> Response<T> {
...
 }
+impl<T: Clone> Response<T> {
+    pub fn get_body(&self) -> Option<T> {
+        self.body.clone().and_then(|b| b.ok())
+    }
+
+    pub fn get_error(&self) -> Option<Error> {
+        self.body.clone().and_then(|b| b.err())
+    }
+}
```
adrianbenavides: How about taking the option's value, instead of cloning it?

```
pub fn take_body(&mut self) -> Option<T> {
    self.body.take().and_then(|b| b.ok())
}

pub fn take_error(&mut self) -> Option<Error> {
    self.body.take().and_then(|b| b.err())
}
```

etorreborre: That's a good point, I propose to consume the `Response` entirely seeing how we use those functions. If more than just the body or the error is needed there's always the `into_parts` method.

adrianbenavides (approved):
🚀

PR #8808, opened 2025-02-21
feat(rust): add support for "launch-configuration" in node's config

adrianbenavides:
Adds support to pass the `launch-configuration` (now renamed to `services`) as part of the node configuration:

```bash
ockam node create '{
  name: n1,
  tcp-listener-address: "127.0.0.1:14001",
  start-default-services: true,
  services: {
    control-api: {
      authentication-token: token,
      backend: true,
      frontend: true,
      http-bind-address: "127.0.0.1:14002",
      node-resolution: relay
  }}}'
```

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/control_api/openapi.rs
+++ b/implementations/rust/ockam/ockam_api/src/control_api/openapi.rs
@@ -46,9 +46,9 @@ configuration or via environment variable, it's then passed in the `Authorizatio
 ## Getting Started
 The easiest way to get development started is to use this API with the Ockam command and run a single node acting both as a frontend and backend.
 ```sh
-ockam node create --foreground -vv --launch-configuration '{
+ockam node create --foreground -vv --configuration '{
   "start_default_services": true,
-  "startup_services": {
+  "services": {
     "control_api": {
       "authentication_token": "my-secret-token",
       "backend": true,
--- a/implementations/rust/ockam/ockam_command/src/node/create/demo_config_files/3.node.services.yaml
+++ b/implementations/rust/ockam/ockam_command/src/node/create/demo_config_files/3.node.services.yaml
@@ -1,0 +1,7 @@
+name: n1
+tcp-listener-address: 127.0.0.1:3333
+services:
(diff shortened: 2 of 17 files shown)


implementations/rust/ockam/ockam_command/src/node/create/demo_config_files/3.node.services.yaml
```diff
@@ -0,0 +1,8 @@
+name: n1
+tcp-listener-address: 127.0.0.1:3333
+start_default_services: true
```
davide-baldo: We can ignore back-compatibility for control node api related fields and change its configuration fields to kebab-case for consistency

implementations/rust/ockam/ockam_api/src/control_api/openapi.rs
```diff
@@ -46,7 +46,7 @@ configuration or via environment variable, it's then passed in the `Authorizatio
 ## Getting Started
 The easiest way to get development started is to use this API with the Ockam command and run a single node acting both as a frontend and backend.
 ```sh
-ockam node create --foreground -vv --launch-configuration '{
+ockam node create --foreground -vv --configuration '{
```
davide-baldo, line 49: I think `startup_services` should become `services`

implementations/rust/ockam/ockam_command/src/run/parser/resource/node.rs
```diff
@@ -155,9 +160,44 @@ impl Node {
...
+    pub services: Option<NamedResources>,
+}
+
+impl Services {
+    /// Parse the services fields into a `ServicesConfig` struct
+    pub fn into_arg(self) -> Result<ServicesConfig> {
+        let mut as_json = serde_json::json!({
+            "services": self.services,
+        });
+        if let Some(start_default_services) = self.start_default_services {
```
davide-baldo, line 176: Same as above, i think we can get away with only the kabeb-case `start-default-services`

davide-baldo (dismissed):

davide-baldo (approved):

PR #8825, opened 2025-02-27
fix(rust): in `node create`, detach background process

adrianbenavides:
Prevents blocking issues in the parent process when the `ockam node create` is run in environments where the terminal session remains open until all subprocesses exit (e.g. through ssh).


etorreborre (approved):
🥤

PR #8829, opened 2025-02-28
feat(rust): add command to retrieve the identity listening at some endpoint

polvorin:
* add "ockam secure-channel peer-info" command to retrieve the identity listening at some endpoint. 
```
pablo@Pablos-MacBook-Pro debug % ./ockam secure-channel peer-info --to  /dnsaddr/orchestrator.ockam.io/tcp/6252/service/api --output json
{
  "change_history": "81825858830101585385f6820181584104ebf9d78281a04f180029c12a74e994386c7c9fee24903f3bfe351497a9952758ee5f4b57d7ed6236ab5082ed85e1ae8c07d5600e0587f652d36727904b3e310df41a656a365d1a7836395d820181584050bf79071ecaf08a966228c712295a17da53994dc781a22103602afe656276ef83ba83a1004845b1e979e0944abff3cd8c7ceef834a8f5eeeca0e8f720fa38f4",
  "identifier": "I84502ce0d9a0a91bae29026b84e19be69fb4203a6bdd1424c85a43c812772a00"
}
```

* add hidden option to `ockam project ticket` to avoid calling orchestrator' controller when generating and enrollment ticket.  This is needed in any self managed scenario:    an authority is started (preconfigured with our identity as valid enroller),  a relay node is started,   a project.json file is generated to use the above authority and relay.   We use `ockam project import` to import the project' definition.  Then try to generate an enrollment ticket with `ockam project ticket`

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/nodes/service/in_memory_node.rs
+++ b/implementations/rust/ockam/ockam_api/src/nodes/service/in_memory_node.rs
@@ -253,10 +253,16 @@ impl InMemoryNode {
         ctx: &Context,
         project: &Project,
         caller_identity_name: Option<String>,
+        skip_controller_call: bool,
     ) -> miette::Result<AuthorityNodeClient> {
         let client = self
             .node_manager
-            .create_authority_client_with_project(ctx, project, caller_identity_name)
+            .create_authority_client_with_project(
+                ctx,
+                project,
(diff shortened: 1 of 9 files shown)


adrianbenavides (approved):
Not 100% convinced about the command name and the `--to` argument name, but we can change that in a separate PR

PR #8832, opened 2025-02-28
feat(rust): added support for multiple concurrent routes to tcp inlets

davide-baldo:
This PR adds the possibility of multiple MultiAddr to the inlets.

There are few new parameters:
- `--ping-timeout`, this is a proxy for ping interval and gets divided by the max. amount of failed pings
- `--target-redundancy`, the amount of redundant routes instantiated, 0 means only one route will be instantiated, can still be specified when a single MultiAddr since dns could resolve into multiple IP addresses. By default it inherits the number of `--to` parameters.
- `--to` can now be passed multiple times

A single MultiAddr can turn into multiple variants:
`/project/default/...` => `/dnsaddr/xx.projects.orchestrator.io/...`  => [ `/ip4/1.1.1.1/...`, `/ip4/2.2.2.2/...`]

A session is created for each route (target redundancy+1) , every time it reconnects it selects a MultiAddr from the selector, making sure every MultiAddr variant gets used(unless redundancy+1 > available variants).
When DNS fails (or return empty) and no other MultiAddr is available, the session will retry every 15 seconds.

The inlet port will be bind upon creation, when a client connection reaches the inlet, a random route will be picked from the available ones, if no route is available it'll wait until one is available.


implementations/rust/ockam/ockam_api/src/cli_state/storage/tcp_portals_repository.rs
```diff
@@ -96,12 +96,12 @@ impl TcpInlet {
         self.bind_addr
     }
 
-    pub fn outlet_addr(&self) -> MultiAddr {
-        self.outlet_addr.clone()
+    pub fn outlet_addr(&self) -> &Vec<MultiAddr> {
```
adrianbenavides: Rename to `outlet_addresses`?

implementations/rust/ockam/ockam_api/src/control_api/protocol/inlet.rs
```diff
@@ -13,6 +13,10 @@ fn retry_wait_default() -> u64 {
     20000
 }
 
+fn ping_timeout_default() -> u64 {
```
adrianbenavides, line 16: Why not use a `Duration` instead? At least I'd include the units I the function name. If it's because of some API arguments constraints the [duration parser](https://github.com/build-trust/ockam/blob/d42d9a0ef72977f8393b5be83b13446c8eb0d8c8/implementations/rust/ockam/ockam_core/src/env/from_string.rs#L121) we use in the command?

implementations/rust/ockam/ockam_api/src/control_api/protocol/inlet.rs
```diff
@@ -65,9 +69,13 @@ pub struct CreateInletRequest {
     #[serde(default = "tcp_inlet_default_bind_address")]
     #[schema(default = tcp_inlet_default_bind_address)]
     pub from: HostnamePort,
-    /// Multiaddress to a TCP Outlet
+    /// Multiaddresses to a TCP Outlet
     #[schema(example = "/project/default/service/forward_to_node1/secure/api/service/outlet")]
-    pub to: String,
+    pub to: Vec<String>,
+    /// Target redundancy for the TCP Inlet routes; 0 means only one route is instantiated
+    /// When omitted, the number of provided Multiaddresses minus one applies
```
adrianbenavides: How are multiple lines processed when generating the API docs? Normally, if they are processed as markdown, line breaks are omitted and these two lines would be printed as a single line (with no spaces between them).

implementations/rust/ockam/ockam_api/src/influxdb/portal.rs
```diff
@@ -132,7 +134,13 @@ impl NodeManagerWorker {
         } = body.tcp_inlet.clone();
 
         //TODO: should be an easier way to tweak the multiaddr
-        let mut issuer_route = outlet_addr.clone();
+        let mut issuer_route = if let Some(first) = outlet_addresses.first() {
```
adrianbenavides, line 138: Could it happen that the first address is no longer active? Or that doesn't matter at this point?

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
```diff
@@ -151,8 +130,12 @@ impl CreateInlet {
         self.prefix_route = route;
     }
 
-    pub fn set_wait_ms(&mut self, ms: u64) {
-        self.wait_for_outlet_duration = Some(Duration::from_millis(ms))
+    pub fn set_ping_timeout(&mut self, ping_timeout: Duration) {
+        self.ping_timeout = Some(ping_timeout)
+    }
+
+    pub fn set_wait_ms(&mut self, wait_for_outlet_duration: Duration) {
```
adrianbenavides: Rename to `set_wait_for_outlet`?

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
```diff
@@ -253,83 +216,78 @@ impl CreateOutlet {
...
-                fmt::INDENTATION,
-                color_primary(r.to_string())
-            )?;
+
+        if self.outlet_routes.is_empty() {
+            writeln!(f, "{}No routes to outlet", fmt::INDENTATION)?;
+        } else {
+            writeln!(f, "{}With routes to outlet:", fmt::INDENTATION)?;
+            for route in &self.outlet_routes {
+                writeln!(f, "{}", &fmt_list!("{route}"))?;
```
adrianbenavides: The fmt_list macro shouldn't be used here. That's only used in `list` subcommands right now to create visual separations from elements. Instead, you should have:

`writeln!(f, "{}{}{}", fmt::INDENTATION, fmt::INDENTATION, color_primary(route))?;`

implementations/rust/ockam/ockam_api/src/nodes/models/portal.rs
```diff
@@ -253,83 +216,78 @@ impl CreateOutlet {
...
-        writeln!(
-            f,
-            "{}Outlet Address: {}",
-            fmt::INDENTATION,
-            color_primary(&self.outlet_addr)
-        )?;
+
+        writeln!(f, "{}Outlet Addresses:", fmt::INDENTATION,)?;
+        for address in &self.outlet_addresses {
+            writeln!(f, "{}", &fmt_list!("{address}"))?;
```
adrianbenavides: Same here about the fmt_list

implementations/rust/ockam/ockam_api/src/nodes/registry.rs
```diff
@@ -133,27 +137,80 @@ impl KafkaServiceInfo {
...
+    }
+
+    pub(crate) async fn stop(&self, context: &Context) -> ockam_core::Result<()> {
+        for session in self.sessions.lock().await.iter_mut() {
+            session.stop().await;
+        }
+        self.tcp_inlet.stop(context)
+    }
+
+    /// Returns all available statues, usually one per session, but could be fewer
```
adrianbenavides: Typo: statues 🗽  -> statuses

implementations/rust/ockam/ockam_transport_tcp/src/portal/inlet_shared_state.rs
```diff
@@ -1,25 +1,131 @@
...
+                let guard = self.routes.read().unwrap();
+                let active_routes: Vec<&InletRouteState> = guard
+                    .iter()
+                    .filter(|(_, route)| !route.is_paused())
+                    .map(|(_, route)| route)
+                    .collect();
+
+                if !active_routes.is_empty() {
+                    break active_routes
+                        .choose(&mut rand::thread_rng())
```
adrianbenavides, line 72: Could this `rand::thread_rng()` call have a noticeable performance impact if this function becomes a hot path?

implementations/rust/ockam/ockam_transport_tcp/src/portal/inlet_shared_state.rs
```diff
@@ -1,25 +1,131 @@
...
+                }
+            }
+            // let's wait until something changes and try again
+            debug!("No active route found for the inlet, waiting for a connection");
+            self.notify.notified().await;
+        }
+    }
+
+    /// Update the route and set pause to false
+    pub fn update_route_and_unpause(
```
adrianbenavides, line 85: What about "resume" instead of "unpause"?

implementations/rust/ockam/ockam_transport_tcp/src/transport/portals.rs
```diff
@@ -37,14 +35,42 @@ impl TcpTransport {
...
+    /// # use ockam_core::{AllowAll, Result, route, Address};
+    /// # async fn test(ctx: Context) -> Result<()> {
+    /// let route_path = route!["outlet"];
+    ///
+    /// let tcp = TcpTransport::get_or_create(&ctx)?;
+    /// let tcp_inlet = tcp.crate_inlet_multi("localhost:4000".parse().unwrap()).await?;
+    /// tcp_inlet.add_route(&ctx, "first_route_key".to_string(), route_path, TcpInletOptions::new())?;
+    /// # tcp_inlet.stop(&ctx)?;
+    /// # Ok(()) }
+    /// ```
```
adrianbenavides, line 75: ❤️

implementations/rust/ockam/ockam_transport_tcp/src/transport/portals.rs
```diff
@@ -212,34 +238,69 @@ impl TcpInlet {
...
     /// reachable, or if we want to switch transport, e.g., from relayed to UDP NAT puncture.
     ///  NOTE: For regular Portals existing TCP connections will still use the old route,
     ///        only newly accepted connections will use the new route.
     ///        For privileged Portals old connections can continue work in case the Identifier of the
     ///        Outlet node didn't change
-    pub fn update_outlet_node_route(&self, ctx: &Context, new_route: Route) -> Result<()> {
-        let mut inlet_shared_state = self.inlet_shared_state.write().unwrap();
-
-        let new_route = Self::build_new_full_route(new_route, inlet_shared_state.route())?;
+    pub fn update_outlet_route_and_unpause(
```
adrianbenavides, line 253: Same comment here about `unpause` VS `resume`

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/node_manager.rs
```diff
@@ -109,135 +118,162 @@ impl NodeManager {
...
     }
 
-    pub async fn delete_inlet(&self, alias: &str) -> Result<InletStatus> {
+    pub async fn delete_inlet(&self, context: &Context, alias: &str) -> Result<InletStatusView> {
         info!(%alias, "Handling request to delete inlet portal");
         if let Some(inlet_to_delete) = self.registry.inlets.remove(alias) {
             debug!(%alias, "Successfully removed inlet from node registry");
-            inlet_to_delete.session.lock().await.stop().await;
+            if let Err(error) = inlet_to_delete.stop(context).await {
+                error!(%alias, %error, "Failed to stop inlet");
```
adrianbenavides, line 265: If we just log the error, should it be a warning instead?

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer/mod.rs
```diff
@@ -0,0 +1,449 @@
...
+    pub(super) context: Context,
+    pub(super) node_manager: Weak<NodeManager>,
+    pub(super) inlet: Arc<TcpInlet>,
+    pub(super) udp_transport: Option<Arc<UdpTransport>>,
+    pub(super) status: Option<InletSessionStatus>,
+    pub(super) parameters: Arc<InletParameters>,
+}
+
+impl InletSessionReplacer {
+    /// Returns a standalone instance of the replacer, it shared the dependencies but not
```
adrianbenavides: Typo `shared` -> `shares`?

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer/mod.rs
```diff
@@ -0,0 +1,449 @@
...
+    /// the status
+    pub(super) fn standalone(&self) -> Result<InletSessionReplacer> {
```
adrianbenavides: I'm not getting the meaning of standalone in this context. Using "python terms", isn't this a "shallow clone"?

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer/mod.rs
```diff
@@ -0,0 +1,449 @@
...
+        } else {
+            return Err(Error::new(
+                Origin::Api,
+                Kind::Internal,
+                "Error while creating additional session. Connection is absent",
+            ));
+        };
+
+        let transport_route = status.connection.transport_route();
+        let additional_sc_route = transport_route.clone() + DefaultAddress::SECURE_CHANNEL_LISTENER;
```
adrianbenavides, line 320: Not the most common use case, but we are limiting this feature to the default secure channel listener. Should this be passed down from a higher layer? Or at least I'd add a TODO if it's important enough. Same with the other `DefaultAddress:*` usages below

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer/selector.rs
```diff
@@ -0,0 +1,540 @@
...
+    /// Returns the original MultiAddr and the route_key.
+    pub fn confirm(
+        mut self,
+        context: &Context,
+        route: Route,
+        options: TcpInletOptions,
+    ) -> ockam_core::Result<(MultiAddr, String)> {
+        self.inlet
+            .update_outlet_route_and_unpause(context, &self.route_key, route, options)?;
+        self.confirmed = true;
```
adrianbenavides, line 36: It's not easy to see what's happening here. If I understood correctly, if the `update_outlet_route_and_unpause` call fails, the `Drop` is triggered and removes the route.

Since this is the only place where `confirmed` is mutated, wouldn't it be the same to move the drop logic here (i.e. using `if let Err(..) = ...`)? If so, the `confirmed` field doesn't seem necessary. Probably consuming the instance isn't needed either.

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer/mod.rs
```diff
@@ -0,0 +1,449 @@
...
+        .await?;
+
+        let options = TcpInletOptions::new()
+            .set_skip_handshake(self.parameters.skip_handshake)
+            .set_enable_nagle(self.parameters.enable_nagle)
+            .with_incoming_access_control(incoming_ac)
+            .with_outgoing_access_control(outgoing_ac);
+
+        let options = if self.udp_puncture_enabled() && self.parameters.disable_tcp_fallback {
+            options.paused()
```
adrianbenavides, line 115: Could you add more context about this? Why are udp inlets paused initially if there is no TCP fallback?

implementations/rust/ockam/ockam_transport_tcp/src/portal/inlet_shared_state.rs
davide-baldo: It should be called once for each new inlet connection, so it shouldn't have any measurable impact. But yeah, we need to be careful with random usage!

implementations/rust/ockam/ockam_transport_tcp/src/portal/inlet_shared_state.rs
davide-baldo: I agree with you! But it's a verb also used in outlets, and I didn't want to make the PR harder to review by refactoring code outside the scope of the PR.
I'll try to incorporate the change in a future PR.

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/node_manager.rs
davide-baldo: My reasoning is that it should never happen (likely a bug), and if we fail to stop an inlet, the relative port will not available anymore for other usages.

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer/mod.rs
davide-baldo: It's a partial shallow clone? :)
Would `clone_without_status()` be more readable?

adrianbenavides: Sounds good to me 👍

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer/selector.rs
davide-baldo: From the moment the route gets selected to when it gets confirmed, things could "go badly" 3 times.

> wouldn't it be the same to move the drop logic here

Yes, it would be logically the same, but you would need to avoid using `?` and manually manage each failure, or create a sub-method and validate the result. But neither would protect against future refactorings. On the other hand, the Drop semantic can guarantee it that it will either be confirmed or removed, like a typical rust lock.

> Probably consuming the instance isn't needed either.

It's not needed, but this prohibits you to confirming the selection multiple times at compile-time. It's a way to prevent misusing of the API.

Since it's synchronization code, which is challenging to test and validate; I'm being careful, and I'm trying to make the code as foolproof as I can even if it increases complexity.

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer/mod.rs
davide-baldo: I've added a comment:

```
// By default, the portal uses the typical route, and switches to UDP puncture once the
// new route gets established.
// But when `tcp_fallback` is disabled, we want *all* traffic to pass through *only*
// the UDP puncture, so we pause the portal until the UDP puncture is complete.
```

implementations/rust/ockam/ockam_api/src/control_api/protocol/inlet.rs
davide-baldo: It's rendered as markdown!
I've added a semicolon to separate the statements.

implementations/rust/ockam/ockam_api/src/control_api/protocol/inlet.rs
davide-baldo: Overall, I took a more "classical" API approach, where the API is low-level as possible, and everything else can be added on top via libraries.
Given the last meeting, I probably took the wrong approach and we should lift the API level.

implementations/rust/ockam/ockam_api/src/influxdb/portal.rs
davide-baldo: In influxDb use case, a new MultiAddr toward the token issuer is derived from the outlet MultiAddr. This is before any connection attempt is made, and doesn't handle any form of high-availability for now.

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer/mod.rs
davide-baldo: Added a TODO to extract it from the original MultiAddr

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/session_replacer/selector.rs
adrianbenavides: Got it, thanks for the context 🙏

adrianbenavides:
I checked that the replacer's messages are OK, and they look good. One difference I noticed is that the inlet now reacts almost instantly when it loses the connection (compared to the ~30 seconds we had before):

image

implementations/rust/ockam/ockam_api/src/cli_state/storage/tcp_portals_repository.rs
```diff
@@ -71,37 +71,37 @@ impl<T: TcpPortalsRepository> TcpPortalsRepository for AutoRetry<T> {
...
+    outlet_addresses: Vec<MultiAddr>,
     alias: String,
     privileged: bool,
 }
 
 impl TcpInlet {
     pub fn new(
         bind_addr: &SocketAddr,
-        outlet_addr: &MultiAddr,
+        outlet_addresses: &[MultiAddr],
```
SanjoDeundiak: NIT: Let's make it a Vector here

implementations/rust/ockam/ockam_api/src/cli_state/storage/tcp_portals_repository.rs
```diff
@@ -71,37 +71,37 @@ impl<T: TcpPortalsRepository> TcpPortalsRepository for AutoRetry<T> {
...
 
-    pub fn bind_addr(&self) -> SocketAddr {
-        self.bind_addr
+    pub fn bind_address(&self) -> SocketAddr {
+        self.bind_address
     }
 
-    pub fn outlet_addr(&self) -> MultiAddr {
-        self.outlet_addr.clone()
+    pub fn outlet_addresses(&self) -> &Vec<MultiAddr> {
```
SanjoDeundiak: NIT: I would return a slice here, unless a caller of that function rely on the fact that it's a Vec (I can't think of such case)

implementations/rust/ockam/ockam_api/src/cli_state/storage/tcp_portals_repository.rs
```diff
@@ -71,37 +71,37 @@ impl<T: TcpPortalsRepository> TcpPortalsRepository for AutoRetry<T> {
...
+        &self.outlet_addresses
     }
 
-    pub fn alias(&self) -> String {
-        self.alias.clone()
+    pub fn alias(&self) -> &str {
```
SanjoDeundiak, line 103: Nice

implementations/rust/ockam/ockam_api/src/cli_state/storage/tcp_portals_repository_sql.rs
```diff
@@ -62,8 +62,14 @@ impl TcpPortalsRepository for TcpPortalsSqlxDatabase {
...
         .bind(node_name)
-        .bind(tcp_inlet.bind_addr().to_string())
-        .bind(tcp_inlet.outlet_addr().to_string())
+        .bind(tcp_inlet.bind_address().to_string())
+        .bind(
+            tcp_inlet
+                .outlet_addresses()
+                .iter()
+                .map(|x| x.to_string())
+                .join("//"),
```
SanjoDeundiak: NIT: A small comment would be great, but overall was easy to understand

implementations/rust/ockam/ockam_api/src/cli_state/storage/tcp_portals_repository_sql.rs
```diff
@@ -158,18 +164,26 @@ struct TcpInletRow {
...
+            .map_err(|e| Error::new(Origin::Api, Kind::Serialization, format!("{e:?}")))
     }
 
-    fn outlet_addr(&self) -> Result<MultiAddr> {
-        MultiAddr::from_str(&self.outlet_addr)
-            .map_err(|e| ockam_core::Error::new(Origin::Api, Kind::Serialization, format!("{e:?}")))
+    fn outlet_addresses(&self) -> Result<Vec<MultiAddr>> {
+        let mut multiaddresses = Vec::new();
+
+        for addr in self.outlet_addr.split("//") {
```
SanjoDeundiak: Could that be shrank into `self.outlet_addr.split("//").map()`? Or is error handling ruining it?

implementations/rust/ockam/ockam_api/src/control_api/backend/inlet.rs
```diff
@@ -155,16 +156,30 @@ async fn handle_tcp_inlet_create(
         )?),
     };
 
+    if request.to.is_empty() {
+        return ControlApiHttpResponse::bad_request("`to` must not be empty");
+    }
+
+    let to = {
+        let mut to = Vec::new();
```
SanjoDeundiak: NIT: Same question here. Or also possible to reserve capacity at least

implementations/rust/ockam/ockam_api/src/control_api/protocol/inlet.rs
SanjoDeundiak: `Duration` is a part of `core` so I don't mind using it everywhere where applicable

implementations/rust/ockam/ockam_command/tests/bats/local/portals.bats
```diff
@@ -142,15 +143,16 @@ teardown() {
 }
 
 @test "portals no handshake - create an inlet/outlet pair and move tcp traffic through it, where the outlet points to an HTTPs endpoint" {
+  skip
```
SanjoDeundiak: Why it no longer pass?

implementations/rust/ockam/ockam_transport_tcp/src/portal/inlet_listener.rs
```diff
@@ -178,30 +171,33 @@ impl Processor for TcpInletListenProcessor {
...
 
         let addresses = Addresses::generate(PortalType::Inlet {
             listener_address: HostnamePort::from(
                 self.inner.local_addr().map_err(TransportError::from)?,
             ),
         });
-
-        let inlet_shared_state = self.inlet_shared_state.read().unwrap().clone();
-
-        if inlet_shared_state.is_paused() {
```
SanjoDeundiak, line 200: What is the motivation for that change?

implementations/rust/ockam/ockam_transport_tcp/src/privileged_portal/workers/internal_processor.rs
```diff
@@ -72,12 +69,6 @@ impl Processor for InternalProcessor {
         match &self.mode {
             // Client -> Inlet packet
             PortalMode::Inlet { inlet } => {
-                let inlet_shared_state = inlet.inlet_shared_state.read().unwrap().clone();
-
-                if inlet_shared_state.is_paused() {
```
SanjoDeundiak, line 75: Could you please elaborate?

implementations/rust/ockam/ockam_transport_tcp/src/privileged_portal/workers/internal_processor.rs
```diff
@@ -88,11 +79,6 @@ impl Processor for InternalProcessor {
                             raw_socket_read_result.ipv4_info.source_ip(),
                             raw_socket_read_result.tcp_info.source_port(),
                         );
-
-                        if connection.their_identifier != inlet_shared_state.their_identifier() {
```
SanjoDeundiak, line 90: Are you removing that check just because it's checked in a different place, or some other reason?

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/node_manager.rs
```diff
@@ -109,135 +118,162 @@ impl NodeManager {
...
+                    !disable_tcp_fallback,
+                    ping_timeout,
+                ))
+            } else {
+                None
+            };
+
+            let mut session =
+                Session::create(ctx, replacer, additional_session_options, ping_timeout)?;
+            session.start_monitoring()?;
```
SanjoDeundiak, line 210: Does it make sense to start with a first session instead?

implementations/rust/ockam/ockam_transport_tcp/src/portal/inlet_shared_state.rs
```diff
@@ -1,25 +1,131 @@
...
+                    break active_routes
+                        .choose(&mut rand::thread_rng())
+                        .copied()
+                        .cloned()
+                        .unwrap();
+                }
+            }
+            // let's wait until something changes and try again
+            debug!("No active route found for the inlet, waiting for a connection");
+            self.notify.notified().await;
```
SanjoDeundiak, line 80: I think for both regular and privileged portals `choose_active_route` would better return an error instead of waiting for a connection. WDYT?

implementations/rust/ockam/ockam_transport_tcp/src/privileged_portal/workers/internal_processor.rs
```diff
@@ -122,15 +110,15 @@ impl Processor for InternalProcessor {
 
                 let portal_packet = OckamPortalPacket::from_tcp_packet(
                     connection.connection_identifier.clone(),
-                    inlet_shared_state.route_index(),
+                    connection.inlet_route_state.route_index(),
```
SanjoDeundiak: It uses locking internally and locks again to get the route ~10 lines below that. That's a small performance penalty, and also brings a possibility of a data race.

implementations/rust/ockam/ockam_transport_tcp/tests/portal.rs
```diff
@@ -283,23 +272,18 @@ async fn portal__tcp_connection_with_invalid_message_flow__should_not_succeed__i
...
     drop(stream);
 
-    tokio::time::sleep(Duration::from_millis(250)).await;
-
     Ok(())
 }
 
 #[allow(non_snake_case)]
-#[ockam_macros::test(timeout = 5000)]
+#[ockam_macros::test(timeout = 5000_0000)]
```
SanjoDeundiak: Probably left from debugging

SanjoDeundiak (changes requested):
This is a big PR, could only consume a part of it so far. Submitting what I have

implementations/rust/ockam/ockam_transport_tcp/src/portal/inlet_listener.rs
davide-baldo: A valid "unpaused" route was already selected here on line 174:
```rust
        let inlet_route_state = self.inlet_shared_state.choose_active_route().await;
```
The condition would be always false

implementations/rust/ockam/ockam_transport_tcp/src/privileged_portal/workers/internal_processor.rs
davide-baldo: We want identifier consistency when we switch from relay to the direct UDP puncture connection.
I moved the check when we update the route, it cannot be updated if the identifier is changed.

`inlet_shared_state.rs:242`

```rust
   if let Some(current_identifier) = &self.their_identifier {
            if let Some(new_identifier) = &their_identifier {
                if current_identifier != new_identifier {
                    return Err(ockam_core::Error::new(
                        Origin::Channel,
                        Kind::Conflict,
                        "Route identifier has changed",
                    ));
                }
            } else {
                return Err(ockam_core::Error::new(
                    Origin::Channel,
                    Kind::Conflict,
                    "Route identifier not found",
                ));
            }
        }
````

implementations/rust/ockam/ockam_api/src/nodes/service/tcp_inlets/node_manager.rs
davide-baldo: We are just spawning new tasks, it should be <<1ms.
If we were to run the first session first, we would delay every other session until the first is established (or fails).

implementations/rust/ockam/ockam_transport_tcp/src/portal/inlet_shared_state.rs
davide-baldo: I think both are acceptable, and it could be an option:
By waiting for it, the connection connects to the TCP inlet, starts sending data, and then it waits until a route is established.
The alternative being a direct rejection of the connection.

The former has the advantage of avoiding the client polling time, and manage to resume the connection as soon as it's available.
The latter has the advantage of better mimicking the availability of the service.

implementations/rust/ockam/ockam_transport_tcp/tests/portal.rs
davide-baldo: nice catch

implementations/rust/ockam/ockam_transport_tcp/src/privileged_portal/workers/internal_processor.rs
davide-baldo: Are you right, accessing the data structure with multiple locks and unlocks is bad.
I've created two methods: `snapshot` and `snapshot_route` that return a consistent view.

implementations/rust/ockam/ockam_api/src/cli_state/storage/tcp_portals_repository_sql.rs
davide-baldo: I managed to collect the Result as well

implementations/rust/ockam/ockam_api/src/control_api/backend/inlet.rs
davide-baldo: I've converted it using iterative tools!

implementations/rust/ockam/ockam_transport_tcp/src/privileged_portal/workers/internal_processor.rs
davide-baldo: same as above, since the route comes from `choose_active_route` it cannot be paused

PR #8854, opened 2025-03-11
Instrument portal messages

etorreborre:
This PR:

 - Sets most of the instrumented spans with the level `TRACE`. The idea is to leave those spans for a low-level analysis of where time is spent, and reintroduce them at the `INFO` level only once we know that they are useful for high-level observability in a project UI.
 
 - Adds an attribute to distinguish host user journeys (all commands from a given node) from project user journey (all commands for a given project).

 - Reworks the spans emitted when a message is sent via a portal:
   - inlet -> outlet: `receive_tcp_message_at_inlet`, then `receive_ockam_message_from_inlet`.  
   - outlet -> inlet: `receive_tcp_message_at_outlet`, then `receive_ockam_message_from_outlet`

 - Additionally, node information (node name + node identifier) and project id have been added to all the spans when the information is available:
   - This is why there is a commit with a large number of changes; the `CliState` had to be shared with the `SpanExporter` to collect that information (if available) when spans are being exported.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/journeys/journeys.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/journeys/journeys.rs
@@ -8,7 +8,7 @@ use crate::{CliState, Result};
 use chrono::{DateTime, Utc};
 use either::Either;
 use ockam_core::{OpenTelemetryContext, OCKAM_TRACER_NAME};
-use opentelemetry::trace::{Link, SpanBuilder, SpanId, TraceContextExt, TraceId, Tracer};
+use opentelemetry::trace::{Link, Span, SpanBuilder, SpanId, TraceContextExt, TraceId, Tracer};
 use opentelemetry::{global, Context, Key, KeyValue};
 use std::collections::HashMap;
 use std::fmt::{Display, Formatter};
@@ -100,7 +100,7 @@ impl CliState {
         event: JourneyEvent,
         attributes: HashMap<&Key, String>,
     ) -> Result<()> {
-        self.add_a_journey_event(event, attributes).await
+        self.add_a_journey_event(event, attributes, None).await
     }
 
     /// This method adds an error event to the project/host journeys
@@ -112,8 +112,9 @@ impl CliState {
         attributes: HashMap<&Key, String>,
     ) -> Result<()> {
         self.add_a_journey_event(
-            JourneyEvent::error(command_name.to_string(), message),
+            JourneyEvent::error(command_name.to_string(), message.clone()),
             attributes,
+            Some(message),
         )
         .await
     }
@@ -126,6 +127,7 @@ impl CliState {
         &self,
         event: JourneyEvent,
         attributes: HashMap<&Key, String>,
+        error_message: Option<String>,
     ) -> Result<()> {
         if !self.is_tracing_enabled() {
             return Ok(());
@@ -135,11 +137,25 @@ impl CliState {
         let (host_journey, project_journey) = self
             .get_journeys(project.clone().map(|p| p.project_id().to_string()))
             .await?;
-        self.make_span_from_journey(&project, &event, &attributes, host_journey, "host")
-            .await?;
+        self.make_span_from_journey(
+            &project,
+            &event,
+            &attributes,
+            &error_message,
+            host_journey,
+            "host",
+        )
+        .await?;
         if let Some(project_journey) = project_journey {
-            self.make_span_from_journey(&project, &event, &attributes, project_journey, "project")
-                .await?;
+            self.make_span_from_journey(
+                &project,
+                &event,
+                &attributes,
+                &error_message,
+                project_journey,
+                "project",
+            )
+            .await?;
         }
         Ok(())
     }
@@ -153,6 +169,7 @@ impl CliState {
         project: &Option<Project>,
         event: &JourneyEvent,
         attributes: &HashMap<&Key, String>,
+        error_message: &Option<String>,
         journey: Journey,
         trace_type: &str,
     ) -> Result<()> {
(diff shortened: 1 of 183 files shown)


implementations/rust/ockam/ockam_api/src/cli_state/journeys/journeys.rs
```diff
@@ -217,25 +277,22 @@ impl CliState {
     }
 
     /// Return a list of journeys for which we want to add spans
```
adrianbenavides: The description should be updated to reflect the changes in the returned objects right? Related to this, is there a central place where the host/project journeys are explained?

implementations/rust/ockam/ockam_api/src/logs/span_exporters.rs
```diff
@@ -38,22 +49,46 @@ impl<S: SpanExporter> DecoratedSpanExporter<S> {
...
+        let span_attributes = self.span_attributes.clone();
+
+        let f = async move {
+            let mut exporter = exporter.lock().await;
+
+            // initialize span attributes from local data if they haven't been initialized yet.
+            let mut span_attributes = span_attributes.lock().await;
+            let attributes = if span_attributes.is_none() {
+                SpanAttributes {
+                    node_info: cli_state.get_default_node().await.ok(),
```
adrianbenavides: Now that we don't have an explicit node name here, how is this initialized for a non-default node?

implementations/rust/ockam/ockam_command/src/command.rs
```diff
@@ -110,16 +112,16 @@ impl OckamCommand {
             return None;
         };
 
-        let app_name = if self.subcommand.is_background_node() {
+        let app_name = if self.subcommand.is_local_node() {
             "local node"
         } else {
             "cli"
```
adrianbenavides, line 118: How relevant are this strings? Not enough to extract them as constants I guess?

implementations/rust/ockam/ockam_command/src/subcommand.rs
```diff
@@ -254,28 +254,33 @@ impl OckamSubcommand {
...
+
+            OckamSubcommand::Authority(cmd) => match &cmd.subcommand {
+                AuthoritySubcommand::Create(cmd) => cmd.foreground || cmd.child_process,
+            },
+            _ => false,
+        }
+    }
+
+    /// Return the node name for an ockam node create command
+    pub fn node_name(&self) -> Option<String> {
```
adrianbenavides, line 275: This function doesn't seem to be used. Should we remove it?

implementations/rust/ockam/ockam_core/src/routing/message/local_message.rs
```diff
@@ -209,53 +209,12 @@ impl LocalMessage {
...
-    /// - A new trace is started
-    /// - The previous trace and the new trace are linked together
-    ///
-    /// We start a new trace here in order to make sure that each transport message is always
-    /// associated to a globally unique trace id and then cannot be correlated with another transport
-    /// message that would leave the same node for example.
-    ///
-    /// We can still navigate the two created traces as one thanks to their link.
-    #[cfg(feature = "std")]
-    pub fn start_new_tracing_context(
```
adrianbenavides, line 230: What are the implications of not using this anymore? Does this mean that we now have a single trace?

implementations/rust/ockam/ockam_transport_tcp/src/portal/portal_worker.rs
```diff
@@ -620,6 +632,10 @@ impl Worker for TcpPortalWorker {
 
                 if remote_packet {
                     let msg = PortalMessage::decode(&payload)?;
+                    if !msg.is_disconnect() {
+                        let span = self.start_span(ctx, &their_identifier)?;
```
adrianbenavides, line 636: A general question about spans. Once a span is created, it's handled by tracing? You don't need to care about it being dropped and stuff like that right?

implementations/rust/ockam/ockam_api/src/cli_state/journeys/journeys.rs
etorreborre: Thanks, I updated the comment.

> 

There's not a full write-up on the subject. There's:

- A description of the [high-level API for journeys](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/cli_state/journeys/journeys.rs#L61).
- A description explaining that [a Journey is a pseudo-trace](https://github.com/build-trust/ockam/blob/develop/implementations/rust/ockam/ockam_api/src/cli_state/journeys/journey.rs#L5).

implementations/rust/ockam/ockam_api/src/logs/span_exporters.rs
etorreborre: You're right, this is problematic. I've reintroduced the optional node name.

implementations/rust/ockam/ockam_command/src/command.rs
etorreborre: There's no place where they would be reused so that's why I left them as literal. 
They try to indicate the difference between events that are resulting from user interactions, get node lists, add a listener, etc... from a running node.

implementations/rust/ockam/ockam_command/src/subcommand.rs
etorreborre: Now it's used again :-).

implementations/rust/ockam/ockam_core/src/routing/message/local_message.rs
etorreborre: Yes, we have now a single trace per message, instead of having one trace per node and having links between those traces (because that would be harder to query with SQL)

implementations/rust/ockam/ockam_transport_tcp/src/portal/portal_worker.rs
etorreborre: When a span is created as above it is started it with a specific tracer (the global one) and the tracer takes care of it.
This is a bit different if you use the `tracing` crate directly because then you have to make sure you enter the span, and then when it is dropped, it will get an end timestamp. 

Your question makes me realize that in the code above we don't end the span. Since we don't track request/responses (we get an independent trace for the response message in a request/response scenario), we don't ever end the span. The better thing would be to end the current span when a transport message leaves the node. At least that would give us an idea of how long it took to process a message on a given node. I will have a look at that later when I come back to improving our data.

adrianbenavides (approved):
🚀

PR #8861, opened 2025-03-14
feat(rust): deprecating `--status-endpoint-port` in favor of `--status-endpoint`

davide-baldo:


implementations/rust/ockam/ockam_api/src/nodes/service/http.rs
```diff
@@ -33,10 +33,10 @@ impl HttpServer {
     pub async fn start(
         context: &Context,
         node_manager: Arc<NodeManager>,
-        port: Port,
+        port: BindAddress,
     ) -> Result<SocketAddr> {
         debug!("Starting HTTP server on port: {port:?}");
```
adrianbenavides: The variable should be renamed here, and the debug text updated (maybe to `Starting HTTP server at: {bind_address:?}`)

implementations/rust/ockam/ockam_api/src/nodes/service/in_memory_node.rs
```diff
@@ -136,7 +136,7 @@ impl InMemoryNode {
         ctx: &Context,
         cli_state: &CliState,
         identity_name: &str,
-        status_endpoint_port: Option<Port>,
+        status_endpoint_port: Option<BindAddress>,
```
adrianbenavides: missing rename here too

adrianbenavides:
To keep it consistent, can you change [this](https://github.com/build-trust/ockam/blob/5d0e430c63d260e7a480b8fc0d449df536a5695d/implementations/rust/ockam/ockam_transport_core/src/hostname_port.rs#L182) to also use `Ipv4Addr::LOCALHOST.to_string()`?

implementations/rust/ockam/ockam_api/src/nodes/models/transport/bind_address.rs
```diff
@@ -0,0 +1,63 @@
...
+
+impl BindAddress {
+    pub fn explicit(address: String, port: u16) -> BindAddress {
+        Self {
+            port: Port::Explicit(port),
+            address,
+        }
+    }
+
+    pub fn try_explicit_or_random(address: String, port: u16) -> BindAddress {
```
adrianbenavides: Not 100% convinced about resurfacing the `Port` functions here. I'd pass a `Port` instance instead and just have a `new` function.

adrianbenavides (approved):
🚀

PR #8864, opened 2025-03-14
Control API Fixes - timeout and inlets

davide-baldo:
When request timeouts, every other request is queued in the meantime, causing outages.
This PR fixes it by spawning a dedicated task for each connection.
Another unrelated paper-cut is to return an explicit error when the inlet name or port are already in-use.

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/control_api/frontend.rs
+++ b/implementations/rust/ockam/ockam_api/src/control_api/frontend.rs
@@ -63,25 +63,34 @@ impl Processor for HttpControlNodeApiFrontend {
             }
         };
 
-        let service = service_fn(|request| {
-            let context = context.clone();
-            Self::route_request(
-                request,
-                context,
-                self.node_manager.clone(),
-                self.node_resolution.clone(),
-                self.authentication_token.clone(),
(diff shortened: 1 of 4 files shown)


implementations/rust/ockam/ockam_api/src/control_api/frontend.rs
```diff
@@ -63,25 +63,34 @@ impl Processor for HttpControlNodeApiFrontend {
...
+            })
+        };
 
-        if let Err(err) = http1::Builder::new()
-            .serve_connection(TokioIo::new(stream), service)
-            .await
-        {
-            error!("HTTP server error: {:?}", err);
-        }
+        tokio::spawn(async move {
```
adrianbenavides, line 86: To make sure I'm following, the problem we had before this fix is that a stuck connection was blocking this processor in particular right?

adrianbenavides (approved):

implementations/rust/ockam/ockam_api/src/control_api/frontend.rs
davide-baldo: correct!

PR #8865, opened 2025-03-17
feat(rust): add env var to configure the orchestrator ui url

adrianbenavides:
Useful when testing the command with a local/dev orchestrator. The custom URL is used in the `enroll` command, when the user is asked to subscribe and create a space.

image


etorreborre (approved):
Nice 🌷.

PR #8870, opened 2025-03-19
feat(rust): unify output of crud tcp-connection commands

adrianbenavides:
Follow up of https://github.com/build-trust/ockam/pull/8867

- Use the same output structure on show/delete/list. 
- Add interactive behavior to show/delete

image


etorreborre (approved):
📟

PR #8871, opened 2025-03-19
feat(rust): unify output of show/list portals commands

adrianbenavides:
- Use same output struct in `tcp-outlet show/list` commands
- Add custom serde methods for the outlet status address to serialize it as a multiaddr route (to keep the existing functionality while using a struct with typed addresses, instead of the [ad-hoc struct](https://github.com/build-trust/ockam/pull/8871/files#diff-110572d0ecb5099ab054d64a4f0a69555dcf8bb5ac454f7673aed15c4664a360L61) with strings)

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/cli_state/storage/tcp_portals_repository_sql.rs
+++ b/implementations/rust/ockam/ockam_api/src/cli_state/storage/tcp_portals_repository_sql.rs
@@ -108,7 +108,7 @@ impl TcpPortalsRepository for TcpPortalsSqlxDatabase {
         )
         .bind(node_name)
         .bind(tcp_outlet_status.to.to_string())
-        .bind(tcp_outlet_status.worker_addr.to_string())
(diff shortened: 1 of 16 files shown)


etorreborre (commented):
[UPDATE] I was wrong there's no `IO` involved in the serde functions.

The functional programmer in me is reluctant to put code doing some IO in deserialization functions:

- Because it can be surprising to get network errors when deserializing something ([see the famous Java example](https://www.reddit.com/r/programminghorror/comments/om7r3p/javas_urlequals_performs_dns_lookups/)).
- Also because the code inside `deserialize_address_from_local_service` cannot be reused elsewhere to do the same string processing.

So I would propose to keep a separate data type that's serialized with strings (after all, that's what we get, we should probably have used `MultiAddr` from the beginning for that field). And then do an additional processing step to get a proper local address.

etorreborre (approved):

PR #8876, opened 2025-03-24
feat(rust): node http api portal endpoints args refinement

adrianbenavides:
## General changes

- Update the fields descriptions of the portals endpoints to match the Command's arguments descriptions
- Add example values for requests/responses so the docs snippets are easier to understand
- Network addresses of format `host:port` are now supported as a string (`localhost:0`) or as an object (`{"hostname": "...", "port": 0"}`) in the requests, and returned as an object in the responses for backwards compatibility
## Outlet
- Add a `Validated` struct for outlet/inlet requests, to bridge the limitations of expressing types in OpenAPI schemas with the fully typed fields we use internally within the NodeManager

- The response now returns both `address` and `name`, and `address` is marked as deprecated. We'll remove it in the future, after existing examples/demos have been updated to use the new `name` field
- The `POST` endpoint:
  - `kind`: is now optional, as in post@inlet
  - `address`: has been renamed to `name`, but it's been added as a valid alias for backwards compatibility

## Inlet

To mimic the shorthand features of the `tcp-inlet create` command, the `POST` endpoint now has the following changes:
- `to`: also supports passing the name of the outlet service, typically used when also passing `via`
- `via`: to define the name of the relay

### Breaking changes

- Inlet's `POST` endpoint, `kind` field: some variants had [a typo](https://github.com/build-trust/ockam/pull/8876/files#diff-75553af96e05a4e8cd6d8729fc6bc74e20975aee074ef98c5299dd1fcde49567R22) that is now fixed
- Inlet response: `current_route` has been [renamed](https://github.com/build-trust/ockam/pull/8876/files#diff-75553af96e05a4e8cd6d8729fc6bc74e20975aee074ef98c5299dd1fcde49567R220) to `internal_route`. This field returns the ockam route, and it's used for debugging purposes only (example: `"0#b9c31602e2cb4d03669bb81662b61ca0 => 0#my-outlet"`)

### DIFF
--- a/implementations/rust/ockam/ockam_api/src/control_api/protocol/outlet.rs
+++ b/implementations/rust/ockam/ockam_api/src/control_api/protocol/outlet.rs
@@ -22,7 +22,7 @@ pub enum OutletKind {
 #[derive(Debug, Serialize, Deserialize, EnumString, ToSchema)]
 #[serde(rename_all = "kebab-case")]
 pub enum OutletTls {
-    /// If the destination uses TLS, the connection will be fully validated.
+    /// The destination is expected to be a TLS endpoint and will be fully validated.
     Validate,
 }
 
--- a/implementations/rust/ockam/ockam_api/src/control_api/protocol/outlet.rs
+++ b/implementations/rust/ockam/ockam_api/src/control_api/protocol/outlet.rs
@@ -1,4 +1,4 @@
-use crate::control_api::protocol::common::HostPort;
+use crate::control_api::protocol::common::{HostPortRequest, HostPortResponse};
 use crate::control_api::ControlApiError;
 use ockam_abac::PolicyExpression;
 use ockam_core::Address;
@@ -43,7 +43,7 @@ pub struct CreateOutletRequest {
     /// Network address where your application is listening to, in the format `<host>:<port>`.
     /// Your TCP Outlet will forward raw TCP traffic to this destination.
     #[schema(example = "dev.environment:1234")]
-    pub to: HostPort,
+    pub to: HostPortRequest,
 
     /// The TLS configuration for the TCP Outlet.
     #[serde(default)]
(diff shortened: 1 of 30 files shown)


implementations/rust/ockam/ockam_api/src/control_api/protocol/outlet.rs
```diff
@@ -14,64 +19,111 @@ pub enum OutletKind {
...
 
-#[derive(Debug, Serialize, Deserialize, Default, ToSchema)]
+#[derive(Debug, Serialize, Deserialize, EnumString, ToSchema)]
 #[serde(rename_all = "kebab-case")]
 pub enum OutletTls {
-    #[default]
-    /// No TLS
-    None,
-    /// The destination uses TLS, the connection will be fully validated.
+    /// If the destination uses TLS, the connection will be fully validated.
```
davide-baldo: > `If the destination uses TLS, the connection will be fully validated.`

The user may think that a plain connection could still work with `validate` but this is not the case

davide-baldo (dismissed):

implementations/rust/ockam/ockam_api/src/control_api/protocol/outlet.rs
adrianbenavides: What do you mean? How can we make the description more precise?

davide-baldo: Something similar to "The destination is expected to be a TLS endpoint and will be fully validated"

davide-baldo (dismissed):

implementations/rust/ockam/ockam_api/src/control_api/protocol/outlet.rs
adrianbenavides: Ok, that's clearer indeed. I've just pushed that change.

davide-baldo (approved):

PR #8885, opened 2025-03-27
fix(rust): `node create` with configuration was swallowing an error produced in a subprocess

adrianbenavides:
When running `node create` with a configuration, the tokio task handle was returning an error before it was ever awaited, blocking the `callback.wait_for_signal()` call.

The task handle and the callback are now awaited in a tokio::select block, as we do in [wait_for_node_callback](https://github.com/build-trust/ockam/blob/985af6947b679fb36a3cd1f33e4310307bcece1e/implementations/rust/ockam/ockam_command/src/node/util.rs#L243).


SanjoDeundiak (approved):
Nice fix! Thank you

PR #8903, opened 2025-04-05
feat(rust): add ai subcommands for enroll and ticket

adrianbenavides:
- most of the changes are moving the `enroll` logic into a separate struct, so it can be reused in both `enroll` commands
- `ai enroll`: the same as regular `enroll` with some differences:
  - it passes a flag to the `enroll` orchestrator endpoint
  - it doesn't have the `force` and `skip_orchestrator_resources_creation` arguments, which default to false and true, respectively
  
- `ai ticket`: the same as `project ticket` for now

In the future, both commands might have a custom output.


adrianbenavides:
Closed in favor of https://github.com/build-trust/ockam/pull/8922

PR #8910, opened 2025-04-08
feat(rust): add shared access to sending messages from a `Context`

SanjoDeundiak:

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/context/context_send.rs
+++ b/implementations/rust/ockam/ockam_node/src/context/context_send.rs
@@ -1,0 +1,86 @@
+use crate::context::ContextState;
+use crate::{Context, NodeError};
+use ockam_core::compat::sync::{Arc, Weak};
+use ockam_core::{Address, Result};
+use ockam_core::{Message, Route};
+
+/// Cloneable [`Context`] variant that allows shared access to sending capabilities from a specified
+/// address. Doesn't own any mailbox, therefore original [`Context`] must be alive to be able
+/// to send the message.
+#[derive(Clone)]
+pub struct ContextSend {
+    primary_address: Address,
+    state: Weak<ContextState>,
+}
+
+impl ContextSend {
+    /// Send a message to an address or via a fully-qualified route
+    ///
+    /// Routes can be constructed from a set of [`Address`]es, or via
+    /// the [`RouteBuilder`] type.  Routes can contain middleware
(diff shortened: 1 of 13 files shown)


implementations/rust/ockam/ockam_node/src/context/context_send.rs
```diff
@@ -0,0 +1,86 @@
+use crate::context::ContextState;
+use crate::{Context, NodeError};
+use ockam_core::compat::sync::{Arc, Weak};
+use ockam_core::{Address, Result};
+use ockam_core::{Message, Route};
+
+/// Cloneable [`Context`] variant that allows shared access to sending capabilities from a certain
+/// address. Doesn't own any mailbox, therefore original [`Context`] should be alive to be able
```
davide-baldo: > should be alive to be able

*must* be alive..?

davide-baldo: > sending capabilities from a certain address.

"certain address" feels overly generic, what about from a "specified address"?

implementations/rust/ockam/ockam_node/src/context/context_send.rs
```diff
@@ -0,0 +1,86 @@
...
+
+        state
+            .send_from_address_impl(route.into(), msg, self.primary_address.clone(), vec![])
+            .await
+    }
+}
+
+impl Context {
+    /// Get cloneable [`Context`] variant that allows shared access to sending capabilities from
+    /// a certain address. Doesn't own any mailbox, therefore original [`Context`] should be alive
```
davide-baldo: same as above, should => must?

implementations/rust/ockam/ockam_node/src/context/context_send.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// to be able to send the message.
+    pub fn get_sending_context(&self) -> ContextSend {
```
davide-baldo, line 80: we should give the possibility to choose a source address other than the primary one

davide-baldo (dismissed):

implementations/rust/ockam/ockam_node/src/context/context_send.rs
SanjoDeundiak: I can imagine that could be required at some point. But I don't think we need it now. So keeping things as simple as possible

davide-baldo (approved):

PR #8921, opened 2025-04-11
feat(rust): expose `Abac`

SanjoDeundiak:


implementations/rust/ockam/ockam_api/src/nodes/service/manager.rs
```diff
@@ -656,6 +657,31 @@ impl NodeManager {
         ))
     }
 
+    pub async fn abac(
```
etorreborre: Maybe `create_abac` would be a bit more explicit.

etorreborre (dismissed):
🛝

etorreborre (approved):

PR #8933, opened 2025-04-17
feat(rust): implement `message.is_local` policy

SanjoDeundiak:


davide-baldo (approved):
LGTM

implementations/rust/ockam/ockam_abac/src/abac/abac.rs
```diff
@@ -58,133 +68,166 @@ impl Abac {
...
+                                    continue;
+                                }
+                            };
+                            if key.find(|c: char| c.is_whitespace()).is_some() {
                                 warn! {
                                     policy = %expression,
                                     id     = %identifier,
-                                    "attribute key is not utf-8"
+                                    key    = %key,
+                                    "attribute key with whitespace ignored"
```
etorreborre, line 194: Did we ever document this anywhere?

implementations/rust/ockam/ockam_abac/src/abac/abac.rs
```diff
@@ -58,133 +68,166 @@ impl Abac {
     }
 }
 
+pub struct OutgoingInfo {
```
etorreborre, line 71: I'm not sure that we need those 2 structs since they contain the same field types. Maybe `MessageAbacInfo` would be enough to extract the information necessary to check the ABAC rules (and be passed directly to `is_authorized`).

implementations/rust/ockam/ockam_abac/src/abac/abac.rs
```diff
@@ -58,133 +68,166 @@ impl Abac {
...
+                        for (key, value) in attrs.attrs() {
+                            let key = match from_utf8(key) {
+                                Ok(key) => key,
+                                Err(_) => {
+                                    warn! {
+                                        policy = %expression,
+                                        id     = %identifier,
+                                        "attribute key is not utf-8"
+                                    }
+                                    continue;
```
etorreborre, line 186: Coming back to a discussion we had before, we would avoid this kind of awkward case if the attributes were well-typed from the beginning :-).

implementations/rust/ockam/ockam_core/src/transport/local_info.rs
```diff
@@ -0,0 +1,91 @@
...
+            .find(|x| x.type_identifier() == TRANSPORT_IDENTIFIER)
+        {
+            Some(local_info) => Self::from_local_info(local_info),
+            None => Err(Self::error_type_id()),
+        }
+    }
+}
+
+impl TransportLocalInfo {
+    /// Mark a `LocalInfo` vector with `IdentitySecureChannelLocalInfo`
```
etorreborre, line 77: s/IdentitySecureChannelLocalInfo/TransportLocalInfo/ ? Here and L83.

implementations/rust/ockam/ockam_core/src/transport/metadata.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn error_format() -> Error {
+        Error::new(
+            Origin::Identity,
+            Kind::Invalid,
+            "invalid format for metadata identifier for transport type",
+        )
+    }
+
+    /// Get the transport type
+    pub fn from_terminal_address_metadata(terminal: &AddressMetadata) -> Result<Self> {
```
etorreborre, line 38: The amount of dynamic typing we do here with putting untyped values in generic messages then decoding them in places we expect them makes me wonder if we could have better types. 

I guess that would require adding a type parameter for `IncomingAccessControl<M: Secured + Transported>` so that it works hand in hand with a `Worker { type Message = M }` and still allows us to check identifier + transport locality at the ABAC level, because we eventually pass around a `SecuredMessage<LocalMessage>`. 

I know that what I'm writing is really fuzzy at this stage but this kind of dynamic typing makes me think that there might be some better solutions in a language like Rust.

etorreborre (approved):
🐇

PR #8934, opened 2025-04-17
feat(rust): full separate postgres data by tenant_id

etorreborre:
This PR adds the `tenant_id` as a primary key for all tables (SQLite or Postgres).

This should guarantee that, in the Postgres case with row-level security enabled, all tables are fully scoped to a given `tenant_id`. In that situation, each tenant id will have it own project data, even if they share the same project.

### DIFF
--- a/implementations/rust/ockam/ockam_node/src/storage/database/migrations/node_migrations/sql/postgres/20250115100000_create_database.sql
+++ b/implementations/rust/ockam/ockam_node/src/storage/database/migrations/node_migrations/sql/postgres/20250115100000_create_database.sql
@@ -21,42 +21,44 @@ CREATE UNIQUE INDEX IF NOT EXISTS name_index ON _rust_migrations (name);
 CREATE TABLE identity
 (
     tenant_id      TEXT NOT NULL, -- Tenant identifier
-    identifier     TEXT NOT NULL UNIQUE,
-    change_history TEXT NOT NULL
+    identifier     TEXT NOT NULL,
+    change_history TEXT NOT NULL,
+    PRIMARY KEY (tenant_id, identifier)
 );
 
 -- This table some local metadata about identities
 CREATE TABLE named_identity
 (
-    tenant_id  TEXT NOT NULL,        -- Tenant identifier
-    identifier TEXT NOT NULL UNIQUE, -- Identity identifier
-    name       TEXT UNIQUE,          -- user-specified name
-    vault_name TEXT NOT NULL,        -- name of the vault used to store the identity keys
-    is_default BOOLEAN DEFAULT FALSE -- boolean indicating if this identity is the default one
+    tenant_id  TEXT NOT NULL,         -- Tenant identifier
+    identifier TEXT NOT NULL,         -- Identity identifier
+    name       TEXT NOT NULL,         -- user-specified name
+    vault_name TEXT NOT NULL,         -- name of the vault used to store the identity keys
+    is_default BOOLEAN DEFAULT FALSE, -- boolean indicating if this identity is the default one
+    PRIMARY KEY (tenant_id, identifier)
 );
 
+CREATE UNIQUE INDEX named_identity_index ON named_identity (tenant_id, name);
 
 -- This table lists attributes associated to a given identity
 CREATE TABLE identity_attributes
 (
     tenant_id   TEXT    NOT NULL, -- Tenant identifier
-    identifier  TEXT PRIMARY KEY, -- identity possessing those attributes
+    identifier  TEXT    NOT NULL, -- identity possessing those attributes
     attributes  BYTEA   NOT NULL, -- serialized list of attribute names and values for the identity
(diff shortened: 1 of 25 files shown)


implementations/rust/ockam/ockam_node/src/storage/database/migrations/node_migrations/sql/postgres/20250115100000_create_database.sql
```diff
@@ -21,42 +21,46 @@ CREATE UNIQUE INDEX IF NOT EXISTS name_index ON _rust_migrations (name);
...
     tenant_id   TEXT    NOT NULL, -- Tenant identifier
-    identifier  TEXT PRIMARY KEY, -- identity possessing those attributes
+    identifier  TEXT    NOT NULL, -- identity possessing those attributes
     attributes  BYTEA   NOT NULL, -- serialized list of attribute names and values for the identity
     added       INTEGER NOT NULL, -- UNIX timestamp in seconds: when those attributes were inserted in the database
     expires     INTEGER,          -- optional UNIX timestamp in seconds: when those attributes expire
     attested_by TEXT,             -- optional identifier which attested of these attributes
-    node_name   TEXT    NOT NULL  -- node name to isolate attributes that each node knows
+    node_name   TEXT    NOT NULL, -- node name to isolate attributes that each node knows
+    PRIMARY KEY (tenant_id, identifier, node_name)
```
polvorin, line 52: node_name wasn't part of the key before,  but I guess it's desired

implementations/rust/ockam/ockam_node/src/storage/database/migrations/node_migrations/sql/postgres/20250115100000_create_database.sql
```diff
@@ -21,42 +21,46 @@ CREATE UNIQUE INDEX IF NOT EXISTS name_index ON _rust_migrations (name);
...
 );
 
-CREATE UNIQUE INDEX identity_attributes_index ON identity_attributes (identifier, node_name);
+CREATE UNIQUE INDEX identity_attributes_index ON identity_attributes (tenant_id, identifier, node_name);
```
polvorin: there is a unique index already on these  created implicitly by the ` primary key ()` definition

implementations/rust/ockam/ockam_node/src/storage/database/migrations/node_migrations/sql/postgres/20250115100000_create_database.sql
```diff
@@ -67,7 +71,8 @@ CREATE TABLE credential
     scope              TEXT  NOT NULL,
     credential         BYTEA NOT NULL,
     expires_at         INTEGER,
-    node_name          TEXT  NOT NULL  -- node name to isolate credential that each node has
+    node_name          TEXT  NOT NULL,  -- node name to isolate credential that each node has
+    PRIMARY KEY (tenant_id, subject_identifier, issuer_identifier, scope)
 );
 
 CREATE UNIQUE INDEX credential_issuer_subject_scope_index ON credential (issuer_identifier, subject_identifier, scope);
```
polvorin: I _think_ this is ok,   these IIUC are credentials a node has retrieved for itself (not credentials presented by other nodes).  So,  `(issuer_identifier, subject_identifier)`   looks to be unique,  at least if there aren't two nodes with same identity.    But.. could we include the tenant_id in the unique index so it is the same than all others?.   I'm not sure if we couldn't end up in a situation where there are different nodes with same identity (I mean.. again that _probably_ is fine here,  but hard to reason about)

polvorin: Looking at the full source there are two index:

```
CREATE UNIQUE INDEX credential_issuer_subject_scope_index ON credential (issuer_identifier, subject_identifier, scope);
CREATE UNIQUE INDEX credential_issuer_subject_index ON credential (issuer_identifier, subject_identifier);
```
Not sure what's the semantic of scope there,  but only 1 of these unique index make sense to keep

polvorin (commented):
Just some comment on indexes

implementations/rust/ockam/ockam_node/src/storage/database/migrations/node_migrations/sql/postgres/20250115100000_create_database.sql
etorreborre: Thanks, I indeed forgot to remove that one.

implementations/rust/ockam/ockam_node/src/storage/database/migrations/node_migrations/sql/postgres/20250115100000_create_database.sql
etorreborre: Fixed both cases.

polvorin (approved):

PR #8952, opened 2025-04-28
feat(rust): add a send_extended method to allow sending messages with an outgoing access control

etorreborre:
This PR adds a `send_extend` method on `Node` and `Context` so that messages can be sent in a fire and forget fashion with an optional `OutgoingAccessControl` to check them.

### DIFF
--- a/implementations/rust/ockam/ockam/Cargo.toml
+++ b/implementations/rust/ockam/ockam/Cargo.toml
@@ -106,6 +106,7 @@ tracing = { version = "0.1", default-features = false }
 ockam_vault = { path = "../ockam_vault", version = "^0.133.0" }
 rand_xorshift = "0.3"
 serde_json = "1.0"
+tokio = "1.43.1"
 trybuild = { version = "1.0", features = ["diff"] }
 
 [package.metadata.cargo-machete]
--- a/implementations/rust/ockam/ockam/Cargo.toml
+++ b/implementations/rust/ockam/ockam/Cargo.toml
@@ -106,7 +106,6 @@ tracing = { version = "0.1", default-features = false }
 ockam_vault = { path = "../ockam_vault", version = "^0.133.0" }
 rand_xorshift = "0.3"
 serde_json = "1.0"
-tokio = "1.43.1"
 trybuild = { version = "1.0", features = ["diff"] }
 
 [package.metadata.cargo-machete]
--- a/implementations/rust/ockam/ockam/src/node.rs
+++ b/implementations/rust/ockam/ockam/src/node.rs
@@ -408,8 +408,6 @@ impl NodeBuilder {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::compat::tokio;
-    use ockam_core::compat::sync::RwLock;
     use ockam_core::{AllowAll, DenyAll};
     use std::time::Duration;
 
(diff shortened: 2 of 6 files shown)


implementations/rust/ockam/ockam_node/src/context/send_message.rs
```diff
@@ -149,18 +210,97 @@ impl Context {
...
+        {
+            incoming_access_control
+        } else {
+            Arc::new(AllowAll)
+        };
+
         let outgoing_access_control: Arc<dyn OutgoingAccessControl> =
-            if let Some(outgoing_access_control) = options.outgoing_access_control {
+            if let Some(outgoing_access_control) = outgoing_access_control {
                 Arc::new(AllOutgoingAccessControl::new(vec![
```
SanjoDeundiak: I know this is not related to your change, but I think `AllowOnwardAddress` is not needed in this case. Both access controls can be `incoming/outgoing_access_control.unwrap_or(AllowAll)`. Would you mind making that change? Or I can do it in a separate PR

implementations/rust/ockam/ockam_node/src/context/send_message.rs
```diff
@@ -149,18 +210,97 @@ impl Context {
...
+        let address = Address::random_tagged("Context.send.detached");
+
+        let outgoing_access_control: Arc<dyn OutgoingAccessControl> =
+            if let Some(outgoing_access_control) = outgoing_access_control {
+                Arc::new(AllOutgoingAccessControl::new(vec![
+                    outgoing_access_control,
+                    Arc::new(AllowOnwardAddress(next.clone())),
+                ]))
             } else {
-                Arc::new(AllowAll)
```
SanjoDeundiak, line 159: AllowAll is fine

implementations/rust/ockam/ockam_node/src/context/send_message.rs
```diff
@@ -149,18 +210,97 @@ impl Context {
...
+                Arc::new(AllowOnwardAddress(next.clone()))
             };
 
+        let mailboxes = Mailboxes::new(
+            Mailbox::new(
+                address.clone(),
+                None,
+                Arc::new(AllowAll),
```
SanjoDeundiak: Can it be `DenyAll`?

implementations/rust/ockam/ockam_node/src/context/send_message.rs
```diff
@@ -99,6 +99,36 @@ impl Context {
...
     }
 
+    /// Using a temporary new context, send a message
+    ///
+    /// This helper function uses [`new_detached`] and [`send`] internally.
+    /// See their documentation for more details.
+    ///
+    /// [`new_detached`]: Self::new_detached
+    /// [`send`]: Self::send
+    pub async fn send_extended<T>(
```
SanjoDeundiak, line 108: Do you have a particular use case in mind why we would need to create a temporary context to send a message, instead of sending it from the existing mailbox? Like `app` address, or whichever is available from where you're sending the message

implementations/rust/ockam/ockam/src/node.rs
```diff
@@ -391,3 +404,84 @@ impl NodeBuilder {
...
+    use ockam_core::{AllowAll, DenyAll};
+    use std::time::Duration;
+
+    #[crate::test]
+    async fn test_send_extended(ctx: &mut Context) -> Result<()> {
+        // create a detached context allowing the sending and receiving of messages
+        let ctx = ctx.new_detached(Address::random_local(), AllowAll, AllowAll)?;
+        let mut node = node(ctx).await?;
+
+        let receiver = Receiver::new();
```
SanjoDeundiak: Would it be simpler to spawn another detached `Context` instead and use `ctx.receive(timeout)`?

SanjoDeundiak: Example: https://github.com/build-trust/ockam/pull/8950/files#diff-c50c85b93df15b9702cc88a25d86e5af555c5648408398f41d33926128680792

implementations/rust/ockam/ockam/Cargo.toml
```diff
@@ -106,6 +106,7 @@ tracing = { version = "0.1", default-features = false }
 ockam_vault = { path = "../ockam_vault", version = "^0.133.0" }
 rand_xorshift = "0.3"
 serde_json = "1.0"
+tokio = "1.43.1"
```
SanjoDeundiak: Why was it required?

etorreborre: To wait a bit in the tests and check if messages were received or not.

implementations/rust/ockam/ockam_node/src/context/send_message.rs
etorreborre: Yes that makes sense.

implementations/rust/ockam/ockam_node/src/context/send_message.rs
etorreborre: How would you add the `outgoing_access_control` to the current mailbox?

implementations/rust/ockam/ockam/src/node.rs
etorreborre: I can try that.

etorreborre: Done.

implementations/rust/ockam/ockam/Cargo.toml
etorreborre: I removed it now.

implementations/rust/ockam/ockam_node/src/context/send_message.rs
SanjoDeundiak: Currently, you can't, but it should be very easy to implement.
I'm actually good with landing this PR, it's a solid change by itself. But if we can avoid creating new mailboxes for the use case you implemented this change for, that would be nice.

SanjoDeundiak (approved):

PR #8954, opened 2025-04-29
refactor: improve cluster repl

adrianbenavides:
- Use random port for the inlet
- React to ctrlc, `:quit`, and `:q` at any point
- Add fallback `docker build` command
- Add env var `OCKAM_USE_DOCKER_CACHE`
- Support big payloads. Now the server responses follow a format where the first line indicates the body length

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/base_command.rs
+++ b/implementations/rust/ockam/ockam_command/src/base_command.rs
@@ -8,16 +8,17 @@ use ockam::transport::SchemeHostnamePort;
 use ockam_api::address::get_free_address;
 use ockam_api::colors::color_primary;
 use ockam_api::orchestrator::ai_platform::node_service_client::AI_API_BASE_URL;
+use ockam_api::terminal::Terminal;
 use ockam_api::{fmt_ok, fmt_separator};
 use ockam_core::env::get_env_ignore_error;
 use ockam_core::TryClone;
 use ockam_node::Context;
-use std::io::{self, Write};
 use std::str::FromStr;
 use std::time::Duration;
 use tokio::io::{AsyncBufReadExt, AsyncReadExt, AsyncWriteExt, BufReader};
 use tokio::task::JoinHandle;
 use tokio::time::timeout;
+use tracing::debug;
 
 #[derive(Clone, Debug, Args, Default)]
 pub struct BaseCommand {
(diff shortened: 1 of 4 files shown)


implementations/rust/ockam/ockam_command/src/base_command.rs
```diff
@@ -47,6 +56,7 @@ impl BaseCommand {
         cmd.cluster_init(ctx, &opts).await?;
         let zone_config = cmd.cluster_create(ctx, &opts).await?;
         let inlet_handle = cmd.cluster_inlet(ctx, &opts, &zone_config).await?;
+        // let inlet_handle = cmd.dummy_inlet(&opts).await?;
```
davide-baldo, line 59: commented out code

implementations/rust/ockam/ockam_command/src/cluster/create.rs
```diff
@@ -39,6 +40,11 @@ pub struct CreateCommand {
     #[arg(long)]
     pub use_public_ecr: bool,
 
+    /// Whether to use the Docker cache when building the image.
+    /// It can be set using the `OCKAM_USE_DOCKER_CACHE` environment variable.
+    #[arg(long)]
```
davide-baldo, line 45: Could a standard clap environment variable work?
`#[arg(long, env = "OCKAM_...`

implementations/rust/ockam/ockam_command/src/cluster/create.rs
```diff
@@ -248,59 +257,130 @@ impl CreateCommand {
...
+                // Store the error for reporting if all commands fail
+                last_error = Some(format!(
+                    "Docker build failed with command 'docker {}' and env vars {:?}: {}",
+                    cmd_args.join(" "),
+                    env_vars,
+                    String::from_utf8_lossy(&output.stderr)
+                ));
+
+                // Log the failed attempt and continue to the next command
+                info!(
```
davide-baldo, line 366: Would the user see the same error twice?

implementations/rust/ockam/ockam_command/src/base_command.rs
adrianbenavides: I left it on purpose for now. I just keep adding it/removing it while working in this command.

implementations/rust/ockam/ockam_command/src/cluster/create.rs
adrianbenavides: I believe the limitation is that it's not parsed the same as we do with our function (e.g. parsing `1` or `true` as boolean).

implementations/rust/ockam/ockam_command/src/cluster/create.rs
adrianbenavides: It's a log, so it's not printed in the terminal unless you enable logs explicitly. In that case, I don't see the problem of logging this more than once.

polvorin (approved):

PR #8956, opened 2025-04-30
feat(rust): add more ways to send and receive messages using `Context`

SanjoDeundiak:


implementations/rust/ockam/ockam_node/src/context/receive_message.rs
```diff
@@ -1,64 +1,20 @@
...
-        self
-    }
-}
+use crate::Context;
+use crate::{debugger, MessageReceiveOptions};
 
 impl Context {
     /// Wait for the next message from the mailbox
-    pub(crate) async fn receiver_next(&mut self) -> Result<Option<RelayMessage>> {
+    pub(crate) async fn receiver_next(
```
etorreborre, line 14: Why isn't that called `receive_next`?

etorreborre (dismissed):
🏎️

implementations/rust/ockam/ockam_node/src/context/receive_message.rs
SanjoDeundiak: I don't know, maybe because it pulls the next message from the receiver. But was called like this from the very beginning)

etorreborre (approved):

PR #8962, opened 2025-05-01
feat: push secrets on `cluster create` command

adrianbenavides:
It will also be run as part of the `ockam` command if the `./secrets.yaml` file exists.

Other fixes included:
- pass relay name in cluster ticket command
- zone config return error if has duplicated pods or containers

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/cluster/inlet.rs
+++ b/implementations/rust/ockam/ockam_command/src/cluster/inlet.rs
@@ -91,6 +91,7 @@ impl InMemoryNodeCommand for InletNodeCommand {
         if let Some(allow) = &self.command.allow {
             node_config["tcp-inlet"]["allow"] = allow.to_string().into();
         }
+        let in_memory = false;
         let node_cmd = crate::node::create::CreateCommand {
             name: node_config.to_string(),
             config_args: ConfigArgs {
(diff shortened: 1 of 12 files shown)


implementations/rust/ockam/ockam_command/src/cluster/inlet.rs
```diff
@@ -28,13 +29,8 @@ before_help = docs::before_help(PREVIEW_TAG),
 after_long_help = docs::after_help(AFTER_LONG_HELP)
 )]
 pub struct InletCommand {
-    /// The Cluster that hosts the Zone.
-    #[arg(long)]
-    pub cluster: Option<String>,
```
polvorin, line 33: we *cant* infer the cluster name in the general case.
Being it  optional is ok.  For simple examples where everything is run by the same person on the same machine, letting it unspecified and retrieving it from orchestrator is fine (it is expected the user is enrolled, as is the one creating the zone, etc).

But breaks as soon as one tries to create the inlet in some machine that is *not* enrolled with ockam ai,   or if one want to create the inlet pointing to someone else cluster,  with a ticket that that other user gives us.

polvorin (changes requested):

implementations/rust/ockam/ockam_command/src/cluster/inlet.rs
adrianbenavides: Yes, my bad... it only makes sense to infer the cluster from the "cluster creator" perspective right? Basically, only this command needs the cluster argument for now

adrianbenavides: Fixed!

polvorin (approved):

PR #8980, opened 2025-05-07
refactor: process "ockam" as a clap command, enabling passing arguments to it

adrianbenavides:
- The `ockam` command now accepts arguments: `--ignore-docker-cache` and `--init-repository`. That also means we can enable logging in the base command, for example: `ockam -vv`
- The `OCKAM_USE_DOCKER_CACHE` env var has been replaced by `OCKAM_IGNORE_DOCKER_CACHE`, so the default behavior now is to use docker cache.


etorreborre (approved):
🟢

PR #8981, opened 2025-05-07
feat(rust): improvements to cluster outlet command

polvorin:
allow for automatically create ticket. Infer outlet name, build long relay names based on cluster and zone (so user don't need to know about our prefix  naming schema).

If `--enrollment-ticket` is not given,  one will be created (assuming the user is the owner of the cluster/zone) and give relay rights for the appropriate relay name.   `from`  is optional,  but instead of defaulting to `"outlet"` its default to the relay name, that is what is expected in this scenario.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/cluster/common_args.rs
+++ b/implementations/rust/ockam/ockam_command/src/cluster/common_args.rs
@@ -133,8 +133,6 @@ pub struct SecretsConfigArg {
 
 #[derive(Clone, Debug, Args, Default)]
 pub struct EnrollmentTicketConfigArg {
-    /// Enrollment ticket
-    /// If not set, one will be created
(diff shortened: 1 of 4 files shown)


implementations/rust/ockam/ockam_command/src/cluster/common_args.rs
```diff
@@ -126,3 +130,33 @@ pub struct SecretsConfigArg {
     #[arg(long, visible_alias = "secrets")]
     pub secrets_config: Option<String>,
 }
+
+#[derive(Clone, Debug, Args, Default)]
+pub struct EnrollmentTicketConfigArg {
+    /// Enrollment ticket
+    /// If not set, one will be created
```
adrianbenavides: This won't be shown, you can remove it or modify the help text below.

adrianbenavides (dismissed):

adrianbenavides (approved):

PR #8992, opened 2025-05-09
feat(rust): use rustyline for handling repl input

polvorin:
support line editing,  history,  multi-line (for multiline copied the same approach in place,  if it starts with  `"""\n`,  it uses that same pattern as the end mark.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4594,8 +4594,8 @@ dependencies = [
  "quickcheck",
  "rand",
  "regex",
- "rustyline",
- "rustyline-derive",
+ "rustyline 14.0.0",
+ "rustyline-derive 0.10.0",
  "serde",
  "serde_json",
  "sqlx",
@@ -4768,6 +4768,7 @@ dependencies = [
  "rustls 0.23.20",
  "rustls-native-certs 0.8.1",
  "rustls-pki-types",
+ "rustyline 15.0.0",
  "semver 1.0.24",
  "serde",
  "serde_json",
@@ -6416,10 +6417,44 @@ dependencies = [
 ]
 
 [[package]]
+name = "rustyline"
+version = "15.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2ee1e066dc922e513bda599c6ccb5f3bb2b0ea5870a579448f2622993f0a9a2f"
+dependencies = [
+ "bitflags 2.6.0",
+ "cfg-if",
+ "clipboard-win 5.4.0",
+ "fd-lock",
+ "home",
+ "libc",
+ "log",
+ "memchr",
+ "nix 0.29.0",
+ "radix_trie",
+ "rustyline-derive 0.11.0",
+ "unicode-segmentation",
+ "unicode-width 0.2.0",
+ "utf8parse",
+ "windows-sys 0.59.0",
+]
+
+[[package]]
 name = "rustyline-derive"
 version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e5af959c8bf6af1aff6d2b463a57f71aae53d1332da58419e30ad8dc7011d951"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.100",
+]
+
+[[package]]
+name = "rustyline-derive"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "327e9d075f6df7e25fbf594f1be7ef55cf0d567a6cb5112eeccbbd51ceb48e0d"
 dependencies = [
  "proc-macro2",
  "quote",
--- a/implementations/rust/ockam/ockam_command/Cargo.toml
+++ b/implementations/rust/ockam/ockam_command/Cargo.toml
@@ -85,6 +85,7 @@ reqwest = { version = "0.12", default-features = false, features = ["json", "rus
 rustls = { version = "0.23.13", default-features = false }
 rustls-native-certs = "0.8.0"
 rustls-pki-types = "1.7.0"
+rustyline = { version = "15.0.0", features = ["derive"] }
 semver = "1.0.23"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
--- a/implementations/rust/ockam/ockam_command/src/cluster/attach.rs
+++ b/implementations/rust/ockam/ockam_command/src/cluster/attach.rs
@@ -16,8 +16,13 @@ use ockam_api::colors::color_primary;
 use ockam_api::orchestrator::ai_platform::node_service_client::AI_API_BASE_URL;
 use ockam_api::{fmt_log, fmt_ok, fmt_separator};
 use ockam_node::{Context, Executor, NodeBuilder};
+use rustyline::config::Configurer;
+use rustyline::error::ReadlineError;
+use rustyline::validate::{ValidationContext, ValidationResult, Validator};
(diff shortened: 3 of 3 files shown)


implementations/rust/ockam/ockam_command/src/base_command.rs
```diff
@@ -524,86 +524,98 @@ impl Display for ReadStatus {
...
+        }
+        let mut rl = Editor::new().map_err(rustyline_error)?;
+        rl.set_max_history_size(50).map_err(rustyline_error)?;
+        let helper = MultiLineValidatorHelper {};
+        rl.set_helper(Some(helper));
+        rl.load_history("history.txt").map_err(rustyline_error)?;
+        loop {
+            // If we can't read from the channel, exit
+            next_r.blocking_recv().unwrap();
+            let readline = rl.readline("> ");
```
adrianbenavides: We don't receive the initial " >" from the server anymore? Or why do we need this?

implementations/rust/ockam/ockam_command/src/base_command.rs
```diff
@@ -524,86 +524,98 @@ impl Display for ReadStatus {
...
-                        eprintln!("Error reading from stdin: {}", e);
+    fn run_repl(mut next_r: Receiver<()>, lines_s: Sender<String>) -> std::result::Result<(), ()> {
+        fn rustyline_error(err: ReadlineError) {
+            eprintln!("Repl error  {:?}", err);
+        }
+        let mut rl = Editor::new().map_err(rustyline_error)?;
+        rl.set_max_history_size(50).map_err(rustyline_error)?;
+        let helper = MultiLineValidatorHelper {};
+        rl.set_helper(Some(helper));
+        rl.load_history("history.txt").map_err(rustyline_error)?;
```
adrianbenavides: The repl fails and exits if the fail doesn't exist. We could store the history files under `./.repl` and have a file per "cluster-pod-inlet".

adrianbenavides: I've added a commit to solve this comment partially. What's left to do is to have different files for different inlets, if that's needed.

adrianbenavides (dismissed):

implementations/rust/ockam/ockam_command/src/base_command.rs
```diff
@@ -256,20 +262,20 @@ impl BaseCommand {
...
 
-        // stdin quit handler
-        let stdin = StdinHandler::start();
-        let mut stdin_rx = stdin.tx.subscribe();
-        let stdin_quit = async {
-            while let Ok(input) = stdin_rx.recv().await {
-                if is_quit_sequence(&input) {
-                    break;
-                }
+        let (quit_tx, mut quit_rx) = tokio::sync::mpsc::channel(1);
```
SanjoDeundiak: May I suggest this?
```
let (quit_tx, quit_rx) = tokio::sync::oneshot::channel();
let mut quit_tx = Some(quit_tx);
ctrlc::set_handler(move || {
    if let Some(quit_tx) = quit_tx.take() {
        let _ = quit_tx.send(());
    }
})
.expect("Error setting exit signal handler");
----
tokio::select! {
    _ = quit_rx => {},
    _ = repl_handle => {},
}
```

implementations/rust/ockam/ockam_command/src/base_command.rs
```diff
@@ -524,86 +525,139 @@ impl Display for ReadStatus {
...
+        let mut rl = Editor::new()
+            .into_diagnostic()
+            .wrap_err("Failed to initialize REPL")?;
+        rl.set_max_history_size(50).into_diagnostic()?;
+        let helper = MultiLineValidatorHelper {};
+        rl.set_helper(Some(helper));
+        rl.load_history(&history_file_pat).into_diagnostic()?;
+        loop {
+            // If we can't read from the channel, exit
+            if next_r.blocking_recv().is_none() {
```
SanjoDeundiak: Any reason why `run_repl` isn't async? This way this read could be async as well

implementations/rust/ockam/ockam_command/src/base_command.rs
```diff
@@ -524,86 +525,139 @@ impl Display for ReadStatus {
...
-                        }
-                    }
-                    Err(e) => {
-                        eprintln!("Error reading from stdin: {}", e);
+    fn run_repl(
+        mut next_r: Receiver<()>,
+        lines_s: Sender<String>,
+        history_file_path: Option<PathBuf>,
+    ) -> Result<()> {
+        let history_file_pat = Self::get_history_file_path(history_file_path)?;
```
SanjoDeundiak: `history_file_path`

implementations/rust/ockam/ockam_command/src/base_command.rs
```diff
@@ -524,86 +525,139 @@ impl Display for ReadStatus {
...
+        let create_history_filecreate_history_file = |path: &PathBuf| -> Result<()> {
+            if !path.exists() {
+                if let Some(parent) = path.parent() {
+                    if !parent.exists() {
+                        std::fs::create_dir_all(parent)
+                            .into_diagnostic()
+                            .wrap_err("Failed to create history directory for REPL")?;
+                    }
+                }
+                std::fs::File::create(path)
```
SanjoDeundiak: Lets restrict permissions on that file to `600`

implementations/rust/ockam/ockam_command/src/base_command.rs
```diff
@@ -524,86 +525,139 @@ impl Display for ReadStatus {
...
+                break;
+            };
+            let readline = rl.readline("> ");
+            match readline {
+                Ok(line) => {
+                    if is_quit_sequence(&line) {
                         break;
                     }
+                    rl.add_history_entry(line.as_str()).into_diagnostic()?;
+                    let _ = lines_s.blocking_send(line);
```
SanjoDeundiak: Can possibly be async as well

implementations/rust/ockam/ockam_command/src/base_command.rs
```diff
@@ -524,86 +525,139 @@ impl Display for ReadStatus {
...
-            let mut read_buffer = String::new();
+#[derive(Hinter, Completer, Highlighter)]
+struct MultiLineValidatorHelper {}
+impl Validator for MultiLineValidatorHelper {
+    fn validate(
+        &self,
+        ctx: &mut ValidationContext,
+    ) -> std::result::Result<ValidationResult, ReadlineError> {
+        let input = ctx.input();
+        if input.starts_with("\"\"\"") {
```
SanjoDeundiak: If we match the ending with `"\n\"\"\""`, should we then match the beginning with `"\"\"\"\n"`?

implementations/rust/ockam/ockam_command/src/base_command.rs
polvorin: we need only for the repl,  and I'm assume we have only 1 repl per zone.

implementations/rust/ockam/ockam_command/src/base_command.rs
polvorin: the input line is taken over by the library.  We do receive that from server still (we could remove if not using nc/telnet),  but what is displayed is whatever the prompt is here.

implementations/rust/ockam/ockam_command/src/base_command.rs
polvorin: it's done by the library already (if it has a different permission, it set it to 600)

SanjoDeundiak (approved):

PR #9011, opened 2025-05-18
feat(rust): use single-connection sqlite by default

SanjoDeundiak:


implementations/rust/ockam/ockam_api/src/cli_state/storage/spaces_repository_sql.rs
```diff
@@ -176,7 +167,11 @@ impl SpacesRepository for SpacesSqlxDatabase {
                 space.users = users;
 
                 // retrieve the subscription
-                space.subscription = self.query_subscription(&space.id).await?;
+                // FIXME: Copy&paste
```
etorreborre, line 170: Why is this necessary for this PR?

SanjoDeundiak: Because it can't work if max number of db connections is 1

implementations/rust/ockam/ockam_node/src/storage/database/sqlx_database.rs
```diff
@@ -308,67 +311,110 @@ impl SqlxDatabase {
...
-            }
-            | DatabaseConfigurationMode::SqliteInMemory { single_connection } => {
-                if *single_connection {
-                    1
-                } else {
-                    MAX_POOL_SIZE
-                }
-            }
-            _ => MAX_POOL_SIZE,
+        let max_pool_size = if configuration.is_single_connection() {
```
etorreborre, line 317: I think it would be better to get the pool size from the environment variable directly when creating the configuration and not get it from this part of the code.

etorreborre (approved):
Let's try this version for now.

PR #9067, opened 2025-06-03
test: add python bats tests

adrianbenavides:
Also adds some logic to override the host:port for the http/repl servers using env vars:
- repl: DEFAULT_HOST_REPL, DEFAULT_PORT_REPL
- http: DEFAULT_HOST_HTTP, DEFAULT_PORT_HTTP


implementations/python/python/ockam/agents/repl.py
```diff
@@ -1,16 +1,22 @@
...
 import re
+import os
 
 from ockam.nodes.message import StreamedConversationSnippet
+from ockam.agents.socket_address import parse_host_and_port
 
-HOST = "127.0.0.1"
-PORT = 7000
+DEFAULT_HOST = os.environ.get("DEFAULT_HOST_REPL", "127.0.0.1")
+DEFAULT_PORT = int(os.environ.get("DEFAULT_PORT_REPL", "7000"))
```
etorreborre, line 16: Same default port as HTTP?

implementations/python/python/ockam/agents/http.py
```diff
@@ -8,24 +8,25 @@
...
 from ..ockam_in_rust_for_python import info, error
 
 
 """
     This class starts an HTTP server allowing a user to interact with a node and its agents.
 """
 
-HOST = "0.0.0.0"
-PORT = 8000
+DEFAULT_HOST = os.environ.get("DEFAULT_HOST_HTTP", "0.0.0.0")
```
etorreborre, line 22: Should it be `0.0.0.0` or `127.0.0.1`?

etorreborre: Actually I think it needs to be `0.0.0.0` in containers.

implementations/python/python/ockam/agents/repl.py
adrianbenavides: copy-paste error. Fixing!

implementations/python/python/ockam/agents/http.py
adrianbenavides: I found strange that we have 0.0.0.0 in the http and 127.0.0.1 in the repl, but I didn't want to touch anything here

implementations/python/examples/01.py
```diff
@@ -26,6 +26,7 @@ async def main(node):
 
     msg = await receiver.receive(timeout=1)
     info(f"Received: {msg}")
+    print(msg)
```
etorreborre: I was hoping that we would avoid using `print` and just use `info` to output the result of the examples.
Could we set the log level at INFO for bats tests and runs assertions on the INFO messages?

adrianbenavides: That wasn't possible. For some reason the bats library was doing weird things with the output. I'll disable these local tests for now and revert the `print` changes. Maybe there is another way to process the logs in the bats tests

etorreborre (approved):

PR #9083, opened 2025-06-04
feat: in "zone init", set the working directory to the template directory

adrianbenavides:
The command will create the template directory if needed and set the working directory to that directory.

This removes the need for the user to cd into an empty directory.


etorreborre (approved):
🌴

PR #9097, opened 2025-06-06
feat: interleave planning and execution steps

etorreborre:
This PR properly displays planning and execution steps when they are interleaved:

 - Example 18 can be used locally for development.
 - Example 008 is for demos.

### DIFF
--- a/implementations/python/examples/18/main.py
+++ b/implementations/python/examples/18/main.py
@@ -1,4 +1,4 @@
-from ockam import Agent, Node, CoTPlanner, Model
+from ockam import Agent, Node, CoTPlanner, ReActPlanner, Model, HttpServer
 
 """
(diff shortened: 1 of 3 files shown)


implementations/python/examples/18/main.py
```diff
@@ -12,9 +12,8 @@ async def main(node):
         node=node,
         name="henry",
         instructions="You are an assistant who solves complex tasks by planning them carefully before solving them.",
-        planner=CoTPlanner(model=Model(name="deepseek-r1")),
+        planner=ReActPlanner(model=Model(name="llama3.2")),
```
davide-baldo, line 15: A typical example would work better with deepseek, maybe we can leave the default model?

davide-baldo (approved):

PR #9100, opened 2025-06-06
fix add planner and mcp streaming capabilities

davide-baldo:


examples/007/images/main/index.html
```diff
@@ -55,38 +55,42 @@
...
+
       async function send(message) {
         const input = document.getElementById("in");
         input.disabled = true;
 
         const out = document.getElementById("out");
         out.textContent = "";
 
-        const response = await fetch("/agents/henry?timeout=300", {
+        const response = await fetch("/agents/henry?stream=true", {
```
etorreborre, line 74: Are you sure that the timeout is not needed anymore?

implementations/python/python/ockam/models/model.py
```diff
@@ -156,41 +156,13 @@ def support_tools(self):
...
-                if "tool_calls" in message:
-                    if len(message["tool_calls"]) == 0:
-                        del message["tool_calls"]
-        else:
-            for message in messages:
-                # convert any tool role to assistant
-                if message["role"] == "tool":
-                    message["role"] = "assistant"
+    def support_forced_assistant_answer(self):
+        if "bedrock" in self.name or "litellm_proxy" in self.name:
```
etorreborre: `return self.name not in ["bedrock", "litellm_proxy"]` should work here

implementations/python/tests/test_normalize_messages.py
```diff
@@ -0,0 +1,167 @@
+from ockam.models.model import normalize_messages
```
etorreborre: So far we have been the unit test file close to the functions it's testing, cf `tool_test`.

etorreborre (dismissed):
🎸

examples/007/images/main/index.html
davide-baldo: In streaming mode, the timeout refers to the time between pieces, the default 60s should be more than enough to reach the first token
(never saw a timeout so far)

implementations/python/python/ockam/models/model.py
davide-baldo: Unfortunately, that syntax is not working, but I've simplified with:

```        return "bedrock" not in self.name and "litellm_proxy" not in self.name```

etorreborre (approved):

PR #9104, opened 2025-06-09
feat: display the url of the logs server

etorreborre:
This PR tells the user where to find the zone logs

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/zone/repl.rs
+++ b/implementations/rust/ockam/ockam_command/src/zone/repl.rs
@@ -1,11 +1,13 @@
 use crate::cluster::common_args::HttpApiArgs;
 use crate::entry_point::RUNTIME;
+use crate::node_command::InMemoryNodeCommand;
 use crate::util::parsers::hostname_parser;
 use crate::util::port_is_free_guard;
 use crate::zone::common_args::{
     EnrollmentTicketConfigArg, ZoneConfigArg, ZoneInletsArgs, ZoneNameOrConfigArg,
 };
 use crate::zone::ctrlc::ZoneCtrlcHandler;
+use crate::zone::get_cluster_name::GetClusterName;
 use crate::zone::zone_config::{Outlet, ZoneConfig};
 use crate::{docs, Command, CommandGlobalOpts, Result};
 use async_trait::async_trait;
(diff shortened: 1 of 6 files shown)


implementations/rust/ockam/ockam_command/src/zone/repl.rs
```diff
@@ -91,6 +93,13 @@ impl ReplCommand {
                 } else {
                     "the portal"
                 };
+                if let Some(logs_address) = logs_address {
+                    opts.terminal.write_line(fmt_log!(
+                        "The zone logs are accessible at http://localhost:{}/",
```
adrianbenavides: For consistency, we should use color_primary for the `http://localhost:{}` url

adrianbenavides: Do we want to show this url only when there is no repl?

etorreborre: Fixed

implementations/rust/ockam/ockam_command/src/zone/repl.rs
```diff
@@ -78,9 +79,18 @@ impl ReplCommand {
...
-            let (executors, repl_address) = self.zone_inlets(&opts, &zone_config).await?;
-            // let (repl_data, rest_inlet_handles) = cmd.dummy_inlet(&opts).await?;
+            let (executors, logs_address, repl_address) =
+                self.zone_inlets(&opts, &zone_config).await?;
+
             opts.terminal.write_line(fmt_separator!())?;
+            if let Some(logs_address) = logs_address {
+                opts.terminal.write_line(fmt_log!(
+                    "Open your browser at: {} to access the zone logs",
+                    color_primary(format!("http://localhost:{}/", logs_address.port()))
```
adrianbenavides: I was checking the [http](https://github.com/build-trust/ockam/pull/9072/files#diff-5650b32e142d51fb1568f6631b7e76fc730556f167ac0d3c10be130db416838fR558-R559) message and they should have a similar structure so they look good when we show both.

etorreborre: I changed the formatting but maybe we should regroup both messages at the same place. What do you think?

adrianbenavides: Yeah, makes sense to me

etorreborre: I made the changes.

adrianbenavides (approved):

PR #9138, opened 2025-06-17
feat: more logging for agents

etorreborre:
This PR:

- Adds a few more logs for: `Model`, `Tool`
- Enables the logging of ockam Rust components.
- Allows the user to specify a string setting the logging configuration from inside the `main` code.

For example:

```
set_log_levels("agent=debug") 
```

Will set the logs of the `agent` actions to the debug level, whereas:
``` 
set_log_levels("debug,agent=info")
```
Will set the logs of all the python modules to `debug` but the ones from `agent` to `info`. 

See some other variations in `logging_test.py`.

### Notes

 - The loggers are loaded lazily when their corresponding class is accessed for the first time (not just loaded).
 - The convention I've used is:
   - At the `INFO` level: "do action", for example "start agent 'x'"
   - At the `DEBUG` level: "the action is done" or "the action is done with result: y"

### DIFF
--- a/implementations/python/python/ockam/agents/agent.py
+++ b/implementations/python/python/ockam/agents/agent.py
@@ -32,13 +32,7 @@ from ..nodes.message import (
 )
 from .names import validate_name
 
-from ..ockam_in_rust_for_python import info, warn, debug
-
-from ..logging.logging import get_logging_config
-import logging.config
-
-logging.config.dictConfig(get_logging_config())
-logger = logging.getLogger("agent")
+from ..ockam_in_rust_for_python import warn
 
 
 class AgentState(Enum):
@@ -258,7 +252,18 @@ class AgentStateMachine:
 
 
 class Agent:
+    _logger = None
     tools: Dict[str, InvokableTool]
+
+    @classmethod
+    def logger(cls):
+        if cls._logger:
+            return cls._logger
+        else:
+            from ..logging.logging import get_logger
+
+            cls._logger = get_logger("agent")
+            return cls._logger
 
     def __init__(
         self,
@@ -273,7 +278,8 @@ class Agent:
         knowledge: KnowledgeProvider,
         max_knowledge_size: int,
     ):
-        logger.info("starting agent")
+        self.logger = Agent.logger()
+        self.logger.info(f"start agent '{name}'")
         self.node = node
 
         self.tools = tools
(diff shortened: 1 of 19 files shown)


implementations/python/python/ockam/agents/agent.py
```diff
@@ -349,7 +354,7 @@ async def start(
         node: NodeProtocol,
         instructions: str,
         name: Optional[str] = None,
-        model: Model = Model(name="llama3.2"),
+        model: Model = None,
```
davide-baldo, line 530: Why remove the explicit model declaration as the default parameter?

davide-baldo (dismissed):
A huge step up in logging!

implementations/python/python/ockam/logging/colored_formatter.py
```diff
@@ -12,13 +12,19 @@ def __init__(self, *args, **kwargs):
...
-        record.asctime = f"{self.GREY}{original_asctime}{self.RESET}"
-        record.name = f"{self.GREY}{record.name}{self.RESET}"
-        return super().formatMessage(record)
+        try:
+            original_asctime = self.formatTime(record, self.datefmt)
+            record.asctime = f"{self.GREY}{original_asctime}{self.RESET}"
+            record.name = f"{self.GREY}{record.name}{self.RESET}"
+            return super().formatMessage(record)
+        except Exception:
+            return ""
```
adrianbenavides, line 30: Wouldn't it be better if we returned the original record if something fails?

implementations/python/python/ockam/agents/repl.py
```diff
@@ -6,20 +6,28 @@
...
+        else:
+            from ..logging.logging import get_logger
+
+            cls._logger = get_logger("repl")
+            return cls._logger
+
     def __init__(
         self, agent_reference, listen_address=f"{DEFAULT_HOST}:{DEFAULT_PORT}", functions=None, timeout=120, stream=True
     ):
+        self.logger = Repl.logger()
```
adrianbenavides, line 30: I don't get this pattern. If I understand correctly, `_logger` is never used, as it's tied to an instance, but it's being used as if it were tied to a global "class" instance.

I'd simplify the `logger` method to return `get_logger("repl")`.

mrinalwadhwa (dismissed):

mrinalwadhwa (approved):

PR #9140, opened 2025-06-17
agent as a state machine

davide-baldo:
Refactored the agent as a state machine, fixing also the issue with streaming + tool + planning.
To make the diagram pass CI, I had to disable `eclint`, but I made sure CI is failing when linting or formatting are not passing

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -49,5 +49,5 @@ end_of_line = unset
 insert_final_newline = unset
 
 [*.py]
-indent_style = space
-indent_size = 4
+indent_style = unset
+indent_size = unset
--- a/implementations/python/Makefile
+++ b/implementations/python/Makefile
@@ -9,7 +9,7 @@ test: build
 
 lint:
 	uv run pylint python examples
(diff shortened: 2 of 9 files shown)


SanjoDeundiak (dismissed):

.editorconfig
```diff
@@ -49,5 +49,5 @@ end_of_line = unset
 insert_final_newline = unset
 
 [*.py]
-indent_style = space
```
SanjoDeundiak, line 52: Why is that change needed?

SanjoDeundiak (approved):

.editorconfig
davide-baldo: Otherwise documentation needs to be aligned to 4 spaces, specifically the diagram was not passing.
But by adding `--exit-non-zero-on-format` to `uv run ruff format examples python tests` the CI fails if the code indentation is not 4

SanjoDeundiak (dismissed):

adrianbenavides (approved):

ckhrysze (approved):

PR #9141, opened 2025-06-17
Redesign memory

SanjoDeundiak:


examples/013/images/main/main.py
```diff
@@ -0,0 +1,21 @@
...
+    agent = await Agent.start(
+        node=node,
+        name="assistant",
+        instructions="You are an assistant who acts like a friend and can maintain the conversation on any topic.",
+        model=model,
+        memory_model=memory_model,
+        memory_embeddings_model=memory_embeddings_model,
+    )
+
+    await Repl.start(agent, "localhost:7000")
```
davide-baldo: is the example using a repl or the html?

implementations/python/pyproject.toml
```diff
@@ -15,7 +15,9 @@ dependencies = [
   "dill>=0.4.0",
   "docstring-parser>=0.16",
   "fastapi>=0.103.1",
-  "litellm @ git+https://github.com/build-trust/litellm.git@v1.72.2",
+  "litellm @ git+https://github.com/build-trust/litellm.git@ba33a4d4b165c276f7eb6a681fbe1ba23466be2f",
+  "mem0ai @ git+https://github.com/build-trust/mem0.git@2f861da0bbd515db1afd0b9a6c79d39273ac0eff",
+  "ollama==0.5.1",
```
davide-baldo: why an explicit ollama dependency?

implementations/python/python/ockam/agents/agent.py
```diff
@@ -359,13 +375,13 @@ async def handle__conversation_snippet(
         messages = snippet.messages
 
         query = ""
-        for message in messages:
+        for message in reversed(messages):
```
davide-baldo, line 375: why did you reverse? shouldn't we search the most recent message?

implementations/python/python/ockam/agents/agent.py
```diff
@@ -388,55 +404,108 @@ async def get_messages_only(self, conversation, scope) -> list[ConversationMessa
...
+        while True:
+            self.memory_knowledge = await AgentMemoryKnowledge.create(self.memory_model, self.memory_embeddings_model)
+            if skipped_message_history:
+                await self.memory_knowledge.add(
+                    scope=scope, conversation=conversation, messages=skipped_message_history
+                )
+
+            # We should actually clear knowledge in this function each time...
+            memory_knowledge = await self.memory_knowledge.search(scope, conversation, query)
+            warning(f"RELATED MEMORY for query: {query}:\n{memory_knowledge}.")
```
davide-baldo: debug logs

implementations/python/python/ockam/agents/agent.py
```diff
@@ -582,12 +674,30 @@ async def start_many(
                 for i in range(number_of_agents):
                     name = secrets.token_hex(12)
                     await Agent.start_agent_impl(
-                        node, instructions, name, model, tools, planner, None, knowledge, max_knowledge_size
+                        node,
+                        instructions,
+                        name,
+                        copy.deepcopy(model),
```
davide-baldo: models and tools should be stateless

implementations/python/python/ockam/agents/agent_memory_knowledge.py
```diff
@@ -0,0 +1,39 @@
...
+
+    async def add(self, scope: Optional[str], conversation: Optional[str], messages):
+        new_messages = messages
+
+        # for message in messages:
+        #     # FIXME: Far from perfect, messages should be assigned unique ids
+        #     if {"scope": scope, "conversation": conversation, "message": message} not in self.added_messages:
+        #         new_messages.append(message)
+
+        warning(f"ADDING MESSAGES: {new_messages}")
```
davide-baldo: debug logging

implementations/python/python/ockam/knowledge/mem0.py
```diff
@@ -0,0 +1,72 @@
...
+
+from litellm import Router
+
+
+def local_config(model: Model, embed_model: Model):
+    return {
+        "vector_store": {
+            "provider": "inmemory",
+            "config": {
+                "embedding_model_dims": 768,
```
davide-baldo: the size of the vector depends on the model that generates the embedding, it's usually around the range of 200~4000

implementations/python/pyproject.toml
SanjoDeundiak: It was required by mem0 to call ollama hosted llm, even though mem0 didn't depend on it. However, I think it's no longer required after I switched to using ockam Model. I'll try removing that dependency. Nice finding

examples/013/images/main/main.py
SanjoDeundiak: Repl doesn't behave as it should unfortunately. Html works fine. I'll disable the Repl

implementations/python/python/ockam/agents/agent.py
SanjoDeundiak: Don't messages start from older?

davide-baldo: yes, they do, and the "last" message being processed is the most recent one, by reversing the array the last messages would be the first one

SanjoDeundiak: So does my code perform as expected here? It looks for the last user message (which is the most recent) and exits the loop

implementations/python/python/ockam/knowledge/mem0.py
SanjoDeundiak: Good to know. I'll see if I can figure out the right number here

implementations/python/python/ockam/agents/agent.py
SanjoDeundiak: I would actually leave all of them for now. They should not be printed unless you specify manually max_tokens in the Model.

implementations/python/python/ockam/agents/agent.py
SanjoDeundiak: Do you think deepcopy hurts in that case?

davide-baldo: when creating 100k agents, yes, it's probably hurting performance

implementations/python/python/ockam/knowledge/mem0.py
davide-baldo: ideally, we should omit the vector size completely

davide-baldo (dismissed):

implementations/python/python/ockam/knowledge/mem0.py
SanjoDeundiak: We can omit the dimensions value when creating a vector column, but that's suboptimal? As a benefit though, we wouldn't need to create separate tables for models with different dimensions value

davide-baldo: > but that's suboptimal?

yes, you can't do full indexing on the field anymore, only partial ones, but indexing with vectors is semantically different compared to traditional indexing, and we probably don't need it

davide-baldo (dismissed):

etorreborre (approved):

PR #9148, opened 2025-06-19
feat: add `identity import/export` commands

adrianbenavides:
```
exported=$(ockam identity export <name>)
OCKAM_HOME=/tmp/dir ockam identity import $exported
```

### DIFF
--- a/implementations/rust/ockam/ockam_vault/src/traits/vault_for_signing.rs
+++ b/implementations/rust/ockam/ockam_vault/src/traits/vault_for_signing.rs
@@ -1,4 +1,7 @@
-use crate::{Signature, SigningKeyType, SigningSecretKeyHandle, VerifyingPublicKey};
+use crate::{
+    Signature, SigningKeyType, SigningSecret, SigningSecretKeyHandle, VaultError,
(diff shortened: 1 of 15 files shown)


implementations/rust/ockam/ockam_vault/src/traits/vault_for_signing.rs
```diff
@@ -35,4 +35,13 @@ pub trait VaultForSigning: Send + Sync + 'static {
         &self,
         signing_secret_key_handle: SigningSecretKeyHandle,
     ) -> Result<bool>;
+
+    /// Get a stored secret by its handle.
+    async fn export_key(
```
SanjoDeundiak, line 43: Should these 2 have a default impl returning an error?

SanjoDeundiak (dismissed):

implementations/rust/ockam/ockam_vault/src/traits/vault_for_signing.rs
adrianbenavides: Done!

SanjoDeundiak (approved):

PR #9151, opened 2025-06-20
feat: substitute env var in zone name field in zone config

adrianbenavides:
Allow defining the zone name from an env var:

```yaml
name: "prefix${NAME}" # or "prefix$NAME"
pods:
  ...
```

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/zone/zone_config.rs
+++ b/implementations/rust/ockam/ockam_command/src/zone/zone_config.rs
@@ -7,7 +7,7 @@ use std::str::FromStr;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ZoneConfig {
-    #[serde(alias = "zone_name")]
+    #[serde(alias = "zone_name", deserialize_with = "deserialize_with_env")]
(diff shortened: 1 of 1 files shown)


implementations/rust/ockam/ockam_command/src/zone/zone_config.rs
```diff
@@ -413,14 +424,25 @@ impl Outlet {
     }
 }
 
+fn deserialize_with_env<'de, D>(deserializer: D) -> Result<String, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    let s = shellexpand::env(&s).map(|c| c.to_string()).map_err(|e| {
```
davide-baldo, line 432: `shellexpand` looks cool!

davide-baldo (approved):

implementations/rust/ockam/ockam_command/src/zone/zone_config.rs
adrianbenavides: Yeah it works great, we also use it in the `node` config.

PR #9153, opened 2025-06-23
feat: add the possibility to restrict the http access with an api key

etorreborre:
This PR adds the possibility to specify an API key and make it mandatory to access the `HttpServer` routes.

By default the API key is accessed by the server with an environment variable named `API_KEY`, and the clients must specify the key in a `X-API-Key` header.

Another option (which might be less secure) is to pass the API key as a query parameter named `api_key` by default.

Usage example:
```python
Node.start(main, http_server=HttpServer(api=Api(), api_key=ApiKey()))

# or 

Node.start(main, http_server=HttpServer(api=Api(), api_key=ApiKey(api_key_env_name="ACME_API_KEY", 
  api_key_parameter_name="acme_api_key")))
```


implementations/python/python/ockam/agents/http.py
```diff
@@ -34,18 +33,28 @@ def logger(cls):
...
+    def get_node(self):
+        return self.node
+
+    def __init__(self, listen_address=f"{DEFAULT_HOST}:{DEFAULT_PORT}", api=None, api_key: ApiKey = None):
+        """
+        Start the HTTP server at the specified address.
+
+        - A custom API can be provided to extend the functionality of the server.
+        - An API key can be set via an environment variable to secure the endpoints.
+        """
```
mrinalwadhwa: let's try to make this comment more `active voice`

SanjoDeundiak (approved):

PR #9155, opened 2025-06-24
feat: in `zone inlet`, try to get `from` and `pod` args from config if not passed

adrianbenavides:
We can omit the `from` and `pod` args when:
- Running the command from a directory that contains the target ockam.yaml file
- Targeting a known inlet/outlet (logs, http, repl)

```
# before
ockam zone inlet --to logs --pod logs-pod --from localhost:3000

# now
ockam zone inlet --to logs
```


etorreborre (dismissed):
🖌️

etorreborre (approved):

PR #9157, opened 2025-06-24
feat: normalize start / end log messages

etorreborre:
This PR improves a number of log messages. Here is an example of a simple output where we send a message to an agent:

image

Notice that:

 - There's a start/end message for each action.
 - The messages all start with a capital letter.
 - The `WARNING` level from Python is displayed as `WARN` (with the same color that would be used by the rust code).

### DIFF
--- a/implementations/python/python/ockam/logging/logging.py
+++ b/implementations/python/python/ockam/logging/logging.py
@@ -11,6 +11,14 @@ FORMAT = DEFAULT_LOG_FORMAT + (" [%(pathname)s:%(lineno)d]" if os.getenv("OCKAM_
 
 LOG_LEVELS = {}
 
+LEVELS: dict[str, int] = {
+    "critical": logging.CRITICAL,
+    "error": logging.ERROR,
+    "warning": logging.WARNING,
+    "info": logging.INFO,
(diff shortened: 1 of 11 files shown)


mrinalwadhwa (dismissed):

implementations/python/python/ockam/logging/logging.py
```diff
@@ -141,3 +144,31 @@ def debug(msg, *args, **kwargs):
...
+            yield
+        except Exception as e:
+            self.logger.info(f"{e}")
+            return
+        self.logger.info(after_msg)
+
+
+class DebugContext(LoggerAware):
+    @contextmanager
+    def info(self, before_msg, after_msg):
```
adrianbenavides: Why is this function called `info`? I'd expect it to be `debug`

etorreborre: Thanks, this is fixed now

adrianbenavides (dismissed):

adrianbenavides (dismissed):

adrianbenavides (dismissed):

adrianbenavides (dismissed):

adrianbenavides (approved):

PR #9158, opened 2025-06-25
test: copy failed tests logs into ockam directory

adrianbenavides:
They are currently stored under `$HOME/.bats-tests`, and the CI job is not uploading that directory in the job's artifacts.


etorreborre (approved):
🧪

PR #9167, opened 2025-06-26
fix: store user when importing an identity

adrianbenavides:
- `identiy import` is backwards compatible with an exported identity that only contains the user email
- The exported identity now includes the user info, which is necessary to retrieve the associated projects
- Fixes the `status` command to retrieve spaces and projects from the controller if possible, and falls back to reading the data from the database if not. Also, filters out the current temporary node from the shown list of nodes


etorreborre (approved):
🎲

PR #9168, opened 2025-06-26
feat: memory + knowledge persistence

etorreborre:
This PR adds some Postgres persistence for:

- The conversations used by the `memory` module.
- The documents handled by the `knowledge` module.
- The conversations handled by the `mem0` module.

The configuration depends on the following environment variables (which are automatically set-up for a deployed pod):

 - `OCKAM_DATABASE_INSTANCE` for example `localhost:5432/postgres`.
 - `OCKAM_DATABASE_USER` for example `postgres` (this is used as the `tenant_id`).
 - `OCKAM_DATABASE_PASSWORD` for example `postgres`.

Notes:

 - The PR relies on a specific `mem0` commit to add a `tenant_id` column for the `mem0_conversation` table.
 - There will possibly some refactoring to do later on to share the database configuration / initialization.
 - All the tables have been created in the RDS database.


implementations/python/python/ockam/memory/memory.py
```diff
@@ -1,12 +1,58 @@
...
+
         self.instructions = []
         self.conversations = defaultdict(lambda: defaultdict(list[dict]))
-        self.lock = threading.RLock()
+
+        self.initialize_database()
+        self.load_conversations()
+
+    def initialize_database(self):
+        db_instance = environ.get("OCKAM_DATABASE_INSTANCE")
```
davide-baldo, line 39: We should use always the same environment variables to load the database, maybe we could abstract and re-use this logic?
https://github.com/build-trust/ockam/blob/develop/implementations/python/python/ockam/knowledge/database.py#L93-L137

davide-baldo (commented):
I was expecting knowledge database class to reflect the new schema, but it's not

implementations/python/python/ockam/memory/memory.py
etorreborre: Yes but I was thinking about doing this refactoring after your refactoring branch has been merged.

davide-baldo (approved):

PR #9172, opened 2025-06-26
Thinking is normalized and exposed

davide-baldo:
`<think>` tag normalization responsibility has been moved to the model, the calling code will always expect `reasoning_content`.
The exposed field is called `thinking`, and it's a boolean that can be present in any message.

It's possible to specify `CACHE_INFERENCE` environment variable to enable inference cache, this is meant to help developers and should not be used in production.

Agent has now `test_full_flow` and `test_full_flow_streaming`, that test the majority of agent flow except memory.

### DIFF
--- a/implementations/python/examples/08.py
+++ b/implementations/python/examples/08.py
@@ -12,8 +12,8 @@ async def main(node):
         node=node,
         name="Assistant",
         instructions="Assistant to solve some complex task ...",
-        model=Model(name="ollama_chat/llama3.2"),
-        planner=CoTPlanner(),
+        model=Model(name="deepseek-r1"),
+        planner=CoTPlanner(model=Model(name="deepseek-r1")),
     )
 
     async for chunk in agent.send_stream("Estimate how many violins are in the world", timeout=60 * 20):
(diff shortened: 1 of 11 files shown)


implementations/python/examples/08.py
```diff
@@ -12,23 +12,29 @@ async def main(node):
...
-        if phase == Phase.THINKING:
-            print(f"\033[94m{text}\033[0m", end="")  # thinking: blue
-        elif phase == Phase.PLANNING:
-            print(f"\033[92m{text}\033[0m", end="")  # planning: green
-        elif phase == Phase.EXECUTING:
-            print(f"\033[90m{text}\033[0m", end="")  # executing: grey
+        # reasoning is always in italic
+        if phase == Phase.PLANNING:  # planning: green
+            if thinking:
+                print(f"\033[3;92m{text}\033[0m", end="")
```
etorreborre, line 30: We should use `info` instead of `print` because this allows us to control the output centrally if necessary.

implementations/python/python/ockam/models/model.py
```diff
@@ -290,33 +302,147 @@ def support_tools(self):
...
-        self.logger.debug(f"Got a response from model '{self.original_name}': {response}")
+        if stream:
+            return self._complete_chat_stream(messages, is_thinking, **kwargs)
+        else:
+            return await self._complete_chat(messages, **kwargs)
+
+    async def _complete_chat_stream(self, messages: List[dict], is_thinking: bool, **kwargs):
+        chunks = None
+
+        if _cache_inference:
```
etorreborre, line 334: Is there an opportunity to refactor with L.380 and maybe L.348 + L.392 / L.410 + L.420?

implementations/python/python/ockam/agents/agent_test.py
```diff
@@ -1,8 +1,311 @@
+from copy import deepcopy
```
etorreborre, line 1: Great test!

etorreborre (dismissed):

implementations/python/examples/08.py
davide-baldo: This is printing a piece of text in streaming mode, using logging would completely break the output

implementations/python/python/ockam/models/model.py
davide-baldo: I did deduplicate a little bit, but couldn't do much

SanjoDeundiak (approved):

etorreborre (approved):

PR #9183, opened 2025-06-30
fix: check for ockam processes before continuing installation

adrianbenavides:
When an ockam process is detected, the user is prompted to decide whether to kill the process or abort the installation:

image


adrianbenavides:
Fixed with a different approach at https://github.com/build-trust/ockam/pull/9184

PR #9193, opened 2025-07-01
feat: support setting zone config and secret config as inline strings

adrianbenavides:
Additionally, the zone config can be set via `OCKAM_ZONE_CONFIG` and the secrets config via `OCKAM_ZONE_SECRETS`.

### DIFF
--- a/implementations/rust/ockam/ockam_command/src/zone/common_args.rs
+++ b/implementations/rust/ockam/ockam_command/src/zone/common_args.rs
@@ -144,11 +144,55 @@ impl ZoneNameLongOrConfigArg {
 
 #[derive(Clone, Debug, Args, Default)]
 pub struct SecretsConfigArg {
-    /// The path to the secrets file, in yaml or json format.
+    /// The path to the secrets file, in yaml or json format, or the inline content of the secrets.
     /// If not set, the `./secrets.yaml` file from the current directory will be used.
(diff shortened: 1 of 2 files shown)


implementations/rust/ockam/ockam_command/src/zone/common_args.rs
```diff
@@ -143,13 +144,57 @@ impl ZoneNameLongOrConfigArg {
...
+                    Ok(None)
+                }
+            }
+        };
+
+        // Try to read as file first, if it fails treat as inline content
+        let contents = if std::path::Path::new(config_input).exists() {
+            std::fs::read_to_string(config_input)
+                .into_diagnostic()
+                .wrap_err(format!("Failed to read secrets file at {}", config_input))?
```
etorreborre, line 189: "Failed to read the secrets definition" maybe since this is not a file in that case?

etorreborre (approved):

implementations/rust/ockam/ockam_command/src/zone/common_args.rs
adrianbenavides: Correct, I'll fix it.

adrianbenavides: Actually, in that branch, we are trying to read the contents from a file explicitly (we check that the path exists right before doing it), so the message is correct.

